PR #193, opened 2014-03-18
Mysql adapter

jquadrin:
TODO 
- [ ] Poolboy integration
- [ ] Transaction support
- [ ] Custom functions test failing


rrouse:
@jquadrin What is your definition of a stable release? I've been wanting to use Ecto but my current database engine is MySQL, so I've stuck with Rails for now.

I noticed mysql-otp release 0.8.0 today. Are we waiting for 1.0?

kotedo:
You could use any database as long as you stick to the implementation details of the database/database adapter.

It's a convenience to have it the Ecto-way. It's erlang nonetheless, so you're still a lot better off than ruby + any ORM

Just my $0.02

Lazyshot:
At the time of his mention of that library it was maybe 2 weeks into
development.

Not sure how stable 0.8.0 is or of any production environments, but it's
still very young. Especially in comparison to emysql, which is 4 years old.

liveforeverx:
I've started elixir implementation of [mysql adaptor](https://github.com/liveforeverx/mariaex) for the same reasons(clear separation and so on), but one more, much more easy integration in ecto too(and, that, why API of an adaptor is very close to postgrex and not to mnesia, for example), only prepared statements are not implemented by me in difference to https://github.com/mysql-otp/mysql-otp, but it should be very easy to do.

That means, that it will be possible on some very close future to interchange the postgrex to mariaex (add mysql adaptor) and it will be mostly working without other changes(or the integration will be very-very easy). 

At the moment, for next week or some weeks, I've another focus and have not so much time to develop it further, but in my personal opinion, it can fit ecto little bit better, as mysql-otp, because of integration will be easier.

True is, that both libraries are young in comparison to emysql, but it is better to integrate something with good architecture, and make it stable/production ready, as to use something with worst architecture from ground up, but it is in my opinion.

jquadrin:
I dont see a problem with implementing mysql-otp and ecto

josevalim:
Hello folks,

Ecto master makes adding SQL adapters with tests much, much easier, so I am closing this pull request (as it is outdated) but a new pull request that adds MySQL support on master is very welcome.

To give an idea on how it should look like. Here is the full postgres implementation:

* https://github.com/elixir-lang/ecto/blob/master/lib/ecto/adapters/postgres.ex
* https://github.com/elixir-lang/ecto/blob/master/lib/ecto/adapters/postgres/connection.ex

`Ecto.Adapters.SQL` handles everything from connection pooling up to transactions. I have also refactored the integration tests so they can be reused across multiple adapters:

* https://github.com/elixir-lang/ecto/tree/master/integration_test/pg

Keep in mind there also some unit tests per adapter too that would need to be ported:

* https://github.com/elixir-lang/ecto/blob/master/test/ecto/adapters/postgres_test.exs

I hope this will make the adapters easier to write and consistent between them, so a new pull request is definitely welcome! Thanks!

josevalim:
@jquadrin I believe much of your initial work can be ported. Let us know if you have any question or if there is something we can do to help!

PR #261, opened 2014-08-19
[WIP] Ecto.Model.assign

MSch:
I've started working on `Ecto.Model.assign` as described in https://github.com/elixir-lang/ecto/issues/258#issuecomment-52421545 and https://github.com/elixir-lang/ecto/issues/246

Currently open questions:
- `assign` should accept both a struct and a module right?

  ```elixir
  assign(%User{name: "martin"}, %{name: "not martin"})
  assign(User, %{name: "not martin"})
  ```

- `assign` should probably be generated by the `Ecto.Model.schema` macro and not use `model.__schema__(:field_names)` at runtime, right?

- `__schema__(:field_names)` doesn't return virtual fields, but there's no `__schema__(:virtual_field_names)`. Is there a reason why?

### DIFF
--- a/lib/ecto/model.ex
+++ b/lib/ecto/model.ex
@@ -76,5 +76,6 @@ defmodule Ecto.Model do
   end
 
   defdelegate assign(model, values), to: Ecto.Model.Assign
+  defdelegate assign(model, values, opts), to: Ecto.Model.Assign
 end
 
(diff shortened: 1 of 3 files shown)


josevalim:
Let's put this on hold as we are revisiting those decisions on the mailing list.

josevalim:
Closing this in favor of the new changesets!

PR #296, opened 2014-10-09
add support for passing custom encoder/decoder/formatter to Adapters.Postgres

tyre:
This would ease the integration of custom encoders like [geo](https://github.com/bryanjos/geo) to extend functionality.


josevalim:
Closing this as we are changing the formater/encoder/decoder in postgrex exactly so it is easier to extend! Then hstore or postgis will be more straight-forward to add! Thank you!

PR #336, opened 2014-12-24
Updated Migration DSL

joshnuss:
This is an updated version of #132.

TODO

- [x] Add more column types (boolean, decimal, etc..), currently only primary key, varchar, datetime, integer
- [x] Add `timestamps/0` function to DSL (maybe this should be a separate PR, since the models should support this notion too)
- [x] Shortcuts for `alter table(..)` (add_column, remove_column, modify_column, rename_column)
- [x] Existence functions (column_exists?, table_exists?, index_exists?)
- [x] Column options, `default`, `scale`, `precision`, others?
- [x] Add `references/1` for creating foreign keys i.e.

```elxir
create table(:products) do
  references(:category)
end
```

It supports a DSL syntax, for creating/dropping/altering objects:

## Change Example:

```elixir
defmodule Repo.Migrations.CreatePosts do
  use Ecto.Migration

  def change do
    create table(:posts) do
      add :title, :string
      add :published_at, :time
    end
  end
end
```

## Up/Down Example:
```elixir
defmodule Repo.Migrations.CreatePosts do
  use Ecto.Migration

  def up do
    create table(:posts) do
      add :title, :string
      add :published_at, :datetime
      timestamps
    end
  end

  def down do
    drop table(:posts)
  end
end
```

## Alter Example:
```elixir
defmodule Repo.Migrations.AlterPosts do
  use Ecto.Migration

  def change do
    alter table(:posts) do
      add :summary, :string
      remove :published_at
      modify :title, :string
      rename :owner, :created_by
    end
  end
end

```


lib/ecto/migration/dsl.ex
```diff
@@ -0,0 +1,297 @@
...
+  Checks if a table exists.
+
+  ## Examples
+
+    if table_exists?(:products) do
+      drop table(:products)
+    end
+
+  """
+  def table_exists?(table_name) do
```
joshnuss: Maybe `exists? column(:products, :name)`?

joshnuss:
Note that the way the postgres adapter is checking if table/index/column exists does not take into account the schema. (will need to be fixed)

test/ecto/migration/runner_test.exs
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  defmodule MockRepo do
+    def adapter do
+      MockAdapter
+    end
+  end
+
+  setup do
+    {:ok, _} = Runner.start_link(MockRepo)
```
joshnuss: But then we put the onus on all the callers to know the pid of the migration runner right?
That would mean we need to modify the Runner & DSL to know and pass along the runner's pid `exists? pid, ...` , `execute(pid, ...)` and `direction(pid, ...)`

josevalim: Hrm, you are right. Nevermind about this for now then.

lib/ecto/migration/dsl.ex
josevalim: It could be... my concern is if people would then try to run something like `drop column(:products, :name)` instead of using remove. :S

Maybe we could allow `exists?(:column)` but only inside an create/alter table?

lib/ecto/migration/dsl.ex
```diff
@@ -0,0 +1,297 @@
...
+
+      create table(:posts) do
+        add :title, :string, default: "Untitled"
+        add :body,  :text
+
+        timestamps
+      end
+
+  """
+  defmacro create(object, do: block) do
```
joshnuss: Fixed in 5ae3aae

joshnuss:
@josevalim thanks for the comments! All those changes are now in.

Whats left:

- Integration tests are failing, can you take a look? Not sure why they are failing, but it has something to do with the Runner being a named process.
- The `object_exists?` queries use pg_attribute, pg_tables and pg_class tables, but should include the schema name (otherwise it will incorrectly report table/column/index exists if defined with the same name in another schema)
- When creating or adding columns we should support some `opts` (default, scale, precision and null come to mind), maybe best to add in another PR?

joshnuss:
Now supports options for precision, scale, default, null & size when adding and modifying columns.

joshnuss:
If the Repo is configured like this:

``` elixir
def conf do
    parse_url "ecto://postgres:postgres@localhost/ecto_simple"
end
```

Is there an easy way to get at the schema name part "ecto_simple"? or do we need to parse that out. (would like to make the object existence queries filter by the correct schema name)

josevalim:
You never have to parse that out. Calling repo.conf should already returned the parsed bits, no?

joshnuss:
Missed that, thank you. will update now

joshnuss:
OK, `object_exists?` queries now taking into account schema name. Thanks for you help on that.

Still trying to figure out the issue with integrations tests passing.

Also noticing some brittleness in the MigratorTests, its using `async: false` but one of every five runs it seems to fail.

josevalim:
> Also noticing some brittleness in the MigratorTests, its using async: false but one of every five runs it seems to fail.

When it fails, try running it with the failing --seed. It could be an ordering issue rather than a race condition one.

joshnuss:
Useful tip.

Seems it is a race condition (used a failing seed and it passes).

Could it be because of the named process? When a process is spawned during a test run and then later killed, does the kill end the process and de-register the name in a single atomic operation?

Manually killing the process in the `on_exit` seems to fix it, is that ok?

``` elixir
setup do
  {:ok, pid} = Ecto.Migration.Runner.start_link(ProcessRepo)
  
  on_exit fn ->
    Process.exit(pid, :normal)
  end
   
  :ok
end
```

joshnuss:
Spoke too soon, still has :racehorse: conditions.

josevalim:
Because you start the process with start_link, it is linked to the test
process and it will shutdown when the test finishes.

What is the error message exactly? Does it happen if you run just that file?

In any case, feel free to not worry about it, I can take a look in a branch
before merging.

-- 


josevalim:
Hello @joshnuss! I have merged this locally, I will make the tests pass, merge to master and have a new Ecto version in probably week. The work you have done here is great! Thank you!

joshnuss:
@josevalim you are most welcome, thanks for all the help!

gjaldon:
Really looking forward to this! :+1:

paulcsmith:
This looks amazing. Thanks @joshnuss

rcdilorenzo:
I hate to re-open this thread, but why did we drop support for `rename_column`? Are we planning to add it?

josevalim:
@rcdilorenzo no problem. The rename_column implementation we had simply didn't work. We will be glad to add it back if someone wants to contribute it. :)

PR #338, opened 2014-12-25
Use Logger instead of IO.puts

yrashk:
IO.puts is not the greatest thing to have in a library :smile:


josevalim:
I have pushed a fix that allows the log level to be configured as we are adding a custom log layer to migrations. That will probably work for your needs. Thanks!

PR #344, opened 2015-01-08
Allow custom encoder, decoder and formatter for the postgres adapter

yrashk:
I want to suggest this functionality. In my particular case I am using it to encode/decode PG's `json` type. Along with that, I have a custom Ecto type for JSON that just passes data through. 

The reason why I needed to implement custom encoding/decoding is that after validation, Ecto changes field's value to the one produced by dump (https://github.com/elixir-lang/ecto/blob/master/lib/ecto/repo/model.ex#L22) which isn't quite convenient. Imagine that you have a JSON field before inserting/saving and then suddenly after that that's a string.

Even if the above behaviour will be eventually fixed, I still suggest that just being able to specify custom encoding/decoding/formatting handlers might sometimes be very helpful to the point of saving somebody's life :) and the implementation isn't very complicated.

If you are in favour of this change, please let me know if you want me to add anything to this PR.

### DIFF
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -18,6 +18,9 @@ if Code.ensure_loaded?(Postgrex.Connection) do
       * `:ssl` - Set to true if ssl should be used (default: false)
       * `:ssl_opts` - A list of ssl options, see ssl docs
       * `:lazy` - If false all connections will be started immediately on Repo startup (default: true)
+      * `:encoder` — Custom encoder function
+      * `:decoder` — Custom decoder function
+      * `:formatter` — Custom formatter function
 
     """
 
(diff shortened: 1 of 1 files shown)


josevalim:
Great, definitely :+1: on this!

I have one question though: is custom functions the best API? I am wondering that, maybe if we passed modules, we could easily allow fallbacks. Something like:

    extensions: [PGis, JSON]

I think this API was designed to be chainable in the first place, is this correct @ericmj?

If, though, we want to keep the current API, we should:

1. Clean up the current implementation. Can we rewrite those as: `Keyword.put_new(..., ..., &custom(&1, &2, opts[:formatter])` or something of sorts?

2. We need tests

josevalim:
We have decided to add this directly to postgrex!

yrashk:
"this" being what exactly?

On Sat, Jan 10, 2015 at 1:45 AM, José Valim <notifications@github.com>
wrote:

> Closed #344 <https://github.com/elixir-lang/ecto/pull/344>.
>
> —
> <https://github.com/elixir-lang/ecto/pull/344#event-216675952>.
>



-- 
Y.
My company is hiring: https://trustatom.com/jobs

josevalim:
`extensions: [PGis, JSON]`

PR #352, opened 2015-01-14
update postgrex version in README

geemus:
updated to match suggested version in postgrex README


josevalim:
It is better to leave the postgrex version as is because then ecto will force the most appropriate one to be downloaded. :) Thank you!

PR #369, opened 2015-01-16
Make `Changeset.cast/4` optionals param optional

BlakeWilliams:
When calling `Changeset.cast/4` you are required to pass a value to
optionals. This makes the optionals parameter default to an empty list
when not present.


josevalim:

PR #382, opened 2015-01-20
Use fragment in docs instead of `+` and `-`

gjaldon:
`+` and `-` no longer part of query expressions?

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -289,7 +289,7 @@ defmodule Ecto.Repo do
 
       MyRepo.update_all(Post, title: "New title")
 
-      MyRepo.update_all(p in Post, visits: p.visits + 1)
+      MyRepo.update_all(p in Post, visits: fragment("? + 1", p.visits))
 
       from(p in Post, where: p.id < 10)
       |> MyRepo.update_all(title: "New title")


lib/ecto/repo.ex
```diff
@@ -405,8 +405,8 @@ defmodule Ecto.Repo do
   ## Examples
 
       MyRepo.transaction(fn ->
-        MyRepo.update(%{alice | balance: alice.balance - 10})
-        MyRepo.update(%{bob | balance: bob.balance + 10})
+        MyRepo.update(%{alice | balance: fragment("? - 10", alice.balance)})
```
josevalim: We shouldn't change those because those are not queries, just straight Elixir code. :)

gjaldon: Oops! Ok will change that now. :D

lib/ecto/repo.ex
```diff
@@ -289,7 +289,7 @@ defmodule Ecto.Repo do
 
       MyRepo.update_all(Post, title: "New title")
 
-      MyRepo.update_all(p in Post, visits: p.visits + 1)
+      MyRepo.update_all(p in Post, visits: fragment("? + 1", p.visits))
 
       from(p in Post, where: p.id < 10)
       |> MyRepo.update_all(title: "New title")
```
jeregrine: If I use this style, p isn't in scope. Is the solution to use
```
fragment("? + 1", :visits)
```
or 
```
fragment("visits + 1")
```

PR #383, opened 2015-01-20
[WIP] MySQL Adapter

jquadrin:
continues #193


josevalim:
That was fast, thank you! Btw, quick question, any reason why you decided to not try `mariaex`? I think it will integrate better given that it mimics postgrex and that `mysql-otp` does not even define an application file. :(

jquadrin:
I wasnt sure, but I'll switch to `mariaex` because of your insight :) thanks

Issue #378, opened 2015-01-20
Cannot increment in Repo.update_all


jeregrine:
http://hexdocs.pm/ecto/Ecto.Adapter.html#c:update_all/5 attepting to to increment a value in an update_all query as is shown in the docs.

Getting
```
== Compilation error on file web/models/card.ex ==
** (Ecto.Query.CompileError) `c.position() + 1` is not a valid query expression.

* If you intended to call a database function, please check the documentation
  for Ecto.Query to see the supported database expressions

* If you intended to call an Elixir function or introduce a value,
  you need to explicitly interpolate it with ^

    lib/ecto/repo/queryable.ex:69: anonymous fn/3 in Ecto.Repo.Queryable.update_all/5
    (elixir) lib/enum.ex:1036: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    lib/ecto/repo/queryable.ex:68: Ecto.Repo.Queryable.update_all/5
    expanding macro: Repo.update_all/2
    web/models/card.ex:84: Card.prepare_cards_for_insert_position/2
    (elixir) expanding macro: Kernel.|>/2
    web/models/card.ex:84: Card.prepare_cards_for_insert_posit
```

jeregrine:
```elixir
  from(c in Card,
         where: c.project_id == ^project_id and c.position > ^position) |> Repo.update_all(position: c.position + 1)
```

josevalim:
You need a fragment as `+` is no longer part of query expressions.

PR #384, opened 2015-01-20
Updating Repo.update_all docs to use fragments

jeregrine:
Lemme know if this is correct syntax in this case.  Fixes my confusion on #378


jeregrine:
HA I didn't even realize #382 existed! Closing.

PR #394, opened 2015-01-24
added Ecto.UUID

### LINKED ISSUE #386: Provide an Ecto.UUID type
Our `uuid` works with binaries and a string based representation can be very welcome.

briksoftware:
Closes #386

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -1,0 +1,52 @@
+defmodule Ecto.UUID do
+  @moduledoc """
+  An Ecto type for UUIDs.
+  """
+
+  @behaviour Ecto.Type
+
+  @doc """
+  The Ecto primitive type.
+  """
+  def type, do: :uuid
+
+  @doc """
+  UUIDs are never blank.
+  """
+  def blank?(_), do: false
+
+  @doc """
+  Casts to UUID.
+  """
+  def cast(<< _::64, ?-, _::32, ?-, _::32, ?-, _::32, ?-, _::96 >> = u), do: {:ok, u}
+  def cast(uuid = << _::128 >>), do: load(uuid)
+  def cast(_), do: :error
+
+  @doc """
+  Converts an string representing a UUID into a binary.
+  """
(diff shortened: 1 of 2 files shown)


lib/ecto/uuid.ex
```diff
@@ -0,0 +1,52 @@
...
+  @doc """
+  Generates a version 4 (random) UUID
+  """
+  def generate do
+    <<u0::48, _::4, u1::12, _::2, u2::62>> = :crypto.strong_rand_bytes(16)
+    <<u0::48, 4::4, u1::12, 2::2, u2::62>> |> encode
+  end
+
+  defp encode(<<u0::32, u1::16, u2::16, u3::16, u4::48>>) do
+    :io_lib.format("~8.16.0B-~4.16.0B-~4.16.0B-~4.16.0B-~12.16.0B", [u0, u1, u2, u3, u4]) |> to_string
```
josevalim: What if we use `Integer.to_string(x, 16)` instead of `:io_lib`? I think it will be clearer and faster!

lib/ecto/uuid.ex
```diff
@@ -0,0 +1,52 @@
...
+
+  @doc """
+  The Ecto primitive type.
+  """
+  def type, do: :uuid
+
+  @doc """
+  UUIDs are never blank.
+  """
+  def blank?(_), do: false
```
josevalim: I think we should consider empty binaries to be empty UUIDs. Thoughts?

josevalim: So we won't even attempt to cast those.

josevalim:
This is beautiful. :heart:

lib/ecto/uuid.ex
briksoftware: That was the way I did it in Crudex, but then I thought, an empty binary is actually no valid UUID so casting should fail in my opinion.

lib/ecto/uuid.ex
briksoftware: Integer.to_string(x, 16) does not do 0-padding, so if an hexlet is 0x02 it will just output "2"

josevalim:
Perfect!

troyk:
This is very awesome, thank you!!! Can't wait to take it for spin

PR #395, opened 2015-01-24
added cast/to_string

### LINKED ISSUE #385: Provide an explicit Ecto.DateTime type with casting

briksoftware:
Closes #385

### DIFF
--- a/lib/ecto/datetime.ex
+++ b/lib/ecto/datetime.ex
@@ -1,4 +1,26 @@
+defmodule Ecto.DateTime.Util do
+  @moduledoc false
+
+  @doc false
+  def zero_pad(val, count) do
+    num = Integer.to_string(val)
+    :binary.copy("0", count - byte_size(num)) <> num
+  end
+
+  @doc false
+  def to_i(string) do
+    String.to_integer(<<string::16>>)
+  end
+
+  @doc false
+  def to_li(string) do
+    String.to_integer(<<string::32>>)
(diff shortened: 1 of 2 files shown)


briksoftware:
I have added very very simple (but supposedly fast) parsing for now

briksoftware:
should I add the +00:00 format and missing timezone to be more flexible?

josevalim:
This looks *great*. I wouldn't add `+00:00` nor timezones but I would also parse `yyyy-mm-dd hh:mm:ss` in the datetime.

Also, should we add a simple validation (just check they are in the proper boundary) or no?

briksoftware:
I am still unsure about `+00:00` since it has the same meaning as `Z` so it would fit in ISO8061 UTC definition. Also adding it costs basically nothing. I am adding the timezone-less format `yyyy-mm-dd hh:mm:ss` as well as `hh:mm:ssZ` for time.

Adding validation like month <= 12, day <= 31, etc would be trivial and can be done in guards. I would not validate the month/day combination however. What do you think?

josevalim:
Let's leave timezone out until someone asks. :) Agreed with everything else
though.

I also just realized that the datetime cast could just call the date and
time ones once we split them. Not sure if it helps implementation wise. :)


-- 


briksoftware:
basic validation is added and time without a timezone can also be read

josevalim:

PR #397, opened 2015-01-25
parse milliseconds

briksoftware:
I made the parser more tolerant, ignoring anything after the seconds. This handles any timezone and most importantly milliseconds in the ".ddd" form. This makes us more compatible with regular JSON encoding which always includes milliseconds when produced from Javascript.

### DIFF
--- a/lib/ecto/datetime.ex
+++ b/lib/ecto/datetime.ex
@@ -127,10 +127,8 @@ defmodule Ecto.Time do
   @doc """
   Casts to time.
   """
-  def cast(<<hour::16, ?:, min::16, ?:, sec::16, ?Z>>),
-    do: from_string_parts(to_i(hour), to_i(min),to_i(sec))
-  def cast(<<hour::16, ?:, min::16, ?:, sec::16>>),
-    do: from_string_parts(to_i(hour), to_i(min),to_i(sec))
+  def cast(<<hour::16, ?:, min::16, ?:, sec::16, _::binary>>),
+    do: from_string_parts(to_i(hour), to_i(min), to_i(sec))
   def cast(%Ecto.Time{} = t),
     do: {:ok, t}
   def cast(_),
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks! Unfortunately it can also parse bad times like "13:23:04###". I think we should explicitly match on ".000" then (and ignore it). :)

josevalim:
Sorry, match on ".xxx" where x can be any digit.

briksoftware:
it is a possibility, but the code would be slower having to parse each possible valid combination and less readable. A time like 13:23:04### although not following any documented format still carries its meaning of 13 hours, 23 minutes, 04 seconds in a readable format. What is the advantage of being strict here?

josevalim:
What if someone types: `13:23:009`? It could have been a mistake. Does the user mean 0 or 9? For the miliseconds, I don't think it would slow down the parser like this:

    ?., m1, m2, m3>> when valid_mili(m1, m2, m3)

And then:

    m1 in ?0..?9 and m2 in ?0..?9 and m3 in ?0..?9

Thoughts?

briksoftware:
well, direct user input of the date at this level is extremely unlikely, isn't it?

What I mean is that to parse this we have to have the following clauses:

1) yyyy-mm-dd hh:mm:ss
2) yyyy-mm-ddThh:mm:ss
3) yyyy-mm-ddThh:mm:ssZ
4) yyyy-mm-ddThh:mm:ss.sss
5) yyyy-mm-ddThh:mm:ss.sssZ

and maybe the "+00:00" timezone also. The current approach using pattern matching is not really suited, IMHO, for so many variants when we also want to validate the parts we do not care about.

My idea was, to have a cast that expects a valid format in Ecto.DateTime which can be parsed very cheaply. The ISO8601 format is not what one would type by hand, so the assumption that whatever library is generating those time strings is not doing it extremely wrong seems not too wild to me.

josevalim:
Right. So we could break into parts. For example:

```
def cast(<<...date..., t_or_space, ...time..., rest::binary>>) when t_or_space in [?\s, ?T] do
  if valid_rest(rest) do ... else :error end
end
```

briksoftware:
yeah, I was about to propose that as an alternative which should make everybody happy. I'll implement this in the next hours.

josevalim:
Awesome, thanks! :D

briksoftware:
done, it should be good now!

josevalim:

chvanikoff:
Hey guys, what about timezones? For example "2015-02-24T02:14:24+0000"
I think it would be easy to add one more case in valid_rest? function and I can send a patch if u don't mind.

PR #398, opened 2015-01-25
Fix documentation example for Ecto.Adapters.SQL.query/4

yrashk:


josevalim:
Oops, we should make this arg optional. :)

josevalim:
Done in master, thanks!

PR #399, opened 2015-01-25
Improve Migration DSL documentation

yrashk:
Describe a way to use migration DSL to use database-specific types


lib/ecto/migration.ex
```diff
@@ -245,7 +245,9 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Adds a column when creating or altering a table.
```
josevalim: Remember to keep the first line of the documentation short as it is used as a summary by ExDoc. :)

PR #404, opened 2015-01-26
add __state__

### LINKED ISSUE #353: Store state (built, loaded, deleted) in model
The idea here is to have a virtual field named `__state__` always generated with default value of `:built`. Loading data from the database (be it on Repo.all, Repo.insert or Repo.update) should set its status to `:loaded` while Repo.delete sets it to `:deleted`.

briksoftware:
Closes #353

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -80,16 +80,17 @@ defmodule Ecto.Integration.RepoTest do
   test "insert, update and delete" do
     post = %Post{title: "create and delete single", text: "fetch empty"}
 
-    assert %Post{} = TestRepo.insert(post)
+    assert %Post{__state__: :loaded} = TestRepo.insert(post)
     assert %Post{} = to_be_deleted = TestRepo.insert(post)
-    assert %Post{} = TestRepo.delete(to_be_deleted)
+    assert %Post{__state__: :deleted} = TestRepo.delete(to_be_deleted)
 
     post = TestRepo.one(Post)
+    assert post.__state__ == :loaded
     assert post.inserted_at
     assert post.updated_at
 
-    post = %{post | text: "coming very soon..."}
-    assert %Post{} = TestRepo.update(post)
+    post = %{post | text: "coming very soon...", __state__: :built}
+    assert %Post{__state__: :loaded} = TestRepo.update(post)
   end
 
   test "insert and update binary inferred type values" do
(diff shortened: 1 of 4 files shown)


lib/ecto/schema.ex
```diff
@@ -196,6 +196,8 @@ defmodule Ecto.Schema do
       Module.register_attribute(__MODULE__, :ecto_assocs, accumulate: true)
       Module.register_attribute(__MODULE__, :ecto_raw, accumulate: true)
 
+      Module.put_attribute(__MODULE__, :struct_fields, {:__state__, :built})
```
josevalim: Beautiful. This way it is not listed as a field anywhere, good catch!

test/ecto/schema_test.exs
```diff
@@ -69,6 +69,10 @@ defmodule Ecto.SchemaTest do
     assert Ecto.Model.primary_key(%SchemaModel{uuid: "hello"}) == "hello"
   end
 
+  test "has __state__ attribute" do
```
josevalim: Can we add an assertion that checks that `SchemaModel.__schema__(:__state__) == nil`?

integration_test/cases/repo.exs
```diff
@@ -80,16 +80,17 @@ defmodule Ecto.Integration.RepoTest do
   test "insert, update and delete" do
     post = %Post{title: "create and delete single", text: "fetch empty"}
 
-    assert %Post{} = TestRepo.insert(post)
+    assert %Post{__state__: :loaded} = TestRepo.insert(post)
```
josevalim: I would be more comfortable if we added those to `test/ecto/repo_test.exs` as they are not part of the integration test per se. Thoughts?

test/ecto/schema_test.exs
briksoftware: ok, will do now

integration_test/cases/repo.exs
briksoftware: That's where I wanted to add them, but the mock repository returns no model at all, it returns `1` apparently... but maybe I am missing something

josevalim: Ok, perfect, let's leave this as is!

test/ecto/schema_test.exs
briksoftware: done

josevalim:

PR #406, opened 2015-01-26
added after_load callback

### LINKED ISSUE #405: add after_load callback
When working on #353, I had the idea that it would be nice to add an after_load callback to the model, which would be invoked every time a model is loaded. This would give the user a chance to set the value of virtual fields in a very convenient way.

Of course, I am up for the task of implementing this if the idea is approved.

briksoftware:
Closes #405

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -314,7 +314,7 @@ defmodule Ecto.Adapters.SQL do
   def all(repo, sql, query, params, opts) do
     %{rows: rows} = query(repo, sql, Map.values(params), opts)
     fields = extract_fields(query.select.fields, query.sources)
-    Enum.map(rows, &process_row(&1, fields))
+    Enum.map(rows, &process_row(&1, fields, repo))
   end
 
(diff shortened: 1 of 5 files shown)


josevalim:
Thanks!

I don't think we should wrap after_load in a changeset because most of the times it won't come from a change. We could easily make it work without a changeset if we change the `Callbacks.__apply__` function to how it was used to be in the past:

https://github.com/briksoftware/ecto/blob/27d23419372866bba2b0c3cb55825f5d643deb39/lib/ecto/model/callbacks.ex#L181

We then need to explicitly document it it doesn't receive a changeset.

briksoftware:
oh! I thought having a changeset for all callbacks was desired. The only advantage of having a changesets there is having a reference to the repo. Should I change it to work without a changeset?

josevalim:
Yeah, I thought about the repo but I think it would be better to have no changeset. Plus doing any Repo query on after_load would be *insane*. :)

briksoftware:
ok, done

josevalim:

PR #407, opened 2015-01-26
Support casting floats and integers in addition to binaries to Decimal

MSch:
I want to use `Ecto.Changeset.cast/4` with a `:decimal` field with `params` coming from JSON. JavaScript's JSON encoding (and the JSON spec) don't distinguish between floats or integers. So using Poison or JSX to decode JSON and assign it to my changeset fails, unless `:decimal` can be casted to from floats and integers. That's what this PR enables.

```javascript
JSON.stringify({a: 1.0}) === '{"a":1}'
```


josevalim:

PR #411, opened 2015-01-27
Join query update all

gjaldon:
refs: https://github.com/elixir-lang/ecto/issues/410

@josevalim is this okay? what else needs done?

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -314,6 +314,19 @@ defmodule Ecto.Integration.RepoTest do
     assert [%Post{}] = TestRepo.all(Post)
   end
 
+  test "delete all with joins" do
+    user = TestRepo.insert(%User{name: "Tester"})
+    post = TestRepo.insert(%Post{title: "foo"})
+    TestRepo.insert(%Comment{text: "hey", author_id: user.id, post_id: post.id})
+    TestRepo.insert(%Comment{text: "foo", author_id: user.id, post_id: post.id})
+    TestRepo.insert(%Comment{text: "bar", post_id: post.id})
+
+    query = from(c in Comment, join: u in User, on: u.id == c.author_id, where: c.post_id in ^[post.id])
+    assert 2 = TestRepo.delete_all(query)
+
+    assert [%Comment{}] = TestRepo.all(Comment)
+  end
+
   test "delete all no entries" do
     assert %Post{id: id1} = TestRepo.insert(%Post{title: "1", text: "hai"})
     assert %Post{id: id2} = TestRepo.insert(%Post{title: "2", text: "hai"})
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -225,6 +225,18 @@ defmodule Ecto.Integration.RepoTest do
     assert %Post{title: "y"} = TestRepo.get(Post, id3)
   end
 
+  test "update all with joins" do
+    query = from p in Post, where: p.id > 0
+    user = TestRepo.insert(%User{name: "Tester"})
+    post = TestRepo.insert(%Post{title: "foo"})
+    comment = TestRepo.insert(%Comment{text: "hey", author_id: user.id, post_id: post.id})
+
+    query = from(c in Comment, join: u in User, on: u.id == c.author_id, where: c.post_id in ^[post.id])
+    TestRepo.update_all(query, text: "hoo")
+
+    assert %Comment{text: "hoo"} = TestRepo.get(Comment, comment.id)
+  end
+
   test "update all with filter" do
     assert %Post{id: id1} = TestRepo.insert(%Post{title: "1"})
     assert %Post{id: id2} = TestRepo.insert(%Post{title: "2"})
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -226,7 +226,6 @@ defmodule Ecto.Integration.RepoTest do
   end
 
   test "update all with joins" do
-    query = from p in Post, where: p.id > 0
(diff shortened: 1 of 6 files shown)


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -494,6 +493,7 @@ if Code.ensure_loaded?(Postgrex.Connection) do
       |> List.flatten
       |> Enum.filter(&(&1 != nil))
       |> Enum.join(" ")
+      |> String.rstrip
```
josevalim: Can we remove the need for this `rstrip`? It is a bit expensive as it needs to traverse the whole string, so I think we should find a way to not need it.

josevalim:
The error you get is because in the planner we are forcing it to contain only wheres:

https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/planner.ex#L286

I would rename this option from `only_where` to `only_filters` and then change this function to also ignore the joins:

https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/planner.ex#L513

i.e. we no longer need to check they are empty. After all this is done, we just need to rename only_where to only_filters in this file too: https://github.com/elixir-lang/ecto/blob/master/lib/ecto/repo/queryable.ex

Then write integration tests here:

https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs

gjaldon:
@josevalim @ericmj @yrashk - could we add `:select` keyword to update_all/2? That way, we can do something like this:
```
assoc(checklist, :checklist_associations)
|> join(:inner, [a], i in Item, i.id == a.checklist_item_id)
|> select([a, i], i)
|> Repo.update_all(text: "great")
```

If you're ok with that, how will the SQL query for that be like? Or to look at this some other way, what's a good way to convert the following SQL query to Ecto's query language?
```
UPDATE checklist_associations m0 SET deleted_at = now() FROM checklist_items AS m1 WHERE m0.checklist_item_id = m1.id AND m0.checklist_id = 1;
```

josevalim:
I don't think we should allow select because we can't really select anything.

> UPDATE checklist_associations m0 SET deleted_at = now() FROM checklist_items AS m1 WHERE m0.checklist_item_id = m1.id AND m0.checklist_id = 1;

I think it would be:

    assoc(checklist, :checklist_associations) |> Repo.update_all(deleted_at: fragment("now()"))

integration_test/cases/repo.exs
```diff
@@ -264,6 +264,18 @@ defmodule Ecto.Integration.RepoTest do
     assert %Post{title: "y"} = TestRepo.get(Post, id3)
   end
 
+  test "update all with joins" do
```
josevalim: We need a test for "delete all with joins" and we should be good to go! :D

josevalim:
Btw, there is one failure in the build, probably because there is some only_where remaining in the test directory. :D

integration_test/cases/repo.exs
```diff
@@ -264,6 +264,18 @@ defmodule Ecto.Integration.RepoTest do
...
+    query = from p in Post, where: p.id > 0
```
josevalim: This query is not used? :)

josevalim:
Ping. :) Anything I can do to help with this one?

gjaldon:
@josevalim sorry for the delay just started with a new job. Will put in some time over the week to get this done. Will let you know if I need help. :)

josevalim:
Don't worry, I was just checking in case something was blocking on me!

josevalim:
Good luck at the new job!

gjaldon:
@josevalim thanks for the well-wishes on my job! :+1: It's my first job that works with Elixir full-time and I plan to stay with it for quite a while. Really happy to get to use Elixir at work! :)

Btw, this is good for review. About joins option in `delete_all/1`, looks like there's no actual JOIN keyword for the DELETE command. JOIN's equivalent is USING when doing DELETE. http://www.postgresql.org/docs/9.0/static/sql-delete.html

Just let me know if there are any changes needed.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -87,21 +87,21 @@ if Code.ensure_loaded?(Postgrex.Connection) do
...
 
     def delete_all(query) do
       sources = create_names(query)
       {table, name, _model} = elem(sources, 0)
 
-      where = where(query.wheres, sources)
-      where = if where, do: " " <> where, else: ""
-      "DELETE FROM #{quote_name(table)} AS #{name}" <> where
+      join  = if query.joins != [], do: using(query.joins, sources)
+      where = if query.joins == [], do: where(query.wheres, sources)
```
josevalim: What if we have where and joins? the syntax won't work?

gjaldon:
Removed :on option for joins in `delete_all/2` and just used :where instead. This means that joins behave a little differently in `delete_all/2` compared to `update_all/2` and `all/1`. Doesn't like there's no way around it since Postgresql's DELETE doesn't support JOIN but instead relies on USING to do joins.

Hope this is okay. Let me know if there's a better way to do this.

josevalim:
> Removed :on option for joins in delete_all/2 and just used :where instead.

We need to support both. So maybe we need to collect the on expressions when building the join and just unify with AND clauses when building the where... does it makes sense?

josevalim:
But don't worry, I can take a look at this tomorrow!

gjaldon:
> We need to support both. So maybe we need to collect the on expressions when building the join and just unify with AND clauses when building the where... does it makes sense?

I've done as you suggested above for `delete_all/2`. Let me know if there's anything else I missed for this functionality.

josevalim:

PR #412, opened 2015-01-27
Ecto.Changeset.convert_params/1 bugfix or behaviour change

whatyouhide:
If the keys in the params are mixed atoms and strings, validations fail and the params with an atom key are marked as missing (if required). For example:

```elixir
defmodule MyModule do
  def changeset(instance, params \\ nil) do
    params |> cast(instance, ~w(f1 f2))
  end
end
```

If I call `changeset/2` like this:

```elixir
MyModule.changeset(%MyModule{}, %{:f1 => "v1", "f2" => "v2"})
```

it will fail saying that one of the fields is required but not present. I added a failing test in the first commit, and the fix in the second commit.

I'm not 300% sure this is a real bug (I kind of smelled something strange when I saw `throw`/`catch` in the current implementation of `Ecto.Changeset.convert_params/1`), but I figured that a bug report *with* a PR for an eventual fix (if this turns out to be a bug) is better than a bug report alone :smile:

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -183,14 +183,12 @@ defmodule Ecto.Changeset do
   end
 
   defp convert_params(params) do
-    :maps.fold fn
-      key, _value, _acc when is_binary(key) ->
-        throw :noop
-      key, value, acc when is_atom(key) ->
-        Map.put(acc, Atom.to_string(key), value)
-    end, %{}, params
-  catch
-    :noop -> params
+    key_to_string = fn
+      {key, _val} = pair when is_binary(key) -> pair
+      {key, val} when is_atom(key)           -> {Atom.to_string(key), val}
+    end
+
+    params |> Enum.map(key_to_string) |> Enum.into(%{})
   end
 
   defp error_on_blank(type, key, value, errors) do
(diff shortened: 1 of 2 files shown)


josevalim:
I don't think we should support mixed. The previous behaviour was on purpose and the throw/catch was to avoid traversing everything if we had binary keys. Maybe we can raise an error if you have mixed keys, instead of ignoring the atom ones, but I think supporting both puts us in a road I don't really think we should go.

whatyouhide:
@josevalim I strongly agree on the fact that we shouldn't become like Ruby/Rails, where the distinction between atoms (well, symbols) and strings is practically none (unless it bites you :imp:).

That said, I think the current behaviour is slightly confusing since

* you expect strings to work (since parameters usually contain string keys since they come from the URL query)
* you also expect atoms to work, because you manipulate pretty much everything else with atoms in Ecto (schema fields, struct keys and so on); this is even more true since with only atom keys *it works*

Furthermore, the arrays of fields you pass to the `cast/4` function can pretty much be everything (as far as I managed to test): only atoms, only strings, atoms and strings. They're all converted to atoms. Probably that's another source of confusion: Ecto lets me use `cast/4` with atoms/string/atoms+strings but then it "fails" when I mix atoms and strings in the parameters.

If we (well, you :smile:) decide to keep the current behaviour, I'd be glad to at least raise some kind of error so that users know what to do to fix it.

What do you think?

josevalim:
Fixed in master. I am just quickly raising!

whatyouhide:
Ok, that's progress :smile: Thanks!

PR #415, opened 2015-01-28
Docs in Ecto.Model.Callbacks

whatyouhide:
* Rephrased words here and there
* Fixed a bunch of typos
* Fixed some code blocks (which were being annoying because of a missing `:do` keyword)


whatyouhide:
Thanks!

PR #418, opened 2015-01-29
Refactor the code in Ecto.Model.Callbacks

whatyouhide:
This is something of a throwaway PR, but I'm sure that if it gets rejected I'll still learn something :smile:.

Dealing with `{local_function}`, `{local_function, args}`, `{module, function}`, `{module, function, args}` is a mess and the way it was handled feels a little bit too magic-y to me. It took me a while to understand the `(module, function, args)` vs. the `(function, args, [])` pattern, where the latter can mean both a call to a local function with args as well as function (`args`) called on the `function` module :octopus:.

This PR makes the code a little more verbose (and not very DRY, but it wasn't very DRY before either - I guess we have to blame not being able to expand private macros at the module level at compile time!), but IMO easier to understand.

I haven't updated the docs for this change (resulting in lots of functions not documented) because I want to hear from you first; if you like the changes I'll update the docs :smiley:.

Thanks!

### DIFF
--- a/lib/ecto/model/callbacks.ex
+++ b/lib/ecto/model/callbacks.ex
@@ -109,12 +109,17 @@ defmodule Ecto.Model.Callbacks do
       before_insert User, :generate_permalink
 
   """
-  defmacro before_insert(function, args \\ []),
-    do: register_callback(:before_insert, function, args, [])
+  defmacro before_insert(function),
+    do: register_callback(:before_insert, function, [])
 
-  @doc """
-  Same as `before_insert/2` but with arguments.
-  """
+  defmacro before_insert(function, args)
+    when is_list(args),
+    do: register_callback(:before_insert, function, args)
+
+  defmacro before_insert(module, function)
+    when is_atom(function),
+    do: register_callback(:before_insert, module, function, [])
+
   defmacro before_insert(module, function, args),
     do: register_callback(:before_insert, module, function, args)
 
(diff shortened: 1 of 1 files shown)


lib/ecto/model/callbacks.ex
```diff
@@ -109,12 +109,17 @@ defmodule Ecto.Model.Callbacks do
       before_insert User, :generate_permalink
 
   """
-  defmacro before_insert(function, args \\ []),
-    do: register_callback(:before_insert, function, args, [])
+  defmacro before_insert(function),
+    do: register_callback(:before_insert, function, [])
+
+  defmacro before_insert(function, args)
+    when is_list(args),
```
josevalim: We can inline the when in the same line as it is not s long one.

lib/ecto/model/callbacks.ex
```diff
@@ -109,12 +109,17 @@ defmodule Ecto.Model.Callbacks do
...
+    do: register_callback(:before_insert, function, args)
+
+  defmacro before_insert(module, function)
+    when is_atom(function),
+    do: register_callback(:before_insert, module, function, [])
 
-  @doc """
```
josevalim: Don't remove the docs. It is another function and we need to document it as such. That's the main issue with this PR, it adds yet another function to document! :)

whatyouhide:
I think I can't push to this PR anymore since it's been closed, so I'll just answer to your concerns with words instead of code :smile:.

I inlined all the `when` clauses; as for the docs, the issue of having an additional function to document can be solved (see [callbacks.ex in my fork](https://github.com/whatyouhide/ecto/blob/callbacks-refactor/lib/ecto/model/callbacks.ex)) with the "optional argument" trick:

```elixir
defmacro before_insert(function, args \\ [])

defmacro before_insert(function, args) when is_list(args),
  do: register_callback(:before_insert, function, args)
 
defmacro before_insert(module, function) when is_atom(function),
  do: register_callback(:before_insert, module, function, [])

defmacro before_insert(module, function, args),
  do: register_callback(:before_insert, module, function, args)
```

This way two functions have to be documented, like they're documented now.

**PS** As I stated in the PR, I didn't remove the docs and expect this to be merged that way :) I only removed them in order to clean up the code to be reviewed; I was going to add them back (and updated for the new "shape" of the functions) in case this turned out to be interesting.

PR #419, opened 2015-01-31
Make test transaction functions idempotent

reset:
This will make the three test transaction functions idempotent. Previous to this change a `FunctionClauseError` exception would be raised when any of the three functions were called out of order or multiple times.

This is especially useful in an umbrella application where one application is dependent upon another for testing purposes. If one or more applications within the umbrella attempt to call begin test transaction the test suite will shutdown in error.


josevalim:
Sweet. I will merge this soon (and add some tests!).

josevalim:

PR #421, opened 2015-01-31
Only drop index if it exists.

alanpeabody:
I don't know if this is the best solution, but I wanted to throw something up to start a discussion about handling the following situation:

```elixir
defmodule Blog.Repo.Migrations.AddPosts do
  use Ecto.Migration

  def change do
    create table(:posts) do
      add :title, :string
      add :body,  :string
      add :slug,  :string
      timestamps
    end

    create index(:posts, [:slug])
  end
end
```

```shell
mix ecto.migrate
mix ecto.rollback

** (Postgrex.Error) ERROR (42704): index "posts_slug_index" does not exist
```

Currently the rollback will fail because dropping the table will drop the indexes, then the drop index will raise an exception.

There may be very good reasons not to add the `IF EXISTS` sql, but this seemed like the simplest approach. If there is a better/preferred approach to this (such as adding something similar to [Rail's index_exists?](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-index_exists-3F)) I am happy to take a stab at it.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -410,7 +410,7 @@ if Code.ensure_loaded?(Postgrex.Connection) do
     end
 
     def execute_ddl({:drop, %Index{}=index}) do
-      "DROP INDEX #{quote_name(index.name)}"
+      "DROP INDEX IF EXISTS #{quote_name(index.name)}"
     end
 
     def execute_ddl(default) when is_binary(default), do: default
--- a/test/ecto/adapters/postgres_test.exs
+++ b/test/ecto/adapters/postgres_test.exs
@@ -434,7 +434,7 @@ defmodule Ecto.Adapters.PostgresTest do
 
   test "drop index" do
     drop = {:drop, %Index{name: "posts$main"}}
-    assert SQL.execute_ddl(drop) == ~s|DROP INDEX "posts$main"|
+    assert SQL.execute_ddl(drop) == ~s|DROP INDEX IF EXISTS "posts$main"|
   end
 
   test "alter table" do


josevalim:
Maybe, when we reverse, we should execute the operations in reverse order?

alanpeabody:
Okay, I will take a stab at reversing.

alanpeabody:
@josevalim So I have been digging into this and I think this is a bit more involved then I expected.

My understanding is that essentially we are just calling the "change" function here: https://github.com/elixir-lang/ecto/blob/master/lib/ecto/migration/runner.ex#L22

Then each add/drop in the change function is just firing the logic as it is run. I don't know where/how to reverse that logic (without changing the DSL to not be a function that is executed).

josevalim:
You are right. Let me think a bit more about it and I will get back to you. Another idea is to add things like drop_if_exists and create_if_exists and make their reverses strict.

joshnuss:
One possibility is making `change` a macro so that the code can be executed backwards when needed:

```
defmodule Migration do
  defmacro change(do: block) do
    {:__block__, [], code} = block
    reversed_block =  {:__block__, [], Enum.reverse(code)}

    quote do
      def change(:forward) do
        unquote(block)
      end

      def change(:reverse) do
        unquote(reversed_block)
      end
    end
  end
end
```

josevalim:
It is unfortunately not that easy. For example:

```
if rainbows? do
  create table ...
  create index ...
end
```

How would we know how to go inside if and traverse its block? :)

alanpeabody:
It seems to me like we could alter the create/drop macros to append actions to a list instead of executing immediately. It would still require change/up/down becoming a macro and changing the DSL however.

That said, I haven't tried anything like this so this isn't much more then naive brainstorming but I would think we could create a DSL like this: 

```elixir
defmodule Blog.Repo.Migrations.AddPosts do
  use Ecto.Migration

  change do
    create table(:posts) do
      add :title, :string
      add :body,  :string
      add :slug,  :string
      timestamps
    end

    if rainbows? do
      create index(:posts, [:slug])
    end
  end
end
```

That expanded to:

```elixir
defmodule Blog.Repo.Migrations.AddPosts do
  use Ecto.Migration

  def change do
    commands = []
    commands = [{:create, table(:posts) do
      add :title, :string
      add :body,  :string
      add :slug,  :string
      timestamps
    end} | commands]

    if rainbows? do
      commands = [{:create, index(:posts, [:slug]) | commands]
    end

    commands
  end
end
```

Then runner is calling something more like:

```elixir
  commands = apply(module, operation, [])
  if direction == :backwards do
    commands = commands.reverse
  end
  timer.tc(#.. execute commands)
```

josevalim:
We could but it would add indirection and possibly mess up error messages,
stacktraces, and other things. It likely isn't worth it. :)

-- 


alanpeabody:
@josevalim makes sense. If you think of an approach you think makes sense I would like to try to tackle it.

PR #422, opened 2015-01-31
Cast integers to float.

alanpeabody:
This is important when JavaScript is building requests because JS only
has floats, but represents 2.0 as 2 when serializing.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -343,6 +343,8 @@ defmodule Ecto.Type do
 
       iex> cast(:float, 1.0)
       {:ok, 1.0}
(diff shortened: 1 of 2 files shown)


josevalim:
Awesome! I think now is a good time to add tests for integer and float casting. You can see examples here:

https://github.com/elixir-lang/ecto/blob/master/test/ecto/type_test.exs#L39

Can you please amend this PR?

alanpeabody:
@josevalim NP, tests added!

josevalim:
Doh, I just realized the tests are in the doctests. Sorry for the derpness, I merged just the first commit.

Issue #391, opened 2015-01-23
Support adding/dropping indexes concurrently


josevalim:

josevalim:
* Docs from Postgres: http://www.postgresql.org/docs/9.2/static/sql-createindex.html
* How Rails implements it: http://robots.thoughtbot.com/how-to-create-postgres-indexes-concurrently-in

This pull request consists of four parts:

1. Adding the option to Ecto.Migration.Index
2. Changing the adapter to be aware of this new option
3. Support the `@disable_ddl_transaction true` inside Ecto.Migration that will automatically remove the transaction in Ecto.Migrator
4. Tests (could all be unit ones)

whatyouhide:
I'm having a shot at this and I have one question.

Since `CREATE INDEX CONCURRENTLY` is Postgres-specific, am I doing the right thing by adding a `:concurrently` field to the `Ecto.Migration.Index` struct? Thanks in advance!

josevalim:
Yes! It should default to false and other databases should likely raise!

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

whatyouhide:
Ok, I'm halfway through the implementation. I'll add the `@disable_ddl_transaction` functionality and then some tests. Thanks!

PR #424, opened 2015-02-01
Add support for adding/dropping indexes concurrently with Postgres

whatyouhide:
I added support for adding/dropping indexes concurrently as specified in #391.

I'm not sure this PR is ready to merge, I think I need a short review first.

The most straightforward part has been adding `@disable_ddl_transaction` to `Ecto.Migration`: what I did was adding a `@before_compile` hook in `Ecto.Migration.__using__/1` in order to define a `__disable_ddl_transaction__/0` function on the migration module. This way, I can retrieve that value in `Ecto.Migrator` and avoid wrapping everything inside a transaction if that's the case.

The part I'm the least sure about is actually adding the `CONCURRENTLY` keyword to the query. I added the `:concurrently` field to the `Ecto.Migration.Index` struct and added the `CONCURRENTLY` only for the Postgres adapter (since only Postgres supports this option AFAIK), but didn't take any measure to ensure that the `:concurrently` field is ignored (or raised against) by other adapters.

What do you think for now?

**PS** I'd also love a small review to the unit test I wrote; I have to check that no transactions are run, but the schema migrator runs one so I can't just do `refute_receive {:transaction, _}`. Let me know!

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -63,6 +63,32 @@ defmodule Ecto.Migration do
 
   Notice not all commands are reversible though. Trying to rollback
   a non-reversible command will raise an `Ecto.MigrationError`.
+
+  ## Transactions
+
+  By default, Ecto runs all migrations inside a transaction. That's not always
+  ideal: for example, PostgreSQL allows to create/drop indexes concurrently but
+  only outside of any transaction (see the [PostgreSQL
+  docs](http://www.postgresql.org/docs/9.2/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY)).
+
+  Migrations can be forced to run outside a transaction by setting the
+  `@disable_ddl_transaction` module attribute to `true`:
+
+      defmodule MyRepo.Migrations.CreateIndexes do
+        use Ecto.Migration
+        @disable_ddl_transaction true
+
+        def change do
+          create index(:posts, [:slug], concurrently: true)
+        end
+      end
+
+  Since running migrations outside a transaction can be dangerous, consider
+  performing very few operations in such migrations.
+
+  See the `index/3` function for more information on creating/dropping indexes
+  concurrently.
+
   """
 
   defmodule Index do
@@ -229,6 +255,17 @@ defmodule Ecto.Migration do
 
   Indexes are non-unique by default.
 
+  ## Adding/dropping indexes concurrently
+
+  PostgreSQL supports adding/dropping indexes concurrently (see the
+  [docs](http://www.postgresql.org/docs/9.4/static/sql-createindex.html)).
+  In order to take advantage of this, the `:concurrently` option needs to be set
+  to `true` when the index is created/dropped.
+
+  **Note**: in order for the `:concurrently` option to work, the migration must
+  not be run inside a transaction. See the `Ecto.Migration` docs for more
+  information on running migrations outside of a transaction.
+
   ## Examples
 
       # Without a name, index defaults to products_category_id_sku_index
@@ -236,6 +273,9 @@ defmodule Ecto.Migration do
 
       # Name can be given explicitly though
       drop index(:products, [:category_id, :sku], name: :my_special_name)
+
+      # Adding an index concurrently
+      create index(:products, [:category_id, :sku], concurrently: true)
 
   """
   def index(table, columns, opts \\ []) when is_atom(table) and is_list(columns) do
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -69,8 +69,19 @@ defmodule Ecto.Migration do
     @moduledoc """
     Defines an index struct used in migrations.
     """
-    defstruct table: nil, name: nil, columns: [], unique: false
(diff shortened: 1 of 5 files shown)


josevalim:
This pull request is **great**! I am adding just some tiny comments/suggestions!

lib/ecto/migration.ex
```diff
@@ -96,6 +107,15 @@ defmodule Ecto.Migration do
...
       def __migration__, do: true
+      @disable_ddl_transaction false
+      @before_compile Ecto.Migration
+    end
+  end
+
+  @doc false
+  defmacro __before_compile__(_env) do
+    quote do
+      def __disable_ddl_transaction__, do: @disable_ddl_transaction
```
josevalim: What if we store this under the __migration__ function?

lib/ecto/migrator.ex
```diff
@@ -56,13 +56,18 @@ defmodule Ecto.Migrator do
...
+    run_attempts = fn ->
       attempt(repo, module, :forward, :up, opts)
         || attempt(repo, module, :forward, :change, opts)
-        || raise Ecto.MigrationError,
-                 message: "#{inspect module} does not implement a `up/0` or `change/0` function"
+        || raise Ecto.MigrationError, message: "#{inspect module} does not implement a `up/0` or `change/0` function"
       SchemaMigration.up(repo, version)
     end
+
+    if module.__disable_ddl_transaction__ do
```
josevalim: We can move this to a function. `run_repo` or something.

test/ecto/migrator_test.exs
```diff
@@ -203,6 +212,18 @@ defmodule Ecto.MigratorTest do
...
   end
 
+  test "the migration can be disabled" do
+    capture_log fn ->
+      up(MockRepo, 0, NoTransactionMigration)
+
+      # Assert there's only one transaction message, which is for when the
+      # SchemaMigration does his thing. If @disable_ddl_transaction was set to
+      # false, we would have *two* {:transaction, _} messages.
+      {:messages, messages} = Process.info(self, :messages)
```
josevalim: You could do:

```elixir
# From the SchemaMigration insert transaction
assert_received {:transaction, _}

# No transaction from migration
refute_received {:transaction, _}
```

In fact, maybe we should add a test showing that repos with DDL transaction do run two transactions?

josevalim:
We just need one extra test here:

https://github.com/elixir-lang/ecto/blob/master/test/ecto/adapters/postgres_test.exs

That's where you can test if concurrently is being properly respected. :)

whatyouhide:
Ok, working on it!

whatyouhide:
Hey @josevalim, I made the changes you suggested. Quick wrap-up:

- Removed `__disable_ddl_transaction__` in favour of using `__migration__`
- Better tests for the transaction
- Added tests for the generated SQL
- Added `CONCURRENTLY` support for dropping indexes too (I completely missed that :|)

Let me know what else I can do :smiley:.

lib/ecto/migration.ex
```diff
@@ -69,8 +69,19 @@ defmodule Ecto.Migration do
...
+              columns: [],
+              unique: false,
+              concurrently: false
+
+    @type t :: %__MODULE__{
+      table: atom,
+      name: atom,
+      columns: [atom | String.t],
+      unique: boolean,
+      concurrently: boolean
```
josevalim: We need to update the docs of the `index` function too, right? Maybe add a couple paragraphs about how to use concurrent indexes and disable_ddl_transaction.

whatyouhide:
I'm usually a docs nerd, I completely forgot to add them here!

Last doubt: as the Postgres docs say, you can't create/drop indexes concurrently inside a transaction. Should we prevent setting the `:concurrently` option to `true` when `@disable_ddl_transaction` has not been set to `true`? We could just raise or we could let the error go down to Postgres/Postgrex. What do you think?

test/ecto/migrator_test.exs
```diff
@@ -45,6 +45,25 @@ defmodule Ecto.MigratorTest do
...
+  defmodule NoTransactionMigration do
+    use Ecto.Migration
+    @disable_ddl_transaction true
+
+    def change do
+      create index(:posts, [:foo])
+    end
+  end
+
+  defmodule TransactionMigration do
```
josevalim: Can we use an existing migration? Avoid defining many modules as they become hard to maintain and affect the testing time considerably. :)

josevalim:
> Last doubt: as the Postgres docs say, you can't create/drop indexes concurrently inside a transaction. Should we prevent setting the :concurrently option to true when @disable_ddl_transaction has not been set to true? We could just raise or we could let the error go down to Postgres/Postgrex. What do you think?

I am not sure. Right now I am more inclined to not do anything, because maybe it is a postgrex only limitation or maybe it will even change with time. I think this path will be fine as long as the docs are clear. :)

test/ecto/migrator_test.exs
whatyouhide: Fixed this by `push --force`'ing so that the history looks niiice!

whatyouhide:
Great, thanks! :smiley: :heart:

PR #428, opened 2015-02-05
TDS Adapter

mobileoverlord:
This is a work in progress. I am going through the integration tests and ensuring that they all pass. Unsure how we can set this up for Travis testing. I am a little new to that tool.

There are some things that I have encountered so far with the generic integration tests that are a little too PG specific. I have marked them in several tests with TODO comments and descriptions about the reason they are too specific.

- [x] Move PG Specific calls from general integration tests to adapter specific tests.
- [ ] Ensure that all integration tests pass
- [x] Add support for Dirty Reads.

### DIFF
--- a/integration_test/cases/lock.exs
+++ b/integration_test/cases/lock.exs
@@ -15,17 +15,17 @@ defmodule Ecto.Integration.LockTest do
   end
 
   setup do
-    %LockCounter{id: 42, count: 1} |> PoolRepo.insert
+    %LockCounter{count: 1} |> PoolRepo.insert
 
     on_exit fn ->
-      PoolRepo.get(LockCounter, 42) |> PoolRepo.delete
+      PoolRepo.get(LockCounter, 1) |> PoolRepo.delete
     end
 
     :ok
   end
 
   test "lock for update" do
-    query = from(p in LockCounter, where: p.id == 42, lock: true)
+    query = from(p in LockCounter, where: p.id == 1, lock: true)
     pid = self
 
     new_pid =
--- a/integration_test/cases/lock.exs
+++ b/integration_test/cases/lock.exs
@@ -13,19 +13,21 @@ defmodule Ecto.Integration.LockTest do
       field :count, :integer
     end
   end
-
+  # TODO
+  # MSSQL Does not allow insertion into identity (serial) columns. 
+  #  Removed and assumed that this is the first
   setup do
-    %LockCounter{count: 1} |> PoolRepo.insert
+    lc = %LockCounter{count: 1} |> PoolRepo.insert
 
     on_exit fn ->
-      PoolRepo.get(LockCounter, 1) |> PoolRepo.delete
+      PoolRepo.delete(lc)
     end
 
-    :ok
+    {:ok, lc: lc}
   end
 
-  test "lock for update" do
-    query = from(p in LockCounter, where: p.id == 1, lock: true)
+  test "lock for update", meta do
+    query = from(p in LockCounter, where: p.id == ^meta.lc.id, lock: true)
     pid = self
 
     new_pid =
--- a/integration_test/cases/lock.exs
+++ b/integration_test/cases/lock.exs
@@ -20,8 +20,13 @@ defmodule Ecto.Integration.LockTest do
 
   test "lock for update" do
     %{id: id} = PoolRepo.insert(%LockCounter{count: 1})
-    query = from(lc in LockCounter, where: lc.id == ^id, lock: true)
-    pid   = self()
+    pid = self()
+
+    # Here we are manually inserting the lock in the query
+    # to test multiple adapters. Never do this in actual
+    # application code: it is not safe and could also break.
+    query = from(lc in LockCounter, where: lc.id == ^id)
+    query = %{query | lock: PoolRepo.lock_for_update}
 
     {:ok, new_pid} =
       Task.start_link fn ->
--- a/integration_test/cases/lock.exs
+++ b/integration_test/cases/lock.exs
@@ -13,53 +13,36 @@ defmodule Ecto.Integration.LockTest do
       field :count, :integer
     end
   end
-  # TODO
-  # MSSQL Does not allow insertion into identity (serial) columns. 
-  #  Removed and assumed that this is the first
+
   setup do
-    lc = %LockCounter{count: 1} |> PoolRepo.insert
-
-    on_exit fn ->
-      PoolRepo.delete(lc)
-    end
-
-    {:ok, lc: lc}
+    on_exit fn -> PoolRepo.delete_all(LockCounter) end
   end
 
-  test "lock for update", meta do
-    query = from(p in LockCounter, where: p.id == ^meta.lc.id, lock: true)
-    pid = self
+  test "lock for update" do
+    %{id: id} = PoolRepo.insert(%LockCounter{count: 1})
+    query = from(lc in LockCounter, where: lc.id == ^id, lock: true)
+    pid   = self()
 
-    new_pid =
-      spawn_link fn ->
-        receive do
-          :select_for_update ->
-            PoolRepo.transaction(fn ->
-              [post] = PoolRepo.all(query)   # this should block until the other trans. commits
-              %{post | count: post.count + 1} |> PoolRepo.update
-            end)
-            send pid, :updated
-        after
-          5000 -> raise "timeout"
-        end
+    {:ok, new_pid} =
+      Task.start_link fn ->
+        assert_receive :select_for_update, 5000
+
+        PoolRepo.transaction(fn ->
+          [post] = PoolRepo.all(query) # this should block until the other trans. commit
+          %{post | count: post.count + 1} |> PoolRepo.update
+        end)
+
+        send pid, :updated
       end
 
     PoolRepo.transaction(fn ->
-      [post] = PoolRepo.all(query)           # select and lock the row
-      send new_pid, :select_for_update       # signal second process to begin a transaction
-      receive do
-        :updated -> raise "missing lock"     # if we get this before committing, our lock failed
-      after
-        100 -> :ok
-      end
-      %{post | count: post.count + 1} |> PoolRepo.update
+      [post] = PoolRepo.all(query)       # select and lock the row
+      send new_pid, :select_for_update   # signal second process to begin a transaction
+      refute_receive :udpated, 100       # if we get this before committing, our lock failed
+      PoolRepo.update(%{post | count: post.count + 1})
     end)
 
-    receive do
-      :updated -> :ok
-    after
-      5000 -> "timeout"
-    end
+    assert_receive :updated, 5000
 
     # Final count will be 3 if SELECT ... FOR UPDATE worked and 2 otherwise
     assert [%LockCounter{count: 3}] = PoolRepo.all(LockCounter)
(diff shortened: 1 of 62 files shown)


josevalim:
This is awesome! It will help a lot in taking the SQL API further. I will review the code and give feedback where I can but I believe this should live in another repository as having a SQL Server up and running is not exactly trivial. Thoughts?

integration_test/cases/lock.exs
```diff
@@ -15,17 +15,17 @@ defmodule Ecto.Integration.LockTest do
   end
 
   setup do
-    %LockCounter{id: 42, count: 1} |> PoolRepo.insert
+    %LockCounter{count: 1} |> PoolRepo.insert
 
     on_exit fn ->
-      PoolRepo.get(LockCounter, 42) |> PoolRepo.delete
+      PoolRepo.get(LockCounter, 1) |> PoolRepo.delete
```
josevalim: I don't think we can rely on the ID being one. So instead we should do somthing like:

```elixir
lc = PoolRepo.insert %LockCounter{count: 1}

on_exit fn -> PoolRepo.delete(lc) end

{:ok, lc: lc}
```

Thoughts?

integration_test/cases/lock.exs
```diff
@@ -15,17 +15,17 @@ defmodule Ecto.Integration.LockTest do
...
     end
 
     :ok
   end
 
   test "lock for update" do
-    query = from(p in LockCounter, where: p.id == 42, lock: true)
+    query = from(p in LockCounter, where: p.id == 1, lock: true)
```
josevalim: Then you will be able to access it here as:

    test "...", meta do
      ... p.id == ^meta.lc.id ...
    end

integration_test/cases/migration.exs
```diff
@@ -147,7 +151,7 @@ defmodule Ecto.Integration.MigrationTest do
       end
 
       def down do
-        execute "DELETE FROM migrations_test WHERE id IN (SELECT id FROM migrations_test LIMIT 1)"
+        execute "DELETE FROM migrations_test WHERE id IN (SELECT id FROM migrations_test)"
```
josevalim: Yeah, we should not even run SQL commands in here! :D

josevalim:
I also think we will have to include the integration_test directory in the package so you can access it from your package too!

mobileoverlord:
I think that it is a great idea to distill out the adaptors into their own packages, especially for setup and CI purposes. I noticed something today that I got stuck on as far as trying to get all the integration tests running and thats the dreaded nil. So I had a conversation with ericmj in regards to how postgres infers what type the data that you are sending it is. In MSSQL I need to pass the data type along to the server for executing the query. In most situations this is fine because I can encode the data type by predicting it when it comes to me in the driver. For example, {{_,_,_},{_,_,_}} is datetime and binaries are strings and uuid's are recognizable. But NULL is where I am having a problem.

In SQL there is no true NULL data type that can be set and sent to the server. Lets say for example, you have a uuid column and you want to insert NULL, you would have to know the column is uuid so you can declare the type. otherwise if you say try to insert a NULL which is being sent represented as a different datatype, the server argues that it cannot implicitly convert an int to uuid even though NULL is NULL.

So this leads me to ask, where should we handle this issue.
One idea is to have ecto pass me the column_info regarding the data type to the driver as part of the params in the query. This would allow me to understand what I should try to encode the data as before sending it to sql. A primary sanity check. Although I am afraid this would maybe require other drivers like postgres to handle this additional data showing up in the query spec.

Another idea is to send everything over the batch t-sql protocol with the tds driver. This filters everything through the parser and could loosen the strict requirements on declaring data types. Typically this is not recommended, as Microsoft urges everything to use the RPC protocol to ensure parameterization of queries benefit from built in prevention of sql injection, escaping and things like that.

Let me know what you think or if you would like additional information. I would love to get this working with your SQL adaptor so we can make it flexible enough to work with many databases.

josevalim:
> One idea is to have ecto pass me the column_info regarding the data type to the driver as part of the params in the query.

This is the correct solution imo.

ericmj:
> > One idea is to have ecto pass me the column_info regarding the data type to the driver as part of the params in the query.

> This is the correct solution imo.

This does indeed sound like the best solution. The only issue I see is that mssql will require more explicit casting because ecto cannot always infer the type and we just let postgresql infer it instead. But this is something we will have to live with.

@mobileoverlord This is looking really great so far. Thanks for you work on this and the mssql driver.

mobileoverlord:
I started realizing this earlier in the development of the driver. 
https://github.com/livehelpnow/tds/blob/master/lib/tds/protocol.ex#L43
I made a Parameter struct to handle things like the parameter name, value and the direction. MSSQL had output parameters for stored procedure calls which I need to still implement. I guess we could make a map for passing params similar to what I'm doing.
In order to handle ecto's passing of raw param values, I am checking and converting them into %Parameters{} upon entry to that method.

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,519 @@
...
+      scale     = Keyword.get(opts, :scale)
+      type_name = ecto_to_db(type)
+
+      cond do
+        type == :serial -> "bigint"
+        pk == true      -> "bigint"
+        size            -> "#{type_name}(#{size})"
+        precision       -> "#{type_name}(#{precision},#{scale || 0})"
+        type == :string -> "#{type_name}(255)"
+        type == :binary -> "#{type_name}(255)"
```
josevalim: This feels wrong. Is there a blob or something similar in SQLServer that does not contain a size?

lib/ecto/adapters/mssql/connection.ex
```diff
@@ -0,0 +1,519 @@
...
+      |> Enum.join(" ")
+    end
+
+    defp escape_string(value) when is_binary(value) do
+      :binary.replace(value, "'", "''", [:global])
+    end
+
+    defp ecto_to_db({:array, t}), do: "nvarchar"
+    defp ecto_to_db(:string),     do: "varchar"
+    defp ecto_to_db(:binary),     do: "varbinary"
```
josevalim: Same here. binary should likely be a blob or something with no size.

josevalim:
@mobileoverlord i have pushed fixes for the integration tests and other things you have mentioned. Do you want to rebase this? Is there anything else I can do? Just let me know!

mobileoverlord:
Thanks José.
I've been working on two aspects.
1. I've been working through the remainder of the integration tests trying to identify remaining issues regarding tests that work with PG but not with MSSQL.

2. I'm still working through the comments you gave in regards to encoding binary types as some all encompassing blob type. So far my research seems to dictate that this may work if I handle the encoding specific to text types prior to hitting the driver since MSSQL uses UCS2 as its default text encoding. I believe what you are proposing is to insert binary types like string, binary, uuid, and bit as straight binary and get MSSQL to cast it implicitly. I'm going to add an additional suite of tests to the TDS driver to test this theory. I agree that if it does work it would be the best solution as strict interpolation between binary types would not be required. I'll let you know what I find. 

Other than that I don't foresee any show stopper issues. We should revisit the beginning of this discussion regarding putting the adapter in its own repo when we get closer to it testing out.

Thanks for your attention and recommendations on this PR. :)

lib/ecto/adapters/mssql/connection.ex
mobileoverlord: I have changed the default string text type to ntext which does not pass a size characteristic. If a size characteristic is passed, I use nvarchar.

mobileoverlord:
I've been unable to find a total solution for implicitly handling binary data types. According to the chart on this page https://msdn.microsoft.com/en-us/library/ms191530.aspx and the way mssql handles text encoding differently between types I have determined that the most appropriate means of handling :string is to use nvarchar and force everything to UCS2 encoding. This works until uuid's. For a uuid to be implicitly converted from nvarchar the uuid string has to contain the dashes in the right place. Now, this isn't so much an issue for insert/update/delete operations as they are exposed in the adapter alongside the types. I think solution is to add a callback to the adapter for select's so adapter specific data encoding can happen prior to hitting the driver. Let me know how possible this is.

josevalim:
@mobileoverlord just to be clear, the issue is when you encode uuid as text, for example, if you have `uuid("dsadsadsa-djhkehsa-dsakjhjdsa")` or when an UUID is passed as parameter? Like `p.uuid == ^uuid`?

If it is the second, there isn't much we can do as we don't always know the type. Can't you encode the UUID as a binary? When you say:

> I have determined that the most appropriate means of handling :string is to use nvarchar and force everything to UCS2 encoding

Does it mean you are also representing binaries as a strings? Something like `\001` and so on?

mobileoverlord:
In my ideal situation I would like to encode :string as nvarchar and :binary as binary and :uuid as uniqueidentifier but I don't know how to do this since all three types are identifiable as binary. So, I've been trying to find a magical encoding type between all three. I understand that the var uuid in ```p.uuid == ^uuid``` cannot be identified, but we would still know the type of field as being uuid.

So I believe the difficulty is since strings are binary and binary is binary and uuid's are binary and there is no way to send SQL all three as a single type the driver needs to know at least what the column expects the type to be, or a way to encode it prior to hitting the driver.

mobileoverlord:
Added raise to column_type when trying to define a list type column.

Here is a list of places where tests need to have array types removed.

Remove array type from model definition
https://github.com/elixir-lang/ecto/blob/master/integration_test/support/models.exs#L10

Repo Tests Failures
https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs#L30-L34
https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs#L99-L106

mobileoverlord:
Oh and the migration of course
https://github.com/elixir-lang/ecto/blob/master/integration_test/support/migration.exs

mobileoverlord:
Additional Test Changes: MSSQL Does not allow inserting into primary key columns
https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs#L132-L148

mobileoverlord:
I think there is 1 or 2 problem(s) remaining which should allow the remainder of my tests to pass. I will demonstrate this problem using the issues we experienced in test validate_unique/3 in repo.exs.

basically this is what happens.
post = TestRepo.insert(%Post{title: "HELLO"})
Title is a :string which means the column gets defined as nvarchar.
INSERT INTO [posts] ([cost], [counter], [inserted_at], [intensity], [public], [text], [title], [updated_at], [uuid], [visits])  OUTPUT INSERTED.[id] , INSERTED.[counter] VALUES (@1, @2, @3, @4, @5, @6, @7, @8, @9, @10)
%Tds.Parameter{direction: :input, name: "@7", type: :string, value: "HELLO"}

Later we try to validate
on_insert = cast(%{"title" => "HELLO"}, %Post{}, ~w(title), ~w())
assert validate_unique(on_insert, :title, on: TestRepo).errors != []

This generates this call
SELECT p0.[title] FROM [posts] AS p0  WHERE (p0.[title] = @1)
%Tds.Parameter{direction: :input, name: "@1", type: nil, value: "HELLO"}

The issue is that since its a select and I don't know the type I default to encoding binary as binary, however, :strings go as nvarchar which is utf16

Therefore HELLO utf8 (binary) doesn't equal HELLO utf16 (binary)

Aside from this, test "get(!) with custom primary key" is kinda pg specific
:)

josevalim:
Strings should always be encoded as strings (UTF-16). When we want to pass a binary, they are going to be explicitly tagged, as we do with UUIDs: https://github.com/elixir-lang/ecto/blob/master/lib/ecto/adapters/postgres/connection.ex#L317-L329

josevalim:
Given what I said above, the get(!) test is wrong as we would need to tag it somehow too. I will work on it.

josevalim:
Closing! :D

PR #429, opened 2015-02-05
Create indexes with USING

### LINKED ISSUE #427: Support GIS and GiST indexes in Postgres
Copying @whatyouhide since you tackled the last index related issue! :D

whatyouhide:
This PR should close #427 since it adds support for GiST and GIN indexes in Postgres but also for all other types of indexes, since it adds support for `USING` when creating indexes.

I added docs for this option as well as a couple of unit tests (testing the generated SQL). I'm not sure if I should add any integration test or any test under `Ecto.Migration` since the change is very simple and just translates to a Postgres feature, without involving much Ecto.

Let me know what you think!


josevalim:

PR #431, opened 2015-02-06
Update the version of the deps in the README

whatyouhide:
I upgraded the version of the postgrex and Ecto dependencies shown in the README "Usage" section, since that is probably where people look at when they're adding Ecto to their deps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Add Ecto as a dependency in your `mix.exs` file. If you are using PostgreSQL, yo
 ```elixir
 defp deps do
   [{:postgrex, ">= 0.0.0"},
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,8 +51,8 @@ Add Ecto as a dependency in your `mix.exs` file. If you are using PostgreSQL, yo
 
 ```elixir
 defp deps do
-  [{:postgrex, ">= 0.0.0"},
```
josevalim: Please leave it `>= 0.0.0`. Ecto will force the proper version.

whatyouhide:
Oh, sorry! Didn't know that and actually it makes sense, I didn't think of that :). Fixed!

PR #433, opened 2015-02-06
Composable casts

### LINKED ISSUE #401: Composable cast?
I have a use case in a webmachine where I create the changeset progressively, each time picking some parts of the params. `cast/4` can work only with models, but couldn't it be extended to support changesets as well?

whatyouhide:
This should close #401.

This PR adds the possibility of composing casts by making `Ecto.Changeset.cast/4` accept a changeset instead of a model as its second argument:

```elixir
changeset = cast(%{title: "Title"}, %Post{}, ~w(title), ~w())
cast(%{title: "Foo", body: "Bar"}, changeset, ~w(title), ~w(body))
```

I think the nicest solution (and the one I've implemented) is to apply the new "cast" to the model in the argument changeset (in the example, `changeset.model`) and then to meticulously merge the argument changeset and the new changeset (I added the `merge_changesets/2` function for that).

Another way to implement this would have been to manually call `cast` with the model in the argument changeset and with merged title, options and params) but that way an already applied cast is just re-applied another time; maybe this solution is nicer though, so I'm more than willing to change the implementation (I could `push --force` just the second commit) if so.

Thanks! :smiley:

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -244,6 +244,48 @@ defmodule Ecto.Changeset do
 
   ## Working with changesets
 
+  @doc """
+  Merges two changesets.
+
+  This function merges two changesets provided they have been applied to the
+  same model (their `:model` field is equal); if the models differ, an
+  `ArgumentError` exception is raised. If one of the changesets has a `:repo`
+  field which is not `nil`, then the value of that field is used as the `:repo`
+  field of the resulting changeset; if both changesets have a non-`nil` and
+  different `:repo` field, an `ArgumentError` exception is raised.
+
+  The other fields are merged with the following criteria:
+
+  * `params` - params are merged (not deep-merged) giving precedence to the
+    params of `changeset2` in case of a conflict. If either changeset has its
+    `:params` field set to `nil`, the resulting changeset will have its params
+    set to `nil` too.
+  * `changes` - changes are merged giving precedence to the `changeset2`
+    changes.
+  * `errors` and `validations` - they are simply concatenated.
+  * `required` and `optional` - they are merged; all the fields that appear
+    in the optional list of either changesets and also in the required list of
+    the other changeset are moved to the required list of the resulting
+    changeset.
+
+  ## Examples
+
+      iex> changeset1 = cast(%{title: "Title"}, %Post{}, ~w(title), ~w(body))
+      iex> changeset2 = cast(%{title: "New title", body: "Body"}, %Post{}, ~w(title body), ~w())
+      iex> changeset = merge(changeset1, changeset2)
+      iex> changeset.changes
+      %{body: "Body", title: "New title"}
+      iex> changeset.required
+      [:title, :body]
+      iex> changeset.optional
+      []
+
+      iex> changeset1 = cast(%{title: "Title"}, %Post{body: "Body"}, ~w(title), ~w(body))
+      iex> changeset2 = cast(%{title: "New title"}, %Post{}, ~w(title), ~w())
+      iex> merge(changeset1, changeset2)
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -209,6 +242,23 @@ defmodule Ecto.Changeset do
...
 
+  defp merge_changesets(%Ecto.Changeset{model: model} = cs1, %Ecto.Changeset{model: model} = cs2) do
+    new_params   = Map.merge(cs1.params, cs2.params)
+    new_changes  = Map.merge(cs1.changes, cs2.changes)
+    new_errors   = cs1.errors ++ cs2.errors
+    new_required = Enum.uniq(cs1.required ++ cs2.required)
+
+    # Merges the :optional field of both changesets and then pulls out all the
+    # fields that have now become required.
+    opt_set = Enum.into(cs1.optional ++ cs2.optional, HashSet.new)
```
josevalim: I think this will be more performant (and clearer) if you simply write it as `Enum.uniq(optional1 ++ optional2) -- required` :)

josevalim:
Beautiful. I have added just one tiny comment!

whatyouhide:
There you go, no more `HashSet`s in favour of a call to `--/2`. I didn't realize that if both lists have no duplicate elements than `--/2` is the same as `Set.difference/2`.

One more thing: should we make `merge_changesets/2` public? I feel like it could come in handy in some cases since maybe you're creating the changesets in different places and then you just want to merge them, as opposite to having to first create one and then `cast/4` on it.

Let me know what you think!

lib/ecto/changeset.ex
```diff
@@ -209,6 +242,21 @@ defmodule Ecto.Changeset do
...
+    new_params   = Map.merge(cs1.params, cs2.params)
+    new_changes  = Map.merge(cs1.changes, cs2.changes)
+    new_errors   = cs1.errors ++ cs2.errors
+    new_required = Enum.uniq(cs1.required ++ cs2.required)
+
+    # Merges the :optional field of both changesets and then pulls out all the
+    # fields that have now become required.
+    new_optional = Enum.uniq(cs1.optional ++ cs2.optional) -- new_required
+
+    %Ecto.Changeset{params: new_params, model: model, valid?: new_errors == [],
```
josevalim: Sorry, this is incomplete. We need to concat the validations too.

josevalim:
Great idea, let's make merge public as long as we guarantee both model and repo fields are equal.

whatyouhide:
Hey José, I'm sorry it took me so long (busy day!) and I've been terrible completely missing validations :disappointed: Pushed a fix for that, I'm going to push another couple of commits later tonight in order to make `merge_changesets/2` public (updating the tests and writing the docs is going to take me a little while, that's why I'm not doing it now).

In the meanwhile, should we call the function `merge_changesets` even if it's part of the `Ecto.Changeset` module? I'm thinking that just `merge` may be too generic especially since `Ecto.Model` imports `Ecto.Changeset`, but maybe we can find a less verbose name. Let me know what you think!

josevalim:
We should call it just merge. :)

-- 


whatyouhide:
Hey @josevalim, I made `merge/2` public and added tests and docs for it. Everything should be ready, let  me know if there's anything else I can do.

Just one thing: you said to fail in `merge/2` if the `:repo` fields of the changesets differ. I think it's better to fail only if they're not `nil` **and** they're different. If I have a changeset I've applied (with a non-`nil` repo) and one I still haven't applied (with a `nil` repo), the final changeset should have the repo of the applied one, right? Let me know if I should change this behaviour before merging :).

Thanks!

josevalim:

PR #434, opened 2015-02-07
Only drop existing indexes when reversing `create index`.

### LINKED ISSUE #423: change doesn't work when it creates a table and an index

alanpeabody:
This fixes the situation where reversing a migration that adds a table
and an index would fail, because dropping the table will remove the
indexes.

Fixes #423

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -200,6 +200,9 @@ defmodule Ecto.Migration do
   @doc """
   Creates an index.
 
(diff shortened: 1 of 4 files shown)


alanpeabody:
Note: When skipping indexes we don't log anything. Should we?

josevalim:
This is beautiful, thank you! And no, no logging required IMO!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

alanpeabody:
NP!

Issue #435, opened 2015-02-08
Boolean column default


ryankask:
I upgraded to master to take advantage of the `fragment` function when setting defaults and I noticed that my migrations broke because I used an Elixir boolean as a default:

      add :is_active, :boolean, default: true, null: false

Should `Ecto.Adapters.Postgres.Connection.default_expr/1` be defined like

    defp default_expr(literal) when is_number(literal) or is_boolean(literal), do: "DEFAULT #{literal}"

or is the intention now to just use binaries for boolean defaults?

      add :is_active, :boolean, default: "true", null: false

josevalim:
@ryankask thanks for reporting! This is definitely a bug and your fix is correct. We need to just add a new column here so we can test it too:

https://github.com/elixir-lang/ecto/blob/master/test/ecto/adapters/postgres_test.exs#L408-L410

:)

josevalim:
@ryankask thanks for reporting! This is definitely a bug and your fix is correct. We need to just add a new column here so we can test it too:

https://github.com/elixir-lang/ecto/blob/master/test/ecto/adapters/postgres_test.exs#L408-L410

A pull request is very appreciated, otherwise I will take a look at it soon!

PR #436, opened 2015-02-08
Allow boolean literals as defaults.

ryankask:
This PR addresses the bug described in #435.

The Postgres docs say: "The key words TRUE and FALSE are the preferred (SQL-compliant) usage."

This just uses `true` and `false` (lowercased). Case doesn't really matter but I wanted to check if this should be changed.


josevalim:
:heart:

PR #437, opened 2015-02-09
Allow unique option in migrations

ryankask:
I frequently use Django you can specify a column to be unique when defining a model field.

I tried this with ecto yesterday and it didn't work so this pull request adds support for this:

    add :slug, :text, unique: true, null: false

Which will generate

    "slug" text UNIQUE NOT NULL

The same thing can accomplished with 

    create index(:posts, [:slug], unique: true)

but seems like a convenient option. I'm not sure if it's desired for Ecto, however.

Note that you can't name the index inline like you can in Postgres. For that you would need to use `index/3` as you would now.

Also, `PRIMARY KEY` implies `UNIQUE NOT NULL` but this doesn't stop you from specifying both (redundant) options. It doesn't matter in Postgres:

    create table posts (slug text unique not null primary key)

works fine.


josevalim:
I think this is great. I will hold on merging it for now though because there is on going work on other adapters and I want to avoid messing it up. We will merge this later and find out a way to notify adapter implementations that they need to catch up! Thanks a lot!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:
@mobileoverlord can you please verify this also works with TDS and add a unit test to tds_ecto? Once you confirm, I am merging this.

mobileoverlord:
Good to go
https://github.com/livehelpnow/tds_ecto/commit/7eb035d10211a414b113336c8d55ffc5ca30eb44

josevalim:
:heart:

PR #439, opened 2015-02-10
Fix test to actually refute receiving :updated

MSch:


josevalim:
I have done this typo approximately 1 gazillion times. Thanks for "udpating" the code!

PR #441, opened 2015-02-10
Fix GenServer.reply/2 calls

MSch:
Not sure what's going on here, not sure what the tests for this should be.


josevalim:
I am adding a test case. Great find!

PR #443, opened 2015-02-10
Cast boolean when returned as integer

mobileoverlord:


josevalim:

PR #446, opened 2015-02-11
WIP: Test nested transactions.

MSch:
@fishcakez this is for you :)

### DIFF
--- a/integration_test/pg/deadlock_test.exs
+++ b/integration_test/pg/deadlock_test.exs
@@ -96,6 +96,18 @@ defmodule Ecto.Integration.DeadlockTest do
 
         assert_tx_aborted
 
+        # Not really sure what and how to test this
+        try do
+          Process.flag(:trap_exit, true)
+          PoolRepo.transaction fn ->
+          end
+        catch
+          class, msg ->
+            Logger.debug inspect([class, msg])
+        after
+          Process.flag(:trap_exit, false)
+        end
+
         # Even aborted transactions can be rolled back.
         PoolRepo.rollback(:deadlocked)
     else


fishcakez:
Here's a trace of `Ecto.Adapters.SQl.Worker` (external) calls, returns and exceptions for the client/worker pair that do not get the lock.
```
** (Dbg) #PID<0.185.0> calls Ecto.Adapters.SQL.Worker.query!/4 with arguments:
    [#PID<0.169.0>, "SELECT pg_advisory_xact_lock($1);", [2], [timeout: 5000]]

** (Dbg) #PID<0.169.0> calls Ecto.Adapters.SQL.Worker.handle_call/3 with arguments:
    [:query, {#PID<0.185.0>, #Reference<0.0.0.2562>}, %{conn: #PID<0.187.0>, link: #PID<0.185.0>, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 1}]

** (Dbg) #PID<0.169.0> Ecto.Adapters.SQL.Worker.handle_call/3 returns:
    {:reply, {:ok, {Ecto.Adapters.Postgres.Connection, #PID<0.187.0>}}, %{conn: #PID<0.187.0>, link: #PID<0.185.0>, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 1}}

** (Dbg) #PID<0.185.0> Ecto.Adapters.SQL.Worker.query!/4 returns:
    %{columns: ["pg_advisory_xact_lock"], command: :select, num_rows: 1, rows: [{:void}]}

** (Dbg) #PID<0.185.0> calls Ecto.Adapters.SQL.Worker.query!/4 with arguments:
    [#PID<0.169.0>, "SELECT pg_advisory_xact_lock($1);", [1], [timeout: 5000]]

** (Dbg) #PID<0.169.0> calls Ecto.Adapters.SQL.Worker.handle_call/3 with arguments:
    [:query, {#PID<0.185.0>, #Reference<0.0.0.2569>}, %{conn: #PID<0.187.0>, link: #PID<0.185.0>, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 1}]

** (Dbg) #PID<0.169.0> Ecto.Adapters.SQL.Worker.handle_call/3 returns:
    {:reply, {:ok, {Ecto.Adapters.Postgres.Connection, #PID<0.187.0>}}, %{conn: #PID<0.187.0>, link: #PID<0.185.0>, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 1}}

** (Dbg) #PID<0.185.0> Ecto.Adapters.SQL.Worker.query!/4 raises:
    ** (Postgrex.Error) ERROR (40P01): deadlock detected
```
The deadlock is detected!
```
** (Dbg) #PID<0.185.0> calls Ecto.Adapters.SQL.Worker.query!/4 with arguments:
    [#PID<0.169.0>, "SELECT 1", [], [timeout: 5000]]

** (Dbg) #PID<0.169.0> calls Ecto.Adapters.SQL.Worker.handle_call/3 with arguments:
    [:query, {#PID<0.185.0>, #Reference<0.0.0.2585>}, %{conn: #PID<0.187.0>, link: #PID<0.185.0>, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 1}]

** (Dbg) #PID<0.169.0> Ecto.Adapters.SQL.Worker.handle_call/3 returns:
    {:reply, {:ok, {Ecto.Adapters.Postgres.Connection, #PID<0.187.0>}}, %{conn: #PID<0.187.0>, link: #PID<0.185.0>, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 1}}

** (Dbg) #PID<0.185.0> Ecto.Adapters.SQL.Worker.query!/4 raises:
    ** (Postgrex.Error) ERROR (25P02): current transaction is aborted, commands ignored until end of transaction block
```
Another query fails and begin a nested transaction:
```
** (Dbg) #PID<0.185.0> calls Ecto.Adapters.SQL.Worker.begin!/2 with arguments:
    [#PID<0.169.0>, [timeout: 5000]]

** (Dbg) #PID<0.169.0> calls Ecto.Adapters.SQL.Worker.handle_call/3 with arguments:
    [{:begin, [timeout: 5000]}, {#PID<0.185.0>, #Reference<0.0.0.2597>}, %{conn: #PID<0.187.0>, link: #PID<0.185.0>, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 1}]

** (Dbg) #PID<0.185.0> Ecto.Adapters.SQL.Worker.begin!/2 raises:
    ** (Postgrex.Error) ERROR (25P02): current transaction is aborted, commands ignored until end of transaction block
```
This exception isn't caught and will causes the client to attempt a rollback
```
** (Dbg) #PID<0.185.0> calls Ecto.Adapters.SQL.Worker.rollback!/2 with arguments:
    [#PID<0.169.0>, [timeout: 5000]]

** (Dbg) #PID<0.169.0> Ecto.Adapters.SQL.Worker.handle_call/3 returns:
    {:noreply, %{conn: nil, link: nil, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 0}}
```
The worker has finished flushing here! We are not out of sync between client and worker.
```
** (Dbg) #PID<0.169.0> calls Ecto.Adapters.SQL.Worker.handle_info/2 with arguments:
    [{:EXIT, #PID<0.187.0>, :normal}, %{conn: nil, link: nil, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 0}]

** (Dbg) #PID<0.169.0> Ecto.Adapters.SQL.Worker.handle_info/2 returns:
    {:noreply, %{conn: nil, link: nil, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 0}}

** (Dbg) #PID<0.185.0> Ecto.Adapters.SQL.Worker.rollback!/2 raises:
    ** (exit) exited in: GenServer.call(#PID<0.169.0>, {:rollback, [timeout: 5000]}, 5000)
        ** (EXIT) time out
```
The call times out because the worker flushed the request message but the client is still alive. This exception bubbles up to the parent transaction and causes another rollback.
```
** (Dbg) #PID<0.185.0> calls Ecto.Adapters.SQL.Worker.rollback!/2 with arguments:
    [#PID<0.169.0>, [timeout: 5000]]

** (Dbg) #PID<0.169.0> calls Ecto.Adapters.SQL.Worker.handle_call/3 with arguments:
    [{:rollback, [timeout: 5000]}, {#PID<0.185.0>, #Reference<0.0.0.2623>}, %{conn: nil, link: nil, module: Ecto.Adapters.Postgres.Connection, params: [otp_app: :ecto, username: "postgres", password: "postgres", database: "ecto_test", hostname: "localhost"], sandbox: false, transactions: 0}]
```
Unfortunately there is no connection, no link and no transactions in the worker state.

fishcakez:
From the trace and the test I think the following issues exist:

* If the connection process replies to the worker with an error the worker resets its state. However a client might catch the exception when it gets the reply and want to continue.
* When the worker wipes is state it calls `Process.exit/2`, which does not guarantee the client has exited because it could trap exits.
* If the client continues after catching an exception, its next request may or may not be flushed by the worker because `Process.exit/2` is asynchronous and so the client could make the call after the worker flushes and then exit due to the exit signal.
* If the client continues after catching an exception inside a transaction, it will assume the transaction is still active when it is not as the connection has been reset by the worker.
* If `begin!` raises an exception at the start of a transaction `rollback!` will be called even though the transaction did not begin.

josevalim:
Yes, this scenario can totally happen. The only work around that comes to mind is to monitor the process when sending an exit, see if it dies in the next x seconds and if not, crash hard.

Another idea is to simply not care and document people to not trap exits when inside transactions but I can see this backfiring.

Thoughts?

fishcakez:
Is it possible to use monitors instead of links? I feel linking does not allow a client to fail nicely in this example. For example a Phoenix process does not trap exits so would not return a `500`. I will have a think and reply better tomorrow.

fishcakez:
Instead of wiping state it should be possible to keep the caller in the state until it exits or sends a message that it is done. If the connection exits the worker can cause the caller to raise (unless its a "valid" rollback to prevent rollback raising hiding the "real" exception).

josevalim:
@fishcakez so basically we don't wipe state if a transaction operation fails. It makes sense... but what if the transaction operation is the final rollback? Should we store the connection as dirty and close it after the transaction is done?

fishcakez:
Ah I see. I think it requires a cast to tell the worker that transaction has finished, for each nested transaction (that is the only way to safely reduce the transaction count).

I am confused why this isn't required for each transaction, including nested transactions. How does rollback work? I assumed it rolled back to the savepoint of the current transaction. Thus calling rollback twice in a row would be equivalent to calling it once?

josevalim:
This is now in master! Thanks everyone.

PR #447, opened 2015-02-11
Check if the current adapter supports DDL transaction

laurocaetano:
It is necessary to make it possible to add other adapters that doesn't
have this support. i.e MySQL -> http://dev.mysql.com/doc/refman/5.0/en/cannot-roll-back.html

cc/ @josevalim

### DIFF
--- a/lib/ecto/adapter/migrations.ex
+++ b/lib/ecto/adapter/migrations.ex
@@ -26,6 +26,12 @@ defmodule Ecto.Adapter.Migrations  do
   @type ddl_object :: Table.t | Index.t
 
   @doc """
+  Checks if the adapter supports ddl transaction.
+
+  """
+  defcallback supports_ddl_transaction? :: boolean
+
+  @doc """
   Executes migration commands.
 
   ## Options
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -119,4 +119,10 @@ defmodule Ecto.Adapters.Postgres do
     args = args ++ ["--quiet", "--host", database[:hostname], "-d", "template1", "-c", sql_command]
     System.cmd("psql", args, env: env, stderr_to_stdout: true)
   end
+
+  @doc false
+
+  def supports_ddl_transaction? do
+    true
+  end
 end
--- a/test/ecto/migrator_test.exs
+++ b/test/ecto/migrator_test.exs
@@ -212,8 +212,10 @@ defmodule Ecto.MigratorTest do
     end
   end
 
-  test "migrations run inside a transaction by default" do
+  test "migrations run inside a transaction if the adapter supports ddl transactions" do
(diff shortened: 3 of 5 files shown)


lib/ecto/adapter/migrations.ex
```diff
@@ -26,6 +26,12 @@ defmodule Ecto.Adapter.Migrations  do
   @type ddl_object :: Table.t | Index.t
 
   @doc """
+  Checks if the adapter supports ddl transaction.
+
+  """
+  defcallback supports_ddl_transaction?
```
josevalim: Let's also tag the return type as boolean:

     defcallback supports_ddl_transaction? :: boolean

test/ecto/migrator_test.exs
```diff
@@ -212,8 +212,10 @@ defmodule Ecto.MigratorTest do
     end
   end
 
-  test "migrations run inside a transaction by default" do
+  test "migrations run inside a transaction if the adapter supports ddl transactions" do
     capture_log fn ->
+      Process.put(:supports_ddl_transaction?, true)
```
josevalim: I think we should also set this to true here:

https://github.com/laurocaetano/ecto/blob/check-ddl-transactions/test/ecto/migrator_test.exs#L227

Otherwise we won't be testing what it is meant to test. :)

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -381,6 +381,10 @@ if Code.ensure_loaded?(Postgrex.Connection) do
     alias Ecto.Migration.Index
     alias Ecto.Migration.Reference
 
+    def supports_ddl_transaction? do
```
josevalim: Should we delegate this to the connection or just implement it directly in Ecto.Adapters.Postgres? I think there is no need for delegation in this case as long as we don't implement it in `__using__` below?

lib/ecto/adapters/postgres.ex
```diff
@@ -119,4 +119,10 @@ defmodule Ecto.Adapters.Postgres do
     args = args ++ ["--quiet", "--host", database[:hostname], "-d", "template1", "-c", sql_command]
     System.cmd("psql", args, env: env, stderr_to_stdout: true)
   end
+
+  @doc false
+
+  def supports_ddl_transaction? do
```
laurocaetano: @josevalim I just moved the `supports_ddl_transaction?` function to the `Ecto.Adapters.Postgres` module. The problem now is that we are defining the behaviour inside the `Ecto.Adapter.Migrations` module and it will warn us that it is not implemented. Shall we move this behaviour definition to another place?

josevalim: No it is fine as is. :)

josevalim:
:heart_eyes_cat:

laurocaetano:
Thanks :heart:

PR #453, opened 2015-02-12
Move case sensitive tests to @tag :case_sensitive

mobileoverlord:
You mentioned there were 2, but this is the only one failing for me.


josevalim:
Merged manually, thanks!

PR #469, opened 2015-02-17
Problem: No console output when migration is already up/down

c-rack:
If you are already at the highest/lowest migration and run ```mix ecto.migrate``` or ```mix ecto.rollback```, the command returns without any output. That might lead to some confusion for the user.

Technically that might be correct, because the task has not done anything. But it would be more user-friendly to display at least the information that there is no higher/lower version to migrate/rollback to.


josevalim:

Issue #305, opened 2014-10-17
Have Query return a map


chadbean:
When selecting multiple values, is it possible to have Query return a map like {:col1 => val1, :col2 => val2} ?

edit: specifically with the `all` function.

josevalim:
1. Allow maps in [select builder](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/builder/select.ex#L33). We will require keys to be atoms, binaries or numbers.

2. [Extract fields from map in the planner](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/planner.ex#L420)

3. [Generate the map after we got results](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/repo/queryable.ex#L128)

whatyouhide:
@josevalim I'm working on it :)

PR #471, opened 2015-02-17
Support maps in `Query.select/3`

whatyouhide:
This PR implements the feature discussed in #305. It adds support for passing maps (other than tuples, lists and values) to selects:

```elixir
from(p in Post) |> select([p], %{title: p.title, answer: 42})
```

This PR is absolutely **not ready for merging**, but I'm opening it with just a few commits only to know if I'm on the right track.

Some basic things that are still missing (other than some refactoring here and there):

- [x] Documentation for this new feature
- [x] Tests that test the actual usage of maps in selects (I'm still not 100% sure where to place such tests)

Let me know what you think for now!

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -41,6 +41,17 @@ defmodule Ecto.Query.Builder.Select do
     {expr, params}
   end
 
+  # Map
+  defp escape({:%{}, _, pairs}, params, vars) do
+    {pairs, params} = Enum.map_reduce pairs, params, fn({k, v}, acc) ->
+      {expr, params} = escape(v, acc, vars)
+      {{k, expr}, params}
+    end
+
+    expr = {:{}, [], [:%{}, [], pairs]}
+    {expr, params}
+  end
+
   # List
   defp escape(list, params, vars) when is_list(list) do
     Enum.map_reduce(list, params, &escape(&1, &2, vars))
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -45,7 +45,7 @@ defmodule Ecto.Query.Builder.Select do
   defp escape({:%{}, _, pairs}, params, vars) do
     {pairs, params} = Enum.map_reduce pairs, params, fn({k, v}, acc) ->
       unless is_atom(k) or is_binary(k) or is_number(k) do
-        raise ArgumentError, "keys can only be atoms/binaries/integers/floats"
+				Builder.error! "keys can only be atoms/binaries/integers/floats"
       end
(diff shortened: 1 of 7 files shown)


lib/ecto/repo/queryable.ex
```diff
@@ -135,6 +135,15 @@ defmodule Ecto.Repo.Queryable do
     {{left, right}, values}
   end
 
+  defp transform_row({:%{}, _, pairs}, from, values) do
+    {result, values} = Enum.map_reduce pairs, values, fn({k, v}, acc) ->
```
josevalim: You can simply reduce here building both the map and retrieving the values in one pass!

lib/ecto/query/builder/select.ex
```diff
@@ -41,6 +41,17 @@ defmodule Ecto.Query.Builder.Select do
     {expr, params}
   end
 
+  # Map
+  defp escape({:%{}, _, pairs}, params, vars) do
+    {pairs, params} = Enum.map_reduce pairs, params, fn({k, v}, acc) ->
```
josevalim: We need to verify if the key here is atom/binary/integer/float.

josevalim:
Regarding the tests, add some unit tests to the builders/select_test.exs, checking the accepted format and a simple test with a couple keys here:

https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/type.exs#L76-L84

whatyouhide:
Hey @josevalim, should have covered everything you pointed out (thanks for that!). I couldn't test that using keys other than atoms/binaries/integers/floats raises an error since it raises it at compile-time and I wouldn't know how to test that :|.

If we're pretty much ready, I'll write some documentation so we can merge :). Thanks!

josevalim:
We are ready, just docs missing, thanks!

whatyouhide:
@josevalim docs added. Thanks to you!

lib/ecto/query/builder/select.ex
```diff
@@ -41,6 +41,20 @@ defmodule Ecto.Query.Builder.Select do
...
+      unless is_atom(k) or is_binary(k) or is_number(k) do
+        raise ArgumentError, "keys can only be atoms/binaries/integers/floats"
```
josevalim: Sorry, I think there is a Builder.error! "..." that would be more appropriate to raise as it contains more info.

josevalim:
Oops, I have added one tiny comment I missed originally, sorry!

whatyouhide:
@josevalim don't worry, actually I'm sorry! I read the code inside `Builder` including `error!/1`, I should have thought of it. Pushed a new commit!

josevalim:

PR #474, opened 2015-02-18
add macro support in query

### LINKED ISSUE #462: Allow macros in query expressions
Allowing people to extend the query language (like the where contents) by providing custom macros.

briksoftware:
Attempt to close #462. I have only one test for now because I am not sure that this PR addresses #462 in full. If this is going in the right direction, let me know what should be done next

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -23,11 +23,11 @@ defmodule Ecto.Query.Builder do
   with `^index` in the query where index is a number indexing into the
   map.
   """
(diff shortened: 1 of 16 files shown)


josevalim:
This is exactly what needs to be done. I don't think anything else is required. We just need to improve the test to check the macro was actually expanded. You can probably do `Macro.to_string(hd(query.wheres).expr)` or something of sorts.

josevalim:
Fantastic work btw!

briksoftware:
@josevalim: Nice to hear! I completed the test as suggested.

PR #475, opened 2015-02-20
Ecto repo timeout

### LINKED ISSUE #448: Support configurable timeouts
The idea is to store the timeout along side the repository pool:

https://github.com/elixir-lang/ecto/blob/master/lib/ecto/adapters/sql.ex#L31

Then we need to change the calls to `__pool__` in that same module to consider it now returns a tuple and properly use it instead of `@timeout`.

briksoftware:
Closes #448. Let me know if anything is missing.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -175,17 +175,18 @@ defmodule Ecto.Adapters.SQL do
   end
 
   defp pool!(repo) do
-    pid = Process.whereis(elem(repo.__pool__, 0))
+    {pid, timeout} = repo.__pool__
+    pid = Process.whereis(pid)
 
     if is_nil(pid) or not Process.alive?(pid) do
       raise ArgumentError, "repo #{inspect repo} is not started"
     end
 
-    pid
+    {pid, timeout}
   end
 
   defp use_worker(repo, timeout, fun) do
-    pool = pool!(repo)
+    {pool, _} = pool!(repo)
     key  = {:ecto_transaction_pid, pool}
 
     {in_transaction, worker} =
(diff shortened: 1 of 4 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -303,7 +317,7 @@ defmodule Ecto.Adapters.SQL do
   @spec rollback_test_transaction(Ecto.Repo.t, Keyword.t) :: :ok
   def rollback_test_transaction(repo, opts \\ []) do
     pool = pool!(repo)
```
josevalim: I was thinking that instead of calling repo.__pool__ twice, we should make `pool!(repo)` return the repo and timeout. What do you think?

test/ecto/adapters/sql_test.exs
```diff
@@ -10,7 +10,13 @@ defmodule Ecto.Adapters.SQLTest do
...
   end
 
+  defmodule RepoWithTimeout do
+    use Ecto.Repo, adapter: Adapter, otp_app: :ecto
+    Application.put_env(:ecto, __MODULE__, timeout: 1500)
+  end
+
   test "stores __pool__ metadata" do
-    assert Repo.__pool__ == Repo.Pool
+    assert Repo.__pool__ == { Repo.Pool, 5000 }
```
josevalim: No spaces after `{` and before `}`. :)

briksoftware: oops, sorry. fixed

lib/ecto/adapters/sql.ex
briksoftware: very good idea actually, I will adapt this and related tests. Sorry for not noticing this before.

briksoftware: done.

josevalim:

fishcakez:
Would it be a good idea to split poolboy timeout and call timeout?

josevalim:
we could do it later if somebody needs it

PR #476, opened 2015-02-20
Make insert and update return a keyword list

laurocaetano:
The motivation is to make it easier to support other adapters, such as MySQL (that doesn't support the `RETURNING` clause).

### DIFF
--- a/lib/ecto/repo/model.ex
+++ b/lib/ecto/repo/model.ex
@@ -28,7 +28,7 @@ defmodule Ecto.Repo.Model do
 
       {:ok, values} = adapter.insert(repo, source, changes, return, opts)
 
-      changeset = load_into_changeset(changeset, model, return, values)
+      changeset = load_into_changeset(changeset, model, values)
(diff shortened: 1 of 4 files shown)


lib/ecto/repo/model.ex
```diff
@@ -119,9 +119,18 @@ defmodule Ecto.Repo.Model do
...
+    update_in changeset.model, &do_load(struct(&1, changes), model, values)
+  end
+
+  defp do_load(struct, model, kv) do
+    loaded = Enum.reduce(kv, {struct}, fn
+      {k,v}, {acc} ->
+        value = Ecto.Type.load!(model.__schema__(:field, k), v)
+        {Map.put(acc, k, value)}
+    end) |> elem(0) |> Map.put(:__state__, :loaded)
+    Ecto.Model.Callbacks.__apply__(struct.__struct__, :after_load, loaded)
```
laurocaetano: Not sure this callback call here is needed, but there is a test expecting this call.

Issue #355, opened 2015-01-14
Support optimistic locking


josevalim:
The idea here is to provide an `Ecto.Model.OptimisticLocking` module that introduces the macro `optimistic_lock :field_name`.  To implement this feature we will need:

1. Introduce the idea of filters into changesets. Filters will be passed to the underlying insert/update/delete operation and filtering the resource to be modified

2. Those filters will now be handled as part of those operations in the repository [alongside the pkfilter](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/repo/model.ex#L60)

3. Because queries can now return [`{:error, :stale}`](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/adapters/sql.ex#L381), we should match them in the Repo and raise `Ecto.StaleModelError`

4. Now with everything in hand, we can implement the `Ecto.Model.OptimisticLocking`. Once `optimistic_lock :field_name` is called, it will define before_update and before_delete callbacks that will read the value of the field in the changeset model and add it as a filter to the changeset and add a change that increments the lock field by one

5. Docs, tests and profit! $$$

whatyouhide:
Pong, I'd like to give this a shot!

On Friday, February 20, 2015, José Valim <notifications@github.com> wrote:

> /cc @briksoftware <https://github.com/briksoftware> @whatyouhide
> <https://github.com/whatyouhide> If any of you want to tackle it, please
> ping here! :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/elixir-lang/ecto/issues/355#issuecomment-75210101>.
>


-- 

Andrea Leopardi
an.leopardi@gmail.com

josevalim:
Merged, thank you @whatyouhide!

PR #477, opened 2015-02-20
Optimistic locking

whatyouhide:
This PR tackles optimistic locking, as described in #355. I implemented pretty much everything @josevalim mentioned. Docs are still missing and tests could be polished a little, but I'd like a quick review first so that I know I'm making things right :smiley:.

Quick roadmap before merging:

- [x] Documentation
- [x] Review the message given by the `Ecto.StaleModelError` exception


whatyouhide:
Whooops, messed everything up by adding optimistic locking to the `Post` model in the integration tests. Fix on its way!

integration_test/support/migration.exs
```diff
@@ -14,6 +14,11 @@ defmodule Ecto.Integration.Migration do
       timestamps
     end
 
+    create table(:optimistically_locked_posts) do
```
josevalim: Can you please check if we can use any of the other resources instead of defining a new one? Some are used just in a handful of tests and would be a perfect fit.

lib/ecto/exceptions.ex
```diff
@@ -117,3 +117,18 @@ end
...
+
+defmodule Ecto.StaleModelError do
+  defexception [:message]
+
+  def exception(opts) do
+    action = Keyword.fetch!(opts, :action)
+    model = Keyword.fetch!(opts, :model)
+
+    msg = """
+    attempted to #{action} a stale #{model}
```
josevalim: `inspect model`

lib/ecto/model/optimistic_locking.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Ecto.Model.OptimisticLocking do
+  defmacro optimistic_locking(field) do
+    hook_name = String.to_atom("optimistic_locking_#{field}")
```
josevalim: Don't do this here otherwise we are assuming an atom is given at compilation time. Do quote `bind_quoted: [field: field]` and convert it inside the quote. Also, you can simply write `:"optimistic_locking_#{field}"`. :)

lib/ecto/exceptions.ex
whatyouhide: About that, what do you think we should print in the exception message? This just prints the model's module, which may not be ideal.

lib/ecto/repo/model.ex
```diff
@@ -150,11 +165,13 @@ defmodule Ecto.Repo.Model do
     Planner.fields(kind, model, Map.take(changeset.changes, fields))
   end
 
-  defp pk_filter(model, struct) do
+  defp filters!(model, struct, %Changeset{filters: filters}) do
+    cs_filters = Enum.to_list(filters)
```
josevalim: Don't covert to a list. Simply add the primary key using `Map.put` and let `Planner.fields` do the conversion once it iterates.

lib/ecto/model/optimistic_locking.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Ecto.Model.OptimisticLocking do
+  defmacro optimistic_locking(field) do
```
josevalim: English question: should this be called `optimistic_locking` or `optimistic_lock`? It feels we should call it the latter because we are passing the name of the field, so the field is effectively the `optimistic_lock`. It is also a bit shorter. :D

If we change the macro name, we should also change the model one.

lib/ecto/repo/model.ex
```diff
@@ -54,16 +57,23 @@ defmodule Ecto.Repo.Model do
...
+      # Callbacks.__apply__/3 "ignores" undefined callbacks and simply returns
+      # the changeset unchanged in case the callback is missing.
       changeset = Callbacks.__apply__(model, :before_update, changeset)
       changes   = validate_changes(:update, model, fields, changeset)
 
-      pk_filter = Planner.fields(:update, model, pk_filter(model, struct))
+      filters = Planner.fields(:update, model, filters!(model, struct, changeset))
 
       if changes == [] do
-        changes = pk_filter
```
josevalim: This should be only the pk_filter. Basically, we only set this if you are doing an update with no changes, so we basically set the ID equal to the current ID so we can at least send something to the database.

josevalim:
This looks great, I have added some comments!

> Strict checking that fields passed to optimistic_locking/1 are integers

I don't think we need to worry about this. Just document they need to be integers and that is it.

lib/ecto/exceptions.ex
josevalim: Oh, sorry. Definitely pass the whole model (the struct) and print it all. You usually want all info (like lock version) when those things happen.

whatyouhide:
@josevalim should have fixed everything you mentioned (thank you so much for the review!). I also added a lot of documentation, hopefully it's clear enough :). Let me know if I can do anything else!

lib/ecto/model/optimistic_lock.ex
```diff
@@ -0,0 +1,117 @@
+defmodule Ecto.Model.OptimisticLock do
+  @moduledoc """
+  Facilities for using the optimistic-locking technique.
```
josevalim: Fantastic docs.

whatyouhide:
Great, thanks!

PR #479, opened 2015-02-21
Update the docs for Ecto.Model.OptimisticLock

whatyouhide:
This PR updates the docs for the changes in f1f86a9316654f9bf419b550ca2ea33cd793aafb. Before, the `Ecto.Model.OptimisticLock` module needed to be `import`'ed but now it needs to be `use`'d.


josevalim:
Ah, yes, good catch. I went with use just for consistency (all others in Ecto.Model need to).

PR #487, opened 2015-02-27
Alias Ecto.Changeset inside Ecto.Changeset

whatyouhide:
With this PR, the `Ecto.Changeset` struct can be referenced as just `Changeset` inside the `Ecto.Changeset` module. I also updated the specs to accept `t` instead of `Ecto.Changeset.t`.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -21,6 +21,8 @@ defmodule Ecto.Changeset do
   * `filters`     - Filters (as a map `%{field => value}`) to narrow the scope of update/delete queries
   """
 
+  alias __MODULE__
+
   import Ecto.Query, only: [from: 2]
 
   defstruct valid?: false, model: nil, params: nil, changes: %{}, repo: nil,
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -28,16 +30,16 @@ defmodule Ecto.Changeset do
...
-                             repo: atom | nil,
-                             model: Ecto.Model.t | nil,
-                             params: %{String.t => term} | nil,
-                             changes: %{atom => term},
-                             required: [atom],
-                             optional: [atom],
-                             errors: [error],
-                             validations: [{atom, atom | {atom, [term]}}],
-                             filters: %{atom => term}}
+  @type t :: %__MODULE__{valid?: boolean(),
```
josevalim: Why not Changeset here then?

whatyouhide: I looked at `Plug.Conn` and it does the same thing there, `__MODULE__` in the struct but `%Conn{}` in the functions. Should I change this?

josevalim: Yes, please. Since we have an alias, let's use it.

whatyouhide:
@josevalim ok, fixed a squashed the commit!

PR #490, opened 2015-02-28
Removing `validate_number` from Ecto lib docs.

djm:
As the function no longer exists (for now).

I've added in an example for `validate_inclusion` instead so that the :age related comment in the prose below still makes sense.


josevalim:

PR #502, opened 2015-03-04
Add source to has assoc

### LINKED ISSUE #478: Allow {source, model} to be given as queryable

Or

tony612:
resolve #478

Please review @josevalim . There's a problem: `source` must implement Ecto.Queryable, but I don't know how to judge this. Is there a way for this? Or should we just limit `source` to be a string?

### DIFF
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -62,6 +62,9 @@ defmodule Ecto.Query.Builder.From do
           # When a binary is used, there is no model
           {1, query(source, nil)}
 
+        {source, model} when is_binary(source) and is_atom(model) ->
+          {1, query(source, model)}
+
         other ->
           {nil, other}
       end
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -17,6 +17,9 @@ defmodule Ecto.Query.Builder.From do
       iex> escape(quote do: p in posts)
       {[p: 0], quote(do: posts)}
 
+      iex> escape(quote do: p in {"posts", MyModel})
+      {[p: 0], quote(do: {"posts", MyModel})}
+
       iex> escape(quote do: [p, q] in posts)
       {[p: 0, q: 1], quote(do: posts)}
 
--- a/lib/ecto/query/builder/join.ex
+++ b/lib/ecto/query/builder/join.ex
@@ -21,6 +21,9 @@ defmodule Ecto.Query.Builder.Join do
       iex> escape(quote(do: x in Sample), [])
       {:x, {nil, {:__aliases__, [alias: false], [:Sample]}}, nil}
 
+      iex> escape(quote(do: x in {"foo", Sample}), [])
+      {:x, {"foo", {:__aliases__, [alias: false], [:Sample]}}, nil}
+
       iex> escape(quote(do: c in assoc(p, :comments)), [p: 0])
       {:c, nil, {0, :comments}}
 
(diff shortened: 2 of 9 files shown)


josevalim:
Hrm, @tony612, this is the wrong approach to the problem. As the issue says, we need to first make `{"source", model}` queryable, this way, adding support for associations is really easy. Making it queryable requires some steps:

1. implementing the queryable protocol for it
2. ensure the query planner knows how to cope with this new format
3. ensure the inspector knows how to distinguish and prints such queryable properly

Just then it makes sense for them to be part of the association.

josevalim:
Btw, the original issue is marked as advanced exactly because it requires going through a couple places in ecto internals. :)

tony612:
@josevalim So should we implement it in both formats(`{"source", Model}` & `Model, source: "source"`), right? I just thought we can pick one format :sweat_smile: 

All right, I'll look into this problem again according to your guide :)

josevalim:
@tony612 first we need to make it work in queries and the only syntax that makes sense in queries is `{"source", Model}`. We can discuss the association syntax later.

tony612:
@josevalim I rewrite my code after reading more doc and code. But I have two problems:
1. My implement broke some Inspect tests.
Inspecting `from p in Post` will become `from p in {"posts", Post}` because `query.from` of query `Post` will be expanded to `{"posts", Post}`. I'm not sure if we should keep the original way. If so, I need to find another way to inspect `{"source", model}`. 
2. I read the code of Planner module, but can't find where to modify. Current Planner seems to work well with new `{"source", Model}` query.

Any idea?

tony612:
I just think of one way to solve problem 1: compare `from`'s `source` with `model.__schema__(:source)`. It works, but I'm not sure it's a good enough

lib/ecto/query/builder/from.ex
```diff
@@ -62,6 +62,9 @@ defmodule Ecto.Query.Builder.From do
           # When a binary is used, there is no model
           {1, query(source, nil)}
 
+        {source, model} when is_binary(source) and is_atom(model) ->
```
josevalim: We need to support something similar in the join builder too. :)

josevalim:
> I just think of one way to solve problem 1: compare from's source with `model.__schema__(:source)`. It works, but I'm not sure it's a good enough

Yup, I think this is the way to go indeed. Good to know no changes are required in the planner. To be generally sure, do a search on `model.__schema__(:source)` and see if there are places assuming lack of source or model.

josevalim:
For example, once we support this in joins too, this code is wrong (as it assumes there is no model when there is a source):

https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/planner.ex#L237

josevalim:
Good job btw! This is a tricky feature and you are progressing well! :D

tony612:
@josevalim I implemented this in joins and added more test/doc. I think it's time to review :)

lib/ecto/query/builder/join.ex
```diff
@@ -40,6 +43,10 @@ defmodule Ecto.Query.Builder.Join do
     {:_, {string, nil}, nil}
   end
 
+  def escape({string, {:__aliases__, _, _} = module}, _vars) when is_binary(string) do
```
josevalim: We also need a `{string, atom}` clause. :)

tony612: @josevalim thx for this. Now it's done. Please review :D

lib/ecto/query/builder/from.ex
tony612: @josevalim Hi, I have a question here.If I change this segment of code to this
```elixir
        {source, {:__aliases__, _, _} = module} when is_binary(source) ->
          {1, query(source, module)}
```
It seems work and the tests pass. Why?

josevalim: This code here is just an optimization. If it can't expand it at compile time, it will fallback to expand at runtime.

PR #507, opened 2015-03-07
[WIP] Add support for microseconds in `Ecto.Time` and `Ecto.DateTime`

whatyouhide:
This PR implements support for microseconds (`usec`) in the `Ecto.Time` and `Ecto.DateTime` structs, as discussed in #491.

I'd like a small review to see if I'm headed in the right direction. For example, `:usec` would be a field for which a `0` default would make sense, but I'm not sure so I'd like additional opinions :). I'd also like to refactor a bit (I can do that in another PR if necessary).

Thanks so much!

### DIFF
--- a/lib/ecto/datetime.ex
+++ b/lib/ecto/datetime.ex
@@ -238,7 +238,7 @@ defmodule Ecto.DateTime do
   """
 
   @behaviour Ecto.Type
-  defstruct [:year, :month, :day, :hour, :min, :sec]
+  defstruct [:year, :month, :day, :hour, :min, :sec, :usec]
 
   @doc """
   The Ecto primitive type.
@@ -250,9 +250,9 @@ defmodule Ecto.DateTime do
   """
   def cast(<<year::4-bytes, ?-, month::2-bytes, ?-, day::2-bytes, sep,
              hour::2-bytes, ?:, min::2-bytes, ?:, sec::2-bytes, rest::binary>>) when sep in [?\s, ?T] do
-    if usec(rest) do
+    if usec = usec(rest) do
       from_parts(to_i(year), to_i(month), to_i(day),
-                 to_i(hour), to_i(min), to_i(sec))
+                 to_i(hour), to_i(min), to_i(sec), usec)
     else
       :error
     end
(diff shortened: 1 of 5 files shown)


lib/ecto/datetime.ex
```diff
@@ -167,29 +167,30 @@ defmodule Ecto.Time do
   def cast(%Ecto.Time{} = t),
     do: {:ok, t}
   def cast(%{"hour" => hour, "min" => min} = map),
-    do: from_parts(to_i(hour), to_i(min), to_i(Map.get(map, "sec", 0)))
+    do: from_parts(to_i(hour), to_i(min), to_i(Map.get(map, "sec", 0)), 0)
```
josevalim: We should do `to_i(Map.get(map, "usec", 0))` here too. :)

lib/ecto/datetime.ex
```diff
@@ -214,12 +215,17 @@ defmodule Ecto.Time do
...
-    %Ecto.Time{hour: hour, min: min, sec: sec}
+    %Ecto.Time{hour: hour, min: min, sec: sec, usec: 0}
   end
 
   defimpl String.Chars do
-    def to_string(%Ecto.Time{hour: hour, min: min, sec: sec}) do
-      zero_pad(hour, 2) <> ":" <> zero_pad(min, 2) <> ":" <> zero_pad(sec, 2)
+    def to_string(%Ecto.Time{hour: hour, min: min, sec: sec, usec: usec}) do
+      str = zero_pad(hour, 2) <> ":" <> zero_pad(min, 2) <> ":" <> zero_pad(sec, 2)
+      if is_nil(usec) or usec == 0 do
```
josevalim: This is meant to be more of a human readable representation. I think we should leave usec out. Wdyt?

lib/ecto/datetime.ex
```diff
@@ -232,7 +238,7 @@ defmodule Ecto.DateTime do
   """
 
   @behaviour Ecto.Type
-  defstruct [:year, :month, :day, :hour, :min, :sec]
+  defstruct [:year, :month, :day, :hour, :min, :sec, :usec]
```
josevalim: Let's default to 0, yes.

lib/ecto/datetime.ex
```diff
@@ -258,37 +264,37 @@ defmodule Ecto.DateTime do
 
   def cast(%{"year" => year, "month" => month, "day" => day, "hour" => hour, "min" => min} = map) do
     from_parts(to_i(year), to_i(month), to_i(day),
-               to_i(hour), to_i(min), to_i(Map.get(map, "sec", 0)))
+               to_i(hour), to_i(min), to_i(Map.get(map, "sec", 0)), 0)
```
josevalim: Let's `to_i(Map.get(map, "usec", 0))` here too. :)

lib/ecto/datetime.ex
```diff
@@ -258,37 +264,37 @@ defmodule Ecto.DateTime do
...
   end
 
   def cast(%{year: year, month: month, day: day, hour: hour, min: min} = map) do
     from_parts(to_i(year), to_i(month), to_i(day),
-               to_i(hour), to_i(min), to_i(Map.get(map, :sec, 0)))
+               to_i(hour), to_i(min), to_i(Map.get(map, :sec, 0)), 0)
```
josevalim: And here!

lib/ecto/datetime.ex
```diff
@@ -337,13 +343,19 @@ defmodule Ecto.DateTime do
...
-      zero_pad(year, 4) <> "-" <> zero_pad(month, 2) <> "-" <> zero_pad(day, 2) <> "T" <>
-      zero_pad(hour, 2) <> ":" <> zero_pad(min, 2) <> ":" <> zero_pad(sec, 2) <> "Z"
+    def to_string(%Ecto.DateTime{year: year, month: month, day: day, hour: hour, min: min, sec: sec, usec: usec}) do
+      str = zero_pad(year, 4) <> "-" <> zero_pad(month, 2) <> "-" <> zero_pad(day, 2) <> "T" <>
+            zero_pad(hour, 2) <> ":" <> zero_pad(min, 2) <> ":" <> zero_pad(sec, 2)
+
+      if is_nil(usec) or usec == 0 do
+        str <> "Z"
+      else
+        str <> "." <> zero_pad(usec, 6) <> "Z"
```
josevalim: Actually, this is meant to be ISO 8601. And afaik it doesn't handle usec.

lib/ecto/datetime.ex
josevalim: Actually, this is meant to be ISO 8601. And afaik it doesn't handle usec.

josevalim:
Thank you, this looks great! I have posted some comments but they are mostly small additions.

lib/ecto/datetime.ex
whatyouhide: I checked the [ISO 8601 page](http://en.wikipedia.org/wiki/ISO_8601) on Wikipedia before adding that. Quoting:

> Decimal fractions may be added to any of the three time elements. However, a fraction may only be added to the lowest order time element in the representation. A decimal mark, either a comma or a dot (without any preference as stated in resolution 10 of the 22nd General Conference CGPM in 2003, but with a preference for a comma according to ISO 8601:2004) is used as a separator between the time element and its fraction. To denote "14 hours, 30 and one half minutes", do not include a seconds figure. Represent it as "14:30,5", "1430,5", "14:30.5", or "1430.5". There is no limit on the number of decimal places for the decimal fraction. However, the number of decimal places needs to be agreed to by the communicating parties.

So it *is* allowed. Nonetheless, if that's for humans it could be left out :) I implemented so that it adds microseconds to the representation only if `:usec` is not `0`. What do you think?

josevalim: Nice! It seems we need to add the Z regardless though.

whatyouhide:
@josevalim pushed a commit which should fix all the stuff you mentioned. Thanks so much for reviewing :heart:

PR #508, opened 2015-03-07
Add microseconds to "Ecto.Schema.timestamps/0" timestamps.

lau:
For "Ecto.Schema.timestamps/0": Use timestamps with microseconds instead of always setting microseconds to 0.

### DIFF
--- a/lib/ecto/model/timestamps.ex
+++ b/lib/ecto/model/timestamps.ex
@@ -23,9 +23,15 @@ defmodule Ecto.Model.Timestamps do
     if get_change changeset, field do
       changeset
     else
-      {date, {h, m, s}} = :erlang.universaltime
-      put_change changeset, field, Ecto.Type.load!(type, {date, {h, m, s, 0}})
+      put_change changeset, field, Ecto.Type.load!(type, timestamp_tuple_with_usec)
     end
+  end
+
+  defp timestamp_tuple_with_usec do
+    erl_timestamp = :os.timestamp
+    {_, _, usec} = erl_timestamp
+    {date, {h, m, s}} =:calendar.now_to_datetime(erl_timestamp)
+    {date, {h, m, s, usec}}
   end
 
   defmacro __before_compile__(env) do


josevalim:
Thanks @lau. I have thought about this previously and I wondered: do we really need this precision? Most database drivers end-up using more compact formats when you don't have usecs. Adding usecs also takes a bit more to compute and serialize. Furthermore, data received from the application front (CLI, forms, apis) likely won't have microseconds. So I am more inclined to not add this precision unless someone says they really need (then they can have their own callbacks or we can support an option). What do you think?

lau:
Yeah, it doesn't seem unreasonable  to not have microseconds by default for these timestamps if it makes a meaningful performance difference.

Updated at and inserted at will be generated by Ecto and not based on input from external sources such as APIs and forms right?

Microseconds can be useful for instance for sorting. Imagine a chat application where there is more than one message per second. With Phoenix and websockets "real time like" applications can make use of precision that's greater than one second.

We have usec now but it is set to 0, right? Do the database drivers use more compact format to store the timestamps now?

ericmj:
We should default to as much as precision as possible. Both mysql and postgrex uses binary protocols now so between application and db server there should be virtually no extra cost.

mobileoverlord:
Tds will be fine with it. Microsoft is pushing everyone to use datetime2 for all new work as it's more a standard.

josevalim:

josevalim:
Gah, mysql does not support this. :( Reverting until we figure it out.

lau:
It seems like MySQL only supports microseconds in 5.7+, but mariaex doesn't support them for any version. Would it not be OK to just let MySQL save datetime and timestamps without microseconds? E.g. a datetime with microseconds is received, but the adapter discards the microseconds/sets them to 0.

Are exceptions raised?

josevalim:
It breaks all integration tests though. So we either need to clean up the integration test suite in a way it won't affect us or figure out another strategy.

laurocaetano:
Actually Mariaex does support microseconds, see: https://github.com/xerions/mariaex/blob/master/lib/mariaex/messages.ex#L234-L235

It will always send the microseconds to the database, but it only will be stored in MySQL 5.7+.

btw.. The integration test fails because the MySQL version running on Travis does not support microseconds.

lau:
When I ran this locally on the branch for this PR, it didn't fail: MIX_ENV=mysql mix test  (that should run the mysql integration test, right?) Running on MySQL 5.6.16
This didn't fail: https://travis-ci.org/elixir-lang/ecto/builds/53486771

But looking at travis, it was the next PR that failed. I will look at the integration tests that failed and report back.

lau:
MySQL 5.6 seems to use the microseconds round up seconds :dizzy_face: So if you send it 14:32:10.700000 it will save it as 14:32:11 ! Am I the only one that thinks that's crazy? Imagine if you tell MySQL to save 2015-12-31 23:59:59.700000. It would be saved as January 1st 2016 instead!

Here are two options:

1) Since MySQL 5.7 is not a general release yet, maybe it would be a good idea for mariaex to simply set all microseconds to 0. And then ignore microseconds in the Ecto integration tests.

or

2) Make travis use MySQL 5.7

laurocaetano:
Yes, it will round up but only in cases that the column has fewer fractional digits, otherwise it will save with full precision.

Example:

```SQL
/* Having columns with less precision */

mysql> CREATE TABLE fractest( c1 TIME(2), c2 DATETIME(2), c3 TIMESTAMP(2) );
Query OK, 0 rows affected (0.33 sec)

mysql> INSERT INTO fractest VALUES 
     > ('17:51:04.777', '2014-09-08 17:51:04.777', '2014-09-08 17:51:04.777');
Query OK, 1 row affected (0.03 sec)

mysql> SELECT * FROM fractest;
+-------------+------------------------+------------------------+
| c1          | c2                     | c3                     |
+-------------+------------------------+------------------------+
| 17:51:04.78 | 2014-09-08 17:51:04.78 | 2014-09-08 17:51:04.78 |
+-------------+------------------------+------------------------+
1 row in set (0.00 sec)


/* Having columns with full precision*/

mysql> CREATE TABLE fractest( c1 TIME(6), c2 DATETIME(6), c3 TIMESTAMP(6) );
Query OK, 0 rows affected (0.02 sec)

mysql> INSERT INTO fractest VALUES ('17:51:04.777000', '2014-09-08 17:51:04.777000', '2014-09-08 17:51:04.777000'); Query OK, 1 row affected (0.00 sec)

mysql> SELECT * FROM fractest;
+-----------------+----------------------------+----------------------------+
| c1              | c2                         | c3                         |
+-----------------+----------------------------+----------------------------+
| 17:51:04.777000 | 2014-09-08 17:51:04.777000 | 2014-09-08 17:51:04.777000 |
+-----------------+----------------------------+----------------------------+
1 row in set (0.00 sec)
```

 See: http://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html

lau:
@laurocaetano Ah, thanks, good to know. It seems that MySQL 5.6 supports microseconds, not just 5.7! I made the integrations tests run by change the datetime to datetime(6) in the adapter :) I'll make a new PR.

PR #510, opened 2015-03-08
Revert "Add microseconds to "Ecto.Schema.timestamps/0" timestamps."

josevalim:
Reverts elixir-lang/ecto#508

Unfortunately not supported by MySQL.


jamescmartinez:
Hi @josevalim! Is `usec: false` going to remain the default for timestamps? It seemed as though the discussion over here was in favor of making it default to true, though I may definitely be misunderstanding. 😄  https://github.com/elixir-ecto/ecto/pull/508

michalmuskala:
Yes, `usec: false` will remain the default, since it's not supported by many databases.

PR #513, opened 2015-03-10
Change/clean validation error handling.

artichokes:
Errors now look as such:
```
[
  post: [
    must_be_longer_than: 2,
    just_another_test: nil
  ],
  number: [
    is_42: false
  ]
]
```

Adding errors to a field will concatenate them to the existing error key or create a new key for the field.

ALL errors must have the type {atom, term}. This means that single atom errors must have `nil` as the second index.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -29,7 +29,7 @@ defmodule Ecto.Changeset do
             errors: [], validations: [], required: [], optional: [],
             filters: %{}
 
-  @type error :: {atom, atom | {atom, [term]}}
+  @type error :: {atom, term}
   @type t :: %Changeset{valid?: boolean(),
                         repo: atom | nil,
                         model: Ecto.Model.t | nil,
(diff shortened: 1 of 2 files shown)


artichokes:
Whoops! Forgot to change `cast/4` and `merge/2`.

artichokes:
Should errors be merged, e.g.,
```
changeset = change(%Post{}, %{title: ""})
  |> add_error(changeset, :title, {must_be_longer_than: 2})
  |> add_error(changeset, :title, {must_be_longer_than: 3})
```
Should it result in:
`[title: [must_be_longer_than: 3]]` or
`[title: [must_be_longer_than: 2, must_be_longer_than: 3]]`?

josevalim:
That's somewhat how they were used to work but I would prefer to keep the errors in the order they are defined which is how it works today. If we are going to move to this direction, then we should make them maps, but maps are somewhat awkward to match on for the empty case.

Let me pull @drewolson and @jeregrine to see if they have any preference.

Thanks for the pull request!

drewolson:
I'm fine with the current error structure. It's easy enough to write a helper that joins errors together by field name for display purposes -- this could even be provided by phoenix.

I find the single atom error case with a value of `nil` pretty bizarre, seems like an indication that this solution has some gaps.

josevalim:
@drewolson if you have the field name, you can easily group them with `Keyword.get_values/2`. General grouping can done easily with `Enum.group_by/2`. :) Closing this then!

PR #514, opened 2015-03-10
support assoc with custom source for Has/BelongsTo

### LINKED ISSUE #478: Allow {source, model} to be given as queryable

tony612:
resolve #478 @josevalim I add it to `BelongsTo` association besice `Has` because I think it also needs it :)

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -210,10 +210,11 @@ defmodule Ecto.Association.Has do
     * `assoc` - The model that is associated
     * `owner_key` - The key on the `owner` model used for the association
     * `assoc_key` - The key on the `associated` model used for the association
+    * `queryable` - The real query to use for querying association
   """
 
   @behaviour Ecto.Association
-  defstruct [:cardinality, :field, :owner, :assoc, :owner_key, :assoc_key]
+  defstruct [:cardinality, :field, :owner, :assoc, :owner_key, :assoc_key, :queryable]
 
   @doc false
   def struct(module, name, opts) do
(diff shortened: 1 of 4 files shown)


josevalim:
Yes, definitely both need it. :)

lib/ecto/association.ex
```diff
@@ -233,11 +234,15 @@ defmodule Ecto.Association.Has do
...
     end
 
-    assoc = Keyword.fetch!(opts, :queryable)
+    queryable = Keyword.fetch!(opts, :queryable)
 
-    unless is_atom(assoc) do
-      raise ArgumentError, "association queryable must be a model, got: #{inspect assoc}"
-    end
+    assoc =
+      case queryable do
```
josevalim: What if we extract this to a function in `Ecto.Association` so both associations can use?

tony612: Yeah, I have this idea, but not sure if we need it in `BelongTo`. I'll refactor this :)

josevalim: If we don't need it in belongs to, then we don't need to worry about it. :)

josevalim:
I have also just realized this is not complete yet. We will have to store the source from the association and, when building the association, we will need to store the source inside the model. For example, if I do this:

    build(post, :comments)

And comments is specified as:

    has_many :comments, {"my_comments", Comment}

`"my_comments"` should be stored somewhere in the built model so when we do the Repo.insert/update/delete, it will change to the proper table. The problem is that we have nowhere to store this in the model and we already have a `__state__` field, so I wanted to avoid adding more. So I thought we could do this:

1. Create a `__meta__` field that will have a `Ecto.Schema.Metadata` struct
2. The Metadata struct will have two fields: :state and :source
3. We will remove the current `__state__` to use the new `__meta__`
4. Associations and loading the resource should set the proper metadata

Since this seems to be another big task, we can do it as part of another pull request, let's get this one merged first. What do you think? Can you tackle this one too? :D

Thanks a lot, this is shaping up fantastic!

tony612:
@josevalim Thanks for your detailed reply. I'll look into this and try it in another PR later :)

tony612:
@josevalim Done with refactoring. I think this can be merged. Then I'll begin with the `build` thing :sunglasses:

josevalim:

PR #515, opened 2015-03-11
Fix mysql datetime and re-add microsecond precision to timestamps

lau:
The default datetime type in MySQL does not have microsecond precision (at least in v5.6).
The first commit fixes this by using datetime(6). With this you avoid MySQL rounding up seconds when any Ecto.DateTime with e.g. usec set to 700000 is saved.

The other commit adds microsecond precision back to timestamps.

If travis runs at least MySQL 5.6, integrations test should pass.

### DIFF
--- a/lib/ecto/model/timestamps.ex
+++ b/lib/ecto/model/timestamps.ex
@@ -23,9 +23,15 @@ defmodule Ecto.Model.Timestamps do
     if get_change changeset, field do
       changeset
     else
-      {date, {h, m, s}} = :erlang.universaltime
-      put_change changeset, field, Ecto.Type.load!(type, {date, {h, m, s, 0}})
+      put_change changeset, field, Ecto.Type.load!(type, timestamp_tuple_with_usec)
     end
+  end
+
+  defp timestamp_tuple_with_usec do
+    erl_timestamp = :os.timestamp
+    {_, _, usec} = erl_timestamp
+    {date, {h, m, s}} =:calendar.now_to_datetime(erl_timestamp)
+    {date, {h, m, s, usec}}
   end
 
   defmacro __before_compile__(env) do
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1,8 +1,11 @@
+Code.require_file "../support/date_time.exs", __DIR__
+
 defmodule Ecto.Integration.RepoTest do
   use Ecto.Integration.Case
 
   require Ecto.Integration.TestRepo, as: TestRepo
   import Ecto.Query
+  import Ecto.Integration.DateTime
 
   alias Ecto.Integration.Post
   alias Ecto.Integration.Comment
@@ -163,7 +166,9 @@ defmodule Ecto.Integration.RepoTest do
 
   test "get(!)" do
     post1 = TestRepo.insert(%Post{title: "1", text: "hai"})
+      |> round_datetime_if_needed(TestRepo.adapter_supports_microsconds?)
     post2 = TestRepo.insert(%Post{title: "2", text: "hai"})
+      |> round_datetime_if_needed(TestRepo.adapter_supports_microsconds?)
 
     assert post1 == TestRepo.get(Post, post1.id)
     assert post2 == TestRepo.get(Post, to_string post2.id) # With casting
@@ -188,7 +193,9 @@ defmodule Ecto.Integration.RepoTest do
 
   test "one(!)" do
     post1 = TestRepo.insert(%Post{title: "1", text: "hai"})
+      |> round_datetime_if_needed(TestRepo.adapter_supports_microsconds?)
     post2 = TestRepo.insert(%Post{title: "2", text: "hai"})
+      |> round_datetime_if_needed(TestRepo.adapter_supports_microsconds?)
 
     assert post1 == TestRepo.one(from p in Post, where: p.id == ^post1.id)
     assert post2 == TestRepo.one(from p in Post, where: p.id == ^to_string post2.id) # With casting
@@ -355,7 +362,9 @@ defmodule Ecto.Integration.RepoTest do
 
   test "joins" do
     p1 = TestRepo.insert(%Post{title: "1"})
+      |> round_datetime_if_needed(TestRepo.adapter_supports_microsconds?)
     p2 = TestRepo.insert(%Post{title: "2"})
+      |> round_datetime_if_needed(TestRepo.adapter_supports_microsconds?)
     c1 = TestRepo.insert(%Permalink{url: "1", post_id: p2.id})
 
     query = from(p in Post, join: c in assoc(p, :permalink), order_by: p.id, select: {p, c})
@@ -367,6 +376,7 @@ defmodule Ecto.Integration.RepoTest do
 
   test "has_many association join" do
     post = TestRepo.insert(%Post{title: "1", text: "hi"})
+      |> round_datetime_if_needed(TestRepo.adapter_supports_microsconds?)
     c1 = TestRepo.insert(%Comment{text: "hey", post_id: post.id})
     c2 = TestRepo.insert(%Comment{text: "heya", post_id: post.id})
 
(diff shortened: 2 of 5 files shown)


lib/ecto/adapters/mysql/connection.ex
```diff
@@ -514,7 +514,7 @@ if Code.ensure_loaded?(Mariaex.Connection) do
 
     defp ecto_to_db(:string),     do: "varchar"
     defp ecto_to_db(:float),      do: "double"
-    defp ecto_to_db(:datetime),   do: "datetime"
+    defp ecto_to_db(:datetime),   do: "datetime(6)"
```
laurocaetano: It will crash (invalid syntax) for users running MySQL `< 5.6` :(

laurocaetano: One possible solution would be checking if the adapter supports microseconds. Something like: https://github.com/elixir-lang/ecto/commit/9860c362fa97a166b8868a97d9fcc87eb6bcca35, but I haven't checked if it's possible in this case.

lau: It would be nice if it was easy to support older versions too. But MySQL 5.6 existed in 2011. It's older than Elixir. Does Ecto need to be held back to support MySQL versions from 2011?

Maybe Ecto should simply require MySQL 5.6+

laurocaetano: I think it's worth supporting older MySQL versions, because not all applications would be brand new. But I'm not sure about what the Ecto team think about it.
cc/ @josevalim @ericmj

josevalim: People don't upgrade databases as often as everything else as it requires replication and complicated setup if you want zero downtime updates and we need to run on legacy databases just as well. Given we were considering at least SQL Server 2008 for TDS, MySQL 5.6 is too recent to restrict only to it.

lib/ecto/model/timestamps.ex
```diff
@@ -23,11 +23,23 @@ defmodule Ecto.Model.Timestamps do
     if get_change changeset, field do
       changeset
     else
-      {date, {h, m, s}} = :erlang.universaltime
-      put_change changeset, field, Ecto.Type.load!(type, {date, {h, m, s, 0}})
+      put_change changeset, field, Ecto.Type.load!(type, timestamp_tuple_with_usec(changeset.repo.adapter))
     end
   end
 
+  defp timestamp_tuple_with_usec(Ecto.Adapters.MySQL) do
```
lau: OK, this passes integration tests on Travis. Instead of looking at the name of the adapter, we could have a boolean function in adapters to determine whether there is consistent microsecond support. What would be a good name for it?

`supports_microsecond_datetimes?` ?

josevalim: Ah, this is interesting. We shouldn't read the adapter configuration though because it is not cached and it is going to affect considerably the insertion/update times. We would need to think about an efficient way of doing this.

laurocaetano: To correctly implement this function, we must have to know what MySQL version is being used, right? Not sure how it would be implemented and if it's going to be efficient.

Another solution would be passing the `precision` option to these date types in the migration, instead of manually setting it to `datetime(6)`.

lau: Mariaex will get date times with microseconds, send them along to MySQL, who might choose to round up to another datetime. Like 2015-12-31 23:59:59.6 to 2016-01-01 00:00:00. This is not just an issue for these automatic timestamps but for any datetime use with microseconds and mysql.

We could change the integration tests to ignore microseconds and seconds when the adapter is MySQL. (Or allow one second difference). But then if the second is 59 and rounded up, then the minute will also be wrong. This can cascade up until the year.

So we could just say: no microseconds when using MySQL and Ecto. Either mariaex or the mysql adapter replaces any microsecond with 0 before sending them along to mysql.

It makes sense that MySQL support should be for legacy databases that use mysql 5.5 or older. Hopefully people will use postgres for new projects.

josevalim: > It makes sense that MySQL support should be for legacy databases that use mysql 5.5 or older. Hopefully people will use postgres for new projects.

Haha, :+1:.

lau:
OK, the library code is now simple again and does not care which adapter is used.

The integration tests have been fixed, so that when MySQL is used, usec will be expected to be 0.

lau:
@josevalim I have gone back to the original change (that was merged and later reverted) and simply fixed the integration tests by expecting MySQL to return 0 microseconds. What do you think?

By the way right now in master there seems to be no general integration test for microseconds and date times. If there were, mysql integration tests would fail unless they (like in this PR) expected MySQL to return 0 microseconds, while other adapters return the microseconds passed to them.

josevalim:
@lau this sounds like a good direction to go, thanks! We need to find a way though to remove the places where we hardcode the MySQL adapter though. Those integration tests are shared by multiple adapters, including third-party ones. So let's suppose someone is writing an adapter for IBM's DBI and they have the same issue. We need to have a solution which does not require we hardcoding the DBI adapter here too. I can probably investigate this further later.

lau:
We could add a boolean function to the adapters: preserves_microseconds?

And then in the tests ask if the adapter preserves microseconds instead of asking if it is called MySQL.

josevalim:
We should not make the adapter API larger because of MySQL though. Maybe it could be added to the repository only for testing but, even though, is still the same idea that we are making everyones lives harder because of MySQL. :(

Maybe we can use ExUnit tags or configuration for that which is then specific only to testing.

lau:
I have made it more generic. The support file is now called date_time.exs instead of mysql.exs.
Also the integration file has been cleaned up a bit. Now the MySQL adapter name is hardcoded in the date_time.exs file instead and only in one place: in that support file.

So maybe we could move that hardcoding to ExUnit configuration somehow?

josevalim:
> So maybe we could move that hardcoding to ExUnit configuration somehow?

I think it is possible but I am not sure how on top of my head.

lau:
If someone wants to make another adapter for a database that doesn't support microseconds, how would they ideally add their adapter to the list of adapters for which microseconds will return 0? With a command line argument?

josevalim:
In the best scenario, some option in their ExUnit.start (see the MySQL test
helper for actual examples). In the worst scenario, some extra function in
their test helper.

-- 


lau:
Now it's fixed by calling a function in the helpers. I looked at the ExUnit docs and couldn't find a way to assign arbitrary options to ExUnit.start (and retrieve them again).

josevalim:
Thank you! I will merge this later when I get to work on Ecto again!

lau:
My pleasure. Great :)

josevalim:

josevalim:
So I merged this and it fails locally because I use MySQL 5.6 which has the complicated rounding approach. :( I think all of our tests should pass for the most recent MySQL versions.

josevalim:
I have reverted this change here: 19dc4a567231a782a4aabe23ca9cbfede5f100ea. Let's please hold on this please. :) We already had some back and forth and I am afraid things will get even worse once we add more adapters.

lau:
All of this is caused by inconsistent behavior from MySQL / the MySQL adapter depending on the version, which is not being tested. So the inconsistency is still there, we just don't have any tests describing and revealing them :( Because there are no integration tests with microseconds being anything other than 0.

josevalim:
@lau If we don't generate microseconds, there is no code in Ecto that can cause the aforementioned problem. The only option for running into this is if you explicitly set the usec before writing.

So I agree we should improve Ecto.DateTime docs to add a note about how different databases handle usec but not doing it by default and leaving it opt-in allows us to say precisely it is the user responsibility once she needs to work with such data. There are other places in Ecto which you can be specific to a given database by opting in, be it via query fragments, setting default values in the database, and so on.

lau:
@josevalim Good point. How about adding an option to the "timestamps" function? For instance you could write `timestamps usec: true` to have timestamps include microseconds.

The default would be false and have the same behavior as today. But if you want usec timestamps, you can set that option. We could then do an explicit test case with the usec option set to true. And tag the test in the same way as for instance the array_type. Then MySQL can skip tests including microseconds in the same elegant way as it array tests are skipped.

How does that sound?

josevalim:
Yes, that sounds perfect to me because:

* the conflicting behaviour is opt-in
* we can write tests for it and cleanly skip them for MySQL

:+1: and thanks for not giving up on this!

josevalim:
FWIW, I have just documented all MySQL limitations in Ecto: https://github.com/elixir-lang/ecto/commit/9c3bb502aeb9be65ac5dc5a5b987f82a89f76673

lau:
Great :+1:

lau:
PR #522 added.

PR #518, opened 2015-03-14
Support custom source in build assoc

### LINKED ISSUE #478: Allow {source, model} to be given as queryable

tony612:
resolve #478 for building assoc

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -48,17 +48,19 @@ defmodule Ecto.Integration.RepoTest do
   test "insert, update and delete" do
     post = %Post{title: "create and delete single", text: "fetch empty"}
 
-    assert %Post{__meta__: %Metadata{state: :loaded}} = TestRepo.insert(post)
+    loaded_meta = %Metadata{state: :loaded, source: "posts"}
+    deleted_meta = %Metadata{state: :deleted, source: "posts"}
+    assert %Post{__meta__: ^loaded_meta} = TestRepo.insert(post)
     assert %Post{} = to_be_deleted = TestRepo.insert(post)
-    assert %Post{__meta__: %Metadata{state: :deleted}} = TestRepo.delete(to_be_deleted)
+    assert %Post{__meta__: ^deleted_meta} = TestRepo.delete(to_be_deleted)
 
     post = TestRepo.one(Post)
     assert post.__meta__.state == :loaded
     assert post.inserted_at
     assert post.updated_at
 
-    post = %{post | text: "coming very soon...", __meta__: %Metadata{state: :built}}
-    assert %Post{__meta__: %Metadata{state: :loaded}} = TestRepo.update(post)
+    post = Ecto.Schema.put_meta(%{post | text: "coming very soon..."}, :state, :built)
+    assert %Post{__meta__: ^loaded_meta} = TestRepo.update(post)
   end
 
   test "insert and update with changeset" do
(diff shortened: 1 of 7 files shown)


tony612:
It seems [a ci job](https://travis-ci.org/elixir-lang/ecto/jobs/54347679) failed for elixir 1.0.2 and OTP 17.4. I'll look into this later

integration_test/cases/repo.exs
```diff
@@ -44,20 +44,21 @@ defmodule Ecto.Integration.RepoTest do
       TestRepo.all(from(p in "posts", where: p.title == "title1", select: p.id))
   end
 
+  alias Ecto.Schema.Metadata
```
josevalim: Let's move the alias to the top of the file pls!

lib/ecto/association.ex
```diff
@@ -205,6 +205,9 @@ defmodule Ecto.Association do
       "or {source, model}, got: #{inspect queryable}"
   end
 
+  def source_from_query({source, model}) when is_binary(source) and is_atom(model), do: source
```
josevalim: We need docs. :)

lib/ecto/association.ex
```diff
@@ -264,8 +267,12 @@ defmodule Ecto.Association.Has do
...
 
   @doc false
-  def build(%{assoc: assoc, owner_key: owner_key, assoc_key: assoc_key}, struct) do
-    Map.put apply(assoc, :__struct__, []), assoc_key, Map.get(struct, owner_key)
+  def build(%{assoc: assoc, owner_key: owner_key, assoc_key: assoc_key,
+              queryable: queryable}, struct) do
+    assoc
+    |> apply(:__struct__, [])
+    |> Map.put(assoc_key, Map.get(struct, owner_key))
+    |> Ecto.Schema.put_meta(:source, Ecto.Association.source_from_query(queryable))
```
josevalim: We should only put the source if there is one from the association. Otherwise it will be set to nil and that is not good. We should write a test for that. :)

lib/ecto/association.ex
```diff
@@ -465,8 +472,9 @@ defmodule Ecto.Association.BelongsTo do
   end
 
   @doc false
-  def build(%{assoc: assoc}, _struct) do
+  def build(%{assoc: assoc, queryable: queryable}, _struct) do
     apply(assoc, :__struct__, [])
+    |> Ecto.Schema.put_meta(:source, Ecto.Association.source_from_query(queryable))
```
josevalim: Same here.

lib/ecto/repo/model.ex
```diff
@@ -14,7 +14,7 @@ defmodule Ecto.Repo.Model do
     struct = struct_from_changeset!(changeset)
     model  = struct.__struct__
     fields = model.__schema__(:fields)
-    source = model.__schema__(:source)
+    source = struct.__meta__.source || model.__schema__(:source)
```
josevalim: Let's *always* read the source from `__meta__.source`. If it is nil, it should be a bug.

lib/ecto/schema.ex
```diff
@@ -200,7 +214,7 @@ defmodule Ecto.Schema do
       Module.register_attribute(__MODULE__, :ecto_assocs, accumulate: true)
       Module.register_attribute(__MODULE__, :ecto_raw, accumulate: true)
 
-      Module.put_attribute(__MODULE__, :struct_fields, {:__state__, :built})
+      Module.put_attribute(__MODULE__, :struct_fields, {:__meta__, %Metadata{state: :built}})
```
josevalim: Let's store the source in the metadata too so it is always available (instead of nil).

josevalim:
The build was a race condition somewhere in our suite (or a database failure). I have restarted it and now all is fine. I have also added some comments to the source.

tony612:
@josevalim Done. Please review

lib/ecto/association.ex
```diff
@@ -198,6 +198,21 @@ defmodule Ecto.Association do
   defp to_lower_char(char) when char in ?A..?Z, do: char + 32
   defp to_lower_char(char), do: char
 
+  @doc """
+  Retrieves assoc from queryable.
```
josevalim: Oh, good catch. :heart:

josevalim:
Looks beautiful. I will merge this when I put some time into Ecto this week. :)

tony612:
@josevalim Awesome!! :clap:

josevalim:
:heart_eyes:

PR #519, opened 2015-03-14
fix the "fail invalid default" test

tony612:
This test is meant to test invalid default, but it just tests same thing with "fail invalid schema" test.


josevalim:

PR #522, opened 2015-03-16
Add usec as option to timestamps

lau:
Adds an option to timestamps for using microseconds.

This defaults to true, but we could change it to false if necessary.

I think true is good because this is the best option for users of Postgres and MSSQL. This is why it was decided to merge the timestamps-with-microseconds as default originally.

With MySQL things will still work with the default being true, although the results can be slightly different depending on the setup. MySQL 5.6 users can set their migrations to use datetime(6) and enjoy the microseconds or set usec to false. Or do nothing and rounding will occur. MySQL 5.5 users can set it to false or do nothing and microseconds will just be saved as 0.

The tests are still clean regardless of the default being true or false.

### DIFF
--- a/lib/ecto/model/timestamps.ex
+++ b/lib/ecto/model/timestamps.ex
@@ -19,13 +19,23 @@ defmodule Ecto.Model.Timestamps do
   @doc """
   Puts a timestamp in the changeset with the given field and type.
   """
-  def put_timestamp(changeset, field, type) do
+  def put_timestamp(changeset, field, type, use_usec) do
     if get_change changeset, field do
       changeset
     else
-      {date, {h, m, s}} = :erlang.universaltime
-      put_change changeset, field, Ecto.Type.load!(type, {date, {h, m, s, 0}})
+      put_change changeset, field, Ecto.Type.load!(type, timestamp_tuple(use_usec))
     end
+  end
+
+  defp timestamp_tuple(_use_usec = true) do
+    erl_timestamp = :os.timestamp
+    {_, _, usec} = erl_timestamp
+    {date, {h, m, s}} =:calendar.now_to_datetime(erl_timestamp)
+    {date, {h, m, s, usec}}
+  end
+  defp timestamp_tuple(_use_usec = false) do
+    {date, {h, m, s}} = :erlang.universaltime
+    {date, {h, m, s, 0}}
   end
 
   defmacro __before_compile__(env) do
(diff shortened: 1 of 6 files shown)


lib/ecto/model/timestamps.ex
```diff
@@ -19,31 +19,42 @@ defmodule Ecto.Model.Timestamps do
...
     if get_change changeset, field do
       changeset
     else
-      {date, {h, m, s}} = :erlang.universaltime
-      put_change changeset, field, Ecto.Type.load!(type, {date, {h, m, s, 0}})
+      put_change changeset, field, Ecto.Type.load!(type, timestamp_tuple(use_usec))
     end
   end
 
+  defp timestamp_tuple(use_usec) when use_usec == true do
```
josevalim: You can match directly: `timestamp_tuple(true)` or `timestamp_tuple(_use_usec = true)`.

lib/ecto/schema.ex
```diff
@@ -195,6 +198,7 @@ defmodule Ecto.Schema do
       import Ecto.Schema, only: [schema: 2]
       @primary_key {:id, :integer, read_after_writes: true}
       @timestamps_type Ecto.DateTime
+      @timestamps_use_usec true
```
josevalim: The default should be false (i.e. we should default to the behaviour that works out of the box on all adapters).

josevalim:
Thank you, I have added some comments!

lib/ecto/schema.ex
josevalim: Sorry, I see your comments now that you decided to go with true. :) Don't worry, let me think about this more and I will merge it later.

josevalim:
This looks great, thank you! Give me a couple hours and I will merge this later.

lib/ecto/model/timestamps.ex
lau: I really like the "_foo = true" form, but didn't have the syntax on the top of my head and didn't get around to look it up before the PR. It's a nice way to label arguments in cases like I think. Fixing this.

lib/ecto/schema.ex
lau: Practically with MySQL 5.5 it doesn't even round up, just discards the microseconds, so if 5.5 users just go by this default, it will work just fine.

With 5.6 going with this default it will also just work and the only practical difference is that timestamps can be up to half a second "too quick".

So these worst case scenarios for defaults are not bad I think.

And you even documented the limitations in MySQL in the adapter. I could add a paragraph there saying something along the lines of: If you use MySQL 5.6 set usec: false on timestamps or make the types datetime(6) in your migrations.

josevalim: > And you even documented the limitations in MySQL in the adapter. I could add a paragraph there saying something along the lines of: If you use MySQL 5.6 set usec: false on timestamps or make the types datetime(6) in your migrations.

That's the issue. Imagine I am writing some code that is meant to be used with any database because it is a library. I am not going to remember adding the usec: false and the code will be broken under certain databases (by broken I mean: if we care so much about this precision, having it half a second "too quick" sounds horrifying). This is also a huge incognito when talking about future adapters. One of the goals of adding MySQL was exactly so we get some diversity in and test our APIs with something besides Postgres. :)

lau: OK, default changed to false :)

josevalim:

PR #525, opened 2015-03-17
Create database in migrate task

### LINKED ISSUE #509: Consider creating the database on mix ecto.migrate
We should support --no-create option to skip it in case we implement this.

tony612:
resolve #509


josevalim:
:heart:

PR #533, opened 2015-03-19
enable setting attributes when building an assoc

### LINKED ISSUE #532: How about adding attributes to `build` function?
So that we can use the built struct more easily.

tony612:
resolves #532

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -72,9 +72,9 @@ defmodule Ecto.Association do
   The struct to build from is given as argument in case default values
   should be set in the struct.
 
-  Invoked by `Ecto.Model.build/2`.
+  Invoked by `Ecto.Model.build/3`.
   """
-  defcallback build(t, Ecto.Model.t) :: Ecto.Model.t
+  defcallback build(t, Ecto.Model.t, %{atom => term} | [Keyword.t]) :: Ecto.Model.t
 
   @doc """
   Returns an association join query.
@@ -295,9 +295,10 @@ defmodule Ecto.Association.Has do
 
   @doc false
   def build(%{assoc: assoc, owner_key: owner_key, assoc_key: assoc_key,
-              queryable: queryable}, struct) do
+              queryable: queryable}, struct, attributes \\ %{}) do
     assoc
     |> apply(:__struct__, [])
+    |> Map.merge(Enum.into(attributes, %{}))
     |> Map.put(assoc_key, Map.get(struct, owner_key))
     |> Ecto.Association.merge_source(queryable)
   end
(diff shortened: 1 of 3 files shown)


paulcsmith:
This will be really helpful! Thanks @tony612

lib/ecto/association.ex
```diff
@@ -295,9 +295,10 @@ defmodule Ecto.Association.Has do
 
   @doc false
   def build(%{assoc: assoc, owner_key: owner_key, assoc_key: assoc_key,
-              queryable: queryable}, struct) do
+              queryable: queryable}, struct, attributes \\ %{}) do
```
josevalim: We don't need to have the default argument in here because this is always invoked as a callback.

josevalim: Same for the builds below. :)

lib/ecto/association.ex
```diff
@@ -295,9 +295,10 @@ defmodule Ecto.Association.Has do
...
     assoc
     |> apply(:__struct__, [])
+    |> Map.merge(Enum.into(attributes, %{}))
```
josevalim: `struct/2` does exactly this! This:

     assoc
     |> apply(:__struct__, [])
     |> Map.merge(Enum.into(attributes, %{}))

Is equivalent to:

     assoc
     |> struct(attributes)

:)

lib/ecto/model.ex
```diff
@@ -96,10 +96,22 @@ defmodule Ecto.Model do
...
+      %Comment{id: nil, post_id: 13, text: "cool"}
+
+      iex> build(post, :comments, %{text: "cool"})
+      %Comment{id: nil, post_id: 13, text: "cool"}
+
+      iex> build(post, :comments, post_id: 1)
+      %Comment{id: nil, post_id: 13}
   """
-  def build(%{__struct__: model} = struct, assoc) do
+  def build(%{__struct__: model} = struct, assoc, attributes \\ %{}) do
```
josevalim: Default arguments here are great. :+1:

lib/ecto/model.ex
```diff
@@ -96,10 +96,22 @@ defmodule Ecto.Model do
       %Comment{id: 13, post_id: 25}
       iex> build(comment, :post)
       %Post{id: nil}
+
+  You can also pass the attributes, which can be a map or
+  a keyword list, to set the struct's fields except the key.
```
josevalim: "except the association key" would be clearer. :)

josevalim:
I have added some minor comments, thank you! <3

tony612:
@josevalim Done. Please review. And `struct` function is very neat!
@paulcsmith Thanks! :smile:

josevalim:

Issue #535, opened 2015-03-22
Timestamps with time zone broken


MSch:
Seems like Ecto 0.9 has broken support for timestamps with time zone. Adding the following after `create table(:posts)` in `integration_test/support/migration.exs` breaks the tests:

```elixir
execute "ALTER TABLE posts ALTER COLUMN inserted_at TYPE timestamp with time zone;"
```

I tried to do some investigating but didn't get far.

----

Related: Specifying `:datetime` in migrations should create a column of type `timestamp with time zone`, not `timestamp without time zone`:

> Long story short – I believe that using timestamps (without time zone) is, in almost every case, a bug. And should be fixed.

http://www.depesz.com/2014/04/04/how-to-deal-with-timestamps/

> Use timestamp with time zone (aka timestamptz) and time with time zone (aka timetz). They store values as UTC, but convert them on selection to whatever your time zone setting is.
> Avoid timestamp without time zone (aka timestamp)

http://justatheory.com/computers/databases/postgresql/use-timestamptz.html

josevalim:
Ecto does not support timestamps with timezones. :) That's why it is "broken" for you. I actually advocate that dates should be saved without timezones in the database and @lau could even give good reasons for such.

Also, because not all databases support timezones, having no timezones by default is the best way to ensure they work with any database (which is one of Ecto goals).

It is likely that the best solution here will be an ecto_pg_timezone package. :)

MSch:
Just to make sure we are talking about the same thing: `timestamp with time zone ` (= `timestamptz`) does not save the time zone in the database, like `timestamp without time zone` (= `timestamp`) it just stores a timestamp without storing any timezone/offset information.

The difference is `timestamptz` is defined to always be in UTC and is transparently converted from/to the time zone that is set for the current connection. (Most applications will set the connection's time zone to UTC, so that no conversion is necessary)

E.g.

```sql
select timestamp '1999-01-08 04:05:06 -8:00'; -- 1999-01-08 04:05:06 TZ information is just discarded
select timestamptz '1999-01-08 04:05:06 -8:00'; -- 1999-01-08 12:05:06+00 The TZ set for the connection is used to convert the timestamp to UTC
select timestamptz '1999-01-08 04:05:06'; -- 1999-01-08 04:05:06+00 if the timestamp is already in UTC nothing happens
```

[...]

PR #537, opened 2015-03-22
Support timestamptz

MSch:
Not ready to be merged, lacks tests. Depends on #535.

I don't know enough about how Postgrex Extensions and Ecto Types play together yet, is it possible to refactor this to use Postgrex de-/encoding of the types instead of duplicating the code?


josevalim:
Damn it, I completely missed this and basically pushed the same thing. Sorry.

PR #540, opened 2015-03-23
Make it possible to set the storage engine for MySQL databases

laurocaetano:
In MySQL database is possible to set the storage engine. This PR adds an option to set the storage engine when creating the database and when creating a table.

Closes https://github.com/elixir-lang/ecto/issues/538

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -116,8 +116,8 @@ defmodule Ecto.Migration do
     @moduledoc """
     Defines a table struct used in migrations.
     """
-    defstruct name: nil, primary_key: true
-    @type t :: %__MODULE__{name: atom, primary_key: boolean}
+    defstruct name: nil, primary_key: true, engine: nil
+    @type t :: %__MODULE__{name: atom, primary_key: boolean, engine: atom}
   end
 
   defmodule Reference do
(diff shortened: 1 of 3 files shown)


lib/ecto/migration.ex
```diff
@@ -246,6 +246,8 @@ defmodule Ecto.Migration do
   ## Options
 
     * `:primary_key` - when false, does not generate primary key on table creation
+    * `:storage_engine` - set the storage engine for MySQL adapters. When not
```
josevalim: We can just call it engine. In the documentation, we should just say: "It customizes the table storage for supported daabases. For MySQL, the default is InnoDB". The idea is that the general documentation comes first, notes for specific databases is added later.

lib/ecto/adapters/mysql.ex
```diff
@@ -29,6 +29,7 @@ defmodule Ecto.Adapters.MySQL do
     * `:ssl` - Set to true if ssl should be used (default: false);
     * `:ssl_opts` - A list of ssl options, see ssl docs;
     * `:connect_timeout` - The timeout for establishing new connections, defaults to `5000`
+    * `:storage_engine` - The storage engine to be used in the current session, i.e: "MYISAM". The default is "INNODB"
```
josevalim: Does it makes sense to have this option for storage_up? Even if the user sets it to MyISAM, the migrations will default to InnoDB, having no effect. I think we can :scissors: this. What do you think?

laurocaetano: Yep, you are right. I'll :scissors:

lib/ecto/migration.ex
laurocaetano: Nice catch.
Thanks :)

laurocaetano:
@josevalim updated

josevalim:

PR #542, opened 2015-03-24
Fix handling of dates before 2000-01-01; epoch offsets need to be signed...

john-vinters:
Modifying the binary matching to return signed integers makes dates/timestamps before 2000-01-01 work correctly, and all tests still pass OK.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -309,6 +309,10 @@ defmodule Ecto.Integration.RepoTest do
     assert %Comment{id: id2} = TestRepo.insert(%Comment{})
     assert 1 = TestRepo.update_all(p in Comment, text: ^text, posted: ^date)
     assert %Comment{text: "hai", posted: ^date} = TestRepo.get(Comment, id2)
+
+    date = {{1955, 11, 12}, {6, 38, 01, 0}}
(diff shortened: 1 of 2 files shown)


laurocaetano:
Could you add some tests for dates before 2000 in `test/ecto/datetime_test.exs`?

laurocaetano:
Btw, I think the fix for `Postgrex` will be the same: https://github.com/ericmj/postgrex/blob/master/lib/postgrex/extensions/binary.ex#L207 

Also, this PR closes https://github.com/elixir-lang/ecto/issues/539

john-vinters:
The problem appears Postgres specific -- using pre 2000 dates in cast etc works correctly, the issue only shows when a real database table is read and the date/timestamp is decoded.

Would the tests be better placed elsewhere as all the tests in `test/ecto/datetime_test.exs` don't touch the database?

laurocaetano:
You are right, thanks! I checked the Postgrex code and there is no bug there. It correctly uses signed integers. 

About the tests, what to you think to place it in `integration_test/cases/repo.ex`? You could just change the test that you mentioned in the issue. (https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs#L308)

josevalim:

Issue #552, opened 2015-04-09
Add function to convert Ecto.Date to Ecto.DateTime (with zeroed-out time)


tobz:
I think there's a very common use case of wanting to search for things based on the date alone, where the underlying column is a datetime i.e. my_little_dt_column > '2015-04-09 00:00:00'

Trying to pass in `Ecto.Date` values on the RHS of a where clause won't compile, so you need to go and build your `Ecto.DateTime` value.  The simplest way I found to do this, based on the available functions, was:

```elixir
{:ok, zero_time} = Ecto.Time.cast("00:00:00")
date_today = Ecto.DateTime.from_date_and_time(Ecto.Date.local, zero_time)
```

Maybe I'm missing something, but it would be much simpler / pretty intuitive (IMO) if Ecto.Date had a `to_datetime` function that handled this for you.

josevalim:
We can either do `Ecto.DateTime.from_date` or `Ecto.Date.to_datetime`. @ericmj any preference?

josevalim:
Actually, the first one is better given we already have to_date in there and from_date_and_time. @tobz can you please send a pull request? Thank you!

tobz:
Word.  I'll get something whipped up.

PR #553, opened 2015-04-09
Add from_date for datetimes.

tobz:
This lets you create a datetime from just a date, which is useful if you're doing comparisons where you want to find all rows for a specific date.


tobz:
This fulfills #552.

josevalim:

PR #555, opened 2015-04-10
allow distinct(true) query expresion

jxs:


lib/ecto/query.ex
```diff
@@ -436,6 +436,10 @@ defmodule Ecto.Query do
     Distinct.build(query, binding, expr, __CALLER__)
   end
 
+  defmacro distinct(query, expr) do
```
josevalim: We need to add docs. Even if it will be just "Adds a distinct clause. See `distinct/3`."

jxs: yeah of course, going to update :)

josevalim:

josevalim:
Actually, I just realized we could do `binding \\ []` in the first definition. I will fix it locally :D

PR #561, opened 2015-04-23
Fixing reference in docs to Ecto.Associations

amutz:
Hello.  I know this is a minor issue, but I tripped up on it just now.  This fixes one spot where the documenting comments refer to the old module name Ecto.Associations, instead of the new one Ecto.Association.


laurocaetano:
:shipit:

PR #567, opened 2015-04-27
Update on readme

janjiss:
In my opinion, Simple does not describe the intent of the module. I am suggesting that it is renamed to SimpleQuery. Close this if you do not agree :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ defmodule Weather do
   end
 end
 
-defmodule Simple do
+defmodule SimpleQuery do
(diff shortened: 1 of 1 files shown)


josevalim:
I think it was supposed to be `Sample`. We can call it `Sample` or `Sample.App`. Could you also please namespace Repo and the Weather under the `Sample` alias?

janjiss:
Does this look better?

josevalim:
Thank you! :heart:

PR #570, opened 2015-04-28
Use correct variable name in documentation

scrogson:
Also, can we get an example of using multiple `from` expressions?

https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query.ex#L244-L246


josevalim:
Good catch!

Queries can't have more than one from, they can have more than one source (one from and another is a join). Can you please update the docs?

josevalim:
Ping?

PR #572, opened 2015-04-29
Initial migration config

gjaldon:
refs: https://github.com/elixir-lang/ecto/issues/566

@josevalim will this work for customizing primary key in `Ecto.Migration.SchemaMigration`? I initially tried testing by checking if `SchemaMigration` has a primary key set but realized later on it gets compiled before I do `Application.put_env` in the tests. 

Should I write a test that checks `SchemaMigration` has the primary key set as expected? If so, how to provide the expected application env variables in compile-time in tests?

### DIFF
--- a/lib/ecto/migration/config.ex
+++ b/lib/ecto/migration/config.ex
@@ -1,0 +1,33 @@
+defmodule Ecto.Migration.Config do
+  @moduledoc false
+
+  @doc """
+  Retrieves the `:ecto_migration` configuration for `:primary_key`
+  """
+  def primary_key do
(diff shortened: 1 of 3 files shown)


josevalim:
Hrm, I don't think we should provide piece-meal configuration like that. I think we should have a single configuration called "schema_migrations_model" that expects a module name. If one is set for the repository, we will use it for all migration operations instead of the default one.

Does it make sense?

gjaldon:
Yep, it does. That way it's more flexible. I'll work on this again later and use the approach you suggested. Thanks for the feedback!

PR #578, opened 2015-05-01
Allow uniqueness constraint on multiple columns

voltone:


josevalim:
Thank you! This is being discussed on #534 and @hahuang65 already said he is sending a pull request for it. Since he has been in the discussion since the beginning, I believe it makes more sense to merge his work (which is truly almost there). :)

voltone:
I saw that discussion, but got the feeling that the proposal was getting a bit unwieldy. It would be more powerful, but the extra complexity might get in the way for a simple use-case as this one.

voltone:
Hmm, I think I may have missed some of the latest comments. I thought the 'scope' was supposed to be a query. But if it can be just a list of field, then that does make my PR redundant. Never mind, I'll wait for that one...

PR #580, opened 2015-05-01
Add :scope option to validate_unique

hahuang65:
This will fulfill https://github.com/elixir-lang/ecto/issues/534

Comments and criticisms are appreciated.

Are there any tests that I'm missing @josevalim @ericmj ?

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -701,6 +701,7 @@ defmodule Ecto.Changeset do
     * `:message` - the message on failure, defaults to "has already been taken"
     * `:on` - the repository to perform the query on
     * `:downcase` - when `true`, downcase values when performing the uniqueness query
+    * `:scope` - a list of other fields to use for the uniqueness query
 
   ## Case sensitivity
 
@@ -736,6 +737,15 @@ defmodule Ecto.Changeset do
       |> update_change(:email, &String.downcase/1)
       |> validate_unique(:email, on: Repo)
 
+  ## Scope
+
+  The `:scope` option allows specifying of other fields that are used to limit
+  the uniqueness check. For exmaple, if our use case limits a user to a single
+  comment per blog post, it would look something like:
+
+      cast(params, model, ~w(comment), ~w())
+      |> validate_unique(:user_id, scope: :post_id, on: Repo)
+
   """
   @spec validate_unique(t, atom, [Keyword.t]) :: t
   def validate_unique(%{model: model} = changeset, field, opts) when is_list(opts) do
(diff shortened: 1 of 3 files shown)


hahuang65:
Actually let's hold off on this for a day, I want to add some integration tests, and maybe look at what to do when they pass in fields that don't exist on the model.

test/ecto/changeset_test.exs
```diff
@@ -564,6 +565,40 @@ defmodule Ecto.ChangesetTest do
     assert changeset.validations == [title: :unique]
   end
 
+  test "validate_unique/3 with scope" do
+    defmodule ScopeRepo do
+      def all(query) do
+        assert query.wheres |> Enum.count == 2
+        query_strings =  query.wheres |> Enum.map(&Macro.to_string(&1.expr))
+        assert query_strings |> Enum.member?("&0.title() == ^0")
```
josevalim: `assert "&0.title() == ^0" in query_strings`

josevalim:
This looks great. Let's just add the tests indeed. Integration tests can be added here: https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs#L140 :)

hahuang65:
@josevalim @ericmj what do you guys think about ignoring fields passed into the scope option if they don't exist?

The current check-in does this, and is tested, but perhaps we should consider throwing an exception?

lib/ecto/changeset.ex
```diff
@@ -733,6 +743,17 @@ defmodule Ecto.Changeset do
       struct = model.__struct__
       query  = from m in struct, select: field(m, ^field), limit: 1
 
+      if opts[:scope] do
+        query = Enum.reduce(opts[:scope], query, fn(field, acc) ->
+          value = get_field(changeset, field)
+          if value do
```
josevalim: I don't think we should ignore the value. If it is nil, it also need to be used in our comparison, no?

josevalim:
I don't think we should ignore nil values. We should just pass everything in, no?

josevalim:
I also checked Rails and they don't discard nils too.

lib/ecto/changeset.ex
hahuang65: Oh, there are two cases here

`validate_unique(changeset, :title, scope: [:field_that_doesnt_exist], on: Repo)`

and

`validate_unique(changeset, :title, scope: [:field_that_exists_but_is_nil], on: Repo)`

hahuang65:
@josevalim Just commented as a line comment.

lib/ecto/changeset.ex
josevalim: No, we don't ignore any. It is going to be much clearer like that, otherwise there will be bugs hiding. If you pass a field that does not exist, it should raise, and `field(m, ^field)` will raise automatically.

hahuang65:
Fixed such that fields that are nil are respected, and made sure that if a field doesn't exist, the validation will raise Ecto.QueryError.

PR #583, opened 2015-05-05
Fix put_change/3, put_new_change/3 and change/2 and add force_change/3

### LINKED ISSUE #582: put_change/3, put_new_change/3 and change/2 should check if the value is different from the one in the model
If the value is the same as in the model, we won't add it to the list of changes. If it is different, we add it.

We will also introduce `force_change/3` which works as `put_change/3` today.

Is someone interested in picking this up? We should release v0.11.1 afterwards with those fixes.

bernardoamc:
The first three methods should not add a change to the list of changes if the changed value is the same as in the model. The latter add the change even if the model has the same value.

Closes #582


josevalim:

PR #591, opened 2015-05-09
Support database defaults during insert.

joshwlewis:
For a bit of background, consider this model:

```elixir
create table(:clients, primary_key: false) do
  add :id, :uuid, primary_key: true, default: fragment("uuid_generate_v4()")
  add :name, :string, null: false
  add :url, :string, null: false
  add :secret, :uuid, null: false, default: fragment("uuid_generate_v4()")

  timestamps
end
```
```elixir
schema "clients"  do
  field :name, :string
  field :url, :string
  field :secret, Ecto.UUID, read_after_writes: true
  timestamps
end
```

My assumption was that creating a new record without specifying a 'secret' would let the database default kick in. Instead, here is what happens:

```elixir
changeset = Client.changeset(%Client{}, %{name: "local", url: "http://localhost:4200"})
Repo.insert(changeset)
# => ** (Postgrex.Error) ERROR (not_null_violation): null value in column "secret" violates not-null constraint
```

I suppose I would have expected Ecto to only insert values that it had data for. However, Ecto is explicitly inserting `NULL`, so database defaults aren't kicking in. I'm thinking that changing this so that only non-null values are inserted is the appropriate fix for this.

What do you guys think here? I'm sort of a newbie to Elixir, so feedback is appreciated.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -117,15 +117,13 @@ defmodule Ecto.Integration.RepoTest do
   test "insert and update with changeset dirty tracking" do
     changeset = Ecto.Changeset.cast(%Post{}, %{}, ~w(), ~w())
 
-    # There is no dirty tracking on insert, even with changesets,
-    # so database defaults never actually kick in.
-    assert %Post{id: pid, counter: nil} = post = TestRepo.insert(changeset)
+    # Make sure the database default gets loaded with read_after_writes.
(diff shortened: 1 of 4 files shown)


joshwlewis:
Oh, and these changes weren't working against mariaex 0.1.0 when inserting primitive types. I'm not sure which commit fixed it, but tests passed after upgrading to 0.1.4.

josevalim:
Yes, this should be fixed and work in master. :)

PR #592, opened 2015-05-11
Support null: false columns with a default.

joshwlewis:
I don't mean to be a pest, but #591 is still an issue in master for me. I'm opening a new PR with some failing tests this time, and hopefully a better error report.

Consider this model:
```elixir
create table(:clients, primary_key: false) do
  add :id, :uuid, primary_key: true, default: fragment("uuid_generate_v4()")
  add :name, :string, null: false
  add :url, :string, null: false
  add :secret, :uuid, null: false, default: fragment("uuid_generate_v4()")
  timestamps
end
```
```elixir
schema "clients"  do
  field :name, :string
  field :url, :string
  field :secret, Ecto.UUID, read_after_writes: true
  timestamps
end
```

When specifying both null: false and a default value, you are effectively prevented from using the database default at all:

```elixir
changeset = Client.changeset(%Client{}, %{name: "local", url: "http://localhost:4200"})
Repo.insert(changeset)
# => ** (Postgrex.Error) ERROR (not_null_violation): null value in column "secret" violates not-null constraint
```

I had the impression that #591 was about defaults, but my context was wrong. This appears to only be an issue when columns have both a default and a null: false.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -117,15 +117,13 @@ defmodule Ecto.Integration.RepoTest do
   test "insert and update with changeset dirty tracking" do
     changeset = Ecto.Changeset.cast(%Post{}, %{}, ~w(), ~w())
 
-    # There is no dirty tracking on insert, even with changesets,
-    # so database defaults never actually kick in.
-    assert %Post{id: pid, counter: nil} = post = TestRepo.insert(changeset)
+    # Make sure the database default gets loaded with read_after_writes.
+    assert %Post{id: pid, counter: 10} = post = TestRepo.insert(changeset)
 
-    # Set the counter to 11, so we can read it soon
+    # Set the counter to 11
     TestRepo.update(%{post | counter: 11})
 
-    # Now, a combination of dirty tracking with read_after_writes,
-    # allow us to see the actual counter value.
+    # Update and validate that counter has not reverted to default
     changeset = Ecto.Changeset.cast(post, %{"title" => "hello"}, ~w(title), ~w())
     assert %Post{id: ^pid, counter: 11, title: "hello"} = post = TestRepo.update(changeset)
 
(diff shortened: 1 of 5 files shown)


josevalim:
Thanks for the report. I don't like discarding nil values because you may have a default in the database and you may want to override it with nil. The only option I have in mind right now is to allow setting a field as read_only, which means it is never written to, so we never send anything to the database. Or maybe we can incorporate this behaviour on read after writes.

This is a tricky issue, thanks for persisting here. :)

joshwlewis:
What if we treated explicitly provided nil attributes differently than those that weren't provided at all?

```elixir
Client.changeset(%Client{}, %{name: "local", secret: nil, url: "http://localhost:4200"})
|> Repo.insert
```
In something like this, we are explicitly providing `secret: nil`, so it would make sense to pass this on to the database and essentially disregard the database default.

```elixir
Client.changeset(%Client{}, %{name: "local", url: "http://localhost:4200"})
|> Repo.insert
```
But in this case, `secret: nil` is implied, not given. I think it would make sense to strip out `secret: nil` here, and let the database decide what to put in that column.

josevalim:
@joshwlewis this is a nice direction. The only complication is that we support default values in the schema too. In your example, if someone does:

    field :secret, :string, default: "foo"

Should we send it to the database because it was explicitly set? Even though nothing came in the changeset? And what if someone does this:

    field :secret, :string, default: "foo"

We have some different features and they are at odds with each other here. Another concern about your solution is how we would make it work for document stores. In them, we don't have default values and, if we send just what was given, we would have an inaccurate description of the document.

In any case, your problem has a solution today, albeit verbose. You can define a before_insert callback that will receive the changeset and simply remove the change from there:

    before_insert :remove_secret
    def remove_secret(changeset) do
      Ecto.Changeset.delete_change(changeset, :secret)
    end

I think even this may work:

    before_insert Ecto.Changeset, :delete_change, [:secret]

The idea is that a read_only option would just simplify this.

josevalim:
Ok, we have decided to fix by supporting something similar to your patch, except we only discard nil read_after_writes values (instead of any nil value). This gives the opposite behaviour of what we had before. Before read_after_writes were always included but most certainly had to be deleted with `delete_change`. Now they are always excluded but you can force them with `force_change`.

Thanks for persisting on this and the other reports, we really appreciate it. :heart:

joshwlewis:
Awesome. That's more in line with what I would have expected from `read_after_writes`. Thanks for the quick turn-around. You guys are doing great work.

PR #597, opened 2015-05-12
adding validate_confirmation

### LINKED ISSUE #595: Add validate_confirmation
Just like validates :email, confirmation: true in Rails.

meatherly:
This should resolve #595

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -940,6 +940,38 @@ defmodule Ecto.Changeset do
     end
   end
 
+
+  @doc """
+  Validates that the given field matches the confirmation param of that field.
+
+  Note that this does not raise a validation error if the confirmation field is nil.
+
+  e.g. `validate_confirmation(changeset, :email)` would look for a param with the key of email_confirmation.
+
+  ## Options
+
+    * `:message` - the message on failure, defaults to "does not match"
+
+  ## Examples
+
+      validate_confirmation(changeset, :email)
+      validate_confirmation(changeset, :password, message: "passwords do not match")
+
+  """
+  @spec validate_confirmation(t, atom, Enum.t) :: t
+  def validate_confirmation(changeset, field, opts \\ []) do
+    validate_change changeset, field, {:confirmation, opts}, fn
+      field, value ->
+        confirmation_value = changeset.params["#{field}_confirmation"]
+        case is_nil(confirmation_value) do
+          true ->
+            []
+          false ->
+            if value != confirmation_value, do: [{:"#{field}_confirmation", message(opts, "does not match")}], else: []
+        end
+      end
+  end
+
   defp message(opts, default) do
     Keyword.get(opts, :message, default)
   end
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -724,4 +724,26 @@ defmodule Ecto.ChangesetTest do
                 |> validate_number(:upvotes, greater_than: 100, less_than: 0, message: "yada")
     assert changeset.errors == [upvotes: {"yada", 100}]
   end
+
+  test "validate_confirmation/3" do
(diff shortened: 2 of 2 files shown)


whatyouhide:
Maybe you could squash your commits into one, so that we keep a nice tidy history? :smiley:

ricn:
Cool! I was planning to solve the issue #595 that I created by myself this weekend but it seems like I don't have to!

meatherly:
@whatyouhide squashed 'em

whatyouhide:
@meatherly Great, thanks :)! :+1:

lib/ecto/changeset.ex
```diff
@@ -940,6 +940,38 @@ defmodule Ecto.Changeset do
...
+    * `:message` - the message on failure, defaults to "does not match"
+
+  ## Examples
+
+      validate_confirmation(changeset, :email)
+      validate_confirmation(changeset, :password, message: "passwords do not match")
+
+  """
+  @spec validate_confirmation(t, atom, Enum.t) :: t
+  def validate_confirmation(changeset, field, opts \\ []) do
```
josevalim: We should not require an `email_confirmation` field. Instead, we can simply look inside the params map.

josevalim: In fact, we need to compare both values before typecast, so I would check if both values in the params map are equal.

josevalim: Also, as in Rails, we should not raise an error if the confirmation field is `nil`, even if different!

meatherly: Even if different?
That sounds like the confirmation validation is not needed? Or are you saying only validate if the confirmation field isn't nil?

ricn: Only validate if confirmation field isn't nil: http://edgeguides.rubyonrails.org/active_record_validations.html#confirmation

lib/ecto/changeset.ex
```diff
@@ -940,6 +940,38 @@ defmodule Ecto.Changeset do
...
+    validate_change changeset, field, {:confimation, opts}, fn
```
crismali: `:confimation` should probably be `:confirmation` :)

test/ecto/changeset_test.exs
```diff
@@ -724,4 +725,21 @@ defmodule Ecto.ChangesetTest do
                 |> validate_number(:upvotes, greater_than: 100, less_than: 0, message: "yada")
     assert changeset.errors == [upvotes: {"yada", 100}]
   end
+
+  test "validate_confimation/3" do
```
crismali: `validate_confimation/3` should probably be `validate_confirmation/3` :)

meatherly: Alrighty then. I'll get working these requests today. @crismali I'll turn on spell check this time lol :p

meatherly:
@josevalim @ricn How does it look now?

lib/ecto/changeset.ex
```diff
@@ -940,6 +940,38 @@ defmodule Ecto.Changeset do
...
+  @spec validate_confirmation(t, atom, Enum.t) :: t
+  def validate_confirmation(changeset, field, opts \\ []) do
+    validate_change changeset, field, {:confirmation, opts}, fn
+      field, value ->
+        confirmation_value = changeset.params["#{field}_confirmation"]
+        case is_nil(confirmation_value) do
+          true ->
+            []
+          false ->
+            if value != confirmation_value, do: [{:"#{field}_confirmation", message(opts, "does not match")}], else: []
```
stevedomin: This feels like it could be cleaned up using guards. Maybe with something like this?

```elixir
case is_nil(confirmation_value) do
  true ->  []
  false when value != confirmation_value -> 
    [{:"#{field}_confirmation", message(opts, "does not match")}]
  false ->  []
end
```

josevalim:
:heart:

Issue #611, opened 2015-05-21
If begin raises in a nested transaction the outer transaction is rolledback to an invalid or previous savepoint at the same level


fishcakez:
Using the `Simple` example and a raw query that raises because `bad` is not a column:
```elixir
defmodule SimpleTransaction do
  def test do
    Simple.Repo.transaction(fn ->
      try do
        Ecto.Adapters.SQL.query(Simple.Repo, "UPDATE weather SET bad = bad + 1", [])
      catch
        _, _ ->
          :ok
      end
      Simple.Repo.transaction(fn -> :ok end)
      :ok
    end)
  end
end
```
Results in the following exception when a transaction is attempted:
```
** (Postgrex.Error) ERROR (invalid_savepoint_specification): no such savepoint
      (ecto) lib/ecto/adapters/sql/worker.ex:60: Ecto.Adapters.SQL.Worker.call!/3
    (simple) lib/simple.ex:13: Simple.Repo.log/2
      (ecto) lib/ecto/adapters/sql.ex:451: Ecto.Adapters.SQL.transaction/3
             lib/simple.ex:50: anonymous fn/0 in SimpleTransaction.test/0
      (ecto) lib/ecto/adapters/sql.ex:442: Ecto.Adapters.SQL.transaction/3
```
Or when a previous transaction has occurred at the same level:
```elixir
defmodule SimpleTransaction do
  def test do
    Simple.Repo.transaction(fn ->
      Simple.Repo.transaction(fn -> :ok end)
      try do
        Ecto.Adapters.SQL.query(Simple.Repo, "UPDATE weather SET bad = bad + 1", [])
      catch
        _, _ ->
          :ok
      end
      Simple.Repo.transaction(fn -> :ok end)
      :ok
    end)
  end
end
```
The begin raises but the rollback succeeds and rolls back to the beginning of the prior transaction at the same level:
```
** (Postgrex.Error) ERROR (in_failed_sql_transaction): current transaction is aborted, commands ignored until end of transaction block
      (ecto) lib/ecto/adapters/sql/worker.ex:60: Ecto.Adapters.SQL.Worker.call!/3
    (simple) lib/simple.ex:13: Simple.Repo.log/2
      (ecto) lib/ecto/adapters/sql.ex:441: Ecto.Adapters.SQL.transaction/3
             lib/simple.ex:51: anonymous fn/0 in SimpleTransaction.test/0
[...]

PR #612, opened 2015-05-22
[WIP] Control transactions from client process

### LINKED ISSUE #585: Do not make queries in Ecto worker

fishcakez:
Resolves #585, #611.

This branch takes a different approach to I suggested in #585 and keeps the transaction state in the client, instead of in the connection processes - on master the state is kept in the worker. Keeping the state in the connection process created significant complexity for the different drivers (postgrex, mariaex etc) and was starting to result in too many lines of code.

This implementation requires heavy use of the process dictionary, which makes logic hard to follow. I will need to add quite a few transaction tests and I would like to check the approach is ok before continuing.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -142,13 +142,71 @@ defmodule Ecto.Adapters.SQL do
   @spec query(Ecto.Repo.t, String.t, [term], Keyword.t) ::
              %{rows: nil | [tuple], num_rows: non_neg_integer} | no_return
   def query(repo, sql, params, opts \\ []) do
-    opts = Keyword.put_new(opts, :timeout, elem(repo.__pool__, 1))
+    {pool, timeout} = repo.__pool__
+    opts = Keyword.put_new(opts, :timeout, timeout)
 
     log(repo, {:query, sql, params}, opts, fn ->
-      use_worker(repo, opts[:timeout], fn worker ->
-        Worker.query!(worker, sql, params, opts)
-      end)
+      query!(repo, pool, sql, params, opts)
     end)
+  end
+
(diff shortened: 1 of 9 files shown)


josevalim:
:+1:

lib/ecto/adapters/sql.ex
```diff
@@ -434,72 +511,160 @@ defmodule Ecto.Adapters.SQL do
...
-      {worker, counter} ->
-        Process.put(key, {worker, counter - 1})
-    end
-
+  defp do_begin(repo, key, %{transactions: [_|_] = trans, depth: n, module: module} = info,
+  opts, timeout) do
+    savepoint = "ecto_#{n}"
+    n = n + 1
+    query = apply(module, :savepoint, [savepoint])
+    info = %{info | transactions: [next: {:savepoint, savepoint}] ++ trans, depth: n}
```
josevalim: Why do we need to keep the `trans` history? Wouldn't checking the depth be enough?

fishcakez: It's a stack, not the history, so transactions are removed on rollback/commit but you may have meant that. Checking the depth is not enough, some other state is required as the current depth could be either `before_begin`, `begin_ok` or `begin_error`. I passed the stack around as I found it easier to understand.

fishcakez:
Ouch those 2 failing tests :(. Fixing those is likely going to add more complexity.

josevalim:
I am closing this. I am almost there with #622, writing the docs now. I have commented the log tests for now, so we can merge this work and fix the log tests once we provide the new log entries.

Thanks @fishcakez, this was a great step for us to get things right. The test cases are also fantastic!

Issue #569, opened 2015-04-28
Support reserving a worker


MSch:
(as discussed at ElixirConf with @josevalim @ericmj)

We have a scenario like this: One `GenServer` `GlobalStuff` that needs to use an Ecto worker from time to time to do its work and a lot of concurrent requests handled by short-lived e.g. `plug` processes all of which need to call `GlobalStuff` while in a transaction before succeeding.

In our load-testing we always run into a deadlock, where we got all Ecto workers checked out by our short-lived processes and in a transaction, waiting for a call to `GlobalStuff` to return. `GlobalStuff` meanwhile is trying to check out an Ecto worker from poolboy, but none is available since all are checked out by the short-lived processes.

Our solution was to manually reserve an Ecto worker for `GlobalStuff`:

```elixir
defmodule PS.Repo do
  @doc ~S"""
  Manually reserve an Ecto worker permanently without creating a transaction (avoid long-running
  Postgres transactions) by pretending to create a transaction.
  """
  def reserve_ecto_worker! do
    worker = :poolboy.checkout(PS.Repo.pool, true, :infinity)
    Ecto.Adapters.SQL.Worker.link_me(worker, :infinity)
    Process.put({:ecto_transaction_pid, PS.Repo.pool}, {worker, 1})
    worker
  end


  @doc "Expose the internal `poolboy` pool"
  def pool do
    __pool__ |> elem(0) |> Process.whereis
  end
end
```

I'd like some public API that exposes this functionality (even better would be a solution that doesn't involve permanently assigning one Postgres connection to a single GenServer, but I don't think there's one)

josevalim:
If we expose an API for reserving, do we also need an API for checking it back in?

MSch:
In our use case the processes that reserve workers do so in `init` and keep the same worker around forever, so there's no need for checking it back in. One performance optimization we have planned would require us to hand the worker from one process to another and back again, but also not giving it back to the pool.

fishcakez:
> One performance optimization we have planned would require us to hand the worker from one process to another and back again, but also not giving it back to the pool.

This sounds like something poolboy would be good at. Have you considered making a repo of 1 connection just for this purpose?

MSch:
@fishcakez not really, I thought the need to either duplicate all involved models and functions using the repo or pass the repo around as argument to them, and then having to figure out how to DRY up the configuration for those two repos would make that a worse workaround.

[...]

PR #614, opened 2015-05-22
Use poolboy's name instead of pid in transaction key

fishcakez:
It is possible for poolboy's pid to change if it crashes or is restarted. Previously this could lead to the following race condition:

* client traps exits
* client looks up pid, checks out a worker and starts a transaction
* poolboy is restarted
* client does not handle exit signal as mid transaction
* client looks up new pid, key has changed and so checks out new worker

Doing `Process.whereis/1`, `Process.is_alive?/1` and then `GenServer.call/2,3` is an anti-pattern. The process could be alive at the check but dead at the call. It is also very defensive programming. It should be more natural to use `try do` when failure is an exceptional circumstance.

This patch is also more efficient as it does not lookup and check the poolboy process is alive everytime the worker is used. I am not sure how cheap or expensive `Process.is_alive?/1` is however avoiding the call is better either way.

### DIFF
--- a/integration_test/pg/deadlock_test.exs
+++ b/integration_test/pg/deadlock_test.exs
@@ -92,7 +92,7 @@ defmodule Ecto.Integration.DeadlockTest do
 
         # At this time the transaction count is actually 0 not 1 because Postgres
         # has killed the tx but Ecto doesn't know/care.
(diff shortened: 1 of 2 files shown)


fishcakez:
@msch this breaks the example code in #569 (you are a naught boy aren't you!)

MSch:
@fishcakez sure, but it just makes it easier for me, since I can use the process name directly instead of having to resolve it :)

PR #615, opened 2015-05-23
Ecto.UUID.load should accept a valid uuid string.

joshwlewis:
Hi! I just found what I think is a bug.

migration:
```elixir
create table(:clients) do
  add :secret,  :string,  null: false, default: fragment("uuid_generate_v4()")
end
```
model:
```elixir
schema "clients" do
  field :secret,  Ecto.UUID, read_after_writes: true
end
before_insert :delete_change, [:secret]
```
Failing script:
```elixir
Repo.insert(%Client{})
# => ** (ArgumentError) cannot load `"5c43eb56-35d6-4eae-adb8-db289befdfe8"` as type Ecto.UUID  
```

So, the database is generating a valid UUID and returning it, but Ecto.UUID doesn't know how to load it.

This seemed like a simple enough thing to fix, so the PR is attached.

Additional backstory on this here: https://github.com/elixir-lang/ecto/pull/592#issuecomment-104573693

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -21,7 +21,7 @@ defmodule Ecto.UUID do
   def cast(_), do: :error
 
   @doc """
-  Converts an string representing a UUID into a binary.
+  Converts a string representing a UUID into a binary.
   """
   def dump(<< u0::64, ?-, u1::32, ?-, u2::32, ?-, u3::32, ?-, u4::96 >>) do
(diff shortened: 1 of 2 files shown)


josevalim:
@joshwlewis shouldn't you rather declare the field as :uuid in the database? It is going to take less space in terms of storage too and definitely work as Ecto.UUID is meant to work on :uuid types.

joshwlewis:
D'oh... yep. It definitely works fine when you use the right database type.

josevalim:
I will reopen up an issue to ensure we emit at least a better error message. :) Thanks for reporting!

PR #617, opened 2015-05-23
Fix formatting of example docs in Ecto.Type

gjaldon:
http://i.glui.me/1F2z6Zl

Example in docs aren't formatted properly. It looks it doesn't recognize the format of `def dump(_), do: :error`.


josevalim:
This is an issue that is fixed on latest ex_doc. I have update ex_doc deps and we should be good the next time we push the docs. Thanks for reporting!

gjaldon:
:+1:

PR #622, opened 2015-05-26
Move transaction state from worker to client

### LINKED ISSUE #609: Test transaction/sandboxing lost if worker exits
If the (single) worker exits when using a test transaction the replacement worker does not sandbox its connection.

Using the simple example after changing the size to 1:
```elixir
:ok = Ecto.Adapters.SQL.begin_test_transaction(Simple.Repo)
poolboy = elem(Simple.Repo.__pool__, 0)
worker = :poolboy.checkout(poolboy, true, :infinity) 
true = :sys.get_state(worker).sandbox
Process.exit(worker, :kill)
worker = :poolboy.checkout(poolboy, true, :infinity)
false = :sys.get_state(worker).sandbox
```

josevalim:
This pull request mixes the ideas in #612 with a proposal outlined [here](https://gist.github.com/josevalim/733988439cfbb964c8e5). Overall, the idea is to make the worker dumb and simply map client <-> worker <-> connection. The benefit is that it is no longer SQL specific.

This also fixes the bug where restart_test_transaction would be unable to catch up if the connection dropped.

I have yet to merge the tests added by @fishcakez on #612, although all integration tests pass but unit tests for the worker needs to be rewritten. If this is merged, we should also improve the docs by "merging" the gist above.

Closes #609 #610 #611 #612


lib/ecto/adapters/sql.ex
```diff
@@ -431,73 +451,174 @@ defmodule Ecto.Adapters.SQL do
...
+  end
+
+  defp worker_fuse(pool, worker, timeout, fun) do
+    fun.()
+  catch
+    kind, reason ->
+      # If it fails, we don't know the connection state
+      # so we need to break the transaction.
+      stack = System.stacktrace()
+      Process.delete({:ecto_transaction_info, pool})
```
josevalim: We can't delete this.

Issue #624, opened 2015-05-28
Introducing id and binary_id


josevalim:
Different databases (or even within the same database), they are two types of primary keys:

  1. the ones (auto) generated in the database, for example, integer serial keys 
  2. the ones (auto) generated in the client, for example, UUID in postgres

In Ecto, only the first is well supported via the `:read_after_writes` option. Read after writes (RAW)  allow us to read a particular field back after performing writes (insert or update).

The motivation for this proposal, however, is the lack of support for the second option. Today you need to use a UUID column and then either: 1. use `:read_after_writes` and generate the value in the database, or 2. define a callback that will generate the field for you on the client. The callback option is considerably verbose though, so many end-up relying on RAW once again which, as we will see next, adds its own problems.

## Read after writes (RAW)

The `:read_after_writes` option can be given per field in your schema:

    field :uuid, Ecto.UUID, read_after_writes: true

It is also important to mention that fields declared as RAW are not sent to the database on insert if they are nil. The trouble with RAW is that they are not properly supported by all databases:

  1. PostgreSQL and MSSQL supports RAW without issues due to RETURNING clauses
  2. MySQL does not support RAW. It can only read the primary key back
  3. NoSQL databases do not support RAW. However, since most of them don't have trigger and similar features, it doesn't really matter, as the value sent is always the value read back

In other words, users that rely heavily on RAW in their schema are likely writing a schema that depends on features that are database specific. Although Ecto embraces the diversity and supports database specific features, we should aim for a common ground as much as possible and proper primary key support is essential for this.

## Auto-generation

Instead of RAW, we actually want to support auto-generation of values:

    field :uuid, Ecto.UUID, autogenerate: true

This pushes the responsibility to the client without any drawback as UUIDs are meant to be generated on the client too. Passing `:autogenerate` is going to invoke a function on the type (likely `generate/0`) before inserting it to the database.

## UUIDs? 

Although we have used only UUID as examples, the truth is that many databases have their own ID type. PostgreSQL support UUIDs that can be generated in the client or server, while MySQL requires them to emulated in the client, Mongo provides ObjectID which is client-based while OrientDB uses RecordID which is generated in the database.

What all those types share in common is that they are ultimately stored as binaries. Hence we propose two new types: `:id` (supporting the current integer fields) and  `:binary_id` (for the binary ones).

[...]

PR #625, opened 2015-05-28
Introduce id, binary_id and autogenerate

josevalim:
See #624 for a deeper discussion.

* [x] Remove :uuid (Ecto.UUID still works)
* [x] Introduce binary literal in queries <<0, 1, 2>>
* [x] Support autogenerate for custom types
* [x] Add id type
* [x] Add autogenerate for id type as primary key
* [x] Add binary_id type
* [x] Add autogenerate for binary_id type as primary key
* [x] Revert the recent behaviour where `read_after_writes` are not sent on insert when nil
* [x] Deprecate `read_after_writes`


josevalim:
This is ready for review. I just need to add two or three unit tests for type normalisation and deprecate read_after_writes.

integration_test/pg/test_helper.exs
```diff
@@ -1,6 +1,10 @@
 Logger.configure(level: :info)
 ExUnit.start
 
+# Configure Ecto for support and tests
```
josevalim: @mobileoverlord you will have to set those too and you can remove the lock_for_update down below.

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -316,7 +316,12 @@ if Code.ensure_loaded?(Mariaex.Connection) do
       raise ArgumentError, "Array type is not supported by MySQL"
     end
 
-    defp expr(%Ecto.Query.Tagged{value: other, type: type}, sources) when type in [:integer, :float] do
+    defp expr(%Ecto.Query.Tagged{value: binary, type: :binary}, _sources) when is_binary(binary) do
```
josevalim: @mobileoverlord this was here in the past. Then we removed and we added back. If you haven't removed it, you should be fine.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -577,6 +582,8 @@ if Code.ensure_loaded?(Postgrex.Connection) do
     end
 
     defp ecto_to_db({:array, t}), do: ecto_to_db(t) <> "[]"
+    defp ecto_to_db(:id),         do: "integer"
```
josevalim: @mobileoverlord you want to add those two types too.

test/ecto/adapters/postgres_test.exs
```diff
@@ -191,6 +191,9 @@ defmodule Ecto.Adapters.PostgresTest do
     query = Model |> select([], "abc") |> normalize
     assert SQL.all(query) == ~s{SELECT 'abc' FROM "model" AS m0}
 
+    query = Model |> select([], <<0, ?a,?b,?c>>) |> normalize
```
josevalim: @mobileoverlord You may want to add this assertion.

test/ecto/adapters/postgres_test.exs
```diff
@@ -45,7 +45,7 @@ defmodule Ecto.Adapters.PostgresTest do
   end
 
   defp normalize(query) do
-    {query, _params} = Ecto.Query.Planner.prepare(query, [])
+    {query, _params} = Ecto.Query.Planner.prepare(query, [], %{})
```
josevalim: @mobileoverlord You need to update your call to Planner.prepare (if you do any).

test/ecto/adapters/postgres_test.exs
```diff
@@ -199,7 +202,7 @@ defmodule Ecto.Adapters.PostgresTest do
   end
 
   test "tagged type" do
-    query = Model |> select([], type(^<<0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15>>, :uuid)) |> normalize
+    query = Model |> select([], type(^"601d74e4-a8d3-4b6e-8365-eddb4c893327", Ecto.UUID)) |> normalize
```
josevalim: @mobileoverlord  `:uuid` is no longer, so this needs to be `Ecto.UUID`.

lib/ecto/repo/model.ex
```diff
@@ -181,7 +201,8 @@ defmodule Ecto.Repo.Model do
   end
 
   defp with_transactions_if_callbacks(repo, adapter, model, opts, callbacks, fun) do
-    if Enum.any?(callbacks, &function_exported?(model, &1, 1)) do
+    if Enum.any?(callbacks, &function_exported?(model, &1, 1)) and
```
josevalim: @michalmuskala Due to this change, you no longer need to declare `def transaction` in your adapter. I mean, you probably should, but you can simply raise now saying transactions are not supported.

josevalim: Sorry, you need to remove the function definition. Otherwise this code will be triggered anyway.

integration_test/pg/test_helper.exs
```diff
@@ -1,6 +1,10 @@
...
+Application.put_env(:ecto, :lock_for_update, "FOR UPDATE")
+Application.put_env(:ecto, :primary_key_type, :id)
```
josevalim: @michalmuskala You can try setting it to `:binary_id` now!

lib/ecto/adapters/sql.ex
```diff
@@ -60,7 +65,22 @@ defmodule Ecto.Adapters.SQL do
       end
 
       @doc false
-      def insert(repo, source, params, returning, opts) do
+      # Nil ids are generated in the database.
+      def insert(repo, source, params, {key, :id, nil}, returning, opts) do
```
josevalim: @michalmuskala Note `insert/6` receives a new argument now. You will need to check for :id and :binary_id in your mongo adapter. Let's talk about this whenever you can.

PR #628, opened 2015-05-31
The code says to import only `from/2`

ma2gedev:


josevalim:

PR #631, opened 2015-05-31
Add join-based tags for some integration tests

jazzyb:
SQLite does not support right outer joins or any joins in update and delete queries.


josevalim:

PR #636, opened 2015-06-01
Conditional load of mysql drivers

tjheeta:
This doesn't quite work properly because conditionally loading aliases doesn't seem to work. Not sure what the correct way to do this is.


lib/ecto/adapters/mysql/connection.ex
```diff
@@ -1,6 +1,18 @@
-if Code.ensure_loaded?(Mariaex.Connection) do
+# Does not compile without specifying this 
+alias Mysqlex, as: MysqlDriver
```
tjheeta: Getting a struct not defined error on compile if this line is not included, which foregoes the point of conditionally including drivers.

PR #637, opened 2015-06-02
Tidy integration tests to support mongo adapter

michalmuskala:
The first commit splits all the integration tests that use joins (that are strictly a SQL feature) to a separate file.

The second one tries to amend some tests that relied on the assumption that primary key is always integer, as it's no longer true. There are some tests that are broken for Mongo that I don't know how to fix. These are:
* [preload with binary_id](https://github.com/elixir-lang/ecto/blob/c209efe05f2e07327435e06f661d32a9780b41a8/integration_test/cases/preload.exs#L255-261) - assumes that binary_id is uuid, that is not true for mongo.
* [insert with user-assigned primary key](https://github.com/elixir-lang/ecto/blob/c209efe05f2e07327435e06f661d32a9780b41a8/integration_test/cases/repo.exs#L102-104) - assumes primary key is integer
* [insert and update with user-assigned primary key in changeset](https://github.com/elixir-lang/ecto/blob/c209efe05f2e07327435e06f661d32a9780b41a8/integration_test/cases/repo.exs#L107-113) - assumes primary key is integer
* [validate_unique/3](https://github.com/elixir-lang/ecto/blob/c209efe05f2e07327435e06f661d32a9780b41a8/integration_test/cases/repo.exs#L188-206) - `validate_unique/3` with `downcase: true` uses SQL fragments `lower(?)` that I'm unable to support.
* [get(!)](https://github.com/elixir-lang/ecto/blob/c209efe05f2e07327435e06f661d32a9780b41a8/integration_test/cases/repo.exs#L236-250) - at line 242 there's a `-1` I'm not sure what to do about.

I haven't considered all the tests that are related to updates or deletes as I haven't implemented them yet, so it's hard to tell what wont work.

### DIFF
--- a/integration_test/cases/joins.exs
+++ b/integration_test/cases/joins.exs
@@ -1,0 +1,333 @@
+defmodule Ecto.Integration.JoinsTest do
+  use Ecto.Integration.Case
+
+  require Ecto.Integration.TestRepo, as: TestRepo
+  import Ecto.Query
+
+  alias Ecto.Integration.Post
+  alias Ecto.Integration.PostUsecTimestamps
+  alias Ecto.Integration.Comment
+  alias Ecto.Integration.Permalink
+  alias Ecto.Integration.User
+  alias Ecto.Integration.Custom
+  alias Ecto.Integration.Barebone
+  alias Ecto.Schema.Metadata
+
+  @tag :update_with_join
+  test "update all with joins" do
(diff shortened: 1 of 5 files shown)


josevalim:

josevalim:
As per our discussion, I will fix second and third items, the other 3 are up to you. :)

Issue #638, opened 2015-06-02
Tag Migrations Relying on Modification and Removal of Columns


jazzyb:
https://github.com/elixir-lang/ecto/blob/master/integration_test/support/migration.exs#L76-L81

SQLite cannot alter or drop columns on tables.  Would it be possible to tag the `modify` and `remove` lines in such a way that they are not run, e.g.:

```elixir
    alter table(:users) do
      if can_alter_column?, do: modify :name, :text
      add :custom_id, :uuid
      if can_drop_column?, do: remove :to_be_removed
      timestamps
    end
```

or Is there some way to tag tests relying on the above migration in such a way that the migration is never run?

I haven't looked too deeply into this yet, so I wanted to open this issue to discuss it.

josevalim:
@jazzyb We should probably move those to a separate test. We were trying to optimize by using the migrations themselves as tests but we should probably move them apart.

jazzyb:
I will work on separating these.  I might not be able to get to it until this weekend.

josevalim:
@jazzyb there is no need to rush. I believe you are the only who needs it anyway. :)

josevalim:
Btw, once SQlite3 tests are passing, feel free to send a pull request to update the table in the README. :) Thank you for working on it!

jazzyb:
Thanks for all your help.  I'll send you a PR once the integration tests are working.

PR #646, opened 2015-06-06

jazzyb:
Moves the modify and remove column tests from integration_test/support/migration.exs to separate tests in integration_test/sql/migration.exs.  The new alter table tests are tagged so that sqlite_ecto can exclude them.

This will address #638.

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -110,36 +110,54 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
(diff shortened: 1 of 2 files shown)


jazzyb:
I'm not sure what's causing the MySQL errors.  I will investigate and commit a fix.

jazzyb:
I can't seem to track down the error.  The MySQL savepoint seems to be disappearing after the test table is created in either of the two new tests, but I don't know why.  Closing this PR until I've fixed the error.

PR #647, opened 2015-06-07
Add ability to rename columns

aforward:
I am looking to add support for renaming columns in ecto migrations.

Here is the documentation I referred to regarding the Postgres and MySQL syntax.

http://www.postgresql.org/docs/9.4/static/sql-altertable.html

https://dev.mysql.com/doc/refman/5.1/en/alter-table.html

### DIFF
--- a/integration_test/support/migration.exs
+++ b/integration_test/support/migration.exs
@@ -68,6 +68,7 @@ defmodule Ecto.Integration.Migration do
 
     create table(:users) do
       add :name, :string
+      add :details, :string
       add :to_be_removed, :string
     end
 
@@ -75,6 +76,7 @@ defmodule Ecto.Integration.Migration do
 
     alter table(:users) do
       modify :name, :text
+      rename :details, :description
       add :custom_id, :uuid
       remove :to_be_removed
       timestamps
--- a/integration_test/support/migration.exs
+++ b/integration_test/support/migration.exs
@@ -76,10 +76,13 @@ defmodule Ecto.Integration.Migration do
 
     alter table(:users) do
       modify :name, :text
-      rename :details, :description
       add :custom_id, :uuid
       remove :to_be_removed
       timestamps
+    end
+
+    alter table(:users) do
+      rename :details, :description, :string
     end
 
     index = index(:users, [:custom_id], unique: true)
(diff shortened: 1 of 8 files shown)


aforward:
I will have to look more closely at the integration tests to ensure the RENAME syntax is correct

```
** (Postgrex.Error) ERROR (syntax_error): syntax error at or near "RENAME"
    (ecto) lib/ecto/adapters/sql.ex:210: Ecto.Adapters.SQL.log_and_check/7
    (ecto) lib/ecto/adapters/postgres.ex:59: Ecto.Adapters.Postgres.execute_ddl/3
    integration_test/support/migration.exs:77: Ecto.Integration.Migration.create_users_table/0
    integration_test/support/migration.exs:23: Ecto.Integration.Migration.change/0
    (stdlib) timer.erl:194: :timer.tc/3
    (ecto) lib/ecto/migration/runner.ex:22: Ecto.Migration.Runner.run/6
    (ecto) lib/ecto/migrator.ex:113: Ecto.Migrator.attempt/6
    (ecto) lib/ecto/migrator.ex:63: anonymous fn/4 in Ecto.Migrator.do_up/4
```

integration_test/support/migration.exs
```diff
@@ -80,6 +81,10 @@ defmodule Ecto.Integration.Migration do
       timestamps
     end
 
+    alter table(:users) do
```
aforward: Renaming a column is not an "action" as defined in
http://www.postgresql.org/docs/9.4/static/sql-altertable.html

So it currently must be run in a separate alter table(:xxx) call.  Is something like that acceptable?  Or, should we detect it in the backend and ensure two multiple alter tables are executed when a rename is used.

josevalim: We don't necessarily need to do separate calls but we do need to generate a special SQL for postgres. So we need to traverse the alter table instructions, finding all rename commands, and then group all queries by using ";" as separator. If that doesn't work, we can try BEGIN END too:

```sql
BEGIN
  ALTER TABLE foo rename 1;
  ALTER TABLE foo rename 2;
  ALTER TABLE foo others
END;
```

Hopefully one of those ideas work. :)

josevalim:
Ping. :)

integration_test/support/migration.exs
aforward: Will do.  I'll attack this on the weekend.

josevalim:
Closing this in favor of #807. Thanks!

aforward:
Sorry I couldn't pick up the slack here.

PR #650, opened 2015-06-08

jazzyb:
Moves the modify and remove column tests from integration_test/support/migration.exs to separate tests in integration_test/sql/migration.exs. The new alter table tests are tagged so that sqlite_ecto can exclude them.

This will address #638.

There are a number of details about the changes I would like to discuss.  I will comment on them inline in the code.

### DIFF
--- a/integration_test/mysql/test_helper.exs
+++ b/integration_test/mysql/test_helper.exs
@@ -43,7 +43,8 @@ defmodule Ecto.Integration.Case do
   end
 
   setup do
-    Ecto.Adapters.SQL.restart_test_transaction(TestRepo, [])
+    Ecto.Adapters.SQL.rollback_test_transaction(TestRepo, [])
+    Ecto.Adapters.SQL.begin_test_transaction(TestRepo, [])
     :ok
   end
 end
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -110,36 +110,54 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule AlterColumnMigration do
+    use Ecto.Migration
+
+    def change do
+      create table(:migration_test_table) do
+        add :to_be_modified, :integer
+        add :to_be_removed, :integer
+      end
+
+      alter table(:migration_test_table) do
+        modify :to_be_modified, :string
+      end
+
+      execute "INSERT INTO migration_test_table (to_be_modified) VALUES ('foo')"
+    end
+  end
+
+  defmodule DropColumnMigration do
+    use Ecto.Migration
+
+    def change do
+      create table(:migration_test_table) do
+        add :to_be_modified, :string
+        add :to_be_removed, :integer
+      end
+
+      alter table(:migration_test_table) do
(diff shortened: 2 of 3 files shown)


integration_test/mysql/test_helper.exs
```diff
@@ -43,7 +43,8 @@ defmodule Ecto.Integration.Case do
   end
 
   setup do
-    Ecto.Adapters.SQL.restart_test_transaction(TestRepo, [])
+    Ecto.Adapters.SQL.rollback_test_transaction(TestRepo, [])
+    Ecto.Adapters.SQL.begin_test_transaction(TestRepo, [])
```
jazzyb: The modify and remove column tests that are being added with this PR require `CREATE` and `ALTER TABLE` statements to be issued to the database.  These statements will cause an implicit commit of any transactions or savepoints in MySQL.  This causes the test transaction that was being restarted here to be released, and any tests that are performed after the new tests will fail with the error "SAVEPOINT ecto_sandbox does not exist".  The above change provides virtually the same functionality as `restart_test_transaction` but does not result in MySQL errors.

I tried a few other changes to work around the error, but this seemed like the most straight-forward.  Is there a better work-around that I should implement instead?

integration_test/sql/migration.exs
```diff
@@ -101,6 +101,75 @@ defmodule Ecto.Integration.MigrationTest do
...
+  end
+
+  @tag :remove_column
+  test "remove column" do
+    import Ecto.Query, only: [from: 2]
+    assert :ok == up(TestRepo, 20090906120000, DropColumnMigration, log: false)
+    assert catch_error(TestRepo.one from p in MigrationTestTable, select: p.to_be_removed)
+    unless TestRepo.adapter == Ecto.Adapters.Postgres do
+      :ok = down(TestRepo, 20090906120000, DropColumnMigration, log: false)
+    end
```
jazzyb: Since the `up/4` migration will implicitly commit the MySQL test transaction, the `down/4` migration will clean-up after the test by dropping the test table.  This is only necessary for MySQL -- for PostgreSQL the `down/4` call is unnecessary because rolling back the test transaction will drop the table anyway.

The test asserts that the `to_be_removed` column was removed by looking for an error when trying to access the non-existent row.  The `down/4` call is wrapped in an if-check because PostgreSQL does not allow any more statements to be made within a transaction once an error has been thrown.  Thus, the test table cannot be dropped by the Postgres adapter.

I do not like calling out the Postgres adapter specifically in a test that is meant to be generic, but I couldn't think of a better way to get around the PostgreSQL error.  Is there a better, more generic way of cleaning up the test for MySQL without running into an error with PostgreSQL?  The assertion also relies on an error being raised.  Is there a generic Ecto way to check the absence of a column in a table?  If there was, that would remove the need for the if-check.

josevalim: Yes. You will need to write the tests to create/change/delete tables that are not used anywhere else (name them migration_1, migration_2 and so on) and not use Ecto.Integration.Case in them as MuSQL autocommits. :(

jazzyb:
New PR #651

PR #651, opened 2015-06-09

jazzyb:
Moves the modify and remove column tests from integration_test/support/migration.exs to separate tests in integration_test/sql/alter_table.exs.  The new tests are tagged to allow exclusion.

Addresses #638.


integration_test/support/migration.exs
```diff
@@ -67,19 +67,13 @@ defmodule Ecto.Integration.Migration do
...
-
-    alter table(:users) do
-      modify :name, :text
+      add :name, :text
       add :custom_id, :uuid
-      remove :to_be_removed
       timestamps
     end
 
+    true = exists? table(:users)
```
josevalim: Let's move this whole thing to the new file. We can keep now the `migration.exs` just the straight-up table definition and let's add all the migration logic checking to the new one. This way everything becomes clearer, wdyt? :D

jazzyb: Sounds good.  In that case do you have a good name for the new test file?  `alter_table.exs` doesn't fit as well if we'll be testing the creating and dropping of tables and indices, too.  Maybe, `table_index.exs`?

josevalim: I would call the new test file "migration.exs" and rename the existing one to "migrator.exs".

josevalim: Or even simpler... just put everything into the same migration.exs. Both files are short.

jazzyb:
@josevalim When you get a chance, can you restart the tests for this commit?  I cannot reproduce the error on my side, but the timeout may be due to the DDoS against Travis right now.

jazzyb:
@josevalim Ping?  ;)  This PR should be ready to go once someone restarts the Travis test.  See my comment above.

josevalim:
Done, sorry for the delay, busy travelling!

josevalim:
Please note that I have separated the migrator tests from migration ones in a later commit. Both you and @mobileoverlord will have to update your requires to add the new entry: https://github.com/elixir-lang/ecto/commit/a489193c62373c5f5c9de706aea81a4275876df3

jazzyb:
Thank you so much!

PR #654, opened 2015-06-10
Update mariaex to 0.2.1

liveforeverx:
mariaex has backwards incompatible changes, no more sanitising of inputs, and performance improvements.


josevalim:
Thanks! There is a failure in the build, you can try it out by running `MIX_ENV=mysql mix test`. Thoughts?

liveforeverx:
Yes, I've tried to delete in some circumstances the same statement twice, it is fixed now in 0.2.1.

PR #657, opened 2015-06-12
Add assertion on DDL concurrency

### LINKED ISSUE #656: [MySQL] - Assert that full concurrency is allowed during the DDL
The [Online DDL documentation](https://dev.mysql.com/doc/refman/5.6/en/innodb-create-index-overview.html) says that MySQL 5.6+ supports concurrency on DDL. Although no syntax change is required in the `CREATE INDEX` or `DROP INDEX` commands, it's possible to assert that the concurrency is allowed using `LOCK=NONE`.

We could add this assertion on the MySQL adapter. @josevalim what do you think?

Thanks @morgo for pointing it on this commit https://github.com/laurocaetano/ecto/commit/0d807e27b65c235b7779b8d8945b2a5adfbdf383#commitcomment-11634210

laurocaetano:
MySQL 5.6+ supports concurrency on DDL statements. This feature is
called Online DDL. There is syntax changes to enable the Online
DDL, but an assertion can be made to check if full concurrency is
allowed during the DDL.

More information can be found here:
https://dev.mysql.com/doc/refman/5.6/en/innodb-create-index-overview.html

Closes #656

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -408,23 +408,23 @@ if Code.ensure_loaded?(Mariaex.Connection) do
     end
 
     def execute_ddl({:create, %Index{}=index}) do
-      if index.concurrently, do: raise(ArgumentError, "CONCURRENTLY is not supported by MySQL")
-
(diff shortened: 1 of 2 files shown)


josevalim:

PR #669, opened 2015-06-16
[WIP] Separate transaction handling and support custom pools

fishcakez:
Work in progress branch to allow custom pools (i.e. other than `poolboy`).

Firstly all `poolboy` usage is abstracted out of `Ecto.Adapters.SQL` to `Ecto.Adapters.Poolboy` so that other adapters can make use of it and the pdict usage is in this module to clean up `Ecto.Adapters.SQL`. The re factor allowed an optimisation where the final log in a transaction or simple query was always moved outside the `poolboy` transaction in some cases (I don't think this was always the case previously).

Secondly the transaction handling that can be abstracted out of the `Ecto.Adapters.Poolboy` is moved to `Ecto.Adapters.Pool.Transaction` so that other pools can make use of it and don't have to reimplement the pdict dance. This change also allows custom pools for SQL adapters.

Thirdly an optimisation is done to `Ecto.Adapters.Poolboy.Worker` (previously `Ecto.Adapters.Worker`) where a call is turned into a cast as it does not need to be synchronous.


josevalim:
@fishcakez this looks great, thank you! I have two questions only:

1. Maybe we should split integration_test into adapter_test and pool_test? It is already getting complex because of multiple adapters support and mixing the pool stuff doesn't help.

2. Can you use the transaction stuff with sbroker? Because, if you cannot, I would rather just merge the pool transaction into poolboy and make transactions part of the pool API.

josevalim:
Ah, there is something else. For example, MongoDB does not provide transactions but its adapter need pool transactions (take connections out of the pool, send something and put it the back). So, in one way or the other, all pool implementations need to be able to just execute something and its current behaviour has no such function.

The issue is that we have two transactions, "pool transactions" (which are stateless) and "adapter transactions" (which are stateful). So even if I just want to check the connection from the pool out for a quick query (stateless), the worker is going through the trouble of monitoring the caller, keeping references and so on, which is likely unnecessary. So we may need one extra function in there called run or execute, which is default part of the Pool behaviour, that just executes some function passing the code without the adapter transaction semantics, worker monitoring and so on.

lib/ecto/adapters/pool/transaction.ex
```diff
@@ -0,0 +1,312 @@
...
+  `{:error, :noconnect}` if a connection is not available.
+  """
+  defcallback open_transaction(Pool.t, timeout) ::
+    {mode, conn, queue_time} | {:error, :noproc | :noconnect} when conn: any
+
+  @doc """
+  Get the connection module and process from the connection data for the current
+  transaction.
+  """
+  defcallback transaction_connection(Pool.t, conn, timeout) ::
```
josevalim: I am assuming you need this function for sbroker? Otherwise, why wouldn't you simply return the connection as a result of open_transaction?

fishcakez:
> Maybe we should split integration_test into adapter_test and pool_test? It is already getting complex because of multiple adapters support and mixing the pool stuff doesn't help.

I agree that re-organisation is required. I would like to test each adapter with each pool and allow non-built in adapters/pools to be tested in any suitable combination. How would we achieve this?

> Can you use the transaction stuff with sbroker? Because, if you cannot, I would rather just merge the pool transaction into poolboy and make transactions part of the pool API.

Yes definitely. A pool for `sbroker` and `dispcount` could be written with this branch's API without any changes to ecto. The pdict stuff is complicated and it makes sense to share it.

> I am assuming you need this function for sbroker? Otherwise, why wouldn't you simply return the connection as a result of open_transaction?

It's not required. We could remove this function and have `open_transaction` return `{mode, worker, {conn_mod, conn}, queue_time}`.

I plan to release an sbroker pool as a separate repo and a new version of sbroker on monday or early next week to improve performance for ecto. Part of this is to allow the broker to send the `conn_mod` and `conn` directly to the client process so a call is not required to retrieve it.

> The issue is that we have two transactions, "pool transactions" (which are stateless) and "adapter transactions" (which are stateful). So even if I just want to check the connection from the pool out for a quick query (stateless), the worker is going through the trouble of monitoring the caller, keeping references and so on, which is likely unnecessary. So we may need one extra function in there called run or execute, which is default part of the Pool behaviour, that just executes some function passing the code without the adapter transaction semantics, worker monitoring and so on.

I agree. It was easier to refactor with just one type of pool transaction. I will add `ask` and `done` callbacks to `Ecto.Adapters.Pool`. This module, rather than the callback module, will need to control the `run` function so it can pickup the connection information inside a transaction. Perhaps:

```elixir
defcallback ask(pool, timeout) :: {:ok, worker, {conn_mod, conn}, queue_time} | {:error, :noproc | :noconnect}
defcallback done(pool, worker, timeout) :: :ok
def run(pool_mod, pool, timeout, (({module, pid}, queue_time) -> result)) :: {:ok, result} | {:error, :noproc | :noconnect} when result: var
```
May need to move `break_transaction` callback to Pool and rename it.

josevalim:
>  I would like to test each adapter with each pool and allow non-built in adapters/pools to be tested in any suitable combination. How would we achieve this?

We could pass the adapter as an environment variable that would be read by PG or MySQL This way you could run it as: `MIX_ENV=pg ECTO_POOL=Ecto.Adapters.Sbroker mix test`. This would require running tests from Ecto though. If you want to run from your project, we would need to include all of integration_test (both adapter and pool test) in the hex package.

> I agree. It was easier to refactor with just one type of pool transaction. I will add ask and done callbacks to Ecto.Adapters.Pool. This module, rather than the callback module, will need to control the run function so it can pickup the connection information inside a transaction.

What not just run then? :) Can you think of any occasion where you would like to ask and done explicitly apart from each other? This way, we don't even need break transaction, it would be all the adapter responsibility.

fishcakez:
> What not just run then? :) Can you think of any occasion where you would like to ask and done explicitly apart from each other? This way, we don't even need break transaction, it would be all the adapter responsibility.

Isn't there an issue with SQL adapters when inside a transaction that a query needs to fetch the worker/connection from the transaction pdict entry. I guess we we fix this by merging `Pool.Transaction` into `Pool`.

fishcakez:
Also I think on IRC with @ericmj  we discussed that if a query times out we should close the connection from the client?

josevalim:
> Isn't there an issue with SQL adapters when inside a transaction that a query needs to fetch the worker/connection from the transaction pdict entry. I guess we we fix this by merging Pool.Transaction into Pool.

Yes, you are right. Maybe we should just merge them then and ensure transaction goes through the fast path (i.e. pool transaction only) if there is no open transaction. This way it is completely transparent for the adapter.

fishcakez:
> We could pass the adapter as an environment variable that would be read by PG or MySQL This way you could run it as: MIX_ENV=pg ECTO_POOL=Ecto.Adapters.Sbroker mix test. This would require running tests from Ecto though. If you want to run from your project, we would need to include all of integration_test (both adapter and pool test) in the hex package.

Ugh the later is confusing. Lets go for the ECTO_POOL approach and keep the pool implementation in ecto.

fishcakez:
I have done the updates discussed except I have not changed the integration tests configuration. The pool tests I have added could use soem reorganised first.

lib/ecto/adapters/pool.ex
```diff
@@ -0,0 +1,450 @@
...
+
+  @doc """
+  Checkout a connection from a pool.
+
+  The connection should not be closed if the calling process exits without
+  returning the connection.
+
+  Returns `{mode, worker, conn, queue_time}` on success, where `worker` is the
+  worker term and conn is a 2-tuple contain the connection's module and
+  pid. The `conn` tuple can be retrieved inside a `transaction/4` with
```
josevalim: There is no `connection/1`.

lib/ecto/adapters/pool.ex
```diff
@@ -0,0 +1,450 @@
...
+      %{conn: _ , mode: ^mode} ->
+        {:error, :already_mode}
+      %{conn: _} = info ->
+        mode(ref, info, mode, timeout)
+      %{} ->
+        {:error, :noconnect}
+    end
+  end
+
+  @doc """
```
josevalim: Do we still need this?

lib/ecto/adapters/pool.ex
josevalim: Gah, ignore me.

josevalim:
The tests look fine. My only comment is, if we don't actually need a database in the pool tests, shouldn't they simply be in `test` instead of integration and we would always run them? Or do you plan to change this at some point?

josevalim:
@fishcakez this looks great! My last concern is regarding the naming... is there a reason why we chose `ask`, `done` and `disconnect` instead of `checkout`, `checkin` and `break`? I am mentioning those because those are the terms you use in the docs (`checkout` and `break` mainly).

lib/ecto/adapters/pool.ex
```diff
@@ -0,0 +1,450 @@
...
+      %{conn: _} = info ->
+        do_run(ref, info, nil, fun)
+      %{} ->
+        {:error, :noconnect}
+    end
+  end
+
+  @doc """
+  Carry out a transaction using a connection from a pool.
+
```
josevalim: I would add something like:

    Once a transaction is opened, all following calls to `run/4` or `transaction/4` will use the same
    connection/worker. If `disconnect/1` is invoked, all operations will return `{:error, :noconnection}`
    until the end of the transaction.

fishcakez: ok

fishcakez:
> The tests look fine. My only comment is, if we don't actually need a database in the pool tests, shouldn't they simply be in test instead of integration and we would always run them? Or do you plan to change this at some point?

I wasn't sure the best way to test custom pools. When I do an `sbroker` PR (and maybe a `dispcount` one) how can the new pool run the same set of tests? I guess we can always run all the tests for all the pools and then test the adapters with `ECTO_POOL` pool for integration tests?

fishcakez:
> @fishcakez this looks great! My last concern is regarding the naming... is there a reason why we chose ask, done and disconnect instead of checkout, checkin and break? I am mentioning those because those are the terms you use in the docs (checkout and break mainly).

I think I use those words because the docs originate from a previous function. I definitely prefer `break` and will change that. I can reword the docs or rename to checkin/checkout as you wish. I preferred ask/done simpler because the words are easier to differentiate between when reading the code.

josevalim:
> I wasn't sure the best way to test custom pools. When I do an sbroker PR (and maybe a dispcount one) how can the new pool run the same set of tests? I guess we can always run all the tests for all the pools and then test the adapters with ECTO_POOL pool for integration tests?

Let's leave it like this for now then until we get another pool. Then we can revisit.

> I think I use those words because the docs originate from a previous function. I definitely prefer break and will change that. I can reword the docs or rename to checkin/checkout as you wish. I preferred ask/done simpler because the words are easier to differentiate between when reading the code.

Another argument for checkin and checkout is that they would be clearer for someone reading the code as it is common terminology. But I don't have strong feelings about this, it is completely up to you.

fishcakez:
> Let's leave it like this for now then until we get another pool. Then we can revisit.

Great.

> Another argument for checkin and checkout is that they would be clearer for someone reading the code as it is common terminology. But I don't have strong feelings about this, it is completely up to you.

OK. Lets call it checkin/checkout.

fishcakez:
Updated

josevalim:
@fishcakez let's merge this?

fishcakez:
Yeah, sure.

fishcakez:
Theres a typo in docs, `:noconnection` should be `:noconnect`. Shall I amend/force push?

josevalim:
@fishcakez you can just push a new commit. Whatever you prefer.

lib/ecto/adapters/poolboy/worker.ex
```diff
@@ -0,0 +1,188 @@
...
+  ## Open transaction
+
+  def handle_call(:open_transaction, {pid, _},
+  %{transaction: nil, mode: mode} = s) do
+    {:reply, {mode, modconn(s)}, monitor(pid, s)}
+  end
+
+  def handle_call(:open_transaction, from,
+  %{transaction: {client, _}, mode: mode} = s) do
+    if Process.is_alive?(client) or mode === :sandbox do
```
josevalim: Why do we need to check if the client is alive?

fishcakez: Is this chain of events possible
* Client crashed and starts exiting
* client :DOWN to poolboy
* poolboy receives :DOWN and sends worker pid to new client
* new client sends open_transaction/checkout
* old client :DOWN to worker

Maybe this is one to ask Concuerror.

fishcakez: We should fix this with a different method. If there is a monitor active and we get a checkout requests we treat it as a break + checkout.

PR #680, opened 2015-06-23
Added validate_subset to Ecto.changeset

zmoshansky:
Creates a validator for lists to check if it is the subset of the given enumerable.

@josevalim The PR with the information discussed in #674 

It does create this warning in the tests that I'm unfamiliar with `[warning] :read_after_writes is deprecated. It was declared for field :x in model Ecto.SchemaTest.AutogenerateFail` after reading about RAW.

I also attempted to cover the following case before realizing that I'm not sure if you can even define an array of arrays in an `Ecto.Model`. This didn't work ` field :topics, {:array, {:array, :string }`, nor would I really expect it to. Thoughts?

```
@whitelist [["a", "b"], ["b", "c"]]
validate_inclusion(:list, [["a", "b"], ["b", "c"], ["a", "c"]], @whitelist)
```

As a follow up to this, is there any value in creating `validate_disjoint_set` (analagous to validate_exclusion) and a `validate_intersection` (at least one item)?

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -729,6 +729,29 @@ defmodule Ecto.Changeset do
   end
 
   @doc """
+  Validates a change, of type enum, is a subset of the given enumerable. Like
+  validate_inclusion for lists.
+
+  ## Options
+
+    * `:message` - the message on failure, defaults to "\#{x} is invalid"
+
+  ## Examples
+
+      validate_subset(changeset, :pets, ["cat", "dog", "parrot"])
+      validate_subset(changeset, :lottery_numbers, 0..99)
+
+  """
+  @spec validate_subset(t, atom, Enum.t) :: t
+  def validate_subset(changeset, field, data, opts \\ []) do
+    validate_change changeset, field, {:subset, data}, fn _, values ->
+      Enum.flat_map values, fn(x) ->
+        if x in data, do: [], else: [{field, message(opts, "#{x} is invalid")}]
+      end
+    end
+  end
+
+  @doc """
   Validates a change is not included in given the enumerable.
 
   ## Options
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -730,7 +730,7 @@ defmodule Ecto.Changeset do
 
   @doc """
   Validates a change, of type enum, is a subset of the given enumerable. Like
-  validate_inclusion for lists.
+  validate_inclusion/4 for lists.
 
   ## Options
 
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -729,6 +729,29 @@ defmodule Ecto.Changeset do
...
+  ## Examples
+
+      validate_subset(changeset, :pets, ["cat", "dog", "parrot"])
+      validate_subset(changeset, :lottery_numbers, 0..99)
+
+  """
+  @spec validate_subset(t, atom, Enum.t) :: t
+  def validate_subset(changeset, field, data, opts \\ []) do
+    validate_change changeset, field, {:subset, data}, fn _, values ->
+      Enum.flat_map values, fn(x) ->
```
josevalim: This can also be: `for x <- value, not x in data, do: [{field, message(opts, "#{x} is invalid")}]`

We also need to figure how to show the error message; If we do interpolation like above, it would be really hard to internationalize it later. :(

lib/ecto/changeset.ex
```diff
@@ -729,6 +729,29 @@ defmodule Ecto.Changeset do
   end
 
   @doc """
+  Validates a change, of type enum, is a subset of the given enumerable. Like
+  validate_inclusion for lists.
```
josevalim: `validate_inclusion/4` for lists.

lib/ecto/changeset.ex
josevalim: Maybe `message(opts, "has an invalid entry")`?

josevalim: If we change to the message above, we need to change the validation because we would add a single error. :S

zmoshansky: Good point about internationalization. Although it would be nice to know what value caused the error, a problem that any array validation would suffer from. Is Ecto.Changeset:46 doing some sort of internationalize work `    less_than:                {&</2,  "must be less than %{count}"}`?

Is it possible to adopt that syntax for multiple errors per field, or is the following preferred to generate a single validation error, free of internationalization problems?

```
case Enum.any?(value, fn(x) -> not x in data end) do
   true -> [{field, message(opts, "has an invalid entry")}]
   false -> []
end
```

josevalim: I am not sure. :( I am thinking we would be fine with a single validation error because if there is a wrong entry, it is very likely an error in the UI (or the user is fiddling with things he is not supposed to). Thoughts?

zmoshansky:
I see your point, It would most likely be a software bug rather than an expected flow where one might present options that are invalid. I'm happy to keep it simple.

josevalim:

PR #698, opened 2015-06-25
Move file_helpers to integration_test/

### LINKED ISSUE #696: Missing requirement when running Ecto 0.12.0 integration tests
I'm trying to run the latest integration tests from Ecto 0.12.0 in hex.pm for sqlite_ecto, and I get the following error:

```
** (Code.LoadError) could not load /vagrant/sqlite_ecto/deps/ecto/test/support/file_helpers.exs
```

The file `ecto/integration_test/cases/migrator.exs` requires a file in the `ecto/test` directory which is not shipped with the ecto hex package.  I had previously been using the github repo as a dependency so I did not notice this earlier.

I've been running the integration tests from the `deps` directory [like so](https://github.com/jazzyb/sqlite_ecto/blob/ecto-0.12.0/integration/sqlite/all_test.exs).  Is there a better way to run the integration tests, or is this a mistake in the hex package?

jazzyb:
fixes #696


jazzyb:
Any chance this can be released as v0.12.1?

josevalim:
Yes, I am thinking I will release v0.12.1 from master with all the latest enhancements since we don't have any backwards incompatible change to the user (and 0.12.0 is broken for mssql/sqlite3 adapters).

jazzyb:
Ecto 0.12.1 works :D

Thanks for your help!

PR #705, opened 2015-06-26
handle on_delete: :nothing, :delete_all, :nillify all on create table in postgres and mysql

hubertlepicki:
This is implementation of feature requested https://github.com/elixir-lang/ecto/issues/703 (add ON DELETE CASCADE/SET NULL implementation)

I would appreciate review, and a suggestion if I should get rid of some code duplication between MySQL and PostgreSQL (the private functions adding clauses) or should it stay like that for readability/clarity.


josevalim:
Perfect as is, thank you! Coupling PG and MySQL may sound interesting at first but I believe we should avoid it. :)

hubertlepicki:
That was my guess.

PR #708, opened 2015-06-26
Run should automatically fuse

josevalim:
The unit tests for the pool are broken but I am sending this to get early feedback from @fishcakez.

### DIFF
--- a/integration_test/mysql/test_helper.exs
+++ b/integration_test/mysql/test_helper.exs
@@ -1,5 +1,6 @@
 Logger.configure(level: :info)
-ExUnit.start exclude: [:array_type, :read_after_writes, :case_sensitive, :uses_usec]
+ExUnit.start exclude: [:array_type, :read_after_writes, :case_sensitive,
+                       :uses_usec, :strict_savepoint]
 
 # Configure Ecto for support and tests
 Application.put_env(:ecto, :primary_key_type, :id)
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -166,8 +166,12 @@ defmodule Ecto.Adapters.SQL do
              %{rows: nil | [tuple], num_rows: non_neg_integer} | no_return
   def query(repo, sql, params, opts \\ []) do
     case query(repo, sql, params, nil, opts) do
-      {:ok, result} -> result
-      {:error, err} -> raise err
+      {{:ok, result}, entry} ->
+        log(repo, entry)
+        result
+      {{:error, err}, entry} ->
+        log(repo, entry)
+        raise err
       :noconnect ->
         # :noconnect can never be the reason a call fails because
         # it is converted to {:nodedown, node}. This means the exit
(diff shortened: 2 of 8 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -514,22 +486,19 @@ defmodule Ecto.Adapters.SQL do
...
 
-    trans_fun = fn(ref, mode, depth, queue_time) ->
-      transaction(repo, ref, mode, depth, queue_time, timeout, opts, fun)
+    trans_fun = fn(ref, {mod, _conn}, mode, depth, queue_time) ->
+      transaction(repo, ref, mod, mode, depth, queue_time, timeout, opts, fun)
     end
 
     case Pool.transaction(pool_mod, pool, timeout, trans_fun) do
-      {:ok, {{:return, result}, entry}} ->
-        log(repo, entry)
```
josevalim: I am aware the goal here was to avoid logging inside the transaction but we already log the begin statement and all other queries inside the transaction, so logging the commit or rollback one is not a big deal and it cleans up the code considerably.

fishcakez:
I'm worried about the situation where `log/2` raises an exception after a begin. This is carried out outside of a fuse in a normal transaction and when setting up a sandbox. If `log/2` raises the `begin` is not rolledback.

josevalim:
@fishcakez very good points. I will investigate it soon.

fishcakez:
> @fishcakez very good points. I will investigate it soon.

Please add tests for these cases because it is easy to regress :).

josevalim:
@fishcakez this should be good to go.

integration_test/mysql/test_helper.exs
```diff
@@ -1,5 +1,6 @@
 Logger.configure(level: :info)
-ExUnit.start exclude: [:array_type, :read_after_writes, :case_sensitive, :uses_usec]
+ExUnit.start exclude: [:array_type, :read_after_writes, :case_sensitive,
+                       :uses_usec, :strict_savepoint]
```
fishcakez: standard mysql...

lib/ecto/adapters/sql.ex
```diff
@@ -348,63 +345,43 @@ defmodule Ecto.Adapters.SQL do
...
         error
     end
   end
 
-  defp savepoint_sandbox(ref, log?, timeout, opts) do
-    case begin(ref, :raw, :sandbox, nil, log?, timeout, opts) do
-      {{:ok, _}, entry} ->
+  defp savepoint_sandbox(repo, ref, mod, entry, timeout, opts) do
+    case begin(repo, mod, :raw, :sandbox, nil, opts) do
+      # We ignore the entry, keeping only the begin transaction one
```
fishcakez: :worried:, why skip the first and not the second? The second makes it clear is a sandbox.

fishcakez: I mean why skip the second and not the first?

josevalim: I can reverse them, yeah.

josevalim:
@michalmuskala this should make things simpler on your side now!

michalmuskala:
@josevalim it definitely is. Updated with michalmuskala/mongodb_ecto@c5cb34b9fa010455e23e8d7d14840837ab626eed

PR #712, opened 2015-06-27
[WIP] Add support for sbroker

fishcakez:
Work in progress :).

Can we drop `stop/1`? It is awkward because it does not work in a supervision tree and as #706 shows we didn't test (or use?) it. I have removed it in a separate commit for now.

Todo:

- [x] Replace use of `Connection` with `GenServer` so not to add an extra dep, was useful to get started though.
- [x] Publish the version of `:sbroker` used and update dep/mix.lock

### DIFF
--- a/integration_test/cases/pool.exs
+++ b/integration_test/cases/pool.exs
@@ -16,9 +16,7 @@ defmodule Ecto.Integration.PoolTest do
 
   test "can start multiple repos" do
     # Can't start a second Repo with the default name
-    {_, pool_name, _} = PoolRepo.__pool__
-    pool_pid = Process.whereis(pool_name)
-    assert {:error, {:already_started, ^pool_pid}} = PoolRepo.start_link()
+    assert {:error, {:already_started, _}} = PoolRepo.start_link()
 
     # Can start a second repo with a different name
     assert {:ok, _second_pool} = PoolRepo.start_link(name: PoolRepo.Second)
--- a/integration_test/pool/pool_run.exs
+++ b/integration_test/pool/pool_run.exs
@@ -5,8 +5,15 @@ defmodule Ecto.Integration.PoolRunTest do
 
   @timeout :infinity
 
-  test "worker cleans up the connection when it crashes" do
-    {:ok, pool} = TestPool.start_link([lazy: false])
+  setup context do
+    case = context[:case]
+    test = context[:test]
+    {:ok, [pool: Module.concat(case, test)]}
+  end
+
+  test "worker cleans up the connection when it crashes", context do
+    pool = context[:pool]
+    {:ok, _} = TestPool.start_link([lazy: false, name: pool])
 
     assert {:ok, conn1} =
       TestPool.run(pool, @timeout, fn({_mod, conn1}, queue_time) ->
@@ -25,8 +32,9 @@ defmodule Ecto.Integration.PoolRunTest do
     end)
   end
 
-  test "nested run has no queue time" do
-    {:ok, pool} = TestPool.start_link([lazy: false])
+  test "nested run has no queue time", context do
+    pool = context[:pool]
+    {:ok, _} = TestPool.start_link([lazy: false, name: pool])
 
     TestPool.transaction(pool, @timeout, fn(_, _, _, _) ->
       TestPool.run(pool, @timeout, fn({_mod, _conn}, queue_time) ->
@@ -35,14 +43,16 @@ defmodule Ecto.Integration.PoolRunTest do
     end)
   end
 
-  test "disconnects if run raises" do
-    {:ok, pool} = TestPool.start_link([lazy: false])
+  test "disconnects if run raises", context do
+    pool = context[:pool]
+    {:ok, _} = TestPool.start_link([lazy: false, name: pool])
 
     assert {:ok, conn} =
       TestPool.run(pool, @timeout, fn({_mod, conn}, _) ->
         conn
       end)
 
+    monitor = Process.monitor(conn)
     assert Process.alive?(conn)
 
     try do
@@ -51,11 +61,12 @@ defmodule Ecto.Integration.PoolRunTest do
       RuntimeError -> :ok
     end
 
-    refute Process.alive?(conn)
+    assert_receive {:DOWN, ^monitor, _, _, _}
   end
 
-  test "do not disconnect if caller dies during run" do
-    {:ok, pool} = TestPool.start_link([lazy: false])
+  test "do not disconnect if caller dies during run", context do
+    pool = context[:pool]
+    {:ok, _} = TestPool.start_link([lazy: false, name: pool])
 
     _ = Process.flag(:trap_exit, true)
     parent = self()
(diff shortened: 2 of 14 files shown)


josevalim:
Yes, let's remove `stop/1`.

integration_test/pool/pool_transaction.exs
```diff
@@ -6,8 +6,15 @@ defmodule Ecto.Integration.PoolTransactionTest do
 
   @timeout :infinity
 
-  test "worker cleans up the connection when it crashes" do
-    {:ok, pool} = TestPool.start_link([lazy: false])
+  setup context do
+    case = context[:case]
+    test = context[:test]
+    {:ok, [pool: Module.concat(case, test)]}
```
josevalim: Do we need this for sbroker or something?

lib/ecto/adapters/sojourn_broker/worker.ex
```diff
@@ -0,0 +1,230 @@
...
+
+    s = %{conn: nil, module: module, params: params, transaction: nil,
+          broker: broker, tag: make_ref(), ref: nil, fun: nil, monitor: nil,
+          mode: :raw, backoff: min_backoff, min_backoff: min_backoff,
+          max_backoff: max_backoff, backoff_threshold: backoff_threshold}
+    {:connect, :init, s}
+  end
+
+  def connect(_, %{conn: nil, module: module, params: params,
+  min_backoff: min_backoff} = s) do
```
josevalim: You can have everything on the same line up to 80 chars. If it is too big, you can do the pattern matching inside the fun itself. :D

mix.exs
```diff
@@ -3,7 +3,7 @@ defmodule Ecto.Mixfile do
 
   @version "0.13.0-dev"
   @adapters [:pg, :mysql]
-  @pools [:poolboy]
+  @pools [:poolboy, :sojourn_broker]
```
josevalim: This means the integration tests will test each pool twice. Well, the unit tests too. I am not sure how to fix it, so you can let it be, I will think about some solutions.

mix.exs
```diff
@@ -34,6 +34,8 @@ defmodule Ecto.Mixfile do
 
   defp deps do
     [{:poolboy, "~> 1.4"},
+     {:sbroker, "~> 0.6", github: "fishcakez/sbroker", branch: "gen"},
```
josevalim: `sbroker` and `connection` should be optional. We will figure out how to make poolboy optional later (please add a TODO on top of the poolboy one).

lib/ecto/adapters/sojourn_broker/worker.ex
josevalim: Same here regarding pattern matching. :)

mix.exs
josevalim: Btw is the `connection` module meant to be used exclusively by sbroker?

integration_test/pool/pool_transaction.exs
fishcakez: The broker needs a name so the workers can message it.

mix.exs
fishcakez: It is for any connection process and nothing to do with `sbroker`. The mongo driver uses it for example https://github.com/ericmj/mongodb/blob/08b63a1aaca3bbf8f8f48daa90bc83630d46455f/lib/mongo/connection.ex#L10. @whatyouhide did a nice blog post about using it: http://andrealeopardi.com/posts/handling-tcp-connections-in-elixir/.

I am keen to remove from here because it would be an optional dep and the would be confusing. Also there could be a version mismatch with a driver.

josevalim: Ok, your call. If you think it would be better to use it with *everything* that is fine for me too.

josevalim: OR even push postgrex and mariaex and friends to use it too.

whatyouhide: Not part of this discussion, but FWIW I'm using it on my OrientDB adapter as well :)

fishcakez: > OR even push postgrex and mariaex and friends to use it too.

Before we do that I think we should see if the "smart" NoSQL connections perform better than the "dumb" SQL connections under network partition. What's good for `poolboy` might be bad for `sbroker` and vice versa.

josevalim: Oops, I meant the Connection module, not sbroker itself. :)

-- 


fishcakez:
I've rebased this and:
* Made `break/3` a cast for `sbroker` (blocking not required as worker checks in)
* Refactored the worker to a `GenServer` to remove the `Connection` dependency

integration_test/cases/pool.exs
```diff
@@ -16,9 +16,7 @@ defmodule Ecto.Integration.PoolTest do
 
   test "can start multiple repos" do
     # Can't start a second Repo with the default name
-    {_, pool_name, _} = PoolRepo.__pool__
-    pool_pid = Process.whereis(pool_name)
-    assert {:error, {:already_started, ^pool_pid}} = PoolRepo.start_link()
+    assert {:error, {:already_started, _}} = PoolRepo.start_link()
```
fishcakez: Must this point to the pool process? In this branch a supervisor (which is registered as `Module.concat(pool_name, Supervisor)`)

integration_test/pool/pool_run.exs
```diff
@@ -51,11 +61,12 @@ defmodule Ecto.Integration.PoolRunTest do
       RuntimeError -> :ok
     end
 
-    refute Process.alive?(conn)
+    assert_receive {:DOWN, ^monitor, _, _, _}
```
fishcakez: break is asynchronous

mix.exs
fishcakez: Have remove `Connection`. Still need to publish a new version of `sbroker`.

lib/ecto/adapters/sojourn_broker.ex
```diff
@@ -0,0 +1,124 @@
...
+
+    children = [broker, worker_sup]
+    sup_opts = [strategy: :rest_for_one, name: Module.concat(name, Supervisor)]
+    Supervisor.start_link(children, sup_opts)
+  end
+
+  @doc """
+  Stop the pool.
+  """
+  def stop(pool) do
```
fishcakez: this should go as `stop/1` is gone.

integration_test/cases/pool.exs
josevalim: No, it does not. :)

lib/ecto/adapters/sojourn_broker.ex
josevalim: So just remove it? :)

integration_test/cases/pool.exs
josevalim: As long as the pool name is the one coming from the `__pool__` tuple, we are good to go.

josevalim:
@fishcakez this looks great to me. Let's just release a new sbroker and get going? I want to clean up the pool structure and tests, but I will do that in another PR. :)

lib/ecto/adapters/sojourn_broker.ex
fishcakez: It was a reminder for myself :P.

fishcakez:
ugh, we need to make this lazy, otherwise we hit the same issue as the sandbox. I'll do this after lunch.

fishcakez:
I added lazy connect and tests. Implementation is clean if only the first connect is lazy - which is all we need.

PR #719, opened 2015-06-28
[WIP] Introduce SQL sandbox

fishcakez:
@josevalim thoughts?

### DIFF
--- a/lib/ecto/adapters/sql/sandbox.ex
+++ b/lib/ecto/adapters/sql/sandbox.ex
@@ -1,0 +1,320 @@
+defmodule Ecto.Adapters.SQL.Sandbox do
+  @moduledoc """
+  Start a pool with a single sandboxed SQL connection.
+  """
+
+  @behaviour Ecto.Adapters.Pool
+
+  @typep log :: (%Ecto.LogEntry{} -> any())
+
+  @doc """
+  Starts a pool with a single sandboxed connections for the given SQL connection
+  module and options.
+
+    * `conn_mod` - The connection module, see `Ecto.Adapters.Connection`
+    * `opts` - The options for the pool and the connections
+
+  """
+  @spec start_link(module, Keyword.t) :: {:ok, pid} | {:error, any}
+  def start_link(conn_mod, opts) do
+    name = Keyword.get(opts, :name)
+    GenServer.start_link(__MODULE__, {conn_mod, opts}, [name: name])
+  end
+
+  @doc false
+  @spec begin(pool, log, Keyword.t, timeout) ::
+    :ok | {:error, :sandbox} when pool: pid | atom
+  def begin(pool, log, opts, timeout) do
+    query(pool, :begin, log, opts, timeout)
+  end
+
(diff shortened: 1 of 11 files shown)


josevalim:
This is beautiful. Can we merge this? :D

lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -0,0 +1,320 @@
...
+    true
+  end
+
+  defp demonitor(%{monitor: mon} = s) do
+    Process.demonitor(mon, [:flush])
+    %{s | fun: nil, ref: nil, monitor: nil}
+  end
+
+  defp dequeue(%{queries: queries} = s) do
+    case :queue.out(queries) do
```
josevalim: Question: why do we need the queue when we didn't need one with a pool of size 1?

fishcakez: We did have a queue: `poolboy`. If an application is using the single connection, the setup in an `ExUnit.Case` gets queued on `poolboy` waiting for the application to hand the single connection back in.

Should probably merge the queues. Having two different queues will reorder requests which might cause race conditions.

josevalim: Would it make more sense then to still have it behind poolboy? We need to ship with a default queue... so we may as well use it?

lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -0,0 +1,320 @@
...
+      {_, {:error, _} = error} ->
+        error
+    end
+  end
+
+  defp do_checkout(pool, ref, fun, timeout) do
+    try do
+      GenServer.call(pool, {:checkout, ref, fun}, timeout)
+    catch
+      :exit, {:timeout, _} = reason ->
```
fishcakez: Should catch `{:noproc, _}` too.

fishcakez: oh never mind we do :)

lib/ecto/adapters/sql/sandbox.ex
fishcakez: Lets try behind poolboy - separate commit as discussed yesterday - before merging?

josevalim: Yes.

mix.lock
```diff
@@ -1,4 +1,5 @@
-%{"decimal": {:hex, :decimal, "1.1.0"},
+%{"dbg": {:hex, :dbg, "1.0.1"},
```
fishcakez: oops

josevalim:
@fishcakez I have merged only the first commit because the code we removed with queue handling from poolboy become pool management, so it is not really worth it.

If you prefer the poolboy version, let me know, and i will pick the other commit too. Thanks a lot!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

fishcakez:
Theres bug in break when :raw

PR #721, opened 2015-06-29
Break raw transaction in sandbox

fishcakez:


josevalim:
Do we need a test?

fishcakez:
> Do we need a test?

It would be a good idea :)

fishcakez:
Really we should re-add as many as possible of the removed sandbox'ed tests from `pool_run.exs` and `pool_transactions.exs`.

josevalim:
Yes, we can add them to the existing `integration_test/sql/test_transaction.exs`.

fishcakez:
Tests uncovered 2 more bugs

josevalim:
How severe are those bugs? Should we get v0.13.1 out?

fishcakez:
The first bug means that if a `Ecto.Adapters.Pool.run` fun raises an exception, the worker is checked in twice. Unfortunately `poolboy` doesn't use references to track messages and so treates the second checkin as a valid checkin. `Ecto.Adapters.SQL.Sandbox` uses references to track messages and crashed which led to the discovery of the bug.

I have not verified but I think the following race condition can occur:

* Run fun raises and fuse breaks
* Worker closes connection and starts a new one
* The first checkin occurs
* A client checks out the worker with `checkout` to do a run
* The second checkin occurs
* A clients checks out the worker with `open_transaction` to do a transaction
* Transaction begins and the query from `run/1` is run inside it.

If so it is severe :(.

The second bug is sandbox only where `break` does not break in raw mode. Ironically the first bug may just crash sandbox. However it can lead to unpredictable behaviour which I think is severe enough on its own.

PR #723, opened 2015-06-29
Add dependent opts

gjaldon:
Closes https://github.com/elixir-lang/ecto/issues/655

@josevalim ready for review. Looking forward to your feedback!


lib/ecto/model/callbacks.ex
```diff
@@ -70,7 +70,6 @@ defmodule Ecto.Model.Callbacks do
   defmacro __using__(_opts) do
     quote do
       import Ecto.Model.Callbacks
-      @before_compile Ecto.Model.Callbacks
```
josevalim: This is not good as it will break Ecto.Model.Callbacsks for everyone using it directly.

lib/ecto/model/dependent.ex
```diff
@@ -0,0 +1,86 @@
...
+  This also does not trigger any lifecycle callbacks on the dependent model.
+  """
+  def nilify_all(%Changeset{repo: repo} = changeset, assoc_queryable, assoc_key) do
+    query = assoc_query(changeset, assoc_queryable, assoc_key)
+    repo.update_all(query, set: [{assoc_key, nil}])
+    changeset
+  end
+
+  defp assoc_query(%Changeset{model: model}, assoc_queryable, assoc_key) do
+    from(a in assoc_queryable, where: field(a, ^assoc_key) == ^model.id)
```
josevalim: can't you use `Ecto.Model.assoc(model, assoc_name)`?

test/ecto/model/dependent_test.exs
```diff
@@ -0,0 +1,121 @@
+Code.require_file "../../support/mock_repo.exs", __DIR__
```
josevalim: Let's drop those unit tests, they are not doing anything really, in favor of integration tests. This way we can drop the changes to mock_repo too. :)

lib/ecto/model/dependent.ex
```diff
@@ -0,0 +1,86 @@
...
+  dependents.
+  """
+
+  alias Ecto.Changeset
+  import Ecto.Query, only: [from: 2]
+
+  @doc """
+  Deletes all records of dependent model and triggers `delete` lifecycle callbacks.
+  """
+  def fetch_and_delete(%Changeset{repo: repo} = changeset, assoc_queryable, assoc_key) do
```
josevalim: Very nice to break each function! Good implementation!

lib/ecto/model/dependent.ex
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  defp assoc_query(%Changeset{model: model}, assoc_queryable, assoc_key) do
+    from(a in assoc_queryable, where: field(a, ^assoc_key) == ^model.id)
+  end
+
+  defmacro __before_compile__(env) do
+    assocs = Module.get_attribute(env.module, :ecto_assocs) |> Enum.reverse
+
+    for {_assoc_name, assoc} <- assocs, Map.from_struct(assoc)[:dependent] do
```
josevalim: `Map.get(assoc, :dependent)`, no need to call from_struct.

test/ecto/model/dependent_test.exs
josevalim: We already have all the associations in there anyway, we just need to set different dependent options.

lib/ecto/association.ex
```diff
@@ -282,6 +284,11 @@ defmodule Ecto.Association.Has do
                            "option, the model should not be passed as second argument"
     end
 
+    if opts[:dependent] && not opts[:dependent] in @dependent_opts do
```
josevalim: In theory the default is `:nothing` (if we want to mirror with the `on_delete` from migrations.

josevalim:
Comments added!

lib/ecto/model/dependent.ex
gjaldon: Glad you like it! :bowtie:

gjaldon:
Thanks for the feedback! Will apply the changes tomorrow. :)

gjaldon:
@josevalim I made changes according to your feedback. The tests are much cleaner now. :)

lib/ecto/model.ex
```diff
@@ -43,7 +43,15 @@ defmodule Ecto.Model do
...
       use Ecto.Model.Timestamps
-      use Ecto.Model.Callbacks
+      import Ecto.Model.Callbacks
+      require Ecto.Model.Dependent
+
+      # The order of `@before_compile`s here is important since the model lifecycle
+      # callbacks introduced by `Ecto.Model.Dependent` will only be recognized by
+      # `Ecto.Model.Callbacks` if its run prior to `Ecto.Model.Callbacks`.
+      @before_compile Ecto.Model.Dependent
+      @before_compile Ecto.Model.Callbacks
```
josevalim: Do we still need to do this? Aren't we invoking it twice?

integration_test/support/models.exs
```diff
@@ -78,6 +86,7 @@ defmodule Ecto.Integration.Comment do
   end
 
   optimistic_lock :lock_version
+  before_delete Ecto.Integration.Model, :callback_check, []
```
josevalim: We can remove those.

integration_test/support/models.exs
```diff
@@ -94,6 +104,8 @@ defmodule Ecto.Integration.Permalink do
     belongs_to :post, Ecto.Integration.Post
     has_many :post_comments_authors, through: [:post, :comments_authors]
   end
+
+  before_delete Ecto.Integration.Model, :callback_check, []
```
josevalim: And this one.

josevalim: I mean, the integration test  gurantees the data is removed, so we don't need to check for a callback.

integration_test/support/models.exs
```diff
@@ -109,11 +122,29 @@ defmodule Ecto.Integration.User do
...
     field :name, :string
     has_many :comments, Ecto.Integration.Comment, foreign_key: :author_id
+    has_many :posts, Ecto.Integration.Post, dependent: :nothing, foreign_key: :author_id
+    has_many :addresses, Ecto.Integration.Address, dependent: :nilify_all
     belongs_to :custom, Ecto.Integration.Custom, references: :bid, type: :binary_id
     timestamps
   end
 end
 
+defmodule Ecto.Integration.Address do
```
josevalim: Could we get by without adding a new model? Or It seems we could add dependent to has_many :comments in the table above.

integration_test/support/models.exs
josevalim: Actually, ignore me. I see this is now testing if before_delete was run in the association. :+1:

josevalim:
I have added a couple more comments. We are almost there!

gjaldon:
@josevalim this should be good now. :)

gjaldon:
Thanks! :+1:

josevalim:
Thank you! Btw, i have renamed the `:dependent` option to `:on_delete`!

gjaldon:
You're welcome! `:on_delete` is clearer than `:dependent` :)

On Mon, Jul 6, 2015 at 7:59 PM, José Valim <notifications@github.com> wrote:

> Thank you! Btw, i have renamed the :dependent option to :on_delete!
>
> —
> <https://github.com/elixir-lang/ecto/pull/723#issuecomment-118833002>.
>



-- 
Gabe Jaldon

yordis:
@josevalim why `belongs_to` do not have something like `ActiveRecord` where I can setup `dependent: :destroy` so it will automatically remove the referenced record?

josevalim:
@yordis because Ecto prefers for the logic to exists at the database level since otherwise you can have data inconsistency errors. And at the database it is always from parent to child.

yordis:
@josevalim but what do you mean by that on this situation,

let say I have the situation where a `place` have some `address` so the `foreign key` is on the `place`.

So when I remove the `place` who is the one that knows about the `address` I want to remove the `address` as well.

Sorry I dont understand what you meant by that

josevalim:
@yordis if the foreign key is on "place", it means different places can point to the same address, which is probably undesirable in your case.

If you want the address to be deleted when the place is deleted, the place is the parent and the foreign key should be on address.

It seems your data design is wrong, no?

yordis:
@josevalim no really, don't forget one to one relationships which is the case for that flag (at least from my experience on Rails)

The reason I put the `address_id` on the place is because I want the place to be the one that knows about the address, no other wise. Just because the design of the application makes the address something that do not have value by their own, they need to be in some context, so place holds the responsibility of knowing the relationship.

That would be the reason and the use case, normally I see it on one to one relationships actually

josevalim:
@yordis we are exclusively talking about a one to one relationship, yes, and it has two sides.

I am not sure how else I can convince you the data design is inadequate, given it allows inconsistent relationships (two different places in the same address) and given the fact the kind of clean up you are advocating is impossible to do at the database level using just constraints.

I also find your definition of responsibility a bit skewed, since you are literally making "address" the parent of a "place". If you make the "place" the parent, Ecto does provide the conveniences for managing the address as a child and automatically handling of the primary key. This way you get the semantics you want at the application level and the proper design at the data level. If you do not value data integrity, then I am afraid Ecto is not going to suit you.

josevalim:
And I want to clarify that, if you actually want the address to be a generic entity in your system that is not necessarily related to places, then it does not make sense to delete an address when a place is deleted, since other entities in your system can be using the same address.

yordis:
>  given the fact the kind of clean up you are advocating is impossible to do at the database level using just constraints.

You are right, that is why ActiveRecord use the callbacks for do it for yourself. It's conveniency  at the end of the day. I don't have to wrap everything in a transaction, remove the other side and then remove this side.

I am never expect any database level constrains because there is non.

> And I want to clarify that, if you actually want the address to be a generic entity in your system that is not necessarily related to places, then it does not make sense to delete an address when a place is deleted, since other entities in your system can be using the same address.

That is the issue, that it is not the this case. I just want to normalize the database and extract the address information to it own table and keep the place referring to it. What is the design flow with it? The business rules says that I dont want to allow to have more than one relationship with the addres, so it's one to one again.

Even when you can make arguing about this particula use case and the way I am handling. I am trying to focus more in the use case of One to One, where you want to owner of the foreign key to manage the other side.

Sorry but I don't see anything wrong, and Ecto could help with conveniences around it.

josevalim:
>You are right, that is why ActiveRecord use the callbacks for do it for
yourself. It's conveniency at the end of the day. I don't have to wrap
everything in a transaction, remove the other side and then remove this
side.

Wrapping in a transaction is not enough to guarantee data integrity!!! This
is a very common misconception because callbacks are not atomic.

First of all, since your data model allows multiple places to have the same
address, deleting an address can cause a place to go without an address.

Second of all, concurrent updates+deletes can leave dangling addresses.

Imagine you have place with ID 1 and address with ID 1 loaded in memory for
deletion. Before deletion is triggered, another request updates the address
to ID 2. Now you do the deletion on the old IDs and ID 2 won't be deleted.

You can say that you are being careful to not violate those cases but then
any further change is walking on a mine field, especially when you could
have all of this guaranteed at the data layer.

> Even when you can make arguing about this particula use case and the way
I am handling. I am trying to focus more in the use case of One to One,
where you want to owner of the foreign key to manage the other side.

No. In any relationship, the foreign key goes to the children, not the
owner. It doesn't change if it is a many to one or one to one. For a place
with many reviews, it goes on reviews. For a place with one address, it
goes on address.

It is not about Ecto missing conveniences, it is Ecto not allowing you to
shoot yourself in the foot. As said previously, if you don't value data
integrity, then Ecto is likely not going to suit you.

-- 


yordis:
> First of all, since your data model allows multiple places to have the same
address, deleting an address can cause a place to go without an address.

That is not the case `One to One` != `One to many`

> For a place with one address, it goes on address.

But why have to be like that? I decided to let `Place` to handle the Address. Address becomes unaware of the `Place` so far but my application rule knows about how to handle the situation.
In an `One to One` relationship either side could have the foreign key and I decided to add it to the `Place` for design reasons. I am not following your arguments.

> it is Ecto not allowing you to shoot yourself in the foot.

Probably, at the end of the day I am expecting Ecto todo exactly the same code that I would actually do. `Wrap in an Transation` -> `Remove the relationships with depends: destroy` -> `Remove the main record` -> `Return the result` 

> Imagine you have place with ID 1 and address with ID 1 loaded in memory for
deletion. Before deletion is triggered, another request updates the address
to ID 2. Now you do the deletion on the old IDs and ID 2 won't be deleted.

Wouldn't the reason why we use transactions? And still this case wouldn't happen, you can't change the Address ID because the business specific rule do not allow you to do it, either you remove it and add it back again or you will always update the same record.

PR #725, opened 2015-06-30
[WIP] Support multiple pools

### LINKED ISSUE #569: Support reserving a worker

MSch:
With these changes I can run multiple `poolboy` instances. I'll try overriding `__pool__` in our app next and see if it works as nicely as we hope.

Fixes #569.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -58,7 +58,10 @@ defmodule Ecto.Repo do
 
   @doc false
   defmacro __using__(opts) do
-    quote bind_quoted: [opts: opts] do
+    {_, adapter, config} = Ecto.Repo.Config.parse(__CALLER__.module, opts)
+    extra_repo_funs = adapter.extra_repo_funs(config)
+
+    quote bind_quoted: [opts: opts, extra_repo_funs: extra_repo_funs] do
       @behaviour Ecto.Repo
 
       {otp_app, adapter, config} = Ecto.Repo.Config.parse(__MODULE__, opts)
@@ -170,6 +173,8 @@ defmodule Ecto.Repo do
       end
 
       defoverridable [log: 1]
+
+      extra_repo_funs
     end
   end
 
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -77,8 +77,9 @@ defmodule Ecto.Repo do
         Ecto.Repo.Config.config(@otp_app, __MODULE__)
       end
 
-      def start_link do
-        @adapter.start_link(__MODULE__, config())
+      def start_link(custom_config \\ []) do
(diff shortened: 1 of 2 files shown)


lib/ecto/repo.ex
```diff
@@ -58,7 +58,10 @@ defmodule Ecto.Repo do
 
   @doc false
   defmacro __using__(opts) do
-    quote bind_quoted: [opts: opts] do
+    {_, adapter, config} = Ecto.Repo.Config.parse(__CALLER__.module, opts)
+    extra_repo_funs = adapter.extra_repo_funs(config)
```
josevalim: This is unfortunately extending the adapter behaviour. :(

lib/ecto/repo.ex
```diff
@@ -74,8 +77,9 @@ defmodule Ecto.Repo do
         Ecto.Repo.Config.config(@otp_app, __MODULE__)
       end
 
-      def start_link do
-        @adapter.start_link(__MODULE__, config())
+      def start_link(custom_config \\ []) do
```
josevalim: I have pushed a different implementation to master to allow pool customization. We now just need the `start_link/1` changes plus guarantee the name and pool given to `start_link/1` have higher priority than the one in `__pool__`. For that, we should have tests.

PR #728, opened 2015-06-30
Support multiple pools

### LINKED ISSUE #569: Support reserving a worker

MSch:
Continuation of #725.

Fixes #569.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -352,10 +352,12 @@ defmodule Ecto.Adapters.SQL do
       """
     end
 
-    {pool_mod, pool, _} = repo.__pool__
+    {default_pool_mod, default_pool_name, _} = repo.__pool__
+    pool_mod = Keyword.get(opts, :pool, default_pool_mod)
+
     opts = opts
       |> Keyword.put(:timeout, Keyword.get(opts, :connect_timeout, 5000))
-      |> Keyword.put(:name, pool)
+      |> Keyword.put_new(:name, default_pool_name)
       |> Keyword.put_new(:size, 10)
 
     pool_mod.start_link(connection, opts)
--- a/integration_test/cases/pool.exs
+++ b/integration_test/cases/pool.exs
@@ -1,0 +1,30 @@
+Code.require_file "../support/file_helpers.exs", __DIR__
+
+defmodule Ecto.Integration.PoolTest do
+  use Ecto.Integration.Case
+
+  require Ecto.Integration.PoolRepo, as: PoolRepo
+
+  defmodule MockPool do
+    def start_link(_conn_mod, opts) do
+      # Custom options are passed through
+      assert PoolRepo.Third == opts[:name]
+      assert :bar == opts[:foo]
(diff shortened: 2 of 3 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -352,10 +352,13 @@ defmodule Ecto.Adapters.SQL do
...
 
-    {pool_mod, pool, _} = repo.__pool__
+    {default_pool_mod, default_pool_name, _} = repo.__pool__
+    pool_mod = Keyword.get(opts, :pool, default_pool_mod)
+    pool_name = Keyword.get(opts, :name, default_pool_name)
+
     opts = opts
       |> Keyword.put(:timeout, Keyword.get(opts, :connect_timeout, 5000))
-      |> Keyword.put(:name, pool)
+      |> Keyword.put(:name, pool_name)
```
josevalim: `Keyword.put_new(:name, default_pool_name)`

integration_test/sql/multiple_pools.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Ecto.Integration.MultiplePoolsTest do
```
josevalim: Let's move this to `integration_test/cases/pool.exs` because I would like to enforce this on all adapters!

josevalim:
Excellent. I have added just two minor comments!

lib/ecto/adapters/sql.ex
MSch: Can do the same for `:pool` right?

MSch:
@josevalim force pushed.

josevalim:

PR #736, opened 2015-07-02
Allow Date/DateTime to cast 1digit month/day strings

orendon:
Minor improvement on `cast` to handle 1 digit month/day strings.
Since is pretty common to have strings like `"2015-8-5"`


josevalim:
Thank you. I am :-1: because this isn't really meant to parse all
inputs.you will need to normalize the date on client before sending it.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

orendon:
@josevalim ahh ok, my understanding was it was handling `Map` and `String` cases. Thanks for the quick reply :smile_cat:

PR #740, opened 2015-07-03
Change migrations to accommodate Mongo

michalmuskala:
All changes should be fully backwards compatible.

### DIFF
--- a/integration_test/cases/migrator.exs
+++ b/integration_test/cases/migrator.exs
@@ -8,6 +8,12 @@ defmodule Ecto.Integration.MigratorTest do
 
   alias Ecto.Integration.Barebone
   alias Ecto.Integration.TestRepo
+  alias Ecto.Migration.SchemaMigration
+
+  setup do
+    TestRepo.delete_all(SchemaMigration)
+    :ok
+  end
 
   defmodule GoodMigration do
     use Ecto.Migration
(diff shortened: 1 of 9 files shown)


integration_test/cases/migrator.exs
```diff
@@ -81,6 +86,9 @@ defmodule Ecto.Integration.MigratorTest do
 
       assert count_entries() == 1
       assert [50] = run(TestRepo, path, :up, to: 50, log: false)
+
+      purge migrations
```
josevalim: Cool, why do we need those? Maybe we should just clean up the migration table before the tests? It would also solve the earlier [0] issue, wouldn't it?

michalmuskala: The `[0]` is there because of https://github.com/elixir-lang/ecto/blob/master/integration_test/pg/test_helper.exs#L59

We could clean up the table here explicitly in a `setup`, and ignore the migration from `test_helper.exs`.

The `purge` calls are needed because without them the compiler complains about redefining modules.

josevalim: Something is odd because we don't get those warnings in Postgres. We are
very careful to only use new names (hence the numbers). Thoughts?



-- 


michalmuskala: If you remove the `purge` but keep the `run(TestRepo, path, :down, all: true, log: false)` that I added to clean up after each test you'll get warnings in both postgres and mysql.

michalmuskala: I think it's because migrator will `Code.load_file`: https://github.com/elixir-lang/ecto/blob/master/lib/ecto/migrator.ex#L213

josevalim: Ah, yes, you need to purge if you run them again. Good point. The point though is to remove those changes altogether and just make sure we clean up the table before running the tests. Can you update this PR?

michalmuskala:
I updated everything as we talked about it:
* `create table(...)` can be executed without a block
* table options can be keywords
* `execute/1` accepts keywords
* SQL databases raise on keywords

PR #742, opened 2015-07-03
Do not execute update callbacks for empty changesets by default.

antipax:
As per the discussion on #733, here is another attempt at #731.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -425,6 +425,10 @@ defmodule Ecto.Repo do
 
   ## Options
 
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks @antipax!

Actually, we don't need to execute any of this at all if there are no changes:

https://github.com/antipax/ecto/blob/no-update-callbacks-for-empty-changesets/lib/ecto/repo/model.ex#L80-L109

So we can just wrap the whole block and rename `force_callbacks` to `force`. :)

antipax:
Cool, done. Thanks for feedback @josevalim

PR #743, opened 2015-07-03
Flatten transactions

josevalim:
/cc @MSch @fishcakez 

This is almost ready, I need to refactor the code and clean some tests that rely on depth, but it is almost all there. Feedback is welcome.

The only thing up to discussion is this test:

https://github.com/elixir-lang/ecto/blob/master/integration_test/sql/transaction.exs#L71

Because the inner transaction fails and we rescue it, we need to make sure all operations inside the parent transaction still fails. Today we fail with `:noconnect` but that is a lie because the connection is there, we have just marked it as tainted. We could raise a specific error though. If so, which?

@fishcakez I would merge your PR before merging this so you don't have to rebase it one more time. Although I think little should be affected.

### DIFF
--- a/integration_test/sql/transaction.exs
+++ b/integration_test/sql/transaction.exs
@@ -68,8 +68,22 @@ defmodule Ecto.Integration.TransactionTest do
     assert [] = TestRepo.all(Trans)
   end
 
+  test "transaction rolls back per repository" do
+    message = "cannot call rollback outside of transaction"
+
+    assert_raise RuntimeError, message, fn ->
+      PoolRepo.rollback(:done)
+    end
+
+    assert_raise RuntimeError, message, fn ->
+      TestRepo.transaction fn ->
+        PoolRepo.rollback(:done)
+      end
+    end
+  end
+
   test "nested transaction partial rollback" do
-    PoolRepo.transaction(fn ->
+    assert PoolRepo.transaction(fn ->
       e1 = PoolRepo.insert!(%Trans{text: "3"})
       assert [^e1] = PoolRepo.all(Trans)
 
@@ -83,20 +97,30 @@ defmodule Ecto.Integration.TransactionTest do
         UniqueError -> :ok
       end
 
-      e3 = PoolRepo.insert!(%Trans{text: "5"})
-      assert [^e1, ^e3] = PoolRepo.all(from(t in Trans, order_by: t.text))
-      assert [] = TestRepo.all(Trans)
-    end)
+      assert {:noconnect, _} = catch_exit(PoolRepo.insert!(%Trans{text: "5"}))
+    end) == {:error, :rollback}
 
-    assert [%Trans{text: "3"}, %Trans{text: "5"}] =
-           TestRepo.all(from(t in Trans, order_by: t.text))
+    assert TestRepo.all(Trans) == []
   end
 
-  test "manual rollback doesnt bubble up" do
+  test "manual rollback doesn't bubble up" do
     x = PoolRepo.transaction(fn ->
       e = PoolRepo.insert!(%Trans{text: "6"})
       assert [^e] = PoolRepo.all(Trans)
       PoolRepo.rollback(:oops)
+    end)
+
+    assert x == {:error, :oops}
+    assert [] = TestRepo.all(Trans)
+  end
+
+  test "manual rollback bubbles up on nested transaction" do
+    x = PoolRepo.transaction(fn ->
+      e = PoolRepo.insert!(%Trans{text: "6"})
+      assert [^e] = PoolRepo.all(Trans)
+      PoolRepo.transaction(fn ->
+        PoolRepo.rollback(:oops)
+      end)
     end)
 
     assert x == {:error, :oops}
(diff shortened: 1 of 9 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -458,43 +449,85 @@ defmodule Ecto.Adapters.SQL do
...
+      {:ok, ref, {mod, _conn}, queue_time} ->
+        try do
+          mode = transaction_mode(pool_mod, pool, timeout)
+          transaction(repo, ref, mod, mode, queue_time, timeout, opts, fun)
+        catch
+          # If any error leaked, it was inside Ecto.
+          # Make sure we break the connection.
+          kind, reason ->
+            stack = System.stacktrace()
+            Pool.break(ref)
```
fishcakez: `break/2` does a blocking disconnect with timeout and a checkin. `close_transaction/2` should come in an `else` clause rather than `after`. Is this misleading? The current behaviour is used so that checkin can occur as early as possible - the connection is useless once broken and we want to release the lock on the worker as soon as possible.

In #712 `break/2` is asynchronous because the worker checks itself in so the back pressure is not required. It is also more efficient (in that case) to send it as a single message.

fishcakez:
> @fishcakez I would merge your PR before merging this so you don't have to rebase it one more time. Although I think little should be affected.

We can merge which ever is ready first. There are no conflicts yet.

lib/ecto/adapters/sql.ex
```diff
@@ -458,43 +449,85 @@ defmodule Ecto.Adapters.SQL do
...
+        after
+          Pool.close_transaction(ref, timeout)
+        end
+
+      {:already_open, ref, _} ->
+        try do
+          {{:return, {:ok, fun.()}}, nil}
+        catch
+          # Taint the connection on any error.
+          kind, reason ->
```
fishcakez: We need to be explicit in docs that this doesn't catch `:ecto_rollback` throw here, taint and return error.

fishcakez:
I think we should make rollback part of `Pool` and not export `taint`. Currently rollback throws do not contain any information relating to the repo or pool.

integration_test/sql/transaction.exs
```diff
@@ -86,13 +86,12 @@ defmodule Ecto.Integration.TransactionTest do
       e3 = PoolRepo.insert!(%Trans{text: "5"})
       assert [^e1, ^e3] = PoolRepo.all(from(t in Trans, order_by: t.text))
       assert [] = TestRepo.all(Trans)
-    end)
+    end) == {:error, :rolledback}
```
MSch: Shouldn't this be present tense `{:error, :rollback}`?

lib/ecto/adapters/sql.ex
josevalim: It doesn't matter because close_transaction is no-op if it was broken. But I guess else is better semantically, what do you think?

lib/ecto/adapters/sql.ex
josevalim: Yup! Docs are pending!

josevalim:
> I think we should make rollback part of Pool and not export taint. Currently rollback throws do not contain any information relating to the repo or pool.

Yes, It was hard to see this at the beginning but I think it is clearer now... do you think we should have `Pool.transaction` that receives `begin_fun`, `fun` and `finalize_fun`? Finalize will rollback or commit. Thoughts?

lib/ecto/adapters/sql.ex
fishcakez: Ah! I prefer `else`. I am not keen on `maybe_do_X` functions though so very biased.

fishcakez:
> Yes, It was hard to see this at the beginning but I think it is clearer now... do you think we should have Pool.transaction that receives begin_fun, fun and finalize_fun? Finalize will rollback or commit. Thoughts?

Sounds good.

josevalim:
This is ready to go. I will merge it tomorrow morning and then get @fishcakez new pool in.

lib/ecto/adapters/pool.ex
```diff
@@ -152,198 +145,217 @@ defmodule Ecto.Adapters.Pool do
...
-      {:ok, fun.(ref, conn, depth, time)}
-    after
-      case Process.put(ref, info) do
-        %{conn: _} ->
-          :ok
-        %{} ->
-          _ = Process.put(ref, Map.delete(info, :conn))
-          :ok
+      value = fun.()
+      if untaint(ref) do
```
fishcakez: is this the correct call here?

lib/ecto/adapters/pool.ex
```diff
@@ -152,198 +145,217 @@ defmodule Ecto.Adapters.Pool do
...
+    :ok
   end
 
-  defp do_transaction(ref, %{depth: depth, conn: conn} = info, time, fun) do
-    depth = depth + 1
-    _ = Process.put(ref, %{info | depth: depth})
+  @doc """
+  Executes the given function giving it the ability to rollback.
+
+  Retunrs `{:ok, value}` if no transaction ocurred,
```
fishcakez: s/Retunrs/Returns

lib/ecto/adapters/pool.ex
josevalim: Yes. We need to untainted always otherwise rollback/commit won't work because the connection would still be tainted.

josevalim: Actually, we could call it on `after`. It would be better. I will clean it up tomorrow.

PR #745, opened 2015-07-04
Fix validate_unique

jonhkr:
I fixed an error in the `validate_unique` validation.
It wasn't checking the model's value. If there were no changes in the field and we had a record in the database with the value `null` for this field, it would result in a validation error, even though the model had the non unique value.

If there were no changes in the field, it would always validate like `is_nil(field(m, ^field))`


josevalim:
@jonhkr good catch. Can you add a test? Or would you prefer if we did it on our side?

jonhkr:
Take a look at 184804d, I think that does the trick. I had to add a new process key to alternate the query assertion.

PR #746, opened 2015-07-04
Upsert

gjaldon:
Closes https://github.com/elixir-lang/ecto/issues/652

@josevalim @ericmj I installed PG 9.5-alpha and the integration test I added for upsert will fail for previous versions of PG. How to deal with that?

### DIFF
--- a/integration_test/cases/upserts.exs
+++ b/integration_test/cases/upserts.exs
@@ -31,12 +31,21 @@ defmodule Ecto.Integration.Upserts do
     end
   end
 
-  test "insert if record exists on update" do
+  test "insert if record does not exist on update" do
+    p1 = TestRepo.update!(%Post{id: 1, title: "1", visits: 1}, if_not_exists: :insert)
+    p1 = TestRepo.get(Post, p1.id)
+    assert p1
   end
 
-  test "ignore if record exists on update" do
+  test "ignore if record does not exist on update" do
+    p1 = TestRepo.update!(%Post{id: 1, title: "1", visits: 1}, if_not_exists: :ignore)
+    p1 = TestRepo.get(Post, p1.id)
+    refute p1
   end
 
-  test "error if record exists on update" do
+  test "error if record does not exist on update" do
+    assert_raise Ecto.StaleModelError, fn ->
+      TestRepo.update!(%Post{id: 1, title: "1", visits: 1}, if_not_exists: :error)
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


josevalim:
@ericmj Can we easily the the PG version in the test helper?

josevalim:
I am holding this until PostgreSQL is out. :) Thank you @gjaldon. :heart:

gjaldon:
Ok sounds good :)

iraasta:
What is a state of this issue? 
Is there any way to achieve upserting in some other, harder way in this moment?

JHKennedy4:
@iraasta trying to solve the same problem at the moment (and debating installing the alpha since this is a side project).

Here's the solution the [Postgres Wiki](https://wiki.postgresql.org/wiki/UPSERT#UPSERT_as_implemented_in_practice) links to on [StackOverflow](http://stackoverflow.com/questions/17267417/how-do-i-do-an-upsert-merge-insert-on-duplicate-update-in-postgresql)

iraasta:
Yeah the problem is the DB I'm forced to use is MySQL right now

JHKennedy4:
Ah, sorry, thought the issue was specific to Postgres. As a short term fix,
could you add gjaldon's changes to the MySQL adapter in your own codebase?
It's only a ten line change. I'm no sure it'll work, but it's a possible
workaround until this gets merged in/Postgres updates

ericmj:
I am dropping a comment here so we don't forget it for the future:

The option should only be available on `insert`, there is no need to provide it for `update` and change the `UPDATE` to an `INSERT` to support it. It may not compose well with future options.

I think we should change the option to: `on_conflict: :update | :error | :ignore`

stevegraham:
Looking forward to this! 😊

stevegraham:
Looking at the code for PG `conflict_target` is specified as id, does this mean that PG won't perform unique index inference with all useable indices as it does when omitting `conflict_target` for the `ON CONFLICT DO NOTHING` strategy?

michalmuskala:
Postgres 9.5 is out http://www.postgresql.org/about/news/1636/

We can think about adding support for upsert.

gjaldon:
Ok, great! I will work on updating this PR over the next few days.

josevalim:
@gjaldon I think we should go back to the drawing board on this one since postgres supports constraints and it would be nice to support them too. We should probably do a proposal comparing different databases, check how it affects insert_all and so on. :)

gjaldon:
@josevalim ok will hold off on working on this until we have a clear plan on how to proceed. Until then, I'll study this and come up with a proposal of my own. If others have their own proposals, feel free to pitch in.

josevalim:
Sounds great! :+1:

josevalim:
Closing this. This feature is still welcome, but we shuold likely start the discussion from scratch.

PR #748, opened 2015-07-06
Embedding

michalmuskala:
This is a first dirty attempt to implement embedded models. Currently only implements `embeds_one`.

There is a lot to improve here, but I wanted to get something out to see, if I'm going in the right direction.

@josevalim thoughts?

### DIFF
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -10,8 +10,8 @@ defmodule Ecto.Integration.TypeTest do
   alias Ecto.Integration.Post
   alias Ecto.Integration.Tag
   alias Ecto.Integration.Custom
-  alias Ecto.Integration.User
-  alias Ecto.Integration.Profile
+  alias Ecto.Integration.Order
+  alias Ecto.Integration.Item
 
   test "primitive types" do
     integer  = 1
@@ -152,14 +152,28 @@ defmodule Ecto.Integration.TypeTest do
 
   @tag :embedded
   test "embeds one" do
-    profile = %Profile{location: "City"}
-    user = TestRepo.insert!(%User{profile: profile})
-    assert TestRepo.get!(User, user.id).profile.location == profile.location
+    order = TestRepo.insert!(%Order{item: %Item{price: 123}})
+    assert %Item{price: 123} = TestRepo.get!(Order, order.id).item
+    assert [%Item{price: 123}] =
+      TestRepo.all(from o in Order, select: o.item)
 
-    user = %User{}
-    user = put_in(user.profile, build_embedded(user, :profile, location: "New"))
-    user = TestRepo.insert!(user)
-    assert TestRepo.get!(User, user.id).profile.location == "New"
+    order = %Order{}
+    order = put_in(order.item, build_embedded(order, :item, price: 123))
+    order = TestRepo.insert!(order)
+    assert %Item{price: 123} = TestRepo.get!(Order, order.id).item
+  end
+
+  @tag :embedded
+  test "embeds many" do
+    order = TestRepo.insert!(%Order{items: [%Item{price: 123}]})
+    assert [%Item{price: 123}] = TestRepo.get!(Order, order.id).items
+    assert [[%Item{price: 123}]] =
+      TestRepo.all(from a in Order, select: a.items)
+
+    order = %Order{}
+    order = put_in(order.items, [build_embedded(order, :items, price: 123)])
+    order = TestRepo.insert!(order)
+    assert [%Item{price: 123}] = TestRepo.get!(Order, order.id).items
   end
 
   @tag :decimal_type
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -150,7 +150,7 @@ defmodule Ecto.Integration.TypeTest do
     assert TestRepo.get!(Post, post.id).meta == %{"world" => "hello"}
   end
 
-  @tag :embedded
+  @tag :map_type
   test "embeds one" do
     order = TestRepo.insert!(%Order{item: %Item{price: 123}})
     assert %Item{price: 123} = TestRepo.get!(Order, order.id).item
@@ -163,17 +163,18 @@ defmodule Ecto.Integration.TypeTest do
     assert %Item{price: 123} = TestRepo.get!(Order, order.id).item
   end
 
-  @tag :embedded
-  test "embeds many" do
-    order = TestRepo.insert!(%Order{items: [%Item{price: 123}]})
-    assert [%Item{price: 123}] = TestRepo.get!(Order, order.id).items
+  @tag :map_type
+  @tag :array_type
+  test "embeds many with array" do
+    tag = TestRepo.insert!(%Tag{items: [%Item{price: 123}]})
+    assert [%Item{price: 123}] = TestRepo.get!(Tag, tag.id).items
     assert [[%Item{price: 123}]] =
-      TestRepo.all(from a in Order, select: a.items)
+      TestRepo.all(from t in Tag, select: t.items)
 
-    order = %Order{}
-    order = put_in(order.items, [build_embedded(order, :items, price: 123)])
-    order = TestRepo.insert!(order)
-    assert [%Item{price: 123}] = TestRepo.get!(Order, order.id).items
+    tag = %Tag{}
+    tag = put_in(tag.items, [build_embedded(tag, :items, price: 123)])
+    tag = TestRepo.insert!(tag)
+    assert [%Item{price: 123}] = TestRepo.get!(Tag, tag.id).items
   end
 
   @tag :decimal_type
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -5,10 +5,13 @@ defmodule Ecto.Integration.TypeTest do
 
   require Ecto.Integration.TestRepo, as: TestRepo
   import Ecto.Query
+  import Ecto.Model
 
   alias Ecto.Integration.Post
   alias Ecto.Integration.Tag
   alias Ecto.Integration.Custom
+  alias Ecto.Integration.User
+  alias Ecto.Integration.Profile
 
   test "primitive types" do
     integer  = 1
@@ -145,6 +148,18 @@ defmodule Ecto.Integration.TypeTest do
     query = from(p in Post, where: p.id == ^post.id)
     TestRepo.update_all(query, set: [meta: %{world: "hello"}])
     assert TestRepo.get!(Post, post.id).meta == %{"world" => "hello"}
+  end
+
+  @tag :embedded
+  test "embeds one" do
+    profile = %Profile{location: "City"}
+    user = TestRepo.insert!(%User{profile: profile})
+    assert TestRepo.get!(User, user.id).profile.location == profile.location
+
+    user = %User{}
+    user = put_in(user.profile, build_embedded(user, :profile, location: "New"))
+    user = TestRepo.insert!(user)
+    assert TestRepo.get!(User, user.id).profile.location == "New"
   end
 
   @tag :decimal_type
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -5,7 +5,6 @@ defmodule Ecto.Integration.TypeTest do
 
   require Ecto.Integration.TestRepo, as: TestRepo
   import Ecto.Query
-  import Ecto.Model
 
   alias Ecto.Integration.Post
   alias Ecto.Integration.Tag
@@ -156,11 +155,6 @@ defmodule Ecto.Integration.TypeTest do
     assert %Item{price: 123} = TestRepo.get!(Order, order.id).item
     assert [%Item{price: 123}] =
       TestRepo.all(from o in Order, select: o.item)
-
-    order = %Order{}
-    order = put_in(order.item, build_embedded(order, :item, price: 123))
-    order = TestRepo.insert!(order)
-    assert %Item{price: 123} = TestRepo.get!(Order, order.id).item
   end
 
   @tag :map_type
@@ -170,11 +164,6 @@ defmodule Ecto.Integration.TypeTest do
     assert [%Item{price: 123}] = TestRepo.get!(Tag, tag.id).items
     assert [[%Item{price: 123}]] =
       TestRepo.all(from t in Tag, select: t.items)
-
-    tag = %Tag{}
-    tag = put_in(tag.items, [build_embedded(tag, :items, price: 123)])
-    tag = TestRepo.insert!(tag)
-    assert [%Item{price: 123}] = TestRepo.get!(Tag, tag.id).items
   end
 
   @tag :decimal_type
--- a/integration_test/support/migration.exs
+++ b/integration_test/support/migration.exs
@@ -58,15 +58,14 @@ defmodule Ecto.Integration.Migration do
     end
 
     create table(:orders) do
-      add :item, :map
-      add :items, :map
+      add :item, :embed
     end
 
     unless :array_type in ExUnit.configuration[:exclude] do
       create table(:tags) do
         add :ints, {:array, :integer}
         add :uuids, {:array, :uuid}
-        add :items, {:array, :map}
+        add :items, {:array, :embed}
       end
     end
   end
--- a/integration_test/support/migration.exs
+++ b/integration_test/support/migration.exs
@@ -5,7 +5,6 @@ defmodule Ecto.Integration.Migration do
     create table(:users) do
       add :name, :text
       add :custom_id, :uuid
-      add :profile, :map
       timestamps
     end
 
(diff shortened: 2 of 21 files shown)


integration_test/support/models.exs
```diff
@@ -173,3 +175,17 @@ defmodule Ecto.Integration.Tag do
...
+defmodule Ecto.Integration.Profile do
+  @moduledoc """
+  This module is used to array
+
+    * Embedding
+
+  """
+  use Ecto.Integration.Model
+
+  schema "" do
```
michalmuskala: We need to deal with this somehow. I propose to have a separate macro called `embedded_schema` to handle this.

lib/ecto/schema.ex
```diff
@@ -754,6 +780,12 @@ defmodule Ecto.Schema do
     Ecto.Model.Callbacks.__apply__(struct.__struct__, :after_load, loaded)
   end
 
+  def __load__(struct, source, fields, map, id_types) do
```
michalmuskala: This will also help with mongo driver, where we already have a map. Is there some more efficient way to handle this?

integration_test/cases/type.exs
```diff
@@ -147,6 +150,18 @@ defmodule Ecto.Integration.TypeTest do
...
   end
 
+  @tag :embedded
+  test "embeds one" do
+    profile = %Profile{location: "City"}
+    user = TestRepo.insert!(%User{profile: profile})
+    assert TestRepo.get!(User, user.id).profile.location == profile.location
+
+    user = %User{}
+    user = put_in(user.profile, build_embedded(user, :profile, location: "New"))
```
michalmuskala: This is rather unfriendly. Is there a better way to handle this?

lib/ecto/type.ex
```diff
@@ -295,6 +295,10 @@ defmodule Ecto.Type do
     dump_array(type, value, [], false)
   end
 
+  def dump({:map, model}, %{__struct__: model} = struct) do
+    {:ok, struct |> Map.from_struct |> Map.delete(:__meta__)}
```
josevalim: This is going to include virtual fields too, we need to not do that.

josevalim: You will need to add a clause to the `of_base_type?` function in this module. Check the current map one for examples. Don't forget to add unit tests for the changes in this module too. Use the `:map` type commit for reference if you want to: 6f6b300273d352a326c040f66b7c849ace7e5474

integration_test/cases/type.exs
josevalim: Yeah, we should consider maybe make `build_embedded/3` already embed it. But let's discuss such APIs later.

integration_test/support/models.exs
```diff
@@ -115,6 +115,7 @@ defmodule Ecto.Integration.User do
     * Timestamps
     * Relationships
     * Dependent callbacks
+    * Embedding
```
josevalim: To make things simpler, what if we flip things around? Let's create an Account model that embeds User instead of User embedding Profile. There are already too many things happening in the User module.

integration_test/support/models.exs
josevalim: Nah, it is fine. :) Just give whatever name and we will document it is simply not used when embedded.

lib/ecto/type.ex
```diff
@@ -367,6 +371,14 @@ defmodule Ecto.Type do
     {:ok, Application.get_env(:ecto, :json_library).decode!(value)}
   end
 
+  def load({:map, model}, value) when is_binary(value) do
+    load({:map, model}, Application.get_env(:ecto, :json_library).decode!(value))
+  end
+
+  def load({:map, model}, map) when is_map(map) do
+    {:ok, model.__schema__(:load, "", map, %{})}
```
josevalim: Instead of passing an empty string, let's pass nil, and change the code to do `source || __schema__(:source)`.

michalmuskala: We also don't have `id_types` from the adapter here. I think we need them.

michalmuskala:
I've added `embeds_many`.

We definitely need to think about `id_types` in `Ecto.Type.load` - I couldn't reuse `User` model because of that, as it uses `binary_id`.

lib/ecto/type.ex
```diff
@@ -291,10 +293,19 @@ defmodule Ecto.Type do
...
+    {:ok, %Ecto.Query.Tagged{value: [], type: :map}}
+  end
+
+  def dump({:array, type}, value) when is_list(value) do
     dump_array(type, value, [], false)
   end
 
+  def dump({:map, model}, %{__struct__: model} = struct) do
+    value = Map.take(struct, model.__schema__(:fields))
+    {:ok, %Ecto.Query.Tagged{value: value, type: :map}}
```
josevalim: Why not simply return the map? If we are going to start tagging maps, then we should be consistent and tag them all.

josevalim:
> We definitely need to think about id_types in Ecto.Type.load - I couldn't reuse User model because of that, as it uses binary_id.

I know what we should do. We should load the embeds inside `__schema__(:load)`. We have all the information there. This doesn't solve the issue though of when just the embed is selected in select. :S We may need to change that part too. Let's go by parts and fix the main one and add a test for selecting an embed.

integration_test/support/migration.exs
```diff
@@ -57,6 +57,11 @@ defmodule Ecto.Integration.Migration do
       add :count, :integer
     end
 
+    create table(:orders) do
+      add :item, :map
+      add :items, :map
```
josevalim: Shouldn't it be `{:array, :map}`?

lib/ecto/schema.ex
josevalim: I have an idea. What if instead of calling `__load__` from the mongo adapter, you simply return a map and we add a clause here that will match the `{:&, _, _}` against the map and call `__load__`? https://github.com/elixir-lang/ecto/blob/master/lib/ecto/repo/queryable.ex#L143

Since we will start handling maps and embeds, I guess we should handle this case automatically too. Keep in mind though the SQL adapters will already return a struct, in such cases, `__load__` should be a no-op (i.e. if a struct is given).

integration_test/support/models.exs
```diff
@@ -149,6 +149,7 @@ defmodule Ecto.Integration.Barebone do
   This module is used to test:
 
     * A model wthout primary keys
+    * Embedding
```
josevalim: We can remove this comment?

lib/ecto/type.ex
```diff
@@ -534,6 +561,8 @@ defmodule Ecto.Type do
   defp of_base_type?(:string, term), do: is_binary(term)
   defp of_base_type?(:map, term),    do: is_map(term) and not Map.has_key?(term, :__struct__)
 
+  defp of_base_type?({:map, model}, %{__struct___: struct}), do: model == struct
```
josevalim: The second argument can be a string... and it will explode in this case. Let's add tests for `{:map, Model}` to `Ecto.TypeTest` please. the proper solution for this case would be something like:

    defp of_base_type?({:map, struct}, %{__struct___: struct}), do: true
    defp of_base_type?({:map, _struct}, _), do: false

integration_test/support/migration.exs
josevalim: As we discussed, let's add an `:embed` type.

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -25,6 +25,8 @@ if Code.ensure_loaded?(Mariaex.Connection) do
 
     def query(conn, sql, params, opts \\ []) do
       params = Enum.map params, fn
+        %Ecto.Query.Tagged{type: :map, value: value} ->
```
josevalim: This should be gone.

michalmuskala:
As we discussed I changed how embeds_many is handled. For now there's only the `:array` container type

lib/ecto/type.ex
```diff
@@ -291,7 +293,20 @@ defmodule Ecto.Type do
     {:ok, %Ecto.Query.Tagged{value: nil, type: type(type)}}
   end
 
-  def dump({:array, type}, value) do
+  def dump({:embed, %{embed: model}}, %{__struct__: model} = struct) do
```
josevalim: check the cardinality is one, otherwise you can accidentally dump a model in an embed_many.

josevalim:
@michalmuskala This is looking great, go ahead with the other tasks in the gist and we can review it again tomorrow!

lib/ecto/type.ex
michalmuskala: I don't check it exactly for recursing in embed_many, I think I need to extract it to a separate function.

lib/ecto/changeset.ex
```diff
@@ -262,7 +262,7 @@ defmodule Ecto.Changeset do
   defp cast_field(param_key, type, params) do
     case Map.fetch(params, param_key) do
       {:ok, value} ->
-        case Ecto.Type.cast(type, value) do
+        case Ecto.Type.cast(type, value, %{}) do
```
michalmuskala: We don't have id_types here. I don't know how big of a problem is that

lib/ecto/model/timestamps.ex
```diff
@@ -23,7 +23,7 @@ defmodule Ecto.Model.Timestamps do
     if get_change changeset, field do
       changeset
     else
-      force_change changeset, field, Ecto.Type.load!(type, timestamp_tuple(use_usec))
+      force_change changeset, field, Ecto.Type.load!(type, timestamp_tuple(use_usec), %{})
```
michalmuskala: It's fine here because we will never get a binary_id in timestamps

lib/ecto/type.ex
```diff
@@ -204,16 +208,13 @@ defmodule Ecto.Type do
...
-  Normalizes a type.
-
-  The only type normalizable is binary_id which comes
-  from the adapter.
-  """
-  def normalize({comp, :binary_id}, %{binary_id: binary_id}), do: {comp, binary_id}
-  def normalize(:binary_id, %{binary_id: binary_id}), do: binary_id
-  def normalize(:binary_id, %{}), do: raise "adapter did not provide a type for :binary_id"
-  def normalize(type, _id_types), do: type
+  defp normalize({comp, :binary_id}, %{binary_id: binary_id}), do: {comp, binary_id}
```
michalmuskala: Should normalize be private now? We don't use it anywhere else

lib/ecto/type.ex
```diff
@@ -263,39 +264,64 @@ defmodule Ecto.Type do
...
 
+      iex> dump(:binary_id, "7d5bed50-e8ec-4a74-b863-eb977f3db92e", %{binary_id: Ecto.UUID})
+      {:ok, %Ecto.Query.Tagged{tag: nil, type: :uuid,
+        value: <<125, 91, 237, 80, 232, 236, 74, 116, 184, 99, 235, 151, 127, 61, 185, 46>>}}
+
   """
-  @spec dump(t, term) :: {:ok, term} | :error
-  def dump(type, nil) do
+  @spec dump(t, term, map) :: {:ok, term} | :error
+  def dump(type, nil, _id_types) do
```
michalmuskala: We need to catch nils here because we don't have id_types when schemas are compiles, and they check the default value by dumping it. I think it won't work with other values though...

lib/ecto/changeset.ex
josevalim: It is fine to pass an empty map.

lib/ecto/type.ex
josevalim: Yes, pls.

lib/ecto/type.ex
josevalim: Yes, it is ok to have this clause here. Just be consistent and do the same for load and cast too.

lib/ecto/type.ex
```diff
@@ -263,39 +264,64 @@ defmodule Ecto.Type do
...
+    {:ok, %Ecto.Query.Tagged{value: nil, type: type(type)}}
+  end
+
+  def dump(type, value, id_types) do
+    dump(normalize(type, id_types), value)
+  end
+
+  defp dump(type, nil) do
```
josevalim: We don't need this clause.

lib/ecto/type.ex
josevalim: This is going to be moved to the `Ecto.Schema.Serializer.dump`, so it will be easy to share.

lib/ecto/type.ex
michalmuskala: We do for arrays, as we use `dump/2` to iterate. We could use `dump/3` but in that case we would need to extract arrays before normalization.

michalmuskala:
I added the serializer, and removed loading logic from `Ecto.Schema`, I'm not entirely sure how to use that in `Ecto.Repo.Model` or `Ecto.Repo.Queryable.to_select` that we talked about earlier.

josevalim:
I will figure that out. Btw, I agree with you we should have an
embedded_schema thing or just embedded. Although I think we should still
keep the code that removes the primary key so we can embed even models not
ready for embedding. I will review those things once I get back home!



-- 


lib/ecto/type.ex
```diff
@@ -256,6 +257,10 @@ defmodule Ecto.Type do
   defp do_match?(type, type), do: true
   defp do_match?(_, _), do: false
 
+  ## load/dump/cast
+
+  alias Ecto.Schema.Serializer
```
josevalim: Do it at the top of the file please.

lib/ecto/type.ex
```diff
@@ -194,6 +194,10 @@ defmodule Ecto.Type do
   @spec type(t) :: t
   def type(type)
 
+  def type({:embed, %{cardinality: :one}}), do: :map
```
josevalim: This will never be called for an embed for Ecto, so we can remove those three clauses.

integration_test/support/migration.exs
josevalim: Ping.

integration_test/support/models.exs
```diff
@@ -171,5 +172,38 @@ defmodule Ecto.Integration.Tag do
...
+defmodule Ecto.Integration.Item do
+  @moduledoc """
+  This module is used to test:
+
+    * Embedding
+
+  """
+  use Ecto.Integration.Model
+
+  # TODO decide what to do with source here
```
josevalim: Let's remove this TODO and keep this model as is. Let's discuss adding an embed shortcut in the future although we shouldn't change this code to use (so we continue testing PKs are not included).

lib/ecto/model.ex
```diff
@@ -154,6 +154,11 @@ defmodule Ecto.Model do
     assoc.__struct__.assoc_query(assoc, values)
   end
 
+  def build_embedded(%{__struct__: model} = struct, assoc, attributes \\ %{}) do
```
josevalim: Let's get rid of this function until we figure out the best way to add this. It seems not necessary (so far).

josevalim: This should also allow us to remove some functions from `Ecto.Embedded`.

lib/ecto/schema/serializer.ex
```diff
@@ -0,0 +1,66 @@
...
+  alias Ecto.Schema.Metadata
+
+  @doc """
+  Loads recursively given model from data.
+
+  Data can be either a map with string keys or a tuple of index and row,
+  where index specifies the place in the row, where data for loading model
+  starts.
+  """
+  def load!(model, source, data, id_types) do
```
josevalim: This function does not really raise, so let's get rid of the `!`.

lib/ecto/schema/serializer.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  @doc """
+  Dumps recursively given model's struct.
+
+  ## Options:
+
+    * `:skip_pk` - whether primary keys should be dumped (default: `false`)
+
+  """
+  def dump!(struct, id_types, opts \\ []) do
```
josevalim: This function does not really raise, so let's get rid of the `!`.

integration_test/support/models.exs
```diff
@@ -171,5 +172,38 @@ defmodule Ecto.Integration.Tag do
...
+
+    * Embedding one model
+    * Embedding many models with map type
+
+  """
+  use Ecto.Integration.Model
+
+  schema "orders" do
+    embeds_one :item, Ecto.Integration.Item
+    # TODO
```
josevalim: Let's remove this TODO and comment too so we can merge this.

josevalim:
@michalmuskala I did one last batch of comments. Let's address those and merge this first step. <3 Then I will briefly review the Serializer stuff and use it in more places.

josevalim:
Then let's talk tomorrow because I think the next steps are cast in changeset and callbacks.

lib/ecto/schema/serializer.ex
michalmuskala: But is uses `Ecto.Type.load!` that may raise that we're not catching.

josevalim:
I have done my changes on top. We are ready to continue the work here (you will need to slightly update your adapter though).

PR #753, opened 2015-07-07
Introduce `Ecto.Repo.after_connect/1` (refs #599)

JonathanTron:
This is a tentative implementation for #599 

In the current implementation `Ecto.Repo.after_connect/1` is called
synchronously with the new connection pid, after a new connection is initiated
but before the connection pid is returned to the pool.

`Ecto.Adapters.SQL` functions can't be called in the hook because the
connection is not yet returned to the pool. This means you can only used
the driver functions to make queries using the connection pid.


MSch:
This is fine for my use case, I also just need to send a query to the DB (and register the pid), nothing that should delay the checkout significantly.

JonathanTron:
@MSch that's great.

I've added a test to ensure `Ecto.Adapters.SQL` sets the repo in the options used by the pool module, but I'm still not sure how to test that `after_connect/1` is called. There is no specific tests about other similar callbacks, `Ecto.Repo.log/1` is tested indirectly when checking that a failure in calling the method would cancel a transaction, but for `after_connect/1` it's problematic because it would mean raising even before the transaction is started...

Does anyone can help me finding a way to test it?

MSch:
I'd go with an integration test that sends a message to a locally registered process saying "yep I was called".

JonathanTron:
Another thing, `Ecto.Adapters.SQL.Sandbox` don't use `Ecto.Adapters.Connection` and I've not updated it to call `Ecto.Repo.after_connect/1`... what should we do about it?

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -11,7 +11,13 @@ if Code.ensure_loaded?(Mariaex.Connection) do
 
     def connect(opts) do
       opts = Keyword.update(opts, :port, @default_port, &normalize_port/1)
-      Mariaex.Connection.start_link(opts)
+      case Mariaex.Connection.start_link(opts) do
+        {:ok, pid} ->
+          repo = Keyword.fetch!(opts, :repo)
+          repo.after_connect(pid)
```
fishcakez: I'm not sure we want to promote this anti-pattern:
```elixir
case start_link() do
  {:ok, pid} = ok ->
    do_stuff()
    ok
  {:error, _} = error ->
    error
end
```
If `do_stuff/0` can crash - unless the crash is because `pid` exited - then `pid` becomes orphaned because the parent doesn't know about `pid`. With the current pools, the parent will crash because the workers do not catch or rescue. This means `pid` will get the workers exit signal and exit immediately. However if a connection is started by a `Supervisor`, the crash will be caught and `pid` orphaned. `pid` will remain alive until the `Supervisor` exits. The most common cause of this is for a call in `do_stuff/0` to timeout. The adapters `Tds` and `Mariaex` both make this mistake.

However because the current workers don't catch the crashes, they will exit if an `after_connect/1` crashes. The `Poolboy` pool will instantly(/blindly) restart the workers forever. This means a pool can hit this loop forever, effectively `max_restarts: :infinity`. This may look like fault tolerance but the application is useless because it is just constantly failing, almost "too fault tolerant". If this is happening and it is an anticipated error we want to handle the error closer to where it occurs (as close as possible), otherwise we want the error to bubble up the supervision tree because there is a bug.

If there is a case where errors should not bubble up the supervision tree, the errors should be caught and some form of circuit breaker and/or backoff should be used so that the process doesn't get stuck in a continuous loop of failure.

With the other two pools (`SojournBroker` and `SQL.Sandbox`) the error does bubble up the supervision tree. However if the `after_connect/1` crashes because of a network issue (e.g. database is down) we don't want the error to bubble up, we want to handle it. Using `Postgrex` the `after_connect/1` will crash if the postgres server is unreachable because it's `start_link/1` returns before trying to connect and any queries will fail. Adapters should be sure to block their `connect/1` calls until a connection is established to prevent this.

This leads us back around to having to catch the `connect/1` call in the pools. However if we catch there we face the orphan problem described above.

I'm unsure of the best solution, I know @josevalim said not to build it into the pools. However I think the `after_connect/1` should either be built into the drivers (`Postgrex`, `Mariex` etc.) or the pools. If we are only going to support native queries it might be better to leave this to the drivers to handle appropriately. However if it is to be part of Ecto (so that all adapters can use the feature and to prevent duplication), it would be better to provide a helper function for the pools to call, which calls `connect/1` and `after_connect/1` and ensures no orphaned processes and catches (and logs?!) `after_connect/1` errors.

@JonathanTron, I am sorry for going on for so long. I wanted to highlight some existing issues in the current adapters.

JonathanTron: @fishcakez thanks for the review and insights. I completely agree with the shortcomings of the solution, but honestly I'm not familiar enough with Ecto/Elixir to define how to overcome them here. 

It seems from your explanation that the issue is more pervasive and need to be addressed at another level. I'm willing to work toward a better implementation if you can guide me in the right direction. Should I wrap the call in a `try/catch`, something different ?

fishcakez: We can have a pattern like this:
```elixir
def connect(mod, params, after_connect) do
  case mod.connect(params) do
    {:ok, pid} ->
       after_connect(mod, after_connect, pid)
    {:error, _} = err ->
       err
end

defp after_connect(mod, after_connect, pid) do
   try do
     after_connect.(pid)
   catch
     {class, reason} ->
       mod.disconnect(pid)
       {class, reason, System.stacktrace()}
   else
     _ ->
       :ok   
   end
end
```
Then a pool can call that helper function with the relevant mod/params/fun.

I am slightly tempted to say we should evaluate `after_connect` in a task so we can have a timeout for it and isolate if from the worker so nothing leaks. Lets see what @josevalim thinks.

JonathanTron: Do you mean something like:

```elixir
defp after_connect(mod, after_connect, pid) do
   try do
     Task.await(Task.async(fn -> after_connect.(pid) end))
   catch
     {class, reason} ->
       mod.disconnect(pid)
       {class, reason, System.stacktrace()}
   else
     _ ->
       :ok   
   end
end
```

josevalim: I am ok with doing it in a task but we shuold first check if after_connect exists in the repo, so we don't spawn 20 tasks in prod to do nothing.

fishcakez: Yeah except we only need to catch the `Task.await/2` exits:
```elixir
catch
   :exit, {:timeout, [Task, :await, [%Task{pid: task_pid}, _]}} ->
      Process.exit(task_pid, :kill)
      {:error, :timeout}
   :exit, {reason, {Task, :await, _}} ->
      {:error, reason}
```
I messed up the catch, its not `{class, reason}`, its `class, reason`. That's what happens when writing and refactoring in a github comment :).

Notice that `timeout` is a special case because the task is still alive. Then we can disconnect the connection as you did.

JonathanTron:
Ok I've tried to improve the call to `Ecto.Repo.after_connect/1` by introducing a new `Ecto.Adapters.Connection.after_connect/2` overridable function which handle the call to `Ecto.Repo.after_connect/1` in a more sensible way.

The call is now run in a `Task` and wrapped in `try/catch` to better handle failures.

@fishcakez I don't know if this was what you imagined? 

Feedback much appreciated.

lib/ecto/adapters/connection.ex
```diff
@@ -12,6 +12,38 @@ defmodule Ecto.Adapters.Connection do
...
+  defmacro __using__(_) do
+    quote do
+
+      @behaviour Ecto.Adapters.Connection
+
+      def after_connect(conn, opts) do
+        repo = opts[:repo]
+        if function_exported?(repo, :after_connect, 1) do
+          try do
+            Task.await(Task.async(fn -> repo.after_connect(conn) end))
```
MSch: Should this use Ecto's timeout value?

lib/ecto/adapters/connection.ex
```diff
@@ -12,6 +12,39 @@ defmodule Ecto.Adapters.Connection do
...
+          try do
+            Task.async(fn -> repo.after_connect(conn) end)
+            |> Task.await(opts[:timeout])
+          catch
+            :exit, {:timeout, [Task, :await, [%Task{pid: task_pid}, _]]} ->
+              Process.exit(task_pid, :kill)
+              {:error, :timeout}
+            :exit, {reason, {Task, :await, _}} ->
+              {:error, reason}
+            class, reason ->
```
fishcakez: We don't need a catch all here because `after_connect` is carried out in the task and we only want to catch errors from the task, which will be of the form: `:exit, {reason, {Task, :await, [task, timeout]}}`.

fishcakez:
I don't think we need to add a callback to `Ecto.Adapters.Connection`. I think it would be simpler to add a helper function to `Ecto.Adapters.Connection` that pool workers call to start a connection (instead of calling the connection module directly). This function can then start the worker and apply the `after_connect` function you have defined in the `__using__` macro if the `after_connect` callback is exported.

fishcakez:
Nevermind the use works, lets stick with it if @josevalim is happy.

JonathanTron:
I think there's still something missing and I'm not sure how to handle it. In a previous message, @josevalim said it would be wise not to generate task when the `after_connect/1` is not present. I've added a check to see if it's exported but given I've defined it with a default in `Ecto.Repo` it will always be defined... can we have something like an optional `defcallback` so that it does not generate warning if it's not defined?

lib/ecto/repo.ex
```diff
@@ -171,7 +171,11 @@ defmodule Ecto.Repo do
         end, ecto_conn_pid: entry.connection_pid)
       end
 
-      defoverridable [log: 1, __pool__: 0]
+      def after_connect(conn) do
```
josevalim: Just don't define this function so the `function_exported?` check works (and pls remove the defcallback).

lib/ecto/adapters/connection.ex
```diff
@@ -12,6 +12,37 @@ defmodule Ecto.Adapters.Connection do
 
   use Behaviour
 
+  defmacro __using__(_) do
+    quote do
+
+      @behaviour Ecto.Adapters.Connection
+
+      def after_connect(conn, opts) do
```
josevalim: Let's please not inject this function. Define it in Ecto.Adapters.Connection and call it directly .

josevalim:
@fishcakez I am torn between between moving the after_connect call to the pool worker or leaving it in the adapters. Thoughts?

In any case, don't worry about me being torn, @JonathanTron. We can always change things later, the current approach is almost ready to ship. I have added some comments.

JonathanTron:
@josevalim Thanks for the feedback I will make the changes.

fishcakez:
> @fishcakez I am torn between between moving the after_connect call to the pool worker or leaving it in the adapters. Thoughts?

We also need to add a disconnect function (using exit signals) that pool workers can call to close connection processes. Having those in the same place would make sense I think. I guess there will always be more adapters than pools so its easier to put the extra code in the pools.

josevalim:
> We also need to add a disconnect function (using exit signals) that pool workers can call to close connection processes. Having those in the same place would make sense I think. I guess there will always be more adapters than pools so its easier to put the extra code in the pools.

Perfect. Do we need to keep the current disconnect after we add the new one?

fishcakez:
> Perfect. Do we need to keep the current disconnect after we add the new one?

I don't think so. I think we can always shutdown with exit signals. Once this is in I'll add the disconnect.

josevalim:
Merged manually and moved the logic to the pool with the help of a `connect/2` function.

josevalim:
Thank you! :heart

JonathanTron:
Thanks to you for handling the last bits :+1:

josevalim:
Team work! :D

PR #755, opened 2015-07-07
Seed SojournBroker.Worker's random number generator for its backoff.

MSch:
/cc @fishcakez

### DIFF
--- a/lib/ecto/pools/sojourn_broker/worker.ex
+++ b/lib/ecto/pools/sojourn_broker/worker.ex
@@ -41,6 +41,10 @@ defmodule Ecto.Pools.SojournBroker.Worker do
     backoff_threshold = div(max_backoff, 3)
     lazy = Keyword.get(worker_opts, :lazy, true)
 
+    # Seed random number generator for backoff. Can't use `now()` since it warns on Erlang 18.
+    {_, sec, micro} = :os.timestamp()
+    :random.seed({sec, micro, :erlang.phash2([make_ref()])})
+
     s = %{conn: nil, module: module, params: params, transaction: nil,
           broker: Process.whereis(broker), tag: tag, ref: nil, fun: nil,
           monitor: nil, backoff: min_backoff, min_backoff: min_backoff,


lib/ecto/pools/sojourn_broker/worker.ex
```diff
@@ -41,6 +41,9 @@ defmodule Ecto.Pools.SojournBroker.Worker do
     backoff_threshold = div(max_backoff, 3)
     lazy = Keyword.get(worker_opts, :lazy, true)
 
+    # Seed random number generator for backoff
+    :random.seed(:erlang.now)
```
lexmag: Could it be `:os.timestamp()`?

fishcakez: `:erlang.now/0` will generate a warning in 18.0+ so we shouldn't introduce it.

This 18.0 docs for `:random` suggest:
```erlang
random:seed(erlang:phash2([node()]), erlang:monotonic_time(), erlang:unique_integer())
```

I guess we could use the has of a reference for a unique enough number:
```elixir
{_, Sec, Micro} = :os.timestamp()
{Sec, Micro, :erlang.phash2([make_ref()])}
```

josevalim: Are we relying on the different workers having different seeds? If not, `:os.timestamp` is fine. If you need something else, maybe even the last example with phash2 of `self()`?

fishcakez: Yes we want all the workers to have different seeds.

fishcakez: The randomess is so that the workers dont backoff in unison.

MSch: As an aside: The actual implementation in rand.erl doesn't use `erlang:phash2([node()])` it uses `erlang:phash2([{node(),self()}])`.

https://github.com/erlang/otp/blob/maint/lib/stdlib/src/rand.erl#L88-L90

fishcakez: Lets copy `:rand` and use that hash.

PR #758, opened 2015-07-08
Shutdown connections with exit signals

fishcakez:
I've used the shutdown procedure used by `Supervisor`, is renaming to shutdown ok?


josevalim:
@fishcakez as I mentioned on IRC, we could also model this after stop. I don't have a preference for any of them, pick whatever you find more appropriate.

PR #767, opened 2015-07-10
Embedding part 2

michalmuskala:
Currently only works for `embeds_one`. We need to think how to represent changes and errors in db arrays.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -27,7 +27,7 @@ defmodule Ecto.Changeset do
 
   defstruct valid?: false, model: nil, params: nil, changes: %{}, repo: nil,
             errors: [], validations: [], required: [], optional: [],
-            filters: %{}
+            filters: %{}, status: nil
 
   @type t :: %Changeset{valid?: boolean(),
                         repo: atom | nil,
@@ -38,10 +38,12 @@ defmodule Ecto.Changeset do
                         optional: [atom],
                         errors: [error],
                         validations: [{atom, String.t | {String.t, [term]}}],
-                        filters: %{atom => term}}
+                        filters: %{atom => term},
+                        status: nil | status}
 
   @type error :: {atom, error_message}
   @type error_message :: String.t | {String.t, integer}
+  @type status :: :insert | :update | :delete
 
   @number_validators %{
     less_than:                {&</2,  "must be less than %{count}"},
@@ -212,26 +214,18 @@ defmodule Ecto.Changeset do
     params = convert_params(params)
     types  = module.__changeset__
 
-    {optional, {changes, errors}} =
-      Enum.map_reduce(optional, {%{}, []},
+    {optional, {changes, errors, valid?}} =
+      Enum.map_reduce(optional, {%{}, [], true},
                       &process_param(&1, :optional, params, types, model, &2))
 
-    {required, {changes, errors}} =
-      Enum.map_reduce(required, {changes, errors},
+    {required, {changes, errors, valid?}} =
+      Enum.map_reduce(required, {changes, errors, valid?},
                       &process_param(&1, :required, params, types, model, &2))
 
-    %Changeset{params: params, model: model, valid?: valid?(changes, errors),
+    %Changeset{params: params, model: model, valid?: valid?,
                errors: Enum.reverse(errors), changes: changes, required: required,
                optional: optional}
   end
-
-  defp valid?(_changes, [_|_]), do: false
-  defp valid?(changes, []), do: Enum.all?(changes, fn
-    {_, [%Changeset{} | _] = changesets} ->
-      Enum.all?(changesets, &Map.get(&1, :valid?))
-    {_, %Changeset{valid?: value}} -> value
-    _                              -> true
-  end)
 
   defp process_param({key, fun}, kind, params, types, model, acc) do
     {key, param_key} = cast_key(key)
@@ -250,29 +244,34 @@ defmodule Ecto.Changeset do
   end
 
   defp do_process_param(key, param_key, kind, params, {:embed, embed},
-                        current, {changes, errors}) do
+                        current, {changes, errors, valid?}) do
     {key,
      case cast_embed(param_key, embed, params, current) do
-       {:ok, embed_changes} ->
-         {Map.put(changes, key, embed_changes), errors}
+       {:ok, embed_changes, embed_valid?} ->
+         {Map.put(changes, key, embed_changes), errors, valid? && embed_valid?}
        :missing ->
-         {changes, error_on_nil(kind, key, current, errors)}
+         {errors, valid?} = error_on_nil(kind, key, current, errors, valid?)
+         {changes, errors, valid?}
        :invalid ->
-         {changes, [{key, "is invalid"}|errors]}
+         {changes, [{key, "is invalid"}|errors], false}
      end}
   end
 
-  defp do_process_param(key, param_key, kind, params, type, current, {changes, errors}) do
+  defp do_process_param(key, param_key, kind, params, type, current,
+                        {changes, errors, valid?}) do
     {key,
      case cast_field(param_key, type, params) do
        {:ok, ^current} ->
-         {changes, error_on_nil(kind, key, current, errors)}
+         {errors, valid?} = error_on_nil(kind, key, current, errors, valid?)
+         {changes, errors, valid?}
        {:ok, value} ->
-         {Map.put(changes, key, value), error_on_nil(kind, key, value, errors)}
+         {errors, valid?} = error_on_nil(kind, key, value, errors, valid?)
+         {Map.put(changes, key, value), errors, valid?}
        :missing ->
-         {changes, error_on_nil(kind, key, current, errors)}
+         {errors, valid?} = error_on_nil(kind, key, current, errors, valid?)
+         {changes, errors, valid?}
        :invalid ->
-         {changes, [{key, "is invalid"}|errors]}
+         {changes, [{key, "is invalid"}|errors], false}
      end}
   end
 
@@ -305,66 +304,17 @@ defmodule Ecto.Changeset do
   defp cast_key(key) when is_atom(key),
     do: {key, Atom.to_string(key)}
 
-  defp cast_embed(param_key, %{cardinality: :one, embed: embed, changeset: fun},
-                  params, current) do
+  defp cast_embed(param_key, embedded, params, current) do
     case Map.fetch(params, param_key) do
       {:ok, value} ->
-        {:ok, apply(embed, fun, [value, current || embed.__struct__()])}
+        case Ecto.Embedded.cast(embedded, value, current) do
+          :error -> :invalid
+          result -> result
+        end
       :error ->
         :missing
     end
   end
-
-  defp cast_embed(param_key, %{container: :array, embed: embed, changeset: fun},
-                  params, current) do
-    case Map.fetch(params, param_key) do
-      {:ok, value} when is_list(value) ->
-        [pk] = embed.__schema__(:primary_key)
-        {pk, param_pk} = cast_key(pk)
-        current = Enum.into(current, %{}, &{Map.get(&1, pk), &1})
-        process_embeds(process_embed_params(value, param_pk, []),
-                       embed, fun, current, [])
-      {:ok, _value} ->
-        :invalid
-      :error ->
-        :missing
-    end
-  end
-
-  defp process_embeds(:error, _mod, _fun,  _current, _acc), do: :invalid
-
-  defp process_embeds([], mod, fun, current, acc) do
-    previous = Enum.map(current, &apply(mod, fun, [%{}, elem(&1, 1)]))
-    {:ok, Enum.reverse(acc, previous)}
-  end
-
-  defp process_embeds([{key, params} | rest], mod, fun, current, acc) do
-    case Map.pop(current, key) do
-      {nil, _current} ->
-        :invalid
-      {model, current} ->
-        changeset = apply(mod, fun, [params, model])
-        process_embeds(rest, mod, fun, current, [changeset | acc])
-    end
-  end
-
-  defp process_embeds([params | rest], mod, fun, current, acc) do
-    changeset = apply(mod, fun, [params, mod.__struct__()])
-    process_embeds(rest, mod, fun, current, [changeset | acc])
-  end
-
-  defp process_embed_params([], _pk, acc), do: Enum.reverse(acc)
-
-  defp process_embed_params([map | rest], pk, acc) when is_map(map) do
-    case Map.fetch(map, pk) do
-      {:ok, value} ->
-        process_embed_params(rest, pk, [{value, map} | acc])
-      :error ->
-        process_embed_params(rest, pk, [map | acc])
-    end
-  end
-
-  defp process_embed_params(_params, _pk, _acc), do: :error
 
   defp cast_field(param_key, :binary_id, params) do
     # Since we don't have the adapter types here,
@@ -400,10 +350,10 @@ defmodule Ecto.Changeset do
     end) || params
   end
 
-  defp error_on_nil(:required, key, nil, errors),
-    do: [{key, "can't be blank"}|errors]
-  defp error_on_nil(_kind, _key, _value, errors),
-    do: errors
+  defp error_on_nil(:required, key, nil, errors, _valid?),
+    do: {[{key, "can't be blank"}|errors], false}
+  defp error_on_nil(_kind, _key, _value, errors, valid?),
+    do: {errors, valid?}
 
   ## Working with changesets
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -41,12 +41,12 @@ defmodule Ecto.Changeset do
                         errors: [error],
                         validations: [{atom, String.t | {String.t, [term]}}],
                         filters: %{atom => term},
-                        status: nil | status,
+                        status: status,
                         types: nil | %{atom => Ecto.Type.t}}
 
   @type error :: {atom, error_message}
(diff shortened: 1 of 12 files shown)


lib/ecto/embedded.ex
```diff
@@ -30,7 +31,32 @@ defmodule Ecto.Embedded do
...
+
+  The new_fun can have formats as the changeset field. Returns the struct
+  with the changeset field updated to have a changeset fun, that can be
+  called with params.
+  """
+  def normalize_changeset(%__MODULE__{embed: module} = embed, new_fun) do
+    update_in embed.changeset, &do_normalize_changeset(&1, module, new_fun)
+  end
+
+  defp do_normalize_changeset({mod, fun}, _module, nil),
```
josevalim: We should not keep anonymous functions around because they are harder to serialize. If you want to normalize, normalize it to `{module, function}` at compile time. Also, we could make this function private and call it when we build the embedding, instead of it being public and calling it later.

lib/ecto/schema.ex
```diff
@@ -778,6 +787,10 @@ defmodule Ecto.Schema do
 
   ## Options
 
+    * `:changeset` - the default changeset function to call during casting,
+      can be overriden in cast/4. Can be either an atom of function name
+      defined in embedded model, or a {module, function} tuple
```
josevalim: `{module, function}`

lib/ecto/changeset.ex
```diff
@@ -225,33 +225,86 @@ defmodule Ecto.Changeset do
                optional: optional}
   end
 
-  defp process_param(key, kind, params, types, model, {changes, errors}) do
+  defp process_param({key, fun}, kind, params, types, model, acc) do
     {key, param_key} = cast_key(key)
+    type = type!(types, key, fun)
```
josevalim: Actually, if a fun was given, we should certify it is an embed. Otherwise we should raise, so you likely want to have both `type!` and `embed!` functions.

lib/ecto/changeset.ex
```diff
@@ -225,33 +225,86 @@ defmodule Ecto.Changeset do
...
+  defp process_param(key, kind, params, types, model, acc) do
+    {key, param_key} = cast_key(key)
+    type = type!(types, key, nil)
+    current = Map.get(model, key)
+
+    do_process_param(key, param_key, kind, params, type, current, acc)
+  end
+
+  defp do_process_param(key, param_key, kind, params, {:embed, embed}, current, {changes, errors}) do
+
```
josevalim: :scissors:

lib/ecto/changeset.ex
```diff
@@ -225,33 +225,86 @@ defmodule Ecto.Changeset do
...
+    {key,
+     case cast_embed(param_key, embed, params, current) do
+       :missing ->
+         {changes, error_on_nil(kind, key, current, errors)}
+       {embed_changes, embed_errors} ->
+         {merge_embed(changes, key, embed_changes, &Map.put/3),
```
josevalim: I think this is too complicated?

    {Map.update(changes, key, embed_changes, &Map.merge(&1, embed_changes)),
     Keyword.update(errors, key, embed_errors, & &1 ++ embed_errors)}

Or something of sorts. You could also just override whatever is in key, I don't think we would validate the same field twice:

    {Map.put(changes, key, embed_changes),
     [{key, embed_errors}|errors]}

lib/ecto/changeset.ex
josevalim: In other words, if you have a function, assert it is an embed, merge the given function into the embed struct, and then call the code as usual.

lib/ecto/embedded.ex
josevalim: Actually, ignore this. Let's keep it simple for now as a simple `:atom`.

lib/ecto/type.ex
```diff
@@ -547,6 +547,10 @@ defmodule Ecto.Type do
 
   """
   @spec cast(t, term, map) :: {:ok, term} | :error
+  def cast({:embed, embed}, value, id_types) do
+    cast_embed(Ecto.Embedded.normalize_changeset(embed, nil), value, id_types)
```
josevalim: As mentioned above, let's get rid of normalize_changeset. Just assume the :changeset will be an atom.

lib/ecto/changeset.ex
```diff
@@ -225,33 +225,86 @@ defmodule Ecto.Changeset do
...
   defp cast_key(key) when is_atom(key),
     do: {key, Atom.to_string(key)}
 
+  defp cast_embed(param_key, %{changeset: fun}, params, model) do
+    case Map.fetch(params, param_key) do
+      {:ok, nil} ->
+        model
+      {:ok, value} ->
+        args = if is_nil(model), do: [value], else: [value, model]
+        changeset = fun.(args)
```
josevalim: When you cast the changeset, you should pass the embed model currently in the parent model! If there is no model, you should create one! I.e. the changeset function in embed should always receive **two** arguments.

josevalim: When you cast the changeset, you should pass the embed model currently in the parent model! If there is no model, you should create one (just call `embed_model.__struct__`). I.e. the changeset function in embed should always receive **two** arguments.

lib/ecto/changeset.ex
michalmuskala: This won't work. We receive a map of `%{key: :new_value}`, and we need to transform it into `%{[:embed, :key] => :new_value}`. The same is with errors, but with keywords.

josevalim: Oh, we shouldn't worry about this as we won't be doing diffs for now. Let's go by parts because managing diffs is definitely harder. Remember we are still unsure how diffs will work when it is the time to run callbacks for embed_one (we got stuck last time when we were discussing this). So let's go diffless. :)

josevalim: Plus, because we need to run callbacks, we shouldn't even build those list keys now, because we still need the whole model to run callbacks. So let's keep this simple.

josevalim: Actually, we shuoldn't even merge the changes and errors. We will need the whole changeset to run callbacks.

michalmuskala: Then we need to figure out what to do with arrays. Have lists with nils where there is no value? Having a structure like that, we could have `[:embed, 0, :key]` as a key.

michalmuskala:
I changed things in comments, and added support for `embeds_many` with array container.

I think that we will need `embedded_schema` to make sure the model has `:binary_id` as primary key.

We should now think what to do with generated changesets.

lib/ecto/changeset.ex
```diff
@@ -220,38 +220,152 @@ defmodule Ecto.Changeset do
       Enum.map_reduce(required, {changes, errors},
                       &process_param(&1, :required, params, types, model, &2))
 
-    %Changeset{params: params, model: model, valid?: errors == [],
+    %Changeset{params: params, model: model, valid?: valid?(changes, errors),
```
josevalim: Instead of doing this check later, we should get this information from `Enum.map_reduce` above as we traverse.

lib/ecto/changeset.ex
```diff
@@ -220,38 +220,152 @@ defmodule Ecto.Changeset do
...
+          "embeds are supported in cast, associations are not)"
+    end
+  end
 
   defp cast_key(key) when is_binary(key),
     do: {String.to_atom(key), key}
   defp cast_key(key) when is_atom(key),
     do: {key, Atom.to_string(key)}
 
+  defp cast_embed(param_key, %{cardinality: :one, embed: embed, changeset: fun},
```
josevalim: Let's move this (or part of it) to Ecto.Embedded.cast so we don't end-up polluting changeset too much with embeds concerns and diffs and what not.

josevalim:
This looks great. From what we discussed, there are a couple changes left:

1. We need to be sure we now dump primary keys for embeds

2. We need to add a status field to the changeset, which is nil for the parent changeset, but, for every embed, it should be one of: `:insert | :update | :delete`

3. That said, we need to write a bunch of tests.

* embeds_one: the child changeset should have status :insert
* embeds_one with custom changeset function: the child changeset should have status :insert
* embeds_one when id on the given embed matches the one in the model: the child changeset should have status :update
* embeds_one when id on the given embed does match the one in the model: the child changeset should have status :insert

With embeds _many, we will have something similar, the difference is that we will always keep the merge of two lists, having all insert, update and delete.

michalmuskala:
I think dumping keys will need to be done later on - during casting we don't have `id_types` from the adapter, and as primary keys in embeds are `:binary_id` we don't know how to dump them. Should we treat them as simple binaries here too, as it's done for regular fields?

When and how should we mark things for delete? There is nothing for this currently. Should we have some magical key like `_delete` to mark it? 

Also with `embeds_one` when we get params without primary key, we assume we're inserting the new model. What should happen with the old one? We probably need to track it somehow in order to run delete callbacks.

In the situation when we get primary key in params, and it does not match the model, I think we should mark it as invalid - it probably means there is an attempt to move this model from somewhere else, but we don't know where from.

Speaking of invalid, we should probably have some better error message than `is invalid` for the case of primary keys not matching - unfortunately I can't think of anything reasonable so I went with `is invalid` for now.

josevalim:
> I think dumping keys will need to be done later on - during casting we don't have id_types from the adapter, and as primary keys in embeds are :binary_id we don't know how to dump them. Should we treat them as simple binaries here too, as it's done for regular fields?

I didn't meant to dump them now. We should change the dump code you wrote in the previous pull request.

> When and how should we mark things for delete? There is nothing for this currently. Should we have some magical key like _delete to mark it?

For now, we will only keep a changeset around with `status: :delete`. We will figure out what to do with those later but I am more inclined to say we won't invoke delete callbacks for embeds.

> Also with embeds_one when we get params without primary key, we assume we're inserting the new model. What should happen with the old one? We probably need to track it somehow in order to run delete callbacks.

It is worth remembering the original is still in the model. So if we need to run callbacks on it, we can do that later on.

> In the situation when we get primary key in params, and it does not match the model, I think we should mark it as invalid - it probably means there is an attempt to move this model from somewhere else, but we don't know where from.

No. For embeds_one, we won't do anything, the original is still in the model and we can use it later to run callbacks. For embeds_many, the matching primary keys are marked as update, the ones not given as delete, and the new ones as insert.

josevalim:
Also please break the existing tests apart and be sure to test all cases I mentioned in the previous comment. We will need them. :)

lib/ecto/embedded.ex
```diff
@@ -30,7 +33,89 @@ defmodule Ecto.Embedded do
...
+    do: Enum.into(current, %{}, &{Map.get(&1, pk), &1})
+
+
+  defp changeset_status(mod, fun, params, nil) do
+    changeset = apply(mod, fun, [params, mod.__struct__()])
+    %{changeset | status: :insert}
+  end
+
+  defp changeset_status(mod, fun, nil, model) do
+    changeset = apply(mod, fun, [%{}, model])
```
josevalim: I don't think we should run validations here. Just create an empty changeset with no params and the model.

lib/ecto/type.ex
```diff
@@ -597,6 +601,17 @@ defmodule Ecto.Type do
     end
   end
 
+  defp cast_embed(embed, value, _id_types) do
```
josevalim: Beautiful.

test/ecto/embedded_test.exs
```diff
@@ -31,10 +44,142 @@ defmodule Ecto.EmbeddedTest do
...
+    assert other.valid?
+    assert michal.model.id == "michal"
+    assert michal.status == :delete
+    assert michal.valid?
+    refute changeset.valid?
+  end
+
+  test "cast embeds_many with invalid params" do
+    changeset = cast(%Author{}, %{"profiles" => "value"}, ~w(profiles))
+    assert changeset.errors == [profiles: "is invalid"]
```
josevalim: TODO Discuss this error message.

test/ecto/embedded_test.exs
```diff
@@ -31,10 +44,142 @@ defmodule Ecto.EmbeddedTest do
...
+    assert other.model.id == "other"
+    assert other.status == :update
+    assert other.valid?
+    assert michal.model.id == "michal"
+    assert michal.status == :delete
+    assert michal.valid?
+    refute changeset.valid?
+  end
+
+  test "cast embeds_many with invalid params" do
```
josevalim: Let's also add an assertion when the parameter is nil. We shuold mark it as invalid too.

michalmuskala:
I rebased onto the failing master build, and it took me half an hour to figure out that wasn't me :smile:

I think this last commit addresses majority of concerns for part 2 of embedding.

lib/ecto/changeset.ex
```diff
@@ -506,10 +583,17 @@ defmodule Ecto.Changeset do
...
   def put_change(%Changeset{} = changeset, key, value) do
-    update_in changeset.changes, &put_change(changeset.model, &1, key, value)
+    changeset = ensure_has_types(changeset)
+    type = Map.get(changeset.types, key)
+    update_in changeset.changes, &put_change(changeset.model, &1, key, value, type)
+  end
+
+  defp put_change(_model, acc, key, value, {:embed, embed}) do
+    value = Ecto.Embedded.wrap_change(embed, value)
+    Map.put(acc, key, value)
```
michalmuskala: Can we do something smarter here?
For embed_one it would be easy to check for changes, but for embed_many it will be hard...

lib/ecto/changeset.ex
```diff
@@ -38,10 +40,13 @@ defmodule Ecto.Changeset do
...
                         errors: [error],
                         validations: [{atom, String.t | {String.t, [term]}}],
-                        filters: %{atom => term}}
+                        filters: %{atom => term},
+                        status: nil | status,
+                        types: nil | %{atom => Ecto.Type.t}}
 
   @type error :: {atom, error_message}
   @type error_message :: String.t | {String.t, integer}
+  @type status :: :insert | :update | :delete
```
josevalim: It can be `nil` too.

lib/ecto/changeset.ex
```diff
@@ -106,17 +111,19 @@ defmodule Ecto.Changeset do
 
   def change(%Changeset{changes: changes} = changeset, new_changes)
       when is_map(new_changes) do
-    %{changeset | changes: get_changed(changeset.model, changes, new_changes)}
+    changeset = ensure_has_types(changeset)
```
josevalim: Is this a smell? Why was a changeset created without types? Maybe we should be raising on put_change if there is no type instead of "ensuring"?

lib/ecto/changeset.ex
```diff
@@ -362,6 +432,13 @@ defmodule Ecto.Changeset do
     raise ArgumentError, message: "different repos when merging changesets"
   end
 
+  defp merge_identical(object, nil, _thing), do: object
+  defp merge_identical(nil, object, _thing), do: object
+  defp merge_identical(object, object, _thing), do: object
+  defp merge_identical(_, _, thing) do
+    raise ArgumentError, message: "different #{thing} when merging changesets"
```
josevalim: No need for `message: `. Can we also say the values we got in the error message?

lib/ecto/changeset.ex
josevalim: What do you mean? I am not sure what you want to achieve here. :)

lib/ecto/embedded.ex
```diff
@@ -22,15 +31,109 @@ defmodule Ecto.Embedded do
...
+    {pk, param_pk} = primary_key(mod)
+    current = process_current(current, pk)
+    map_changes(params, param_pk, mod, fun, current, [], true)
+  end
+
+  def cast(_embed, _params, _current) do
+    :error
+  end
+
+  def wrap_change(%Embedded{cardinality: :one}, value) do
```
josevalim: We need docs. And let's call this function "change". :)

test/ecto/embedded_test.exs
```diff
@@ -31,10 +44,147 @@ defmodule Ecto.EmbeddedTest do
...
       %Ecto.Embedded{field: :profile, cardinality: :one, owner: Author,
-                     embed: Profile, container: nil}
+                     embed: Profile, container: nil, on_cast: :required_changeset}
 
     assert Author.__schema__(:embed, :profiles) ==
       %Ecto.Embedded{field: :profiles, cardinality: :many, owner: Author,
-                     embed: Profile, container: :array}
+                     embed: Profile, container: :array, on_cast: :changeset}
+  end
+
```
josevalim: Extra new line. :D :scissors:

josevalim:
I have added a couple more comments. We are almost there.

The last change we need to do is to set the `:status` for the parent changeset as soon as we call `Repo.insert/update/delete` and add test for it. :)

lib/ecto/changeset.ex
michalmuskala: Can a changeset be created only with functions from this module (I know it can, but do we support it)?
If not you are right - we can make sure we always have types.

lib/ecto/changeset.ex
michalmuskala: For regular fields we check whether the new value is equal to the one in the model, and in that case we either skip adding this to the list of changes or remove it from that list. 

> If the change is already present, it is overridden with
> the new value, also, if the change has the same value as
> the model, it is not added to the list of changes.

I was thinking about something similar, but I think its too hard to do - at least now.

michalmuskala:
I addressed all the issues

lib/ecto/changeset.ex
```diff
@@ -506,10 +580,16 @@ defmodule Ecto.Changeset do
   """
   @spec put_change(t, atom, term) :: t
   def put_change(%Changeset{} = changeset, key, value) do
```
josevalim: Let's add a clause that checks here if `types: nil` and, if so, we raise a nice error message.

PR #768, opened 2015-07-10
Decode result sets outside of pool

josevalim:
Hello @elixir-lang/ecto-adapters and @liveforeverx.

* [x] Postgres
* [x] MySQL
* [ ] MSSQL
* [ ] SQLite3
* [ ] MongoDB

This pull request introduces two big changes to adapters in Ecto:

1. Now we expect result sets to return a list of lists instead of lists of tuples

2. We now support decoding queries outside of the connection and outside of the pool, this means we will hold the connection for shorter period of times

Changing 1 is straight-forward but cumbersome. Changing 2 is a bit more complex so let's talk about it more.

You will notice that we now require a new function to be implemented by SQL adapters called `decode/2`, this function will receive the result set returned by `query/4` and a mapper function invoked when traversing the result. The idea is that adapters should be able to return a result set without being decoded so we can do it later on **once**, after putting the connection back in the pool.

Here is how the changes for the Postgres adapter look like: https://github.com/elixir-lang/ecto/commit/54429aa4397d41a11e5d00faa677fa2c164dacbf#diff-ee25bf8df5589fbf812f652738bfbdae

In order to support this feature, we did two changes to the Postgres driver:

1. First we moved the response decoding from server to client: https://github.com/ericmj/postgrex/commit/c99653c2e07bcc4030661f9d824836d076fc10d0

2. Then we allow the decoding to be controlled with an option: https://github.com/ericmj/postgrex/commit/2ea75b9e8f51ed79f02597fedda50e532d70a234

If for some reason you can't do client side decoding, you can emulate it (but we don't recommend it), which is what the MySQL adapter does today (we will change it soon™):

https://github.com/elixir-lang/ecto/commit/54429aa4397d41a11e5d00faa677fa2c164dacbf#diff-975ad8a7b1f8e27ac59595bcdbdd0a07R31

If you have any questions, please let me know!

### DIFF
--- a/integration_test/pg/deadlock_test.exs
+++ b/integration_test/pg/deadlock_test.exs
@@ -75,7 +75,7 @@ defmodule Ecto.Integration.DeadlockTest do
   end
 
   defp pg_advisory_xact_lock(key) do
-    %{rows: [{:void}]} =
+    %{rows: [[:void]]} =
       Ecto.Adapters.SQL.query(PoolRepo, "SELECT pg_advisory_xact_lock($1);", [key])
   end
 end
--- a/integration_test/sql/transaction.exs
+++ b/integration_test/sql/transaction.exs
@@ -229,15 +229,6 @@ defmodule Ecto.Integration.TransactionTest do
   test "transaction exit includes :timeout on query timeout" do
     assert match?({:timeout, _},
       catch_exit(PoolRepo.transaction(fn ->
-        PoolRepo.insert!(%Trans{text: "12"}, [timeout: 0])
-      end)))
-
-    assert [] = PoolRepo.all(Trans)
-  end
-
-  test "transaction exit includes :timeout on nested query timeout" do
-    assert match?({:timeout, _},
-      catch_exit(PoolRepo.transaction(fn ->
         PoolRepo.transaction(fn ->
           PoolRepo.insert!(%Trans{text: "13"}, [timeout: 0])
         end)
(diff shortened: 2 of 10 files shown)


liveforeverx:
MySQL adapter, released as v0.4.1

josevalim:
Thanks you @liveforeverx! MySQL is the first to be ready!

jazzyb:
I'm having some difficulty understanding the decode changes.  What exactly does "decode" mean in this context?  I've glanced at the Postgrex and Mariaex drivers, and decoding looks like some sort of conversion from binary to Elixir types.  Is this correct?

josevalim:
Yes, exactly. Maybe you don't have much decoding to do due to the nature of
the C library. If that's the case, you can just have decode be a no-op, as
in the earlier MySQL commit linked at the top.

-- 


jazzyb:
Okay.  Thanks for the info, José.

michalmuskala:
Is there something changing for MongoDB? I looked through the changes and it seems nothing changed in the `Adapter` interface. I also run all tests on the PR branch and all seem to pass.

josevalim:
I merged this but we still need to keep track of the adapters work. I will open up an issue for that.

josevalim:
@michalmuskala yes, you need to do those changes too. The first step is to change the mongo driver to do the decoding in the client. Talk to @ericmj, you can probably send a PR to the driver. Then you can do manual decoding, similar to how we added to postgrex.

josevalim:
Let's continue the discussion in #773.

PR #769, opened 2015-07-10
added a commented description to the has_many/has_one :through example of Schema

elbow-jason:
After some confusion about has_many/has_one :through I was requested to improve the has_many/has_one :through documentation. Hopefully the comment will prevent other people from having the same issue.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -478,6 +478,13 @@ defmodule Ecto.Schema do
           has_one :permalink, Permalink
           has_many :comments_authors, through: [:comments, :author]
 
+          # From the has_many :through example above, in the list
+          # `[:comments, :author]` the `:comments` refers to the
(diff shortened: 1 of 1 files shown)


lib/ecto/schema.ex
```diff
@@ -478,6 +478,13 @@ defmodule Ecto.Schema do
           has_one :permalink, Permalink
           has_many :comments_authors, through: [:comments, :author]
 
+          # From the has_many :through example above, in the list
```
josevalim: Can we move this comment to right above the `has_many :through` thing?

elbow-jason: will do

elbow-jason:
I moved the comment up and corrected the wording

josevalim:
Merged manually, thank you

elbow-jason:
:+1:

PR #774, opened 2015-07-12
Log DB's connection ids

MSch:
I've started working on https://github.com/elixir-lang/ecto/issues/667

Right now this is just a proof of concept. I've got some questions /cc @ericmj @fishcakez @josevalim

- The mapping between the DB `conn_id` and Ecto's `conn_pid` is kept in ETS, managed via a separate process.

  Advantages: a) No need to store this meta information in the Worker (where else could it be stored?) and somehow get it to `Ecto.Adapters.SQL.query` so it can go into the LogEntry. b) Also can be fetched from other parts (e.g. exometer) so it can query the DB for stats on what's happening.

  Disadvantages: Either Ecto becomes an application with it's own top supervisor where the global GenServer+ETS table live or every Repo becomes a supervisor that has a separate GenServer+ETS table for the mapping.

  Right now I think the recommended pattern (at least what `mix phoenix.new` creates) is to have `YourApp.Repo` as a worker in `YourApp`'s supervisor. But with sbroker the Repo is no longer a worker it should be a supervisor. So I think that is already something that needs to be thought about.

  Right now (because I noticed the problem with poolboy worker vs sbroker supervisor and because I need a place to put my GenServer) I'm thinking it would be nice to have all repo processes in Ecto's supervision tree, not in the supervision tree of my app. (That would also make the configuration simpler, no need for the `otp_app` key any more, config becomes just `config :ecto, YourApp.repo, ...`.)

- Storing the mapping in ETS and cleaning the table when the connection process dies means that the mapping is then no longer available for logging. I'm not sure if that can actually happen, or if in that case the whole query won't be logged anyhow, since it never reaches `Repo.log`?

- I've introduced the `Ecto.Adapter.fetch_conn_id(conn_pid)` callback, that returns either `{:ok, conn_id}` as an arbitrary db-specific term or `:ignore` if the connection doesn't support it. Is `Ecto.Adapter` that a good place to put the callback?

- Is the way I call `register_conn_id` in `Ecto.Adapters.Connection` ok?

### DIFF
--- a/integration_test/mysql/test_helper.exs
+++ b/integration_test/mysql/test_helper.exs
@@ -1,4 +1,4 @@
-Logger.configure(level: :info)
+Logger.configure(level: :debug)
 ExUnit.start exclude: [:array_type, :read_after_writes, :case_sensitive,
                        :uses_usec, :strict_savepoint]
 
--- a/integration_test/pg/test_helper.exs
+++ b/integration_test/pg/test_helper.exs
@@ -1,4 +1,4 @@
-Logger.configure(level: :info)
+Logger.configure(level: :debug)
 ExUnit.start
 
 # Configure Ecto for support and tests
--- a/lib/ecto/adapter.ex
+++ b/lib/ecto/adapter.ex
@@ -39,6 +39,8 @@ defmodule Ecto.Adapter do
   defcallback start_link(repo, options) ::
               {:ok, pid} | :ok | {:error, {:already_started, pid}} | {:error, term}
 
+  defcallback fetch_conn_id(pid) :: {:ok, term} | :ignore | {:error, term}
+
   @doc """
   Returns the binary id type for this adapter.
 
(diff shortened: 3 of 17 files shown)


josevalim:
Thanks @MSch! What I would do is to Ecto.Repo.start_link start an Ecto.Repo.Supervisor which would start the adapter as a child (a supervisor) and your worker too.

However, after looking at the changes, they still look like too much complexity to push to both adapters and ecto. Maybe it is easier to handle it in your app now that we have after_connect? It will be much cleaner for you to handle this in your app as you wouldn't need to worry about all the indirection and we don't add complexity to Ecto itself.

MSch:
I'm already handling it in my app, was just trying to get it upstream as well.

> However, after looking at the changes, they still look like too much complexity to push to both adapters and ecto.

What specifically do you think is too much complexity? Introducing `Ecto.Repo.Supervisor` or just the actual conn_id to conn_pid mapping?

josevalim:
@MSch both. :)

PR #775, opened 2015-07-12
Implementes table renaming

bcardarella:
Usage:

```elixir
def change
  rename table(:foo), table(:bar)
end
```

This will rename the table "foo" to "bar".

Works with backward migrations


bcardarella:
@josevalim green with integration test

josevalim:
Beautiful!

PR #778, opened 2015-07-13
Make psql invocation fail fast

maxnordlund:
This enables the `ON_ERROR_STOP` variable for faster failures and the `--no-psqlrc` flag for added robustness.

### DIFF
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -130,7 +130,7 @@ defmodule Ecto.Adapters.Postgres do
     end
 
     host = database[:hostname] || System.get_env("PGHOST") || "localhost"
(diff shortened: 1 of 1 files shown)


josevalim:
AFAIK, ON_ERROR_STOP only makes sense when running multiple commands. All we ever run in this case though is a CREATE or DROP DB. So I don't think we need it?

maxnordlund:
Better safe then sorry I guess. The manual says:

> EXIT 
> psql returns 0 to the shell if it finished normally, 1 if a fatal error of its own occurs (e.g. out of memory, file not found), 2 if the connection to the server went bad the session was not interactive, and 3 if an error occurred in a script and the variable ON_ERROR_STOP was set.

It should also future proof it if multiple commands are needed.

josevalim:
Alright, thank you!

PR #780, opened 2015-07-14
Raise Ecto.QueryError

### LINKED ISSUE #766: Raise ecto query errors in MySQL and Postgres
@michalmuskala is doing this in his mongo adapter and we should do the same, otherwise users are stuck with errors like these that do not emit the query:

https://github.com/drewolson/scrivener/blob/master/lib/scrivener.ex#L153

@michalmuskala can you please give us a hand in here and do this change for pg/mysql too?

michalmuskala:
Resolves #766


josevalim:

PR #782, opened 2015-07-14
Clarify syntax for the array type

wsmoak:
Is there any convention in the `mix docs` format that says "replace this with a valid value"?

If you're reading a shell script and you see $FOO you know to either set an environment variable or replace that bit.

I got stuck on `{:array, inner_type}` here and did not 'get' that I should replace the text inner_type with, for example `:string`. It is of course blindingly obvious in retrospect, but it wasn't at the time. :)

This adds a note to clarify what to do with `inner_type`.


josevalim:

Issue #671, opened 2015-06-22
Support changeset constraints


josevalim:
We will start with unique_constraint (and add unique_index to migrations too).

wojtekmach:
Based on conversation on https://github.com/elixir-lang/ecto/pull/787 here's a quick'n'dirty example of how constraint errors could be translated into changeset errors: https://github.com/wojtekmach/ecto/commit/365a924724a9786bb487df08ab06d87b9b8ce64a.

michalmuskala:
That looks great, but I think we would need to handle errors somehow at the adapter level - we can't handle all adapters at the repo level.

wojtekmach:
@michalmuskala yup, totally agreed. That was just a proof-of-concept.

josevalim:
In master!

PR #787, opened 2015-07-14
Add note about validate_unique not being safe under high load

wojtekmach:
Based on: http://api.rubyonrails.org/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -876,6 +876,11 @@ defmodule Ecto.Changeset do
   reason, you may want to trigger the unique validations as last
   in your validation pipeline.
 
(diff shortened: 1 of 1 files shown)


josevalim:
Yup. And it is going away soon. :)

wojtekmach:
> Yup. And it is going away soon. :)

Good to hear :) I assume a unique contstrain validation error (from DB) will be transformed into a changeset error or something like that? If so (or however else this is going to be solved), let me know if you need any help with that

josevalim:
@wojtekmach that's exactly it. I am not sure yet how it will work but it is issue #671.

PR #788, opened 2015-07-14
Handle adding references/foreign key constraint in alter table

hubertlepicki:
So this is solution for https://github.com/elixir-lang/ecto/issues/722

Please note that MySQL is pretty picky about matching data types on columns foreign keys are being defined with, so in addition to adding foreign key constraint, we need to make sure it is converted to BIGINT UNSIGNED.

I have added integration tests as well, thanks for the tip @josevalim. They are passing for MIX_ENV=mysql and MIX_ENV=pg on my machine running MariaDB and PostgreSQL 9.4 / Arch Linux.

Please let me know if there's anything that can be done better, I'm happy to ammend.

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -65,6 +65,32 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule AlterForeignKeyMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:alter_fk_migration_users)
+
+      create table(:alter_fk_migration_posts) do
+        add :alter_fk_migration_user_id, :integer
+      end
+
+      alter table(:alter_fk_migration_posts) do
+        add :content, :text
+        modify :alter_fk_migration_user_id, references(:alter_fk_migration_users, on_delete: :nilify_all)
+      end
+
+      execute "INSERT INTO alter_fk_migration_users (id) VALUES ('1')"
+      execute "INSERT INTO alter_fk_migration_posts (id, content, alter_fk_migration_user_id) VALUES ('1', 'text', '1')"
+      execute "DELETE FROM alter_fk_migration_users"
+    end
+
+    def down do
+      drop table(:alter_fk_migration_posts)
+      drop table(:alter_fk_migration_users)
+    end
+  end
+
   defmodule DropColumnMigration do
     use Ecto.Migration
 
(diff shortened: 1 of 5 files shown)


integration_test/sql/migration.exs
```diff
@@ -65,6 +65,31 @@ defmodule Ecto.Integration.MigrationTest do
...
+        add :alter_fk_migration_user_id, :integer
+      end
+
+      alter table(:alter_fk_migration_posts) do
+        modify :alter_fk_migration_user_id, references(:alter_fk_migration_users, on_delete: :nilify_all)
+      end
+
+      execute "INSERT INTO alter_fk_migration_users (id) VALUES ('1')"
+      execute "INSERT INTO alter_fk_migration_posts (id, alter_fk_migration_user_id) VALUES ('1', '1')"
+      execute "DELETE FROM alter_fk_migration_users"
```
josevalim: Maybe we shouldn't remove them, so we can assert in the test there is one entry here?

integration_test/sql/migration.exs
```diff
@@ -65,6 +65,31 @@ defmodule Ecto.Integration.MigrationTest do
...
+
+    def up do
+      create table(:alter_fk_migration_users)
+
+      create table(:alter_fk_migration_posts) do
+        add :alter_fk_migration_user_id, :integer
+      end
+
+      alter table(:alter_fk_migration_posts) do
+        modify :alter_fk_migration_user_id, references(:alter_fk_migration_users, on_delete: :nilify_all)
```
josevalim: Just for sanity checking purpose, can you add another column at the same time you modify? Just to make sure the SQL syntax won't choke.

josevalim:
Beautiful, thank you, I have added some comments!

integration_test/sql/migration.exs
hubertlepicki: I might be missing the point why we would want to do that. I am deleting the record to validate that ON CASCADE clause gets executed...

integration_test/sql/migration.exs
hubertlepicki: this has been now addressed, seems like it works (tests still pass)

hubertlepicki:
@josevalim I have addressed one of the issues, not the other one (with delete). I don't get what you would like to change in this part. The reason why I am deleting from `users` table is to verify that the foreign key gets nullified in `posts` table, i.e. foreign constant's ON DELETE clause gets executed. If that ain't right please explain how you see that :)

josevalim:

josevalim:
I missed the constraint bit. thank you!

hubertlepicki:
cool

PR #791, opened 2015-07-15
fis integration_test/sql to pass under mysql

hubertlepicki:
When I was running tests yesterday, I noticed the integration_test/sql failure under MySQL. Thisi s because of the assertion in one of tests that checks if lock_for_update has been set globally. This is a direct result of change ebcda6bbc875e51b73863524edc03d9ef8de62ed when the option was copied to pg/test_helper.exs but *not* to mysql/test_helper.exs. I am uncertain if that was intentional or simply frogotten by @josevalim to do so.

As far as I can tell, simply adding the option back makes the integration_test/sql to pass again under MySQL. I assume those test should pass under all SQL adapters...

Is that the correct behavior and good fix?

### DIFF
--- a/integration_test/mysql/all_test.exs
+++ b/integration_test/mysql/all_test.exs
@@ -1,4 +1,5 @@
 Code.require_file "../sql/escape.exs", __DIR__
+Code.require_file "../sql/lock.exs", __DIR__
 Code.require_file "../sql/migration.exs", __DIR__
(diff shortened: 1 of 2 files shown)


hubertlepicki:
Actually it looks like the lock tests are being disabled intentionally, in this commit: 3a79060f55f37ae2be208a28a0c0f9228efd2d7f by @laurocaetano but I'm struggling to understand why, since the tests are passing...

PR #796, opened 2015-07-16
make date/time/datetime use Ecto.*

### LINKED ISSUE #759: Make date/time/datetime use Ecto.*

tony612:
Resolve #759 

@josevalim I find a problem: `load` and `dump` functions of Ecto.DateTime/Date/Time will raise an error if the params can't match, but Ecto.Type will return `:error`. So we'd better let them of Ecto.* return `:error` for unmatched params? But I'm not sure if it's suitable for Ecto.*. Any idea?


josevalim:
Beautiful. Let's fix the ecto types to return `:error` too.

tony612:
@josevalim Done :)

josevalim:

PR #799, opened 2015-07-17
Add uses_msec tag to integration tests

jazzyb:
SQLite can calculate integrals down to the millisecond (but not the microsecond) so this commit distinguishes between the two in `integration_test/cases/interval.exs`.


josevalim:

PR #806, opened 2015-07-18
Dry run casts embeds as dry run

### LINKED ISSUE #795: Dry run should cast changesets as dry run too
This function should call all embeds recursively passing `:empty` param:

https://github.com/elixir-lang/ecto/blob/master/lib/ecto/changeset.ex#L207-L220

/cc @michalmuskala

michalmuskala:
Closes #795

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -10,6 +10,14 @@ defmodule Ecto.ChangesetTest do
     schema "" do
       field :name
     end
+
+    def changeset(params, model \\ %Author{}) do
+      cast(model, params, ~w(name))
+    end
+
+    def custom_changeset(params, model \\ %Author{}) do
+      cast(model, params, ~w(), ~w(name))
+    end
   end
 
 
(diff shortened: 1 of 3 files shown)


test/ecto/changeset_test.exs
```diff
@@ -118,6 +127,37 @@ defmodule Ecto.ChangesetTest do
...
+    changeset = cast(%Post{authors: [%Author{}]}, :empty, ~w(authors))
+    [author_changeset] = changeset.changes.authors
+    assert author_changeset.model == %Author{}
+    assert author_changeset.params == nil
+    assert author_changeset.changes == %{}
+    assert author_changeset.errors == []
+    assert author_changeset.validations == []
+    assert author_changeset.required == [:name]
+    assert author_changeset.optional == []
+    assert author_changeset.action == :update
```
michalmuskala: I'm not sure this is the correct behaviour. What should be the action here? Should it depend on primary key presence?

josevalim: There no params... so there isn't anything really we can do... I am thinking that, if it is an embeds_many, we should just create a single empty changeset in the list (with no action) and that is it.

josevalim: Actually, ignore me. What you have today is excellent.

PR #807, opened 2015-07-18
Rename column

bcardarella:
This adds support for renaming columns. It comes in two forms:

*Alter Table Form*
```elixir
alter table(:foo) do
  rename :bar, :baz
end
```

*Outside Alter Table Form*
```elixir
rename table(:foo), :bar, :baz
```

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -108,6 +108,24 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule RenameColumnMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:rename_col_migration) do
+        add :to_be_renamed, :integer
+      end
+
+      rename table(:rename_col_migration), :to_be_renamed, to: :was_renamed
+
+      execute "INSERT INTO rename_col_migration (was_renamed) VALUES (1)"
+    end
+
+    def down do
+      drop table(:rename_col_migration)
+    end
+  end
+
   defmodule OnDeleteMigration do
     use Ecto.Migration
 
@@ -281,7 +299,14 @@ defmodule Ecto.Integration.MigrationTest do
     :ok = down(TestRepo, 20090906120000, DropColumnMigration, log: false)
   end
 
-  @tag :rename
+  @tag :rename_column
+  test "rename column" do
+    assert :ok == up(TestRepo, 20150718120000, RenameColumnMigration, log: false)
+    assert 1 == TestRepo.one from p in "rename_col_migration", select: p.was_renamed
+    :ok = down(TestRepo, 20150718120000, RenameColumnMigration, log: false)
+  end
+
+  @tag :rename_table
   test "rename table" do
     assert :ok == up(TestRepo, 20150712120000, RenameMigration, log: false)
     assert :ok == down(TestRepo, 20150712120000, RenameMigration, log: false)
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -14,6 +14,17 @@ if Code.ensure_loaded?(Mariaex.Connection) do
       Mariaex.Connection.start_link(opts)
     end
 
+    def query(conn, sqls, params, opts) when is_list(sqls) do
+      Enum.reduce sqls, {}, fn(sql, response) ->
+        case response do
+          {:error, _} ->
+            response
+          _ ->
+            query(conn, sql, params, opts)
+        end
+      end
+    end
+
     def query(conn, sql, params, opts \\ []) do
       params = Enum.map params, fn
         %Ecto.Query.Tagged{value: value} -> value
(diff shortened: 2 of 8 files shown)


integration_test/sql/migration.exs
```diff
@@ -111,6 +111,29 @@ defmodule Ecto.Integration.MigrationTest do
...
+  defmodule RenameColumnMigration do
+    use Ecto.Migration
+
+    def up do
+      create table(:rename_col_migration) do
+        add :to_be_renamed_inside_alter, :integer
+        add :to_be_renamed_outside_alter, :integer
+      end
+
+      alter table(:rename_col_migration) do
```
josevalim: Can you please add another operation inside the alter table? You can add any column you want. The reason for this is that some operations can only exist by themselves inside `alter table`, so we always need to add another operation to help us catch those cases.

josevalim:
Thank you! Please also note the build failed. :)

bcardarella:
Closing, there are some bugs I need to fix. I'll reopen in a bit

bcardarella:
@josevalim I might need some insight on this. I can do a rename in mysql with the following:

```mysql
select @column_type := COLUMN_TYPE from INFORMATION_SCHEMA.COLUMNS  where table_name = 'bar' and COLUMN_NAME = 'baz' limit 1;
set @rename_stmt = concat('alter table bar change column baz boogies ', @column_type);
prepare rename_stmt from @rename_stmt;
execute rename_stmt;
```

The prepared statement will prevent from having to run a query, retrieve and store the type in ecto, then run another query interpolating the type. (what ActiveRecord currently does) However, this presents issues with how Ecto is architected. From the `change_column` function in `adapters/mysql/connection.ex` I have no context for the `alter table` statement that wraps the the rename command, I can out output an line command to be appended onto the alter statement.

I would prefer to have an elegant solution for `rename` that doesn't require passing options to explicitly pass the column type. Do you have thoughts if this is possible or not?

If this is not possible then I'll just add `rename/3` that takes a type atom that is intended only for use with mysql

josevalim:
@bcardarella you could pass the table into column_changes:

https://github.com/elixir-lang/ecto/blob/master/lib/ecto/adapters/mysql/connection.ex#L450

`column_changes` is private so you can change it anyway you want.

However, after looking at all SQL queries for renaming columns, I think we should go with your original intuition and provide the rename command **outside** of alter table because it is often hard to combine rename with the other alter table operations. So something like:

    rename table(:posts), :title, :new_title

Or, even better:

    rename table(:posts), :title, to: :new_title

integration_test/sql/migration.exs
bcardarella: Postgres does not support other actions in the alter statement:

>All the actions except RENAME and SET SCHEMA can be combined into a list of multiple alterations to apply in parallel.

bcardarella: Is there a way to guard against this?

josevalim: Let's suppose rename only outside of alter table. I would suggest:

    rename table(:rename_cel_migration), :old_column, to: :new_column

bcardarella: :+1:

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -14,6 +14,17 @@ if Code.ensure_loaded?(Mariaex.Connection) do
       Mariaex.Connection.start_link(opts)
     end
 
+    def query(conn, sqls, params, opts) when is_list(sqls) do
+      Enum.reduce sqls, {}, fn(sql, response) ->
```
bcardarella: @josevalim I want to make sure you're OK with this. `Mira.Connection.query` only accepts one command at a time. In order to get the MySQL prepared statement working I had to split it up into its separate commands and process each one at a time. This `Enum.reduce` statement will loop through each command. If an `{:error, message}` is encountered by any of the commands it will complete the loop but no longer process any more commands. The error will be returned up the chain as the last command executed.

josevalim: I would prefer not to but I can take a stab at it before merging the PR, so no worries. Just let me know when this is ready for review.

bcardarella: Its green so its ready

lib/ecto/migration.ex
```diff
@@ -474,6 +474,17 @@ defmodule Ecto.Migration do
   end
 
   @doc """
+  Renames a column outside of the `alter` statement. Note that unlike `rename/3` it is not necessary
```
josevalim: The first line of docs should always be a short summary. Just a note, I have already fixed this locally. :)

josevalim:
Merged! To fix the problem you were facing, I added multiple statement support `execute_ddl` only: 732dcb0316d04a3b28c459e5133faadaf726e520

josevalim:

bcardarella:
@josevalim very nice, much better :)

PR #815, opened 2015-07-20
Skip embedded changes with action :update and no changes

### LINKED ISSUE #810: Ensure a changeset where action == update and changes == %{} is not stored as change for embeds
We need to ensure this happens for embeds_one and it also happens for embeds_many in the case **all** entries in the list match the criteria above.

AFAIK we do this for `embeds_one` but there are no tests. We need to ensure and test both cases.

michalmuskala:
Closes #810


josevalim:

PR #816, opened 2015-07-20
Rollback reference fix

gjaldon:
Closes https://github.com/elixir-lang/ecto/issues/804

To be able to reverse the commands on rollback, I had to change all migration commands to `queue` instead of `execute`.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -432,26 +432,18 @@ if Code.ensure_loaded?(Mariaex.Connection) do
     alias Ecto.Migration.Index
     alias Ecto.Migration.Reference
 
-    def execute_ddl({:create, %Table{} = table, columns}) do
-      engine = engine_expr(table.engine)
+    def execute_ddl({command, %Table{} = table, columns}) when command in [:create, :create_if_not_exists] do
+      engine  = engine_expr(table.engine)
       options = options_expr(table.options)
+      if_not_exists = if command == :create_if_not_exists, do: " IF NOT EXISTS", else: ""
 
-      "CREATE TABLE #{quote_table(table.name)} (#{column_definitions(columns)})" <> engine <> options
+      "CREATE TABLE" <> if_not_exists <> " #{quote_table(table.name)} (#{column_definitions(columns)})" <> engine <> options
     end
 
-    def execute_ddl({:create_if_not_exists, %Table{}=table, columns}) do
-      engine = engine_expr(table.engine)
-      options = options_expr(table.options)
+    def execute_ddl({command, %Table{name: name}}) when command in [:drop, :drop_if_exists] do
+      if_exists = if command == :drop_if_exists, do: " IF EXISTS", else: ""
 
-      "CREATE TABLE IF NOT EXISTS #{quote_table(table.name)} (#{column_definitions(columns)})" <> engine <> options
-    end
-
-    def execute_ddl({:drop, %Table{name: name}}) do
-      "DROP TABLE #{quote_table(name)}"
-    end
-
-    def execute_ddl({:drop_if_exists, %Table{name: name}}) do
-      "DROP TABLE IF EXISTS #{quote_table(name)}"
+      "DROP TABLE" <> if_exists <> " #{quote_table(name)}"
     end
 
     def execute_ddl({:alter, %Table{}=table, changes}) do
@@ -471,18 +463,8 @@ if Code.ensure_loaded?(Mariaex.Connection) do
                 if_do(index.concurrently, "LOCK=NONE")])
     end
 
-    def execute_ddl({:create_if_not_exists, %Index{}=index}) do
-      create = "CREATE#{if index.unique, do: " UNIQUE"} INDEX"
-      using  = if index.using, do: "USING #{index.using}", else: []
-
-      assemble([create,
-                "IF NOT EXISTS",
-                quote_name(index.name),
-                "ON",
-                quote_table(index.table),
-                "(#{Enum.map_join(index.columns, ", ", &index_expr/1)})",
-                using,
-                if_do(index.concurrently, "LOCK=NONE")])
+    def execute_ddl({:create_if_not_exists, %Index{}}) do
+      raise Ecto.QueryError, query: "CREATE INDEX IF NOT EXISTS", message: "Not supported in MySQL"
     end
 
     def execute_ddl({:drop, %Index{}=index}) do
@@ -492,12 +474,8 @@ if Code.ensure_loaded?(Mariaex.Connection) do
                 if_do(index.concurrently, "LOCK=NONE")])
     end
 
-    def execute_ddl({:drop_if_exists, %Index{}=index}) do
-      assemble(["DROP INDEX",
-                "IF EXISTS",
-                quote_name(index.name),
-                "ON #{quote_table(index.table)}",
-                if_do(index.concurrently, "LOCK=NONE")])
+    def execute_ddl({:drop_if_exists, %Index{}}) do
+      raise Ecto.QueryError, query: "DROP INDEX IF EXISTS", message: "Not supported in MySQL"
     end
 
     def execute_ddl({:rename, %Table{}=current_table, %Table{}=new_table}) do
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -463,9 +463,8 @@ if Code.ensure_loaded?(Mariaex.Connection) do
                 if_do(index.concurrently, "LOCK=NONE")])
     end
 
-    def execute_ddl({:create_if_not_exists, %Index{}}) do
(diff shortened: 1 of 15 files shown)


lib/ecto/migration/runner.ex
```diff
@@ -110,6 +127,8 @@ defmodule Ecto.Migration.Runner do
   Checks if a table or index exists.
   """
   def exists?(object) do
+    execute_commands()
```
gjaldon: This is the only function that implicitly triggers executing of all commands. Thought it would be okay to do this since it seems like it's only used in tests. Let me know if it should be changed.

josevalim:
Thank you @gjaldon. I will need to think about this, it is a major change. I don't think we should change the user facing `execute` to `queue` though. But let me think about this and I will get back to you.

gjaldon:
@josevalim ok, copy that. Thought to change `execute` to `queue` since it doesn't execute the commands right away anymore. Thinking about it now, we could still use `execute` for consistency with the rest of the commands like `create`. `execute` is still a great name even though the internal behavior changes.

stevedomin:
nice one @gjaldon!

I don't think there is any need to change the name to `queue`, `execute` doesn't carry any notion of time nor does it imply that the action will happen immediately

josevalim:
Hey @gjaldon! I know what is left to finish this.

1. Let's keep `execute` (instead of `queue`)

2. Let's add two new macros: `create_if_not_exists` and `drop_if_exists`. They receive the same argument as create and drop today (indexes and table).

3. Let's remove the `exists?` code today as well as the `ddl_exists?` callback and their implementation

4. Let's add a `flush` to Ecto.Migration that simply calls `Runner.execute_commands` (we can also rename `execute_commands` to `flush`). `flush` will be part of the public API and can be called multiple times. Let's update the tests to call this new function instead of `Runner.execute_commands` everywhere.

What do you think?

gjaldon:
@josevalim - I've started implementing the items you mentioned above. I think `flush` is a much better name than `execute_commands` and removes the awkwardness of using `execute`. The `create_if_not_exists` and `drop_if_exists` are a much better approach too and cleans up the code quite a lot.

My latest push is what tackles most items except `create_if_exists` and `drop_if_exists` in mysql. It is causing errors in the integration tests that are unclear and will need help with. I tested the `create table if exists` query in mysql and it works fine in the cli.

lib/ecto/migration/runner.ex
```diff
@@ -56,14 +58,31 @@ defmodule Ecto.Migration.Runner do
   end
 
   @doc """
-  Executes command tuples or strings.
+  Queued migration commands are executed.
+
+  Reverses the order commands are executed when doing a rollback
+  on a change/0 function.
+  """
+  def flush do
```
josevalim: You need to reset the commands in the agent. Otherwise flushing twice will execute the same commands twice! Try to do just a simple `get_and_update` call to the agent that retrieves the commands, direction and resets the commands in one pass!

lib/ecto/migration/runner.ex
```diff
@@ -74,16 +93,14 @@ defmodule Ecto.Migration.Runner do
...
-      Agent.get_and_update __MODULE__, fn state ->
-        {operation, object} = state.command
-        {{operation, object, Enum.reverse(state.subcommands)},
-         %{state | command: nil, subcommands: []}}
-      end
-    execute(command)
+    Agent.update __MODULE__, fn state ->
+      {operation, object} = state.command
+      command = {operation, object, Enum.reverse(state.subcommands)}
+      %{state|command: nil, subcommands: [], commands: [command|state.commands]}
```
josevalim: Nitpick: `%{state | command: nil, ...}`

lib/ecto/migration/runner.ex
josevalim: Let's add a test that flushing twice does not run things twice.

lib/ecto/migration.ex
```diff
@@ -513,15 +554,13 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Checks if a table or index exists.
-
-  ## Examples
-
-      exists? table(:products)
+  Queued migration commands are executed.
```
josevalim: "Executes queue migration commands".

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -432,35 +432,28 @@ if Code.ensure_loaded?(Mariaex.Connection) do
...
+      options = options_expr(table.options)
 
-    def ddl_exists(%Index{name: name}) do
-      """
-      SELECT COUNT(1)
-        FROM information_schema.statistics
-       WHERE table_schema = SCHEMA()
-         AND index_name = '#{escape_string(to_string(name))}'
-      """
+      "CREATE TABLE #{quote_table(table.name)} (#{column_definitions(columns)})" <> engine <> options
```
josevalim: We can share the logic between `:create` and `:create_if_not_exists` and avoid the duplication here.

josevalim: The same for drop, indexes and the other adapters.

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -478,6 +463,10 @@ if Code.ensure_loaded?(Mariaex.Connection) do
                 if_do(index.concurrently, "LOCK=NONE")])
     end
 
+    def execute_ddl({:create_if_not_exists, %Index{}}) do
+      raise Ecto.QueryError, query: "CREATE INDEX IF NOT EXISTS", message: "Not supported in MySQL"
```
gjaldon: Is this the correct error type to use?

gjaldon:
@josevalim this should be ready for review :)

lib/ecto/migration/schema_migration.ex
```diff
@@ -38,4 +28,42 @@ defmodule Ecto.Migration.SchemaMigration do
   def down(repo, version) do
     repo.delete_all from(p in __MODULE__, where: p.version == ^version), @opts
   end
+
+  defp create_migrations_table(Ecto.Adapters.MySQL = adapter, repo) do
```
josevalim: Why couldn't you use `CREATE TABLE IF NOT EXISTS` here? We should never special case behaviour for adapters.

lib/ecto/adapters/mysql/connection.ex
josevalim: No, it should be something like this: https://github.com/gjaldon/ecto/blob/rollback-reference-fix/lib/ecto/adapters/mysql/connection.ex#L488 :)

lib/ecto/migration/schema_migration.ex
gjaldon: If I use `create_if_not_exists` here, I get errors that have to do with savepoints in MYSQL. Using the old approach for MySQL fixes it. This was the weird error that you and @liveforeverx were discussing in irc. `create_if_not_exists` for MySQL works fine outside of schema migrations though.

Should we revert `ddl_exists?` so we can use it here? Any idea on how we can use `create_if_not_exists` for MySQL here?

josevalim: I will investigate this then. Thanks.

lib/ecto/migration.ex
```diff
@@ -342,13 +383,13 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Executes arbitrary SQL or a keyword command in NoSQL databases.
+  Queues arbitrary SQL or a keyword command in NoSQL databases.
 
   ## Examples
 
-      execute "UPDATE posts SET published_at = NULL"
+      queue "UPDATE posts SET published_at = NULL"
```
josevalim: Those docs are still outdated.

lib/ecto/migration.ex
```diff
@@ -342,13 +383,13 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Executes arbitrary SQL or a keyword command in NoSQL databases.
+  Queues arbitrary SQL or a keyword command in NoSQL databases.
```
josevalim: Here too.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -672,6 +668,7 @@ if Code.ensure_loaded?(Postgrex.Connection) do
 
     defp assemble(list) do
       list
+      |> Enum.reject(& &1 == "")
```
josevalim: Please don't add this. Just make "" and empty list and it will be flattened out. :)

josevalim:
@gjaldon please address the other comments then and I will take a look at the MySQL one before merging! Thank you! :heart:

josevalim:
Merged manually, thank you!

gjaldon:
:+1:

gjaldon:
@josevalim btw, what was the problem with `:create_if_not_exists` with the MySQL adapter in schema_migrations? Was it working fine or did you have to make a change?

josevalim:
We were running the DDL statement inside tests and that would commit the sandbox transaction. It was fixed here: https://github.com/elixir-lang/ecto/commit/200bea68f65c881387ebd6d0384daa22591ab184

PR #817, opened 2015-07-20
Adding using option to modify

jeregrine:
This is only supported in postgres. It is useful for when postgres cannot autocast a field to its new type. The default behaviour is to drop the column and then change it. 

And postgres surprisingly doesn't know how to cast from json to jsonb so my test shows an example..

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -65,6 +65,25 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule AlterColumnMigrationWithUsing do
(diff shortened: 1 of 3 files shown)


integration_test/sql/migration.exs
```diff
@@ -65,6 +65,25 @@ defmodule Ecto.Integration.MigrationTest do
     end
```
jeregrine: blerg just realized this is the sql migration test and this will only work in pg.

jeregrine: Not sure how to test this just for pg.

PR #831, opened 2015-07-23
Ensure embedded models have primary keys on update and delete

### LINKED ISSUE #830: Ensure update/delete actions in embeds have an id
When can do it while we apply the callbacks because that's when ids are generated and therefore managed.

michalmuskala:
Closes #830

I've also split test into couple smaller ones - it was hard to find the one failing with so many cases in one test.


josevalim:

PR #834, opened 2015-07-24
Intruduce changesets for associations

michalmuskala:


lib/ecto/changeset/relation.ex
```diff
@@ -0,0 +1,174 @@
...
+  def change(_related, nil, nil), do: {:ok, nil, false, true}
+
+  def change(%{related: mod} = related, value, current) do
+    {pk, _} = primary_key(mod)
+    cast_or_change(related, value, current, pk, pk, &do_change(&1, &2, mod))
+  end
+
+  defp do_change(value, nil, mod) do
+    fields    = mod.__schema__(:fields)
+    embeds    = mod.__schema__(:embeds)
```
josevalim: We will need to include `:assocs` here?

lib/ecto/changeset.ex
```diff
@@ -729,6 +738,9 @@ defmodule Ecto.Changeset do
         case Map.get(types, key) do
           {:embed, embed} ->
             {key, Ecto.Embedded.apply_changes(embed, value)}
+          {:assoc, _assoc} ->
+            raise ArgumentError, "associations should be removed from changeset" <>
```
josevalim: What is the idea here?

michalmuskala: We will be calling `Repo.insert!/update!/delete!` for the assocs, and using the returned model from there, and not applying the changes from the changeset (repo functions do that already for us)

josevalim:
Really nice. I think we should also port some of the tests from test/ecto/embedded_test.exs to test/ecto/association_test.exs. You should probably change the latter first though to make sure they use the same model names, it will help when moving things over and finding corner cases.

lib/ecto/changeset.ex
josevalim: I see, perfect.

michalmuskala:
Yes, we will need the tests. Making this thing actually work is the nest step ;)
For now it's mainly renaming.

paulcsmith:
Thanks for your work @michalmuskala This is going to be really helpful :D

lib/ecto/changeset.ex
sumerman: We are using `apply_changes` to build a preview before saving into the DB. Right now, with no nested changesets support we have to manually propagate application in order to build this preview. It would be sad to keep this mess even when Ecto will be able to handle this for saving.

josevalim: @sumerman are you advocating for `apply_changes` to also apply the association changes?

sumerman: @josevalim yes, it makes perfect sense for me. /cc @michalmuskala

josevalim: If so, this would definitely be possible to do, we just need to make sure we strip all associations from changes before we call `apply_changes/1` inside the repository.

michalmuskala: Yeah, we could definitely do that, and the code is mostly there with embedded models.

sumerman: @michalmuskala that would be great! thanks in advance.

michalmuskala:
@josevalim Ready for review!

josevalim:

gjaldon:
Great! Looking forward to using this :)

PR #856, opened 2015-08-05
Fix handling of nested embeds on delete

michalmuskala:


lib/ecto/repo/model.ex
```diff
@@ -252,21 +252,14 @@ defmodule Ecto.Repo.Model do
...
+  defp delete_changes(changeset, _embeds, assocs) do
     {assoc_changes, changeset} = pop_assoc_changesets(changeset, assocs)
     if map_size(assoc_changes) > 0 do
       raise ArgumentError, "nested association changes are not allowed on delete, " <>
                            "but got `#{inspect assoc_changes}`"
     end
 
-    changeset = %{changeset | changes: %{}}
-    Ecto.Changeset.change(changeset, embeds)
+    %{changeset | changes: %{}}
```
michalmuskala: Do we want to keep this?

test/ecto/repo_test.exs
```diff
@@ -894,4 +906,16 @@ defmodule Ecto.RepoTest do
...
+    sub_embed = %SubEmbed{id: @uuid, y: "xyz"}
+    embed = %MyEmbed{id: @uuid, x: "xyz", sub_embed: sub_embed}
+    TestRepo.delete!(%MyModel{id: 1, embed: embed})
+    assert [{:after_delete, MyModel}, {:after_delete, MyEmbed},
+            {:after_delete, SubEmbed}, {:before_delete, SubEmbed},
+            {:before_delete, MyEmbed}, {:before_delete, MyModel} | _] =
+      Agent.get(CallbackAgent, &get_models/1)
+  end
+
+  # Nested assocs on delete are handled with the on_delete setting and tested there
```
josevalim: Wrong line?

lib/ecto/repo/model.ex
josevalim: We can kill this I think.

lib/ecto/embedded.ex
```diff
@@ -87,12 +101,20 @@ defmodule Ecto.Embedded do
...
     check_action!(action, function, model)
-    callback = callback_for(type, action)
+    callback = callback_for(:before, action)
     Ecto.Model.Callbacks.__apply__(model, callback, changeset)
     |> generate_id(callback, model, embed, adapter)
-    |> apply_callbacks(model.__schema__(:embeds), adapter, function, type)
+    |> apply_callbacks(model.__schema__(:embeds), adapter, function, :before)
+  end
+
+  defp apply_callback(%Changeset{model: %{__struct__: model}, action: action} = changeset,
```
josevalim: Did we break them apart because of `check_action!`?

test/ecto/repo_test.exs
michalmuskala: All previous tests come in pairs for embeds and assocs. This one doesn't. I wanted to make clear why. Is there a better place to put this?

lib/ecto/embedded.ex
michalmuskala: No, because the order of running callbacks was wrong. We want to run parent callbacks first on before, and child callbacks first on after. It was wrong before with running all child callbacks after parents.

PR #859, opened 2015-08-06
Nested relations fixes

michalmuskala:


michalmuskala:
I finally managed to mark the whole association as invalid if we can't cast the primary key

lib/ecto/changeset/relation.ex
```diff
@@ -272,35 +278,38 @@ defmodule Ecto.Changeset.Relation do
...
-    primary_keys =
-      case module.__schema__(:primary_key) do
-        []  -> raise Ecto.NoPrimaryKeyFieldError, model: module
-        pks -> pks
-      end
-    param_keys = Enum.map(primary_keys, &Atom.to_string/1)
+  defp do_get_pk(model_or_params, {key, type}) do
+    case Ecto.Type.cast(type, do_get_pk(model_or_params, key)) do
+      {:ok, value} -> value
+      :error       -> :error
```
michalmuskala: This is a little bit hackish.
In case we can't cast a key we'll get `[:error]` instead of `[1]` or something, and that obviously won't match any id or binary_id, but I'm not sure it's ok to do that.

josevalim: If we can't cast, return the original value.

michalmuskala: good point :+1:

PR #872, opened 2015-08-12
Sort preloads in integration tests

michalmuskala:
MongoDB does not guarantee order of the results

### DIFF
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -35,8 +35,8 @@ defmodule Ecto.Integration.PreloadTest do
 
     # With assoc query
     assert [p3, p1, p2] = TestRepo.preload([p3, p1, p2], :comments)
-    assert [%Comment{id: ^cid1}, %Comment{id: ^cid2}] = p1.comments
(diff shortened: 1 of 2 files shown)


integration_test/cases/preload.exs
```diff
@@ -35,8 +35,8 @@ defmodule Ecto.Integration.PreloadTest do
 
     # With assoc query
     assert [p3, p1, p2] = TestRepo.preload([p3, p1, p2], :comments)
-    assert [%Comment{id: ^cid1}, %Comment{id: ^cid2}] = p1.comments
-    assert [%Comment{id: ^cid3}, %Comment{id: ^cid4}] = p2.comments
+    assert [%Comment{id: ^cid1}, %Comment{id: ^cid2}] = p1.comments |> Enum.sort_by(&(&1.id))
```
josevalim: Let's have a `sort_by_id` function in this module in special since we use it so many times...

josevalim: Just address this and go ahead and :shipit: !

PR #876, opened 2015-08-13
get_field/3 and fetch_field/2 for relations return models

### LINKED ISSUE #875: get_field for nested changesets should filter out deleted entries for has_many relations
When I use `Changeset.get_change` I expect it to return what it returns now, but for the `Changeset.get_field` the expectation is to return the value as if it comes from the model.

michalmuskala:
Closes #875


josevalim:

PR #885, opened 2015-08-16
Adding a validation error to the "[field]_confirmation" field instead of the main field

JustMikey:
I think it's more logical to add an error to the field which an user is supposed to enter the confirmation information into


josevalim:
Thank you. I am sticking with Rails convention that adds it to the regular attribute. Unless there is a strong evidence one is better than the other, I would keep it this way. :)

JustMikey:
@josevalim Hmm, I may be wrong but I think Rails adds the error to the [field]_confirmation field as well

josevalim:
You are right. Rails adds it to the `#{field}_confirmation`, thank you. @jeregrine @chrismccord @drewolson any preference?

drewolson:
Adding the error to `#{field}_confirmation` makes sense to me.

jeregrine:
:+1:

PR #888, opened 2015-08-17
Allows migration commands to be pipeable

### LINKED ISSUE #887: change to rename/2
Thoughts on the following change to `rename/2`, having it return the table that is being acted upon. This way we can write the following in a migration:

```elixir
table(:foo)
|> rename :bar, to: :baz
|> rename :qux, to: quxx
```

bcardarella:
* create
* drop
* rename

Closes #887


josevalim:

PR #889, opened 2015-08-18
fix(Ecto.Adapters.SQL): ensure binary_ids work with unique_constrant

Gazler:
Previously when using binary_ids, invalid constraints would raise:

 > (CaseClauseError) no case clause matching: {:invalid, [unique:
 > "users_email_index"]}

This adds the missing clause to the case statement for binary_ids.

I wanted to add a test for this too, but couldn't find the correct location for it. If someone could point me in the right direction then I'll add it in.


Gazler:
I was going to add the test to https://github.com/elixir-lang/ecto/blob/master/test/ecto/changeset_test.exs#L754

josevalim:
Ah, adding a test for this feature is a bit complicarted. I will take a look at it.

josevalim:
test in master! Thank you!

PR #892, opened 2015-08-19
Add reload to Repo

jsteiner:
This adds support for reloading models.

```
post = Repo.insert(%Post{text: "original"})

# post is updated elsewhere to have text "changed"

post = Repo.reload(post)

post.text # => "changed"
```


michalmuskala:
Should this support `reload!` that will raise if the record is not there?

jsteiner:
Probably, @michalmuskala. Updating...

josevalim:
Thanks but we want to ship with a small API in Repo by default. You can add conveniences to repos in your own application but we will keep it light from Ecto side.

Issue #891, opened 2015-08-19
False positive "missing parameter" errors on subsequent cast calls


sumerman:
`cast(cast(ch0, %{x: ..., y: ...}, ...), %{z: ...}, ...)`  will mark `x` and `y` as absent, although they are present in the changes set.

josevalim:
Can you give a more concrete example? What is the third parameter in both cases?

josevalim:
Ping? I wanted to release Ecto 1.0 today and this is holding it back.

sumerman:
Ouch, I'm on vacation right now. I can try to come up with the example in a couple of hours, or you can release it as is and fix this later.

josevalim:
If I understand this one, I don't think it is a bug. The required parameters are meant to check the values being given, not the ones in the changeset. So this likely isn't a bug, so no worries.

However, if you have a couple minutes today to help us solve #890, that would be preferred, because that one seems closer to a bug.

Thank you and enjoy the vacation! :)

sumerman:
@josevalim agreed, it's rather usability issue than a bug. Since it's quite common to have `Model.changeset/2` wrapper over the `cast` and other validations it would be nice to have such wrappers composable.

josevalim:
Ok, thank you for confirming the issues. I think we are good to go, we won't be fixing this one as the current behaviour is per design. :) Cheers!

sumerman:
@josevalim won't be fixing for 1.0 or ever?

josevalim:
Ever.

josevalim:
Actually, ignore me. We should fix this. We already check the model, we should likely check the changes to.

sumerman:
Hah :) Thanks :+1:

Issue #890, opened 2015-08-19
cast(cast(model, :empty, ...), params, ...) sometimes leaves relations in invalid state


sumerman:
Not sure if it's a bug or just behaviour to be documented. Below I'll try to explain what happens as I understand it.

As expected `cast(model, :empty, ...)` marks all generated changes as invalid. Later on, if `params` of subsequent `cast` lead to generation of a change that is considered no-op (`:skip`) the original "invalid" change stays in the changeset. So in the end we can have an inconsistent changeset with top-level `valid?` set to true, but with nested changesets having their `valid?` fields set to false.

josevalim:
Can you give me a more concrete example here too? It could even be a simple file or a simple test case that defines the models and shows the scenario happening. It does feel like a bug but without a way to reproduce it, I am thinking it will be really hard to fix it. :(

josevalim:
Ping? I wanted to release Ecto 1.0 today and this is holding it back.

sumerman:
@josevalim it's not present in `v0.16.0`, sorry for disturbance.

PR #893, opened 2015-08-21
Test cases for #890 and #891

sumerman:
#890 was fixed in `0.16.0` though, by not-generating invalid empty changes for relations. The test case can be used to prevent regressions in the future.
#891 is still present, but it may be an inconvenient behaviour rather than a bug.

### DIFF
--- a/test/ecto/association_test.exs
+++ b/test/ecto/association_test.exs
@@ -660,6 +660,26 @@ defmodule Ecto.AssociationTest do
     end
   end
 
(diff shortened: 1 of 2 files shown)


sumerman:
Checks obviously fail since it's a tests-only PR.

josevalim:
We have pushed a test for #890 when we fixed it. :) About #891, we are going with a won't fix, as we are casting the parameters and not the changes in changeset. Thank you!

Issue #896, opened 2015-08-22
Nested relation changes ordering sometimes breaks DB constraints.


sumerman:
It's probably an uncommon case, but during the update I have to destroy and recreate the whole set of related models and I have a uniqueness constraint for these models. Unfortunately changes list is ordered in a such way that insertions come first and only then deletions. So, during `Repo.insert ` exception is risen.

For now I have this (relation field is renamed) as a workaround: 
```elixir
Map.update!(res, :changes, &(Map.update(&1, :relation_field, [], fn d -> Enum.reverse d end)))
```

P.S.: I'm on `v0.16.0` now.

michalmuskala:
We create deletions after insertions, as it's impossible to know what to delete without the information what to keep. We need to traverse all the params, and only then we can know what to delete. 

A possible workaround is to have a specIal param, and explicitly mark models for deletion based on it, similar to how José does this in his article http://blog.plataformatec.com.br/2015/08/working-with-ecto-associations-and-embeds/ in the "deleting todo items" part.

sumerman:
@michalmuskala but it's a way to get into uniqueness troubles in general. Why not to reverse the changes list after params traversal is done?

josevalim:
@michalmuskala We have two options: we order in the changeset and then we just perform the order in the changeset in the repo or we order in the repo. I prefer the first, thoughts? Also, can you help me tackle this one?

michalmuskala:
I agree we should do this during casting/changing -this gives more flexibility. It should be fairly easy to place the autogenerated ones at the beginning. Should we also move those that the user explicitly sets the action to delete? I don't think so.

I'll be able to take this one but only in the evening. I'm traveling today the whole day.

josevalim:
> Should we also move those that the user explicitly sets the action to delete? I don't think so.

I don't think so too but we need to make sure that the order given as parameter is the respected order (if we don't yet).

PR #897, opened 2015-08-22
Test case to support #896

sumerman:
Test case to support #896.
This is almost my case distilled. CI should fail for this PR because it's test-only.


sumerman:
PR branch was rebased on master. CI should pass now.

josevalim:
Thank you!

PR #898, opened 2015-08-22
When casting/changing many relations, keep deletes in front

### LINKED ISSUE #896: Nested relation changes ordering sometimes breaks DB constraints.

michalmuskala:
Closes #896

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -233,13 +233,16 @@ defmodule Ecto.Changeset.Relation do
   end
 
   defp map_changes([], _pks, fun, current, acc, valid?, skip?) do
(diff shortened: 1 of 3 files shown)


josevalim:

PR #901, opened 2015-08-23
Structure dump and load tasks

### LINKED ISSUE #814: Create test db from development schema
Copied from https://github.com/phoenixframework/phoenix/issues/1035

I am migrating a node.js application over to Phoenix. the database already exists. My development migrations are to normalize the schema with many table/column renames. I currently cannot test the application without creating the test db manually from the development db as a template. This is fine for my development machine, but doesn't work within the context of running the application on Travis where that development db doesn't exist at all.

bcardarella:
Support for ecto to dump and load current db structure (schema)

Closes #814 

- [x] per-repo schema dumping & loading
- [x] integration tests
- [x] override source/destination files

### DIFF
--- a/lib/ecto/adapters/mysql.ex
+++ b/lib/ecto/adapters/mysql.ex
@@ -139,23 +139,28 @@ defmodule Ecto.Adapters.MySQL do
     end
   end
 
-  defp run_with_mysql(sql_command, database) do
-    unless System.find_executable("mysql") do
-      raise "could not find executable `mysql` in path, " <>
+  defp run_with_mysql(sql_command, opts) do
+    args = ["--silent", "--execute", sql_command]
+    run_with_cmd("mysql", opts, args)
+  end
+
+  defp run_with_cmd(cmd, opts, opt_args) do
+    unless System.find_executable(cmd) do
+      raise "could not find executable `#{cmd}` in path, " <>
             "please guarantee it is available before running ecto commands"
     end
 
     env =
-      if password = database[:password] do
+      if password = opts[:password] do
         [{"MYSQL_PWD", password}]
       else
         []
       end
 
-    host = database[:hostname] || System.get_env("MYSQL_HOST") || "localhost"
-    port = database[:port] || System.get_env("MYSQL_TCP_PORT") || "3306"
-    args = ["--silent", "-u", database[:username], "-h", host, "-P", to_string(port), "-e", sql_command]
-    System.cmd("mysql", args, env: env, stderr_to_stdout: true)
+    host = opts[:hostname] || System.get_env("MYSQL_HOST") || "localhost"
+    port = opts[:port] || System.get_env("MYSQL_TCP_PORT") || "3306"
+    args = ["--user", opts[:username], "--host", host, "--port", to_string(port)] ++ opt_args
+    System.cmd(cmd, args, env: env, stderr_to_stdout: true)
   end
 
   @doc false
@@ -185,5 +190,19 @@ defmodule Ecto.Adapters.MySQL do
   def insert(_repo, %{schema: schema}, _params, returning, _opts) do
     raise ArgumentError, "MySQL does not support :read_after_writes in schemas for non-primary keys. " <>
                          "The following fields in #{inspect schema} are tagged as such: #{inspect returning}"
+  end
+
+  @doc false
+  def structure_dump(config) do
+    run_with_cmd("mysqldump", config, ["--no-data", "--routines", config[:database]])
+  end
+
+  @doc false
+  def structure_load(config, path) do
+    args = [
+      "--execute", "SET FOREIGN_KEY_CHECKS = 0; SOURCE #{path}; SET FOREIGN_KEY_CHECKS = 1",
+      "--database", config[:database]
+    ]
+    run_with_cmd("mysql", config, args)
   end
 end
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -136,8 +136,59 @@ defmodule Ecto.Adapters.Postgres do
     end
   end
 
+  defp run_with_psql(sql_command, opts) do
+    args = ["--quiet",
+            "--set", "ON_ERROR_STOP=1",
+            "--set", "VERBOSITY=verbose",
+            "--no-psqlrc",
+            "--dbname", "template1",
+            "--command", sql_command]
+    run_with_cmd("psql", database, args)
+  end
+
+  defp run_with_cmd(cmd, opts, opt_args) do
+    unless System.find_executable(cmd) do
+      raise "could not find executable `#{cmd}` in path, " <>
+            "please guarantee it is available before running ecto commands"
+    end
+
+    env =
+      if password = opts[:password] do
+        [{"PGPASSWORD", password}]
+      else
+        []
+      end
+    env = [{"PGCONNECT_TIMEOUT", "10"} | env]
+
+    args = []
+
+    if username = opts[:username] do
+      args = ["-U", username|args]
+    end
+
+    if port = opts[:port] do
+      args = ["-p", to_string(port)|args]
+    end
+
+    host = opts[:hostname] || System.get_env("PGHOST") || "localhost"
+    args = ["--host", host|args]
+
+    args = args ++ opt_args
+    System.cmd(cmd, args, env: env, stderr_to_stdout: true)
+  end
+
   @doc false
   def supports_ddl_transaction? do
     true
+  end
+
+  @doc false
+  def structure_dump(config) do
+    run_with_cmd("pg_dump", config, ["--schema-only", "--no-acl", "--no-owner", config[:database]])
+  end
+
+  @doc false
+  def structure_load(config, path) do
+    run_with_cmd("psql", config, ["--quiet", "--file", path, config[:database]])
   end
 end
--- a/lib/mix/ecto.ex
+++ b/lib/mix/ecto.ex
@@ -41,6 +41,33 @@ defmodule Mix.Ecto do
     Enum.reverse(acc)
   end
 
+
+  @doc """
+  Parses the file option from the given list.
+
+  If no file option is given we assume `{priv_dir(repo)}/structure.sql`
+  """
+  def parse_file(args, repo) do
+    parse_file(args, [], repo)
+  end
+
+  defp parse_file([key, path | t], _acc, _repo) when key in ~w(--file -f) do
+    path
+  end
+
+  defp parse_file([_|t], acc, repo) do
+    parse_file t, acc, repo
+  end
+
+  defp parse_file([], [], repo) do
+    Path.relative_to(repo_priv(repo), Mix.Project.app_path)
+    |> Path.join("structure.sql")
+  end
+
+  defp parse_file([], acc, _repo) do
+    Enum.reverse(acc)
+  end
+
   @doc """
   Ensures the given module is a repository.
   """
@@ -115,6 +142,27 @@ defmodule Mix.Ecto do
   @spec migrations_path(Ecto.Repo.t) :: String.t
   def migrations_path(repo) do
     Path.join(repo_priv(repo), "migrations")
+  end
+
+  @doc """
+  Dumps the database structure
+
+  Will dump the structure of the database via the given repo's adapter
+  """
+  @spec structure_dump(Ecto.Repo.t) :: no_return
+  def structure_dump(repo) do
+    {dump, 0} = repo.__adapter__.structure_dump(repo.config)
+    dump
+  end
+
+  @doc """
+  Loads the database structure
+
+  Will load the structure into a database via the given repo's adapter
+  """
+  @spec structure_load(Ecto.Repo.t, String) :: no_return
+  def structure_load(repo, path) do
+    {_, 0} = repo.__adapter__.structure_load(repo.config, path)
   end
 
   @doc """
--- a/lib/mix/tasks/ecto.dump.ex
+++ b/lib/mix/tasks/ecto.dump.ex
@@ -1,0 +1,32 @@
+defmodule Mix.Tasks.Ecto.Dump do
+  use Mix.Task
+  import Mix.Ecto
+  import Mix.Generator
+
+  @shortdoc "Dumps the current environment's database structure into a structure file"
+
+  @moduledoc """
+  Dumps the current environment's database structure into a structure file
+
+  ## Example
+
+      mix ecto.dump
+
+  ## Command line options
+
+    * `-r`, `--repo` - the repo to load the structure info into.
+    * `-f`, `--file` - the path of the file to dump into. Will default to `{priv_dir(repo)}/structure.sql`
+  """
+
+  def run(args \\ []) do
+    no_umbrella!("ecto.dump")
+
+    [repo] = parse_repo(args)
+    ensure_repo(repo, args)
+
+    path = parse_file(args, repo)
+
+    create_directory Path.dirname(path)
+    create_file path, structure_dump(repo), force: true
+  end
+end
--- a/lib/mix/tasks/ecto.load.ex
+++ b/lib/mix/tasks/ecto.load.ex
@@ -1,0 +1,30 @@
+defmodule Mix.Tasks.Ecto.Load do
+  use Mix.Task
+  import Mix.Ecto
+
+  @shortdoc "Loads the current environment's database structure from a previously dumped structure file"
+
+  @moduledoc """
+  Loads the current environment's database structure from a previously dumped structure file
+
+  ## Example
+
+      mix ecto.load
+
+  ## Command line options
+
+    * `-r`, `--repo` - the repo to load the structure info into.
+    * `-f`, `--file` - the path of the file to load from. Will default to `{priv_dir(repo)}/structure.sql`
+  """
+
+  def run(args \\ []) do
+    no_umbrella!("ecto.load")
+
+    [repo] = parse_repo(args)
+    ensure_repo(repo, args)
+
+    path = parse_file(args, repo)
+
+    structure_load(repo, path)
+  end
+end
--- a/test/mix/tasks/ecto.load_test.exs
+++ b/test/mix/tasks/ecto.load_test.exs
@@ -1,0 +1,79 @@
+defmodule Mix.Tasks.Ecto.LoadTest do
+  use ExUnit.Case
+
+  import Support.FileHelpers
+  import Mix.Tasks.Ecto.Load, only: [run: 0, run: 1]
+
+  @tmp_path Path.join(tmp_path, inspect(Ecto.Dump))
+
+  defmodule Adapter do
+    def structure_load(_config, _structure), do: {:ok, 0}
+    defmacro __before_compile__(_), do: :ok
+  end
+
+  defmodule BadAdapter do
+    def structure_load(_config, _structure), do: {:error, 2}
+    defmacro __before_compile__(_), do: :ok
+  end
+
+  defmodule Repo do
+    def __repo__, do: true
+    def __adapter__, do: Adapter
+
+    def config do
+      [priv: "tmp/#{inspect(Ecto.Load)}/repo", otp_app: :ecto]
+    end
+  end
+
+  defmodule OtherRepo do
+    def __repo__, do: true
+    def __adapter__, do: Adapter
+
+    def config do
+      [priv: "tmp/#{inspect(Ecto.Load)}/other_repo", otp_app: :ecto]
+    end
+  end
+
+  defmodule BadRepo do
+    def __repo__, do: true
+    def __adapter__, do: BadAdapter
+
+    def config do
+      [priv: "tmp/#{inspect(Ecto.Load)}/bad_repo", otp_app: :ecto]
+    end
+  end
+
+  setup_all do
+    Mix.Project.config()
+    |> Keyword.get(:app)
+    |> Application.put_env(:app_repo, Repo)
+
+    on_exit fn ->
+      Mix.Project.config()
+      |> Keyword.get(:app)
+      |> Application.delete_env(:app_repo)
+    end
+
+    :ok
+  end
+
+  test "raises no exception when run" do
+    run()
+  end
+
+  test "raises no exception when run on a different repo" do
+    run(["-r", to_string(OtherRepo)])
+  end
+
+  test "raises no exception when run on a different structure file" do
+    filename = Path.join(@tmp_path, "repo/structure.sql")
+
+    run(["-f", filename])
+  end
+
+  test "raises an exception when the structure loading fails" do
+    assert_raise MatchError, fn ->
+      run(["-r", to_string(BadRepo)])
+    end
+  end
+end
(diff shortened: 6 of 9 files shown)


bcardarella:
@josevalim I need a review and some guidance on getting this completed.

Mysql & Postgres dumper/loader are in there but untested. The mix tasks are tested but I'm not sure if these tests are any good. It proved tricky to properly test when shelling out. When compared to ActiveRecord's schema task tests they're simply asserting that the method is called with certain arguments, not something that seems do-able in ExUnit.

bcardarella:
In addition to guidance on testing out the shell command, I think some of the workflow needs some thinking out.

There should probably be a few guards and graceful fallbacks for certain situations. For example, attempting to load a schema into a database already loaded, or a non-existent database.

bcardarella:
Currently the version constraint is an exact match which is likely too strict. I considered including a matcher type in the meta data for the schema. Maybe match on MAJOR.MINOR ? I realize that psql and mysql don't follow semver and these version constraints might not make sense. Thoughts?

josevalim:
Thanks @bcardarella. I will take a look at it, but just later. I have switched cycles to Ecto 1.0 and Elixir 1.1. Focus on Ecto 1.1 will be a couple weeks ahead now.

lib/mix/tasks/ecto.schema.load.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Mix.Tasks.Ecto.Schema.Load do
+  use Mix.Task
+  import Mix.Ecto
+  import Mix.Generator
+
+  @shortdoc "Loads the current environment's database from schemak"
```
duggiefresh: s/schemak/schema/

bcardarella:
@josevalim @chrismccord and I paired on this recently. We're looking to get some feedback on the direction of how we're doing to version/database type constraints before completing the PR. I'm happy to discuss on IRC

xtagon:
Rails has a feature where you can switch to a `structure.sql` file instead of the default `schema.rb`. Will Ecto be dumping/loading an Elixir DSL or plain SQL, and will that be configurable?

bcardarella:
@xtagon out of curiosity, what value do you find in the structure dumps?

xtagon:
@bcardarella I do a lot of PostgreSQL-specific stuff in migrations and schemas and, at least with Rails, sometimes it doesn't do quite what I meant to do in the schema.rb, but I can always count on the structure.SQL being correct.

josevalim:
We will only do the SQL specific variant in Ecto.

@bcardarella sorry it took me too long to give feedback on this. I knew we had important changes coming so I wanted to see how Ecto would look like after the work. Good news is that the work is done and we are close to Ecto 2.0.

For now, I have two comments related to this PR:

1. I don't think we should call it schema.dump or schema.load because schema already has a very distinct meaning in Ecto. I would go the following:

      * at the adapter, add `storage_dump` and `storage_load` callbacks
      * expose it as `Ecto.Storage.dump/load`
      * for the mix task, we can either go with `ecto.storage.dump/load` or straight `ecto.dump/load` (to mirror create/drop)

    Thoughts?

2. Using `psql` was very error prone and that's why Ecto 2.0 directly starts a connection with Postgres to perform operations, even on `storage_up`/`storage_down`. Can we implement load and dump using SQL instead of the command line or such is not possible?

bcardarella:
@josevalim 

> 1. rename from `schema` to `storage`

Sure, I'l rename.

> 2. don't use `psql`

I don't know for certain, I can investigate. I agree it would be better not to use the command line. Does this also apply to mysql?

bcardarella:
It appears that mysql's migration task is till using command line.

josevalim:
We haven't run into issues with MySQL, so don't worry about it. That can be either because less people are using it, so the issues did not show up yet, or because it is truly less cumbersome than psql.

bcardarella:
@josevalim It doesn't seem possible to do a true schema dump from within the connection. The closest is the `COPY` command but that is per table and doesn't give us all the meta information (like permissions) that the `pgdump` result would.

Tica2:
Possible `dump` command send to server can only store dump-file to server... But client don't know server file system structure and don't have access to it (see cloud instance).
Only valid option is local tool that know to make dump.

paulcsmith:
I'm wondering if this should be done every time the database is migrated/rolled back as well? I think it would be really easy to migrate and forget to do a dump. Maybe a mix alias would be fine, but I think it's important enough to do the dump by default. What do you think?

josevalim:
Yup. It is something we will streamline once all is good. :D

josevalim:
@bcardarella let me know when you have pushed the changes. :)

bcardarella:
@josevalim the changes are done but we still have some outstanding questions: https://github.com/elixir-lang/ecto/pull/901#issuecomment-160505500

paulcsmith:
`ecto.storage.dump` sounds to me like it is going to dump the actual contents of the database. I wonder if `ecto.dump` (as suggested [here](https://github.com/elixir-lang/ecto/pull/901#issuecomment-174443932)) might make that more clear? Especially since it mirrors ecto.create/drop as mentioned in the comment. Or maybe `ecto.dump_sql/load_sql` could work? Either way I'm excited to see this merged in :+1: :D

sorentwo:
I have to admit I'm disappointed to see a reversion to needing `psql`. Schema loading and dumping will be performed far more often than database creation, which makes it an even larger dependency on the `psql` command line utility. It seems technically possible to create a schema only dump over the connection, but looks daunting: http://doxygen.postgresql.org/pg__dump_8c_source.html

bcardarella:
@sorentwo I would also prefer to do it over the connection but I don't think it is currently possible in a guaranteeable way

bcardarella:
@josevalim the solution that @chrismccord and I came up with for the version db type and version constraint is to have that as in the actual file name. For example, on my machine the structure file will dump as `MyApp.Repo-postgres-9.4.storage`

This file name will allow us to derive the repo, db type, and version. From there there is the project can optionally opt into a version constraint pattern in their application config.

josevalim:
Do we need DB version in there? what does PG says about pg_dump and compatibility?

bcardarella:
@josevalim I think we would, for example the db dumps from postgres 9.3 -> 9.4 are not always compatible

sorentwo:
I think the version should definitely be included. Neither forward or backward compatibility are not guaranteed.

paulcsmith:
I think if the version is included, it should be major and minor version, but no patch version. It seems that's what's happening, but just wanted to add that just in case. If you include the patch version that could get pretty annoying if someone does a patch upgrade and everyone needs to update for something so minor. It would also change the filename more often than is probably necessary. Maybe that's not a big deal though.

Another thing to keep in mind with storing the version in the filename, is what happens when someone running postgres 9.3 starts the project and then someone with 9.4 comes in and generates a schema later. There would then be 2 files. 

So should the dump/load tasks fail if there are two files or if the version is out of sync? Or maybe the version is stored in the file in a comment at the top that could be parsed out? I don't know what would be best, but I think it would be good to fail fast if there is a version mismatch.

josevalim:
If the version is out of sync it probably just wouldn't find the sync file.

bcardarella:
the version constraint option that can be put into the application config allows for a customization on how restrictive you want to be. For example, by default there is no restriction. Or you can define one that is compatible with `Version.match?` https://github.com/bcardarella/ecto/blob/bc-schema-dump-and-loader/test/mix/tasks/ecto.load_test.exs#L45

bcardarella:
@paulcsmith I strongly disagree with that. There are many organizations that aren't flexible about patch level version changes to their database.

paulcsmith:
@josevalim Good point. I think this will still be messed up sometimes though :S. If someone upgrades their db and the task doesn't find the schema file some people will just create a new one on accident. I've seen this happen :P It's not a big deal, but something to be aware of. 

Storing the version in the filename could also be slightly painful because you have to remove the file and then add the new one. Bigger than that though is that the history is harder to track when you rename the file. Diffs might be messed up as well if you upgrade the db and run a migration in the same commit. Maybe this isn't a big deal, but I thought I'd mention in case it sparks some ideas. 

Maybe storing the the dump file as `storage_dump.sql` and the database/version in a separate like `storage.dump_version` would work? That way you can check them both in, you have full git history, nice diffs, and you can give nice errors like "You're running Postgres 9.4.1, but the dump file is for Postgres 0.3."  if you wanted to.

@bcardarella That makes sense. Since it's configurable I think that's awesome!

bcardarella:
@paulcsmith if you upgrade the database then you should not just rename the file. This is something we want to discourage. At that point someone should create a new dump to replace the current one. The structure that the upgraded DB is producing could differ from the previous dump file.

paulcsmith:
Yeah, they should generate a new file, I'm just saying that if the version is in the filename then you will lose the git history and it makes it harder to see what changed. By storing the dump as one file, and the version as another you keep both. 

You could still throw an error if you attempt to create or load the with a version mismatch. I think that is a good idea. Definitely not proposing that it just keeps on going even if it's mismatched :)

bcardarella:
Originally the PR was embedding this meta data at the head of the file, commented out. This presented two problems:

1. Messing with the integrity of the db dump could present some edge cases
2. We were eval'ing the meta data back in during the load task, which could be used for malicious purposes

The file name itself presented an opportunity to solve both of these issues. In addition, to get around your git issue, one could simply `diff` the two files.

paulcsmith:
Those are great points. So would storing the version in a file solve those issues? That way you get the git history with it. Also the diff would work in pull requests so when people review it they can see what changed much more easily.

If that won't work though then no worries. Just thought I'd mention it as a possibility

bcardarella:
I can't imagine that one is updating their database frequently enough where this becomes an issue. Nor would I think it would be a good idea to update the database *and* make structure changes in the same commit. (not saying it doesn't happen) Finally, as previously stated, different db versions may dump the structure in different ways. So the diff becomes meaningless.

paulcsmith:
:+1:

lib/ecto/adapters/postgres.ex
```diff
@@ -145,8 +145,67 @@ defmodule Ecto.Adapters.Postgres do
...
+  @doc false
+  def database_info do
+    { output, _ } = System.cmd("psql", ["--version"])
+    [ _, _, version ] = Regex.run(~r/^(\w+) \(\w+\) (.+)$/, output)
+    %{ type: "psql", version: version }
+  end
+
+  @doc false
+  def storage_dump(config) do
+    { output, _ } = run_with_cmd("pg_dump", config, ["-s", "-x", "-i", "-O", config[:database]])
```
derekprior: The `-i` option has been deprecated (and ignored) since PostgreSQL 8.4. It was subsequently removed in 9.5, which means this task won't run on 9.5.

derekprior: Also, what do you think about expanding these flags to their full names so it's more clear to readers what they are doing? `--schema-only`, `--no-acl`, `--no-owner`?

josevalim: Great suggestions @derekprior!

derekprior:
I tried this out on a project with 9.4.4 and then updated to 9.5.0 to see exactly what happens. The dump that was generated was indeed different in a way that would be annoying to see in diffs, though it was compatible. Under 9.5, there's even now a comment in the dump that tells you the version the database was running and the version of pg_dump.

On the surface, locking to a particular version does seem like it will make the schema file itself more interesting and less noisy in diffs and the like. Unfortunately, things aren't so straightforward.

1. Even users with the same version of postgres installed can have different extensions (e.g. postgis) enabled that can ever so slightly change the dumped schema even if you don't use postgis features at all (the extension might be enabled). This would causes breakages for users who try to load the schema without that extension, even though the extension is actually unused.

2. A large enough collection of developers seldom have the same exact version of postgres across their systems. It's my experience that people tend to have whatever was available when they first `brew install postgres` and then they update from there when they need to use a feature of a later version. This requires a lot more thought to set up. This isn't necessarily bad, but it's certainly more hoops to jump through since installing older versions isn't always quite so obvious in various package managers.

3. To my knowledge there's no great way to quickly switch running postgres versions or run multiple versions of postgres at the same time that is well documented and works across varying platforms. For most people, updating postgres is a pain. If you work on a SOA that has various databases, they might not all run the same version of postgres at the same time. The best way around this would be to use a docker setup for local development, but that still seems to be a minority of developers.

All of this is to say that I think it's difficult to get a meaningful, consistent, safe, and portable schema dump from a developer environment. I think it's important to consider the use case - is it for getting developers up and running? Or is it for production environments? If you had to set up a second production DB instance, would you use a dump generated by a development environment or would you use a dump or backup from your other production instance? 

I view the use case as being for developers. In that case I feel like being loose in the versions (maybe warning when they differ) is probably best. I think it's more likely that developers will be hindered by version locking than saved by it in some way that wouldn't have presented itself without the locking.

I like the idea of keeping a dump meta file that has the version of the database and dumper in use and letting that show up in diffs. Then it's on reviewers to decide if differences there are important.

bcardarella:
@derekprior by default this PR does not put any version constraint restrictions. You have to opt into it.

bcardarella:
>  different extensions being installed

This is true but there isn't much we can do about this.

> different versions of postgres installed for developers

The origin of this PR was meant to address this exactly. In Rails there is no restriction. Too often there are schema loads into incompatible database versions (or types) and cause larger issues. Preventing that from happening all together is the goal here. Developing and deploying on the same language version and database version is a best practice that should be highly encouraged. However, like I indicated in my previous comment this is a behavior that has to be opted into. We don't want to add this restriction by default as it will cause a lot of pain. However, for certain organizations that do see the merit in this they have that option. For other organizations that want a looser versioning scheme they are free to go as loose as they want within what is compatible with `Version.match?`

> running multiple versions of postgres

I'm sorry, I don't follow the logic here.

> which environment does one dump from?

I would always suggest dumping from whichever environment your organization would consider to be the source of truth.

derekprior:
> @derekprior by default this PR does not put any version constraint restrictions. You have to opt into it.

If that's the case, carry on. That sounds great to me. I guess I'm confused by this: https://github.com/elixir-lang/ecto/pull/901/files#diff-0ac5844cc3b4f634295277213de1ced2R140

bcardarella:
@derekprior go deeper, it will eventually hit https://github.com/elixir-lang/ecto/pull/901/files#diff-0ac5844cc3b4f634295277213de1ced2R163

derekprior:
> I'm sorry, I don't follow the logic here.

Sorry, that was unclear. What I'm trying to say is it is not uncommon for people to work on multiple projects at time that target different versions of Postgres (either patch or minor versions).

It's great to see that the locking is optional, but it appears that the file name will *always* include the version even when it's not locked? Seems like this would be a problem for a lot of people. It's actually a giant pain to install a precise patch version of Postgres via homebrew, for instance. Having to switch that version as you write migrations for various projects would be quite a bummer.

bcardarella:
@derekprior I think you're looking at this backwards. The version in the file name is not an absolute lock. It is just the version of the source of truth. It is up to you to define the version constraint type. So if you want to lock to the minor version you would just declare `~> 9.4` which will allow for `9.4.0` - `9.x.x`

derekprior:
I think maybe we're talking past each other. I can hop into IRC later tonight if you'll be around?

derekprior:
Ok, I've read through everything again and got some answers (I notce now that `load` takes a filename which alleviates some concern). I have one lingering major concern, but before I get into that...

Is `ecto.storage.dump` something that is intended to be run with every migration by every developer, like is done in Rails? Or is it something that is expected to be run from the "source of truth" at a moment when you decide to take a schema snapshot.

If it's the former then I think my earlier concerns are still valid (I can provide a clearer scenario for this). If it's the latter then I don't think my concern will matter.

lib/mix/ecto.ex
```diff
@@ -104,6 +104,87 @@ defmodule Mix.Ecto do
...
+    repo_priv(repo)
+  end
+
+  @doc """
+  Add me
+  """
+  @spec storage_filename(Ecto.Repo.t) :: String.t
+  def storage_filename(repo) do
+    database_info = repo.__adapter__.database_info
+    "#{inspect(repo)}-#{database_info.type}-#{database_info.version}.storage"
```
BlakeWilliams: Should the adapter decide what extension to use here?

lib/ecto/adapters/mysql.ex
```diff
@@ -186,4 +191,26 @@ defmodule Ecto.Adapters.MySQL do
     raise ArgumentError, "MySQL does not support :read_after_writes in schemas for non-primary keys. " <>
                          "The following fields in #{inspect schema} are tagged as such: #{inspect returning}"
   end
+
+  @doc false
+  def database_info do
+    { output, _ } = System.cmd("mysql", ["--version"])
```
lexmag: Needless spaces in pattern here and below.
Also in few other places in the PR. :bowtie:

test/mix/tasks/ecto.load_test.exs
```diff
@@ -0,0 +1,74 @@
...
+      [priv: "tmp/#{inspect(Ecto.Load)}", otp_app: :ecto]
+    end
+  end
+
+  defmodule RepoConstraint do
+    def __repo__, do: true
+    def __adapter__, do: Adapter
+
+    def config do
+      [ priv: "tmp/#{inspect(Ecto.Load)}",
```
lexmag: Misindentation.

lib/mix/ecto.ex
josevalim: I think the whole storage_filename should either be part of the adapter API (instead of `database_info`) or it should be simply transparent (i.e. the adapter only imports schema_load/schema_dump and is responsible for checking if the file exists). I am more inclined towards the second but if you prefer some other approach, then please let me know. :)

lib/mix/ecto.ex
```diff
@@ -118,6 +118,87 @@ defmodule Mix.Ecto do
...
+    adapter = repo.__adapter__
+    storage_info = extract_storage_info(adapter, filename)
+    database_info = adapter.database_info
+
+    database_match?(storage_info, database_info) && version_match?(repo, storage_info, database_info)
+  end
+  defp database_match?(storage_info, database_info) do
+    storage_info[:type] == database_info[:type]
+  end
+  defp version_match?(repo, storage_info, database_info) do
```
josevalim: :-1: for this dance. Here is another approach: we will always try to load some particular version. If the version is not supported, you can pass an explicit file to the schema.load/dump, this way we don't need to guess and it is more explicit.

lib/mix/tasks/ecto.dump.ex
```diff
@@ -0,0 +1,22 @@
...
+  import Mix.Ecto
+  import Mix.Generator
+
+  @shortdoc "Dumps the current environment's database storage to disk"
+
+  @moduledoc """
+  """
+
+  def run(args) do
+    no_umbrella!("ecto.storage.dump")
```
josevalim: "ecto.dump"

lib/mix/tasks/ecto.load.ex
```diff
@@ -0,0 +1,20 @@
...
+  import Mix.Ecto
+  import Mix.Generator
+
+  @shortdoc "Loads the current environment's database from storagek"
+
+  @moduledoc """
+  """
+
+  def run([file_name|args]) do
+    no_umbrella!("ecto.storage.dump")
```
josevalim: "ecto.load"

lib/mix/tasks/ecto.load.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Mix.Tasks.Ecto.Load do
+  use Mix.Task
+  import Mix.Ecto
+  import Mix.Generator
+
+  @shortdoc "Loads the current environment's database from storagek"
+
+  @moduledoc """
```
josevalim: Needs docs.

lib/mix/tasks/ecto.dump.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Mix.Tasks.Ecto.Dump do
+  use Mix.Task
+  import Mix.Ecto
+  import Mix.Generator
+
+  @shortdoc "Dumps the current environment's database storage to disk"
+
+  @moduledoc """
```
josevalim: Needs docs.

lib/mix/tasks/ecto.load.ex
```diff
@@ -0,0 +1,20 @@
...
+  """
+
+  def run([file_name|args]) do
```
josevalim: Is the filename required? I thought we should guess first?

test/mix/tasks/ecto.load_test.exs
```diff
@@ -0,0 +1,74 @@
...
+    def database_info, do: %{type: "foo", version: "3.0.0"}
+    defmacro __before_compile__(_), do: :ok
+  end
+
+  defmodule RepoAdapterConstraint do
+    def __repo__, do: true
+    def __adapter__, do: AdapterConstraint
+
+    def config do
+      [ priv: "tmp/#{inspect(Ecto.Load)}",
```
lexmag: Misindentation.

lib/ecto/adapters/postgres.ex
```diff
@@ -136,8 +136,67 @@ defmodule Ecto.Adapters.Postgres do
...
+
   @doc false
   def supports_ddl_transaction? do
     true
   end
+
+  @doc false
+  def database_info do
+    {output, _} = System.cmd("psql", ["--version"])
+    [ _, _, version ] = Regex.run(~r/^(\w+) \(\w+\) (.+)$/, output)
```
lexmag: Needless space after `version`.

lib/mix/ecto.ex
josevalim: Ignore me, the current approach is fine.

josevalim:
I am wondering if the current file name approach is the best way to go. schema.load and schema.dump is meant to be used by development teams, it is not meant to be used in production. Even if the whole team is not using the same PG version, it is very likely the dumps will be the same between them. If not, then they need to organize themselves to ensure this is the case. So my proposal is to drop the naming schema for now.

josevalim:
Alternatively, we can always require the name to be given on both `ecto.load` and `ecto.schema` calls. Mix aliases would make it easy to shortcut them. :)

bcardarella:
@josevalim thank you for the feedback, I'll try to address this week

bcardarella:
@josevalim ready for review, a few things:

1. I addressed most, if not all, of the comments
2. Per our IRC chat I've rolled back the version constraints in favor of letting the loader fail
3. Supports multiple repos but assumes use of the `priv_dir(repo)` instead of dumping repo name as file name
4. Testing the schema load and dump functionality in each adapter is missing. I was unsure how best to do this. I could just run the command as part of an integration test but with instances like the dumper I don't think asserting that the content makes any sense. I could dump from a db, drop that db, create it, then load to test a full cycle?

bcardarella:
Added integration tests

bcardarella:
I've normalized on "structure" to avoid confusion with Ecto's schema's

bcardarella:
@josevalim I just saw that Ecto 2.0 RC is out. Is this PR still on track for 2.0?

josevalim:
Yes, it is. :)

josevalim:
Closing in favor of the PR (which should be merged tomorrow!)

josevalim:
Oops, I closed the wrong thing. :D

josevalim:

bcardarella:
Awesome!

PR #902, opened 2015-08-23
Removes constraint assertion from "has_many assoc with constraints"

jazzyb:
The unique constraint assertion causes the Sqlite.Ecto tests to fail.  It seems unnecessary since multiple other tests assert unique constraints.  If it is actually necessary, then I can change the commit to simply tag the test with `:unique_constraint`.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -625,13 +625,8 @@ defmodule Ecto.Integration.RepoTest do
 
   test "has_many assoc with constraints" do
     author = TestRepo.insert!(%User{name: "john doe"})
-    p1 = TestRepo.insert!(%Post{title: "hello", author_id: author.id})
+    TestRepo.insert!(%Post{title: "hello", author_id: author.id})
     TestRepo.insert!(%Post{title: "world", author_id: author.id})
-
-    # Asserts that `unique_constraint` for `uuid` exists
-    assert_raise Ecto.ConstraintError, fn ->
-      TestRepo.insert!(%Post{title: "another", author_id: author.id, uuid: p1.uuid})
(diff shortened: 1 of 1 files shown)


josevalim:

sumerman:
how to make sure than, that test actually tests something (aka constraints are there)?

jazzyb:
@sumerman Unique constraints are tested elsewhere: https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs#L195

The primary assertion for this test is https://github.com/elixir-lang/ecto/blob/master/integration_test/cases/repo.exs#L639.  I believe the constraint assertion was superfluous.

sumerman:
@jazzyb It is superfluous, intentionally. This assert works as a precondition — should somebody change the other part of the code in the incompatible way, this case will fail. Without this assert there is no way to tell if test actually tests the thing, or it passes only because the assumption that was made holds no longer. Also, the last assert is not the "primary" one, the most important thing in this case is the `update!` line before. You can see the details in #896. /cc @josevalim

jazzyb:
I don't completely understand, but I may have misunderstood the purpose of the test.  If it is the case that the assertion is necessary, I'll be happy to undo the changes and tag the test with `:unique_constraint`.

sumerman:
@jazzyb The problem with the assertion is that it's not direct. The things that I want to assert are: 
- unique constraints are supported by the DB;
- particular constraint is present in the DB.

Otherwise the rest of the test makes no sense. Therefore, I think that tagging with `:unique_constraint` is a better solution.

josevalim:
Good points all. Let's go with tagging.

-- 

PR #911, opened 2015-08-30
validate_length/3 for lists

janpieper:
`validate_length/3` only supports strings at the moment, but i needed to validate the length of a list and after searching for a validator for this, i ended up extending Ecto with this functionality.

```elixir
changeset = changeset(%{"topics" => ["Politics"]}) |> validate_length(:topics, min: 1)
changeset.valid? # true
changeset.errors # []

changeset = changeset(%{"topics" => []}) |> validate_length(:topics, min: 1)
changeset.valid? # false
changeset.errors # [topics:{"should have at least %{count} items", count: 1}]

changeset = changeset(%{"topics" => ["Politics", "Economy"]}) |> validate_length(:topics, max: 1)
changeset.valid? # false
changeset.errors # [topics:{"should have at most %{count} items", count: 1}]

changeset = changeset(%{"topics" => ["Politics", "Economy"]}) |> validate_length(:topics, is: 1)
changeset.valid? # false
changeset.errors # [topics:{"should have %{count} items", count: 1}]
```

This functionality is quite useful if you have arrays in your model:

```elixir
schema "posts" do
  field :title
  field :body
  field :uuid, :binary_id
  field :decimal, :decimal
  field :upvotes, :integer, default: 0
  field :topics, {:array, :string}                 # <----
  field :published_at, Ecto.DateTime
  has_many :comments, Ecto.ChangesetTest.Comment
  has_one :comment, Ecto.ChangesetTest.Comment
end
```

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -982,17 +982,22 @@ defmodule Ecto.Changeset do
   end
 
   @doc """
-  Validates a change is a string of the given length.
+  Validates a change is a string or list of the given length.
 
   ## Options
 
-    * `:is` - the string length must be exactly this value
-    * `:min` - the string length must be greater than or equal to this value
-    * `:max` - the string length must be less than or equal to this value
+    * `:is` - the length must be exactly this value
+    * `:min` - the length must be greater than or equal to this value
+    * `:max` - the length must be less than or equal to this value
     * `:message` - the message on failure, depending on the validation, is one of:
-      * "should be %{count} characters"
-      * "should be at least %{count} characters"
-      * "should be at most %{count} characters"
+      * for strings:
+        * "should be %{count} characters"
+        * "should be at least %{count} characters"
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you. I am torn between providing a new function or reusing this one. I am inclined towards the latter just because I cannot come up with a decent name for the a new function. :)

janpieper:
@josevalim I think `validate_length` is a good name for a validator that validates the length of a list. If you want to be more descriptive, we could use `validate_list_length` instead, but in this case I would vote for renaming `validate_length` to `validate_string_length` (BC break!) to be more consistent in naming.

wojtekmach:
if you want to add a separate function, I think a decent name could be `validate_count` (as in `Enum.count/1`) and keep `validate_length` as in `String.length/1`, however I guess it won't be 100% consistent either, because there's a `Kernel.length/1` for lists

josevalim:
:heart:

PR #919, opened 2015-09-04
Allow external primary keys in nested relations

michalmuskala:
For now only the embeds tests are updated. There are other failures due to changes.

We also need a better error message then `error!` - not that helpful :smile:

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -262,19 +262,15 @@ defmodule Ecto.Changeset.Relation do
   defp map_changes([map | rest], pks, fun, current, acc, valid?, skip?) when is_map(map) do
     pk_values = get_pks(map, pks)
 
-    {model, current} =
+    {model, current, allowed_actions} =
       case Map.fetch(current, pk_values) do
         {:ok, model} ->
-          {model, Map.delete(current, pk_values)}
+          {model, Map.delete(current, pk_values), [:update, :delete]}
         :error ->
-          if Enum.all?(pk_values, &is_nil/1) do
-            {nil, current}
-          else
-            raise Ecto.UnmachedRelationError, new_value: map, old_value: Map.values(current), cardinality: :many
-          end
+          {nil, current, [:insert]}
       end
 
-    changeset = fun.(map, model)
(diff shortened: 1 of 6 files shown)


lib/ecto/changeset/relation.ex
```diff
@@ -262,19 +262,15 @@ defmodule Ecto.Changeset.Relation do
...
+    {model, current, actions} =
       case Map.fetch(current, pk_values) do
         {:ok, model} ->
-          {model, Map.delete(current, pk_values)}
+          {model, Map.delete(current, pk_values), [:update, :delete]}
         :error ->
-          if Enum.all?(pk_values, &is_nil/1) do
-            {nil, current}
-          else
-            raise Ecto.UnmachedRelationError, new_value: map, old_value: Map.values(current), cardinality: :many
```
josevalim: We can now also remove this exception from the exceptions file. :)

michalmuskala:
I've implemented the main functionality, and fixed tests.

I have two questions though:
  1. We check that the action is `insert` if the model was not related previously. Should we check that the action is `update` or `delete` if the model was related? I'm doing this currently, but I'm not sure it's the right behaviour.
  2. Any ideas about what error should we raise? I tried to come up with something, but I don't like any of my ideas :disappointed:

josevalim:
> We check that the action is insert if the model was not related previously. Should we check that the action is update or delete if the model was related? I'm doing this currently, but I'm not sure it's the right behaviour.

Didn't we do it before anyway since we could only update it? Anyway, I think this is the way to go. Better to be more restrictive, specially given we can have security issues if we are not careful.

> Any ideas about what error should we raise? I tried to come up with something, but I don't like any of my ideas

Suggestion: cannot #{action} model #{model} because it (does not exist)|(already exists) in the related assoc/embed.

So for insert, we will say we cannot insert because it already exists. For update/delete, it does not exist.

What do you think?

michalmuskala:
I adopted the error format you suggested. Should this be a regular `RuntimeError` or do we need a special one?

josevalim:
runtime error is fine for now i think. :)

michalmuskala:
Then I think this is ready.

josevalim:

josevalim:

PR #921, opened 2015-09-04
Use EEX for changeset error messages

kurtisnelson:
Right now serializing errors in JSONAPI format is painful, the detail field ends up being "must be at least %{count} characters" and there is no easy way to sub in the count value. If the error message strings instead use EEX, it is trivial to output them correctly, as shown below.

```elixir
defmodule App.ChangesetView do
  use App.Web, :view

  def render("error.json", %{changeset: changeset}) do
    errors = Enum.map(changeset.errors, fn {field, detail} ->
      %{
        source: %{ pointer: "/data/attributes/#{field}" },
        title: "Invalid Attribute",
        detail: render_detail(detail)
      }
    end)

    %{errors: errors}
  end

  def render_detail({message, values}) do
    EEx.eval_string message, values
  end

  def render_detail(message) do
    message
  end
end
```

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -134,11 +134,11 @@ defmodule Ecto.Changeset do
                         field: atom, message: error_message}
 
   @number_validators %{
-    less_than:                {&</2,  "must be less than %{count}"},
-    greater_than:             {&>/2,  "must be greater than %{count}"},
-    less_than_or_equal_to:    {&<=/2, "must be less than or equal to %{count}"},
-    greater_than_or_equal_to: {&>=/2, "must be greater than or equal to %{count}"},
-    equal_to:                 {&==/2, "must be equal to %{count}"},
+    less_than:                {&</2,  "must be less than <%= count %>"},
+    greater_than:             {&>/2,  "must be greater than <%= count %>"},
+    less_than_or_equal_to:    {&<=/2, "must be less than or equal to <%= count %>"},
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you but we should not use a *template language* for interpolation in error messages. Besides being a possible security issue in the future (if you allow those messages to come from an i18n system), it will also be slow. Please use `String.replace/3`.

josevalim:
If you want to do it programmatically:

```elixir
{message, opts} = error_message_in_a_tuple
Enum.reduce opts, message, fn {k, v}, acc ->
  String.replace(acc, "%{#{k}}", to_string(v))
end
```

It is only 4 LOC. :)

michalmuskala:
@kurtisnelson You can also look at the `Poison.Encoder` protocol implementation for `Ecto.Changeset` in `phoenix_ecto`: https://github.com/phoenixframework/phoenix_ecto/blob/master/lib/phoenix_ecto/json.ex#L2-L45

wojtekmach:
An alternative approach would be to expose a public API in Ecto to do the interpolation, perhaps something like this?

```elixir
  test "error_messages" do
    changeset =
      changeset(%{"title" => "hello"})
      |> validate_length(:title, max: 3)

    assert changeset.errors == [title: {"should be at most %{count} characters", count: 3}]
    assert Ecto.Changeset.error_messages(changeset) == [title: "should be at most 3 characters"]
  end
```

If this seems like a good idea I can open up a PR for this

josevalim:
I want to wait with that until we figure out how I18n is going to work out. :)

PR #933, opened 2015-09-08
Set recursive true on the `ecto.*` mix tasks.

jamesotron:
This allows you to run `ecto.create` for example in an umbrella app and
it will create all the repositories in the child apps.

I'm not entirely certain this is the correct way to make this work - what
happens when one of the child apps *doesn't* use ecto, for example?

If there's a better way to make this work then please let me know. I ran
into this because I'm using Phoenix inside an umbrella app.

### DIFF
--- a/lib/mix/tasks/ecto.drop.ex
+++ b/lib/mix/tasks/ecto.drop.ex
@@ -19,6 +19,8 @@ defmodule Mix.Tasks.Ecto.Drop do
 
   """
 
+  @recursive true
+
   @doc false
   def run(args) do
     repo = parse_repo(args)
--- a/lib/mix/tasks/ecto.drop.ex
+++ b/lib/mix/tasks/ecto.drop.ex
@@ -19,8 +19,6 @@ defmodule Mix.Tasks.Ecto.Drop do
 
   """
 
-  @recursive true
-
   @doc false
   def run(args) do
     repo = parse_repo(args)
--- a/lib/mix/tasks/ecto.gen.migration.ex
+++ b/lib/mix/tasks/ecto.gen.migration.ex
@@ -29,6 +29,8 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
 
   """
 
+  @recursive true
+
   @doc false
   def run(args) do
     no_umbrella!("ecto.gen.migration")
--- a/lib/mix/tasks/ecto.gen.migration.ex
+++ b/lib/mix/tasks/ecto.gen.migration.ex
@@ -29,8 +29,6 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
 
   """
 
-  @recursive true
-
   @doc false
   def run(args) do
     no_umbrella!("ecto.gen.migration")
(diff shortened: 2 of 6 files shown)


lib/mix/tasks/ecto.drop.ex
```diff
@@ -19,6 +19,8 @@ defmodule Mix.Tasks.Ecto.Drop do
 
   """
 
+  @recursive true
```
josevalim: I wouldn't add this one just because it could end-up being very, very, very dangerous.

lib/mix/tasks/ecto.gen.migration.ex
```diff
@@ -29,6 +29,8 @@ defmodule Mix.Tasks.Ecto.Gen.Migration do
 
   """
 
+  @recursive true
```
josevalim: Generators should not be recursive, I believe.

lib/mix/tasks/ecto.gen.repo.ex
```diff
@@ -26,6 +26,8 @@ defmodule Mix.Tasks.Ecto.Gen.Repo do
 
   """
 
+  @recursive true
```
josevalim: Generators should not be recursive, I believe.

lib/mix/tasks/ecto.drop.ex
jamesotron: Good call :+1:

lib/mix/tasks/ecto.gen.migration.ex
jamesotron: Yah. I shouldn't do `sed` driven development.

jamesotron:
Updated as per your comments @josevalim. Thanks.

josevalim:
Thanks @jamesotron. I also not sure myself what is the best way to proceed here. You are correct that the task will fail if it does not have a repository. Maybe the best option is to not make those recursive after all but create aliases in your app. Something like:

    aliases: ["ecto.create_all": ["ecto.create -r Foo.Repo", "ecto.create -r Bar"]]

Or just allow multiple repositories to be given:

    mix ecto.create -r Foo.Repo -r Bar.Repo

Which you can invoke manually or by creating a quick alias in your umbrella.

lib/mix/tasks/ecto.rollback.ex
```diff
@@ -40,6 +40,8 @@ defmodule Mix.Tasks.Ecto.Rollback do
 
   """
 
+  @recursive true
```
josevalim: This can be also dangerous? We want to rollback one and then we end-up rolling back all of them.

jamesotron:
Yeah. I'd really like if there was a better way to do this, but I can't quite visualise what that might be. It'd be ace if we could have task dependencies like Make or Rake, or were able to namespace tasks by their app in umbrellas. I'm running into a similar issue with `espec` also.

I'll have a go with the `aliases` idea. Thanks.

josevalim:
> Yeah. I'd really like if there was a better way to do this, but I can't quite visualise what that might be. It'd be ace if we could have task dependencies like Make or Rake, or were able to namespace tasks by their app in umbrellas. I'm running into a similar issue with espec also.

Task dependencies can be done with aliases:

    [aliases: ["ecto.reset": ["ecto.drop", "ecto.create", "ecto.migrate"]]

Regarding the namespacing, we could likely have a task that does just that: mix do --in foo ecto.drop. Or something of sorts.

PR #939, opened 2015-09-09
show re-assignment for preload in docs

mitchellhenke:
I wasn't sure if there's a better way to show it or if there's a reason the docs don't have it, but the lack of re-assignment for the models being preloaded can be misleading to those that are newer and more prone to forgetting about immutability.

http://hexdocs.pm/ecto/Ecto.Repo.html#c:update/2 shows re-assignment for update.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -317,9 +317,9 @@ defmodule Ecto.Repo do
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


josevalim:
:+1: for clarity.

mitchellhenke:
Thanks :heart:

PR #940, opened 2015-09-10
Add unique_ids function in Repo.Preloader

simonprev:
This function only calls the ids function with the same parameters but after collecting all the ids, it goes through `Enum.uniq/1` to only use unique ids in the final query.

I don’t know if a better solution would have been to only add the id if it’s not present while in the list comprehension `for id <- ...` https://github.com/elixir-lang/ecto/blob/master/lib/ecto/repo/preloader.ex#L114


josevalim:
Thank you!

Issue #943, opened 2015-09-10
Primary key checking for nested associations also forbids replacement of association changes without PKs


sumerman:
This https://github.com/elixir-lang/ecto/commit/9e0a750559338a2b90477ae60d2e8a711a9b68d3 commit introduces non-obvious behaviour change. 

In the situation like:
```
changeset = cast(changeset, ~w(nested), [], %{nested: [%{foo: 1, bar: 2}]})
... # there is no insert inside
changeset = cast(changeset, ~w(nested), [], %{nested: [%{foo: 2, bar: 1}]})
```
the last line causes very cryptic runtime error:
```
cannot insert related App.Model, because it already exists in the parent model.
```
although I would expect it to replace the previous list of related model.

josevalim:
Thank you for reporting. Can you give more information on how to reproduce it? Does your nested models have an ID in them? And, curiosity, why are you setting the same data twice?

sumerman:
I've added couple of testcases.
This line https://github.com/elixir-lang/ecto/pull/944/files#diff-0a0b8860bcb20e0d769453b48598f7dbR807
and this https://github.com/elixir-lang/ecto/pull/944/files#diff-0a0b8860bcb20e0d769453b48598f7dbR899 are failing.

>And, curiosity, why are you setting the same data twice?
We have a wizard-like UI, where we render the whole model on every page with only part of it being visible and editable.

josevalim:
Fixed on master. That commit pointed another bug in our code, which has now been fixed, thank you!

josevalim:
Please give it a try and let us know how it goes!

sumerman:
:+1: works as expected, thank you!

PR #944, opened 2015-09-10
test case to support #943

sumerman:
This is test only PR, so travis is expected to report failure.


josevalim:
Merged locally, thank you! :heart:

PR #946, opened 2015-09-11
Enable Ecto.DateTime.cast to support a unix timestamp

benwilson512:


benwilson512:
Do we want a handy way for going from an Ecto.Datetime to a unix timestamp btw?

lib/ecto/datetime.ex
```diff
@@ -376,6 +376,7 @@ defmodule Ecto.DateTime do
     * a tuple with `{{year, month, day}, {hour, min, sec}}` as integers or binaries
     * a tuple with `{{year, month, day}, {hour, min, sec, usec}}` as integers or binaries
     * an `Ecto.DateTime` struct itself
+    * a unix timestamp
```
josevalim: a unix timestamp as integer?

josevalim:
We are closing this as it assumes UTC conversion. Tks @lau

PR #947, opened 2015-09-12
Allow escaping of ? in fragment string

ctlevi:
Fragment strings can escape question mark with a backslash

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -205,8 +205,8 @@ defmodule Ecto.Query.Builder do
 
   defp split_binary(query), do: split_binary(query, "")
   defp split_binary(<<>>, consumed), do: [consumed]
-  defp split_binary(<<63, rest :: binary >>, consumed), do: [consumed | split_binary(rest, "")]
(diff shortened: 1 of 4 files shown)


lib/ecto/query/builder.ex
```diff
@@ -203,6 +203,12 @@ defmodule Ecto.Query.Builder do
     error! "`#{Macro.to_string(other)}` is not a valid query expression"
   end
 
+  defp split_binary(query), do: split_binary(query, "")
+  defp split_binary(<<>>, consumed), do: [consumed]
+  defp split_binary(<<63, rest :: binary >>, consumed), do: [consumed | split_binary(rest, "")]
```
josevalim: You can use `??` (63) and `?\\` (92) :)

josevalim:
That's awesome, I have added one tiny comment and we are ready to go!

lib/ecto/query/builder.ex
ctlevi: Very cool! I must have missed that in the elixir tutorial.

josevalim:
:heart:

PR #952, opened 2015-09-13
[WIP] Introduce on_replace :raise and :mark_as_invalid

### LINKED ISSUE #949: on_replace should allow :raise and :mark_as_invalid
The default should be `:raise` with an instructive error message, otherwise data can be removed by simply not sending it.

michalmuskala:
Closes #949

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -163,8 +163,13 @@ defmodule Ecto.Changeset.Relation do
     {:ok, Enum.map(current, &do_cast(relation, :empty, &1)), false, false}
   end
 
-  def cast(%{cardinality: :one}, nil, _current) do
-    {:ok, nil, false, false}
+  def cast(%{cardinality: :one} = relation, nil, current) do
+    case current && on_replace(relation, current) do
+      :error ->
+        :error
+      _ ->
+        {:ok, nil, false, false}
+    end
   end
 
   def cast(%{cardinality: :many} = relation, params, current) when is_map(params) do
@@ -254,16 +259,23 @@ defmodule Ecto.Changeset.Relation do
   @doc """
   TODO
   """
-  def on_replace(%{on_replace: :mark_as_invalid}, changeset_or_model) do
-    %{Changeset.change(changeset_or_model) | valid?: false}
+  def on_replace(%{on_replace: :mark_as_invalid}, _changeset_or_model) do
+    :error
   end
 
-  def on_replace(%{on_replace: :raise}, _changeset_or_model) do
-    raise "Nested associations and embedded models, by default, cannot be deleted " <>
-      "by simply ommiting the data or overriding it with new one. " <>
-      "If you want to enable this bahaviour, you should set the relation's " <>
-      "`on_replace` option to `:delete`. You can also consider setting the child " <>
-      "changeset's action explicitly to `:delete`"
+  def on_replace(%{on_replace: :raise, field: name, owner: owner}, _) do
+    raise """
+    you are attempting to change relation #{name}
+    of #{inspect owner}, but there is missing data.
+
+    By default, if the parent model contains N children, at least the same
+    N children must be given on update. In other words, it is not possible
+    to orphan embed nor associated records, attempting to do so results
+    in this error message.
+
+    It is possible to change this behaviour by setting :on_replace when
+    defining the relation. Check has_*/embed_* docs for more info.
+    """
   end
 
   def on_replace(%{__struct__: module} = relation, changeset_or_model) do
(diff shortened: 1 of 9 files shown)


lib/ecto/changeset/relation.ex
```diff
@@ -252,6 +251,27 @@ defmodule Ecto.Changeset.Relation do
...
+  def on_replace(%{on_replace: :mark_as_invalid}, changeset_or_model) do
+    %{Changeset.change(changeset_or_model) | valid?: false}
+  end
+
+  def on_replace(%{on_replace: :raise}, _changeset_or_model) do
+    raise "Nested associations and embedded models, by default, cannot be deleted " <>
+      "by simply ommiting the data or overriding it with new one. " <>
+      "If you want to enable this bahaviour, you should set the relation's " <>
+      "`on_replace` option to `:delete`. You can also consider setting the child " <>
+      "changeset's action explicitly to `:delete`"
```
michalmuskala: @josevalim any ideas how to improve this message?

lib/ecto/changeset/relation.ex
```diff
@@ -252,6 +251,27 @@ defmodule Ecto.Changeset.Relation do
...
 
+  @doc """
+  TODO
+  """
+  def on_replace(%{on_replace: :mark_as_invalid}, changeset_or_model) do
+    %{Changeset.change(changeset_or_model) | valid?: false}
+  end
+
+  def on_replace(%{on_replace: :raise}, _changeset_or_model) do
+    raise "Nested associations and embedded models, by default, cannot be deleted " <>
```
josevalim: start with lower case

josevalim: s/embedded/related

lib/ecto/changeset/relation.ex
josevalim:
```
"""
you are attempting to change relation #{...} but there is missing data.

By default, if the parent model contains N children, at least the same
N children must be given on update. In other words, it is not possible
to orphan embed nor associated records, attempting to do so results
in this error message.

It is possible to change this behaviour by setting :on_replace when
defining the relation. You may for example set it to `:delete`, to
automatically delete records when they are replaced/orphaned,
or `:nillify` to simply nilify the foreign key.

The default behaviour is `:raise`, although `:mark_as_invalid`
can be used if you want to transform this into a changeset error.
"""
```

lib/ecto/changeset/relation.ex
```diff
@@ -252,6 +251,27 @@ defmodule Ecto.Changeset.Relation do
     Changeset.change(current, changes) |> put_new_action(:update)
   end
 
+  @doc """
+  TODO
+  """
+  def on_replace(%{on_replace: :mark_as_invalid}, changeset_or_model) do
+    %{Changeset.change(changeset_or_model) | valid?: false}
```
josevalim: I don't think we should do this. We probably should mark the whole field as invalid by triggering this clause: https://github.com/elixir-lang/ecto/blob/master/lib/ecto/changeset.ex#L424. This way, we will at least show an error message to the user instead of silently failing.

lib/ecto/changeset/relation.ex
josevalim:
```
"""
you are attempting to change relation #{...} but there is missing data.

By default, if the parent model contains N children, at least the same
N children must be given on update. In other words, it is not possible
to orphan embed nor associated records, attempting to do so results
in this error message.

It is possible to change this behaviour by setting :on_replace when
defining the relation. Check has_*/embed_* docs for more info.
"""

Comment (michalmuskala) [inline lib/ecto/changeset/relation.ex]: I really like the second one :+1:

File: lib/ecto/changeset/relation.ex (hunk 3 above)
Comment (michalmuskala) [inline lib/ecto/changeset/relation.ex]: But how should it behave on `put_change` in this case?

Comment (josevalim) [inline lib/ecto/changeset/relation.ex]: We need to start marking things as invalid on put_change too.

Comment (michalmuskala) [inline lib/ecto/changeset/relation.ex]: Only with assocs/embeds or with other things too?

Comment (josevalim) [inline lib/ecto/changeset/relation.ex]: Only with those. We typically assume the other ones are valid (they will fail on dump). Hrm... now I am not so sure.

Comment (josevalim) [inline lib/ecto/changeset/relation.ex]: What do you think?

Comment (michalmuskala) [inline lib/ecto/changeset/relation.ex]: I'm not sure. On one hand it would be nice to signal this somehow. On the other hand this would be inconsistent. 

And what should be marked as invalid in case of `has_one`? The new changeset?

Comment (josevalim) [inline lib/ecto/changeset/relation.ex]: We would always add the error to the relation field in the parent changeset. I think this is the best way to go because if we support `put_change(changeset, :relation, on_replace: :mark_as_invalid)`, i would expect the field to be marked as invalid right there. So we should change how `put_change` works slightly then for relations.

File: lib/ecto/changeset/relation.ex [hunk 4]
```diff
@@ -153,16 +155,13 @@ defmodule Ecto.Changeset.Relation do
 
   Sets correct `state` on the returned changeset
   """
-  def cast(%{cardinality: :one} = relation, :empty, current) do
-    {:ok, current && do_cast(relation, :empty, current), false, false}
-  end
-
-  def cast(%{cardinality: :many} = relation, :empty, current) do
-    {:ok, Enum.map(current, &do_cast(relation, :empty, &1)), false, false}
-  end
```
michalmuskala: It seems we weren't using those two `:empty` cases.

michalmuskala:
@josevalim ready for review

Issue #953, opened 2015-09-13
Possible bug using custom `atom` type


parkerl:
This is a bit of a long story but someone asked on slack about an Ecto `atom` datatype.

 I thought I would try using a custom type. Everything was great until I tried querying. Then compilation failed with:

Here is my silly implementation: 
https://github.com/parkerl/curious_ecto/blob/possible_bug/lib/atom_type_problem.ex

```
== Compilation error on file lib/atom_type_problem.ex ==
** (Ecto.Query.CompileError) `:monkey` is not a valid query expression
    expanding macro: Ecto.Query.where/3
    lib/atom_type_problem.ex:62: Curious.TypedTable.run_example/0
    expanding macro: Ecto.Query.from/2
    lib/atom_type_problem.ex:62: Curious.TypedTable.run_example/0
```

I managed to track down the issue.

`Ecto.Query.Builder.Filter.build/5 ` call `Builder.escape/5` with `expr` which has the value:

```
{:==, [line: 2],
 [{{:., [line: 2], [{:t, [line: 2], nil}, :atom_type]}, [line: 2], []},
  :monkey]}
```

from the query

```
Curious.Repo.all( from t in Curious.TypedTable, where: (t.atom_type == :monkey))
```

This matches in `Builder` around line 132

```
  # comparison operators
  def escape({comp_op, _, [left, right]} = expr, type, params, vars, env) when comp_op in ~w(== != < > <= >=)a do
```

Around line 144 is where things go wrong, `right` is the atom `:monkey`.
```
    {right, params} = escape(right, rtype, params, vars, env)
```

There is no match among the literals that start around line 116 and so we fail out to line ~205

```
  # Everything else is not allowed
  def escape(other, _type, _params, _vars, _env) do
``

I have fixed this by adding:

```
  def escape(atom, type, params, vars, _env) when is_atom(atom),
    do: {literal(atom, type, vars), params}
```

I will submit a PR shortly but maybe I'm too far out on the edge to consider merging this?

PR #954, opened 2015-09-13
Fixes atom use in queries

parkerl:
Related to #953 I have added support for atoms. Not sure if this is a terrible idea of if I have implemented it safely. Also, the test for `escape` are a bit opaque. Halp!?


josevalim:
Thank you but if people want to have atoms that become strings behind the scenes, or even an integer, reproducing enum types, they can interpolate the atom: `^:foo`. I wouldn't have it as a literal or as part of the query syntax exactly because it is not supported by any type by default. :)

josevalim:
Other than that, your patch was perfect. :)

parkerl:
@josevalim Thanks for commenting!

PR #958, opened 2015-09-14
Fix sbroker matching a client in conn_exit

MSch:
There are not tests for that branch and I'm not sure how to add some. Is it even worth it? /cc @fishcakez


fishcakez:
Ah! Great catch @msch! The fix is perfect. This branch exists so that blocking to start the connection process does not block the process that just checked out the connection.

It should be possible to force this branch by using `sys:suspend` on a worker in a pool of 1 worker then kill the connection, then checkout the process, then `sys:resume`. I can add a test if you like.

MSch:
Thanks! I'd appreciate it.

fishcakez:
Test added in https://github.com/elixir-lang/ecto/commit/38af7abeaea10d88b72712c647a8c5c000cee521

PR #960, opened 2015-09-14
Add warnings and clarifications to various Time and Date function docs.

lau:


lib/ecto/datetime.ex
```diff
@@ -167,6 +167,9 @@ defmodule Ecto.Date do
 
   @doc """
   Returns an `Ecto.Date` in local time.
+
+  WARNING: Using the local time of the server will often lead to
+  intermittent bugs. Please use the `utc/0` function instead.
```
josevalim: Can we expand on why it leads to bugs? Maybe at least say that local time may go back and forth in time and therefore it is not always increasing, due to DST changes?

lau: Yeah, I could expand on it. Maybe I'll do another PR sometime. I will have to phrase it in a good way without it becoming a huge wall of text.

josevalim: I have already pushed some changes, please let me know what you think.

lau: DST is one reason not to use it. But it is more than that: Why would someone use local time instead of UTC? Because they are running a server that is not running UTC. That is a bad idea in the first place.

Apart from that: The local time relies on a magic global mutable variable which is the "local" time zone setting. If you run the same software on another server (it could be a VM) with a different time zone setting you can get different results. There are countries that today do not observe DST, but still their UTC offset differ.

When you do need to know about local times, you can take the UTC timestamp and use tzdata to convert it to any timezone by passing an explicit identifier. And in case of repeated time due to DST or time zone changes, you also know the offset so that you can convert back to UTC.

To sum up some problems with just asking for local time:
- Implicitly assuming that the server is set to something it should not be in the first place (not UTC). 
- Implicitly assuming that the mutable global variable of "local time" is set to something you want but have not specified. And that this setting never changes.
- When time time is ambiguous (e.g. DST autumn or timezone changes) you do not have the UTC offset to find out if the time you got was the first occurrence or the repetition.

TLDR: Local times requires bad server settings in order to get less valuable data (compared to UTC) and you cannot really be sure which timezone they are based on.

Sometimes people do need to compare with or convert to a specified local time. Not "local time" as in the server's local time setting, but rather an explicit timezone somewhere in the world. The solution here is to use UTC in combination with the tz database.

josevalim: Beautiful!

PR #964, opened 2015-09-15
Underscores the filename when generating a migration file using Mix

### LINKED ISSUE #963: mix ecto.gen.migration should generate consistent filenames
When generating migrations with `mix ecto.gen.migration` you can get filename inconsistencies:

`mix ecto.gen.migration CreateFooForBar` # CreateFooForBar.exs
`mix ecto.gen.migration create_foo_for_bar` # create_foo_for_bar.exs

I think that we should downcase and underscore the filename regardless of how the user typed the name for the migration.

kyledecot:
Fixes #963


josevalim:

Issue #976, opened 2015-09-22
--quiet for mix ecto.rollback


Tica2:
`mix ecto.migrate --quiet`

Is quiet
But

`mix ecto.rollback  --all --quiet`

In not

For testing case I use
```elixir
Mix.Task.run "ecto.rollback", ["--all", "--quiet"]
Mix.Task.run "ecto.migrate", ["--quiet"]
```
Insted:
```elixir
Mix.Task.run "ecto.create", ["--quiet"]
Mix.Task.run "ecto.migrate", ["--quiet"]
```

Log:

> mix ecto.rollback  --all --quiet
12:07:33.943 [info]  == Running App.Repo.Migrations.IndexUsers.down/0 forward
12:07:33.946 [info]  execute "DROP INDEX IF EXISTS users_username_lower_unique"
12:07:33.947 [info]  drop index if exists users_username_index
12:07:33.950 [info]  == Migrated in 0.0s
12:07:33.986 [info]  == Running App.Repo.Migrations.TableUsers.down/0 forward
12:07:33.986 [info]  drop table if exists users
12:07:33.987 [info]  == Migrated in 0.0s
12:07:34.020 [info]  == Running App.Repo.Migrations.General.down/0 forward
12:07:34.021 [info]  execute "DROP TABLE IF EXISTS version"
12:07:34.021 [info]  == Migrated in 0.0s

Thanks,
@Tica2

PR #977, opened 2015-09-22
Ensures that rolling back can be quiet

kyledecot:
This is in regards to #976 We should also be able to pass `quiet` when rolling back


josevalim:
:heart:

Tica2:
:+1:

PR #982, opened 2015-09-27
Changed to_binary to to_string in query module docs

JustMikey:
It looks like `to_binary` method doesn't exist anymore


JustMikey:
I accidently added two commits here :(

josevalim:
Merged manually, thank you.

josevalim:
No problem, I saw that and picked only the correct one. :)

PR #983, opened 2015-09-27
Imported Ecto.Changeset to Ecto.Model

JustMikey:
In one of the previous commits `import Ecto.Changeset` and `import Ecto.Query, only: [from: 2]` were deleted and I don't know about the latter but I think the former must still be present because otherwise ecto gives an error and doesn't compile


josevalim:
You should import it explicitly, it is not Ecto's responsibility.

JustMikey:
oh ok, my mistake, I am sorry

josevalim:
no worries, thank you for the contributions!

PR #984, opened 2015-09-29
make Ecto.DateTime.cast support over 6 decimal places

bgentry:
ISO8601 has no limit on the number of decimal places for the decimal fraction [1]. Some other languages (in particular Golang) serialize times with nanosecond precision by default. This results in the following behavior from Ecto:

```elixir
iex(8)> Ecto.DateTime.cast("2015-09-29T05:03:49.5470521Z") 
:error
iex(9)> Ecto.DateTime.cast("2015-09-29T05:03:49.547052Z")  
{:ok, #Ecto.DateTime<2015-09-29T05:03:49.547052Z>}
```

Rather than continuing to return an error when the input contains over 6 decimal digits, this change makes Ecto.DateTime truncate any decimal portion beyond 6 digits.

[1] https://en.wikipedia.org/wiki/ISO_8601#Times

### DIFF
--- a/lib/ecto/datetime.ex
+++ b/lib/ecto/datetime.ex
@@ -50,6 +50,8 @@ defmodule Ecto.DateTime.Utils do
   """
   def usec("." <> rest) do
     case parse(rest, "") do
+      {int, rest} when byte_size(int) > 6 and is_iso_8601(rest) ->
+        usec("." <> String.slice(int, 0, 6))
       {int, rest} when byte_size(int) in 1..6 and is_iso_8601(rest) ->
         pad = String.duplicate("0", 6 - byte_size(int))
         String.to_integer(int <> pad)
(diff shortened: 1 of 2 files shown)


lib/ecto/datetime.ex
```diff
@@ -50,6 +50,8 @@ defmodule Ecto.DateTime.Utils do
   """
   def usec("." <> rest) do
     case parse(rest, "") do
+      {int, rest} when byte_size(int) > 6 and is_iso_8601(rest) ->
+        usec("." <> String.slice(int, 0, 6))
```
bgentry: as an alternative, we could just alter the clause below to match `> 1` instead of `in 1..6`, and then take a `String.slice(int, 0, 6)` or something prior to doing any padding or `to_integer` parsing.

josevalim: No, this is great. Thank you!

josevalim:
Unfortunately I released a new Ecto version yesterday, so it may take a while for a new one. In any case, I have backported it to the v1.0 branch, so you can depend on a stable branch if you want to.

Tica2:
This is because sqlite has functions that datetime without nanosecond or with 3 miliseconds
```sql 
sqlite> select datetime('now');
2015-09-28 19:43:12
sqlite> select strftime('%Y-%m-%d %H:%M:%f', 'now');
2015-09-28 19:43:24.310
sqlite>
```

PR #987, opened 2015-10-03
Doc fix for sample code

iwarshak:
I believe the sample code is incorrect, and the required param should be ```[:body]``` and not ```[:string]```


michalmuskala:
Thank you! :heart:

PR #995, opened 2015-10-07
Raise helpful error when using wrong Ecto types

### LINKED ISSUE #994: Confusing error when using :datetime, :date, :time and :uuid as field types in schema definition
I think adding a helpful error when the type is `:date` or `:datetime`, `:uuid`, `:time` would be nice. Something like:

`(ArgumentError) invalid or unknown type :datetime for field :created_at. Maybe you meant to set the type to Ecto.DateTime?`

Here was my problem:

```elixir
defmodule MyApp.Repo.Migrations.CreateArticles do
  use Ecto.Migration

  def change do
    create table(:articles) do
      add :published_at, :datetime
    end
  end
end
```

and a model

```elixir
defmodule MyApp.Article do
  use Ecto.Model

  schema "articles" do
    # Figured this would also be `:datetime` like in the migration
    field :published_at, :datetime
  end
end
```

It seemed to me that `:datetime` would just work since that's what I used in the migration. This threw an error: `(ArgumentError) invalid or unknown type :datetime for field :published_at`. I was pretty confused by this so I tried `date_time`. Eventually I realized I needed `Ecto.DateTime`. This isn't a big deal, but I believe it can be improved. 

I can work on a PR for this if it's something you want

paulcsmith:
Right now if you use :datetime, :date, :time, or :uuid it will give an
error like this:

(ArgumentError) invalid or unknown type :datetime for field :published_at

To make it a bit easier to guide users in the right direction the
message adds:

Maybe you meant to set the type to Ecto.DateTime?

Closes #994

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -123,8 +123,8 @@ defmodule Ecto.Schema do
   `:string`               | UTF-8 encoded `string`  | "hello"
   `:binary`               | `binary`                | `<<int, int, int, ...>>`
   `{:array, inner_type}`  | `list`                  | `[value, value, value, ...]`
-  `:decimal`              | [`Decimal`](https://github.com/ericmj/decimal) | 
-  `:map`                  | `map` | 
+  `:decimal`              | [`Decimal`](https://github.com/ericmj/decimal) |
+  `:map`                  | `map` |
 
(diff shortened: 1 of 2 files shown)


josevalim:

PR #996, opened 2015-10-07
Correctly shutdown adapter application after migrating

vanstee:
Looks like there was a simple mistake when shutting down the adapter application after migrating. Both `Repo`s and `Adapter`s have an `@adapter` and the `Repo`'s was used in the migration instead of the application name in the `Adapter`, which results in the following error after migrating: 

```
** (MatchError) no match of right hand side value: {:error, {Ecto.Adapters.Postgres, {'no such file or directory', 'Elixir.Ecto.Adapters.Postgres.app'}}}
    lib/mix/tasks/ecto.migrate.ex:53: Mix.Tasks.Ecto.Migrate.run/2
    (mix) lib/mix/task.ex:291: Mix.Task.run_alias/3
    (mix) lib/mix/task.ex:243: Mix.Task.run/2
    (mix) lib/mix/task.ex:291: Mix.Task.run_alias/3
    (mix) lib/mix/task.ex:243: Mix.Task.run/2
    (mix) lib/mix/cli.ex:55: Mix.CLI.run_task/2
```

Not sure if this is the best way to grab the adapters application name, so feel free to fix this up a bit.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -118,6 +118,9 @@ defmodule Ecto.Adapters.SQL do
         :ok
       end
 
+      @doc false
+      def __adapter__, do: @adapter
+
       defoverridable [prepare: 2, execute: 6,
                       insert: 6, update: 7, delete: 5,
                       execute_ddl: 3, embed_id: 1,
--- a/lib/mix/tasks/ecto.migrate.ex
+++ b/lib/mix/tasks/ecto.migrate.ex
@@ -43,7 +43,7 @@ defmodule Mix.Tasks.Ecto.Migrate do
   @doc false
   def run(args, migrator \\ &Ecto.Migrator.run/4) do
     repo = parse_repo(args)
-    adapter = repo.__adapter__
+    adapter = repo.__adapter__.__adapter__
 
     {opts, _, _} = OptionParser.parse args,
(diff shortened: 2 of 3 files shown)


josevalim:
Thank you @vanstee! I had to go with a slightly different fix in master, can you please give it a try and let me know if it works?

vanstee:
@josevalim Thanks. You're code looks much better than mine :) But, it's actually not quite working for me.

Killing the repo process ends up killing everything due to the `start_link` I think. So, the migration runs just fine. But, if you are running a series of mix tasks, like you would with `ecto.reset` for instance, the whole mix task process just blows up after the repo is killed and the next mix task is never run (in this case the seeds file isn't run).

There are a couple of quick fixes, like calling `Process.flag(:trap_exit, true)` or maybe just using `start` instead of `start_link` in the migration task? Pretty sure you'll have a better idea of what's appropriate here.

Also here's a phoenix app to test things out with if you need a quick example app. You should see the "Done. Yay!" text printed at the end once everything is working. https://github.com/vanstee/migrate_broken

josevalim:
That's my bad. Change the reason from shutdown to normal and give it a try?
If it works, please send a PR.

-- 


vanstee:
Oh yep that totally works. PR incoming. Thanks @josevalim

PR #998, opened 2015-10-08
prefix option for migration modules

Dania02525:


josevalim:
This looks excellent. There is just one change left: we need to write some tests for the generated SQL. For postgres, we can simply pass the prefix in this same test:

https://github.com/elixir-lang/ecto/blob/master/test/ecto/adapters/postgres_test.exs#L564-L570

We need to do a similar change for MySQL too and we are good to go!

Dania02525:
I have added also the reference prefixes, since if creating prefixed tables, the references might have failed if unable to add prefixes to them as well.  Feature is documented, and added all tests for table prefixes, index prefixes, and reference prefixes to both postgres test and mysql test :-)

josevalim:

Tica2:
:+1 :+1 Checkout for testing ...

Issue #874, opened 2015-08-13
Support count with distinct (and possibly other aggregators too)


josevalim:
First we need to check how well supported in SQL databases this is.

vanstee:
Looks like all the SQL databases that Ecto has an adapter for currently support `COUNT(DISTINCT expr)`

* PostgreSQL http://www.postgresql.org/docs/9.4/static/sql-expressions.html#SYNTAX-AGGREGATES
* MySQL https://dev.mysql.com/doc/refman/5.7/en/group-by-functions.html#function_count-distinct
* MSSQL https://msdn.microsoft.com/en-us/library/ms175997.aspx
* SQLite3 https://www.sqlite.org/lang_aggfunc.html

josevalim:
Beautiful, thanks for doing the research @vanstee!

vanstee:
Figured I'd do a little bit more work on this. Looks like ecto supports distinct via fragments like this:

```
from user in User, select: count(fragment("DISTINCT ?", user.country))
```

So I'm guessing this PR is for something a little fancier to avoid any raw sql. What do you think of an API like this?

```
from user in User, select: count(distinct: user.country)
```

To keep things simple we could just require that the value of the `distinct` key has to be a field for now, but still allow anyone to build up a more complex distinct query with a fragment. While I'm at it I can apply this to the other aggregate functions as well.

parkerl:
@vanstee I like you proposed API. However I'm wondering if simply `count_distinct` would be clearer and easier to document than overloading `count`? Just a thought. 

Either way I love that you are working on adding this to the query API!

josevalim:
`count_distinct` sounds good. It increases the API surface but it is the simplest. What do you think @ericmj / @michalmuskala?

michalmuskala:
I like `count_distinct` too. It's clear what it does. It will also keep the AST for adapters clean. Something like `count(dictinct: model.field)` would add yet another case.

On the other hand this will only allow us to use `dictinct` with `count`. As @vanstee outlined above we could support `dictinct` in other aggregate functions too (I'm not sure how useful distinct is with a `sum` though). In this case I don't like adding `sum_distinct`, `min_dictinct`, etc...

josevalim:
Here is another idea:

    count(model.field, :distinct)

It is still making the API wider (`count/1` and `count/2`) but it is fairly easy to match on the AST side and does not introduce `count_distinct`, `sum_distinct` and friends.

[...]

PR #1024, opened 2015-10-16
Adds the option to do a count distinct query

parkerl:
- allows the atom ":distinct" to be passed to the count macro
  - implements count distinct for postgres and mysql
  - issue #874

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -1,7 +1,7 @@
 defmodule Ecto.Query.Builder do
   @moduledoc false
 
-  @aggregators ~w(count)a
+  @distinct ~w(count)a
 
   alias Ecto.Query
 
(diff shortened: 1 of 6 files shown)


lib/ecto/query/builder.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Ecto.Query.Builder do
   @moduledoc false
 
+  @aggregators ~w(count)a
```
josevalim: Sorry, lets call it `@distinct`.

josevalim:
@parkerl we just need docs, as you said. They can be added to Ecto.Query.API. :)

parkerl:
@josevalim like that?

parkerl:
@josevalim sorry I missed your comment earlier. Also found another bit of cruft.

josevalim:

josevalim:
Thank you for improving Ecto!

PR #1035, opened 2015-10-21
Ecto.UUID optimization

evax:
This changes the UUID representation to a struct containing both the binary and string representations.
This avoids unnecessary conversions when actively generating UUIDs (as opposed to letting the server generate them).

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -7,6 +7,7 @@ defmodule Ecto.UUID do
   """
 
   @behaviour Ecto.Type
+  defstruct [:binary, :string]
 
   @doc """
   The Ecto primitive type.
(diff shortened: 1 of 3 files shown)


michalmuskala:
I'm :-1:, because it will be a major breaking change in the API

evax:
@michalmuskala it's not really a change in the API, things like references and autogeneration will continue to work as is. The only code that will require updating is code assuming a string (e.g. logging the UUID without inspecting).

evax:
That's the motivation behind patch:
ecto_uuid

ericmj:
We cannot do this because like @michalmuskala said it's a major breaking change. It's very surprising that `Base.decode16/2` takes 15% of time, SQL generation, encoding values and building messages in the driver should take a lot more time. Please investigate why decoding takes so much time instead.

josevalim:
I just made this **80 times** faster in master and v1.0 branch, so please give it a try. :D

Also, if you find any other hotspot, please let us know. We did very few optimization passes in Ecto so I bet there are many things we could make really fast with little effort.

josevalim:
Also huge thanks to @sasa1977 for `mix profile.fprof`, it made optimizing it really easy :D

evax:
Thanks @josevalim (and @sasa1977)!

PR #1044, opened 2015-10-22
Inferred rollback for "create index" is broken for MySQL

maxim:
~~This is just a test case for now.~~  This is a test case and a possible fix. If you run the test on mysql without the fix, it fails as follows.

```
1) test correctly infers how to drop index (Ecto.Integration.MigrationTest)
     integration_test/sql/migration.exs:272
     ** (ArgumentError) MySQL adapter does not support drop if exists for index
     stacktrace:
       (ecto) lib/ecto/adapters/mysql/connection.ex:701: Ecto.Adapters.MySQL.Connection.error!/2
       (ecto) lib/ecto/adapters/mysql.ex:84: Ecto.Adapters.MySQL.execute_ddl/3
       (ecto) lib/ecto/migration/runner.ex:74: anonymous fn/2 in Ecto.Migration.Runner.flush/0
       (elixir) lib/enum.ex:1387: Enum."-reduce/3-lists^foldl/2-0-"/3
       (ecto) lib/ecto/migration/runner.ex:72: Ecto.Migration.Runner.flush/0
       (stdlib) timer.erl:181: :timer.tc/2
       (ecto) lib/ecto/migration/runner.ex:23: Ecto.Migration.Runner.run/6
       (ecto) lib/ecto/migrator.ex:113: Ecto.Migrator.attempt/6
       (ecto) lib/ecto/migrator.ex:93: anonymous fn/4 in Ecto.Migrator.do_down/4
       (ecto) lib/ecto/migrator.ex:83: Ecto.Migrator.down/4
       integration_test/sql/migration.exs:274
```

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -253,12 +253,25 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule InferredDropIndexMigration do
+    use Ecto.Migration
+
+    def change do
+      create index(:posts, [:title])
+    end
+  end
+
   import Ecto.Query, only: [from: 2]
   import Ecto.Migrator, only: [up: 4, down: 4]
 
   test "create and drop table and indexes" do
     assert :ok == up(TestRepo, 20050906120000, CreateMigration, log: false)
     assert :ok == down(TestRepo, 20050906120000, CreateMigration, log: false)
+  end
+
+  test "correctly infers how to drop index" do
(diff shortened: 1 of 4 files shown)


maxim:
Added a possible fix for this, just never use `:drop_if_exists` for indexes. I believe Rails solves this by providing an explicit `:if_exists` option. We can add this going forward, but this is probably a bad idea as a default.

maxim:
Another possible solution would be to remove "if exists" part from MySQL only, but I think the way Rails does it (see prev. comment) is a bit better.

josevalim:

PR #1045, opened 2015-10-22
Convert interpolation to concatenation in mysql driver

alirz23:


josevalim:
Why? The interpolation is alright. Unless there is a bigger reason, we should leave it as is.

alirz23:
I see the other config params are being concatenated like `database` and `extra` I figured it will make it unified but if you think its not vital thats fine

josevalim:
Ok, closing then! Thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1047, opened 2015-10-23
Migration manager

Dania02525:
Tests passing for migration manager, including stopping itself when no more migrations, and staying alive when still containing migrations.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -169,7 +169,6 @@ defmodule Ecto.Migration do
   end
 
   alias Ecto.Migration.Runner
-  alias Ecto.Migration.Manager
 
   @doc false
   defmacro __using__(_) do
@@ -661,11 +660,11 @@ defmodule Ecto.Migration do
   end
 
   def prefix(opts_prefix) when is_nil(opts_prefix) do
-    Manager.get_prefix(self)
+    Runner.prefix()
   end
 
   def prefix(opts_prefix) do
-    case {Manager.get_prefix(self), Manager.get_prefix(self) == opts_prefix} do
+    case {Runner.prefix(), Runner.prefix() == opts_prefix} do
       {nil, _} -> opts_prefix
       {prefix, true} -> prefix
       {_, false} -> raise Ecto.MigrationError, 
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -169,6 +169,7 @@ defmodule Ecto.Migration do
   end
 
   alias Ecto.Migration.Runner
+  alias Ecto.Migration.Manager
 
   @doc false
   defmacro __using__(_) do
@@ -220,7 +221,7 @@ defmodule Ecto.Migration do
   defp do_create(object, command, block) do
     quote do
       table = %Table{} = unquote(object)
-      Runner.start_command({unquote(command), table})
+      Runner.start_command({unquote(command), %{table | prefix: prefix(table.prefix)}})
 
       if table.primary_key do
         add(:id, :serial, primary_key: true)
@@ -267,7 +268,7 @@ defmodule Ecto.Migration do
 
   """
   def create(%Index{} = index) do
-    Runner.execute {:create, index}
+    Runner.execute {:create, %{index | prefix: prefix(index.prefix)}}
   end
 
   def create(%Table{} = table) do
@@ -286,7 +287,7 @@ defmodule Ecto.Migration do
 
   """
   def create_if_not_exists(%Index{} = index) do
-    Runner.execute {:create_if_not_exists, index}
+    Runner.execute {:create_if_not_exists, %{index | prefix: prefix(index.prefix)}}
   end
 
   def create_if_not_exists(%Table{} = table) do
@@ -301,7 +302,7 @@ defmodule Ecto.Migration do
         []
       end
 
-    Runner.execute {command, table, columns}
+    Runner.execute {command, %{table | prefix: prefix(table.prefix)}, columns}
   end
 
   @doc """
@@ -314,7 +315,7 @@ defmodule Ecto.Migration do
 
   """
   def drop(%{} = object) do
-    Runner.execute {:drop, object}
+    Runner.execute {:drop, %{object | prefix: prefix(object.prefix)}}
     object
   end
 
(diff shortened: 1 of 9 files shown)


lib/ecto/migrator.ex
```diff
@@ -31,10 +31,10 @@ defmodule Ecto.Migrator do
   This function ensures the migration table exists
   if no table has been defined yet.
   """
-  @spec migrated_versions(Ecto.Repo.t) :: [integer]
-  def migrated_versions(repo) do
-    SchemaMigration.ensure_schema_migrations_table!(repo)
-    SchemaMigration.migrated_versions(repo)
+  @spec migrated_versions(Ecto.Repo.t, atom) :: [integer]
+  def migrated_versions(repo, prefix) do
```
josevalim: Let's pass the prefix here as part of options to be consistent with `up` and `down`. :)

lib/ecto/migrator.ex
```diff
@@ -47,7 +47,7 @@ defmodule Ecto.Migrator do
   """
   @spec up(Ecto.Repo.t, integer, Module.t, Keyword.t) :: :ok | :already_up | no_return
   def up(repo, version, module, opts \\ []) do
-    versions = migrated_versions(repo)
+    versions = migrated_versions(repo, Keyword.get(opts, :prefix, nil))
```
josevalim: `Keyword.get(opts, :prefix, nil)` can be `opts[:prefix]` here and below (since it will return `nil` if there is no key)

lib/ecto/migration.ex
```diff
@@ -658,4 +659,17 @@ defmodule Ecto.Migration do
   defp validate_type!(%Reference{} = reference) do
     reference
   end
+
+  def prefix(opts_prefix) when is_nil(opts_prefix) do
```
josevalim: Excellent. I think this function should be private though. :)

lib/ecto/migration/runner.ex
```diff
@@ -16,7 +17,12 @@ defmodule Ecto.Migration.Runner do
   """
   def run(repo, module, direction, operation, migrator_direction, opts) do
     level = Keyword.get(opts, :log, :info)
-    start_link(repo, direction, migrator_direction, level)
+    ensure_manager_started()
+    Manager.put_migration(self(), Keyword.get(opts, :prefix, nil))
```
josevalim: I think the prefix should be stored in the runner. :) All the agent should really do is to return the runner pid from the current migration process.

lib/ecto/migration/runner.ex
```diff
@@ -16,7 +17,12 @@ defmodule Ecto.Migration.Runner do
   """
   def run(repo, module, direction, operation, migrator_direction, opts) do
     level = Keyword.get(opts, :log, :info)
-    start_link(repo, direction, migrator_direction, level)
+    ensure_manager_started()
```
josevalim: We should always start the manager when Ecto starts! This can be done in two steps:

1. Adding `mod: {Ecto, []}` to `def application` in `mix.exs` so it invokes the callback in the Ecto module

2. By adding `use Application` in Ecto and returning a supervision tree where the manager is the only child

3. From there on, we should just assume the manager is available and running :)

josevalim:
Thank you @Dania02525! I have added some comments, we are definitely moving to the proper direction!

lib/ecto/migration.ex
Dania02525: This is imported with "use" and fails in migrator_test.exs if they are private (although, they could be private aside from that.  should I remake the prefix functions within the test, or leave them public?

josevalim: Can we not call it from tests? It is an implementation detail, we can probably test this indirectly via `create table` and friends, what do you think?

lib/ecto/migration/runner.ex
Dania02525: I've got this set up in the latest commit, and it starts the manager fine when using iex -S mix.  However, in the tests, I get a flood of errors about no process, at the call to put_migration in the test setup.  The app should be up at this point, what might be the cause of this?

lib/ecto/migration.ex
Dania02525: I'm thinking it won't work in normal use either, since in the migration modules the use imports only public functions.  We would still need the prefix/1 when the create macro is expanded I think.

lib/ecto/migration/runner.ex
Dania02525: I've got this set up in the latest commit, and it starts the manager fine when using iex -S mix. However, in the tests, I get a flood of errors about no process, at the call to put_migration in the test setup. The app should be up at this point, what might be the cause of this?

lib/ecto/migration.ex
josevalim: The macro can just call the function directly in runner: `Ecto.Migration.Runner.put_prefix(...)`. For this module itself, we can import it.

Dania02525: this fails because I'm setting the table prefix inside a quoted block in migration, that then gets sent to runner (where the private prefix functions aren't available).  I still need the prefix functions in migration for simply executed commands like just create(:table), but I think I need to duplicate them in runner.  I don't think this is a testing problem, I think its a real bug in the commit :-)

josevalim: No problem then. Leave things as is, we can always find another solution later. :)

Dania02525: We can make these private if this can be improved to allow the table to be prefixed before being sent into runner:
```elixir
defmacro alter(object, do: block) do
    quote do
      table = %Table{} = unquote(object)
      Runner.start_command({:alter, %{table | prefix: prefix(table.prefix)}})
      unquote(block)
      Runner.end_command
    end
  end
```
any ideas?

Dania02525:
I'm still seeing a no process failure when running the tests at the test setup where I put_migration, but I think the supervisor tree is set up to run properly with the application, since the manager works fine when I start the app with iex -S mix.  Is the app started differently for tests?

Dania02525:
@josevalim I have this working now, My problem was that Exunit's on_exit callback is run as a separate process, therefore the self() retrieved from functions in runner it called referred to the callback pid, rather than the test pid that started the migration.  I manually stopped the runner within each test in migrations test- is there a better DRYer way?

josevalim:
This has been merged locally. Thank you @Dania02525!

josevalim:

PR #1048, opened 2015-10-23
Allows a keyword list to be passed to where

parkerl:
- keyword arguments will be joined with "and" as a where clause
 - an empty list will return an empt clause
 - Issue #969


josevalim:
Perfect! I will merge it and do some tiny improvements. Soon I will ping on the issue the next steps. :) Thank you so much @parkerl and great job!

PR #1049, opened 2015-10-24
update Ecto.Changeset.unique_contraint/3 docs

mkaszubowski:
The docs don't explain why there aren't any error messages for uniqueness validation when other validation error messages are present. This can be misleading for beginners and since it is a common feature implemented while learning, it can be useful to have it in the docs.


josevalim:

PR #1053, opened 2015-10-27
Allow interpolated keyword lists in where clauses

parkerl:
- Issue #969

### DIFF
--- a/lib/ecto/query/builder/filter.ex
+++ b/lib/ecto/query/builder/filter.ex
@@ -50,6 +50,10 @@ defmodule Ecto.Query.Builder.Filter do
   list multiple key value pairs will be joined with "and".
   """
   @spec escape(:where | :having, Macro.t, Keyword.t, Macro.Env.t) :: {Macro.t, %{}}
+  def escape(kind, {:^, _, [var]}, _, _)  when is_tuple(var) do
+    {quote(do: Ecto.Query.Builder.Filter.runtime!(unquote(kind), unquote(var))), %{}}
+  end
+
   def escape(_kind, [], _vars, _env) do
     {true, %{}}
   end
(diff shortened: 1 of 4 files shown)


lib/ecto/query/builder/filter.ex
```diff
@@ -74,5 +78,23 @@ defmodule Ecto.Query.Builder.Filter do
...
+  end
+
+  def runtime!(kind, params) when is_list(params) do
+    params
+      |> Enum.map(fn
+        {field, value} when is_atom(field) ->
+          value = %Ecto.Query.Tagged{tag: nil, type: {0, field}, value: value}
+          {:==, [], [to_field(field), value]}
+        {_, _acc} ->
+          raise(ArgumentError, message: "expected a keyword list in #{kind}, " <>
```
josevalim:
```elixir
raise ArgumentError, "expected a keyword list in #{kind}, got: `#{inspect params}`"
```

Rationale: 

1. for `ArgumentError` (and all exceptions since 1.1), we can pass a binary (and it will become message: "message")
2. No need for `Macro.to_string(params)` because now we are working with values and no AST. So inspect will do it. :)

lib/ecto/query/builder/filter.ex
```diff
@@ -74,5 +78,23 @@ defmodule Ecto.Query.Builder.Filter do
...
+    true
+  end
+
+  def runtime!(kind, params) when is_list(params) do
+    params
+      |> Enum.map(fn
+        {field, value} when is_atom(field) ->
+          value = %Ecto.Query.Tagged{tag: nil, type: {0, field}, value: value}
+          {:==, [], [to_field(field), value]}
+        {_, _acc} ->
```
josevalim: `_acc` -> `_value` or just `_`.

test/ecto/query/builder/filter_test.exs
```diff
@@ -4,6 +4,14 @@ defmodule Ecto.Query.Builder.FilterTest do
   import Ecto.Query.Builder.Filter
   doctest Ecto.Query.Builder.Filter
 
+  def strip_heredoc(string) do
```
josevalim: There is no need for strip_heredoc. Just build the parts using `<>` and we will be fine!

josevalim:
this is awesome! I have added a couple more comments and we are ready to go!

test/ecto/query/builder/filter_test.exs
parkerl: Ahh but I had so much fun figuring that out :relaxed: . Sorry I will get out of the bike shed.

parkerl:
@josevalim In the process of trying to knock the last item off your task list, using the new functionality for `get_by`, I realized I don't think I'm handling custom types correctly. I'm trying to figure it out but would appreciate a bump in the right direction.

parkerl:
I have pushed what I have so you can see the issue.

josevalim:
Merged manually as we talked on IRC, thank you so much!

PR #1056, opened 2015-10-28
Optional bindings for where, order_by, limit, offset and update

### LINKED ISSUE #1055: Remove bindings from most operations in Ecto.Query
Quoting @paulcsmith:

> To make this work well across the Query API, I think that making the binding optional for order_by, limit, where, and whatever other query functions are needed would make writing queries so much simpler.

This issue is currently assigned to him. :)

paulcsmith:
Closes #1055 

I didn't do the query functions where it seemed like a binding was required. I figured I'd get this out there for feedback. Was a bit easier than I was expecting so maybe I missed something?

Any query functions that I missed?

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -476,13 +476,25 @@ defmodule Ecto.Query do
   than one where expression, they are combined with an `and` operator. All
   where expressions have to evaluate to a boolean value.
 
+  `where` also accepts a keyword list where the field given as key is going to
+  be compared with the given value. The fields will always refer to the source
+  given in `from`.
+
   ## Keywords example
 
       from(c in City, where: c.state == "Sweden")
+      from(c in City, where: [state: "Sweden"])
+
+  It is also possible to interpolate the whole keyword list, allowing you to
+  dynamically filter the source:
+
+      filters = [state: "Sweden"]
+      from(c in City, where: ^filters)
 
   ## Expressions example
 
       City |> where([c], c.state == "Sweden")
+      City |> where(state: "Sweden")
 
   """
   defmacro where(query, binding \\ [], expr) do
@@ -505,6 +517,7 @@ defmodule Ecto.Query do
   ## Expressions example
 
       City |> order_by([c], asc: c.name, desc: c.population)
+      City |> order_by(asc: :name) # Sorts by the cities name
 
   ## Atom values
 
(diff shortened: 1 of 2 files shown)


test/ecto/query_test.exs
```diff
@@ -13,6 +13,14 @@ defmodule Ecto.QueryTest do
     end
   end
 
+  test "query functions do not require binding" do
+    from(p in "posts") |> limit(1)
+    from(p in "posts") |> order_by([asc: :title])
+    from(p in "posts") |> where(title: "foo")
+    from(p in "posts") |> offset(1)
+    from(p in "posts") |> update(set: [title: "foo"])
```
paulcsmith: I'm not sure about these tests, but I couldn't find another set of tests that looked like a good place to put them. I used the same style as other tests in this file. Is this sufficient or should I add more robust tests elsewhere? If so, where?

parkerl:
Nice!

lib/ecto/query.ex
```diff
@@ -476,16 +476,20 @@ defmodule Ecto.Query do
   than one where expression, they are combined with an `and` operator. All
   where expressions have to evaluate to a boolean value.
 
+  `where` also accepts a keyword list
```
josevalim: What about this for docs:

> `where` also accepts a keyword list where the field given as key is going to be compared with the given value. The fields will always refer to the source given in `from`.
> 
> It is also possible to interpolate the whole keyword list, allowing developers to dynamically filter the source:
>
>     filters = [state: "Sweden"]
>     from(c in City, where: ^filters)

It would be nice to add similar notes to `order_by`, which also supports dynamic fields. :)

test/ecto/query_test.exs
josevalim: Those tests are perfect. :)

lib/ecto/query.ex
paulcsmith: Yeah I like it! Thanks for the feedback

lib/ecto/query.ex
```diff
@@ -476,16 +476,28 @@ defmodule Ecto.Query do
...
   ## Keywords example
 
       from(c in City, where: c.state == "Sweden")
+      from(c in City, where: [state: "Sweden"])
+
+  It is also possible to interpolate the whole keyword list, allowing you to
+  dynamically filter the source:
+
+      filters = [state: "Sweden"]
+      from(c in City, where: ^filters)
```
paulcsmith: Documentation added. It seems the docs for `order_by` already had a note about interpolating a whole keyword list.

josevalim:

PR #1057, opened 2015-10-29
Raises the appropriate error if nil is given in a kw where

parkerl:


parkerl:
@josevalim here it is as we discussed. However, does this part make sense in this error?

> "If you want to check if a value is (not) nil, use is_nil/1 instead"

There is not way to pass `is_nil` into the kw list.

lib/ecto/query/builder/filter.ex
```diff
@@ -65,6 +65,7 @@ defmodule Ecto.Query.Builder.Filter do
   end
 
   def escape(kind, expr, vars, env) when is_list(expr) do
+    assert_no_nils!(expr, Ecto.Query.CompileError)
```
josevalim: We need to add this check inside the map_reduce below, otherwise, if it is not a keyword list, but a list, it will fail with an weird error and the last clause currently in map_reduce would never be triggered. The same for `runtime!` below. We would also avoid traversing the list one extra time. :)

josevalim:
Maybe we should clarify it a bit more? "Pass a full query expression and use is_nil/1 instead" Or something of sorts?

parkerl:
I'm working on fixing this error. Interesting it does not fail for me locally.

lib/ecto/query/builder/filter.ex
parkerl: Ha, ok I had that...should have commited it. Then I thought I should just reuse.

parkerl:
@josevalim better?

josevalim:

Issue #1070, opened 2015-11-04
Separate logging query time and loading time


paulcsmith:
Taken from the discussion here: https://groups.google.com/forum/#!topic/elixir-ecto/Px92gzBNlkg

Me:
> Recently I was messing around with a query that loaded around 10,000 records. It was a simple `Repo.all(User)` query. In the logs it said `query: 600ms`. This seemed odd to me that the query took so long to run. I tried the query in `psql` and it ran in 30ms.

> It seems like the log is counting both the query time *and* the time it took to load the records. I think it would be helpful (and more accurate) to make the query time show how long the actual query took to run and then a separate `load` time. That way it's simpler to see if it's a slow running query or something else.

Jose:
> Yes. That's because a good chunk of the processing happens in the adapter and we can't make a distinction between query time and load time. We hope to improve this in the future, James is already working on reading the socket data on the client, which should make things faster and easier to measure. Please open up an issue so we don't forget to retackle this

fishcakez:
Added in https://github.com/elixir-lang/ecto/commit/ac8c9013fb12134dd7f0795a1a05080d532e7690

paulcsmith:
Right on! Thanks @fishcakez

PR #1073, opened 2015-11-04
Allow decoding result without blocking connection

fishcakez:
Warning: WIP while working on required changes to postgrex and benchmark comparisons.

Trying again to handle decoding in the client process. See ericmj/postgrex/pull/108 for an explanation of the proposed changes to the postgresql driver.


fishcakez:
I've added a second commit that resolves #1070. Note that approximately the `decode_time` logged in this PR would be split approximately evenly between `query_time` and `decode_time` when using master.

fishcakez:
I've updated this to use postgrex 0.10.0.

josevalim:

PR #1079, opened 2015-11-12
Postgres: Match on error codes instead of localized messages

sldab:
The pattern matching on localized error messages fails in non-English setups.
Setting `VERBOSITY=verbose` makes psql output distinct error codes.
Checking on those is much more reliable.


josevalim:

PR #1081, opened 2015-11-13
Allow anonymous functions for casting associations

paulcsmith:
This allows user's to pass a custom function for casting associations.

    cast(post, params, ~w(), [{:author, &my_custom_changeset/2}])

Also clarifies documentation on what it expects the value to be. When I
read "key value" I thought it wanted a Map or Keyword list. This commit
changes it to state that it expects a 2 item tuple and adds an example
for clarity.

This can be helpful if your changesets are defined somewhere other than the model itself. It is also helpful even if it is defined on the model because it makes it very clear which function is used

    cast(post, params, ~w(), [{:author, &Author.optional_changeset/2}])

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -179,8 +179,12 @@ defmodule Ecto.Changeset.Relation do
                    &do_cast(relation, &1, &2))
   end
 
-  defp do_cast(%{related: model, on_cast: fun} = meta, params, nil) do
-    {:ok, apply(model, fun, [meta.__struct__.build(meta), params])
+  defp do_cast(%{related: model, on_cast: fun} = meta, params, nil) when is_function(fun) do
+    {:ok, fun.(meta.__struct__.build(meta), params) |> put_new_action(:insert)}
+  end
+
+  defp do_cast(%{related: model, on_cast: fun_name} = meta, params, nil) do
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset/relation.ex
```diff
@@ -179,8 +179,12 @@ defmodule Ecto.Changeset.Relation do
                    &do_cast(relation, &1, &2))
   end
 
-  defp do_cast(%{related: model, on_cast: fun} = meta, params, nil) do
-    {:ok, apply(model, fun, [meta.__struct__.build(meta), params])
+  defp do_cast(%{related: model, on_cast: fun} = meta, params, nil) when is_function(fun) do
```
paulcsmith: I wonder if it would be helpful to check that the arity is 2, and if it isn't raise a helpful message explaining what the function should look like and that the arity should be 2?

josevalim:
@paulcsmith this is perfect. However, I want to remove associations support from cast and have explicit cast_embed and cast_assoc, so I would like to punt on this until we have the newly proposed functions.

josevalim:
Actually, never mind. Let's merge this and we can port when the time comes. thank you!

paulcsmith:
You're welcome. Thanks for an awesome lib! 

BTW, I am excited about the idea of cast_embed and cast_assoc. Seems a lot simpler conceptually. Can't wait to try it out!

PR #1087, opened 2015-11-16
Remove Range.t from validate_number/3 typespec

sorentwo:
While a range would be nice, that isn't the current behavior. The `validate_number/3` function only accepts keywords and raises an error when a range is provided.

Alternatively, I'd like to explore adding range support to `validate_number/3`.


josevalim:

PR #1090, opened 2015-11-17
Increase Ecto.Time.utc/local and Ecto.DateTime.utc/local precision

sldab:
So far, Ecto.Time.utc/local and Ecto.DateTime.utc/local only provide precision up to seconds, which led to subtle bugs in my application.
This patch increases precision up to usecs while still keeping compatibility with OTP 17.

### DIFF
--- a/lib/ecto/datetime.ex
+++ b/lib/ecto/datetime.ex
@@ -339,20 +339,30 @@ defmodule Ecto.Time do
 
   @doc """
   Returns an `Ecto.Time` in local time.
+  `precision` can be `:sec` or `:usec.`
 
   WARNING: The local time is often not always increasing due
   to DST changes, which can lead to bugs. Please prefer the
-  `utc/0` function instead.
+  `utc/1` function instead.
   """
-  def local do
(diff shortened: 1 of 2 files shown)


josevalim:
> So far, Ecto.Time.utc/local and Ecto.DateTime.utc/local only provide precision up to seconds, which led to subtle bugs in my application.

Can you please expand on the subtle bugs?

Also note that our timestamps already support usec as long as you give the `usec: true` flag.

sldab:
I used the `usec: true` flag and was inserting records that store state transitions. Several states can be inserted within 1 second. Now I expected my query which checks for states with `inserted_at` greater than `Ecto.DateTime.utc` to return the latest states only, but it also returns no longer valid states.

Is there a reason for not including usec in `Ecto.DateTime.utc`?

josevalim:
It is ok. I just wouldn't make it the default because it is a breaking change. Likely make it optional.

sldab:
Does this make more sense?

josevalim:

PR #1099, opened 2015-11-26
Add support for exclusion constraints

### LINKED ISSUE #1098: Support EXCLUDE constraint in changeset
I am using the EXCLUDE constraint in Postgres and currently there is no way to check for it using an Ecto Changeset. Adding this should be trivial (I do not know if other DB also support this) since Postgrex already maps this error with the error code :exclusion_violation.

I can tackle this issue myself, but I wanted to know if this is a desired addition. The function would have the following signature

```elixir
exclude_constraint(changeset, field, opts)
```

and behave just like the unique_constraint function.

Comments?

bitgamma:
Closes #1098


josevalim:
Thank you! We also need an integration test. Please let me know if you have
any questions!

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

bitgamma:
The problem is, that as far as I know EXCLUDE is only supported by Postgres (and I implemented support only for Postgres), so the test would be adapter specific. Should I add a "extra.exs" script (or constraints.exs) under integration_test/pg to test this scenario? I will need to create a simple table with an EXCLUDE constraint and then try to violate it to write a complete test.

michalmuskala:
You could use `@tag :exclude_constraint` in the integration suite to make it easy to turn off this tests for any adapter that does not support this feature.

josevalim:
Since it is PG specific, I would just define everything inside PGs test
directory. Both the migration and the test itself. :)

-- 


bitgamma:
The tests are done. Sorry it took so long!

bitgamma:
Just wanted to check: is there anything else needed before you can merge?

lexmag:
@bitgamma please squash everything into single commit. :bowtie:

PR #1101, opened 2015-11-27
Move pattern match to function signature for consistency

beedub:
It looks like similar functions are pattern matching in the function signature

(new to elixir)


josevalim:
The reason we have kept as today is to avoid very long lines, which may be difficult to read. Also, most projects will likely not merge pull request doing style changes, so keep that in mind. In any case, if you are looking for a way to contribute, please check our issues tracker! Thank you!

beedub:
Makes sense. Thanks for the prompt reply!

Issue #1089, opened 2015-11-17
Fragment interpolation and sigil questions


parkerl:
I have been playing around with fragments after a few questions came up on Slack. Queries using `fragment` fail to compile with either of the following:

Interpolation with a variable.
```
    clause = "1 = ?"
    from p in "posts", where: fragment(^clause, 1)
```

With a sigil.
```
    from p in "posts", where: fragment(~S|1 = 1|)
```

Confusingly, an interpolated value with no variables does compile. However, neither the Postgres or MySQL adapter will allow interpolated values in fragments.

```
      clause = "1 = ?"
      from p in "posts", where: fragment(^clause)
```

Should either of the top two be supported?

josevalim:
The first should not be supported because it leads itself to SQL injection. The second could be supported but it don't see much purpose. To support it, we could simply call Macro.expand on the first argument of fragment and check if it returns a keyword list or a string. The third should raise at compile time, and implementing the previous suggestion should be enough to make it raise.

Thanks for the report! :)

parkerl:
@josevalim just to clarify should we...

- Raise a better error message for the first case? The current error is `fragment(...) expects the first argument to be a string for SQL fragments, a keyword list, or an interpolated value, got: '^clause'` which seems to imply interpolated fragments are ok. This should also take care of the confusion around the last case.

- Sigils might be too edge case. The only case I an make for them is if you need to write a regex or `like` that matches on double quotes. `fragment("ILIKE '\"'")` maybe looks better as `fragment(~S{ILIKE '"'})` and it would seems expected that a sigil should work where a string dose. 

(As a side note, some of this came up from someone trying to use a regex with a `?`. I am going to also make a PR documenting the escaping of that case in `fragment`)

josevalim:
> Raise a better error message for the first case? The current error is fragment(...) expects the first argument to be a string for SQL fragments, a keyword list, or an interpolated value, got: '^clause' which seems to imply interpolated fragments are ok. This should also take care of the confusion around the last case.

Oh, an interpolated value is OK if it is a keyword list. So if there is an interpolated value, we need to check at runtime if it is a keyword list and raise otherwise. We did a similar trick in a previous PR we were working on.

> Ignore sigils as too edge case.

It is fine to support sigils. Call `Macro.expand(first_arg, caller | env)` before the check and it should become a string just fine.

parkerl:
@josevalim great. I will get on this as soon as I can.

[...]

PR #1109, opened 2015-12-01
Raise at runtime if anything other than a keyword list is interpolated

parkerl:
Looking for some feedback. Thanks

- Issue #1089

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -65,10 +65,11 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:fragment, [], [escaped]]}, params}
   end
 
-  def escape({:fragment, _, [{:^, _, _} = expr]}, _type, params, vars, env) do
-    {escaped, params} = escape(expr, :any, params, vars, env)
-
-    {{:{}, [], [:fragment, [], [escaped]]}, params}
+  def escape({:fragment, _, [{:^, _, [var]} = _expr]}, _type, params, _vars, _env) do
+    expr = quote do
+      Ecto.Query.Builder.runtime_validate!(unquote(var))
+    end
+    {{:{}, [], [:fragment, [], [expr]]}, params}
   end
 
   def escape({:fragment, _, [query|frags]}, _type, params, vars, env) when is_binary(query) do
@@ -216,6 +217,15 @@ defmodule Ecto.Query.Builder do
   # Everything else is not allowed
   def escape(other, _type, _params, _vars, _env) do
     error! "`#{Macro.to_string(other)}` is not a valid query expression"
+  end
+
+  def runtime_validate!(kw) do
+    unless Keyword.keyword?(kw) do
+      raise ArgumentError, "to prevent sql injection, only a keyword list may be interpolated " <>
+                           "as the first argument to `fragment/1` with the `^` operator, got `#{inspect kw}`"
+    end
+
+    kw
   end
 
   defp split_binary(query), do: split_binary(query, "")
--- a/test/ecto/query_test.exs
+++ b/test/ecto/query_test.exs
@@ -266,4 +266,11 @@ defmodule Ecto.QueryTest do
     assert excluded_query.preloads == base.preloads
     assert excluded_query.assocs == base.assocs
   end
+
+  test "fragment/1 raises at runtime when interpolation is not a keyword list" do
+    assert_raise ArgumentError, ~r|only a keyword list.*1 = ?|, fn ->
+      clause = "1 = ?"
+      from p in "posts", where: fragment(^clause)
+    end
+  end
 end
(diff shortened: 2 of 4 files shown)


lib/ecto/query/builder.ex
```diff
@@ -65,10 +65,11 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:fragment, [], [escaped]]}, params}
   end
 
-  def escape({:fragment, _, [{:^, _, _} = expr]}, _type, params, vars, env) do
-    {escaped, params} = escape(expr, :any, params, vars, env)
-
-    {{:{}, [], [:fragment, [], [escaped]]}, params}
+  def escape({:fragment, _, [{:^, _, [var]} = expr]}, _type, params, vars, env) do
```
lexmag: There are several "unused variable" warnings.

lib/ecto/query/builder.ex
```diff
@@ -218,6 +219,12 @@ defmodule Ecto.Query.Builder do
     error! "`#{Macro.to_string(other)}` is not a valid query expression"
   end
 
+  def runtime_fragment!([{_,_}|_] = kw) when is_list(kw), do: kw
+  def runtime_fragment!(expr) do
```
lexmag: There is "unused variable" warning.

lib/ecto/query/builder.ex
```diff
@@ -218,6 +219,12 @@ defmodule Ecto.Query.Builder do
...
+    raise ArgumentError, "Only a keyword list may be interpolated as the first " <>
```
lexmag: Exception message should be non-capitalized and with no trailing punctuation.

test/ecto/query_test.exs
```diff
@@ -266,4 +266,13 @@ defmodule Ecto.QueryTest do
     assert excluded_query.preloads == base.preloads
     assert excluded_query.assocs == base.assocs
   end
+
+  test "fragment/1 raises  at runtime when interpolation is not a keyword list" do
+    assert_raise ArgumentError, "Only a keyword list may be interpolated as the first " <>
+                                "argument to `fragment/1` with the `^` operator.",
+    fn ->
```
lexmag: It'd be better to keep the `fn ->` on the previous line, as for now it's harder to follows `assert_raise` definition.
Maybe even better to assign message to variable before an assertion.

parkerl:
Thanks @lexmag, I cleaned that up. What do you think of the overall implementation? Right track?

lib/ecto/query/builder.ex
```diff
@@ -218,6 +219,12 @@ defmodule Ecto.Query.Builder do
     error! "`#{Macro.to_string(other)}` is not a valid query expression"
   end
 
+  def runtime_fragment!([{_,_}|_] = kw) when is_list(kw), do: kw
+  def runtime_fragment!(expr) do
+    raise ArgumentError, "to prevent sql injection, only a keyword list may be interpolated " <>
```
lexmag: I think it should rather be `error!/1` call instead of plain `raise/2`.

lib/ecto/query/builder.ex
```diff
@@ -218,6 +219,12 @@ defmodule Ecto.Query.Builder do
     error! "`#{Macro.to_string(other)}` is not a valid query expression"
   end
 
+  def runtime_fragment!([{_,_}|_] = kw) when is_list(kw), do: kw
```
lexmag: Shall we consider empty list here as input?
We could use `Keyword.keyword?/1` function, don't think we expect really long lists there.

lexmag:
@parkerl :+1:, looks like a right direction.

lib/ecto/query/builder.ex
```diff
@@ -218,6 +219,12 @@ defmodule Ecto.Query.Builder do
...
+  def runtime_fragment!(expr) do
+    raise ArgumentError, "to prevent sql injection, only a keyword list may be interpolated " <>
+                         "as the first argument to `fragment/1` with the `^` operator; `#{expr}` was given"
```
josevalim: "as the first argument to `fragment/1` with the `^` operator, got: `#{expr}`"

lib/ecto/query/builder.ex
parkerl: @lexmag I didn't use `error!/1` as it raises an `Ecto.Query.CompileError` which doesn't seem correct in this case. No?

josevalim: That's what we use in other places, so I think we are fine: https://github.com/parkerl/ecto/blob/fragment_interpolation_validation/lib/ecto/query/builder.ex#L434

lexmag: That't what I thought initially – `ArgumentError` should be in runtime, but [there is what made me think other way about](https://github.com/elixir-lang/ecto/blob/da82346505b6182ba62b7d560df30599dbe81e4c/lib/ecto/query/builder/update.ex#L154-L160).
Then we should fix those calls instead (not related to the current PR). :smile:

parkerl:
What else do I need to do on this one? I think I've handled all the issues above. Thanks.

josevalim:

PR #1122, opened 2015-12-07
Add support for exclusion constraints

### LINKED ISSUE #1098: Support EXCLUDE constraint in changeset

bitgamma:
Closes #1098, replaces #1099

### DIFF
--- a/integration_test/pg/constraint_test.exs
+++ b/integration_test/pg/constraint_test.exs
@@ -1,0 +1,69 @@
+defmodule Ecto.Integration.ConstraintTest do
+  use ExUnit.Case, async: true
+
+  alias Ecto.Integration.TestRepo
+  import Ecto.Migrator, only: [up: 4, down: 4]
+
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you, I will review this later!

josevalim:
Merged, thank you so much! Are there any other constraint we should add? check constraints maybe?

bitgamma:
Thanks for merging! I thought about CHECK constraints as well, but then I realized that they are actually better implemented using validations (which are already supported). I am not aware of cases where CHECK requires access to the db, but I might be wrong.

PR #1132, opened 2015-12-10
[WIP] add cast_assoc and cast_embed

### LINKED ISSUE #950: Add cast_embed and cast_assoc
Today, casting of relations is done via the `cast/4` function which is already too cluttered. The main issue is that the lack of options in `cast` forces us to configure things like on_cast and on_replace at the module level, which is global and poorly configurable.

The idea of this feature is to introduce specific functions for casting embed and assocs, so they can be properly customized at the changeset level.

We should allow options in `put_change` for similar reasons.

Once this is in, we should eventually deprecate passing relations to `cast/4` itself.

michalmuskala:
Closes #950


josevalim:
Merged locally, thanks!

PR #1135, opened 2015-12-10
Specs for changeset functions

alco:


lib/ecto/changeset.ex
```diff
@@ -1419,6 +1422,7 @@ defmodule Ecto.Changeset do
       |> unique_constraint(:email)
 
   """
+  @spec unique_constraint(t, atom, options) :: t
```
michalmuskala: Why use in most places `Keyword.t` and in others `options` for the options?

alco: @michalmuskala I added those specs according to the usage of `options` in the functions. Those with the type `options` rely on Access, so maps would also work. Where it says `Keyword.t`, the function uses `Keyword` for fetching the options.

josevalim: Please use `Keyword.t` throughout. Using access is an implementation detail mostly, does not imply we want to support all access based types. ;)

josevalim:
Thank you! I have added just one tiny comment!

alco:
Updated.

josevalim:

PR #1138, opened 2015-12-15
Add Ecto.Multi

### LINKED ISSUE #1114: Introducing Ecto.Multi
Ecto.Multi is a data structure that stores multiple operations to run inside a transaction. It will also help clean up many cases of `Repo.transaction` as well as help reduce the use of model callbacks.

Let's see an example:

```elixir
user_changeset = User.changeset(%User{}, params)
log_changeset = Log.changeset(%Log{}, event: "user updated")

multi =
  Ecto.Multi.new
  |> Ecto.Multi.update(:user, user_changeset)
  |> Ecto.Multi.insert(:log, log_changeset)
  |> Ecto.Multi.run(:charge_credit_card, fn changes_so_far -> {:ok, _} | {:error _} end)

case Repo.transaction(multi) do
  {:ok, %{user: user, log: log}} ->
    ...
  {:error, key_that_errored, %{user: user_changeset}} ->
    ...
end
```

All operations in `Ecto.Multi` are tagged and stored in the order they are defined. Duplicated keys should raise. Once built, the underlying `Ecto.Multi` struct can be given to `Repo.transaction` which will return `{:ok, map_of_structs}` or `{:error, key_that_errored, map_of_changesets}` where `key_that_errored` is the first key that failed the operation.

### API

* `Ecto.Multi.new()` - returns a new `Ecto.Multi` struct
* `Ecto.Multi.insert(multi, key, changeset_or_struct, opts)` - stores an insert to be performed on transaction
* `Ecto.Multi.update(multi, key, changeset, opts)` - stores an update to be performed on transaction
* `Ecto.Multi.delete(multi, key, changeset_or_struct, opts)` - stores a delete to be performed on transaction
* `Ecto.Multi.run(multi, key, fun)` and `Ecto.Multi.run(multi, key, mod, fun, args)` - stores a function that will be executed if the previous operations succeed. The function will receive a map with the result of all the previous operations (which must have succeeded up to this point). Returns `{:ok, value}` or `{:error, value}`
* `Ecto.Multi.all|update_all|delete_all(query, opts)` - query based operations
* `Ecto.Multi.to_list(multi)` - converts it to a list in the format `[user: {:update, args}, log: {:insert, args}, ...]`
* `Ecto.Multi.prepend|append(multi, other_multi)` - prepends or appends the operations in other multi

`Ecto.Repo.transaction/2` will be changed to accept `Ecto.Multi` structs with the semantics specified above. It is worth pointing out that, although `insert` and `delete` accept "structs", it always return changesets in `{:error, key_that_errored, map_of_changesets}`. The only exceptions are the `run/3` and `run/5` functions which will return the `value` in `{:ok, value}` and `{:error, value}`, or `:skipped` if the operation failed before the function was executed.

### Implementation

I was thinking Ecto.Multi could be a struct with a list, that keeps the reverse ordering of operations, and a map, which keeps the key-value mapping. It is better than a keyword list because every time we add a new operation we want to check the key is new and that would be faster against a map.

##### A: Why `Ecto.Multi` and not something like `Ecto.Transaction`?

Because `Repo.transaction/2` is potentially only one of the many functions that may use it. For example, Mongo does not provide transaction support but supports performing multiple operations at once.

### Acknowledgements

* @robconery for #1009
* @michalmuskala for https://github.com/michalmuskala/transchange and for pushing towards the idea

/cc @solnic @msch @chrismccord @jeregrine

michalmuskala:
Closes #1114

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -1,0 +1,121 @@
+defmodule Ecto.Multi do
+  alias __MODULE__
+  alias Ecto.Changeset
+
+  defstruct operations: [], names: MapSet.new
+
+  @type run :: (t, Keyword.t -> {:ok | :error, any})
+  @type operation :: Changeset.t | run
+  @type name :: atom
+  @type t :: %__MODULE__{operations: [{name, operation}]}
+
+  @spec new :: t
+  def new do
+    %Multi{}
+  end
+
+  @spec insert(t, name, Changeset.t | Model.t) :: t
+  def insert(multi, name, %Changeset{} = changeset) do
+    add_changeset(multi, :insert, name, changeset)
+  end
+
+  def insert(multi, name, struct) do
+    add_changeset(multi, :insert, name, Changeset.change(struct))
+  end
+
+  @spec update(t, name, Changeset.t) :: t
+  def update(multi, name, %Changeset{} = changeset) do
+    add_changeset(multi, :update, name, changeset)
+  end
+
+  @spec delete(t, name, Changeset.t | Model.t) :: t
+  def delete(multi, name, %Changeset{} = changeset) do
+    add_changeset(multi, :delete, name, changeset)
+  end
+
+  def delete(multi, name, struct) do
+    add_changeset(multi, :delete, name, Changeset.change(struct))
+  end
+
+  defp add_changeset(multi, action, name, changeset) do
+    add_operation(multi, name, put_action(changeset, action))
+  end
+
+  defp put_action(%{action: nil} = changeset, action) do
+    %{changeset | action: action}
+  end
+
+  defp put_action(%{action: action} = changeset, action) do
+    changeset
+  end
+
+  defp put_action(%{action: original}, action) do
+    raise ArgumentError, "you provided a changeset with an action already set " <>
+      "to #{original} when trying to #{action} it"
+  end
+
+  @spec run(t, name, run) :: t
+  def run(multi, name, run) when is_function(run, 2) do
+    add_operation(multi, name, run)
+  end
+
+  @spec run(t, name, module, function, args) :: t
+    when function: atom, args: [any]
+  def run(multi, name, mod, fun, args) do
+    add_operation(multi, name, {mod, fun, args})
+  end
+
+  defp add_operation(%Multi{operations: operations, names: names} = multi, name, operation)
+      when is_atom(name) do
+    if MapSet.member?(names, name) do
+      raise "#{name} is already a member of the Ecto.Multi: \n#{inspect multi}"
+    else
+      %{multi | operations: [{name, operation} | operations],
+                names: MapSet.put(names, name)}
+    end
+  end
+
+  @spec apply(t, Ecto.Repo.t, wrap, return, Keyword.t) ::
+      {:ok, results} | {:error, {name, errors}}
+    when results: %{name => Ecto.Schema.t | any},
+         errors: %{name => Changeset.t | :skipped | any},
+         wrap: ((() -> any), Keyword.t -> {:ok | :error, any}),
+         return: (any -> no_return)
+  def apply(%Multi{} = multi, repo, wrap, return, opts \\ []) do
+    # TODO skip transaction if changesets are invalid
+    multi.operations
+    |> Enum.reverse
+    |> apply_operations(repo, wrap, return, opts)
+  end
+
+  defp apply_operations(operations, repo, wrap, return, opts) do
+    wrap.(opts, fn ->
+      do_apply_operations(operations, repo, return, opts, %{})
+    end)
+  end
+
+  defp do_apply_operations([], _repo, _return, _opts, acc) do
+    acc
+  end
+
+  defp do_apply_operations([{name, operation} | rest], repo, return, opts, acc) do
+    case apply_operation(operation, acc, repo, opts) do
+      {:ok, value} ->
+        do_apply_operations(rest, repo, return, opts, Map.put(acc, name, value))
+      {:error, value} ->
+        return.(%{name => value})
+    end
+  end
+
+  defp apply_operation(%Changeset{action: action} = changeset, _acc, repo, opts) do
+    apply(repo, action, [changeset, opts])
+  end
+
+  defp apply_operation(run, acc, _repo, opts) when is_function(run, 2) do
+    apply(run, [acc, opts])
+  end
+
+  defp apply_operation({mod, fun, args}, acc, _repo, opts) do
+    apply(mod, fun, [acc, opts | args])
+  end
+end
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -20,7 +20,7 @@ defmodule Ecto.Multi do
   end
 
   def insert(multi, name, struct) do
-    add_changeset(multi, :insert, name, Changeset.change(struct))
+    insert(multi, name, Changeset.change(struct))
   end
 
   @spec update(t, name, Changeset.t) :: t
@@ -34,11 +34,11 @@ defmodule Ecto.Multi do
   end
 
   def delete(multi, name, struct) do
-    add_changeset(multi, :delete, name, Changeset.change(struct))
+    delete(multi, name, Changeset.change(struct))
   end
 
   defp add_changeset(multi, action, name, changeset) do
-    add_operation(multi, name, put_action(changeset, action))
+    add_operation(multi, name, {:changeset, put_action(changeset, action)})
   end
 
   defp put_action(%{action: nil} = changeset, action) do
@@ -51,18 +51,30 @@ defmodule Ecto.Multi do
 
   defp put_action(%{action: original}, action) do
     raise ArgumentError, "you provided a changeset with an action already set " <>
-      "to #{original} when trying to #{action} it"
+      "to #{inspect original} when trying to #{action} it"
   end
 
   @spec run(t, name, run) :: t
   def run(multi, name, run) when is_function(run, 2) do
-    add_operation(multi, name, run)
+    add_operation(multi, name, {:run, run})
   end
 
   @spec run(t, name, module, function, args) :: t
     when function: atom, args: [any]
   def run(multi, name, mod, fun, args) do
-    add_operation(multi, name, {mod, fun, args})
+    add_operation(multi, name, {:run, {mod, fun, args}})
+  end
+
+  @spec update_all(t, name, Ecto.Queryable.t, Keyword.t) :: t
+  def update_all(multi, name, queryable, updates) do
+    query = Ecto.Queryable.to_query(queryable)
+    add_operation(multi, name, {:update_all, query, updates})
+  end
+
+  @spec delete_all(t, name, Ecto.Queryable.t) :: t
+  def delete_all(multi, name, queryable) do
+    query = Ecto.Queryable.to_query(queryable)
+    add_operation(multi, name, {:delete_all, query})
   end
 
   defp add_operation(%Multi{operations: operations, names: names} = multi, name, operation)
@@ -91,15 +103,15 @@ defmodule Ecto.Multi do
   defp check_operations_valid(operations) do
     operations
     |> Enum.find(fn
-      {_, %{valid?: valid?}} -> not valid?
-      _                      -> false
+      {_, {:changeset, %{valid?: valid?}}} -> not valid?
+      _                                    -> false
     end)
     |> do_check_operations_valid(operations)
   end
 
   defp do_check_operations_valid(nil, operations),
     do: {:ok, operations}
-  defp do_check_operations_valid({name, changeset}, _operations),
+  defp do_check_operations_valid({name, {:changeset, changeset}}, _operations),
     do: {:error, {name, changeset, %{}}}
 
   defp apply_operations({:ok, operations}, repo, wrap, return, opts) do
@@ -125,15 +137,14 @@ defmodule Ecto.Multi do
     end
   end
 
-  defp apply_operation(%Changeset{action: action} = changeset, _acc, repo, opts) do
-    apply(repo, action, [changeset, opts])
-  end
-
-  defp apply_operation(run, acc, _repo, opts) when is_function(run, 2) do
-    apply(run, [acc, opts])
-  end
-
-  defp apply_operation({mod, fun, args}, acc, _repo, opts) do
-    apply(mod, fun, [acc, opts | args])
-  end
+  defp apply_operation({:changeset, changeset}, _acc, repo, opts),
+    do: apply(repo, changeset.action, [changeset, opts])
+  defp apply_operation({:run, {mod, fun, args}}, acc, _repo, opts),
+    do: apply(mod, fun, [acc, opts | args])
+  defp apply_operation({:run, run}, acc, _repo, opts),
+    do: apply(run, [acc, opts])
+  defp apply_operation({:update_all, query, updates}, _acc, repo, opts),
+    do: {:ok, repo.update_all(query, updates, opts)}
+  defp apply_operation({:delete_all, query}, _acc, repo, opts),
+    do: {:ok, repo.delete_all(query, opts)}
 end
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -4,13 +4,13 @@ defmodule Ecto.Multi do
 
   defstruct operations: [], names: MapSet.new
 
-  @type run :: (t, Keyword.t -> {:ok | :error, any}) | {module, atom, [any]}
-  @type operation :: {:changeset, Changeset.t} |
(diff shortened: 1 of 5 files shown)


lib/ecto/multi.ex
```diff
@@ -0,0 +1,121 @@
...
+    if MapSet.member?(names, name) do
+      raise "#{name} is already a member of the Ecto.Multi: \n#{inspect multi}"
+    else
+      %{multi | operations: [{name, operation} | operations],
+                names: MapSet.put(names, name)}
+    end
+  end
+
+  @spec apply(t, Ecto.Repo.t, wrap, return, Keyword.t) ::
+      {:ok, results} | {:error, {name, errors}}
```
josevalim: `{:error, name, errors}`?

lib/ecto/repo.ex
```diff
@@ -681,4 +681,16 @@ defmodule Ecto.Repo do
...
   """
   @callback log(Ecto.LogEntry.t) :: any
+
+  @doc false
+  def transaction(adapter, repo, opts, fun) when is_function(fun, 0) do
+    adapter.transaction(repo, opts, fun)
+  end
+
+  def transaction(adapter, repo, opts, %Ecto.Multi{} = multi) do
+    wrap   = &adapter.transaction(repo, &1, &2)
```
josevalim: `&adapter.transaction(repo, opts, &1)`. This way you don't need to pass the options to multi.

lib/ecto/multi.ex
michalmuskala: We can't really with rollback. We would need to repack the error after returning from transaction.

lib/ecto/repo.ex
michalmuskala: But I still want to pass the options to all the repo functions. So we respect something like `log: false`.

josevalim: I see, we still don't need to pass it to `adapter.transaction`, do we?

michalmuskala:
@josevalim it lacks docs, but otherwise is ready for review

lib/ecto/multi.ex
```diff
@@ -0,0 +1,186 @@
+defmodule Ecto.Multi do
+  alias __MODULE__
+  alias Ecto.Changeset
+
+  defstruct operations: [], names: MapSet.new
+
+  @type run :: (t, Keyword.t -> {:ok | :error, any}) | {module, atom, [any]}
+  @type operation :: {:changeset, Changeset.t} |
```
josevalim: Is this really public?

lib/ecto/multi.ex
```diff
@@ -0,0 +1,186 @@
...
+                     {:run, run} |
+                     {:update_all, Ecto.Query.t, Keyword.t} |
+                     {:delete_all, Ecto.Query.t}
```
josevalim: Let's make it `{:update_all, Ecto.Query.t, Keyword.t, Keyword.t}` and `{:delete_all, Ecto.Query.t, Keyword.t}` the last one being options.

josevalim: We should do the same for `{:changeset, Changeset., Keyword.t}`

lib/ecto/multi.ex
```diff
@@ -0,0 +1,186 @@
...
+  def merge(%Multi{names: names1, operations: ops1} = lhs,
+            %Multi{names: names2, operations: ops2} = rhs,
+            joiner) do
+    if MapSet.disjoint?(names1, names2) do
+      %Multi{names: MapSet.union(names1, names2),
+             operations: joiner.(ops1, ops2)}
+    else
+      common = MapSet.intersection(names1, names2) |> MapSet.to_list
+      raise """
+      When merging following Ecto.Multi:
```
josevalim: start with lower case

lib/ecto/repo.ex
```diff
@@ -681,4 +681,21 @@ defmodule Ecto.Repo do
 
   """
   @callback log(Ecto.LogEntry.t) :: any
+
+  @doc false
+  def transaction(adapter, repo, opts, fun) when is_function(fun, 0) do
```
josevalim: Let's move it to `Ecto.Repo.Queryable`.

josevalim: or Ecto.Repo.Transaction

lib/ecto/repo.ex
```diff
@@ -83,8 +83,8 @@ defmodule Ecto.Repo do
         @adapter.stop(__MODULE__, pid, timeout)
       end
 
-      def transaction(opts \\ [], fun) when is_list(opts) do
-        @adapter.transaction(__MODULE__, opts, fun)
+      def transaction(opts \\ [], fun_or_multi) when is_list(opts) do
```
josevalim: Let's deprecate opts being the first argument on all cases.

lib/ecto/multi.ex
```diff
@@ -0,0 +1,189 @@
...
+
+  defstruct operations: [], names: MapSet.new
+
+  @type run :: (t -> {:ok | :error, any}) | {module, atom, [any]}
+  @typep operation :: {:changeset, Changeset.t, Keyword.t} |
+                      {:run, run} |
+                      {:update_all, Ecto.Query.t, Keyword.t} |
+                      {:delete_all, Ecto.Query.t, Keyword.t}
+  @type name :: atom
+  @opaque t :: %__MODULE__{operations: [{name, operation}], names: MapSet.t}
```
michalmuskala: Should this be `@opaque`?

lib/ecto/repo.ex
```diff
@@ -654,7 +654,8 @@ defmodule Ecto.Repo do
       end)
 
   """
-  @callback transaction(Keyword.t, fun) :: {:ok, any} | {:error, any}
+  @callback transaction(fun | Ecto.Multi.t, Keyword.t) ::
+    {:ok, any} | {:error, any} | {:error, atom, any, %{atom => any}}
```
michalmuskala: I don't like that we can have so different arguments and return values. Should we make a different function that would run a `Multi`?

michalmuskala:
@josevalim updated

michalmuskala:
@josevalim ready for review. I don't think there's anything more to add

lib/ecto/multi.ex
josevalim: It can be public although the names field must be opaque.

josevalim: Actually, let's leave it opaque for now.

lib/ecto/multi.ex
```diff
@@ -0,0 +1,189 @@
+defmodule Ecto.Multi do
+  alias __MODULE__
+  alias Ecto.Changeset
+
+  defstruct operations: [], names: MapSet.new
+
+  @type run :: (t -> {:ok | :error, any}) | {module, atom, [any]}
+  @typep operation :: {:changeset, Changeset.t, Keyword.t} |
+                      {:run, run} |
+                      {:update_all, Ecto.Query.t, Keyword.t} |
```
josevalim: Should we support `:insert_all` now?

lib/ecto/multi.ex
```diff
@@ -0,0 +1,189 @@
...
+  end
+
+  defp put_action(%{action: original}, action) do
+    raise ArgumentError, "you provided a changeset with an action already set " <>
+      "to #{inspect original} when trying to #{action} it"
+  end
+
+  @spec run(t, name, run) :: t
+  def run(multi, name, run) when is_function(run, 1) do
+    add_operation(multi, name, {:run, run})
```
josevalim: You can add this one as: `{:run, {:erlang, :apply, [run, []]}}`

lib/ecto/multi.ex
```diff
@@ -0,0 +1,189 @@
...
+    |> Enum.reverse
+    |> Enum.map(&format_operation/1)
+  end
+
+  defp format_operation({name, {:changeset, changeset, opts}}),
+    do: {name, {changeset.action, changeset, opts}}
+  defp format_operation(other),
+    do: other
+
+  @spec apply(t, Ecto.Repo.t, wrap, return) :: {:ok, results} | {:error, {name, error, results}}
```
josevalim: This should be either private API or be moved into `Ecto.Repo.Transaction`. I prefer the latter.

lib/ecto/multi.ex
```diff
@@ -0,0 +1,189 @@
...
+                     operation) when is_atom(name) do
+    if MapSet.member?(names, name) do
+      raise "#{inspect name} is already a member of the Ecto.Multi: \n#{inspect multi}"
+    else
+      %{multi | operations: [{name, operation} | operations],
+                names: MapSet.put(names, name)}
+    end
+  end
+
+  def to_list(%Multi{operations: operations}) do
```
josevalim: Do we need this function for now?

lib/ecto/multi.ex
```diff
@@ -0,0 +1,189 @@
...
+         return: (any -> no_return)
+  def apply(%Multi{} = multi, repo, wrap, return) do
+    multi.operations
+    |> Enum.reverse
+    |> check_operations_valid
+    |> apply_operations(repo, wrap, return)
+  end
+
+  defp check_operations_valid(operations) do
+    case Enum.find(operations, &invalid_operation?/1) do
```
josevalim: `Enum.find_value(...) || {:ok, operations}`?

lib/ecto/multi.ex
josevalim: If we want to keep it in Multi, then I suggest for us to remove the wrap function, allowing anyone to call it directly. For example: `Ecto.Multi.apply(multi, repo, opts)`.

lib/ecto/repo.ex
josevalim: It is fine. Let's just make clear in the docs you get the four item tuple only if a multi is given.

josevalim:
@michalmuskala this is beautiful. I have added only a couple comments. Once we settle them and add docs, we are good to go. :)

lib/ecto/multi.ex
michalmuskala: Yes, we should. I'll add it

lib/ecto/multi.ex
michalmuskala: My use case is allowing `Mongo.Ecto` to add a custom wrapper that would run a multi without a transaction.

josevalim: So let's call it `__apply__` and add a `@doc false`. :)

lib/ecto/multi.ex
michalmuskala: It's problematic because I can't easily add an argument to the run function in that case:

```elixir
  defp apply_operation({:run, {mod, fun, args}}, acc, _repo),
    do: apply(mod, fun, [acc | args])
```

Or am I missing something? Is there a way to call a fun as mfa and add additional argument later?

lib/ecto/multi.ex
michalmuskala: :+1:

lib/ecto/multi.ex
```diff
@@ -0,0 +1,195 @@
...
+    add_operation(multi, name, {:update_all, query, updates, opts})
+  end
+
+  @spec delete_all(t, name, Ecto.Queryable.t, Keyword.t) :: t
+  def delete_all(multi, name, queryable, opts \\ []) when is_list(opts) do
+    query = Ecto.Queryable.to_query(queryable)
+    add_operation(multi, name, {:delete_all, query, opts})
+  end
+
+  defp add_operation(%Multi{operations: operations, names: names} = multi, name,
```
lexmag: I think variables extraction on its own looks be more readable:
```elixir
defp add_operation(%Multi{} = multi, name, operation) when is_atom(name) do
  %{operations: operations, names: names} = multi
  if MapSet.member?(names, name) do
  #...
```
anyway we need to assign whole struct to a variable. :bowtie:

michalmuskala: :+1:

michalmuskala:
I've added some documentation. It most probably has to be refined, but it's a good start.

lib/ecto/multi.ex
lexmag: Same can be done there: https://github.com/elixir-lang/ecto/pull/1138/files#diff-290dd885b452ddaec406658d1c820044R157.

lib/ecto/multi.ex
```diff
@@ -0,0 +1,364 @@
...
+
+      test "dry run password_reset" do
+        account = %Account{password: "letmein"}
+        multi = Service.password_reset(account, params)
+
+        assert [
+          {:account, {:update, account_changeset, []}},
+          {:log, {:insert, log_changeset, []}},
+          {:sessions, {:delete_all, query, []}}
+        ] = Ecto.Multi.to_list(multi)
```
paulcsmith: This might just be preference/bias, but I think I've typically seen assertions ordered as `actual_value == expected_value`. If that's the case, maybe the order of these arguments should be switched?

lib/ecto/multi.ex
```diff
@@ -0,0 +1,364 @@
...
+  directly modifying them is not advised.
+  `Ecto.Multi.to_list/1` returns a canonical representation of the structure
+  that can be used for introspection.
+
+  ## Changesets
+
+  If Multi contains operations that accept changesets (like `insert/4`,
+  `update/4` or `delete/4`) they will be checked before starting transaction
+  if any of them isn't invalid. In case any has errors, the transaction won't
+  be even started and the error will be immediately returned.
```
paulcsmith: "even be"

lib/ecto/multi.ex
```diff
@@ -0,0 +1,364 @@
...
+
+  ## Run
+
+  Multi allows you to run arbitrary functions as part of your transaction via
+  the `run/3` and `run/5`. Those functions will receive changes so far as the
+  first argument and have to return `{:ok, value}` or `{:error, value}` as
+  their result. Returning an error will abort any further operations and
+  make the whole multi fail.
```
paulcsmith: I think it would be great to include an example of using `run` in the example `Service` module below. What do you think?

test/ecto/multi_test.exs
```diff
@@ -0,0 +1,272 @@
...
+    changeset = Changeset.change(struct)
+    multi     =
+      Multi.new
+      |> Multi.delete(:comment, struct)
+
+    assert multi.names      == MapSet.new([:comment])
+    assert multi.operations == [{:comment, {:changeset, %{changeset | action: :delete}, []}}]
+  end
+
+  test "run with fun" do
```
paulcsmith: This is a great test name :)

lib/ecto/multi.ex
```diff
@@ -0,0 +1,364 @@
...
+  end
+
+  @doc """
+  Adds a function to run as part of the multi
+
+  The function should return either `{:ok, value}` or `{:error, value}`, and
+  receives changes so far as an argument.
+  """
+  @spec run(t, name, run) :: t
+  def run(multi, name, run) when is_function(run, 1) do
```
paulcsmith: It looks like this guard will block function capture like:

`Ecto.Multi.run(multi, :charge_stripe, &charge_client(&1, "product_id")`

If this function clause accepts a function, my thought would be that it would accept any number of args greater than 1, so that you could do function captures like the one above. One could usethe  run/5 version of this, but I usually prefer function capture since it will blow up at compile time if the function with that arity doesn't exist. What do you think?

paulcsmith:
@michalmuskala This looks so awesome. Can't wait to give it a try :)

lib/ecto/multi.ex
michalmuskala: It's using `=` to pattern match and assign parts of the value to variables to assert later. It's not possible with `==`.

Maybe an example of that later will help with why that's done.

lib/ecto/multi.ex
michalmuskala: Yeah, we probably should.

lib/ecto/multi.ex
paulcsmith: Oh my bad. I didn’t notice that it was just one `=` :) Yeah I think an example would be helpful. I was wondering where the account_changeset, log_changeset, etc. came from because I didn’t notice the pattern match.

lib/ecto/multi.ex
michalmuskala: `&charge_client(&1, "product_id")` will produce an anonymous function of arity 1 since it accepts only one additional argument. 

We'll always call this function with one argument - changes so far, so we should blow up early if given wrong functions.

paulcsmith: Oh, I didn’t realize that that’s the way function capture worked. Very cool. Thanks for the explanation :)

lib/ecto/multi.ex
michalmuskala: @paulcsmith Do you have any idea what we could do there? I'm out of ideas.

josevalim:

michalmuskala:
:tada: :balloon: :tada:

PR #1139, opened 2015-12-15
Update changeset docs to import Ecto.Changeset

drewolson:
Using Ecto.Schema in Ecto 1.1 does not import Ecto.Changeset. We now
import it manually in the examples.


josevalim:

PR #1142, opened 2015-12-15
Add validate_required/2

paulcsmith:
This is being added as discussed in the mailing list:
https://groups.google.com/forum/#!topic/elixir-ecto/KDVPSglEdVw

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1042,6 +1042,41 @@ defmodule Ecto.Changeset do
   end
 
   @doc """
+  Validates that one or more fields are present in the changeset.
+
+  If the value of a field is `nil`, an empty string `""`, or a string with only
+  white space `"   "` it will mark the changeset as invalid.
+
+  You can pass a single field name or a list of field names that are required.
+
+  ## Examples
+
+      validate_required(changeset, :title)
+      validate_required(changeset, [:title, :body])
+
+  """
+  @spec validate_required(t, list()) :: t
+  def validate_required(changeset, fields) when is_list(fields) do
+    Enum.reduce fields, changeset, fn(field, changeset) ->
+      changeset |> validate_required(field)
+    end
+  end
+
+  @spec validate_required(t, atom) :: t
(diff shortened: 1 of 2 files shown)


michalmuskala:
Maybe we should make the behaviour configurable? I saw @chrismccord proposed it:
```elixir
|> validate_required(fields, with: &(not is_nil(&1)))
```
The default behaviour would be the one implemented in the PR.

We can also make the message configurable.

paulcsmith:
That's an interesting thought. I'm leaning toward not doing that though because I think you could just write your own function fairly easily at that point. It might be nice to get a validation message for free, but I don't know if it's that big of a win. It also seems like this might be something used in multiple models, so it might be better to extract your own function and use it elsewhere anyway. Maybe that's an incorrect assumption though?

michalmuskala:
I was thinking about this a little bit more and I came to conclusion that what I want is a function similar to `validate_change` that would be always invoked, something like `validate_param` or `validate_field`. `validate_required` could be easily implemented in terms of such a function.

paulcsmith:
I like the name validate_field and I like the idea as well. Would be cool if it could handle multiple fields name as well, but maybe that's out of scope. 

Any comments on this particular PR? Should I go ahead and add a validate_field function that validate_required uses under the hood?

josevalim:
:-1: until we have at least one more function that needs it. :)

paulcsmith:
Sounds good to me! :)

paulcsmith:
Is there anything else this PR needs?

josevalim:
I haven't checked it yet but if there is anything left I will let you know @paulcsmith! I am working on the adapter/associations part for now, I will focus on changesets later. Happy holidays!

paulcsmith:
Sounds great! BTW I'll test the new associations stuff on ExMachina after the holidays. Happy holidays to you too!

erikreedstrom:
@paulcsmith It seems the PR does not include an `:message` option as other validation functions do. I believe with the addition of gettext, this should be available.

PR #1143, opened 2015-12-16
Update moduledoc for Ecto.Query: not is_nil

maxbeizer:
Tonight I was trying to write an Ecto query to find all resources where a given attribute is not nil. The [Ecto.Query docs](http://hexdocs.pm/ecto/Ecto.Query.html) state that this is a valid query:

```elixir
from p in query, where: p.published_at != nil
```
I kept getting [this error](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/builder.ex#L149-L152), which I believe came from https://github.com/elixir-lang/ecto/pull/1057 . 

This PR updates the moduledoc to use `not/1` and `is_nil/1`

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -261,7 +261,7 @@ defmodule Ecto.Query do
   require the use of `in`:
 
       def published(query) do
-        from p in query, where: p.published_at != nil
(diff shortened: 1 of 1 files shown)


josevalim:

Issue #883, opened 2015-08-15
Support partial indexing


ricn:
In ActiveRecord you can do this:

```
add_index :orders, :billed, where: "billed = false"
```

It would be nice if we could do it like this in Ecto:

```
create index(:orders, [:billed], where: "billed = false")
```

The only way I have figured out to do this in Ecto is to use the execute function:

```
execute("CREATE INDEX index_orders_on_billed_idx ON orders(billed) WHERE billed = false")
```

josevalim:
We are feature frozen for Ecto 1.0 but we could add this after v1.0 is out. The main question is: how do these options work between databases? It would be nice if we could use Ecto's query syntax here. But we probably won't be able to.

ricn:
Sqlite seems to support the same syntax that PG uses. Not sure about the other ones but I can look it up.

ricn:
Postgres, Sqlite & MSSQL have support for partial indexes (sometimes referred to as filtered indexes) & they all support it by appending a WHERE clause to the CREATE INDEX statement. 

MySQL does not seem to support partial indexes.

matthewrudy:
I had a go at this.
https://github.com/elixir-lang/ecto/compare/master...matthewrudy:create-partial-indexes

Currently just send a string into the `WHERE` clause

``` elixir
def change do
  create index(:posts, [:permalink], unique: true, where: "public IS TRUE")
end
```

I haven't tried using ecto query syntax yet, but the Postgres implementation of partial indexes supports full query syntax, so it should be useful.
http://www.postgresql.org/docs/8.0/static/indexes-partial.html

josevalim:
A PR is welcome. Let's just support passing a string for now.

matthewrudy:
@josevalim PRed https://github.com/elixir-lang/ecto/pull/1152

PR #1152, opened 2015-12-18
Add support for partial indexes

matthewrudy:
Comes from issue #883.

This PR adds support for the `:where` option.
It currently only allow a string or atom for the `:where`,
which gets added to the string automatically.

``` elixir
create index(:products, [:user_id], where: "price = 0", name: :free_products_index)
```

becomes the SQL
``` sql
CREATE INDEX "free_products_index" ON "products" ("user_id") WHERE price = 0
```

### DIFF
--- a/test/ecto/adapters/postgres_test.exs
+++ b/test/ecto/adapters/postgres_test.exs
@@ -669,6 +669,16 @@ defmodule Ecto.Adapters.PostgresTest do
            ~s|CREATE UNIQUE INDEX "posts_permalink_index" ON "posts" ("permalink")|
   end
 
+  test "create unique index with condition" do
+    create = {:create, index(:posts, [:permalink], unique: true, where: "public IS TRUE")}
+    assert SQL.execute_ddl(create) ==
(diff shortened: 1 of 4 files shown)


test/ecto/adapters/postgres_test.exs
```diff
@@ -669,6 +669,16 @@ defmodule Ecto.Adapters.PostgresTest do
            ~s|CREATE UNIQUE INDEX "posts_permalink_index" ON "posts" ("permalink")|
   end
 
+  test "create unique index with condition" do
+    create = {:create, index(:posts, [:permalink], unique: true, where: "public IS TRUE")}
+    assert SQL.execute_ddl(create) ==
+           ~s|CREATE UNIQUE INDEX "posts_permalink_index" ON "posts" ("permalink") WHERE public IS TRUE|
```
matthewrudy: note, I would normally do one case per test
but this style was copied from the `test "create index with prefix"` above

josevalim:

matthewrudy:
Thanks @josevalim

kidbombay:
Is there documentation on how to use this in a constraint on a changeset?

AndrewDryga:
@kidbombay take a look at `Ecto.Changeset.check_constraint/2`.

PR #1162, opened 2015-12-21
Make DateTime.cast support a less awkward subset of ISO 8601 datetime…

gradenb:
… format. Week dates and ordinal dates are still missing.

### DIFF
--- a/lib/ecto/date_time.ex
+++ b/lib/ecto/date_time.ex
@@ -62,6 +62,14 @@ defmodule Ecto.DateTime.Utils do
   def usec(rest) when is_iso_8601(rest), do: 0
   def usec(_), do: nil
 
+  def add_with_wrap(current, n, first..last = range) do
+    Enum.at(range, rem(current - first + n, last + (1 - first)))
+  end
+
+  def get_wrap_count(current, n, first..last) do
+    trunc Float.floor((current - first + n) / (last + (1 - first)))
+  end
+
   @doc """
   Compare two datetimes.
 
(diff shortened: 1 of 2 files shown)


gradenb:
My goal here was to make the subset of ISO 8601 strings that Ecto.{Date,Time,DateTime}.cast can handle less awkward. This branch should make them able to handle everything except week dates and ordinal dates. I was running into issues myself when trying to cast strings in the following format:
`2015-01T23:50:07+0000`
This should be the same as `2015-01T23:50:07Z`, but the functions could not handle it.

gradenb:
Is this something that is likely to get accepted? I'm sure the changes need review from more experience Elixir developers :)

michalmuskala:
I agree it would be nice to be a little bit more flexible out of the box, on the other hand I feel like this adds much complexity.

I also believe this makes you loose information - specifically the one about the time-zone. There would be no way to recover this information once the data is converted into `Ecto.DateTime`. If you care about time-zones, you should really consider one of the more advanced time libraries - timex or calendar - they both have ecto bindings via custom types.

I know the `+0000` timezone issue comes mostly from ruby's `DateTime` (that's how it renders UTC times), contrary to `Time` class that uses the `Z` format.
Maybe as a compromise we could support both `Z` and `+0000` time-zone indicators?

\cc @josevalim

gradenb:
@michalmuskala Thanks for the feedback. I agree with your comments. I think it would be much nicer to support any ISO 8601 string eventually, but I guess there should be more discussions before tackling that. I know that it would be nice to me personally if `+0000` was supported. I'm not sure how much this would help others though. I wouldn't mind changing my pull request to add just that support if that is desired. I'm mainly looking for ways to get involved and this seemed like low hanging fruit. :)

josevalim:
The goal of Ecto is exactly to do some subset. If you need robust parsing, then you need to use a proper date time library. :) Thank you for the PR in any case!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1163, opened 2015-12-22
Switch pg storage management away from psql

### LINKED ISSUE #854: Mix ecto.create uses psql
Is it necessary for Ecto to use `psql` for database creation? It causes a problem for me because I am using Docker (in all environments) and I don't have Postgres in my application container. It is making running automated tests difficult.

DAE think it would make more sense to use Ecto itself for database creation? I'm interested in contributing a patch for this, but first want to see if it would be welcome and get your thoughts on this.

sorentwo:
This moves `storage_up` and `storage_down` for the Postgres adapter to use an Ecto/Postgrex connection, rather than relying on `psql`.

This closes #854.

### DIFF
--- a/integration_test/pg/storage_test.exs
+++ b/integration_test/pg/storage_test.exs
@@ -15,12 +15,12 @@ defmodule Ecto.Integration.StorageTest do
         password: "password1234" ]
   end
 
-  def drop_database do
-    :os.cmd 'psql -U postgres -c "DROP DATABASE IF EXISTS storage_mgt;"'
-  end
-
   def create_database do
     :os.cmd 'psql -U postgres -c "CREATE DATABASE storage_mgt;"'
+  end
+
+  def drop_database do
+    :os.cmd 'psql -U postgres -c "DROP DATABASE IF EXISTS storage_mgt;"'
   end
 
   setup do
@@ -30,22 +30,17 @@ defmodule Ecto.Integration.StorageTest do
 
   test "storage up (twice in a row)" do
     assert Postgres.storage_up(correct_params) == :ok
-    assert Postgres.storage_up(correct_params) == {:error, :already_up}
+    assert Postgres.storage_up(correct_params) == :ok
   end
 
   test "storage up (wrong credentials)" do
-    refute Postgres.storage_up(wrong_user) == :ok
+    assert Postgres.storage_up(wrong_user) == {:error, "connection not available"}
   end
 
   test "storage down (twice in a row)" do
     create_database
 
     assert Postgres.storage_down(correct_params) == :ok
-    assert Postgres.storage_down(correct_params) == {:error, :already_down}
-  end
-
-  test "storage down (wrong credentials)" do
-    create_database
-    refute Postgres.storage_down(wrong_user) == :ok
+    assert Postgres.storage_down(correct_params) == :ok
   end
 end
--- a/integration_test/pg/storage_test.exs
+++ b/integration_test/pg/storage_test.exs
@@ -3,6 +3,8 @@ defmodule Ecto.Integration.StorageTest do
 
   alias Ecto.Adapters.Postgres
 
+  @moduletag :capture_log
+
   def correct_params do
     Ecto.Repo.Supervisor.parse_url(
       Application.get_env(:ecto, :pg_test_url) <> "/storage_mgt"
--- a/integration_test/pg/test_helper.exs
+++ b/integration_test/pg/test_helper.exs
@@ -71,12 +71,15 @@ defmodule Ecto.Integration.Case do
   end
 end
 
+{:ok, _} = Application.ensure_all_started(:postgrex)
+
 # Load up the repository, start it, and run migrations
-_   = Ecto.Storage.down(TestRepo)
+:ok = Ecto.Storage.down(TestRepo)
 :ok = Ecto.Storage.up(TestRepo)
 
 {:ok, _pid} = TestRepo.start_link
 {:ok, _pid} = PoolRepo.start_link
 
 :ok = Ecto.Migrator.up(TestRepo, 0, Ecto.Integration.Migration, log: false)
+
 Process.flag(:trap_exit, true)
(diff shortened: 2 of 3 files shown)


integration_test/pg/test_helper.exs
```diff
@@ -69,12 +69,15 @@ defmodule Ecto.Integration.Case do
   end
 end
 
+{:ok, _} = Application.ensure_all_started(:postgrex)
+
+Process.flag(:trap_exit, true)
+
 # Load up the repository, start it, and run migrations
-_   = Ecto.Storage.down(TestRepo)
+# _   = Ecto.Storage.down(TestRepo)
```
sorentwo: This is commented out to get the tests passing in general. When it is enabled it will invariably result in:

```
** (MatchError) no match of right hand side value: {:error, %Postgrex.Error{message: "tcp recv: timeout", postgres: nil}}
```

The `storage_test` itself and the `ecto.create_drop_test` don't normally have such a timeout problem. I'm not sure what is causing the timeout, and could really use some help.

sorentwo:
This has taken an embarrassing amount of time for such small code changes, and I'm still getting timeout errors that I don't understand.

integration_test/pg/storage_test.exs
```diff
@@ -33,19 +27,10 @@ defmodule Ecto.Integration.StorageTest do
     assert Postgres.storage_up(correct_params) == {:error, :already_up}
   end
 
-  test "storage up (wrong credentials)" do
-    refute Postgres.storage_up(wrong_user) == :ok
```
sorentwo: Incorrect credentials cause an exit, which I'm not sure should be caught.

```
20:51:08.074 [error] GenServer #PID<0.229.0> terminating
** (stop) %Postgrex.Error{message: nil, postgres: %{code: :invalid_authorization_specification, file: "miscinit.c", line: "362", message: "role \"randomuser\"
 does not exist", pg_code: "28000", routine: "InitializeSessionUserId", severity: "FATAL"}}
```

Any advice here?

lib/ecto/adapters/postgres.ex
```diff
@@ -79,66 +81,40 @@ defmodule Ecto.Adapters.Postgres do
...
+      error -> error
     end
   end
 
-  defp run_with_psql(database, sql_command) do
-    unless System.find_executable("psql") do
-      raise "could not find executable `psql` in path, " <>
-            "please guarantee it is available before running ecto commands"
-    end
+  defp run_query(opts, sql) do
```
GeorgeTaveras1231: How do you feel about making `run_query/2` return `{:error,  {code, message}}`? This will simplify the calling code to:
```elixir
case run_query(opts, command) do
  {:ok, _result} -> :ok
  {:error, {:duplicate_database, _message}} -> {:error, :already_up}
  {:error, {_, message}} -> {:error, message} # This would also be more like the original implementation
end
```

lib/ecto/adapters/postgres.ex
```diff
@@ -79,66 +81,40 @@ defmodule Ecto.Adapters.Postgres do
...
-      String.contains?(output, "3D000") -> {:error, :already_down}
-      true                              -> {:error, output}
+    command = "DROP DATABASE #{opts[:database]}"
+
+    case run_query(opts, command) do
+      {:ok, _result} ->
+        :ok
+      {:error, %Postgrex.Error{postgres: %{code: :invalid_catalog_name}}} ->
+        {:error, :already_down}
+      error -> error
```
GeorgeTaveras1231: Just a heads up, this is not backwards compatible. This used to return `{:error, String}` and is now returning `{:error, Postgrex.Error}`

sorentwo:
@GeorgeTaveras1231 Thanks for the suggestions, I've updated `run_query` to destructure the error, much better.

integration_test/pg/storage_test.exs
josevalim: Let's catch any Postgrex.Error exit and print the error messages in a nice readable way.

josevalim: To do this, you may need to start the connection and run the query inside a separate process (it can be one started with `Task.start_link`), monitor it and receive its error status (which may be Postgrex.Error).

GeorgeTaveras1231:
No problem @sorentwo, glad to help

btw, I don't know if this should be done as part of this same PR, but we should not forget about the [`mysql` dependency here](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/adapters/mysql.ex#L143-L159).

integration_test/pg/storage_test.exs
fishcakez: When we add DBConnection the GenServer log message is much improved as the exception is raised by DBConnection so we get to see the message (and a stacktrace). If it is awkward to handle a nice log we can leave it for now.

sorentwo:
I've made some changes to use Task.Supervisor.async_nolink with Task.yield, which works well aside from random timeouts. With ericmj/postgrex#128 being merged I think this will have to be reworked again though.

sorentwo:
The connection is now closed after the query is run, which prevents a timeout for back-to-back invocations of the `storage_up` or `storage_down`.

sorentwo:
@josevalim This has been updated with the db_connection changes, some cleanup, and squashed into a single commit. CI seems to be having s3 errors, so I'm not sure of the build status now.

sorentwo:
Ok, everything is passing now. This is ready to go from my end.

lib/ecto/adapters/postgres.ex
```diff
@@ -80,66 +80,48 @@ defmodule Ecto.Adapters.Postgres do
...
+
+    task = Task.Supervisor.async_nolink(pid, fn ->
+      {:ok, conn} = Postgrex.Connection.start_link(opts)
+
+      value = Ecto.Adapters.Postgres.Connection.query(conn, sql, [], [])
+      GenServer.stop(conn)
+      value
+    end)
+
+    case Task.yield(task) do
```
josevalim: You think the default timeout of 5000 is reasonable? Maybe we should use the timeout values coming from opts?

sorentwo: I have to admit that I didn't test this on anything under-powered, but it never exceeded even 1000 in my local usage/testing.

PR #1171, opened 2015-12-26
[1170] Cast must validate that the key be present if required

mrkaspa:
This PR is to solve https://github.com/elixir-lang/ecto/issues/1170


josevalim:
Thank you but this fix is incorrect. I will post the proper fix in the issue in case you want to tackle it.

PR #1174, opened 2015-12-26
raise when an association is defined in embedded_schema

### LINKED ISSUE #1166: Raise if embedded_schema is given and an association is defined

tony612:
resolves #1166


josevalim:

PR #1176, opened 2015-12-28
@schema_prefix for Ecto schema

Tica2:
Define prefix for schema with @schema_prefix attribute
Issues https://github.com/elixir-lang/ecto/issues/1006

For query we need to agree about format:
We current have: `%{query | prefix: "foo"}` 

I see: 
1) use `prefix` key as binary (current usage) to replace all schema prefixes (or only is nil)
    define second key `:build_prefix` as map for all possible schema prefixes.

2) use only `prefix` key as binary (current usage) or map for all possible schema prefixes.

Example to alter prefix on query is here (without error)
https://github.com/Tica2/ecto/commit/7c94a87f45a9ec83f40f111d60e32fac4ee8d4db#diff-975ad8a7b1f8e27ac59595bcdbdd0a07L443

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -35,6 +35,9 @@ defmodule Ecto.Schema do
       to `{:id, :id, autogenerate: true}`. When set to
       false, does not define a primary key in the schema;
 
+    * `@schema_prefix` - configures the schema prefix. Defaults `nil`
+      generate queries without prefix;
+
     * `@foreign_key_type` - configures the default foreign key type
       used by `belongs_to` associations. Defaults to `:integer`;
 
@@ -221,6 +224,8 @@ defmodule Ecto.Schema do
   used for runtime introspection of the schema:
 
   * `__schema__(:source)` - Returns the source as given to `schema/2`;
+  * `__schema__(:prefix)` - Returns optional prefix for source provided by
+    `@schema_prefix` schema attribute;
   * `__schema__(:primary_key)` - Returns a list of primary key fields (empty if there is none);
 
   * `__schema__(:fields)` - Returns a list of all non-virtual field names;
@@ -279,6 +284,7 @@ defmodule Ecto.Schema do
       @foreign_key_type :id
       @before_compile Ecto.Schema
       @ecto_embedded false
+      @schema_prefix nil
 
       Module.register_attribute(__MODULE__, :ecto_fields, accumulate: true)
       Module.register_attribute(__MODULE__, :ecto_assocs, accumulate: true)
@@ -311,6 +317,7 @@ defmodule Ecto.Schema do
   """
   defmacro schema(source, [do: block]) do
     quote do
+      prefix = Module.get_attribute(__MODULE__, :schema_prefix)
       source = unquote(source)
 
       unless is_binary(source) do
@@ -320,7 +327,7 @@ defmodule Ecto.Schema do
       Module.register_attribute(__MODULE__, :changeset_fields, accumulate: true)
       Module.register_attribute(__MODULE__, :struct_fields, accumulate: true)
       Module.put_attribute(__MODULE__, :struct_fields,
-                           {:__meta__, %Metadata{state: :built, source: {nil, source}}})
+                           {:__meta__, %Metadata{state: :built, source: {prefix, source}}})
 
       primary_key_fields =
         case @primary_key do
@@ -347,7 +354,7 @@ defmodule Ecto.Schema do
       Module.eval_quoted __ENV__, [
         Ecto.Schema.__defstruct__(@struct_fields),
         Ecto.Schema.__changeset__(@changeset_fields),
-        Ecto.Schema.__schema__(source, fields, primary_key_fields),
+        Ecto.Schema.__schema__(prefix, source, fields, primary_key_fields),
         Ecto.Schema.__types__(fields),
         Ecto.Schema.__assocs__(assocs),
         Ecto.Schema.__embeds__(embeds),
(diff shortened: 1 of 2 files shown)


lib/ecto/schema.ex
```diff
@@ -221,6 +224,8 @@ defmodule Ecto.Schema do
   used for runtime introspection of the schema:
 
   * `__schema__(:source)` - Returns the source as given to `schema/2`;
+  * `__schema__(:prefix)` - Returns optional prefix for source provided by
```
josevalim: Instead of `__schema__(:prefix)`, I believe we should make `__schema__(:source)` return a tuple. Then we check everywhere we call `__schema__(:source)` to verify it still makes sense.

josevalim:
For queries we don't need to do anything. We just need to change Ecto.Queryable implementation for atoms to consider the prefix we should return from `__schema__(:source)`. :) Thanks for the PR!

lib/ecto/schema.ex
Tica2: not sure

```elixir
__schema__(:source) == "table" # when @schema_prefix not present
__schema__(:source) == {"foo_prefix", "table"} # when @schema_prefix "foo_prefix"
```
or
```elixir
__schema__(:source) == {nil, "table"} # when @schema_prefix not present
__schema__(:source) == {"foo_prefix", "table"} # when @schema_prefix "foo_prefix"
```

Tica2:
@josevalim Perfect! Ecto.Queryable solve problem where `prefix` key is not set in query %{} 

I consider to have a way to alter all prefixes:

%{query | prefix: "foo"} Alter all prefixes?
or
Alter only prefixes with null?

Tica2:
Ecto.Queryable.to_query()
return {"table", model_module}

do not support prefix!

Tica2:
Again:

https://github.com/Tica2/ecto/commit/7c94a87f45a9ec83f40f111d60e32fac4ee8d4db#diff-975ad8a7b1f8e27ac59595bcdbdd0a07L443

any source in sources is 3-tuple
{"prefix"."table", "t1", App.Models.Table}
here we need to alter prefix

Tica2:
more reference to __schema__(:source) assume is Binary:
 
https://github.com/elixir-lang/ecto/blob/master/lib/ecto/changeset.ex#L1614
https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/planner.ex#L312

lib/ecto/schema.ex
josevalim: The second. always a tuple is better.

josevalim:
> I consider to have a way to alter all prefixes:

in `Ecto.Queryable` for `Atom` and for `Tuple` we will never have a previous prefix, so we can always change it. :)

josevalim:
> https://github.com/elixir-lang/ecto/blob/master/lib/ecto/changeset.ex#L1614
> https://github.com/elixir-lang/ecto/blob/master/lib/ecto/query/planner.ex#L312

So get only the second element in both cases. :) You can pattern match on it. :)

Tica2:
ok, is easy to fix in ecto.
Problem is with external deps that assume   `__schema__(:source)` return binary 

this is my concern

josevalim:
@Tica2 master is Ecto 2.0. We can break backwards compatibility. :)

josevalim:
Ping. :)

Tica2:
prepare new PR with
```__schema__(:source) == {"prefix", "table"}```

fix tests that fails...

**Comment by Tica2** [comment]
if we want:
```__schema__(:source) == {"prefix", "table"}```
In many places code assume ```source``` is 2-Tuple ```{"table", Model}```
in combination with ```__schema__(:source) == {"prefix", "table"}```

now we have:
2-Tuple ```{"table", Model}```
or
3-Tuple ```{{"prefix", "table"}, Model}```

and exploding number of problems ...
 
For instance: 
```Ecto.Queryable```  look like:



```elixir
defprotocol Ecto.Queryable do
  @moduledoc """
  Converts a data structure into an `Ecto.Query`.
  """

  @doc """
  Converts the given `data` into an `Ecto.Query`.
  """
  def to_query(data)
end

defimpl Ecto.Queryable, for: Ecto.Query do
  def to_query(query), do: query
end

defimpl Ecto.Queryable, for: BitString do
  def to_query(source) when is_binary(source),
    do: %Ecto.Query{from: {{nil, source}, nil}}
end

defimpl Ecto.Queryable, for: Atom do
  def to_query(module) do
    try do
      %Ecto.Query{from: {module.__schema__(:source), module}}
    rescue
      UndefinedFunctionError ->
        message = if :code.is_loaded(module) do
          "the given module is not queryable"
        else
          "the given module does not exist"
        end

        raise Protocol.UndefinedError,
             protocol: @protocol,
                value: module,
          description: message
    end
  end
end

  def to_query(from = {{schema_prefix, source}, model}) when is_binary(source) and is_atom(model),
    do: %Ecto.Query{from: from}

  def to_query({nil, source}) when is_binary(source),
    do: %Ecto.Query{from: {{nil, source}, nil}}

  def to_query({nil, model}) when is_atom(model),
    do: %Ecto.Query{from: {{nil, nil}, model}}

  def to_query({source, model}) when is_binary(source) and is_atom(model),
    do: %Ecto.Query{from: {{nil, source}, model}}
end
```

I think `__schema__(:prefix)` is best

Tica2:
If user work with plain "table_name" as string then do not have prefix. 
Is require to use Module to have static prefix for each Module.

josevalim:
That's the point. We should handle all those cases accordingly, we shouldn't just replace it by source altogether. But no worries, I have some time and I can merge this and change accordingly. :)

Tica2:
it is not easy, keep me in loop...

josevalim:
@tica2 i have pushed a commit. my concern with not adding the prefix to `__schema__(:source)` is because it would be easy to forget using prefix in a bunch of places and this was, in fact, happening. So I changed `__schema__(:source)` to return a tuple, fixed all places that broken, double checked the code and then reverted them to be two separate clauses. This way, we have the best of both worlds and now we can be sure we didn't forget everything. :) Thank you!

Tica2:
Check now #1176
Yes! you're right. New source code need to take care of prefixes.

PR #1177, opened 2015-12-29
Support many to many

josevalim:
Still WIP.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -803,7 +803,7 @@ defmodule Ecto.Schema do
       The keys are inflected from the schema names.
 
     * `:on_delete` - The action taken on associations when the parent record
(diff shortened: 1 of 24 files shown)


michalmuskala:
Should we consider splitting all the assocs modules into separate files?

lib/ecto/schema.ex
```diff
@@ -770,6 +770,89 @@ defmodule Ecto.Schema do
...
+      for more info.
+
+    * `:defaults` - Default values to use when building the association
+
+  ## Removing data
+
+  If you attempt to remove associated `many_to_many` data, be it by
+  setting `:on_replace` to `:delete`, `:on_delete` to `:delete_all`
+  or by using `Ecto.Changeset.put_assoc/3` and `Ecto.Changeset.cast_assoc/3`,
+  **Ecto will always remove data from the joint schema and never from
```
michalmuskala: s/joint/join/

PR #1180, opened 2015-12-31
Improved support for PG check and exclusion constraints

nathanl:
Postgres supports ["check constraints"](http://www.postgresql.org/docs/9.5/static/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS), which validate a row to say that,
for example, `price` must be greater than `0`, or `end_time` must be
greater than `start_time`. This PR is meant to add support for such
checks.

I did **not** ask about this on the mailing list, so if you don't want this
feature, that's OK. :blush: I'm new to Elixir and Phoenix and wanted
to see if I could figure this out. So I'm asking for feedback before I
go further.

FWIW, of all the constraints [Postgres supports](http://www.postgresql.org/docs/9.5/static/ddl-constraints.html), check constraints are
the only ones Ecto doesn't yet support, so I thought this might be a useful
addition.

What works so far:

- Adding a constraint when adding a column in a migration
- Capturing the error returned by Postgres if a constraint is violated
- Turning that into a readable error on the changeset

What's still missing:

- Ability to `modify` an existing column to add a constraint
- Ability to remove constraints
- Ability to add table constraints (not tied to a single column; 
eg, `end_time` must be greater than `start_time`)

I'm currently tinkering with this in my `Rumbl` app from "Programming Phoenix".
I added an `other_description` column to `videos` and gave it a
constraint, like this:

```elixir
def change do
  alter table(:videos) do
    add :other_description, :string,
      check: check("char_length(other_description) < 5",
      name: "other_description_must_be_short")
  end
end
```

Then I added the constraint to `Video` like this:

```elixir
def changeset(model, params \\ :empty) do
   model
   |> cast(params, @required_fields, @optional_fields)
   |> foreign_key_constraint(:category_id)
   |> check_constraint(:other_description, name: :other_description_must_be_short)
end
```

With that much, adding a long description gives the user this error:

constraint_error_no_message_given

If I instead use

```elixir
|> check_constraint(:other_description,
  name: :other_description_must_be_short,
  message:  "must be less than 5 characters"
)
```

The user sees:

constraint_error_with_message

Should I keep working on this? If so, what feedback do you have so far?

### DIFF
--- a/integration_test/pg/check_constraint_test.exs
+++ b/integration_test/pg/check_constraint_test.exs
@@ -1,0 +1,69 @@
+defmodule Ecto.Integration.CheckConstraintTest do
+  use ExUnit.Case, async: true
+
+  alias Ecto.Integration.TestRepo
+  import Ecto.Migrator, only: [up: 4, down: 4]
+
+  defmodule CheckConstraintMigration do
+    use Ecto.Migration
+
+    @table table(:products)
+
+    def change do
+      create @table do
+        add :price, :integer
+      end
+      create constraint(@table.name, "positive_price", check: "price > 0")
+    end
+  end
+
+  defmodule CheckConstraintModel do
+    use Ecto.Integration.Schema
+
+    schema "products" do
+      field :price, :integer
+    end
+  end
+
+  test "creating, using, and dropping a check constraint" do
+    assert :ok == up(TestRepo, 20120806000000, CheckConstraintMigration, log: false)
+
+    # When the changeset doesn't expect the db error
+    changeset = Ecto.Changeset.change(%CheckConstraintModel{}, price: -10)
+    exception =
+      assert_raise(Ecto.ConstraintError, ~r/constraint error when attempting to insert model/, fn ->
+        TestRepo.insert(changeset)
+      end
+      )
+    assert exception.message =~ "check: positive_price"
+    assert exception.message =~ "The changeset has not defined any constraint."
+
+    # When the changeset does expect the db error, but doesn't give a custom message
+    changeset = Ecto.Changeset.change(%CheckConstraintModel{}, price: -10)
+    {:error, changeset} =
+      changeset
+      |> Ecto.Changeset.check_constraint(:price, name: :positive_price)
+      |> TestRepo.insert()
+    assert changeset.errors == [price: "violates check 'positive_price'"]
+    assert changeset.model.__meta__.state == :built
+
+    # When the changeset does expect the db error and gives a custom message
+    changeset = Ecto.Changeset.change(%CheckConstraintModel{}, price: -10)
+    {:error, changeset} =
+      changeset
+      |> Ecto.Changeset.check_constraint(:price, name: :positive_price, message: "price must be greater than 0")
+      |> TestRepo.insert()
+    assert changeset.errors == [price: "price must be greater than 0"]
+    assert changeset.model.__meta__.state == :built
+
+    # When the change does not violate the check constraint
+    changeset = Ecto.Changeset.change(%CheckConstraintModel{}, price: 10)
+    {:ok, changeset} =
+      changeset
+      |> Ecto.Changeset.check_constraint(:price, name: :positive_price, message: "price must be greater than 0")
+      |> TestRepo.insert()
+    assert is_integer(changeset.id)
+
+    assert :ok == down(TestRepo, 20120806000000, CheckConstraintMigration, log: false)
+  end
+end
--- a/integration_test/pg/exclusion_constraint_test.exs
+++ b/integration_test/pg/exclusion_constraint_test.exs
@@ -1,4 +1,4 @@
-defmodule Ecto.Integration.ConstraintTest do
+defmodule Ecto.Integration.ExclusionConstraintTest do
   use ExUnit.Case, async: true
 
   alias Ecto.Integration.TestRepo
@@ -7,61 +7,58 @@ defmodule Ecto.Integration.ConstraintTest do
   defmodule ExcludeConstraintMigration do
     use Ecto.Migration
 
-    @table table(:exclude_constraint_migration)
+    @table table(:non_overlapping_ranges)
 
-    def up do
+    def change do
       create @table do
         add :from, :integer
         add :to, :integer
       end
-      execute "ALTER TABLE exclude_constraint_migration " <>
-              "ADD CONSTRAINT overlapping_ranges EXCLUDE USING gist (int4range(\"from\", \"to\") WITH &&)"
-    end
-
-    def down do
-      drop @table
+      create constraint(@table.name, :cannot_overlap, exclude: ~s|gist (int4range("from", "to", '[]') WITH &&)|)
     end
   end
 
   defmodule ExcludeConstraintModel do
     use Ecto.Integration.Schema
 
-    schema "exclude_constraint_migration" do
+    schema "non_overlapping_ranges" do
       field :from, :integer
       field :to, :integer
     end
   end
 
-  test "exclude constraint exception" do
+  test "creating, using, and dropping an exclude constraint" do
     assert :ok == up(TestRepo, 20050906120000, ExcludeConstraintMigration, log: false)
 
-    changeset = Ecto.Changeset.change(%ExcludeConstraintModel{}, from: 0, to: 1)
+    changeset = Ecto.Changeset.change(%ExcludeConstraintModel{}, from: 0, to: 10)
     {:ok, _} = TestRepo.insert(changeset)
+
+    non_overlapping_changeset = Ecto.Changeset.change(%ExcludeConstraintModel{}, from: 11, to: 12)
+    {:ok, _} = TestRepo.insert(non_overlapping_changeset)
+
+    overlapping_changeset = Ecto.Changeset.change(%ExcludeConstraintModel{}, from: 9, to: 12)
 
     exception =
       assert_raise Ecto.ConstraintError, ~r/constraint error when attempting to insert model/, fn ->
-        changeset
+        overlapping_changeset
         |> TestRepo.insert()
       end
-
-    assert exception.message =~ "exclude: overlapping_ranges"
+    assert exception.message =~ "exclude: cannot_overlap"
     assert exception.message =~ "The changeset has not defined any constraint."
 
     message = ~r/constraint error when attempting to insert model/
     exception =
       assert_raise Ecto.ConstraintError, message, fn ->
-        changeset
+        overlapping_changeset
         |> Ecto.Changeset.exclude_constraint(:from)
         |> TestRepo.insert()
       end
-
-    assert exception.message =~ "exclude: overlapping_ranges"
+    assert exception.message =~ "exclude: cannot_overlap"
 
     {:error, changeset} =
-      changeset
-      |> Ecto.Changeset.exclude_constraint(:from, name: :overlapping_ranges)
+      overlapping_changeset
+      |> Ecto.Changeset.exclude_constraint(:from, name: :cannot_overlap)
       |> TestRepo.insert()
-
     assert changeset.errors == [from: "violates an exclusion constraint"]
     assert changeset.model.__meta__.state == :built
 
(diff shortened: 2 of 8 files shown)


josevalim:
Thank you @nathanl! This is a welcome addition and you can look at #1122 for guidance on the integration tests.

The only change I would do is to not define the constraint alongside the name. This makes things really more complex when there is no specific field or when changing the table. In other words, is it possible that, instead of writing something like:

    add :name, :string, check(...)

I would prefer if we did:

    create check_constraint(:table, ..., opts)

As the later is much more manageable. If we can do the later, we may even add the same for exclude constraints.

Thank you!

nathanl:
> Thank you @nathanl! This is a welcome addition...

Yaaay! :smile: 

> I would prefer if we did...

Ah, yes - that would make it simpler. I'll work on that as soon as I can.

nathanl:
Updated. The following now works:

```elixir
defmodule Rumbl.Repo.Migrations.AddCheckToVideos do
  use Ecto.Migration

  def up do
    create check_constraint(:videos, check: "char_length(description) < 20", name: "description_must_be_short")
  end

  def down do
    drop check_constraint(:videos, name: "description_must_be_short")
  end

end
```

and in `Video`:

```elixir
  def changeset(model, params \\ :empty) do
    model
    |> cast(params, @required_fields, @optional_fields)
    |> foreign_key_constraint(:category_id)
    |> check_constraint(:description, name: :description_must_be_short, message: "must be less than 20 characters") # message key is optional
  end
```

Please review carefully; I'm very new to this codebase and to some extent relied on copy and pasting existing code. Eg, I don't quite understand why I needed to define a function head for `execute_ddl` in `Connection` in order to make the migration work; the SQL string from `command` in `Runner` seems to be what's executed by the migration.

lib/ecto/migration.ex
```diff
@@ -263,7 +278,11 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Creates an index or a table with only `:id` field.
+  Creates one of the following:
+
+  - an index
```
josevalim: Please use `*` and indent with 2 spaces :)

lib/ecto/migration.ex
```diff
@@ -656,6 +690,28 @@ defmodule Ecto.Migration do
...
+      drop   check_constraint(:products, name: "price_must_be_positive")
+
+  ## Options
+
+    * `:check` - The expression to evaluate on a row. Required when creating.
+    * `:name` - The name of the constraint - required.
+
+  """
+  def check_constraint(table, opts) do
+    case Keyword.get(opts, :name) do
```
josevalim: `Keyword.get(opts, :name) || raise ArgumentError...`

lib/ecto/migration.ex
```diff
@@ -685,6 +741,10 @@ defmodule Ecto.Migration do
     reference
   end
 
+  defp validate_type!(%CheckConstraint{} = check_constraint) do
```
josevalim: Unused function?

josevalim:
> why I needed to define a function head for execute_ddl in Connection in order to make the migration work; the SQL string from command in Runner seems to be what's executed by the migration.

That's only for `execute "command"`. Everything else goes through the adapter.

integration_test/pg/check_constraint_test.exs
```diff
@@ -0,0 +1,88 @@
+defmodule Ecto.Integration.CheckConstraintTest do
+  use ExUnit.Case, async: true
+
+  alias Ecto.Integration.TestRepo
+  import Ecto.Migrator, only: [up: 4, down: 4]
+
+  defmodule TableCreationMigration do
```
josevalim: Please merge the migrations together.

josevalim:
Thank you @nathanl! This seems to be all technically correct. I have left some comments regarding styles and I have two other remarks:

1. Please follow Ecto's convention on things like assertions. So use `assert left == right`, without parens. If the line is too long, break after `==`.

2. I am not happy with this API:

        check_constraint(table, name: name, check: ...)

    Name is actually required (although consistent with indexes) and check is optional but it is kind of redundant. I am wondering if we should move to something like this:

        constraint(table, name, check: ...)

    The goal of using `constraint` is that we could add exclusion checks to it in the future too:

        constraint(table, name, exclusion: ...)

    The code, of course, would need to check if one of them is set.

Thoughts?

lib/ecto/changeset.ex
```diff
@@ -1395,6 +1395,26 @@ defmodule Ecto.Changeset do
...
+  ## Options
+
+    * `:message` - the message in case the constraint check fails.
+      Defaults to something like "violates check 'products_price_check'"
+    * `:name` - the name of the constraint. Required.
+
+  """
+  def check_constraint(changeset, field, opts \\ []) do
+    Keyword.get(opts, :name) || raise ArgumentError, "must supply a name for the check constraint"
+    constraint = opts[:name]
```
josevalim: Even better, you can have it here:

    constraint = opts[:name] || raise ArgumentError, ...

test/ecto/adapters/postgres_test.exs
```diff
@@ -715,6 +715,28 @@ defmodule Ecto.Adapters.PostgresTest do
...
+  test "create check constraint, specifying constraint name" do
+    create = {:create, check_constraint(:products, check: "price > 0", name: "price_must_be_positive")}
+    assert(
+    SQL.execute_ddl(create) == ~s|ALTER TABLE "products" ADD CONSTRAINT "price_must_be_positive" CHECK (price > 0)|
+    )
+  end
+
+  test "create check constraint, not specifying constraint name" do
+    exception =
+      assert_raise(ArgumentError, ~r/supply a name/, fn ->
```
josevalim: No need to check this here. We check on Ecto.Migration tests. So you can erase this test.

test/ecto/changeset_test.exs
```diff
@@ -828,6 +828,22 @@ defmodule Ecto.ChangesetTest do
...
+    assert changeset.constraints ==
+           [%{type: :check, field: :title, constraint: "title_must_be_short",
+              message: "violates check 'title_must_be_short'"}]
+
+    changeset = change(%Post{}) |> check_constraint(:title, name: :title_must_be_short, message: "cannot be more than 15 characters")
+    assert changeset.constraints ==
+           [%{type: :check, field: :title, constraint: "title_must_be_short",
+              message: "cannot be more than 15 characters"}]
+
+    assert_raise(ArgumentError, ~r/supply a name/, fn ->
```
josevalim: Style change, just drop the parens from assert_raise call.

josevalim:
Note to myself: make sure MySQL raises when using those constraints.

integration_test/pg/check_constraint_test.exs
nathanl: I'm happy to merge them, but the reason I initially made them separate was that I was afraid that if `down` did both `drop check_constraint...` and `drop table`, it would return `:ok` and appear to have worked (the constraint would be destroyed along with the table) even if `drop check_constraint` didn't work. Not a problem?

BTW I also used a different version timestamp than what I'd seen in other migrations because otherwise I got intermittent test failures, saying a migration was already up. Seemed like a race condition, maybe because of versions + `async: true`.

josevalim: > Not a problem?

Not a problem.

> maybe because of versions + async: true.

Yes exactly, different versions are required. :+1:

nathanl:
@josevalim Yep, I like the API you describe much better, and I think it will generalize well to exclusion constraints. I'll change `check_constraint` to be `constraint` as you suggest.

nathanl:
@josevalim I've added a bunch of commits; I'll rebase on master before we merge this, but I think doing that makes it hard to see previous comments so I'll wait until you say this is mergeable.

I've done this:

- Add migration support for `create constraint(table_name, "positive_price", check: "price > 0")`
- Add migration support for `create constraint(table_name, :overlapping_ranges, exclude: ~s|gist (int4range("from", "to", '[]') WITH &&)|`
- Made `create constraint...` reversible so the user can just define a `change` in their migration
- Improved tests and docs for constraints

So far I've left the separate functions in `Changeset` for `check_constraint` and `exclude_constraint` because I thought it seemed more natural than combining them and passing `:check` or `:exclusion`, but I'm happy to change that if you disagree.

I'd like to ensure that the MySQL raises a nice error if you try to add a check or exclusion constraint, as you mentioned. That's all I see as still missing unless you see something else.

I wonder if we might want to enable `create constraint(table_name, unique: [:column1, :column2...])` also, but that could be a different PR.

nathanl:
One philosophical question: given that PostgreSQL will only check one constraint at a time, but validations can check multiple attributes at once, what do you think is the best recommendation for using check constraints?

Eg, you wouldn't want the user to submit, get validation errors, submit, get a constraint error, fix that, submit, get another constraint error, etc.

At the moment, I'm thinking I'd do every "single row" check with validations, and maybe add some very basic check constraints as insurance in case of a bad changeset definition somewhere (since constraints are global), or to protect against bad data from another code base (a peer application or whatever replaces this one someday).

josevalim:
@nathanl I believe you are able to implement most check constraints in Elixir itself, so I would do that so you can give quick feedback to the user and leave the database constraint for the exceptional cases. I will review the PR soon. :)

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -702,6 +719,19 @@ if Code.ensure_loaded?(Postgrex.Connection) do
     defp null_expr(true), do: "NULL"
     defp null_expr(_), do: []
 
+    defp new_constraint_expr(%Ecto.Migration.Constraint{check: nil, exclude: nil})  do
```
nathanl: I defined matching function heads for the two bad cases so that I could give a friendlier message than "no matching function", but I'll delete these if you think that's overdoing it. :)

josevalim: This whole function is beautiful. :)

lib/ecto/migration.ex
```diff
@@ -177,6 +177,23 @@ defmodule Ecto.Migration do
...
+    """
+    defstruct name: nil,
+              table: nil,
+              check: nil,
+              exclude: nil
+
+    @type t :: %__MODULE__{
+      name: atom,
+      table: atom,
+      check: String.t(),
```
josevalim: `String.t | nil` and below.

lib/ecto/migration/runner.ex
```diff
@@ -238,6 +241,18 @@ defmodule Ecto.Migration.Runner do
   defp command({:rename, %Table{} = table, current_column, new_column}),
     do: "rename column #{current_column} to #{new_column} on table #{quote_table(table.prefix, table.name)}"
 
+  defp command({:create, %Constraint{check: nil, exclude: nil}}),
+    do: raise ArgumentError, "a constraint must have either a check or exclude option"
+  defp command({:create, %Constraint{check: check, exclude: exclude}}) when is_binary(check) and is_binary(exclude),
+    do: raise ArgumentError, "a constraint must not have both check and exclude options"
+  defp command({:create, %Constraint{check: check, exclude: nil} = constraint}) when is_binary(check),
```
josevalim: You don't need the `exclude: nil`, nor `check: nil` below.

lib/ecto/adapters/postgres/connection.ex
josevalim: Actually, I just realized you did those checks on the runner, so you can get rid of the first two clauses here and the nil checks below.

integration_test/pg/exclusion_constraint_test.exs
```diff
@@ -7,60 +7,59 @@ defmodule Ecto.Integration.ConstraintTest do
...
         add :from, :integer
         add :to, :integer
       end
-      execute "ALTER TABLE exclude_constraint_migration ADD CONSTRAINT overlapping_ranges EXCLUDE USING gist (int4range(\"from\", \"to\") WITH &&)"
-    end
-
-    def down do
-      drop @table
+      create constraint(@table.name, :cannot_overlap, exclude: ~s|gist (int4range("from", "to", '[]') WITH &&)|
+      )
```
josevalim: This should be in the previous line. :)

josevalim:
@nathanl I have added small comments, just details really. Everything else is perfect! Let's rebase, squash and ship it. :) Thank you!

lib/ecto/adapters/postgres/connection.ex
josevalim: Oops, I didn't see your comment when I added mine. Just to clarify: I think those two clauses should go because you already check for them in the runner.

nathanl:
@josevalim I think it's ready to go now. I'm excited to get a patch in!

josevalim:

PR #1184, opened 2016-01-03
Test entry in SQL logs

fishcakez:


josevalim:
I moved the log tests to repo.exs to avoid adding even more files. We can move it back to a separate file if we need more log specific tests. :)  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1185, opened 2016-01-04
test for @schema_prefix for 2 tables

Tica2:
This test fails.


josevalim:
We are not going to do it cross schema because it is not guarantee to work. We are going to use it only for `from` and you are the one responsible for keeping it together.

Tica2:
This is idea to have schema. To work with tables from one or more schemes.
Not happy at all with this model.

I can't join `hr.persons` and `main.cities` 
I will wait for more users to ask for this.
Thanks.

PR #1191, opened 2016-01-07
index/unique_index can now take a single atom

kyledecot:
When starting to use phoenix/ecto it through me for a loop that I couldn't do:

    add index(:posts, :title)

Instead you'd normally have to do:

    add index(:posts, [:title])

This PR allows for a single atom to be passed to `index` as well as `unique_index`

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -445,9 +445,14 @@ defmodule Ecto.Migration do
       create index(:products, [:user_id], where: "price = 0", name: :free_products_index)
 
   """
-  def index(table, columns, opts \\ []) when is_atom(table) and is_list(columns) do
-    index = struct(%Index{table: table, columns: columns}, opts)
-    %{index | name: index.name || default_index_name(index)}
+  def index(table, columns, opts \\ []) when is_atom(table) do
+    cond do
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you for the PR but I am missing the rationale.  :)

In Phoenix/Ecto, we actually prefer to be explicit and specific about which arguments we receive, so we avoid common practices as seen in Rails where a method/function would receive all kinds of arguments and the framework would have to figure it out. The spec is clear that the index is always over a list of columns or expressions, so I would keep the argument a list only.

kyledecot:
Okay, makes sense. I'll keep that in mind for future PRs. Thanks @josevalim

paulcsmith:
@josevalim Thanks for the explanation, it's always helpful to hear the rationale behind these decisions! 

I have also been bitten by this a couple times, and then had to go look at the docs to remember that it's supposed to be a list :P. I wonder if Ecto could help the developer out a bit here. What do you think of adding a guard to make sure it's a list, and then adding a fallback clause that raises and lets the user know it is expecting a list if you pass something else instead? `"Ecto.Migration.index/2 expects columns to be a list, got :title instead"`

PR #1197, opened 2016-01-10
Typos in schema.ex

jamilabreu:
here you go :) while I have you, I’m working on a new phoenix app and looking to use many_to_many - but it seems 2.0 isn’t out yet. are there any work-arounds until then?


josevalim:
You can use has_many :through. It will work just fine for reading but it will require more work for manipulating.

PR #1198, opened 2016-01-11
Provide ownership based sandbox

josevalim:
A pool for concurrent transactional tests.

The sandbox pool is implemented on top of an ownership mechanism.
When started, the pool is in automatic mode, which means using
the repository will automatically check connections out as with
any other pool. The only difference is that connections are not
checked back in automatically but by explicitly calling `checkin/2`.

The `mode/2` function can be used to change the pool mode to
`:manual`. In this case, each connection must be explicitly
checked out before use. This is useful when paired with
`checkout/2` which by default wraps the connection in a transaction.
This means developers have a safe mechanism for running concurrent
tests against the database.

## Example

The first step is to configure your database to use the
`Ecto.Adapters.SQL.Sandbox` pool. You set those options in your
`config/config.exs`:

    config :my_app, Repo,
      pool: Ecto.Adapters.SQL.Sandbox

Since you don't want those options in your production database, we
typically recommend to create a `config/test.exs` and add the
following to the bottom of your `config/config.exs` file:

    import_config "config/#{Mix.env}.exs"

Now with the test database properly configured, you can write
transactional tests:

    # At the end of your test_helper.exs
    # Set the pool mode to manual for explicitly checkouts
    Ecto.Adapters.SQL.Sandbox.mode(TestRepo, :manual)

    defmodule PostTest do
      # Once the model is manual, tests can also be async
      use ExUnit.Case, async: true

      setup do
        # Explicitly get a connection before each test
        :ok = Ecto.Adapters.SQL.Sandbox.checkout(TestRepo)
      end

      test "create comment" do
        # Use the repository as usual
        assert %Post{} = TestRepo.insert!(%Post{})
      end
    end


integration_test/cases/pool.exs
```diff
@@ -39,20 +38,14 @@ defmodule Ecto.Integration.PoolTest do
   end
 
   test "starts repo with after_connect" do
-    assert {:ok, _} = MockRepo.start_link(lazy: false, name: MockRepo.AfterConnect, query_cache_owner: false)
+    assert {:ok, pid} = MockRepo.start_link(lazy: false, query_cache_owner: false)
```
fishcakez: Lazy doesnt exist anymore we need to remove it

lib/ecto/adapters/sql.ex
```diff
@@ -210,122 +205,42 @@ defmodule Ecto.Adapters.SQL do
...
+      @doc false
+      def __pool__, do: {unquote(name), unquote(Macro.escape(config))}
+    end
   end
 
-  @spec rollback_test_transaction(Ecto.Repo.t, Keyword.t) :: :ok
-  def rollback_test_transaction(repo, opts \\ []) do
-    test_transaction(:rollback, repo, opts)
-  end
+  defp normalize_pool(Ecto.Adapters.SQL.Sandbox),
```
fishcakez: Should we normalize the old Ecto.Pools.* to DBConnection equivalents too?

josevalim: I am not sure yet. That's what is left to figure out, the user API.

lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -1,290 +1,236 @@
...
-    sandbox_begin(s, opts)
-  end
-  defp handle_request(:rollback, opts, %{status: :sandbox} = s) do
-    sandbox_rollback(s, opts)
-  end
-  defp handle_request(:rollback, _, s) do
-    {:ok, [], s}
-  end
+    def child_spec(_module, _opts, _child_opts) do
+      raise "should never be invoked"
```
fishcakez: We should use child_spec instead of start_link from Repo.Supervisor

fishcakez: oh not here but in general :D

lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -1,290 +1,227 @@
...
-      {:ok, state} ->
-        s = %__MODULE__{mod: mod, state: state, status: :run, adapter: adapter}
-        {:ok, s}
-      {:error, _} = error ->
-        error
-    end
-  end
+        setup do
+          # Explicitly get a connection before each test
+          :ok = Ecto.Adapters.SQL.Sandbox.checkout(TestRepo)
```
maxnordlund: Don't you need to have a teardown block as well, returning the connection? Or is that handled automatically?

fishcakez: When the owner exits the connection is automatically returned to the pool

josevalim: Great question though. I will add it to the docs.

youroff:
Is this setup suitable for integration tests? I followed the instructions but in my test, where access to db performed somewhere in the app, when I connect with test websocket client, I get: `RuntimeError{message: "cannot find ownership process for #PID<0.399.0>. This may happen if you have not explicitly checked out or the checked out process crashed"}` Also this might be worth including into sample app, it took a while until I found this.

josevalim:
@youroff It should be possible but you need some extra work. For example, you can pass the owner pid as a parameter:

    self |> :erlang.pid_to_list |> List.to_string |> Base.url_encode64

Then in your socket you will need to do:

    pid = encoded |> Base.url_decode64! |> String._to_list |> :erlang.list_to_pid
    Ecto.Adapters.SQL.Sandbox.allow(Repo, pid, self())

We need to make this easier and test-only but I am not sure yet how.

youroff:
I'm not sure that I understand who's the owner here. Is this a main app process? Or Supervisor, that runs Database pool? Or process where test is performing? Sorry, if this is a dumb question, playing with elixir only for two weeks.

josevalim:
@youroff no, they are not dumb questions. The owner process is the test that checks out the connection from the sandbox. In this case, it is the test process. The one running your test. Once the request starts, you need to tell it which connection to use, that's why you get the error. So one approach right now is to use what I proposed above, which is to encode the test process pid somehow into the parameters sent to the websocket and decode it there for testing purposes. We are investigating simpler solutions.

youroff:
@josevalim ah, I see now. The need of polluting app code with test-related snippets looks hacky, but bearable. Are there other scenarios of Sandbox usage? I assume if there's manual, then there might be :auto or something, in case if I'm ok with sync test suite for integration? Or I can't mix async and sync tests?

josevalim:
Yes, your tests would need to be sync. The issue with :auto is that you don't have a transaction, so they will pollute the DB. I have opened an issue to support shared mode, which is a mixture of auto and manual: #1223

youroff:
Oh, very cool. Thank you, the solution you proposed totally worked. And thanks a lot for Elixir itself, it is absolutely awesome!

youroff:
I've got another issue. In dev mode with `pool: Ecto.Adapters.SQL.Sandbox` in config, sometimes I see this error:
`18:29:17.983 [error] Postgrex.Protocol (#PID<0.420.0>) disconnected: ** (DBConnection.Error) client timeout`
Is this config for tests only, so shouldn't it go to test.exs from the beginning? Or I have something unrelated?

josevalim:
It is test only. :)

youroff:
Yeah, right, my bad :) But config example is really misleading, probably worth mentioning that it goes to test.exs. Now it's reading as if it stays at the top and the rest of configs per dev go to sub-files.

fishcakez:
> I've got another issue. In dev mode with pool: Ecto.Adapters.SQL.Sandbox in config, sometimes I see this error:
> 18:29:17.983 [error] Postgrex.Protocol (#PID<0.420.0>) disconnected: ** (DBConnection.Error) client timeout
> Is this config for tests only, so shouldn't it go to test.exs from the beginning? Or I have something unrelated?

This is a bug in DBConnection's ownership pool I will open an issue there. Thank you for the report.

josevalim:
@youroff I have pushed shared mode that will make your scenario much easier with the downside of not allowing concurrent tests. You can read a direct comparison of those modes here: https://github.com/elixir-lang/ecto/commit/dc9c2f940511999bcb5ca1d6c47ee8ef7f5fde61#diff-75c3d8c8b421ecc99c1ef78229abeb6aR54

youroff:
Awesome, thank you @josevalim! :+1: :+1: :+1:

fishcakez:
> I've got another issue. In dev mode with pool: Ecto.Adapters.SQL.Sandbox in config, sometimes I see this error:
> 18:29:17.983 [error] Postgrex.Protocol (#PID<0.420.0>) disconnected: ** (DBConnection.Error) client timeout
> Is this config for tests only, so shouldn't it go to test.exs from the beginning? Or I have something unrelated?

In DBConnection master the ownership pool (the base of Sandbox) supports setting `ownership_timeout: :infinity` which will allow using the Sandbox from shell/dev.

youroff:
@fishcakez So, should I use Sandbox outside of test env? I think I still don't understand what is Sandbox, could you point at docs or sources where this abstraction is introduced?

josevalim:
@youroff with @fishcakez's commit, now you can use it for dev. The best way to read more about it is in the docs I have linked earlier.

PR #1199, opened 2016-01-11
Add support for views

sorentwo:
http://www.postgresql.org/docs/9.5/interactive/rules-views.html
http://www.postgresql.org/docs/9.5/interactive/sql-createview.html
http://www.postgresql.org/docs/9.5/interactive/rules-materializedviews.html

This is an initial stab at adding direct support for creating, altering, dropping, and using views. This is an important feature of Postgres that doesn't get much love, especially in the Rails/ActiveRecord world where there isn't any native support for it. I figured I'd open a PR as an RFC and start some of the work to scope it out.

I imagine the primary use to look like this:

```elixir
def change do
  create view(:products, materialized: true, prefix: :outdated, query: """
    SELECT * FROM products WHERE inserted_at > now() - '1 month'::interval
  """)
end

# CREATE MATERIALIZED VIEW outdated_products AS SELECT * FROM products WHERE ...
```

A few more of the options and details are in the View struct within `lib/ecto/migration.ex`. What do you think?

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -552,7 +552,7 @@ if Code.ensure_loaded?(Postgrex.Connection) do
 
     # DDL
 
-    alias Ecto.Migration.{Table, Index, Reference, Constraint}
+    alias Ecto.Migration.{Table, Index, Reference, Constraint, View}
 
     @drops [:drop, :drop_if_exists]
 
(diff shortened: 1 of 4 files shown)


josevalim:
I am :-1: because most of it is going to be SQL, so adding our own syntax is only going to make things more confusing. `execute/1` is your friend here.

michalmuskala:
I agree with @josevalim. Given that the main part will be SQL it won't be cross database anyway - I'm not sure I see advantage over a simple `execute/1`.

Also "prefix" is something that already has a meaning in Ecto as a schema/database. It would be confusing to introduce a different thing with the same name (at least it confused me, at first).

sorentwo:
Fair point regarding the bulk of it being raw SQL. It occurred to me after writing out that example that there isn't any practical way to use query building to construct the SQL, which negates the whole point. Thanks for weighing in!

Side note, I didn't intend to introduce `prefix` with an alternate meaning, I followed the other examples with a misunderstanding of the meaning =)

PR #1203, opened 2016-01-12
docs(Ecto.Schema): missing "can" on foreign key type

Gazler:


michalmuskala:
I believe the docs are incorrect in here, because the default type is actually `:id` and not `:integer`. Could you change it as well while you're at it?

michalmuskala:
Thank you! :heart:

PR #1206, opened 2016-01-17
Add support for prepared queries

fishcakez:
Add support for prepared by introducing `prepare_execute` adapter callback, and `prepare_execute` and `execute` sql connection callbacks.

Note there is a theoretical potential of hash collisions causing a query to always fail. Naming should switch to using `unique_integer` instead of a hash. Though we will need to ensure there are no race conditions so that the same integer is always used.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -58,6 +58,11 @@ defmodule Ecto.Adapters.SQL do
       def prepare(:delete_all, query), do: {:cache, @conn.delete_all(query)}
 
       @doc false
+      def prepare_execute(repo, meta, prepared, params, preprocess, opts) do
+        Ecto.Adapters.SQL.prepare_execute(repo, meta, prepared, params, preprocess, opts)
+      end
+
+      @doc false
       def execute(repo, meta, prepared, params, preprocess, opts) do
         Ecto.Adapters.SQL.execute(repo, meta, prepared, params, preprocess, opts)
       end
@@ -141,11 +146,17 @@ defmodule Ecto.Adapters.SQL do
   def to_sql(kind, repo, queryable) do
     adapter = repo.__adapter__
 
-    {_meta, prepared, params} =
+    plan =
       Ecto.Queryable.to_query(queryable)
       |> Ecto.Query.Planner.query(kind, repo, adapter)
-
-    {prepared, params}
+    case plan do
+      {:execute, _meta, %{statement: statement}, params} ->
+        {statement, params}
+      {:execute, _meta, prepared, params} ->
+        {prepared, params}
+      {:prepare_execute, _meta, prepared, params, _} ->
+        {prepared, params}
+    end
   end
 
   @doc """
@@ -199,10 +210,14 @@ defmodule Ecto.Adapters.SQL do
   end
 
   defp query(repo, sql, params, mapper, opts) do
+    sql_call(repo, :query, sql, params, mapper, opts)
+  end
+
+  defp sql_call(repo, callback, sql, params, mapper, opts) do
     {pool, default_opts} = repo.__pool__
     conn = get_conn(pool) || pool
     opts = [decode_mapper: mapper] ++ with_log(repo, params, opts ++ default_opts)
-    repo.__sql__.query(conn, sql, params, opts)
+    apply(repo.__sql__, callback, [conn, sql, params, opts])
   end
 
   ## Worker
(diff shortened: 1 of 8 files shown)


josevalim:
@fishcakez this is fantastic! Thank you! :heart:

lib/ecto/adapters/sql.ex
```diff
@@ -141,11 +146,17 @@ defmodule Ecto.Adapters.SQL do
...
     adapter = repo.__adapter__
 
-    {_meta, prepared, params} =
+    plan =
       Ecto.Queryable.to_query(queryable)
       |> Ecto.Query.Planner.query(kind, repo, adapter)
-
-    {prepared, params}
+    case plan do
+      {:execute, _meta, %{statement: statement}, params} ->
```
josevalim: Curiosity, why do we have the statement sometimes? Is this when it comes after prepare_execute? I am afraid this is being specific to PG. Can't we call DBCOnnection to string or something? Or can we get rid of the difference in the first place?

fishcakez: We only use it here when its not PG specific I think. Yes this should be to_string, the same as logging. Good catch.

lib/ecto/adapters/sql.ex
```diff
@@ -311,17 +326,50 @@ defmodule Ecto.Adapters.SQL do
...
+
+  def prepare_execute(repo, %{select: %{fields: fields}}, prepared, params, preprocess, opts) do
+    mapper = &process_row(&1, preprocess, fields)
+    {query, result} = sql_call!(repo, :prepare_execute, prepared, params, mapper, opts)
+    %{rows: rows, num_rows: num} = result
+    {query, {num, rows}}
+  end
+
+  @doc false
+  def execute(repo, _meta, %{} = prepared, params, nil, opts) do
```
fishcakez: Here we only match on map, when the query has been prepared and call execute. We are assuming the query is a struct because of DBConnection.Query. We could tag database prepared queries with a tuple to make the code clearer, e.g.`{:prepared, query}`.

lib/ecto/adapters/sql.ex
josevalim: I think there is a bit of confusion because we have three states and execute is representing two. The states are: planned for execute, planned for prepare_execute and prepared for execute. It also doesn't help that our adapter prepare is not the database prepare.

So I'd suggest at least one change: let's rename our adapter prepare to plan.

Regarding the double meaning in execute, should we introduce three APIs? execute_planned, prepare_and_execute_planned and execute_prepared.

Btw, I can work on those changes after merged. I'd just like to know if they make sense. :)

josevalim: Another option is to remove this knowledge from Ecto and pass it to the adapter completely.

1. The call to Ecto.Query.Planner would return a key indicating `:nocache` (execute_planned),  `:cache` (prepare_execute_planned), `:cached` (execute_prepared).

2. This would be sent to the adapter execute which would act accordingly. This way the adapter is a single function, decoupled from SQL semantics, and you can do as you wish from the adapter side. The only difference is that in the prepare_execute case, the adapter would have to call the anonymous function as we won't do it in the repository.

fishcakez: All of these sounds better. Yes the double meaning is awkward and I did try an approach without but I wanted to have minimal impact on the API. Shall I rework this to use unique integers and then you can refactor to a better API? The unique integers may impact the decision.

fishcakez: One issue we have that is confusing is that are so many different ways to prepare now

josevalim: Yes, please proceed as you where. :) Let me know once we are good to go!

fishcakez: I am happy to proceed.

fishcakez:
@josevalim I don't know if there was a miscommunication but please feel free to merge and improve/continue working on the the API. I think you would like to and I certainly would like you to!

josevalim:
No miscommunication. I just didn't have time to work on it today. :(

fishcakez:
> No miscommunication. I just didn't have time to work on it today. :(

No problem, I was just concerned I was blocking you when I didnt mean to be so wanted to be clear. When you finish your work on this could you tag postgrex 0.11?

josevalim:
Definitely! :D Thanks for all the work @fishcakez!

fishcakez:
Controlling this at the adapter level might be better because we may need to turn off prepared queries for those using pgbouncer in transaction mode.

PR #1209, opened 2016-01-20
improving schema docs for @primary_key

Tica2:
type (:id or :binary_id) is confusing (we can put more here) and 
also {:id, :id, autogenerate: true} not clear who is pk and who is type

I hope is more clear now.


josevalim:

PR #1210, opened 2016-01-20
Support for composite primary keys in schema

Tica2:
composite primary keys in schema
Fix https://github.com/elixir-lang/ecto/issues/135

supported version:

```elixir
defmodule HelloPhoenix.Player do
  use HelloPhoenix.Web, :model

  @primary_key false
  schema "players" do
    field :name, :string, primary_key: true
    field :position, :string, primary_key: true
    field :number, :integer

    timestamps
  end
  . . .
```
Feedback is Welcome.
Thanks!


lib/ecto/schema.ex
```diff
@@ -398,8 +401,9 @@ defmodule Ecto.Schema do
 
   """
   defmacro field(name, type \\ :string, opts \\ []) do
+    primary_key = Keyword.get(opts, :primary_key, false)
     quote do
-      Ecto.Schema.__field__(__MODULE__, unquote(name), unquote(type), false, unquote(opts))
```
josevalim: What if instead of passing `primary_key` as argument, we simply pass it as part of options and `__field__` can fetch it out? We just need to make sure to pass `primary_key: true` for the field in `@primary_key`.

josevalim:
Really, really nice. I have added a comment!

lib/ecto/schema.ex
Tica2: I do not see the value,

Default @primary_key 
use explicit call:
Ecto.Schema.__field__(__MODULE__, name, type, true, opts)
https://github.com/elixir-lang/ecto/pull/1210/files#diff-cb8b14078d10148b5f2f3f8635c75892R342

is good this signature for __field__

michalmuskala:
I'm not so sure about supporting `Repo.get(schema, keys)` with a list. It's not clear what order the values should be in.

I think it would be best to support only single-pk models with `get` and promote use of `get_by` for composite keys.

josevalim:
@tica2 @michalmuskala both very good points.

Tica2:
order defined in schema,
`get_by` can be used anyway ...

josevalim:
@tica2 let me know when this is ready for merge. :)

Tica2:
@josevalim This part for schema is ready.
For migrations is more work to do because need to check more adapters. (pg, mysql first)

josevalim:
Ok, I would like to merge them altogether though. So let me know if I can be some help on the migrations bit. :)

Tica2:
I start to work for migrations with composite keys.
I think we can close https://github.com/elixir-lang/ecto/issues/135 and open new one for migrations, but is not require.

Tica2:
Support for composite primary keys in migrations done!
Is working for pg.

I think mysql is disable.
When try to enable I have one error:
`** (UndefinedFunctionError) undefined function Ecto.Adapters.MySQL.Connection.connection/1`

josevalim:
MySQL is disabled but the unit tests are still available. Just add one to test/ecto/adapters/mysql_test  similar to how you did to postgres and we are good to go!

Tica2:
Mysql migrations for composite primary keys is also ready.

josevalim:

PR #1216, opened 2016-01-23
Do not specify timezone in iso8601

lau:
datetime function. We do not know if the datetime
is in UTC or not, so we should not assume that it is.

I noticed that inspect also uses the iso8601 function and adds a Z.
This signals to the world that the datetime is UTC even though we do not know if it is or not.

### DIFF
--- a/lib/ecto/date_time.ex
+++ b/lib/ecto/date_time.ex
@@ -520,13 +520,8 @@ defmodule Ecto.DateTime do
   end
 
   @doc """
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you! The docs were pretty clear though that it expects a datetime in UTC as argument. So I think we should keep the iso function as is BUT we should change `inspect` to use to_string. Thoughts?

josevalim:
We don't need the timezone specification, so let's remove it.

PR #1219, opened 2016-01-26
Correct comment on preload example [ci skip]

derekprior:
The existing comment does not entirely match what the code does.


michalmuskala:
Thank you! :heart:

PR #1221, opened 2016-01-27
Include :invalid in the typespec for Changeset.cast/4

paulcsmith:
I ran dialyzer and almost all of my changeset functions threw errors because `:invalid` wasn't in the typespec. Since you can pass `:invalid` as the second argument, I think it makes sense to add it to the typespec.

I'm new to dialyzer and typespecs though so I may have gotten this wrong. I think this is the correct way to solve the issue though


sorentwo:
I'm not a dialyzer pro, but that looks right to me. :+1:

josevalim:

PR #1230, opened 2016-02-01
Add supervisor as top process to builtin pools

### LINKED ISSUE #1208: Top pool process is always supervisor which is a bug
As per IRC discussion with @fishcakez :
the top pool process may or may not be a supervisor. It should be pool dependent. Currently it is always listed as a supervisor in the children of the repo supervisor. Its a bug.
https://github.com/elixir-lang/ecto/blob/v1.1/lib/ecto/repo/supervisor.ex#L115

`fishcakez_`: chvanikoff: yeah, we need the pool to define the child spec to get it right.
`micmus`: fishcakez_: most probably we should go to the adapter level, not all adapters are using ecto's pools (at least now)
`fishcakez_` micmus: yeah it would be the adapter level, and the sql adapter would pass it to the pool
`fishcakez_` the repo suoervisor has no ideal about pooling

fishcakez:
Closes #1208


fishcakez:
Just as a side note the Mongo pool already has a supervisor and does not have this issue on any version.

josevalim:

PR #1234, opened 2016-02-02
A typo in assoc.exs integration test?

DNNX:
I don't know whether it is a copy&paste error or the titles of the two posts were made the same intentionally.

If the same titles is the intention and posts having same titles cover some weird edge-case regression, then feel free to reject this PR and close it.

If it wasn't the intention, then I think it is worth merging in order not to confuse readers like me :) 

What do you think?


josevalim:

PR #1245, opened 2016-02-11
Warn when migrations directory is not found

### LINKED ISSUE #1239: Warn when configured priv directory does not exist
Now when the directory configured by the specified :priv key under the repository configuration does not exist running Ecto mix tasks gives no hint about that. Example:

`#project$ mix ecto.rollback --all --repo Project.Repo2; mix ecto.migrate --all --repo Project.Repo2`
`10:03:03.030 [info]  Already down`
`10:03:03.702 [info]  Already up`

That's IMHO nearly a silent fail... I think there should be a check if the given directory exists and at least some warning (if not even an error) message should inform about this situation.

(I can offer to implement the solution (e.g. `ensure_repo_priv` and `ensure_migrations_path` functions or a direct directory existence check in the `repo_priv` and `migrations_path` functions) if that might be of any help)

tokafish:
Closes #1239

### DIFF
--- a/lib/mix/ecto.ex
+++ b/lib/mix/ecto.ex
@@ -23,7 +23,7 @@ defmodule Mix.Ecto do
   defp parse_repo([], []) do
     if app = Keyword.get(Mix.Project.config, :app) do
       case Application.get_env(app, :app_repo) do
-        nil -> 
+        nil ->
           case Application.get_env(app, :app_namespace, app) do
             ^app -> app |> to_string |> Mix.Utils.camelize
             mod  -> mod |> inspect
@@ -94,6 +94,19 @@ defmodule Mix.Ecto do
     Application.ensure_all_started(app)
   after
     Logger.add_backend(:console, flush: true)
+  end
+
+  @doc """
+  Ensures the given repository's migrations path exists on the filesystem.
+  """
+  @spec ensure_migrations_path(Ecto.Repo.t) :: Ecto.Repo.t | no_return
+  def ensure_migrations_path(repo) do
+    path = Path.relative_to(migrations_path(repo), Mix.Project.app_path)
+    if File.dir?(path) do
+      repo
+    else
+      Mix.raise "could not find migrations directory #{inspect path} for repo #{inspect repo}"
+    end
   end
 
   @doc """
--- a/test/mix/tasks/ecto.migrate_test.exs
+++ b/test/mix/tasks/ecto.migrate_test.exs
@@ -2,6 +2,14 @@ defmodule Mix.Tasks.Ecto.MigrateTest do
   use ExUnit.Case
 
   import Mix.Tasks.Ecto.Migrate, only: [run: 2]
+  import Support.FileHelpers
+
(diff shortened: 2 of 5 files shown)


lib/mix/ecto.ex
```diff
@@ -97,6 +97,19 @@ defmodule Mix.Ecto do
   end
 
   @doc """
+  Ensures the given repository's migrations path exists on the filesystem.
+  """
+  @spec ensure_migrations_path(Ecto.Repo.t) :: Ecto.Repo.t | no_return
+  def ensure_migrations_path(repo) do
+    path = Path.relative_to(migrations_path(repo), Mix.Project.app_path)
```
josevalim: Just use migrations_path(repo) and use Path.relative_to_cwd when building the error message. :)

test/mix/tasks/ecto.rollback_test.exs
```diff
@@ -2,6 +2,15 @@ defmodule Mix.Tasks.Ecto.RollbackTest do
   use ExUnit.Case
 
   import Mix.Tasks.Ecto.Rollback, only: [run: 2]
+  import Support.FileHelpers
+
+  migrations_path = Path.join([tmp_path, inspect(Ecto.Migrate), "migrations"])
+
+  setup do
+    File.rm_rf!(unquote(migrations_path))
```
josevalim: No need to remove it given we just need to check if it exists. :)

test/mix/tasks/ecto.migrate_test.exs
```diff
@@ -2,6 +2,15 @@ defmodule Mix.Tasks.Ecto.MigrateTest do
   use ExUnit.Case
 
   import Mix.Tasks.Ecto.Migrate, only: [run: 2]
+  import Support.FileHelpers
+
+  migrations_path = Path.join([tmp_path, inspect(Ecto.Migrate), "migrations"])
+
+  setup do
+    File.rm_rf!(unquote(migrations_path))
```
josevalim: No need to remove it given we just need to check if it exists. :)

josevalim:
Thank you. I have added some comments!

lib/mix/ecto.ex
tokafish: @josevalim thanks! That was my first thought as well, but I noticed that (within the tests, at least) `migrations_path` was returning `_build/shared/lib/ecto/tmp/Ecto.Migrate/migrations`, instead of just `tmp/Ecto.Migrate/migrations` like I expected.

I believe this is due to [this line](https://github.com/elixir-lang/ecto/blob/master/lib/mix/ecto.ex#L113). I noticed that gen.migration [uses](https://github.com/elixir-lang/ecto/blob/master/lib/mix/tasks/ecto.gen.migration.ex#L43) Mix.Project.app_path, so I thought that would be the way to proceed.

I'm happy to make any adjustments, but I could use some guidance on the correct path.



Application.app_dir

josevalim: You are right, this looks good to go!

josevalim:

PR #1258, opened 2016-02-19
fix(Mix.Ecto): ignore ensure_migrations in an umbrella project

Gazler:
Previously attempting to run `mix ecto.migrate` or `mix test` which runs
`mix ecto.migrate` from the root of an umbrella app would result in the
following error:

 > ** (RuntimeError) Trying to access Mix.Project.app_path for an
 > umbrella project but umbrellas have no app

This is caused by calling:

    Path.relative_to(migrations_path(repo), Mix.Project.app_path)

Which was introduced in 9b5ede7 to
ensure the migrations directory exists. This check is now ignored in an
umbrella application.


josevalim:

PR #1261, opened 2016-02-20
Potential regression on skipping preload?

mitchellhenke:
I wrote a test below that is currently failing for me on master and 2.0.0-beta.0, but passes on v1.1.3.  I didn't see anything in the changelog about it, so I wasn't sure if this was an intended change or not?

Edit:
If it helps, it looks the test passes if I set `author_id` and `post_id`:
`c1 = %{c1 | author: u1, author_id: u1.id, post: p1, post_id: p1.id}`

Willing to take a go at fixing it if this is indeed not the intended functionality.

### DIFF
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -442,6 +442,18 @@ defmodule Ecto.Integration.PreloadTest do
     assert u.custom.bid == c.bid
   end
 
+  test "skip preloads with association set but without id set" do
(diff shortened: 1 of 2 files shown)


josevalim:
That sounds like a regression. Let's give the PR a try. :)

mitchellhenke:
Sounds good, will do!

mitchellhenke:
I reordered and changed the conditional logic so all tests pass now.  I'm guessing there is a better way to format all of it though?

PR #1266, opened 2016-02-21
Clarify the purpose of validate_required

VladShcherbin:


michalmuskala:
I'm not sure this makes it any clearer. Aren't all validation functions intended for validation of user input?

josevalim:
As @michalmuskala, all of the validation functions are meant for user input, they are not meant to enforce business rules. That's the whole goal of changesets casting altogether. It is not something specific here. :)

PR #1269, opened 2016-02-21
Add call to :calendar.valid_date to cast/1 in Ecto.Date and Ecto.DateTime

Theemuts:
This fixes the issue where dates which don't exist (like Februari 29th, 2015) are cast successfully to Ecto.Date and Ecto.DateTime structs.


josevalim:

PR #1292, opened 2016-03-05
Improve error when a variable is accidentally passed to assoc/2

parkerl:
I mess this up all the time. Currently you get

```
** (Ecto.Query.CompileError) expected literal atom or interpolated value in field/2, got: `{:some_assoc, [line: 405], nil}`
```

which always surprises me since I wasn't using `field` in my query. Then I have to track it down (for the 12th time). 

This PR tries to improve on that for slow learners like me by returning:

```
You passed the variable some_assoc to `assoc/2`. Did you mean to pass the atom :some_assoc?
```

### DIFF
--- a/lib/ecto/query/builder/join.ex
+++ b/lib/ecto/query/builder/join.ex
@@ -68,6 +68,7 @@ defmodule Ecto.Query.Builder.Join do
 
   def escape({:assoc, _, [{var, _, context}, field]}, vars, _env)
       when is_atom(var) and is_atom(context) do
+    ensure_field!(field)
     var   = Builder.find_var!(var, vars)
     field = Builder.quoted_field!(field)
     {:_, nil, {var, field}, %{}}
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder/join.ex
```diff
@@ -190,4 +191,9 @@ defmodule Ecto.Query.Builder.Join do
       "invalid join qualifier `#{inspect qual}`, accepted qualifiers are: " <>
       Enum.map_join(@qualifiers, ", ", &"`#{inspect &1}`")
   end
+
+  defp ensure_field!({var, _, _}) when var != :^ do
+    Builder.error! "You passed the variable #{var} to `assoc/2`. Did you mean to pass the atom :#{var}?"
```
josevalim: error messages start with lowercase :) Also please wrap `#{var}` in backticks.

parkerl:
@josevalim yup, I'm a bit rusty. ;-) Should the second sentence be lower case as well?

josevalim:
No, just the leading one. :)

parkerl:
@josevalim updated.

josevalim:

PR #1293, opened 2016-03-05
Attempt to clarify to schema vs migration field types

parkerl:
At least 3 times last week issues came up on the slack channel due to confusion between Ecto types and types that can be used in a migration. I have attempted to raise the visibility of this in the documentation for both Ecto.Migration and Ecto.Schema. Hope it helps.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -64,6 +64,18 @@ defmodule Ecto.Migration do
   Notice not all commands are reversible though. Trying to rollback
   a non-reversible command will raise an `Ecto.MigrationError`.
 
+  ## Field Types
+
+  The Ecto primitive types are mapped to the appropriate database
+  type by the various database adapters. For example, `:string` is converted to
+  `:varchar`, `:datetime` to the underlying `:datetime` or `:timestamp` type,
+  `:binary` to `:bits` or `:blob`, and so on.
+
+  Similarly, you can pass any field type supported by your database
+  as long as it maps to an Ecto type. For instance, you can use `:text`,
+  `:varchar` or `:char` in your migrations as `add :field_name, :text`.
+  In your Ecto schema, they will all map to the same `:string` type.
+
   ## Prefixes
 
   Migrations support specifying a table prefix or index prefix which will target either a schema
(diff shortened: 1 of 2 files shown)


lib/ecto/migration.ex
```diff
@@ -64,6 +64,17 @@ defmodule Ecto.Migration do
   Notice not all commands are reversible though. Trying to rollback
   a non-reversible command will raise an `Ecto.MigrationError`.
 
+  ## Field Types
+
+  The Ecto primitive types are mapped to the appropriate database
+  type by the various database adapters. For example, `string` is converted to
```
josevalim: `:string` is converted to `:varchar`, `:datetime` ...

lib/ecto/migration.ex
```diff
@@ -64,6 +64,17 @@ defmodule Ecto.Migration do
...
+  varchar, `datetime` to the underlying datetime or timestamp type,
+  `binary` to bits or blob, and so on.
+
+  In addition to these mappings, any field type supported by your database
```
josevalim: What about this paragraph be something like:

> Similarly, you can pass any field type supported by your database
> as long as it maps to an Ecto type. For instance, you can use `:text`,
> `:varchar` or `:char` in your migrations as `add :field_name, :text`.
> In your Ecto schema, they will all map to the same `:string` type.

Thoughts?

lib/ecto/migration.ex
parkerl: @josevalim I stole this from the docs for `add`. Should I update those as well?

lib/ecto/migration.ex
parkerl: Love it.

parkerl:
@josevalim updated.

lib/ecto/schema.ex
```diff
@@ -137,6 +137,13 @@ defmodule Ecto.Schema do
   **Note:** For the `:array` type, replace `inner_type` with one of
   the valid types, such as `:string`.
 
+  ### Database types
+
+  In migrations the Ecto primitive types shown above are mapped to the
+  appropriate database type by the various database adapters.
+  However, Ecto allows for the use of any type supported by your
+  database. See the documentation for Ecto.Migration.
```
josevalim: `Ecto.Migration`

josevalim:
Yes, let's fix both! Thank you for all of these updates! :)

parkerl:
Will fix shortly

parkerl:
@josevalim updated, ready for review.

josevalim:

PR #1294, opened 2016-03-05
Refer to schema rather than model

parkerl:
Are we still referring to a module where Ecto.Schema is used as a model? If so then this change may not make sense. I was trying to make the docs very specific as to where the types are used. No hard feelings if it is not a good change.


josevalim:
Thank you! I believe however we should just call it `schema` and not `schema function`, no?

parkerl:
That is fine with me. Just trying to be specific. I will update.

parkerl:
@josevalim updated.

josevalim:

PR #1313, opened 2016-03-12
Sets the binding key for `validate_number` errors to `number`.

### LINKED ISSUE #1308: Bad Gettext interpolation keys when validating numbers
### Environment

* Ecto version (mix deps): 2.0.0-beta.2

### Current behavior

In Phoenix, we translate Ecto errors using `Gettext.dngettext/6`. In the [relevant snippet of code](https://github.com/phoenixframework/phoenix/blob/f43f8a7fc4101bfe271dfadd6f378b3ed27df741/installer/templates/new/web/views/error_helpers.ex#L29):

```elixir
  def translate_error({msg, opts}) do
    # Because error messages were defined within Ecto, we must
    # call the Gettext module passing our Gettext backend. We
    # also use the "errors" domain as translations are placed
    # in the errors.po file. On your own code and templates,
    # this could be written simply as:
    #
    #     dngettext "errors", "1 file", "%{count} files", count
    #
    Gettext.dngettext(MyApp.Gettext, "errors", msg, msg, opts[:count], opts)
  end
```

So, we pass `opts[:count]` as the 5th argument to `Gettext.dngettext/6`: according to [the documentation for `Gettext.dngettext/6`](https://hexdocs.pm/gettext/Gettext.html#dngettext/6), that argument should be a `non_neg_integer` (as it's the number of items that will allow gettext to decide how to pluralize the translation).

In Ecto, we use the following error messages to validate numbers (code is [here](https://github.com/elixir-lang/ecto/blob/2d1a2eecdc691c86aa98f261e9eddb880a7c411c/lib/ecto/changeset.ex#L164-L170)):

```elixir
  @number_validators %{
    less_than:                {&</2,  "must be less than %{count}"},
    greater_than:             {&>/2,  "must be greater than %{count}"},
    less_than_or_equal_to:    {&<=/2, "must be less than or equal to %{count}"},
    greater_than_or_equal_to: {&>=/2, "must be greater than or equal to %{count}"},
    equal_to:                 {&==/2, "must be equal to %{count}"},
  }
```

Here, we use `count` as the interpolation key for the value of the number field we're validating. However, `validate_number/3` also supports `Decimal`s and floats, and if there's an error, they end up being passed in the bindings (i.e., `[count: #Decimal<1.2>]`).

All of this amounts to a `FunctionClauseError` when `translate_error/1` from the Phoenix error helper is called with `opts = [count: x, ...]` with `x` being a non-integer.

### Expected behavior

This should not fail, of course :smile_cat:.

### Proposed solution

My proposed solution is to change the error messages in Ecto to not use `count` as the interpolation key (which I also believe being misleading but that's another story :grin:), and replace it with something like `number` instead. Then, in the Phoenix error helper, we could do:

```elixir
Gettext.dngettext(MyApp.Gettext, "errors", msg, msg, opts[:count] | | 1, opts)
```

so that we're safe from when `:count` is not in the passed options; this makes sense to me as many times (and number validation is one of these cases) we use plural translations just as a workaround, but we never actually use the `msgid_plural`.

I'd be happy to work on this if you want :)

amencarini:
Given that there is not going to be a singular and plural version of
the errors related to `validate_number`, this changes the relevant key,
in order for this to be handled as a singular translation with a binding
rather than a singular/plural in Gettext.

Fixes #1308

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1050,7 +1050,7 @@ defmodule Ecto.Changeset do
     new_errors =
       for field <- fields,
           missing?(changeset, field),
-          do: {field, message}
+          do: {field, {message, []}}
 
     case new_errors do
       [] -> %{changeset | required: fields ++ required}
@@ -1086,7 +1086,7 @@ defmodule Ecto.Changeset do
   @spec validate_format(t, atom, Regex.t, Keyword.t) :: t
   def validate_format(changeset, field, format, opts \\ []) do
     validate_change changeset, field, {:format, format}, fn _, value ->
-      if value =~ format, do: [], else: [{field, message(opts, "has invalid format")}]
+      if value =~ format, do: [], else: [{field, {message(opts, "has invalid format"), []}}]
     end
   end
 
@@ -1106,7 +1106,7 @@ defmodule Ecto.Changeset do
   @spec validate_inclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_inclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:inclusion, data}, fn _, value ->
-      if value in data, do: [], else: [{field, message(opts, "is invalid")}]
+      if value in data, do: [], else: [{field, {message(opts, "is invalid"), []}}]
     end
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you!

With this change though, I think it would be best if we normalized all error fields for all validations to be in the format of:

    {message, opts}

`opts` is going to be an empty list for most validations but at least it would be consistent. In any case, when `:message` is given, it still must be only a string, like in `validate_required(changeset, :name, message: "must be given")`.

With all that said. Could you please update your PR to also change the other validations? You don't need to amend, it can be a new commit on top of the existing one. Thank you! :)

amencarini:
Ah, yes, I initially misinterpreted the message you left on the issue. I changed all validation errors, let me know if that needs further amendments!

josevalim:

PR #1320, opened 2016-03-16
Fix insert all to work like the docs when specifying prefix

samterrell:
`insert_all` wasn't working like described in https://github.com/elixir-lang/ecto/blob/master/lib/ecto/repo.ex#L429 with regards to sending a `{"prefix", "source"}` tuple.
> It expects a schema (`MyApp.User`) or a source (`"users"` or
> `{"prefix", "users"}`) as first argument.

I'm not really sure what the design of some of this is, so let me know if I did it wrong. I *could* make a pickier gaurd with matching, if that would make a more friendly error.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -18,6 +18,10 @@ defmodule Ecto.Repo.Schema do
     do_insert_all(repo, adapter, nil, {nil, table}, rows, opts)
   end
 
+  def insert_all(repo, adapter, {_prefix, _source} = table, rows, opts) do
+    do_insert_all(repo, adapter, nil, table, rows, opts)
+  end
+
   defp do_insert_all(_repo, _adapter, _schema, _source, [], opts) do
     if opts[:returning] do
       {0, []}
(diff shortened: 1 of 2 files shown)


lib/ecto/repo/schema.ex
```diff
@@ -18,6 +18,10 @@ defmodule Ecto.Repo.Schema do
     do_insert_all(repo, adapter, nil, {nil, table}, rows, opts)
   end
 
+  def insert_all(repo, adapter, table, rows, opts) when is_tuple(table) do
```
josevalim: Can we check for `{_prefix, _source} = table` instead of the `is_tuple` guard? Thank you! :heart:

samterrell: I had thought about that, too. Maybe even checking `is_binary` to match the previous.

and thank you man! You're my hero.

josevalim:
Beautiful, can you squash all changes into a single commit? Then I will merge it right away!

samterrell:
Sure thing.

josevalim:

PR #1325, opened 2016-03-16
Update cast and relation errors to tuples

amencarini:
This follows up the changes applied to error messages to accommodate for a keyword list together with a string for easier localization of errors.


josevalim:

wojtekmach:
Does it mean that all errors are now always held as `{message, opts}`? If so, I think we need to update docs for `Changeset.traverse_errors/2` https://github.com/elixir-lang/ecto/blob/master/lib/ecto/changeset.ex#L1744 along with typespecs. I'm happy to send a patch

PR #1333, opened 2016-03-22
Correct calculation of elapsed time during migration tasks

ftaebi:
:timer.tc returns the elapsed time in microseconds, so the current value is wrong by a factor of 10.


josevalim:

PR #1334, opened 2016-03-23
Ecto.Multi documentation errors

mgwidmann:
Also noticed that `to_list/1` was missing documentation, so I added something. If theres more you think we should say just let me know.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -20,9 +20,9 @@ defmodule Ecto.Multi do
   ## Changesets
 
   If Multi contains operations that accept changesets (like `insert/4`,
-  `update/4` or `delete/4`) they will be checked before starting transaction
-  if any of them isn't invalid. In case any has errors, the transaction won't
-  even be started and the error will be immediately returned.
+  `update/4` or `delete/4`) they will be checked before starting the transaction.
+  If any changeset has errors, the transaction won't even be started and the error
+  will be immediately returned.
 
   ## Run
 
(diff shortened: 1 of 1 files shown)


lib/ecto/multi.ex
```diff
@@ -65,7 +65,7 @@ defmodule Ecto.Multi do
           # under keys we used for naming the operations.
         {:error, failed_operation, failed_value, changes_so_far} ->
           # One of the operations failed. We can access the operation's failure
-          # value (like changeset for operations on changesets) to prepare
+          # value (like changeset for operations on changesets) to prepare a
           # proper response. We also get access to results of any operations
           # that succeeded before the indicated operation failed.
```
mgwidmann: I was also curious if it would be worth mentioning that the successful operations were rolled back? This makes it seem like maybe it wasn't run in a transaction, for the DBs that support transactions obviously.

For DBs like Mongo, I suspect these would be the successful operations that actually did succeed and the developer would be responsible for cleaning up at this point.

lib/ecto/multi.ex
```diff
@@ -305,6 +305,10 @@ defmodule Ecto.Multi do
     end
   end
 
+  @doc """
+  Transforms the `Ecto.Multi` into a list of operations to be performed. Inspecting
```
josevalim: @michalmuskala any reason why we didn't document this?

michalmuskala: I think it was my mistake. It should be documented & public API, we even mention it in moduledocs.

lib/ecto/multi.ex
michalmuskala: I think it's reasonable to mention it.

Mongo won't be using `Repo.transaction` in the first place, so we don't need to be concerned by that here.

lib/ecto/multi.ex
mgwidmann: Thats why I added it :smile: 

Anything you'd like to add?

josevalim:
@michalmuskala please merge it at your discretion. :)

michalmuskala:
Thank you! :heart:

PR #1336, opened 2016-03-25
Added custom message option to cast_assoc and cast_embed

acarpe:


josevalim:
Thank you @acarpe. Because we have both "can't be blank" and "is invalid" messages, what if we support `:on_blank` and `:on_invalid` parameters for messages? @michalmuskala thoughts on how the options should be named considering our existing API?

michalmuskala:
How about `blank_message` and `invalid_message`? I think it fits better into other functions where we use `message` for the validation information.

josevalim:
Let's go with `required_message` and `invalid_message`. That would also mirror the `validate_required` bit. @acarpe can you please amend your commit accordingly? Thank you!

acarpe:
Changes done, you can review it :)

lib/ecto/changeset.ex
```diff
@@ -511,10 +515,10 @@ defmodule Ecto.Changeset do
     end
   end
 
-  defp missing_relation(%{changes: changes, errors: errors} = changeset, name, current, required?, relation) do
+  defp missing_relation(%{changes: changes, errors: errors} = changeset, name, current, required?, relation, opts) do
     current_changes = Map.get(changes, name, current)
     if required? and Relation.empty?(relation, current_changes) do
-      %{changeset | errors: [{name, {"can't be blank", []}} | errors], valid?: false}
+      %{changeset | errors: [{name, {(opts[:required_message] || "can't be blank"), []}} | errors], valid?: false}
```
michalmuskala: Should we change `message/2` to accept additional message key?
````
defp message(opts, key \\ :mesage, default)
```

acarpe:
@michalmuskala modified the message function.

michalmuskala:
:shipit:

josevalim:

PR #1337, opened 2016-03-27
Add ON_UPDATE support in migrations

stavro:
Piggybacking off of https://github.com/elixir-lang/ecto/pull/705

Ecto currently supports adding `ON DELETE` column constraints in migrations. 

Analogous to `ON DELETE` there is also `ON UPDATE` which is invoked when a referenced column is changed (updated). The possible actions are the same as `ON DELETE`.

This PR supports `ON UPDATE` on column definitions.

Would love Ecto to support this!

:heart: :purple_heart: :blue_heart: :green_heart:

Thoughts?

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -71,7 +71,7 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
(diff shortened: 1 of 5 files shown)


josevalim:

PR #1338, opened 2016-03-27
Adds inherited table support for Ecto migrations.

kipcole9:
Adds syntax to the table() macro to permit the creation of inherited tables
for adapters that support that functionality (currently limited to Postgres).

Given the migration:

    defmodule InheritedMigration do
      use Ecto.Migration

      def up do
        create table(:things) do
          add :name, :string
        end

        create table(:posts, inherits: :things) do
          add :content, :text
        end
      end

      def down do
        drop table(:posts)
        drop table(:things)
      end
    end

Then the table :posts will be created and inherit the columns from :things
as well as having the additional columns defined on :posts.  This is the
native Postgres behaviour of table inheritance.

This native behaviour does not inherit the primary key definition, doesn't
inherit any indexes and doesn't inherit any triggers.

By default, this commit will copy the primary key definition, indexes and
triggers from the inherited table to the new table since this is the likely
expected behaviour.

To modify the behaviour, the inheritance can be given options:

    create table(:posts, inherits: [:things, primary_keys: false, indexes: false, triggers: false) do
      add :content, :text
    end

This configuration will reproduce the default Postgres behaviour.  Each option
can of course be set separately.

Tests have been added and the tests specifically related to inheritance can be
run by:

    mix test --only inheritance

This is the first of three pull requests, the next two build upon this one:

* Add Ecto.Schema support for inherited tables that will copy the fields,
primary keys and associations from another `schema`.

* Eco.Query support that will polymorphically assign retrieved rows to the
appropriate `schema`

### DIFF
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -37,6 +37,11 @@ defmodule Ecto.Adapter.Migration  do
   Checks if the adapter supports ddl transaction.
   """
   @callback supports_ddl_transaction? :: boolean
+  
+  @doc """
+  Checks if the adapter supports inherited tables
+  """
+  @callback supports_inherited_tables? :: boolean
 
   @doc """
   Executes migration commands.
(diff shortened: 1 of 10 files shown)


josevalim:
Closing in favor of #1340.

PR #1339, opened 2016-03-27
Remove old code for `Changeset.traverse_errors/2` tests

wojtekmach:
Also, call the fn argument `msg` everywhere, since this is how it's called in docs.

### DIFF
--- a/test/ecto/changeset/embedded_test.exs
+++ b/test/ecto/changeset/embedded_test.exs
@@ -702,13 +702,10 @@ defmodule Ecto.Changeset.EmbeddedTest do
       |> Changeset.cast_embed(:profile)
       |> Changeset.add_error(:name, "is invalid")
 
-    errors = Changeset.traverse_errors(changeset, fn
-      {err, opts} ->
-        err
-        |> String.replace("%{count}", to_string(opts[:count]))
-        |> String.upcase()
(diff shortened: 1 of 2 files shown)


test/ecto/changeset/embedded_test.exs
```diff
@@ -726,13 +723,10 @@ defmodule Ecto.Changeset.EmbeddedTest do
...
-      {err, opts} ->
-        err
-        |> String.replace("%{count}", to_string(opts[:count]))
-        |> String.upcase()
-      err ->
-        String.upcase(err)
+    errors = Changeset.traverse_errors(changeset, fn {msg, opts} ->
+      msg
+      |> String.replace("%{count}", to_string(opts[:count]))
+      |> String.upcase()
```
wojtekmach: see that the this custom function (`msg_fun` in traverse_errors docs) is duplicated in many tests and in the docs. 

Perhaps it would be worth it, to supply it as the default one for `traverse_errors/2`? It wouldn't catch all possible cases, but it could be a good starting point. Perhaps it could be just do `String.replace` for all opts (`count` etc) that Ecto ships with?

josevalim: Nah, it is alright. :)

PR #1340, opened 2016-03-28
Eco.Schema support inherited tables

kipcole9:
Adds schema support for inherited tables.  Built on the previous pull request that adds migration support for inherited tables so this PR can be accepted instead of the first one if adding schema support for inherited tables is also accepted.

## Feature additions

There are three macros added to Ecto.Schema:

* `inheritable` marks a schema as inheritable by defining the field `:_type` that
is is a column alias for tableoid::regclass::text (more on column alias' later)

* `inherit` copies the fields and associations from another schema module to
this one.  Since all attributes of fields are copied, including the definition
of the primary key, setting `@primary_key false` on inherited schemas would
normally be appropriate.  `inherit` would normally by invoked on a schema
that inherits from a schema marked as `inheritable`

* `include` is exactly the same as `inherits` but expresses a different
intent - it would normally be invoked on a schema that is not marked
as `inheritable`

For example:

    defmodule Thing do
      use Ecto.Schema

      schema "things" do
        inheritable
        field :name, :string
      end
    end

    defmodule Post do
      use Ecto.Schema

      schema "posts" do
        inherit Thing
        field :content, :text
      end
    end

In summary, `includes` and `inherits` copy field and association definitions
from another schema module.  If the inherited schema includes a `:_type`
field (as it would if inheriting a schema marked as `inheritable`)
then Ecto.Query will in a default select retrieve a `:_type` field that is
the name of the underlying table from which a row is derived.  Since
queries on inherited tables can retrieve rows from multiple different
but related tables this provides a way to know from which underlying
table the row was sourced.

## Additional option to the `field` macro

The macro field gains a new option `:alias_for` which will be used by
Ecto.Query to substitute an expression which is then aliased to the field name.
The `:_type` field is defined this way:

    defmodule Thing do
      use Ecto.Schema

      schema "things" do
        field :_type, :string, alias_for: fragment("%{table}.\"tableoid\"::regclass::text\")
      end
    end

The expression for an alias can be anything that is accepted by Ecto.Query.select() except
that parameters are not supported.

There is one supported interpolation `%{table}` that will be substituted with the table
alias at query exection time.

For example:

    iex> query = from t in Thing
    iex> Repo.all query

    SELECT t0."tableoid"::regclass::text AS _type, .... FROM "things"

This technique can also be used to include SQL functions as part of a field
definition although clearly such fields cannot be updated and hence should not
be included in a changeset.

    defmodule Other do
      use Ecto.Schema

      schema "things" do
        field :first_name
        field :second_name
        field :full_name, :string, alias_for: fragment(coalesce("%{table}.first_name, ' ', %{table}.last_name))
      end
    end

## Tests

Tests are defined with `@tag :inheritance` and can be run by:

    mix test --only inheritance

### DIFF
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -37,6 +37,11 @@ defmodule Ecto.Adapter.Migration  do
   Checks if the adapter supports ddl transaction.
   """
   @callback supports_ddl_transaction? :: boolean
+  
+  @doc """
+  Checks if the adapter supports inherited tables
+  """
+  @callback supports_inherited_tables? :: boolean
 
   @doc """
   Executes migration commands.
--- a/lib/ecto/adapters/mysql.ex
+++ b/lib/ecto/adapters/mysql.ex
@@ -162,6 +162,11 @@ defmodule Ecto.Adapters.MySQL do
   def supports_ddl_transaction? do
     true
   end
+  
+  @doc false
+  def supports_inherited_tables? do
+    false
+  end
 
   @doc false
   def insert(repo, %{source: {prefix, source}, autogenerate_id: {key, :id}}, params, [key], opts) do
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -140,4 +140,9 @@ defmodule Ecto.Adapters.Postgres do
   def supports_ddl_transaction? do
     true
   end
+  
+  @doc false
+  def supports_inherited_tables? do
+    true
+  end
 end
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -384,9 +384,9 @@ if Code.ensure_loaded?(Postgrex) do
       "$#{ix+1}"
     end
 
-    defp expr({{:., _, [{:&, _, [idx]}, field]}, _, []}, sources, _query) when is_atom(field) do
-      {_, name, _} = elem(sources, idx)
-      "#{name}.#{quote_name(field)}"
+    defp expr({{:., _, [{:&, _, [idx]}, field]}, _, []}, sources, query) when is_atom(field) do
+      {_table, name, schema} = elem(sources, idx)
+      field_or_alias(field, name, sources, schema, query)
     end
 
     defp expr({:&, _, [idx, fields, _counter]}, sources, query) do
@@ -397,9 +397,13 @@ if Code.ensure_loaded?(Postgrex) do
           "Please specify a schema or specify exactly which fields from " <>
           "#{inspect name} you desire")
       end
-      Enum.map_join(fields, ", ", &"#{name}.#{quote_name(&1)}")
+      Enum.map_join(fields, ", ", &field_or_alias(&1, name, sources, schema, query))
     end
 
+    defp expr({:{}, [], fragment}, sources, query) do
+      expr(List.to_tuple(fragment), sources, query)
+    end
+    
     defp expr({:in, _, [_left, []]}, _sources, _query) do
       "false"
     end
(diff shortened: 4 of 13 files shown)


josevalim:
Thanks @kipcole9!

I am honestly a bit worried about us doing the opposite of PG in the migration case by automatically inheriting PKs, indexes and friends. This could give developers the impression those tables are automatically kept in sync which is not true (if you change those definitions in the parent, the child one won't change). So that's a :-1: from my side.

I would expect developers to get the inheritance behaviour today by doing:

    create table(:bar, options: "INHERITS(foo)") do
      ...
    end

I will investigate the rest of the PR. `alias_for` is something we wanted to add one way or the other but I am afraid the current definition will push too much database specific logic to the schema.

kipcole9:
José, no problem.

alias_for just uses the same path as select() and the implementation is in the adapter (so ast resolution is adapter specific) so I think it wouldn’t be too hard to implement the MySQL connection stuff if someone if feeling motivated.

Cheers, —Kip

josevalim:
@kipcole9 thanks. Btw, do you agree with the automatic inherits comment?

Btw, can you tell me for what do you use the alias_for? Is it to refer to the type from the parent table?

kipcole9:
José, yes, on reflection I agree with you.  It works well for my use case, but the doctrine of least surprises says that it shouldn’t be automatic.  However I do think its a useful thing to provision in some more specific way.  Not inheriting primary key definitions, indexes and triggers automatically by Postgres makes it easy to forget.  And its not simple to identify them from the parent and instantiate on the child.  Perhaps a macro to do that part if specifically invoked?

And yes, I use the alias_for to alias the tableoid which then provisions the name of the underlying table from which the row came.  Since a select() on a parent table can retrieve rows from one or more child tables, its valuable to know which actual table a row came from.  My original implementation was a string substitution on the derived SQL and that definitely didn’t feel right, hence the alias_for implementation which has other potential value anyway.

I have a third PR that I was holding back that will use the tableoid column (aliased as _type in this implementation) to populate the correct schema module dynamically.  I’ve done some performance testing on my implementation and it seems to hold up well:

## Ecto.Query.LoadTest
load data without dynamic schema lookup     1000000   34.41 µs/op
load data with dynamic schema lookup        1000000   34.46 µs/op

So even if the first PR doesn’t fit into Ecto comfortably, its possible the schema support and then the additional dynamic schema selection (which I haven’t submitted) may have some value.  Let me know if so and I’ll finish packaging it up.

Cheers, —Kip

josevalim:
Thanks @kipcole9. I will investigate and let you know. It is very likely you will be able to implement the missing features in a separate package. For example, I could imagine something that provides something like:

    inherit table(:foo), from: :bar

Or something similar that would the automatic reflection. I will post comments about the rest of the features today.

josevalim:
Sorry for the delay @kipcole9. As per my previous comments, I believe this would likely work better as a separate package. This way you can implement the index/pk definitions/etc inheritance mechanism and make it opt-in (or opt-out) as you prefer.

However I believe you wouldn't be able to implement the alias_for functionality in a separate package. It would need to be supported by Ecto. My concern though is that alias_for supports only a one-to-one mapping and I think we would benefit more if we had a generic mechanism that would allow, for example,  3 different columns with different types to become a single field in the schema. This is somewhat frequently asked although I am not sure how to implement it without coupling the schema to a particular repository implementation.

kipcole9:
José, no problem.  The reason I ‘invaded’ the schema mechanism is that an inherited table may have additional columns beyond those that are inherited.  So for now at least I’ll keep my own fork with the inheritance stuff.

Not sure I understand what you mean by "3 different columns with different types to become a single field in the schema” - but if its of value and I can do it then I’ll see what I can do.

josevalim:
> Not sure I understand what you mean by "3 different columns with different types to become a single field in the schema” - but if its of value and I can do it then I’ll see what I can do.

Sometimes, people want a field named "first_name" in their struct to map to a column named "firstName" in the database. Other times, they want the same column name but with a different type (your case). And some other times, they want 3 columns in their database to compose a single field in their schema.

Does it make sense?

PS: I will close this issue for now. I agree a fork is (unfortunately) the best way to go for now due to the schema limitations but I would advise you to move at least the migration stuff to a separate project out of the fork so it is less things for you to maintain inside the fork itself. :)

kipcole9:
OK, got it for at least the first case (column naming) and the second (my case).  I’ll need to do some work to enable updating of aliased columns in the first case though.  The third one is covered, at least in Postgres, by a composite type and which obviously works well in Ecto today - or am I still not getting it?

josevalim:
The third one could work as a composite type but sometimes composite types are not an option (existing database, using MySQL or SQLServer, etc).

PR #1341, opened 2016-03-28
Include the type that failed to cast in changeset errors

wojtekmach:
See: https://github.com/elixir-lang/ecto/issues/1295

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -370,7 +370,7 @@ defmodule Ecto.Changeset do
          {errors, valid?} = error_on_nil(kind, key, Map.get(changes, key, current), errors, valid?)
          {changes, errors, valid?}
        :invalid ->
-         {changes, [{key, {"is invalid", []}} | errors], false}
+         {changes, [{key, {"is invalid", [type: type]}} | errors], false}
      end}
   end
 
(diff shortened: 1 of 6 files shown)


josevalim:
Beautiful! Can you also do so for `cast_relation`? Please check if there are other places where we use "is invalid" as well.

wojtekmach:
Thanks for mentioning relations, I thought about it for a while. What would be the most useful 'type' value in these cases? Just 'type: :assoc' or something more elaborate like: 'type: {:assoc, %Ecto.Association.Has{...}}'?

AFAIR the only remaining place is 'put_relation' but I'll double check and update accordingly.

wojtekmach:
In a way, the only valid types for cast_relation and others are maps or lists of maps, so perhaps 'type: {:array, :map}'? I prefer something with assoc/embed though, idk :)

josevalim:
Yes, `{:array, :map}` sounds good as well for the cast case.

wojtekmach:
@ericmj or @michalmuskala do you have a preference about the best representation? Personally, I like `{:assoc, :one}` and `{:assoc, :many}` the most because it conveys the most information.

michalmuskala:
If those are to be shown to the end user (like an API consumer) I think it makes sense to make `has_one` a `:map` and `has_many` an `{:array, :map}` - that's the form the client needs to send data in.

wojtekmach:
> If those are to be shown to the end user (like an API consumer) I think it makes sense to make has_one a :map and has_many an {:array, :map} - that's the form the client needs to send data in.

Yeah, that makes sense too. My point of view was that these errors would go through `Changeset.traverse_errors` anyway, and so the developer would always have a chance to customize it (e.g. transform it to `{:array, :map}` if desired).

Anyway, I decided to move forward with `:map`, `{:array, :map}` and I just updated the PR. Let me know if anything is missing!

josevalim:

PR #1354, opened 2016-04-05
Make order_by work in has_many through association

YoussefKababe:
Fix the order_by issue in has_many through associations.


josevalim:
Thank you @YoussefKababe. More important than the fix would be a test case. Can you write a test case inside the integration_test directory that reproduces your issue using the existing schemas? Thank you!

YoussefKababe:
@josevalim I updated the tests and fixed what I did before!

josevalim:

YoussefKababe:
:smile: :+1: :heart:

PR #1364, opened 2016-04-11
Use simpler syntax to get post for delete example

radar:
Also includes a very small typo fix.


josevalim:

radar:
That was fast!! 🚄💨

PR #1365, opened 2016-04-11
Remove weird wording around final para of optimistic_lock docs

radar:
I think this may have been intended as "keep in mind", but it makes just as much sense without the "Finally, keep in mind" words too.


josevalim:

Issue #1366, opened 2016-04-12
BIT(1) type cannot load `<<1>>` as type :boolean


paulgithub:
### Precheck

When reading a boolean value from the database which is represented as a BIT(1) in the MySQL server (version 5.5.29) and using this value in model as a :boolean

There is a suggested fix for this on the mariaex github issue: https://github.com/xerions/mariaex/issues/107 that works.

### Environment

* Elixir version (elixir -v): 1.2.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): MySQL version 5.5.29
* Ecto version (mix deps): 1.1.4
* Database adapter and version (mix deps): mariaex version 0.7.3
* Operating system: OSX

### Current behavior

error / stacktrace

** (exit) an exception was raised:
** (ArgumentError) cannot load <<1>> as type :boolean
(ecto) lib/ecto/schema.ex:982: Ecto.Schema.load!/3
(ecto) lib/ecto/schema.ex:974: anonymous fn/3 in Ecto.Schema.do_load/4
(elixir) lib/enum.ex:1473: Enum."-reduce/3-lists^foldl/2-0-"/3
(ecto) lib/ecto/schema.ex:972: Ecto.Schema.do_load/4
(ecto) lib/ecto/schema.ex:957: Ecto.Schema.load/6
(ecto) lib/ecto/adapters/sql.ex:519: anonymous fn/3 in Ecto.Adapters.SQL.process_row/3
(elixir) lib/enum.ex:1151: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
(ecto) lib/ecto/adapters/sql.ex:513: Ecto.Adapters.SQL.process_row/3
(elixir) lib/enum.ex:1088: Enum."-map/2-lists^map/1-0-"/2
(ecto) lib/ecto/adapters/sql.ex:283: Ecto.Adapters.SQL.decode/2
(stdlib) timer.erl:166: :timer.tc/1
(ecto) lib/ecto/adapters/sql.ex:278: Ecto.Adapters.SQL.decode/3
(ecto) lib/ecto/adapters/sql.ex:225: Ecto.Adapters.SQL.query/5
(ecto) lib/ecto/adapters/sql.ex:183: Ecto.Adapters.SQL.query!/5
(ecto) lib/ecto/adapters/sql.ex:481: Ecto.Adapters.SQL.execute/6
[...]

PR #1369, opened 2016-04-13
BIT(1) type cannot load  as type :boolean #1366

paulgithub:
I have not been able to test this code it is to facilitate for the use of BIT(1) in MySQL. I was able to test the previous pull request on branch v1.1 and it works. I have sent this as it was requested by Jose in a previous attempt I made at a pull request was wrong.

I believe https://github.com/liveforeverx thought he had updated this to echo 2.0 but could not find the changes there. Hence this update.


josevalim:

PR #1383, opened 2016-04-20
Small fixes to CHANGELOG about constraints

nathanl:
- `exclusion_constraint` is actually named `exclude_constraint`. (Although I like `exclusion_constraint` better - it matches Postgres terminology - so maybe we should rename the function?)
- Check constraint examples makes more sense with message on the price field
- Give a feasible table name


josevalim:
> exclusion_constraint is actually named exclude_constraint.

Sounds great. Can you please send a PR for this change as well? Thank you!

josevalim:

PR #1384, opened 2016-04-20
Call them exclusion constraints, not exclude constraints

nathanl:
This matches the PostgreSQL docs

Per @josevalim's request in https://github.com/elixir-lang/ecto/pull/1383


nathanl:
Note - when I do `mix test.all` I get a couple of errors from `MIX_ENV=mysql mix test` in `integration_test/cases/joins.exs`, but I also get those errors on master.

josevalim:
Thank you! 

Others have also reported the MySQL issue but we cannot reproduce it. What is your MySQL version?

nathanl:
@josevalim I opened an issue with details - https://github.com/elixir-lang/ecto/issues/1387

PR #1395, opened 2016-04-24
Limit tagging of values

michalmuskala:
Given we have the mechanism of dupers, it should be the adapter that decides whether a value should be tagged or not.
Tagged values should never reach the adapter in `insert` or `update` unless the adapter tags them itself.

Right now I removed the tagging for `Ecto.UUID`, but we should probably also find a way to do something similar for `:binary`.


lib/ecto/uuid.ex
```diff
@@ -36,8 +36,8 @@ defmodule Ecto.UUID do
     catch
       :error -> :error
     else
-      binary ->
-        {:ok, %Ecto.Query.Tagged{type: :uuid, value: binary}}
+     binary ->
```
princemaple: :eyes:

michalmuskala:
@josevalim I managed to get rid of all tagging outside of adapters, and found a [compiler bug](https://github.com/elixir-lang/elixir/issues/4553) at the same time. 

Ready for review.

michalmuskala:
It's not yet ready after all. There are still issues with autogenerated values.

michalmuskala:
I updated with version that ensures autogenerated values are always dumped.
This creates some additional work when autogenerating `Ecto.DateTime` values since we wrap them in structs (in order for the autogeneration function to return full type), just to immediately unwrap them when dumping. However this issue should go away with Elixir 1.3 date types when, hopefully, we'll be operating everywhere on them.

On the other hand this makes it so that the only place an adapter has to deal with raw values are `dumpers` and `loaders` callbacks - we have only one place where values are actually converted between types. Besides that all values are in the shape requested by the adapter, which simplifies adapters significantly. I think this outweighs the downsides.

lib/ecto/adapters/postgres.ex
```diff
@@ -73,10 +73,9 @@ defmodule Ecto.Adapters.Postgres do
   @behaviour Ecto.Adapter.Structure
 
   ## Support arrays in place of IN
-  def dumpers({:embed, _} = type, _), do: [&Ecto.Adapters.SQL.dump_embed(type, &1)]
-  def dumpers(:binary_id, type), do: [type, Ecto.UUID]
-  def dumpers({:in, sub}, {:in, sub}), do: [{:array, sub}]
-  def dumpers(_, type), do: [type]
+  @doc false
+  def dumpers({:in, sub}, {:in, sub}), do: super({:array, sub}, {:array, sub})
```
josevalim: Why not return `{:array, sub}` directly?

lib/ecto/adapters/sql.ex
```diff
@@ -38,19 +38,15 @@ defmodule Ecto.Adapters.SQL do
       ## Types
 
       @doc false
-      def autogenerate(:id), do: nil
-      def autogenerate(:embed_id), do: Ecto.UUID.autogenerate()
-      def autogenerate(:binary_id), do: Ecto.UUID.autogenerate()
+      def autogenerate(:id),        do: nil
+      def autogenerate(:embed_id),  do: Ecto.UUID.generate()
```
josevalim: This is reverting the work I did in unifying those APIs for 2.0 :(

lib/ecto/adapters/sql.ex
```diff
@@ -38,19 +38,15 @@ defmodule Ecto.Adapters.SQL do
...
-      def loaders({:embed, _} = type, _), do: [&Ecto.Adapters.SQL.load_embed(type, &1)]
-      def loaders(:binary_id, type), do: [Ecto.UUID, type]
-      def loaders(_, type), do: [type]
+      defdelegate loaders(primitive, type), to: Ecto.Adapters.SQL
 
       @doc false
-      def dumpers({:embed, _} = type, _), do: [&Ecto.Adapters.SQL.dump_embed(type, &1)]
-      def dumpers(:binary_id, type), do: [type, Ecto.UUID]
-      def dumpers(_, type), do: [type]
+      defdelegate dumpers(primitive, type), to: Ecto.Adapters.SQL
```
josevalim: Let's not delegate and rather inline as it was before. These functions are called *a lot*.

lib/ecto/uuid.ex
```diff
@@ -100,9 +100,7 @@ defmodule Ecto.UUID do
 
   # Callback invoked by autogenerate fields.
   @doc false
-  def autogenerate do
-    %Ecto.Query.Tagged{type: :uuid, value: bingenerate()}
-  end
+  def autogenerate, do: generate()
```
josevalim: Unfortunately this is expensive because we generate the binary to become a UUID and then to become a binary again.

lib/ecto/adapters/sql.ex
michalmuskala: But it is wrong in many places.
For SQL databases we want values in embeds to be application-level types, while in regular models to be database level types.
With Mongo, I want the value to always be database level, because there's no JSON encoding in the middle.

michalmuskala: With mongo I do generate the same value in both places.

josevalim: Right, but you can call those whatever you want in Mongo. :)

michalmuskala: The problem is that in one place we later loaded the autogenerated value (embeds) and in the other we didn't.
In mongo I need raw value in both places. So we either need to dump in both places (but then we need to have different type for embed_id and regular one for SQL databases), or let the adapter control it.

josevalim: I see. The autogenerate API is still a bit everywhere unfortunately. autogenerate for embeds generate an Ecto value, for binary ids the DB value and the autogenerate API Ecto values.

I don't mind much about the autogenerate in adapters bit but the one in custom types is bothering me a lot because of the amount of unecessary paths. Here is what is happening for UUID:

  1. autogenerate calls "bingenerate" which calls "generate"
  2. ecto then dumps the type converting "generate" back into "bingenerate"
  3. the type is loaded after the record is modified, going from "bingenerate" to "generate"

So we have:

    bingenerate -> generate -> bingenerate -> generate

Previously we had:

  1. autogenerate calls "bingenerate"
  3. the type is loaded after the record is modified, going from "bingenerate" to "generate"

I think we can at least skip step 3 though by storing those values. I will give it a try. It is still one extra pass compared to the ideal case but I don't think we can avoid it if we want to keep the adapters responsible for what they are doing.

josevalim: @michalmuskala is it safe to say that you never see the `{{..., ..., ...}, {..., ..., ...}}` for datetimes in the Mongo adapter?

michalmuskala: Yes, I never see them. At least I never want to see them ;)

michalmuskala: After the changes I made the autogenerate in custom types returns the custom value, and not the database representation, so I'd say it makes more sense.

The downside, as you said is with things that we can generate directly in the database version. Maybe we should have something to short-circuit the first step in database dumping of those types, so we only keep the adapter-specific logic?

PR #1403, opened 2016-04-27
Support required has_many in traverse_errors

moklett:
A required has_many association already reports a Changeset error when the value is empty:

    iex> changeset = %Author{}
    iex>   |> Changeset.cast(%{posts: []}, ~w())
    iex>   |> Changeset.cast_assoc(:posts, required: true)
    iex> changeset.errors
    [posts: {"can't be blank", []}]

However, calling `Changeset.traverse_errors` on this changeset currently results in a blank set of errors for `posts`:

    iex> Ecto.Changeset.traverse_errors(changeset, &(&1))
    %{posts: []}

This happened because the errors on the nested association -- of which there were none -- overwrote the error on the parent for the association name, i.e. it was the equivalent of the following:

    iex> Map.put(%{posts: [{"can't be blank", []}]}, :posts, [])
    %{posts: []}

The fix has been implemented by merging the nested errors only if there is no error on the parent for the association.  I believe this is safe because if there is an error on the association itself, we should not care about errors on the children.

### DIFF
--- a/test/ecto/changeset/has_assoc_test.exs
+++ b/test/ecto/changeset/has_assoc_test.exs
@@ -402,6 +402,11 @@ defmodule Ecto.Changeset.HasAssocTest do
     assert changeset.required == [:posts]
     assert changeset.errors == []
 
+    changeset = cast(%Author{}, %{posts: []}, :posts, required: true)
+    assert changeset.required == [:posts]
+    assert changeset.changes == %{posts: []}
+    assert changeset.errors == [posts: {"can't be blank", []}]
+
     changeset = cast(%Author{posts: []}, %{}, :posts, required: true)
     assert changeset.required == [:posts]
     assert changeset.changes == %{}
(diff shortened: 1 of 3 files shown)


test/ecto/changeset/has_assoc_test.exs
```diff
@@ -790,4 +795,46 @@ defmodule Ecto.Changeset.HasAssocTest do
...
+    assert changeset.errors == [posts: {"can't be blank", []}]
+    assert Changeset.traverse_errors(changeset, &(&1)) == %{posts: [{"can't be blank", []}]}
+
+    changeset = cast(%Author{posts: []}, %{"posts" => nil}, :posts, required: true)
+    assert changeset.errors == [posts: {"is invalid", [type: {:array, :map}]}]
+    assert Changeset.traverse_errors(changeset, &(&1)) == %{posts: [{"is invalid", [type: {:array, :map}]}]}
+
+    changeset = cast(%Author{}, %{posts: []}, :posts, required: true)
+    assert changeset.errors == [posts: {"can't be blank", []}]
+    assert Changeset.traverse_errors(changeset, &(&1)) == %{posts: [{"can't be blank", []}]}
```
moklett: This `assert` (line 838) is the ONLY ONE that will fail without the code change from this PR.  The other assertions I have added are to establish a baseline for `traverse_errors`, since it didn't seem well-tested for the `has_many` association cases.

josevalim:
Beautiful report, fix and tests. Thank you! 

test/ecto/changeset/has_assoc_test.exs
moklett: And, for the record, this would be the failure:

```ex
  1) test traverses changeset errors with has_many when required (Ecto.Changeset.HasAssocTest)
     test/ecto/changeset/has_assoc_test.exs:819
     Assertion with == failed
     code: Changeset.traverse_errors(changeset, &(&1)) == %{posts: [{"can't be blank", []}]}
     lhs:  %{posts: []}
     rhs:  %{posts: [{"can't be blank", []}]}
     stacktrace:
       test/ecto/changeset/has_assoc_test.exs:838
```

PR #1409, opened 2016-05-02
Better errors on field that is both required and invalid

moklett:
Using `cast/3` and `validate_required/2` together on a field that is of the wrong type will produce 2 errors that seem conflicting.

For example, given the following schema:

```ex
  defmodule Post do
    use Ecto.Schema

    schema "posts" do
      field :body
    end
  end
```

Casting the wrong type for `body` while also marking it as required produces 2 errors on the `body` field:

```ex
iex> changeset = cast(%Post{}, %{body: :invalid}, [:body]) |> validate_required([:body])
iex> changeset.changes
%{}
iex> changeset.errors
[body: {"can't be blank", []}, body: {"is invalid", [type: :string]}]
```

This change prevents the "can't be blank" error when the field as already been deemed invalid, since the field is not actually blank.

---

The way I have fixed this, ANY error that is present on a field when `validate_required` is called will prevent the "can't be blank" error from being added.  This _seems_ like it would be expected in most cases, but will it be potentially surprising to the user?  If we do go this way, perhaps a note should be added to the `validate_required/2` docs indicating that any fields with a pre-existing error will be skipped?

The other way I thought of fixing this was to change the type check such that the field is NOT removed from the `changes` map, but this seems to change even more pre-existing behavior (and breaks more tests).

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -46,6 +46,28 @@ defmodule Ecto.ChangesetTest do
     assert changeset.valid?
   end
 
+  test "cast/3: optional field of wrong type is marked as invalid" do
+    params = %{"body" => :world}
+    struct = %Post{}
+
+    changeset = cast(struct, params, ~w(body))
+    assert changeset.changes == %{}
+    assert changeset.errors == [body: {"is invalid", [type: :string]}]
+    refute changeset.valid?
+  end
+
+  test "cast/3: required field (via validate_required/2) of wrong type is marked as invalid" do
+    params = %{"body" => :world}
+    struct = %Post{}
+
+    changeset = cast(struct, params, [:body])
+                |> validate_required([:body])
+
+    assert changeset.changes == %{}
+    assert changeset.errors == [body: {"is invalid", [type: :string]}]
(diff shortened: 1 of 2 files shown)


test/ecto/changeset_test.exs
```diff
@@ -46,6 +46,28 @@ defmodule Ecto.ChangesetTest do
...
+    params = %{"body" => :world}
+    struct = %Post{}
+
+    changeset = cast(struct, params, ~w(body))
+    assert changeset.changes == %{}
+    assert changeset.errors == [body: {"is invalid", [type: :string]}]
+    refute changeset.valid?
+  end
+
+  test "cast/3: required field (via validate_required/2) of wrong type is marked as invalid" do
```
moklett: I'm curious if others think this is the right way to test this.  The setup uses both `cast` and `validate_required`, since that is the use case that spawned this change.  However, we can also test it in `validate_required` just by showing that pre-existing errors on a field prevent further errors.

Note that both tests I added are patterned after similar tests in `cast/4` that check for required vs. optional fields that have an invalid type.

josevalim:

PR #1417, opened 2016-05-05
Add join example with condition

VladShcherbin:


josevalim:

PR #1421, opened 2016-05-06
Fix `from/2` example in documentation

parkerl:
Currently this example throws
```
 ** (Ecto.Query.CompileError) unbound variable `c` in query
    (ecto) expanding macro: Ecto.Query.select/3
           iex:17: (file)
    (ecto) expanding macro: Ecto.Query.from/2
           iex:17: (file)
```


parkerl:
Opening this as a PR separate from #1420 because it seems like it might require less discussion.

PR #1422, opened 2016-05-07
Remove default length specifier from PostgreSQL string type

johngrimes:
The absence of a length specifier is a better default - the 255 character limit
is arbitrary and is actually inferior from a performance perspective within
PostgreSQL.

See http://www.postgresql.org/docs/current/static/datatype-character.html.


josevalim:
I am afraid that not having a limit can leave applications exposed to attacks that continuously send large data to string fields. If you don't want a limit, then maybe you should be using text or another type?

johngrimes:
Yes, I think you're right. I hadn't thought about the security aspect.

I think that probably the best way for me to do it right now is define string fields as text and define a max length validation on the model - which I would have to do anyway, even if it was a varchar(255).

josevalim:
Yes! YOu can also use :varchar as a type in your migrations if you want. :)

PR #1429, opened 2016-05-14
Docs: Replaced the ~w() list usage in cast() and with atom lists

carusso:
Hi, first pull request ever. :)

I noticed in the Ecto 2.0 docs that the Changeset cast examples still used ~w(title) for the required and optional fields.  In my own code, I received an error from using string lists in that parameter.

I replaced all the examples in the Changeset docs with [:title] atom lists.

Cheers,
Chris Russo

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -39,7 +39,7 @@ defmodule Ecto.Changeset do
 
         def changeset(user, params \\ %{}) do
           user
-          |> cast(params, ~w(name email age))
+          |> cast(params, [:name, :email, :age])
           |> validate_required([:name, :email])
           |> validate_format(:email, ~r/@/)
           |> validate_inclusion(:age, 18..100)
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you and congratulations! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

paulcsmith:
I'm curious why atoms are now required? I thought it was more convenient to use `~w()`, but maybe there's some other internal issues that string keys caused?

wojtekmach:
@paulcsmith I don't think atoms are required, we've recently upgraded hex_web to ecto-2.0.0-rc.5 and we're still using strings for the field lists. Looking at the typespecs it does accept both: https://github.com/elixir-lang/ecto/blob/v2.0.0-rc.5/lib/ecto/changeset.ex#L303:L305. I'm also curious about rationale behind this change although it's not a big deal (I personally prefer  `~w()`/`~w()a` though :))

paulcsmith:
@wojtekmach I believe that `cast` still accepts string fields, but `validate_required` does not, at least on rc4. I would expect it to be the same for `cast` and `validate_required` but it doesn't seem to be

wojtekmach:
@paulcsmith good point about `validate_required`

PR #1434, opened 2016-05-17
Deprecate Ecto.Changeset.cast/4

wojtekmach:


test/ecto/changeset/belongs_to_test.exs
```diff
@@ -61,7 +62,7 @@ defmodule Ecto.Changeset.BelongsToTest do
 
   test "cast belongs_to with invalid params" do
     changeset = cast(%Author{}, %{"profile" => %{name: nil}}, :profile)
-    assert changeset.changes.profile.changes == %{}
+    assert changeset.changes.profile.changes == %{name: nil}
```
wojtekmach: this is seems to be the only non-obvious change in tests (apart from silencing warnings later in the PR)

lib/ecto/changeset.ex
```diff
@@ -313,38 +313,43 @@ defmodule Ecto.Changeset do
   Converts the given `params` into a changeset for `data`
   keeping only the set of `required` and `optional` keys.
   """
-  # TODO: Deprecate this function so it finally emit warnings.
   def cast(data, params, required, optional)
 
-  def cast(_data, %{__struct__: _} = params, _required, _optional) do
+  def cast(data, params, required, optional) do
+    IO.puts :stderr, "warning: `Ecto.Changeset.cast/4` is deprecated, " <>
```
josevalim: We are going to do this only on 2.1 (sorry, I should have added this information to the TODO). So what if we bring the TODO back and comment the warning for now? Everything else is great to go though :)

josevalim:
I have added one comment and we are good to go. Basically, let's not emit the warning and bring the TODO back, but the remaining changes are good. :)

lib/ecto/changeset.ex
wojtekmach: sounds good!

wojtekmach:
@josevalim Done, I removed the deprecation in af23442. Build failure seems unrelated.

josevalim:

PR #1439, opened 2016-05-18
Include ecto_repos setup instructions in the readme

joshprice:
Without this you get the following error: 

```
✪ mix ecto.create
warning: could not find repositories for application :my_app.

You can avoid this warning by passing the -r flag or by setting the
repositories managed by this application in your config/config.exs:

    config :my_app, ecto_repos: [...]

The configuration may be an empty list if it does not define any repo.
```


josevalim:

PR #1444, opened 2016-05-19
Fix string casting in MySQL adapter

tpunt:
When using to Ecto's `type` function to cast a value to type `:string` with the MySQL adapter, a SQL syntax error is produced. This is because casting to `varchar` is invalid in MySQL - we must instead cast to type `char` ([see here](http://stackoverflow.com/a/15368852/4530326)).

This works fine when comparing values that have been casted to `char` to columns of either type `char` or `varchar`.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -420,7 +420,7 @@ if Code.ensure_loaded?(Mariaex) do
     end
 
     defp expr(%Ecto.Query.Tagged{value: other, type: type}, sources, query) do
-      "CAST(#{expr(other, sources, query)} AS " <> ecto_to_db(type, query) <> ")"
+      "CAST(#{expr(other, sources, query)} AS " <> ecto_cast_to_db(type, query) <> ")"
     end
 
     defp expr(nil, _sources, _query),   do: "NULL"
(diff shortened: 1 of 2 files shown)


josevalim:
This function is also used by migrations and this change unfortunately
means that string columns will be marked as char. We need to handle it
somewhat different. Can you also please include a test?

On Thursday, May 19, 2016, Thomas Punt <notifications@github.com> wrote:

> When using to Ecto's type function to cast a value to type :string with
> the MySQL adapter, a SQL syntax error is produced. This is because casting
> to varchar is invalid in MySQL - we must instead cast to type char (see
> here <http://stackoverflow.com/a/15368852/4530326>).
>
> This works fine when comparing values that have been casted to char to
> columns of either type char or varchar.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/elixir-lang/ecto/pull/1444
> Commit Summary
>
>    - Fix string casting in MySQL adapter
>
> File Changes
>
>    - *M* lib/ecto/adapters/mysql/connection.ex
>    <https://github.com/elixir-lang/ecto/pull/1444/files#diff-0> (2)
>
> Patch Links:
>
>    - https://github.com/elixir-lang/ecto/pull/1444.patch
>    - https://github.com/elixir-lang/ecto/pull/1444.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/elixir-lang/ecto/pull/1444>
>


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

tpunt:
Hmm... I see what you mean. I accidentally ran the test suite from the wrong branch, so didn't catch this. I'll see if I can find a different way to patch this, as well as add a test :)

josevalim:

PR #1454, opened 2016-05-26
Adding opts to the Repo.start_link behaviour

jeregrine:
Currently the [Repo.start_link implementation](https://github.com/jeregrine/ecto/blob/a4ee7e44a34309f075bdec8ef014578a3e66efe9/lib/ecto/repo.ex#L113) accepts an options keyword list but the behaviour does not reflect this.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -233,12 +233,16 @@ defmodule Ecto.Repo do
 
   @doc """
   Starts any connection pooling or supervision and return `{:ok, pid}`
-  or just `:ok` if nothing needs to be done.
+  or just `:ok` if nothing needs to be done. 
 
   Returns `{:error, {:already_started, pid}}` if the repo is already
   started or `{:error, term}` in case anything else goes wrong.
+
+  ## Options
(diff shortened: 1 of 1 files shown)


lib/ecto/repo.ex
```diff
@@ -233,12 +233,16 @@ defmodule Ecto.Repo do
...
   Starts any connection pooling or supervision and return `{:ok, pid}`
-  or just `:ok` if nothing needs to be done.
+  or just `:ok` if nothing needs to be done. 
 
   Returns `{:error, {:already_started, pid}}` if the repo is already
   started or `{:error, term}` in case anything else goes wrong.
+
+  ## Options
+
+  See the configuration for the module documentation.
```
michalmuskala: I think we can say:
```
See the configuration in the moduledoc for options shared between adapters,
for adapter-specific configuration see the adapter's documentation.
```

Or something similar. This module mentions just a few options since most of them are directly dependent on the adapter you use.

jeregrine: Good call

Issue #1450, opened 2016-05-23
Maybe `validate_confirmation/3` should require the confirmation field


asonge:
The docs say that

> Note that this does not add a validation error if the confirmation field is nil. Note “email_confirmation” does not need to be added as a virtual field in your schema.

The problem is that if I do not add it as a virtual field in my schema, I cannot use `validate_required/3` to make sure the changeset contains the parameter. Virtual fields seem very model-like compared to the schema/representation model we are after in Ecto 2.

It seems that we should either provide a way to validate the existence of a parameter *or* `validate_confirmation/3` should have an option to require the confirmation to exist. This seems particularly important as these confirmations are important for security. I can imagine someone using a confirmation to verify the name of a resource that the user has to type in to delete. It might be easy to accidentally forget to include this parameter and enable delete without confirmation.

josevalim:
A `required: true` option is fine then. The default should be false.

michalmuskala:
I think it should be the default to check the confirmation field even exists on the schema.

A possible way to not enforce a virtual field on a schema is to have a `with` option:
```elixir
|> validate_confirmation(:password, with: &get_field(&1, :password_confirmation))
|> validate_confirmation(:foo, with: &(&1.params["foo_confirmation"]))
```

josevalim:
@michalmuskala then it is better to just write your own confirmation code. :P

michalmuskala:
That's probably good point.

But I still think we should error if the confirmation field does not exist. This can lead to subtle bugs.

asonge:
Just to make a short case for making it true by default:

- For those making normal UI apps, it might be a surprise that they have to pass an extra option to enable requiring it to exist
- For those making API's without UI, they won't include the confirmation validator at all
- For those making normal UI apps that also have API functionality, and they turn to the manual when the API request fails, they know that if they turn off the confirmation requirement, and that they need to add extra code somewhere UI-specific to enforce that the parameter exists.

josevalim:
> For those making normal UI apps, it might be a surprise that they have to pass an extra option to enable requiring it to exist

They mostly don't care because the UI will be passing the confirmation field anyway. If the UI doesn't, that's where the bug is.

> For those making normal UI apps that also have API functionality, and they turn to the manual when the API request fails, they know that if they turn off the confirmation requirement, and that they need to add extra code somewhere UI-specific to enforce that the parameter exists.

Opposite to a default that works for both today? :)

[...]

PR #1467, opened 2016-06-03
Add required option to validate_confirmation/3

alexauritt:
Currently `validate_confirmation/3` does not add an error message if the specified field's expected corresponding confirmation field is nil. This PR adds an optional `required` flag, which defaults to false.

Re: #1450

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1348,13 +1348,16 @@ defmodule Ecto.Changeset do
   validation will check if both "email" and "email_confirmation"
   in the parameter map matches.
 
-  Note that this does not add a validation error if the confirmation
-  field is nil. Note "email_confirmation" does not need to be added
-  as a virtual field in your schema.
+  Note that if the confirmation field is nil or missing, by default this does
+  not add a validation error. You can specify that the confirmation field is
+  required in the options (see below). Note "email_confirmation" does not need
+  to be added as a virtual field in your schema.
 
   ## Options
 
     * `:message` - the message on failure, defaults to "does not match"
+    * `:required` - boolean, sets whether existence of confirmation parameter
+      is required for addition of error. Defaults to false
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1372,15 +1375,20 @@ defmodule Ecto.Changeset do
...
       case Map.fetch(changeset.params, error_param) do
         {:ok, ^value} -> []
         {:ok, _}      -> [{error_field, {message(opts, "does not match confirmation"), []}}]
-        :error        -> []
+        :error        -> confirmation_missing(opts, error_field)
       end
     end
   end
 
+  defp confirmation_missing(opts = [required: true], error_field) do
```
michalmuskala: You can't pattern match on keyword like that. What if I pass `[message: "foo", required: true]` as options?

lib/ecto/changeset.ex
```diff
@@ -1372,15 +1375,20 @@ defmodule Ecto.Changeset do
...
+    [{error_field, {message(opts, "must be included in this changeset"), []}}]
```
michalmuskala: I think we can use the `can't be blank` message we use for normal required field validation.

lib/ecto/changeset.ex
```diff
@@ -1348,13 +1348,16 @@ defmodule Ecto.Changeset do
   validation will check if both "email" and "email_confirmation"
   in the parameter map matches.
 
-  Note that this does not add a validation error if the confirmation
-  field is nil. Note "email_confirmation" does not need to be added
-  as a virtual field in your schema.
+  Note that if the confirmation field is nil, by default this does not add
```
michalmuskala: Maybe we should say "nil or missing"?

michalmuskala:
@alexauritt this looks good :+1: I added some comments.

alexauritt:
@michalmuskala thanks so much for the thoughtful comments. Let me know if you spot anything else. :smile:

lib/ecto/changeset.ex
alexauritt: good call. updated.

lib/ecto/changeset.ex
alexauritt: Updated

lib/ecto/changeset.ex
alexauritt: Good catch

michalmuskala:
Thank you! :heart:

Issue #1468, opened 2016-06-04
Improve error when trying to name an association the same as its foreign_key


narrowtux:
When trying to do `belongs_to` with a custom `:foreign_key` that is the same as the name of the association, ecto crashes with 

```
** (exit) an exception was raised:
    ** (Ecto.QueryError) /path/to/project/deps/ecto/lib/ecto/association.ex:392: field `AppsrvWeb.Organization.parent` in `where` does not exist in the schema in query:
from o in AppsrvWeb.Organization,
  where: o.parent in ^[2],
  select: {o.parent, o}
        (elixir) lib/enum.ex:1473: Enum."-reduce/3-lists^foldl/2-0-"/3
        (elixir) lib/enum.ex:1151: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/enum.ex:1473: Enum."-reduce/3-lists^foldl/2-0-"/3
        (ecto) lib/ecto/repo/queryable.ex:119: Ecto.Repo.Queryable.execute/5
        (ecto) lib/ecto/repo/queryable.ex:40: Ecto.Repo.Queryable.all/4
        (ecto) lib/ecto/repo/preloader.ex:171: Ecto.Repo.Preloader.fetch_query/8
        (ecto) lib/ecto/repo/preloader.ex:111: Ecto.Repo.Preloader.preload_assoc/10
        (elixir) lib/task/supervised.ex:89: Task.Supervised.do_apply/2
        (elixir) lib/task/supervised.ex:40: Task.Supervised.reply/5
```

This error message should be fixed to tell that the association name has to be different from the `foreign_key`. The documentation should also state that

alexauritt:
@michalmuskala -- possibly above my pay grade, but happy to take a look

michalmuskala:
Sure, go ahead.

alexauritt:
@narrowtux -- to clarify, you're getting that error when you execute a query on the child table (while preloading the parent), and you're saying we should have a clearer error when we compile the association/fk in the child schema, correct?

Could you post some example schemas and a query? I'm not able to reproduce this as I am in fact getting a compilation error when I use the same name for the association and fk.

e.g. if I define:

```
defmodule Demo.Author do
  use Ecto.Schema

  schema "authors" do
    field :name, :string
    field :email, :string
  end
end

defmodule Demo.Post do
  use Ecto.Schema

  schema "posts" do
    field :title, :string
[...]

PR #1471, opened 2016-06-06
Improve error message when fk matches association

alexauritt:
Improve error message is foreign_key is identical to corresponding association name. Re: #1468

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1178,6 +1178,10 @@ defmodule Ecto.Schema do
     opts = Keyword.put_new(opts, :foreign_key, :"#{name}_id")
     foreign_key_type = opts[:type] || Module.get_attribute(mod, :foreign_key_type)
 
+    if name == Keyword.get(opts, :foreign_key) do
+      raise ArgumentError, "foreign_key #{inspect name} must be distinct from corresponding association name"
+    end
+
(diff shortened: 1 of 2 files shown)


lib/ecto/schema.ex
```diff
@@ -1174,17 +1174,25 @@ defmodule Ecto.Schema do
...
   def __belongs_to__(mod, name, queryable, opts) do
     check_options!(opts, @valid_belongs_to_options, "belongs_to/3")
+    opts = set_foreign_key!(mod, name, opts)
 
+    struct =
+      association(mod, :one, name, Ecto.Association.BelongsTo, [queryable: queryable] ++ opts)
+    Module.put_attribute(mod, :changeset_fields, {name, {:assoc, struct}})
+  end
+
+  defp set_foreign_key!(mod, name, opts) do
```
josevalim: Why are you breaking it into a separate function? :)

alexauritt: @josevalim -- no need to break it up. Thought it might be bit more readable. You prefer as it was? :)

alexauritt: You're right. Reads better w/o the separate function. :smile:

josevalim:

narrowtux:
Woah very fast. Nice work :)

PR #1473, opened 2016-06-08
Ecto.Changeset.validate_change/3 should keep errors consistent.

linjunpop:
I am working with `traverse_errors/2`, which's doc says `The error message function receives an error tuple {msg, opts}`

I follow this doc and set the error message to `[title: "is foo"]`, then when I traverse errors, the message is not formatted as `{msg, opts}`.

I'm not sure we should change to doc or change to code:

```
    case new do
      []    -> changeset
      [_|_] -> %{changeset | errors: new ++ errors, valid?: false}
    end
```

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1014,7 +1014,7 @@ defmodule Ecto.Changeset do
       iex> changeset = change(%Post{}, %{title: "foo"})
       iex> changeset = validate_change changeset, :title, fn
       ...>   # Value must not be "foo"!
-      ...>   :title, "foo" -> [title: "is foo"]
+      ...>   :title, "foo" -> [{:title, {"is foo", []}]
       ...>   :title, _     -> []
       ...> end
       iex> changeset.errors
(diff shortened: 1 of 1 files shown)


linjunpop:
I've saw tests on `validate_change` does `assert changeset.errors == [title: "oops"]`,
but tests with `validate_format`(https://github.com/elixir-lang/ecto/blob/56a5869f2c6e40d8af92bfc7e4e4bb4c8391e487/test/ecto/changeset_test.exs#L654) does `assert changeset.errors == [title: {"has invalid format", []}]`

these are two different data structures, is this expected?

josevalim:
You should change your code. Use `add_error` to add errors messages. It is expected for all error messages to be tuples.

linjunpop:
But if

>  It is expected for all error messages to be tuples.

Shouldn't the doc and tests to be written as tuples?

josevalim:
@linjunpop yes, definitely. if the tests are out of date, please send a PR. :)

linjunpop:
@josevalim OK, I will prepare another PR. :)

josevalim:
@linjunpop No worries, you are right, I have just fixed this in master to support both formats. :)

linjunpop:
Cool

PR #1482, opened 2016-06-12
Add Ecto.Multi.merge/2

### LINKED ISSUE #1464: Support prepend and append with functions in Ecto.Multi

michalmuskala:
The function passed to merge accepts an initial multi - that allows us to easily guard against duplicated operations, that would cause issues later on when merging results.

This still needs docs.

Closes #1464

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -24,12 +24,6 @@ defmodule Ecto.Multi do
   If any changeset has errors, the transaction won't even be started and the error
   will be immediately returned.
 
-  `insert/4`, `update/4` and `delete/4` also accept a function which receives
-  the changes so far and returns a changeset. This allows introspection and use
-  of database generated values, such as ids, from previous steps. However when
-  using changesets wrapped in functions the changeset can not be validated prior
-  to starting the transaction.
-
   ## Run
 
   Multi allows you to run arbitrary functions as part of your transaction via
@@ -108,7 +102,6 @@ defmodule Ecto.Multi do
   @type merge :: (map -> t) | {module, atom, [any]}
   @typep schema_or_source :: binary | {binary | nil, binary} | Ecto.Schema.t
   @typep operation :: {:changeset, Changeset.t, Keyword.t} |
-                      {:changeset_fun, atom, changeset_fun, Keyword.t} |
                       {:run, run} |
                       {:merge, merge} |
                       {:update_all, Ecto.Query.t, Keyword.t} |
@@ -211,7 +204,8 @@ defmodule Ecto.Multi do
   """
   @spec merge(t, module, function, args) :: t
     when function: atom, args: [any]
-  def merge(multi, mod, fun, args) do
+  def merge(%Multi{} = multi, mod, fun, args)
+      when is_atom(mod) and is_atom(fun) and is_list(args) do
     Map.update!(multi, :operations, &[{:merge, {:merge, {mod, fun, args}}} | &1])
   end
 
@@ -220,15 +214,11 @@ defmodule Ecto.Multi do
 
   Accepts the same arguments and options as `Ecto.Repo.insert/3` does.
   """
-  @spec insert(t, name, Changeset.t | Ecto.Schema.t | changeset_fun, Keyword.t) :: t
+  @spec insert(t, name, Changeset.t | Ecto.Schema.t, Keyword.t) :: t
   def insert(multi, name, changeset_or_struct, opts \\ [])
 
   def insert(multi, name, %Changeset{} = changeset, opts) do
     add_changeset(multi, :insert, name, changeset, opts)
-  end
-
-  def insert(multi, name, fun, opts) when is_function(fun, 1) do
-    add_operation(multi, name, {:changeset_fun, :insert, fun, opts})
   end
 
   def insert(multi, name, struct, opts) do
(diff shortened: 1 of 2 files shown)


michalmuskala:
@ericmj @fishcakez - you were asking for this feature, what do you think about this?

lib/ecto/multi.ex
```diff
@@ -234,8 +243,8 @@ defmodule Ecto.Multi do
     add_changeset(multi, :delete, name, changeset, opts)
   end
 
-  def delete(multi, name, run, opts) when is_function(run, 1) do
-    add_operation(multi, name, {:changeset_fun, :delete, run, opts})
+  def delete(multi, name, fun, opts) when is_function(fun, 1) do
```
josevalim: We are no longer supporting those. :)

test/ecto/multi_test.exs
```diff
@@ -330,4 +330,57 @@ defmodule Ecto.MultiTest do
...
+    |> Multi.merge(fn _, multi ->
+      assert_raise RuntimeError, ~r":run is already a member", fn ->
+        Multi.run(multi, :run, fun)
+      end
+    end)
+    |> Multi.run(:run, fun)
+
+
+    Multi.new
+    |> Multi.merge(fn _, multi -> Multi.run(multi, :run, fun) end)
```
josevalim: It seems we are passing a multi only to avoid name conflicts, right? I would rather not pass a multi at all and, when merging, figure out if there is a name error. The API will be simpler and less confusing. WDYT?

michalmuskala: I was wondering about that as well, what would be a better approach. I'm open to changing it.

josevalim: I think the approach where we don't pass a multi would be less confusing
and easier to compose (you can use any function that returns a multi
without having to pass yours).

On Sunday, June 12, 2016, Michał Muskała <notifications@github.com> wrote:

> In test/ecto/multi_test.exs
> <https://github.com/elixir-lang/ecto/pull/1482#discussion_r66727275>:
>
> > +  end
> > +
> > +  test "merge does not allow repeated operations" do
> > +    fun = fn _ -> {:ok, :ok} end
> > +
> > +    Multi.new
> > +    |> Multi.merge(fn _, multi ->
> > +      assert_raise RuntimeError, ~r":run is already a member", fn ->
> > +        Multi.run(multi, :run, fun)
> > +      end
> > +    end)
> > +    |> Multi.run(:run, fun)
> > +
> > +
> > +    Multi.new
> > +    |> Multi.merge(fn _, multi -> Multi.run(multi, :run, fun) end)
>
> I was wondering about that as well, what would be a better approach. I'm
> open to changing it.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/elixir-lang/ecto/pull/1482/files/ace09344c55e5c4227c379f4fe5140544a43a6bb#r66727275>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbtiTFzAUn_y7WzzBNb8ZyFm7WwHFks5qLHOmgaJpZM4Iz4Hm>
> .
>


-- 


michalmuskala:
@josevalim updated

lib/ecto/multi.ex
```diff
@@ -364,37 +368,56 @@ defmodule Ecto.Multi do
...
+  defp apply_fun(fun, acc),              do: apply(fun, [acc])
+
+  defp merge_results(changes, new_changes, names) do
+    new_names = new_changes |> Map.keys |> MapSet.new
+    case MapSet.intersection(names, new_names) |> MapSet.to_list do
+      [] ->
+        {Map.merge(changes, new_changes), MapSet.union(names, new_names)}
+      common ->
+        raise "merging results of multi, found operations: #{inspect common} " <>
+          "already present in the enclosing multi"
```
michalmuskala: I'm not sure about the error message here

josevalim: "cannot merge multi, the following operations were found in both Ecto.Multi: #{inspect common}"

?

test/ecto/multi_test.exs
```diff
@@ -330,4 +271,72 @@ defmodule Ecto.MultiTest do
...
+  end
+
+  test "merge rollbacks on errors" do
+    error = fn _ -> {:error, :error} end
+    ok    = fn _ -> {:ok, :ok} end
+
+    multi =
+      Multi.new
+      |> Multi.run(:outside_ok, ok)
+      |> Multi.merge(fn _ ->
```
josevalim: We should also test that the merged operations run before any upcoming operation in the parent changeset. For example, we can add `outside_error` at the end and ensure it won't be triggered.

michalmuskala: done

michalmuskala:
@josevalim Ready for review

josevalim:
:shipit:

fishcakez:
Thank you @michalmuskala :heart:, this is very nice :D

wojtekmach:
@michalmuskala `Multi.merge` is a really nice new abstraction, thanks!

I recently ported `Multi.insert + fn` to `Multi.merge` and it was very straightforward:
- [before](https://github.com/hexpm/hex_web/blob/master/web/controllers/api/release_controller.ex#L135:L139)
- [after](https://github.com/wojtekmach/hex_web/blob/878a66c6298507cf1f9f396b8ec805b4b3a0d9d8/web/controllers/api/release_controller.ex#L141:L145)

The former has a bit less code and to me is more readable (it's not a big difference though) and now it should be trivial to re-implement in terms of `.merge` - if that would seem like a good idea and/or you get feature requests - feel free to let me know ;-)

michalmuskala:
I think implementing `insert`/`update`/`delete` in terms of merge is a good idea, I'm not sure this should be part of ecto itself, given it's basically a one-liner:

```elixir
def insert(multi, name, fun) do
  Multi.merge(multi, fn data -> Multi.new |> Multi.insert(name, fun.(data)) end)
end
```

\cc @josevalim

PR #1483, opened 2016-06-13
Delegate prefix to migration runner

bokay:
When running the migrations for tenants with `mix ecto.migrate --prefix tenant0`, we should not run the migrations related to the public namespace.

This commit will allow to do the following:
```
def change do
  unless prefix do
    create_table(:tenants, prefix: :public) do
      add :identifier, :string
    end
  end
end
```

Discussion: https://github.com/elixir-lang/ecto/issues/1005#issuecomment-225506810
/cc @josevalim

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -519,6 +519,13 @@ defmodule Ecto.Migration do
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


josevalim:

PR #1484, opened 2016-06-14
correct relation! error msg

kagemusha:
all relations are now supported so unfound association is probably a naming error


josevalim:

Issue #1425, opened 2016-05-12
Support LATERAL joins


ashleyw:
http://www.postgresql.org/docs/9.3/static/queries-table-expressions.html#QUERIES-LATERAL
http://blog.heapanalytics.com/postgresqls-powerful-new-join-type-lateral/

It would be immensely valuable when combined with `subquery`, for instance getting the most recent comments on a post:

```
post = Repo.one(
  from p in Post,
  where: p.id == 123,
  lateral_join: cs in subquery(
    from c in Comment,
    order_by: [desc: c.inserted_at],
    where: c.post_id == p.id,
    limit: 5
  ),
  preload: [comments: cs]
)
```

It doesn't appear that using `INNER JOIN LATERAL` is possible even with a `fragment`, since `join` wraps the fragment in additional parentheses (i.e. `INNER JOIN (LATERAL (SELECT …))`), which isn't valid AFAIK.

josevalim:
If you would like to send a PR, it would be appreciated. :)

ashleyw:
I dived in and took a look, but I don't think I'm capable of implementing this I'm afraid. I've only been using Elixir for a few months. :)

josevalim:
@ashleyw no problem, we will tackle this after 2.0 is out, unless someone decides to handle it first. :)

aphillipo:
I started having a look into this yesterday and I came up with two ways to do it:

1) Change the Join Qualifier (qual) to include the join type (INNER), the JOIN keyword and optionally LATERAL, which would make the atom something like: :inner_join_lateral :left_join_lateral - we would need to change the adaptors to support this, removing " JOIN " which makes the code less explicit.

2) Change the %JoinExpr to add a :join_qual_suffix part of the struct, this would support other possible queries than LATERAL in the <join_qualifier> JOIN <:join_qual_suffix>  should they exists in the future.  However implementing this seems like there are a hell of a lot of moving parts to change; this would turn all of the join building macros from /7 to /8 and I started getting lost in the changes.

Obviously it's a classic trade off:

1) is simpler but feels very inflexible and if there is anything that adds things it means more entries for each supported join type 

2) gives extra possibilities for the future but if I'm honest seems difficult!

Let me know if you want me to push through with one of these and maybe I can ask questions if I get stuck!  Thanks!

ashleyw:
[...]

PR #1496, opened 2016-06-18
[WIP] Add lateral join for Postgres adapter.

bernardoamc:
The current code supports lateral joins like:

```
query =
  Game 
  |> join(:inner_lateral, [g], gs in fragment("SELECT * FROM games_sold AS gs WHERE gs.game_id = ? ORDER BY gs.sold_ON LIMIT 2", g.id))
  |> select([g, gs], {g.name, gs.sold_on})

[debug] QUERY OK db=61.6ms
SELECT g0."name", f1."sold_on"
FROM "games" AS g0
INNER JOIN LATERAL (SELECT * FROM games_sold AS gs WHERE gs.game_id = g0."id" ORDER BY gs.sold_ON LIMIT 2) AS f1 ON TRUE []
```

I created a [small app](https://github.com/bernardoamc/lateral_join) with the instructions required to test these changes locally.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -231,14 +231,14 @@ if Code.ensure_loaded?(Mariaex) do
         %JoinExpr{on: %QueryExpr{expr: expr}, qual: qual, ix: ix, source: source} ->
           {join, name} = get_source(query, sources, ix, source)
           qual = join_qual(qual)
-          "#{qual} JOIN " <> join <> " AS #{name} ON " <> expr(expr, sources, query)
+          "#{qual} " <> join <> " AS #{name} ON " <> expr(expr, sources, query)
       end)
     end
 
-    defp join_qual(:inner), do: "INNER"
-    defp join_qual(:left),  do: "LEFT OUTER"
-    defp join_qual(:right), do: "RIGHT OUTER"
-    defp join_qual(:full),  do: "FULL OUTER"
+    defp join_qual(:inner), do: "INNER JOIN"
+    defp join_qual(:left),  do: "LEFT OUTER JOIN"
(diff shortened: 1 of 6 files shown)


josevalim:

aptinio:
For reference: this fixes #1425

axelson:
Are lateral joins supported using keywords? Or only with expressions?

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -231,14 +231,14 @@ if Code.ensure_loaded?(Mariaex) do
...
     end
 
-    defp join_qual(:inner), do: "INNER"
-    defp join_qual(:left),  do: "LEFT OUTER"
-    defp join_qual(:right), do: "RIGHT OUTER"
-    defp join_qual(:full),  do: "FULL OUTER"
+    defp join_qual(:inner), do: "INNER JOIN"
+    defp join_qual(:left),  do: "LEFT OUTER JOIN"
+    defp join_qual(:right), do: "RIGHT OUTER JOIN"
+    defp join_qual(:full),  do: "FULL OUTER JOIN"
```
michalmuskala: Should we raise a message stating it's not supported if we encounter an unsupported join type, or is failing with FunctionClauseError enough?

josevalim: In master there is surely a raise.

josevalim:
> Are lateral joins supported using keywords? Or only with expressions?

Both.

ivanovaleksey:
Hi, lateral join can be executed only using `fragment`? Can I use `subquery`?

alex88:
@ivanovaleksey I don't think so since subquery doesn't let you use alias from the main query IIRC

PR #1502, opened 2016-06-20
simplify validate_change doc

mitchellhenke:
The current documentation uses an anonymous function with multiple clauses, which can be harder to understand, especially if someone isn't familiar with it.  I'm not sure if my change illustrates the best way to use validate_change, but hopefully it's closer.


josevalim:

PR #1508, opened 2016-06-23
Implements ON DELETE RESTRICT for PostgreSQL

evadne:
**Changes**
1. Changes `Reference` to support `ON DELETE RESTRICT`.
2. Amends documentation.

**Verification**
It works for my app.

**Questions**
1. Shall we also patch this for other databases?
2. Do we need additional test cases written?

Thank you!


josevalim:
ON DELETE restrict is the same as not specifying an ON DELETE at all, it is equivalent to `:nothing`. So why do we need this option? :) Thank you!

evadne:
Sorry I was trying to move an AR migration over. You can disregard this pull request.

PR #1516, opened 2016-06-26
Add Getting Started Guide

radar:
Continuing on from https://groups.google.com/forum/#!topic/elixir-ecto/9aFPh4BFo5U

This PR adds a "Getting Started" guide to Ecto, showing readers how to create, read, update and delete records in a new project. 

It'll need a bit of a tidy up, but it's a start.

- [x] Move "friends" example app into Ecto repo
- [ ] Get technical proof done by an Ecto expert
- [x] Re-read for sense-making
- [ ] Link to functions when they are first used
- [ ] Add `delete!` example failing with a `no_assoc_constraint`.


josevalim:
Thank you! We can keep the the pages at the project root, no need to put it
in lib. Also, great idea for including the examples project. I will proof
read everything before merge. :)

On Sunday, June 26, 2016, Ryan Bigg <notifications@github.com> wrote:

> Continuing on from
> https://groups.google.com/forum/#!topic/elixir-ecto/9aFPh4BFo5U
>
> This PR adds a "Getting Started" guide to Ecto, showing readers how to
> create, read, update and delete records in a new project.
>
> It'll need a bit of a tidy up, but it's a start.
>
>    - Move "friends" example app into Ecto repo
>    - Get technical proof done by an Ecto expert
>    - Re-read for sense-making
>
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/elixir-ecto/ecto/pull/1516
> Commit Summary
>
>    - Add Getting Started Guide
>
> File Changes
>
>    - *A* lib/ecto/pages/Getting Started.md
>    <https://github.com/elixir-ecto/ecto/pull/1516/files#diff-0> (679)
>    - *M* mix.exs
>    <https://github.com/elixir-ecto/ecto/pull/1516/files#diff-1> (3)
>
> Patch Links:
>
>    - https://github.com/elixir-ecto/ecto/pull/1516.patch
>    - https://github.com/elixir-ecto/ecto/pull/1516.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/elixir-ecto/ecto/pull/1516>, or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbuztpqJfOGzIVn6yVhLvPFFqZjqcks5qPchygaJpZM4I-dVC>
> .
>


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

radar:
I've moved the guide to `guides` at the root, similar to how "railties" does it.

josevalim:
Excellent. Once we have the example app, let's ship it!

radar:
@josevalim did you have a chance to proofread this yet? I think it should be proofed before it is shipped to ensure accuracy.

josevalim:
@radar Sorry, I was waiting for the example project before starting on that. Is that ok?

radar:
Oh yeah that's perfectly fine. I'll get it to you today or tomorrow. I'm going through and doing a proof myself first.

radar:
Ok, example app has been placed into `example/friends` and I've read through the guide and double checked things. Should be good for your eyes now :)

dustinfarris:
Thanks for this!

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+mix ecto.migrate
+```
+
+If we found out that we made a mistake in this migration, we could run `mix ecto.rollback` to undo the changes in the migration. We could then fix the changes in the migration and run `mix ecto.migrate` again. If we ran `mix ecto.rollback` now, it would delete the table that we just created.
+
+We now have a table created in our database. The next step that we'll need to do is to create the schema.
+
+## Creating the schema
+
+The schema is an Elixir representation of data from our database. Each schema doesn't necessarily need to be tied to a table; they can be tied to database views too.
```
dustinfarris: Could rephrase here, perhaps: "Schemas are commonly associated with a database table, however they can be associated with a database view as well."

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+
+  schema "people" do
+    field :first_name, :string
+    field :last_name, :string
+    field :age, :integer
+  end
+end
+```
+
+This model defines the schema from the database that this model maps to. In this case, we're telling Ecto that the `Friends.Person` model maps to the `people` table in the database, and the `first_name`, `last_name` and `age` fields in that table. The second argument passed to `field` tells Ecto how we want the information from the database to be represented in our model.
```
dustinfarris: This is the first mention of the word "model", seems abrupt and undefined in the scope of this document.  Perhaps we don't want to use the word model since Ecto is removing it?

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+
+Let's take a look at how we can insert data into the database.
+
+## Inserting data
+
+We can insert a new record into our `people` table with this code:
+
+```elixir
+person = %Friends.Person{}
+Friends.Repo.insert person
```
dustinfarris: I think parens are [recommended](https://github.com/niftyn8/elixir_style_guide) for all function calls.  This documentation would be a great place to set the example if that's correct.

examples/friends/test/friends_test.exs
```diff
@@ -0,0 +1,8 @@
+defmodule FriendsTest do
+  use ExUnit.Case
+  doctest Friends
+
+  test "the truth" do
+    assert 1 + 1 == 2
+  end
+end
```
michalmuskala: Let's remove this file or add some very basic tests

examples/friends/mix.exs
```diff
@@ -0,0 +1,36 @@
...
+  #
+  # Or git/path repositories:
+  #
+  #   {:mydep, git: "https://github.com/elixir-lang/mydep.git", tag: "0.1.0"}
+  #
+  # Type "mix help deps" for more examples and options
+  defp deps do
+    [
+      {:ecto, "2.0.1"},
+      {:postgrex, "0.11.2"}
```
michalmuskala: Let's use `path` dependency for ecto with a comment on how to use this in a proper app, similar to the other example: https://github.com/elixir-ecto/ecto/blob/master/examples/simple/mix.exs#L16-L19

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+```elixir
+defp deps do
+  [
+    {:ecto, "2.0.1"},
+    {:postgrex, "0.11.2"}
+  ]
+end
+```
+
+Ecto provides the common querying API, but we need the Postgrex adapter installed too, as that is what Ecto uses to speak in terms a PostgreSQL database can understand.
```
michalmuskala: It's a technicality, but `postgrex` is a driver, not adapter. The adapter is built into ecto.

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+
+  ...
+```
+
+This piece of configuration will start the Ecto process which receives and executes our application's queries. Without it, we wouldn't be able to query the database at all!
+
+There's one final bit of configuration that we'll need to add ourselves, since the generator does not add it. Underneath the configuration in `config/config.exs`, add this line:
+
+```elixir
+config :friends, ecto_repos: [Friends.Repo]
```
michalmuskala: Should this be done by the generator?

/cc @josevalim

michalmuskala:
I'm not sure how linking works in `ex_doc` with custom pages, but it should definitely be possible to link simply to modules or functions without adding explicit links.

guides/Getting Started.md
michalmuskala: Let's replace "model" with "schema".

I think a paragraph about the relation of ecto types and database types could be a good idea as well.
Ecto types are used in schema and define how the data should be represented in elixir. Types in migrations correspond to database types - both does not need to match necessarily.

examples/friends/test/friends_test.exs
radar: The idea of this `examples/friends` is to show the final point of the getting started guide. There are no tests in the guide, and this file isn't even mentioned so I haven't touched it.

guides/Getting Started.md
radar: imo, yes. I found it weird that it wasn't done already.

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+  { :ok, person } ->
+    # do something with person
+  { :error, changeset } ->
+    # do something with changeset
+end
+```
+
+**NOTE:** `changeset.valid?` will not check constraints (such as `uniqueness_constraint`). For that, you will need to attempt to do an insert and check for errors. It's for this reason it's best practice to try inserting data and validation the returned tuple from `Friends.Repo.insert` to get the correct errors, rather than relying on the changeset itself.
+
+If the insertion of the changeset succeeds, then you can do whatever you wish with the `person` returned in that result. If it fails, then you have access to the changeset and its errors. In the failure case, you may wish to present these errors to the end user.
```
michalmuskala: We could mention the `Ecto.Changeset.traverse_errors/2` function here that allows to easily format errors to show to the end user.

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+```
+
+The code between the angle brackets `<...>` here shows the Ecto query which has been constructed. We could construct this query ourselves with almost exactly the same syntax:
+
+```elixir
+require Ecto.Query
+Ecto.Query.from p in Friends.Person, order_by: [asc: p.id], limit: 1
+```
+
+We need to `require Ecto.Query` here so that the module is available for us to use. Then it's a matter of calling the `from` function from `Ecto.Query` and passing in the code from between the angle brackets. As we can see here, `Ecto.Query.first` saves us from having to specify the `order` and `limit` for the query.
```
michalmuskala: Require is needed to enables macros from a given module. Regular functions can be used freely without this.

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+** (Ecto.MultipleResultsError) expected at most one result but got 3 in query:
+
+from p in Friends.Person
+
+    lib/ecto/repo/queryable.ex:67: Ecto.Repo.Queryable.one/4
+```
+
+This happens because Ecto doesn't know what one record out of all the records
+returned that we want. Ecto will only return a result if we are explicit in
+our querying about which result we want.
```
michalmuskala: It might be good idea to mention here that when no record is returned `Repo.one` will return `nil`.

radar:
All comments addressed in the above commits.

---

> I'm not sure how linking works in ex_doc with custom pages, but it should definitely be possible to link simply to modules or functions without adding explicit links.

I don't have any clue on how to do that.

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+last_name = "Smith"
+Friends.Person |> Ecto.Query.where(last_name: ^last_name) |> Friends.Repo.all
+```
+
+Or:
+
+```elixir
+last_name = "Smith"
+Ecto.Query.from(p in Friends.Person, where: p.last_name == ^last_name) |> Friends.Repo.all
+```
```
michalmuskala: It might be good idea to explain why this is needed - the pin operator instructs the query builder to use parametrised SQL queries protecting against SQL injection

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+
+This means that you can also use a `case` statement to do different things depending on the outcome of the `update` function:
+
+```elixir
+case Friends.Repo.update(changeset) do
+  {:ok, person} ->
+    # do something with person
+  {:error, changeset}
+    # do something with changeset
+end
```
michalmuskala: I wonder - maybe it might be a good idea to introduce the `Ecto.Changeset.change/2` function here?

It allows to mark fields for change without validations and type-casting that is usually involved with the changeset function.

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+
+```elixir
+person = Friends.Repo.get(Friends.Person, 1)
+Friends.Repo.delete(person)
+{:ok,
+ %Friends.Person{__meta__: #Ecto.Schema.Metadata<:deleted>, age: 29,
+  first_name: "Ryan", id: 2, last_name: "Bigg"}}
+```
+
+Similar to `insert` and `update`, `delete` returns a tuple. If the deletion succeeds, then the first element in the tuple will be `:ok`, but if it fails then it will be an `:error`.
```
michalmuskala: Should `delete!` and `update!` be mentioned too?

examples/friends/test/friends_test.exs
michalmuskala: Makes sense

michalmuskala:
I added couple minor comments. Overall this looks fantastic. Great work, @radar!

guides/Getting Started.md
```diff
@@ -0,0 +1,696 @@
...
+  valid?: false>}
+```
+
+This means that you can also use a `case` statement to do different things depending on the outcome of the `update` function:
+
+```elixir
+case Friends.Repo.update(changeset) do
+  {:ok, person} ->
+    # do something with person
+  {:error, changeset}
```
gmile: Missing arrow `->`?

josevalim:
@radar just a heads up, I am currently travelling but I plan to tackle this as soon as I am back. :)

guides/Getting Started.md
radar: Done in 05549c9.

guides/Getting Started.md
radar: Thanks! I wasn't sure about the wording here and it's good to get confirmation of it. I've fixed this up in cc18463.

guides/Getting Started.md
radar: Thanks. Something else I did not know. Addressed in 1bfd8fe.

guides/Getting Started.md
radar: Are you suggesting that I should use `Ecto.Changset.change/2` here instead of `Friends.Person.changeset(person, %{first_name: ""})`? Should I use it like this:

```elixir
changeset = Ecto.Changeset.change(person, %{first_name: ""})
```

guides/Getting Started.md
radar: I've just added a mention of `update!` in the updating section. I am not sure how to get a `delete!` to fail though. Do you have an example of that?

radar:
> I added couple minor comments. Overall this looks fantastic. Great work, @radar!

Thank you very much @michalmuskala :)

OvermindDL1:
> I'm not sure how linking works in ex_doc with custom pages, but it should definitely be possible to link simply to modules or functions without adding explicit links.

For ex_doc all I do is just do:
```elixir
This will link to `MyModule.SubModule.function/3` within the same project when `mix doc` is run.
```
You can link to a Module, a specific function, even to source code.  It actually seems pretty powerful.

guides/Getting Started.md
michalmuskala: In general you use `cast` for dealing with external data that requires conversion and can't be trusted. `change` is much simpler in what it does - it simply adds the value to the changes, exactly as given.

Usually changeset functions use `cast` - but there might be valid use cases for `change` as well, for example:

```
def activate_changeset(struct) do
  change(struct, active: true)
end
```

guides/Getting Started.md
michalmuskala: `delete` might fail on constraints, a good example is the `no_assoc_constraint` that checks if there aren't any children associated to this parent before removing it.

michalmuskala:
@radar I'm sorry it took me so long to get back to you with your questions.

guides/Getting Started.md
radar: Thanks! Just fixed this in 16726a8.

guides/Getting Started.md
radar: Do you think an example of `delete` using a changeset with `no_assoc_constraint` would be fitting here?

radar:
@OvermindDL1 I just tried that and it doesn't provide a link for me.

OvermindDL1:
@radar Hmm, works for me.  See https://hexdocs.pm/ldap_ex/LDAPEx.Client.html for an example of in-page linking and https://hexdocs.pm/permission_ex/getting_started.html for cross-page linking that I did.  Check the corresponding github repositories for the code (which is also linked via the <> tags on the doc pages for quick access).

radar:
Thanks for the example! I'll give it a shot.

On Jul 6 2016, at 1:32 am, OvermindDL1 &lt;notifications@github.com&gt; wrote:  

> [@radar](https://github.com/radar) Hmm, works for me. See
<https://hexdocs.pm/ldap_ex/LDAPEx.Client.html> for an example of in-page
linking and <https://hexdocs.pm/permission_ex/getting_started.html> for cross-
page linking that I did. Check the corresponding github repositories for the
code (which is also linked via the &lt;&gt; tags on the doc pages for quick
access).

>

> —  
> You are receiving this because you were mentioned.  
> Reply to this email directly, [view it on GitHub](https://github.com/elixir-
ecto/ecto/pull/1516#issuecomment-230513673), or [mute the
thread](https://github.com/notifications/unsubscribe
/AAAKfxFMjf4Jui046UbvKkee6gRdfDpzks5qSnj7gaJpZM4I-
dVC).

josevalim:
Just a heads up I am allocating time to review this and hopefully merge it this weekend! 🙌

guides/Getting Started.md
josevalim: Yes, that would be a good example but we can also just mention it (we don't need to do it in practice). Up to you. :D

josevalim:
This is fantastic, thank you!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

I will merge this and do a last pass. I will also remove the simple application in favor of this one. :)

PR #1518, opened 2016-06-27
Fix unique_index complex constraints example

aaronrenner:
It looks like the two `unique_index` examples for complex constraints got mixed up slightly. This change just shows that the `unique_index` and `unique_constraint` should have the same name.


josevalim:

PR #1519, opened 2016-06-27
Permissive Multi naming

alexauritt:
resolves: #1505 
also see earlier discussion on closed PR: #1506


alexauritt:
@michalmuskala: We want to allow ANY term for a name, yes? Also I could use some guidance on where best to update the documentation.

michalmuskala:
Yes, I think we shouldn't impose any requirements at that point.

alexauritt:
@michalmuskala -- alright then -- gonna strip it down and keep it simple. Let me know if we're missing anything.

michalmuskala:
Thank you! ❤️

PR #1524, opened 2016-06-29
Fixes incorrectly named argument in @callback for Repo.one!

bradurani:
This was causing an incorrect argument name in the docs


fishcakez:
Thank you :heart:

Issue #1527, opened 2016-06-30
`Ecto.Changeset.unique_constraint/3` error messages not respected in many_to_many relationships


fourcube:
### Environment

* Elixir version (elixir -v): Erlang/OTP 18 [erts-7.3] [source] [64-bit] [smp:4:4] [async-threads:10] [kernel-poll:false]; Elixir 1.3.1
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.5
* Ecto version (mix deps): master
* Database adapter and version (mix deps): postgrex 0.11.2
* Operating system: Ubuntu 16.04

### Current behavior

Under certain conditions. unique key violations do not report the error messages that were defined through the `Ecto.Changeset.unique_constraint/3` function. This only happens when a changeset is applied to the database that inserts duplicate rows into a join table (e.g. because of a many-to-many relationship).

### Details

After my [post](https://groups.google.com/forum/#!topic/elixir-ecto/pwFw47WMFMo) to the mailing list @josevalim advised me to open up an issue here. I'll just copy the example that I posted to the mailing list, since it is pretty complete.

My little phoenix app deals with people that get assigned to tasks. I want to use a join table, because I will attach data to the task assignments later on. Therefor I have defined the following models:

```elixir
# My Task schema
schema "tasks" do
  field :name, :string
  field :description, :string, default: ""
  field :date, Ecto.Date

  many_to_many :workers, Worker, join_through: WorkerTaskAssociation

  timestamps()
end

# My Worker schema
schema "workers" do
  field :name, :string
  field :active, :boolean, default: false
  field :archived, :boolean, default: false

  many_to_many :tasks, Task, join_through: WorkerTaskAssociation

  timestamps()
end

# The join table schema
schema "workers_tasks" do
  belongs_to :worker, Worker
  belongs_to :task, Task

  timestamps()
end 
```

This is how my join table is defined

```elixir
create table(:workers_tasks) do
  add :worker_id, references(:workers)
  add :task_id, references(:tasks)
  timestamps
end

# It also has a unique index over the 'worker_id' and 'task_id' fields
create index(:workers_tasks, [:worker_id, :task_id], unique: true) 
```

In one of my unit tests, I try to validate that a certain error is thrown, when a worker is assigned to a task twice.

```elixir
# Create a task
task = %{name: "foo", date: %{day: 15, month: 2, year: 2016}}
task = Task.create_changeset(%Task{}, task)
[...]

PR #1528, opened 2016-06-30
Fix unique constraint errors for many_to_many inserts (#1527)

fourcube:
Under certain conditions. unique key violations do not report the error messages that were defined through the `Ecto.Changeset.unique_constraint/3` function. This only happens when a changeset is applied to the database that inserts duplicate rows into a join table (e.g. because of a many-to-many relationship).

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -233,6 +233,13 @@ defmodule Ecto.Association do
               {[changeset|changesets], structs, halt, valid?}
             {:ok, struct} ->
               {[changeset|changesets], [struct | structs], halt, valid?}
+            {:error, {error_changeset, :insert_join}} ->
+              # Roll the errors originating from a join into the parent_changeset
+              [last_changeset | changesets] = changesets
+              errors = last_changeset.errors ++ error_changeset.errors
+              merged_changeset = %{last_changeset | errors: errors, valid?: error_changeset.valid?}
+
+              {[merged_changeset|changesets], structs, halted?(halt, changeset, error_changeset), false}
             {:error, error_changeset} ->
               {[error_changeset|changesets], structs, halted?(halt, changeset, error_changeset), false}
           end
@@ -939,7 +946,7 @@ defmodule Ecto.Association.ManyToMany do
           data = [{join_owner_key, owner_value}, {join_related_key, related_value}]
 
           case insert_join(repo, join_through, data, opts, constraints) do
-            {:error, changeset} -> {:error, changeset}
+            {:error, changeset} -> {:error, {changeset, :insert_join}}
             _                   -> {:ok, related}
           end
         else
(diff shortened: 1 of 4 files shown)


fourcube:
This is probably not the best way to implement this fix. I'd be happy to change stuff around and learn something during the process!

michalmuskala:
Since the constraint is added to the parent changeset, I'm not sure I understand why the error appears suddenly in the child changeset.

josevalim:
Thank you @fourcube! I really appreciate that you included the tests, sometimes it is the ones that take the longest to write. However, could you please try to write such tests using the existing schemas and associations? We usually try to keep them to a minimum otherwise they become harder and harder to maintain. ❤️

fourcube:
@michalmuskala the code is propagating all parent changeset constraints to the child changeset (the one that performs the insert into the join table). In the original code, the join table row was inserted through an `Ecto.Repo.insert!` call, raising an error and not keeping any reference to the child changeset. Now it is using a "normal" `insert` so the error is returned together with the child changeset. 

Maybe I can find a way to get rid of the child changeset (like in the original code) and roll the errors into the parent changeset.

@josevalim sure that's understandable! I'll change the test case so it uses the existing schemas.

fourcube:
I've changed the code so it no longer returns the insert changeset created during the call to `insert_join`. The test cases now use the existing schemas.

fourcube:
@michalmuskala even though I have changed things like you and @josevalim suggested, I am not 100% sure about the correctness of my solution. 

Would you like me to take another stab at this? If so, could you point me in the right direction so I can get this into a mergeable state?

josevalim:
Hi @fourcube! At first glance those changes look good for me. However, I don't have time right now to fully review it. But no worries, if any changes are necessary in the future and you are not around, we can take care of them ourselves. Thank you1

lib/ecto/association.ex
```diff
@@ -233,6 +233,18 @@ defmodule Ecto.Association do
               {[changeset|changesets], structs, halt, valid?}
             {:ok, struct} ->
               {[changeset|changesets], [struct | structs], halt, valid?}
+            {:error, {error_changeset, :insert_join}} ->
+              # Roll the errors originating from a join into the parent_changeset
+              case changesets do
```
josevalim: Shouldn't we add the errors to the `changeset` instead of revisiting the `changesets` list?

josevalim:
@fourcube, sorry for the delay. I am a bit worried about implicitly moving the constraints from the parent changeset to the parent. Although I think it would be OK as they don't change how the database behaves and keys are meant to be unique anyway. The alternative would be to tell which association the constraint belongs to in `validate_unique` but I think it would be rather more complex with no gains.

@fourcube do you think we could add some docs to many_to_many about constraints in the join table and mentioning that they can be directly added to the parent changeset since writing to the join table is hidden away from the developer?

lib/ecto/association.ex
fourcube: I'm not sure at which point in the code I can update the parent changeset (the one which contains the constraint and from which the insert_join originates).

fourcube:
@josevalim thanks for taking another look at this.

I will add some documentation on the constraint propagation.

josevalim:

Issue #1531, opened 2016-07-01
Add comment option to migration


yaBliznyk:
I think this must looks like this:

```
create table(:table_name, comment: "Table comment") do
      add :title, :string, size: 150, null: false, comment: "My comment to title"
end
```

I try execute this code and not seen comment in table.
http://joxi.ru/DmB7j3QuNon3q2
"Комментарий" is comment in Russian))

http://joxi.ru/82QWlGkf1G0X3r
Documentation don`t have option "comment"

eniolopes:
I took the liberty to look into doing this for PostgreSQL.

The way comments are added to tables and columns on PostgreSQL is through a separate query, which means there is no special constraint to add the comment when creating the table. I've got something working for adding the comment to a table [here](https://github.com/eniolopes/ecto/commit/8d47685d71438df439b7d7ec6331e0f6e088ddaa).
Please advise if this approach is ok, I'm not sure if adding  `comment` to `%Table{}` is desired or not.

Following this same implementation for columns will be a little tricky, because options for columns are only extracted via `column_definitions/2`, which means I won't be able to concatenate the query to add comments at the end. What I can think of is having a function that builds comments queries for columns and call it in `excute_ddl/3`.

PS: I wasn't intending to implement this without asking permission first, sorry. I actually was just reading through the code to see if I could find it and it turned out not so hard of change.

josevalim:
@eniolopes it is totally fine to go ahead and implement it. :D Putting it in `%Table` is the best option for now.

However, if the comment option is not really supported by columns and we would have to collect them, I would rather prefer to not support them. Something else to consider is how comments are going to work with alter table comments.

eniolopes:
@josevalim cool, will go ahead then. I think we will be fine supporting it for columns as well. Will take a look at the behaviour for alter tables. Thanks :)

PR #1534, opened 2016-07-02
Adds support for commenting on PostgreSQL

eniolopes:
TODO:

- [x] Support comments when doing changes to tables and columns (ALTER TABLE)
- [x] Support comments on constraints and anything else supported by us on [this list](https://www.postgresql.org/docs/9.5/static/sql-comment.html)

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -604,7 +604,8 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     def execute_ddl({:create, %Constraint{}=constraint}) do
-      "ALTER TABLE #{quote_table(constraint.prefix, constraint.table)} ADD #{new_constraint_expr(constraint)}"
+      "ALTER TABLE #{quote_table(constraint.prefix, constraint.table)} ADD #{new_constraint_expr(constraint)}" <>
+      comment_on(:constraint, constraint.name, constraint.comment)
     end
 
     def execute_ddl({:drop, %Constraint{}=constraint}) do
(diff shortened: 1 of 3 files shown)


eniolopes:
@josevalim any ideas on why this would be falling? locally tests are all green

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -803,6 +829,8 @@ if Code.ensure_loaded?(Postgrex) do
       <<?", name::binary, ?">>
     end
 
+    defp single_quote(value), do: "'#{value}'"
```
josevalim: Shouldn't we also escape any quote inside `value`?

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -626,6 +633,25 @@ if Code.ensure_loaded?(Postgrex) do
       end
     end
 
+    defp comment_on(_database_object, _name, nil), do:  ""
+    defp comment_on(:column, {table_name, column_name}, comment) do
+      column_name = quote_table(table_name, column_name)
+      "; COMMENT ON COLUMN #{column_name} IS #{single_quote(comment)}"
+    end
+
+    defp comment_on(database_object, name, comment) do
```
josevalim: Instead of doing Atom.to_string |> String.upcase, we should explicitly list everything we support. :)

josevalim:
The build failures are unrelated, so please go ahead!

eniolopes:
@josevalim please advise if commenting should be supported when renaming tables and columns, I think these are the only ones that I didn't add support for it.

Also, should anything be added to the integration tests specifically for this? If so, please advise how it should be done.

I want to be able to finish this today. Thanks for the support and opportunity to contribute ❤️

eniolopes:
This close https://github.com/elixir-ecto/ecto/issues/1531

josevalim:

PR #1536, opened 2016-07-04
Remove unnecessary Map.delete

atsheehan:
I may be missing something subtle (or obvious), but the `Map.delete` call is only made when the key does not exist in the map, so the function should always return the map unchanged. Is there a scenario where `Map.delete(changes, key)` will have an effect on `changes`?


josevalim:

Issue #1538, opened 2016-07-05
Umbrella app mix test migration alias error


hunterboerner:
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 18 [erts-7.3] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.3.1
```

* Database and version: PostgreSQL 9.4.4
* Ecto version (mix deps):
```
* ecto 2.0.2 (Hex package) (mix)
  locked at 2.0.2 (ecto) b02331c1
```
* Database adapter and version (mix deps):
```
* postgrex 0.11.2 (Hex package) (mix)
  locked at 0.11.2 (postgrex) 139755c1
```
* Operating system: macOS El Capitan

### Current behavior

I'm having trouble using this in an umbrella app. I have app A which defines schemas and has migrations and I have app B that depends on app A (in_umbrella: true). If I add the alias to mix.exs for both of those apps and then run the tests separately, everything works. However, when I run mix test in the umbrella app, it fails.

```
Randomized with seed 54646
==> AppA
....................................................

Finished in 0.5 seconds
52 tests, 0 failures

Randomized with seed 188921
** (DBConnection.OwnershipError) cannot find ownership process for #PID<0.46.0>.

When using ownership, you must manage connections in one
of the three ways:

  * By explicitly checking out a connection
  * By explicitly allowing a spawned process
  * By running the pool in shared mode

The first two options require every new process to explicitly
check a connection out or be allowed by calling checkout or
allow respectively.

The third option requires a {:shared, pid} mode to be set.
If using shared mode in tests, make sure your tests are not
async.

If you are reading this error, it means you have not done one
of the steps above or that the owner process has crashed.

See Ecto.Adapters.SQL.Sandbox docs for more information.
    (db_connection) lib/db_connection.ex:722: DBConnection.checkout/2
    (db_connection) lib/db_connection.ex:623: DBConnection.run/3
    (db_connection) lib/db_connection.ex:845: DBConnection.run_meter/3
    (db_connection) lib/db_connection.ex:464: DBConnection.prepare_execute/4
    (ecto) lib/ecto/adapters/postgres/connection.ex:82: Ecto.Adapters.Postgres.Connection.execute/4
    (ecto) lib/ecto/adapters/sql.ex:228: Ecto.Adapters.SQL.sql_call/6
[...]

PR #1545, opened 2016-07-07
port fix from #1538 to ecto.rollback

paulswartz:
#1538 fixed `mix ecto.migrate` when running with the Sandbox. `ecto.rollback` has the same issue and this ports the fix from that change.


josevalim:

PR #1548, opened 2016-07-08
Export Ecto.Adapters.SQL.query/query! directly in Ecto.Repo module

### LINKED ISSUE #1515: Export Ecto.Adapters.SQL.query/query! directly in SQL repositories

imranismail:
This adds the `Ecto.Adapters.SQL.query/query!` function directly to the Repo module.

I've used a guard clause against an attribute that defines the sql adapters in the `Ecto.Queryable` module

This should close #1515

### DIFF
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -3,8 +3,6 @@ defmodule Ecto.Repo.Queryable do
   # for query related functionality.
   @moduledoc false
 
-  @sql_adapters [Ecto.Adapters.Postgres, Ecto.Adapters.MySQL]
-
   alias Ecto.Query
   alias Ecto.Queryable
   alias Ecto.Query.Planner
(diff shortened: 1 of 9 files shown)


michalmuskala:
I'm really not sure about this functionality. While I can see it's usefullness, it adds SQL-specific functions to the repo interface. A function like this for mongo would be great too, the problem is there are more than just one functions possible.

/cc @josevalim

lib/ecto/repo/queryable.ex
```diff
@@ -3,6 +3,8 @@ defmodule Ecto.Repo.Queryable do
   # for query related functionality.
   @moduledoc false
 
+  @sql_adapters [Ecto.Adapters.Postgres, Ecto.Adapters.MySQL]
```
michalmuskala: There are other not built-in SQL adapters. We can't assume we know all of them.

josevalim:
@imranismail the implementation should be added to Repo. The adapter should generate those functions in the repository when the adapter functionality is included. Would this change reduce your concerns, @michalmuskala?

Tica2:
Yes! is up to adapter to generate functions in the repository or not.

michalmuskala:
Yes, adapter generating those in `before_compile` is great 👍

imranismail:
Alright, I'll make those changes 😁

josevalim:
Thank you! Can you please change some tests? We have some tests in integration_test/sql/query.exs (or something similar that used the "old way", we just need to rewrite them to use the new way. :)

josevalim:
We should also make the third argument optional in Ecto.Adapters.SQL as well.

imranismail:
Changed some tests that were using the old way and made `Ecto.Adapters.SQL.query/query!` third argument optional.

josevalim:

PR #1552, opened 2016-07-09
add error from `validate_required([:name, :email])`

koolhazcker:
Or maybe it would be better to just remove [`validate_required([:name, :email])`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L43) validation from the changeset? It's not even mentioned in the [paragraph](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L50-L52) below the code: 

> In the `changeset/2` function above, we define two validations -
   one for checking the e-mail format and another to check the age -
   as well as a unique constraint in the email field.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -47,9 +47,10 @@ defmodule Ecto.Changeset do
         end
       end
 
-  In the `changeset/2` function above, we define two validations -
-  one for checking the e-mail format and another to check the age -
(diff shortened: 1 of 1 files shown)


josevalim:
@koolhazcker we should also update the paragraph above. Or remove validate_required. It is up to you. Can you please do any of the changes above? :) Thank you!

josevalim:

PR #1555, opened 2016-07-10
fix typo

koolhazcker:
Params can only have keys of `String.t` type, if I read [this](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L148) correctly.


josevalim:

PR #1557, opened 2016-07-10
fix typo

koolhazcker:
Since `cast/4` is deprecated.

Should I also edit [this](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L596-599) paragraph? Is the `optional` field still there?

> `required` and `optional` - they are merged; all the fields that appear
    in the optional list of either changesets and also in the required list of
    the other changeset are moved to the required list of the resulting
    changeset.


josevalim:
Yes, please fix the other paragraph as optional is gone. :)

josevalim:

PR #1559, opened 2016-07-10
replace cast/4 with cast/3 in docs

koolhazcker:
Since `cast/4` is deprecated.


josevalim:

PR #1560, opened 2016-07-10
add Ecto.Changeset.validate_acceptance/2

Kr00lIX:
`validate_acceptance/2` encapsulates the pattern of wanting to validate the acceptance of a terms of service check box (or similar agreement). 

Example:
```elixir
# Model
defmodule Shop.Booking do
  use Shop.Web, :model

  schema "bookings" do
    field :email, :string
    field :terms, :boolean, virtual: true
  end

    struct
    |> cast(params, [:email, :terms])
    |> validate_acceptance(:terms)
  end
end
```

```slim
/ Slim template
= form_for @changeset, fn f ->
  .form-group
    = label f, :email, class: "control-label"
    = text_input f, :email, class: "form-control"

  .checkbox
    label
      = checkbox f, :terms
      | Terms

```

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1400,6 +1400,30 @@ defmodule Ecto.Changeset do
     Keyword.get(opts, key, default)
   end
 
+  @doc """
+  Validates that a checkbox was checked.
+
+  ## Options
+
+    * `:message` - the message on failure, defaults to "must be accepted"
+
+  ## Examples
+
+      validate_acceptance(changeset, :terms_of_service)
+      validate_acceptance(changeset, :rules, message: "please accept rules")
+
+  """
+  @spec validate_acceptance(t, atom, Keyword.t) :: t
+  def validate_acceptance(%{params: params} = changeset, field, opts \\ []) do
+    param = Atom.to_string(field)
+    value = Map.get(params, param)
+
+    case Ecto.Type.cast(:boolean, value) do
+      {:ok, true} -> changeset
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @Kr00lIX. I think functionality such as above would be OK but I would prefer if we didn't have to add a field to the schema for this to work. This means we would need to check it directly in the parameters and we would need to validate it works well with Phoenix HTML.

Kr00lIX:
@josevalim, thank you for review. Did you mean check this params in changeset like code bellow?
```elixir 
defmodule Shop.Booking do
  use Shop.Web, :model

  schema "bookings" do
    field :email, :string
  end

    struct
    |> cast(params, [:email])
    |> validate_acceptance(params, :terms)
  end
end
```
with error in changest for invalid
`changeset.errors == [terms: {"must be accepted", []}]`

josevalim:
@Kr00lIX you don't need to pass the parameters per se because they are in the changeset. It is just that implementation wise we will look if params has such key and if so we will validate it is true.

Kr00lIX:
@josevalim, updated code, Is it ok now?

leejarvis:
This looks good. I think it would be good if the function signature matched `validate_required` though and accept a list of fields. WDYT?

josevalim:
@leejarvis validate_required is rather the exception than the rule so I would keep it as is for now.

josevalim:

leejarvis:
Yep fair enough ✨

PR #1563, opened 2016-07-11
replace downcase(?) with lower(?) in the Fragments example

koolhazcker:
I understand that the use of `downcase(?)` was just for the sake of example, but wouldn't it be better to use the real function instead (both for postgres and mysql)?


michalmuskala:
Thank you! ❤️

PR #1567, opened 2016-07-11
replace downcase(?) with lower(?) in the Fragments example

koolhazcker:
Same reasoning as in [#1563](https://github.com/elixir-ecto/ecto/pull/1563).

> I understand that the use of `downcase(?)` was just for the sake of example, but wouldn't it be better to use the real function instead (both for postgres and mysql)?

Sorry for submitting that again. Should have changed it all in one PR.


josevalim:

PR #1568, opened 2016-07-11
typo?

koolhazcker:
I am not sure if that's actually a typo, but should not the URL be the same as in [`Ecto` docs](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto.ex#L48)?


josevalim:
The scheme does not matter, so both are correct. I think I am more inclined to use "ecto" even to avoid confusion. :)

koolhazcker:
Ah, okay.

PR #1572, opened 2016-07-13
Keep lists of integers as lists when inspecting queries

michalmuskala:
@josevalim I'm opening this as a PR as I'm not sure about this change.

Is there a better way than to pattern match on the result of `Macro.to_string/2`?

### DIFF
--- a/lib/ecto/query/inspect.ex
+++ b/lib/ecto/query/inspect.ex
@@ -150,12 +150,10 @@ defimpl Inspect, for: Ecto.Query do
   end
 
   defp expr_to_string({:^, _, [ix]}, _, _, %{params: params}) do
-    escaped =
-      case Enum.at(params || [], ix) do
-        {value, _type} -> Macro.escape(value)
-        _              -> {:..., [], nil}
-      end
-    Macro.to_string {:^, [], [escaped]}
+    case Enum.at(params || [], ix) do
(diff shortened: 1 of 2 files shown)


lib/ecto/query/inspect.ex
```diff
@@ -155,7 +155,12 @@ defimpl Inspect, for: Ecto.Query do
         {value, _type} -> Macro.escape(value)
         _              -> {:..., [], nil}
       end
-    Macro.to_string {:^, [], [escaped]}
+    case Macro.to_string({:^, [], [escaped]}) do
```
josevalim: I think we should change the whole function to something like this:

```elixir
  defp expr_to_string({:^, _, [ix]}, _, _, %{params: params}) do
    case Enum.at(params || [], ix) do
      {value, _type} -> "^" <> inspect(value, char_lists: :as_lists)
      _              -> "^..."
    end
  end
```

Thoughts?

michalmuskala: Makes sense

michalmuskala: Should we `Macro.escape` the value before inspecting?

josevalim: I don't think so. Macro escaping is actually forcing it to be inspected.

michalmuskala:
@josevalim updated

josevalim:

Issue #1525, opened 2016-06-29
`Ecto.Repo.Assoc.merge/3` failure


OvermindDL1:
### Precheck

Quick google revealed nothing.  Following post is slightly sanitized for job reasons but is accurate.

### Environment

* Elixir version:  Elixir 1.3.1, Erlang/OTP 18
* Database and version:  PostgreSQL 9.5
* Ecto version:  ecto 2.0.1
* Database adapter and version (mix deps):  postgrex 0.11.2
* Operating system:  Windows 10 and Debian both

### Current behavior

Built a SQL query of the form:
```elixir
    now = Ecto.Date.utc
    from s in Student,
      join: ssems in assoc(s, :student_sems),
      join: sreqs in assoc(s, :student_reqs),
      where: ssems.uid == s.uid
        and ssems.start <= ^now
        and ssems.end >= ^now,
      preload: [:student_sems, student_reqs: sreqs],
      select: s
```
Yes the :student_sems not being linked to ssems is on purpose, the issue happens without it as well.

A reduced test is:
```elixir
    from s in Student,
      join: sreqs in assoc(s, :student_reqs),
      preload: [student_reqs: sreqs]
```
The Student schema is:
```elixir
  @primary_key {:uid, :integer, autogenerate: false}
  schema "students" do
    field :preferred_name, :string
    field :personal_email, :string
    field :comment, :string

    has_many :student_reqs, Student.Requirement, foreign_key: :uid
    has_many :student_sems, Student.Semester, foreign_key: :uid

    # Why the heck do these not want to link, plus still need access to the intermediate table regardless and it seems these do not supply that...
    # many_to_many :requirements, Requirement,
    #   join_through: Student.Requirement,
    #   join_keys: [uid: :uid, requirement_id: :id]
    #
    # many_to_many :semesters, Semester,
    #   join_through: Student.Semester,
    #   join_keys: [uid: :uid, semester_id: :id]

    timestamps
  end
```
And the Student.Requirement schema is:
```elixir
  @primary_key false
  schema "students_requirements" do
    belongs_to :student, Student, foreign_key: :uid, references: :uid, primary_key: true
    belongs_to :requirement, Requirement, primary_key: true
    field :date, Ecto.Date, primary_key: true
    field :location, :string
    field :method, :string
    field :comment, :string

    timestamps
  end
```

Trying to run the above (reduced testcase) query (just via Repo.all) returns this:
```elixir
[debug] QUERY OK db=16.0ms
[...]

PR #1574, opened 2016-07-14
fix pattern matching error #1525

slavaZim:
[https://github.com/elixir-ecto/ecto/issues/1525](url) referencing this issue

### DIFF
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -52,6 +52,7 @@ defmodule Ecto.Integration.Post do
     has_many :users_comments, through: [:users, :comments]
     has_many :comments_authors_permalinks, through: [:comments_authors, :permalink]
     timestamps()
+    has_one :post_user_composite_pk, Ecto.Integration.PostUserCompositePk
   end
 
   def changeset(schema, params) do
--- a/lib/ecto/repo/assoc.ex
+++ b/lib/ecto/repo/assoc.ex
@@ -37,9 +37,10 @@ defmodule Ecto.Repo.Assoc do
   defp merge([struct|sub_structs], {keys, dict, sub_dicts}, parent_key) do
     child_key =
       if struct do
-        [{_, key}] = Ecto.primary_key!(struct)
-        key || raise Ecto.NoPrimaryKeyValueError, struct: struct
+        Ecto.primary_key!(struct)
+        |> Enum.map(&elem(&1, 1))
       end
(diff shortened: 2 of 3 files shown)


lib/ecto/repo/assoc.ex
```diff
@@ -37,9 +37,13 @@ defmodule Ecto.Repo.Assoc do
   defp merge([struct|sub_structs], {keys, dict, sub_dicts}, parent_key) do
     child_key =
       if struct do
-        [{_, key}] = Ecto.primary_key!(struct)
-        key || raise Ecto.NoPrimaryKeyValueError, struct: struct
+        Ecto.primary_key!(struct)
+        |> Enum.reduce([], fn(x, acc) ->
+          {_, key} = x
+          [key | acc]
+        end)
```
michalmuskala: Wouldn't `|> Enum.map(&elem(&1, 1))` be enough?

lib/ecto/repo/assoc.ex
```diff
@@ -37,9 +37,13 @@ defmodule Ecto.Repo.Assoc do
...
       end
```
michalmuskala: In that case we also need an `else` clause to return an empty list

lib/ecto/repo/assoc.ex
```diff
@@ -37,9 +37,13 @@ defmodule Ecto.Repo.Assoc do
...
+    if Enum.empty?(child_key), do: raise Ecto.NoPrimaryKeyValueError, struct: struct
```
michalmuskala: The error should be raised if any of the primary keys has value `nil`, not if there are no primary keys.

slavaZim: @michalmuskala, so should it be raised if at least one is nil in case of composite keys?

integration_test/support/schemas.exs
```diff
@@ -262,3 +263,19 @@ defmodule Ecto.Integration.PostUserCompositePk do
     timestamps()
   end
 end
+
+defmodule Ecto.Integration.AccosiationWithCompositePk do
```
josevalim: Can we try to use one of the existing schemas instead of introducing a new one?

lib/ecto/repo/assoc.ex
michalmuskala: Yes, I think so. It doesn't make sense to have `NULL` in one of the parts of a composite key, does it?

PR #1576, opened 2016-07-14
Handle not loaded associations in get/fetch_field

michalmuskala:
We should do exactly the same thing that we do when we try to
cast/change the associations when you try to access them and they are
not loaded


michalmuskala:
Signed off by @josevalim over on IRC

PR #1578, opened 2016-07-15
Clear logic on missing? method

asiniy:
I think this approach is more obvious since we have inner logic conditions. Here logic conditions are placed on one level


josevalim:

Issue #1554, opened 2016-07-09
`cast` mentioning field not in the scheme produces confusing error message


marick:
Here is a definition of a schema:

```elixir
  schema "users" do
    field :display_name, :string
  end
```
I want to validate `display_name` in the model. But I make a typo:

```elixir
    cast(starting_value, additions, ~w(display_name_TYPO), [])
```

The error message doesn't show that the name was not in the schema:

```
iex(2)> User.cs_this_fails(%User{}, %{})
** (ArgumentError) argument error
             :erlang.binary_to_existing_atom("display_name_typo", :utf8)
      (ecto) lib/ecto/changeset.ex:422: Ecto.Changeset.cast_key/1
      (ecto) lib/ecto/changeset.ex:390: Ecto.Changeset.process_param/6
    (elixir) lib/enum.ex:1246: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
      (ecto) lib/ecto/changeset.ex:371: Ecto.Changeset.do_cast/6
```

As an Elixir/Phoenix novice, it took me a long time to figure this out. I ask that there be a better error message, something like `:display_name_TYPO is not part of the schema`. 

### Environment

* Elixir 1.3.0
* psql (PostgreSQL) 9.5.3
* ecto 2.0.2 (Hex package) (mix)

josevalim:
@marick agreed, thank you. If you would like to submit a PR, we would appreciate it. I would suggest wrapping the String.to_existing_atom/1 call into a `try/rescue ArgumentError` and raise an error message such as: "could not convert the parameter "..." into an atom. This typically implies there is no field with such name in the schema".

mattludwigs:
@josevalim @marick my fork has the updates for this so I can submit a PR if that is okay?

PR #1580, opened 2016-07-16
better missing schema field error message during cast

mattludwigs:
Provide a more helpful error message when trying to use a field in a schema that does exist when calling `cast`.
Issue: #1554

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -424,7 +424,7 @@ defmodule Ecto.Changeset do
       {String.to_existing_atom(key), key}
     rescue
       ArgumentError ->
-        raise "could not convert the parameter #{key} into an atom, #{key} is not in schema"
+        raise "could not convert the parameter #{key} into an atom, #{key} is not in the schema."
     end
   end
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -419,8 +419,14 @@ defmodule Ecto.Changeset do
...
   end
 
-  defp cast_key(key) when is_binary(key),
-    do: {String.to_existing_atom(key), key}
+  defp cast_key(key) when is_binary(key) do
+    try do
+      {String.to_existing_atom(key), key}
+    rescue
+      ArgumentError ->
+        raise "could not convert the parameter #{key} into an atom, #{key} is not in the schema."
```
josevalim: Let's go with this message please:

    raise ArgumentError, "could not convert the parameter `#{key}` into an atom, `#{key}` is not a schema field"

The changes were:

    1. Wrap keys in backticks
    2. Remove trailing dot
    3. Raise it still as ArgumentError

Can you please amend? Thank you!

josevalim:

Issue #1584, opened 2016-07-18
Repo.insert replaces embedded Changeset with Struct


seangeo:
Hi there,

We've found a problem with Changesets being replaced for embedded objects when a Repo.insert fails. It seems to happen when inserted into a table that has both an association and an embedded record and the associated record has a association of it's own.

I've created a minimal, failing example of this in a Phoenix app at https://github.com/sendle/ecto_embed_insert_bug

To describe the schema, it looks something like this:

[Parent](https://github.com/sendle/ecto_embed_insert_bug/blob/master/web/models/parent.ex)

```elixir
defmodule Parent do
  schema "parents" do
    timestamps
  end
end
```

[Child](https://github.com/sendle/ecto_embed_insert_bug/blob/master/web/models/child.ex)
```elixir
defmodule Child do
  schema "children" do      
    belongs_to :parent, EctoEmbedInsertBug.Parent
  end

  def changeset(struct, params \\ %{}) do
    struct
    |> cast(params, [:parent_id])
    |> assoc_constraint(:parent)
  end
end
```

[GrandChild](https://github.com/sendle/ecto_embed_insert_bug/blob/master/web/models/grand_child.ex)
```elixir
defmodule Grandchild do
  defmodule Embed do
    use EctoEmbedInsertBug.Web, :model

    embedded_schema do
      field :something, :string
    end

    def changeset(data, params) do
      data
      |> cast(params, [:something])    end
  end

  schema "grand_children" do
    belongs_to :child, Child
    embeds_one :embed, GrandChild.Embed

    timestamps()
  end

  def changeset(struct, params \\ %{}) do
    struct
    |> cast(params, [:child_id])
    |> cast_embed(:embed, required: true)
    |> cast_assoc(:child)
    |> assoc_constraint(:child)
  end
end
```

If you attempt to create changeset for a `GrandChild`, along with a `Child` but referencing a Parent that doesn't exist, like so:

```elixir
changeset = GrandChild.changeset(%GrandChild{}, 
   %{embed: %{something: "test"}, child: %{parent_id: 123}})
```

You correctly get a changeset with a nested changeset at `changeset.changes.embed`.

If you then try to insert that changeset, using `Repo.insert(changeset)`, you correctly get an error back that contains a changeset for the child with the error about the missing parent, but the embed changeset at `changeset.changes.embed` has been converted into a `GrandChild.Embed` struct, which means passing the changeset to `traverse_errors/2` fails.

[...]

PR #1587, opened 2016-07-18
Fixes embed changeset corruption when parent insert fails.

seangeo:
This fixes the bug reported in #1584. The issue was caused by the fact that embedded changes are already replaced in the changeset with their target struct before the parent records in the changeset are processed. So if a parent insert causes an error, we end up returning a changeset with the replaced struct, not the original changeset for the embedded object. I've fixed this by checking if the changeset is still valid after attempting to process the parents, and only after that check are the embed changesets converted to structs.

I've gone for a minimal set of changes to get this test to pass (and not break anything else). But if there is anything I should refactor, let me know.


josevalim:
@seangeo great job! Can you please make sure we also don't have the same bug on updates? :D

seangeo:
Will do. Thanks 👍

josevalim:

PR #1589, opened 2016-07-19
More flexible repo constraint violation matching

vietord:
Allows changeset constraints to map to repo constraints as a suffix, as an alternative to the default behavior of exact matching. 

This allows users more flexibility in mapping low-level backend errors to higher-level errors, and is a key piece of functionality for high-scale applications which utilize advanced Repo management techniques which are incompatible with the current constraint mapping. 

Because constraints and accesses are implicitly scoped to a table access, there should be no confusion as a result of similar constraint match clauses for different tables. Additionally one could, if desired, use this to write a wildcard style matching or other flexible aggregation for constraints in normal single-table cases.

Motivating Use Case
====

I'm using Ecto to manage a very large dataset which takes advantage of Postgres partitioning to operate efficiently. A result of this is that what appears as one Repo is actually in the backend many individual tables, transparently routed to within postgres and completely compatible with Ecto, except when constraint violations are raised. 

The key trouble is that index names must be unique, and so they are prefixed by the table names. When you insert to the root table, and PG performs the actual insert to the appropriate partition, constraint violations result in a different name than expected

A constraint defined as:

`|> unique_constraint(:column_a, name: :table_column_a_idx)`

May result in: 
```
     ** (Ecto.ConstraintError) constraint error when attempting to insert struct:
     
         * unique: table_partition0_column_a_idx
     
     If you would like to convert this constraint into an error, please
     call unique_constraint/3 in your changeset and define the proper
     constraint name. The changeset defined the following constraints:
     
         * unique: table_column_a_idx
```

Adding a `match:` option to the changeset `unique_constraint` with supported values of `:exact` and `:suffix` allow us to write a constraint which look like:

`|> unique_constraint(:column_a, name: :column_a_idx, match: :suffix)`

And for any access of `table` that causes a constraint violation will return:

`errors: [column_a: {"has already been taken", []}]`

Even when the the underlying repo returns an error matching: 
```
column_a_idx
table_partition0_column_a_idx
table_partition1_column_a_idx
...
```

Not Implemented
=====
1. `match:` option for changeset constraints other than `unique_constraint`. 
2. Match types other than `:exact` and `:suffix`

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1852,7 +1852,7 @@ defmodule Ecto.Changeset do
     constraint = opts[:name] || "#{get_source(changeset)}_#{field}_exclusion"
     message    = message(opts, "violates an exclusion constraint")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :exclude, to_string(constraint), :exact, field, {message, []})
+    add_constraint(changeset, :exclude, to_string(constraint), match_type, field, {message, []})
   end
 
   defp no_assoc_message(:one), do: "is still associated to this entry"
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -158,7 +158,7 @@ defmodule Ecto.Changeset do
 
   @type error :: {String.t, Keyword.t}
   @type action :: nil | :insert | :update | :delete | :replace
-  @type constraint :: %{type: :unique, constraint: String.t,
+  @type constraint :: %{type: :unique, constraint: String.t, match: atom,
                         field: atom, message: error}
   @type data :: map()
   @type types :: Keyword.t | map()
@@ -1624,6 +1624,13 @@ defmodule Ecto.Changeset do
     * `:name` - the constraint name. By default, the constraint
       name is inflected from the table + field. May be required
       explicitly for complex cases
+    * `:match` - how the changeset constraint matches against the
+      repo constraint. The default is :exact, which requires an
+      exact match. Also available is :suffix, which causes any repo constraint
+      violation which ends_with this constraint name to be mapped to this
+      constraint. This allows for constraint consolidation, and supports
+      cases wherein the repo generates multiple backend constraints
+      for a single given input constraint. (e.g. table inheritance)
 
   ## Complex constraints
 
@@ -1669,7 +1676,8 @@ defmodule Ecto.Changeset do
   def unique_constraint(changeset, field, opts \\ []) do
     constraint = opts[:name] || "#{get_source(changeset)}_#{field}_index"
     message    = message(opts, "has already been taken")
-    add_constraint(changeset, :unique, to_string(constraint), field, {message, []})
+    match_type = if opts[:match] == :suffix, do: :suffix, else: :exact
+    add_constraint(changeset, :unique, to_string(constraint), match_type, field, {message, []})
   end
 
   @doc """
@@ -1842,9 +1850,9 @@ defmodule Ecto.Changeset do
   defp no_assoc_message(:one), do: "is still associated to this entry"
   defp no_assoc_message(:many), do: "are still associated to this entry"
 
-  defp add_constraint(changeset, type, constraint, field, error)
-       when is_binary(constraint) and is_atom(field) and is_tuple(error) do
-    update_in changeset.constraints, &[%{type: type, constraint: constraint,
+  defp add_constraint(changeset, type, constraint, match \\ :exact, field, error)
+       when is_binary(constraint) and is_atom(field) and is_tuple(error) and is_atom(match)  do
+    update_in changeset.constraints, &[%{type: type, constraint: constraint, match: match,
                                          field: field, error: error}|&1]
   end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -172,6 +172,7 @@ defmodule Ecto.Changeset do
   }
 
   @relations [:embed, :assoc]
+  @match_types [:exact, :suffix]
 
   @doc """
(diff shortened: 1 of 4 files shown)


test/ecto/repo_test.exs
```diff
@@ -307,4 +307,46 @@ defmodule Ecto.RepoTest do
...
+  end
+
+  test "that an unknown constraint violation falls through and raises an exception" do
+    my_schema = %MySchema{id: 1}
+    changeset =
+      put_in(my_schema.__meta__.context, {:invalid, [invalid_constraint_type: "my_schema_foo_index"]})
+      |> Ecto.Changeset.change(x: "foo")
+      |> Ecto.Changeset.unique_constraint(:foo)
+    assert_raise Ecto.ConstraintError, fn ->
+     TestRepo.update(changeset)
```
josevalim: Indentation wrong.

test/ecto/repo_test.exs
```diff
@@ -307,4 +307,46 @@ defmodule Ecto.RepoTest do
...
+  end
+
+  test "insert fails to maps repo constraint violation to changeset constraint with non-matching name" do
+    my_schema = %MySchema{id: 1}
+    changeset =
+      put_in(my_schema.__meta__.context, {:invalid, [unique: "foo_table_custom_foo_index"]})
+      |> Ecto.Changeset.change(x: "foo")
+      |> Ecto.Changeset.unique_constraint(:foo, name: "custom_foo_index")
+    assert_raise Ecto.ConstraintError, fn ->
+     TestRepo.update(changeset)
```
josevalim: Indentation wrong.

lib/ecto/changeset.ex
```diff
@@ -1669,7 +1676,8 @@ defmodule Ecto.Changeset do
   def unique_constraint(changeset, field, opts \\ []) do
     constraint = opts[:name] || "#{get_source(changeset)}_#{field}_index"
     message    = message(opts, "has already been taken")
-    add_constraint(changeset, :unique, to_string(constraint), field, {message, []})
+    match_type = if opts[:match] == :suffix, do: :suffix, else: :exact
```
josevalim: `match_type = Keyword.get(opts, :match, :exact)` ?

lib/ecto/changeset.ex
```diff
@@ -1842,9 +1850,9 @@ defmodule Ecto.Changeset do
   defp no_assoc_message(:one), do: "is still associated to this entry"
   defp no_assoc_message(:many), do: "are still associated to this entry"
 
-  defp add_constraint(changeset, type, constraint, field, error)
-       when is_binary(constraint) and is_atom(field) and is_tuple(error) do
-    update_in changeset.constraints, &[%{type: type, constraint: constraint,
+  defp add_constraint(changeset, type, constraint, match \\ :exact, field, error)
```
josevalim: It would be best if we didn't use the default argument so we are always forced to explicitly pass the option.

josevalim:
@vietord looks great to me! I have added some (mostly stylistic) comments. :)

vietord:
Thanks! Updated with fixes, reworded the docs, and added match_type to the check and exclusion constraints makes sense for those two also. 

One additional change while shifting to Keyword.get():  
I added constraint match type validation to add_constraint. I like it better than the original behavior of squashing to the default because that would _really_ hide typos, and nobody needs to spend an hour trying to figure out why their `match: :sufifix` doesn't work.

josevalim:
Sounds great!

lib/ecto/changeset.ex
```diff
@@ -158,7 +158,7 @@ defmodule Ecto.Changeset do
 
   @type error :: {String.t, Keyword.t}
   @type action :: nil | :insert | :update | :delete | :replace
-  @type constraint :: %{type: :unique, constraint: String.t,
+  @type constraint :: %{type: :unique, constraint: String.t, match: atom,
```
josevalim: Instead of atom, maybe we should use :exact | :suffix then?

test/ecto/repo_test.exs
```diff
@@ -307,4 +307,46 @@ defmodule Ecto.RepoTest do
...
+    changeset =
+      put_in(my_schema.__meta__.context, {:invalid, [unique: "foo_table_custom_foo_index"]})
+      |> Ecto.Changeset.change(x: "foo")
+      |> Ecto.Changeset.unique_constraint(:foo, name: "custom_foo_index")
+    assert_raise Ecto.ConstraintError, fn ->
+      TestRepo.update(changeset)
+    end
+  end
+
+  test "that an unknown constraint violation falls through and raises an exception" do
```
josevalim: Let's remove "that an".

lib/ecto/changeset.ex
```diff
@@ -1831,20 +1842,27 @@ defmodule Ecto.Changeset do
...
   defp no_assoc_message(:one), do: "is still associated to this entry"
   defp no_assoc_message(:many), do: "are still associated to this entry"
 
-  defp add_constraint(changeset, type, constraint, field, error)
-       when is_binary(constraint) and is_atom(field) and is_tuple(error) do
-    update_in changeset.constraints, &[%{type: type, constraint: constraint,
+
+  defp add_constraint(changeset, type, constraint, match, field, error)
+       when is_binary(constraint) and is_atom(field) and is_tuple(error) and is_atom(match)  do
+    unless match in @match_types, do: raise(ArgumentError, "Invalid match type: #{inspect match}. Allowed match types: #{inspect @match_types}")
```
josevalim: error messages should start with lower case.

josevalim:
Thank you! This looks good to me. Shall we ship it? :D

vietord:
Great! Want me to squash commits or anything? 

As far as I can tell the travis-ci failure is not related to code changes, but I can't request a re-run on it.

josevalim:

PR #1593, opened 2016-07-20
Allow to alter primary keys on tables

xcambar:
According to the documentation, `add/3` (https://hexdocs.pm/ecto/Ecto.Migration.html#add/3) works for both `CREATE` and`ALTER`.

This PR brings the code that was missing for `ALTER`.

PR #1594, opened 2016-07-20
Fixes SQL syntax on alter with primary key

xcambar:
Follow up on #1593, the SQL syntax generated by the PR is invalid (missing comma).
This could not be caught by the test suite, because the expected value was incorrect as well...

This has been tested on a real-world project, but you're more than encouraged to test further.

Should an integration test be added to avoid future regressions?

Sorry for the inconvenience...

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -287,6 +287,19 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule AlterPrimaryKeyMigration do
+    use Ecto.Migration
+
+    def change do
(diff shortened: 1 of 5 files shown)


josevalim:
Yes, let's try to add a test to the integration suite we already have on integration_test. There are a couple migration tests there as well. :)

xcambar:
✔️  add an Integration test

josevalim:
The build failed for MySQL, can you please take a look?

xcambar:
Build fixed, the issue with MySQL has been described here: https://github.com/elixir-ecto/ecto/issues/1595. Quite an edge case :D

josevalim:

PR #1596, opened 2016-07-21
Fixes MySQL syntax for the table structure in CREATE

### LINKED ISSUE #1595: Creating an empty table in mysql
### Current behavior

While working on #1594, I noticed that creating a table without any column (unusual, isn't it? 😉 ),
the generated SQL query for MySQL is invalid:

```
  defmodule DuntMigration do
    use Ecto.Migration

    def change, do
      create table(:dummy)
    end
  end

=> CREATE TABLE "dummy" () ENGINE = INNODB;
```

### Expected behavior

It should generate `CREATE TABLE "dummy" ENGINE = INNODB;`.

I'll gladly do it, feel free to assign me.

xcambar:
fixes #1595

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -487,18 +487,26 @@ if Code.ensure_loaded?(Mariaex) do
 
     alias Ecto.Migration.{Table, Index, Reference, Constraint}
 
+    defp wrap_in_parentheses(""), do: ""
+    defp wrap_in_parentheses(str), do: "(#{str})"
+
     def execute_ddl({command, %Table{} = table, columns}) when command in [:create, :create_if_not_exists] do
       engine  = engine_expr(table.engine)
       options = options_expr(table.options)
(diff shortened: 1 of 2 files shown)


lib/ecto/adapters/mysql/connection.ex
```diff
@@ -725,10 +732,13 @@ if Code.ensure_loaded?(Mariaex) do
       <<?`, name::binary, ?`>>
     end
 
-    defp assemble(list) do
+    defp assemble(list), do: assemble(list, " ")
+    defp assemble(list, joiner) do
       list
       |> List.flatten
-      |> Enum.join(" ")
+      |> Enum.reject(&is_nil(&1))
```
josevalim: Can we drop the is_nil check? I think removing both nil and empty strings is a smell, we should only remove the empty string ones.

josevalim:
@xcambar excellent, thank you! I have added one tiny comment.

xcambar:
I removed the is_nil check by making sure that all the functions implied in CREATE statements return strings.

josevalim:

Issue #1579, opened 2016-07-15
Add source metadata to log entries


michalmuskala:
When looking at the logs, usually the part you're most interested in is the table, against which you're running the query. Unfortunately, today it's relatively hard to get it, since it's hidden in the `FROM` clause behind all the explicit selects.

I propose to add an additional `source` field to `Ecto.LogEntry` that would be filled on a best-effort basis (it's hard to extract it for subqueries, for example). If present the field would be displayed on the first line of the log, for example:
```
[debug] QUERY OK source=events db=3.2ms queue=0.1ms
SELECT e0."id", e0."name", e0."description", e0."inserted_at", e0."updated_at" FROM "events" AS e0 []
```

Troush:
I tried to handle this.
But got stuck with question:
How to get proper table name?

Regex query statement is a bad idea, I think. Or not?

josevalim:
@Troush when building the log entry struct, we typically have the query or the source already separated, so we should pass such information as well.

Troush:
@josevalim I was inspecting https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/sql.ex#L512
And I can't find any `source` or `table` info there.  Just plain `statement` and other info.

michalmuskala:
Closed by https://github.com/elixir-ecto/ecto/commit/0ea701ca6df04026a84c8a4f9a91b42bd2b97196

PR #1597, opened 2016-07-22
Add source metadata to log entries

Troush:
Solves #1579

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -68,6 +68,13 @@ defmodule Ecto.Adapters.SQL do
 
       @doc false
       def execute(repo, meta, query, params, process, opts) do
+        table =
+          case Map.fetch(meta, :sources) do
+            {:ok, {{table, module}}} ->
+              table
+            _ -> nil
+          end
+        opts = Keyword.put(opts, :source, table)
         Ecto.Adapters.SQL.execute(repo, meta, query, params, process, opts)
       end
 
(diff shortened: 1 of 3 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -124,6 +126,24 @@ defmodule Ecto.Adapters.SQL do
         :ok
       end
 
+      def do_get_sources(meta) do
```
Troush: This is a bit messy, but I don't figure how to implement it more readable.

lib/ecto/adapters/sql.ex
```diff
@@ -504,17 +524,22 @@ defmodule Ecto.Adapters.SQL do
...
       {false, opts} -> opts
     end
   end
 
-  defp log(repo, params, entry) do
+  defp log(repo, params, entry, opts) do
     %{connection_time: query_time, decode_time: decode_time,
       pool_time: queue_time, result: result, query: query} = entry
+    source =
+      case Keyword.fetch(opts, :source) do
```
josevalim: `Keyword.get(opts, :source)`

lib/ecto/adapters/sql.ex
josevalim: We don't need to get all sources, only the first one (which is the one used in `from`).

lib/ecto/log_entry.ex
```diff
@@ -79,6 +80,9 @@ defmodule Ecto.LogEntry do
   defp ok_error({:ok, _}),    do: "OK"
   defp ok_error({:error, _}), do: "ERROR"
 
+  defp ok_source(nil),    do: ""
+  defp ok_source(source), do: " source: #{inspect(source)}"
```
josevalim: We should probably use "source=" or just inspect the source itself.

lib/ecto/adapters/sql.ex
Troush: If we can provide such feature, why not?
This not so much LOC, but ability to show all targeted tables is a big plus.

josevalim: Because we will be logging too much information that will get in the way of reading the rest. I am already on the fence about this feature because the source exists in the query, so this is just a convenience. If we are going to log all sources, then it is best to just read the query.

michalmuskala:
This would require each adapter to extract this data themselves. Would it be possible to do that at some other level?

What if instead of constructing the `Ecto.LogEntry` directly in the adapter we provide a function accepting the `query` itself and a list of fields, something like:
```elixir
defmodule Ecto.LogEntry do
  # ...

  def new(query, fields) do
    struct(%Ecto.LogEntry{source: get_source(query)}, fields})
  end
end
```

This would also allow us to easily extend the log entry with some metadata in the future, should we need this.

josevalim:
> This would require each adapter to extract this data themselves. Would it be possible to do that at some other level?

I don't think there is anything wrong with that. Specially because the source does not always come from Ecto.Query.

Troush:
Ready, with only "from" source

michalmuskala:
@josevalim that's a fair point. So maybe instead of query, we should be passing the `meta` part we pass into every adapter call?

Troush:
@josevalim ping :purple_heart:

josevalim:
Looks good to me! I will merge it later when focusing on Ecto again. :D

Troush:
Okay, thnx! Just now I know that this is okay and I can do something else.

Quick feedback on this, from my project running this branch: this is even more helpfult that I was imagine in the beginning :heart:

josevalim:
Mergeed manually!

PR #1598, opened 2016-07-22
Exposes the migration documentation differently

xcambar:
This PR offers to reshape a bit the documentation about the index/table/reference/constraint:

* It separates in distinct files the different modules instead of keeping everything in `lib/ecto/migration.ex`
* The complete doc is no longer in the doc of the function called in the module (as `Ecto.Migration.references/3`) but it has been moved to the module itself, this IMHO allow for better redabiility

⚠️ This PR only applies changes to the "references" method of the migration. The goal here is to kickstart the discussion.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -487,18 +487,26 @@ if Code.ensure_loaded?(Mariaex) do
 
     alias Ecto.Migration.{Table, Index, Reference, Constraint}
 
+    defp wrap_in_parentheses(""), do: ""
+    defp wrap_in_parentheses(str), do: "(#{str})"
+
     def execute_ddl({command, %Table{} = table, columns}) when command in [:create, :create_if_not_exists] do
       engine  = engine_expr(table.engine)
       options = options_expr(table.options)
-      if_not_exists = if command == :create_if_not_exists, do: " IF NOT EXISTS", else: ""
-      pk_definition = case pk_definition(columns) do
-        nil -> ""
-        pk -> ", #{pk}"
-      end
+      if_not_exists = if command == :create_if_not_exists, do: "IF NOT EXISTS", else: ""
 
-      "CREATE TABLE" <> if_not_exists <>
-        " #{quote_table(table.prefix, table.name)}" <>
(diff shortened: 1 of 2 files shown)


josevalim:
I am honestly not a fan because:

1. There is no need to break into multiple files
2. Most of the people are worried about the migration function and not the structure returned by the function. In fact, most of the times you don't need to care about the structure at all.

josevalim:
In any case, thank you for sending the PR and starting the conversation! :D

xcambar:
I confess I went too far by moving the structs into different files.

Allow me a bit of a backstory here, maybe it will spawn new ideas:
My PKs are all UUIDs. When building references, I was ignor ant of the fact that `references` defaults to ints and not the type of the foreign PK (for good reasons), and I had a hard time figuring out the correct params for `references`.
My first idea was the doc. "Yay, a module is called Reference, that must be there". Bummer, no doc... Only a type that I did/could not read (my bad).
It took me a very long time to figure out (because I'm still new to Elixir) that the doc of `references` was actually in `Ecto.Migration`.

Hence, the PR.

Maybe it's just me, but from my newcomer's perspective, in the end, I had the feeling that the experience could have been enhanced when the needs go a bit "beyond" the default values.

I am looking forward to your thoughts, and please feel free to close if you think this use case is not relevant for further discussion.

josevalim:
@xcambar that helps a lot, thanks for sharing.

So what if we add a note to the moduledocs saying: "Only adapters interact with this data-structure directly. To create a reference, please call Ecto.Migration.references/2" or something of sorts?

xcambar:
👍
I'll update the PR as time allows.

lib/ecto/migration/reference.ex
```diff
@@ -0,0 +1,30 @@
...
+         May be `:nothing`, `:update_all` or `:nilify_all`.
+         Defaults to `:nothing`.
+
+  """
+
+  defstruct name: nil, table: nil, column: :id, type: :serial, on_delete: :nothing, on_update: :nothing
+  @type t :: %__MODULE__{table: atom, column: atom, type: atom, on_delete: atom, on_update: atom}
+end
+
+
```
lexmag: Unnecessary newline. ✂️

PR #1624, opened 2016-08-07
Poison Encoder for Decimal should return a JSON Number, not String

IbrahimZananiri:
Poison `Encoder` protocol implementation currently encodes Decimal to double-quoted JSON `String`.
That causes unexpected behavior in object-mapping on clients due to the confused JSON numeric type.
This PR changes the `Encoder` implementation to encode Decimal to JSON `Number` type.

**Before:**
```
iex(5)> Poison.encode!(%{"number": Decimal.new("2.5")})  
"{\"number\":\"2.5\"}"
```
**After this change:**
```
iex(5)> Poison.encode!(%{"number": Decimal.new("2.5")})  
"{\"number\":2.5}"
```

### DIFF
--- a/lib/ecto/poison.ex
+++ b/lib/ecto/poison.ex
@@ -1,6 +1,6 @@
 if Code.ensure_loaded?(Poison) do
   defimpl Poison.Encoder, for: Decimal do
-    def encode(decimal, _opts), do: <<?", Decimal.to_string(decimal)::binary, ?">>
+    def encode(decimal, _opts), do: Decimal.to_string(decimal)
   end
 
   defimpl Poison.Encoder, for: [Ecto.Date, Ecto.Time, Ecto.DateTime] do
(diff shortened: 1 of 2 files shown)


michalmuskala:
This behaviour is intentional. JSON number is encoded as a floating point number which is, by definition, inexact. Decimal type is designed to represent exact numbers - encoding with a float would loose all that precision. They only way to represent an exact decimal number in JSON is with a string.

IbrahimZananiri:
@michalmuskala I see your point.
I was wondering why precision was not the responsibility of the deserializing end. It is because most deserialization libraries parse a non-integer JSON `Number` as float.

I checked JSON marshaling implementations by other Decimal libraries in Go, Rails, they encode Decimal as string too.

If anyone else is wondering, I found this comment satisfying, from the the Rails source for Decimal encoder:
>A BigDecimal would be naturally represented as a JSON number. Most libraries, however, parse non-integer JSON numbers directly as floats. Clients using those libraries would get in general a wrong number and no way to recover other than manually inspecting the string with the JSON code itself.
>That's why a JSON string is returned. The JSON literal is not numeric, but if the other end knows by contract that the data is supposed to be a BigDecimal, it still has the chance to post-process the string and get the real value.

https://github.com/rails/rails/blob/v5.0.0/activesupport/lib/active_support/core_ext/object/json.rb#L107

Thank you!

PR #1625, opened 2016-08-10
Allow single column indexes to not use a list

mgwidmann:
The following code crashes when running a migration:

```elixir
create unique_index(:users, :email)
```

When Ecto is expecting instead

```elixir
create unique_index(:users, [:email])
```

Since the error message is a no such function error, it's a bit confusing what the user has done wrong.

Instead, this PR just wraps it in a list so everyone is happy.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -498,19 +498,23 @@ defmodule Ecto.Migration do
       create index(:products, [:user_id], where: "price = 0", name: :free_products_index)
 
   """
-  def index(table, columns, opts \\ []) when is_atom(table) and is_list(columns) do
+  def index(table, columns, opts \\ [])
+  def index(table, columns, opts) when is_atom(table) and is_list(columns) do
     index = struct(%Index{table: table, columns: columns}, opts)
     %{index | name: index.name || default_index_name(index)}
   end
+  def index(table, column, opts) when is_atom(table) and is_atom(column), do: index(table, [column], opts)
 
   @doc """
   Shortcut for creating a unique index.
 
   See `index/3` for more information.
   """
-  def unique_index(table, columns, opts \\ []) when is_atom(table) and is_list(columns) do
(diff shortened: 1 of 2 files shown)


lib/ecto/migration.ex
```diff
@@ -498,19 +498,23 @@ defmodule Ecto.Migration do
       create index(:products, [:user_id], where: "price = 0", name: :free_products_index)
 
   """
-  def index(table, columns, opts \\ []) when is_atom(table) and is_list(columns) do
+  def index(table, columns, opts \\ [])
+  def index(table, columns, opts) when is_atom(table) and is_list(columns) do
     index = struct(%Index{table: table, columns: columns}, opts)
     %{index | name: index.name || default_index_name(index)}
   end
+  def index(table, column, opts), do: index(table, [column], opts)
```
josevalim: We should add `is_atom(table)`, otherwise giving it a non-atom will make it an infinite loop.

lib/ecto/migration.ex
```diff
@@ -498,19 +498,23 @@ defmodule Ecto.Migration do
...
   Shortcut for creating a unique index.
 
   See `index/3` for more information.
   """
-  def unique_index(table, columns, opts \\ []) when is_atom(table) and is_list(columns) do
+  def unique_index(table, columns, opts \\ [])
+  def unique_index(table, columns, opts) when is_atom(table) and is_list(columns) do
     index(table, columns, [unique: true] ++ opts)
   end
+  def unique_index(table, column, opts), do: unique_index(table, [column], opts)
```
josevalim: We should add `is_atom(table) and is_atom(column)`, otherwise giving it a non-atom will make it an infinite loop.

josevalim:
@mgwidmann I have added some comments, thank you!

mgwidmann:
Pushed and rebased... 👍

josevalim:

PR #1626, opened 2016-08-10
Add field name to schema load error

andrewtimberlake:
Before the change, when invalid data exists in a schema, the error doesn’t indicate which field contains the problem.

```
** (ArgumentError) cannot load `"[\"postal\", \"home\", \"work\"]"` as type {:array, :string} in schema MyModule.MyModel
```

after the change, the error message includes the field name (which is very useful when there are a lot of fields of a similar type in a schema)

```
** (ArgumentError) cannot load `"[\"postal\", \"home\", \"work\"]"` as type {:array, :string} for :my_array_field in schema MyModule.MyModel
```

**Note**  There are no tests for this change because I was not sure how to test this.


josevalim:

PR #1634, opened 2016-08-18
Update sandbox.ex

DamirSvrtan:


lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -221,16 +221,16 @@ defmodule Ecto.Adapters.SQL.Sandbox do
           ** (DBConnection.ConnectionError) owner #PID<> timed out
           because it owned the connection for longer than 15000ms
 
-  If you have a long running test, the timeout for the connection ownership may
+  If you have a long running test (or you're debugging with IEx.pry), the timeout for the connection ownership may
   be too short.  You can increase the timeout by setting the
-  `:ownership_timeout` options for your repo config:
+  `:ownership_timeout` options for your repo config (preferably in `config/test.exs`):
```
michalmuskala: The `config/test.exs` is a phoenix convention that is not enforced by elixir or ecto. It might be confusing here.

DamirSvrtan: Hey @michalmuskala, thnx for the comment, forgot about that!

I can remove it, or just add a note that it's how it is in Phoenix.

However, that should also be said on line 26:

```
  The first step is to configure your database to use the
  `Ecto.Adapters.SQL.Sandbox` pool. You set those options in your
  `config/config.exs` (or preferrably `config/test.exs`) if you
  haven't yet:
```

michalmuskala: Ah, I was not aware we're already using it. So I guess it's fine. I would only change it to mention `config.exs` first, and later the `test.exs`, just like the fragment you're citing.

DamirSvrtan:
@michalmuskala did it!

michalmuskala:
Thank you! ❤️

PR #1635, opened 2016-08-18
Allow column specification for MySQL index

immon:
Now it is allowed to `create index(:table, ["token(16)"])` like for
postgres adapter.

Additionally fixed two test cases where column specification should not be
quoted.
btw, MySQL does not have functional indexes

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -647,6 +647,8 @@ if Code.ensure_loaded?(Mariaex) do
     defp default_expr(:error),
       do: []
 
+    defp index_expr(literal) when is_binary(literal),
+      do: literal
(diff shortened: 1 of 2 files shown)


immon:
I'm not sure why tests are failing for ELIXIR_ERL_OPTIONS=""

josevalim:
Hi @immon! I think we should continue escaping columns names. What if we support fragments, like we do in the `:default` option, to provide functional indexes? So it would look like this: `create index(:table, [fragment("token(16)")])`. The patch would be very similar, except it would match on `{:fragment, _}` key in the adapters and return the second argument.

immon:
I don't feel strong enough to write something more complicated. I just copy&pasted the code we have in postgres adapter (and fixed failing tests). Sorry.

josevalim:

PR #1636, opened 2016-08-18
Fix some typos in Ecto.Changeset docs

DNNX:
There are a few misspellings and typos here and there.

Initially I stumbled upon `a exclusion` when reading the docs, which is an obvious mistake. I decided to run the file through a spell checker and found a few more suggestions to improve the docs. Hence, this PR was born.

By no means I'm an expert in English language (not even a native speaker), so it would be good if someone who knows better than me had a look and double-checked the diff. Don't blame the spell checker if you find incorrect "fixes", because some of them are introduced by me, a human.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -75,12 +75,12 @@ defmodule Ecto.Changeset do
   ## Empty values
 
   Many times, the data given on cast needs to be further pruned, specially
-  regarding to empty values. For example, if you are gathering data to be
-  cast from the command line or through an HTML form or any other text
-  based format, it is likely those means cannot express nil values. For
+  regarding empty values. For example, if you are gathering data to be
+  cast from the command line or through an HTML form or any other text-based
+  format, it is likely those means cannot express nil values. For
   those reasons, changesets include the concept of empty values, which are
   values that will be automatically converted to nil on `cast/3`. Those
-  values are stored in the changeset `empty_values` field and defaults to
+  values are stored in the changeset `empty_values` field and default to
   `[""]`.
 
   ## The Ecto.Changeset struct
@@ -684,7 +684,7 @@ defmodule Ecto.Changeset do
   then falls back on the data, finally returning `:error` if
   no value is available.
 
-  For relations, this functions will return the changeset
+  For relations, these functions will return the changeset
   original data with changes applied. To retrieve raw changesets,
   please use `fetch_change/2`.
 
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -74,13 +74,13 @@ defmodule Ecto.Changeset do
 
   ## Empty values
 
-  Many times, the data given on cast needs to be further pruned, specially
-  regarding to empty values. For example, if you are gathering data to be
-  cast from the command line or through an HTML form or any other text
-  based format, it is likely those means cannot express nil values. For
+  Many times, the data given on cast needs to be further pruned, especially
```
josevalim: It should be `specially`.

lib/ecto/changeset.ex
```diff
@@ -721,7 +721,7 @@ defmodule Ecto.Changeset do
   then falls back on the data, finally returning `default` if
   no value is available.
 
-  For relations this functions will return the changeset data
+  For relations, this functions will return the changeset data
```
josevalim: For relations, `these functions`

lib/ecto/changeset.ex
```diff
@@ -1486,7 +1486,7 @@ defmodule Ecto.Changeset do
   Optimistic locking works by keeping a "version" counter for each record; this
   counter gets incremented each time a modification is made to a record. Hence,
   in order to use optimistic locking, a field must exist in your schema for
-  versioning purpose. Such field is usually an integer but other types are
+  the versioning purpose. Such field is usually an integer but other types are
```
josevalim: No `the` here.

lib/ecto/changeset.ex
```diff
@@ -1605,7 +1605,7 @@ defmodule Ecto.Changeset do
       Defaults to "is invalid"
     * `:name` - the name of the constraint. Required.
     * `:match` - how the changeset constraint name it matched against the
-      repo constraint, may be `:exact` or `:suffix`. Default is `:exact`.
+      repo constraint, may be `:exact` or `:suffix`. The default is `:exact`.
```
josevalim: Let's use `Defaults to`.

lib/ecto/changeset.ex
```diff
@@ -1647,10 +1647,10 @@ defmodule Ecto.Changeset do
     * `:message` - the message in case the constraint check fails,
       defaults to "has already been taken"
     * `:name` - the constraint name. By default, the constraint
-      name is inflected from the table + field. May be required
+      name is inferred from the table + field. May be required
       explicitly for complex cases
     * `:match` - how the changeset constraint name it matched against the
-      repo constraint, may be `:exact` or `:suffix`. Default is `:exact`.
+      repo constraint, may be `:exact` or `:suffix`. The default is `:exact`.
```
josevalim: `Defaults`

josevalim:
Awesome, I have added a couple comments and we should be good to go!

DNNX:
@josevalim please take a look again.

josevalim:

PR #1637, opened 2016-08-19
[wip] Fix #1610 -- Add option to scope schema relations

### LINKED ISSUE #1610: Association-induced compile time dependencies
### Environment

* Ecto version (mix deps): 1.1.5 & 2.0.3
* Operating system: debian jessie

### Current behavior

I have a project that contains lots of ecto models (over a 100) with lots of relationships between them. I recently noticed that changing one model would trigger the compilation of every other model.

By using `mix xref graph` I noticed that a given module would have a compile time dependency for every other related model. For example, if I have something like:

```elixir
## schema_a.ex
defmodule SchemaA do
  use Ecto.Schema
  schema "tableA" do
    belongs_to :b, SchemaB
  end
end

## schema_b.ex
defmodule SchemaB do
  use Ecto.Schema
  schema "tableB" do
    field :lol, :string
    has_many :as, SchemaA
    has_one :C, SchemaC
  end
end

## schema_c.ex
defmodule SchemaC do
  use Ecto.Schema
  schema "tableC" do
    field :lol, :string
    has_many :as, SchemaA
    has_one :C, SchemaC
  end
end
```

The dependency graph will look like this: (which means that changing any of those files will require recompilation of every other)

screenshot from 2016-08-03 01-11-41

@josevalim was kind enough to help me identify the problem & suggest a workaround using `Module.concat/1`:

```elixir
defmodule SchemaC do
  use Ecto.Schema
  schema "tableC" do
    field :lol, :string
    has_many :as, Module.concat(["SchemaA"])
    has_one :C, Module.concat(["SchemaC"])
  end
end
```

(I recall him mentioning that phoenix's router uses a similar trick with the `scope` clause)

After I applied the workaround, the number of recompiled files after a change to a particular model changed from 226 to 6. (which drastically reduced the recompilation times & made everyone happy in the office).

Does it make sense for Ecto to provide facilities to avoid such problem? If not, where would be a good place to document this behavior?

rranelli:
As discussed in #1610, this PR adds the option to scope Ecto relations so that no compile-time dependencies are created. 

I still have some doubts about the implementation/tests, so I wanted to make sure I'm on the right track before heading out to write the docs. I have added inline comments to the points I have doubts.

I couldn't find a good way to decide whether or not to scope custom types -- like Ecto.Date. I struggled for some time with this and wasn't able to get to a solution that looked good.

Fixes #1610

### DIFF
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -26,7 +26,7 @@ defmodule Ecto.Integration.Post do
   use Ecto.Integration.Schema
   import Ecto.Changeset
 
-  schema "posts" do
+  schema Ecto.Integration, "posts" do
     field :counter, :id # Same as integer
     field :title, :string
     field :text, :binary
@@ -40,21 +40,21 @@ defmodule Ecto.Integration.Post do
     field :meta, :map
     field :links, {:map, :string}
     field :posted, Ecto.Date
-    has_many :comments, Ecto.Integration.Comment, on_delete: :delete_all, on_replace: :delete
-    has_one :permalink, Ecto.Integration.Permalink, on_delete: :delete_all, on_replace: :delete
+    has_many :comments, Comment, on_delete: :delete_all, on_replace: :delete
+    has_one :permalink, Permalink, on_delete: :delete_all, on_replace: :delete
     has_many :comments_authors, through: [:comments, :author]
-    belongs_to :author, Ecto.Integration.User
-    many_to_many :users, Ecto.Integration.User,
+    belongs_to :author, User
+    many_to_many :users, User,
       join_through: "posts_users", on_delete: :delete_all, on_replace: :delete
-    many_to_many :customs, Ecto.Integration.Custom,
+    many_to_many :customs, Custom,
       join_through: "posts_customs", join_keys: [post_id: :uuid, custom_id: :bid],
       on_delete: :delete_all, on_replace: :delete
-    many_to_many :unique_users, Ecto.Integration.User,
-      join_through: Ecto.Integration.PostUserCompositePk
+    many_to_many :unique_users, User,
+      join_through: PostUserCompositePk
     has_many :users_comments, through: [:users, :comments]
     has_many :comments_authors_permalinks, through: [:comments_authors, :permalink]
     timestamps()
-    has_one :post_user_composite_pk, Ecto.Integration.PostUserCompositePk
+    has_one :post_user_composite_pk, PostUserCompositePk
   end
 
   def changeset(schema, params) do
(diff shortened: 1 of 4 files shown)


integration_test/support/schemas.exs
```diff
@@ -26,7 +26,7 @@ defmodule Ecto.Integration.Post do
   use Ecto.Integration.Schema
   import Ecto.Changeset
 
-  schema "posts" do
+  schema Ecto.Integration, "posts" do
```
rranelli: I'm in doubt on whether I should replicate the assocs integration test (integration_test/cases/assoc.exs) & the associated schemas (integration_test/support/schemas.exs) to use a scoped schema in the tests.

I scoped the schemas here because it was easier, but I would like to hear other opinions on this.

lib/ecto/schema.ex
```diff
@@ -319,6 +320,16 @@ defmodule Ecto.Schema do
   end
 
   @doc """
+  TODO
+  """
+  defmacro schema(scope, source, [do: block]) do
```
rranelli: I also thought about exposing the `@association_scope` attribute directly instead of defining `schema/3`.

I personally think that this would be better since I'm not scoping custom field types.

lib/ecto/schema.ex
```diff
@@ -596,7 +607,8 @@ defmodule Ecto.Schema do
   """
   defmacro has_many(name, queryable, opts \\ []) do
     quote do
-      Ecto.Schema.__has_many__(__MODULE__, unquote(name), unquote(queryable), unquote(opts))
+      queryable = Ecto.Schema.__build_assoc_module__(@association_scope, unquote(queryable))
```
josevalim: We should do as little work inside of quote as possible. So all of those functions do not actually need to use `@association_scope`. Instead, inside `__has_many__` and so on, you can run:

    Module.get_attribute(module, :association_scope) |> Ecto.Schema(queryable)

I would change all functions below to do the work inside `__has_many__`, `__belongs_to__` and so on instead of inside the quote. :)

integration_test/support/schemas.exs
josevalim: This is totally fine!

lib/ecto/schema.ex
josevalim: Great question. I was in doubt about this myself. I would like to talk to @chrismccord and @michalmuskala to be sure but, one of the points to have it not in an attribute, is to mirror the property that aliases usually have of being lexical. If we make it an attribute, people can put it inside `web/web.ex` and that's going to change how the code inside `schema` behaves in a non-obvious way.

lib/ecto/schema.ex
rranelli: That's indeed a good Idea!

I think I've been able to simplify the code quite a bit following it <3

lib/ecto/schema.ex
michalmuskala: I missed this discussion, but yes I'm not a big fan of the module attribute - to be honest I'm not a fan of solving this with the additional scope either - I would love to see some more generic solution in the compiler to prevent situations like this.

josevalim: Do we have any ideas though? :(

PR #1641, opened 2016-08-23
Fix #1640 -- Suboptimal SQL generated by Ecto.assoc

### LINKED ISSUE #1640: Suboptimal SQL generated by Ecto.assoc
### Environment

* Database and version: PostgreSQL 9.5
* Ecto version (mix deps): master/HEAD
* Database adapter and version: postgrex 0.11.2
* Operating system: Debian 8 & OS X 10.11

### Current behavior

I noticed a performance regression between:

    Ecto.assoc(parent, :children) |> order_by(asc: :name) |> Repo.all

and

    Child |> where(parent_id: ^1) |> order_by(asc: :name) |> Repo.all

Depending on the involved data PostgreSQL executes the first query 8-11x slower. The first one generates this SQL:

    SELECT c0."id", c0."name", c0."parent_id" FROM "children" AS c0 \
    WHERE (c0."parent_id" = ANY(array[1])) ORDER BY c0."name";

and the second one:

    SELECT c0."id", c0."name", c0."parent_id" FROM "children" AS c0 \
    WHERE (c0."parent_id" = 1) ORDER BY c0."name";

`EXPLAIN ANALYZE` reveals the problem:

    Sort  (cost=513.60..526.10 rows=5000 width=36) (actual time=42.814..43.376 rows=5000 loops=1)
      Sort Key: name
      Sort Method: quicksort  Memory: 583kB
      ->  Index Scan using children_parent_id_name_index on children c0  (cost=0.29..206.41 rows=5000 width=36) (actual time=0.021..3.271 rows=5000 loops=1)
            Index Cond: (parent_id = ANY ('{1}'::integer[]))
    Planning time: 0.111 ms
    Execution time: 43.816 ms

instead of

    Index Scan using children_parent_id_name_index on children c0  (cost=0.29..367.51 rows=5000 width=36) (actual time=0.022..4.292 rows=5000 loops=1)
      Index Cond: (parent_id = 1)
    Planning time: 0.117 ms
    Execution time: 4.959 ms

It seems like PostgreSQL is not able to use the multi-column index for sorting in case of the ANY-condition and has to fall back to performing a quicksort.

### Expected behavior

In case of a single value `Ecto.assoc/2` should generate SQL with a simple equal-condition. This would be very easy to implement: just by adding this method to `Ecto.Association.Has` and `Ecto.Association.BelongsTo`:

    def assoc_query(%{queryable: queryable, related_key: related_key}, query, [value]) do
      from x in (query || queryable),
        where: field(x, ^related_key) == ^value
    end

Alternatively ectos postgres-adapter could use `IN(1)` instead of `ANY(array[1])`. PostgreSQL is able to use the index in this case.

It seems like the failure of PostgreSQL – but it would be better if the SQL generated by ecto would work performant with the current versions.

yolk:
See #1640 for details

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -388,6 +388,12 @@ defmodule Ecto.Association.Has do
   end
 
   @doc false
+  def assoc_query(%{queryable: queryable, related_key: related_key}, query, [value]) do
+    from x in (query || queryable),
+      where: field(x, ^related_key) == ^value
+  end
+
+  @doc false
   def assoc_query(%{queryable: queryable, related_key: related_key}, query, values) do
     from x in (query || queryable),
       where: field(x, ^related_key) in ^values
@@ -719,6 +725,12 @@ defmodule Ecto.Association.BelongsTo do
     from o in owner,
       join: q in ^queryable,
       on: field(q, ^related_key) == field(o, ^owner_key)
+  end
+
+  @doc false
+  def assoc_query(%{queryable: queryable, related_key: related_key}, query, [value]) do
+    from x in (query || queryable),
+      where: field(x, ^related_key) == ^value
   end
 
   @doc false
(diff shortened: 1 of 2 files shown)


josevalim:

Issue #1627, opened 2016-08-12
Support on_replace: :update for has_one/belongs_to


michalmuskala:
Since those are one-to-one relationships it makes sense to allow editing them without passing the original id, they can be always uniquely identified. Currently ecto will not like that and will require to always pass the association id.

When `on_replace` is set to `:edit`, the associated record will be always edited with new params, regardless if they contain the id or what is the value of the id field.

josevalim:
@michalmuskala :+1:. Shall we call it `:update` though?

michalmuskala:
Yes, upgrade is better.

eniolopes:
Can I work on this one?

josevalim:
Yes, please!

eniolopes:
👍

PR #1643, opened 2016-08-23
Support on_replace: :update for has_one/belongs_to

eniolopes:
Addresses https://github.com/elixir-ecto/ecto/issues/1627

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -312,7 +312,8 @@ defmodule Ecto.Association.Has do
 
   @behaviour Ecto.Association
   @on_delete_opts [:nothing, :nilify_all, :delete_all]
-  @on_replace_opts [:raise, :mark_as_invalid, :delete, :nilify, :update]
+  @on_replace_opts [:raise, :mark_as_invalid, :delete, :nilify]
+  @has_one_on_replace_opts [:update]
   defstruct [:cardinality, :field, :owner, :related, :owner_key, :related_key, :on_cast,
              :queryable, :on_delete, :on_replace, defaults: [], relationship: :child]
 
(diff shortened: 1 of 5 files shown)


eniolopes:
@josevalim as spoken yesterday, I'm no entirely sure where to go now, need some guidance.

lib/ecto/changeset/relation.ex
```diff
@@ -244,7 +248,7 @@ defmodule Ecto.Changeset.Relation do
       single_change(new, current, fun, [:update, :delete], true)
     else
       case on_replace(relation, current) do
-        {:ok, _} -> single_change(new, nil, fun, [:insert], false)
+        {:ok, changeset} -> single_change(new, nil, fun, [changeset.action], false)
```
josevalim: If we are on a update, we need to call something similar to the line 248 above, passing both `new` and `current`, to ensure the new fields will be merged on top of current. Maybe changing [this line](https://github.com/eniolopes/ecto/blob/469f1f90f1647e70c823741c572f10e869ac1faa/lib/ecto/changeset/relation.ex#L247) to `if new_pks.(new) == current_pks.(current) or relation.on_change == :update do` is a better way to move forward?

michalmuskala:
@eniolopes ping

eniolopes:
@michalmuskala I've done the change requested by José, I think we have a working solution for `belongs_to`. Starting the work on `has_one` now.

eniolopes:
@josevalim @michalmuskala I think I'm done with this, sorry for the delay. The tests I added will be enough? Let me know if there is anything else left to do here.

lib/ecto/changeset/relation.ex
```diff
@@ -240,11 +240,11 @@ defmodule Ecto.Changeset.Relation do
   end
 
   defp single_change(relation, new, current_pks, new_pks, fun, current) do
-    if new_pks.(new) == current_pks.(current) do
+    if new_pks.(new) == current_pks.(current) or relation.on_replace == :update do
```
josevalim: Let's pattern match `%{on_replace: on_replace} = relation` on the signature instead. :)

lib/ecto/association.ex
```diff
@@ -343,15 +344,21 @@ defmodule Ecto.Association.Has do
...
-
     unless on_delete in @on_delete_opts do
       raise ArgumentError, "invalid :on_delete option for #{inspect name}. " <>
         "The only valid options are: " <>
         Enum.map_join(@on_delete_opts, ", ", &"`#{inspect &1}`")
     end
 
-    unless on_replace in @on_replace_opts do
+    on_replace = Keyword.get(opts, :on_replace, :raise)
+    cardinality = Keyword.get(opts, :cardinality, :raise)
```
josevalim: `Keyword.fetch!`. I believe the cardinality is always present.

josevalim: We should also change this line: https://github.com/eniolopes/ecto/blob/d844662db4c6cd33769ba4a1b5d4174fd02bed5b/lib/ecto/association.ex#L369

It now can use the `:cardinality` variable.

josevalim:

PR #1653, opened 2016-08-26
Allow {"prefix", "source"} in from

michalmuskala:
This mirrors what we have in insert_all - it was surprising that you
could do:

    insert_all({"prefix", "source"}, [%{x: 1])

But using the same in update_all or delete_all would fail:

    update_all({"prefix", "source"}, set: [x: 0])

\cc @josevalim I'm opening this as a PR, as I'm not 100% convinced about this feature.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -265,6 +265,11 @@ defmodule Ecto.Query do
   prefix, so operations like update and delete will be applied to the
   proper prefix. In case you want to manually set the prefix for new data,
   specially on insert, use `Ecto.put_meta/2`.
+
+  Schemaless queries can use `{"prefix", "source"}` as the query source
+  to directly create a query with the required prefix. This is especially
+  useful when working with `Ecto.Repo.update_all/3` and mirrors the source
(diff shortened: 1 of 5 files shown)


josevalim:
@michalmuskala update_all and delete_all receive a query and the prefix can be set there. The only exception is indeed `insert_all`. Therefore I wouldn't add this. :)

michalmuskala:
On one hand I agree with you, on the other hand setting a prefix on query is significantly more complex than setting the prefix for `insert_all` and `update_all`. The current way to do this for update is:

```elixir
query = %{Ecto.Queryable.to_query("my_table") | prefix: "my_prefix"}
Repo.update_all(query, set: ...)
```
compared to: `Repo.insert_all({"my_prefix", "my_table"}, [...])`

I think aligning both uses is the right path following the rule of least surprise. We should unify the API to support the same thing everywhere.

josevalim:
Sometime ago I have proposed a prefix option to be given to all, insert_all, update_all, insert, etc. That would be my preferred approach instead of adding a special case to update_all/delete_all.

PR #1654, opened 2016-08-26
Feature/add method repo exists

shhavel:
Adds method `Ecto.Repo.exists?/3`
```elixir
MyRepo.exists?(User, email: "testuser@example.com") # => true / false
MyRepo.exists?(Post) # => true / false
```
Please review.
Thanks

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -536,6 +536,7 @@ defmodule Ecto.Integration.RepoTest do
     post1 = TestRepo.insert!(%Post{title: "1", text: "hai"})
     post2 = TestRepo.insert!(%Post{title: "2", text: "hello"})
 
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you but at this moment we would prefer to keep the repository surface API as small as possible.

hayesgm:
Is there a good place where these features should go? I feel like I often re-write a few common functions like this into each Phoenix app (e.g. `MyApp.Repo.exists?/1`, `MyApp.Repo.first/1`, `MyApp.Repo.last/1`, etc).

josevalim:
Ecto.Query.first and Ecto.Query.last are part today of Ecto query API. You can do:

    foo |> where(...) |> first |> Repo.one!
    foo |> where(...) |> last |> Repo.one!

We could introduce exists? but I am not sure it gives us anything?

PR #1659, opened 2016-08-29
Add Join Schema example to many_to_many docs

lukerollans:
What it says on the tin!

I noticed that a "join" schema example was not present in the docs for `many_to_many` associations. A common question amongst newcomers to Ecto is how to manage them. I personally believe this method is the easiest and most efficient

1. Supports timestamps.
2. Distinct changeset function allows easy counter caching, for starters.
3. In the case of a web app, this allows for logical extraction of assoc creation (in this case) to a controller named `UserOrganizationController`, for example, which keeps the app restful and reasonable.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -934,6 +934,7 @@ defmodule Ecto.Schema do
         schema "user_organisation" do
           belongs_to :user, User
           belongs_to :organization, Organization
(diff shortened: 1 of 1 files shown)


josevalim:

romseguy:
It would have be nice to have the advantage (i.e "This gives you the advantage of a unique changeset to update counter caching and whatnot.") of this method listed in the docs, I had to google for it.

Issue #1661, opened 2016-08-30
validate_number for decimal raises: no function clause matching in Decimal.compare/2


kacperix:
### Environment

* Elixir version: 1.3.2
* Database and version: PostgreSQL 9.5.4
* Ecto version: 2.0.4
* Database adapter and version (mix dips): postgrex 0.11.2
* Operating system: OS X (10.11.6)

### Current behavior

One of the form fields is amount where user can put decimal value, e.g. 2.5.
In model changset defines `validate_number` function with `greater_than` option set to `0`.
Whatever value (lower than 0, 0, even greater than 0) is put into this field exception `no function clause matching in Decimal.compare/2` is raised.

#### Error stack

```
[error] #PID<0.341.0> running XPenser.Endpoint terminated
Server: localhost:4000 (http)
Request: POST /expenses
** (exit) an exception was raised:
    ** (FunctionClauseError) no function clause matching in Decimal.compare/2
        (decimal) lib/decimal.ex:275: Decimal.compare(#Decimal<-2>, 0)
        (ecto) lib/ecto/changeset.ex:1350: Ecto.Changeset.validate_number/6
        (elixir) lib/enum.ex:2587: Enum.do_find_value/3
        (ecto) lib/ecto/changeset.ex:1063: Ecto.Changeset.validate_change/3
        (xpenser) web/controllers/expense_controller.ex:17: XPenser.ExpenseController.create/2
        (xpenser) web/controllers/expense_controller.ex:1: XPenser.ExpenseController.action/2
        (xpenser) web/controllers/expense_controller.ex:1: XPenser.ExpenseController.phoenix_controller_pipeline/2
        (xpenser) lib/xpenser/endpoint.ex:1: XPenser.Endpoint.instrument/4
        (xpenser) lib/phoenix/router.ex:261: XPenser.Router.dispatch/2
        (xpenser) web/router.ex:1: XPenser.Router.do_call/2
        (xpenser) lib/xpenser/endpoint.ex:1: XPenser.Endpoint.phoenix_pipeline/1
        (xpenser) lib/plug/debugger.ex:122: XPenser.Endpoint."call (overridable 3)"/2
        (xpenser) lib/xpenser/endpoint.ex:1: XPenser.Endpoint.call/2
        (plug) lib/plug/adapters/cowboy/handler.ex:15: Plug.Adapters.Cowboy.Handler.upgrade/4
[...]

PR #1662, opened 2016-08-30
Update doc for validate_number method: add example with decimal

shhavel:
#1661


josevalim:
Thank you but I think we should automatically convert those to decimal. So I think we can push a better fix for #1661. :)

PR #1663, opened 2016-08-31
Allow validate_number compare decimal with number

shhavel:
#1661 
#1662

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1365,7 +1365,7 @@ defmodule Ecto.Changeset do
   end
 
   defp validate_number(field, %Decimal{} = value, message, spec_key, _spec_function, target_value) do
-    result = Decimal.compare(value, target_value)
+    result = Decimal.compare(value, Decimal.new(target_value))
     case decimal_compare(result, spec_key) do
       true  -> nil
       false -> [{field, {message, number: target_value}}]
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1365,6 +1365,7 @@ defmodule Ecto.Changeset do
   end
 
   defp validate_number(field, %Decimal{} = value, message, spec_key, _spec_function, target_value) do
+    target_value = if is_number(target_value), do: Decimal.new(target_value), else: target_value
```
michalmuskala: It should be safe to always call `Decimal.new/1` here

test/ecto/changeset_test.exs
```diff
@@ -799,6 +799,13 @@ defmodule Ecto.ChangesetTest do
     changeset = changeset(%{"decimal" => Decimal.new(1.5)})
                 |> validate_number(:decimal, greater_than_or_equal_to: Decimal.new(1.5))
     assert changeset.valid?
+
+    changeset = changeset(%{"decimal" => Decimal.new(-1)})
+                |> validate_number(:decimal, greater_than_or_equal_to: -1.5)
+    assert changeset.valid?
+    changeset = changeset(%{"decimal" => Decimal.new(1.5)})
+                |> validate_number(:decimal, greater_than_or_equal_to: 1.5)
+    assert changeset.valid?
```
michalmuskala: It's have one test with a float and one with an integer

lib/ecto/changeset.ex
shhavel: @michalmuskala Ok, will remove condition than. Thanks

test/ecto/changeset_test.exs
shhavel: Will add one test with an integer and one with float instead two tests with float.

michalmuskala:
Thank you! ❤️

Issue #1375, opened 2016-04-16
Inline embeds


michalmuskala:
Many changes introduced in Ecto 2.0 move it away from being just a database library and bring it closer to being a general data mapping solution.

We'd like to propose an enhancement to schemas and changesets that targets easy processing of external data, and makes Ecto more robust when it comes to data mapping. The aim of this proposal is to make Ecto schemas easier to use as a first line of your application, processing external data and getting it into structured format, that can be easily processed by other layers.

For a long time Ecto had a perfect solution for mapping such data - embeds. With the recent [removal of the `__meta__` field from embeds](https://github.com/elixir-lang/ecto/commit/e4eb4695a46e88c697511e3e981c223c3d2b1e6b) and removal of callbacks, embeds are even closer to being regular elixir structs with additional type information than to full-blown Ecto schemas. They make it easy to declare typed structs with custom validation logic that can be completely separate from the database. Using embedded schemas as pure data structure gives you the power of changesets and Ecto types for modeling your data. Those could be a part of your applications architecture in a role similar to form objects or commands known from various OO solutions. The biggest issue is that they become rather verbose when dealing with nested data, what is a common situation in many APIs - each level of nesting requires defining separate modules with schema declaration.

## Usage

We'd like to propose ability to define embedded schemas inline in the parent schema. 

```elixir
defmodule NewOrderCommand do
  use Ecto.Schema
  
  embedded_schema do
    field :user_id, Ecto.UUID
    
    embeds_one :address, Address do
      field :street, :string
      field :country, :string
      field :location, Geo.Point
    end
    
    embeds_many :items, Item do
      field :item_id, Ecto.UUID
      field :quantity, :integer
    end
  end
end
```
This would define two additional embedded schemas `NewOrderCommand.Address` and `NewOrderCommand.Item`. It's not yet clear if we'll go with the `embeds_one/many` accepting a block route or introduce separate `inline_embeds_one/many` macros.

The new schemas could be later used similarly to separately defined embeds with the exception that it woudn't be possible to define changeset functions that `cast_embed` expects. This shouldn't be an issue though, because we can use `with` option for passing arbitrary changeset function.

```elixir
def changeset(struct, params) do
  struct
  |> cast(params, [:user_id])
  |> cast_embed(:address, with: &address_changeset/2)
  |> cast_embed(:items, with: &item_changeset/2)
  |> validate_required([:user_id, :address, :items])
end

defp address_changeset(struct, params) do
  struct
[...]

Issue #1638, opened 2016-08-19
Support `@primary_key false` in embedded schemas


chrismccord:
As discussed with @michalmuskala, we should remove the primary key requirement on embedded schemas. I think having it as the default can remain and would allow this to be a forwards compatible change. The issue with the current behavior is for drivers like postgres, having an auto generated primary key on the embedded schema is mostly like pointless data for the majority of users, so it would be nice to add `@primary_key false` and skip the `id` key generated on embedded schemas when desired. I would be happy to help with this change, but I'm tapped out until after ElixirConf and Phoenix 1.3 ships. Thoughts?

josevalim:
I am fine with removing the requirement but keep it in mind we needs IDs if you want to track changes using put_assoc/put_embed in changesets. Otherwise it is hard for us to know how the data in the DB maps to the one sent by the user.

chrismccord:
😻 😻 😻

PR #1668, opened 2016-09-02
Introduce inline embeds

michalmuskala:
Closes https://github.com/elixir-ecto/ecto/issues/1375

This still needs documentation and possibly a little bit more testing.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1065,6 +1065,41 @@ defmodule Ecto.Schema do
       # Update the order
       changeset = Repo.update!(changeset)
 
+  ## Inline embedded schema
+
+  The schema module can be defined inline in the parent schema in simple
+  cases:
+
+      defmodule Parent do
+        use Ecto.Schema
+
+        schema "parents" do
+          field :name, :string
+
+          embeds_one :child, Child do
+            field :name, :string
+            field :age,  :integer
+          end
+        end
+      end
+
+  Defining embedded schema in such a way will define a `Parent.Child` module
+  with the appropriate struct. In order to properly cast the embedded schema.
+  When casting the inline-defined embedded schemas you need to use the `:with`
+  option of `cast_embed/3` to provide the proper function to do the casting.
+  For example:
+
+      def changeset(schema, params) do
+        schema
+        |> cast(params, [:name])
+        |> cast_embed(:child, with: &child_changeset/2)
+      end
+
+      defp child_changeset(schema, params) do
+        schema
+        |> cast(params, [:name, :age])
+      end
+
   ## Encoding and decoding
 
   Because many databases do not support direct encoding and decoding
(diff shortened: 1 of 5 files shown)


josevalim:
:heart:

michalmuskala:
Ready for review

lib/ecto/schema.ex
```diff
@@ -1139,9 +1196,66 @@ defmodule Ecto.Schema do
...
+  Defining embedded schema in such a way will define a `Parent.Child` module
+  with the appropriate struct. In order to properly cast the embedded schema.
+  When casting the inline-defined embedded schemas you need to use the `:with`
+  option of `cast_embed/3` to provide the proper function to do the casting.
+  For example:
+
+      def changeset(schema, params) do
+        schema
+        |> cast(params, [:name])
+        |> cast_embed(:children, with: &child_changeset/2)
```
josevalim: Do we raise a nice error message in case you don't? Should we?

michalmuskala: The current error will be along the lines of:

```
     ** (UndefinedFunctionError) function Module.changeset/2 is undefined or private. Did you mean one of:

           * __changeset__/0
```

which, I agree, is not the friendliest. 

If we were to raise something we should consider doing something in [here](https://github.com/elixir-ecto/ecto/blob/042aa92c62883572917477b2acaad3297590775c/lib/ecto/changeset.ex#L557), where we assume a `changeset/2` function will exist.

josevalim: I think we could do something for this particular case, which means we could define `changeset/2` in the embed module and raise a nice error message. Thoughts?

michalmuskala: What if in the casting code we checked if the function is defined and otherwise raised something along the lines of:

> the association module does not define a `changeset/2` function. You need to provide
> the function explicitly using the `:with` option.

This would probably cover more cases than just this and should cover this one as well.

josevalim: :+1:

michalmuskala:
Updated

michalmuskala:
Build errors fixed

josevalim:
:shipit:

josevalim:

lib/ecto/schema.ex
```diff
@@ -1143,14 +1200,71 @@ defmodule Ecto.Schema do
...
+  The schema module can be defined inline in the parent schema in simple
+  cases:
+
+      defmodule Parent do
+        use Ecto.Schema
+
+        schema "parents" do
+          field :name, :string
+
+          embeds_one :children, Child do
```
vsushkov: Should we change documentation here and use `embeds_many` example in order to avoid confusion?

michalmuskala: Yes! Thank you for spotting that, it's a mistake. Updated in master.

scottmessinger:
@michalmuskala This is awesome! One concern: is an `id` still required on `embeds_one` schemas? As Jose noted on the [mailing list back in February](https://groups.google.com/forum/#!topic/elixir-ecto/mS2tM5t8QPo), an ID on an embedded map doesn't make much sense. For arrays, an id on each map makes sense. From what I can tell in [this line](https://github.com/elixir-ecto/ecto/pull/1668/files#diff-08587a116d45885d97f1e77faf88f539R175), the ID is still required and autogenerated.

josevalim:
The ID will always be generated but issue #1638 will allow it to be skipped if you want to.

PR #1670, opened 2016-09-05
Disable lexical tracker for association modules

### LINKED ISSUE #1610: Association-induced compile time dependencies

michalmuskala:
This will prevent the compiler from creating a compile-time dependency
on the association module. It's safe to do, since in that place we
don't call any functions on the association module.

This solves an issue of excessive recompilation in Ecto projects with
heavy use of the associations.

Closes #1610

---

I verified on a sample project with `xref` that the dependency is indeed not created.

This is an alternative to https://github.com/elixir-ecto/ecto/pull/1637 that is significantly less intrusive and will allow older projects to benefit from it as well.

https://github.com/elixir-ecto/ecto/pull/1668 conflicts with those changes, it doesn't matter which one is merged first.

\cc @josevalim

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -595,6 +595,7 @@ defmodule Ecto.Schema do
 
   """
   defmacro has_many(name, queryable, opts \\ []) do
+    queryable = expand_alias(queryable, __CALLER__)
     quote do
       Ecto.Schema.__has_many__(__MODULE__, unquote(name), unquote(queryable), unquote(opts))
     end
@@ -651,6 +652,7 @@ defmodule Ecto.Schema do
 
   """
   defmacro has_one(name, queryable, opts \\ []) do
+    queryable = expand_alias(queryable, __CALLER__)
     quote do
       Ecto.Schema.__has_one__(__MODULE__, unquote(name), unquote(queryable), unquote(opts))
     end
@@ -829,6 +831,7 @@ defmodule Ecto.Schema do
   See `many_to_many/3` for more information on this particular approach.
   """
   defmacro belongs_to(name, queryable, opts \\ []) do
+    queryable = expand_alias(queryable, __CALLER__)
     quote do
       Ecto.Schema.__belongs_to__(__MODULE__, unquote(name), unquote(queryable), unquote(opts))
     end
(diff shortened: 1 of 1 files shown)


michalmuskala:
I verified on a sample project with `xref` that the dependency is indeed not created.

michalmuskala:
Closed by accident

josevalim:
Beautiful. I was planning to work on it just now. I have two small suggestions:

1. Let's move this to a private function and expand only the if the AST is `{:__aliases__, ..., ...}`. That's because maybe you are invoking another module to get the association name

2. Let's do the same for the `field` second argument.

rranelli:
Doing the same for `field` also solves the problem with custom attributes :)

That's indeed a nicer solution.

josevalim:
Actually, we shouldn't do it for fields given we have this code: https://github.com/elixir-ecto/ecto/blob/c07a8eebefa424c7021d1e0b86c0fa4c5d447f64/lib/ecto/schema.ex#L1474

josevalim:
I assume not doing it for fields is OK as you have nowhere as many custom fields as resources.

michalmuskala:
Updated.

I agree that custom types are not a big issue. The issue with associations was mainly the cascading recompilation, which does not happen with type modules.

PR #1671, opened 2016-09-06
Update docs: Ecto.Pools.Poolboy is now DBConnection.Poolboy

balexand:
It looks like `Ecto.Pools.Poolboy` no longer exists.


fishcakez:
Thank you :heart:

PR #1682, opened 2016-09-11
Fix typo in comments regarding foreign key type.

StevenXL:
- Comment erroneously suggested that the type of the foreign key in a
`belongs_to` association defaults to `:id`, when it in fact defaults to
`:integer`.


josevalim:

PR #1683, opened 2016-09-11
Partial denormalization: duplicate parent fields

shhavel:
Hi, I understand that this PR seems odd and probably will not be merged.
But maybe this kind of feature is interesting or there is already ability to do this.

### The problem

**Store chosen foreign key(s) of a parent table for better queries performance.**

### Example

normilized

Model `Group` has many `posts` (model `Post`),
model `Post` has many `comments` (model `Comment`)

table `posts` stores foreign key `group_id`
table `comments` stores foreign key `post_id`

To check if user has access to the `Comment` I need to check `group_id` which is stored in `posts`.
So I need to do join table or an extra query on every read.

### Solution

partiallly_denormilized

Add column `group_id` to table `comments` and copy value from model `Post`
on creating `Comment` and on updating `Post` group (for all post's comments).

It is easy to do this almost everywhere but not in `cast_assoc`.

Please see my changes and maybe you have some better suggestions.

Thanks

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -999,4 +999,68 @@ defmodule Ecto.Integration.RepoTest do
     Process.put(:on_log, fn _ -> flunk("logged") end)
     TestRepo.insert!(%Post{title: "1"}, [log: false])
   end
+
+  ## Denormalization
+
+  test "sync duplicated parent fields on insert" do
+    defmodule Comment do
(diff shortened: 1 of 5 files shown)


josevalim:
@shhavel you can probably implement this feature with a function call that checks the changeset after doing your changes and trying to figure out if the change should be propagated. But yes, it doesn't belong in Ecto per se. :)

Issue #1526, opened 2016-06-30
Upsert


atsheehan:
There was previous discussion on adding the **UPSERT** feature to Ecto in issue #652 and PR #746 by @gjaldon. I'm not sure if this the best place to revisit the topic, but I figured I'd open this issue to leave my notes and start a discussion. I can move this to the mailing list if it's more appropriate there.

# Adapters

UPSERT is handled differently (or not at all) by different databases. The PostgreSQL wiki has a [brief overview of how UPSERTs are handled in other systems](https://wiki.postgresql.org/wiki/UPSERT#UPSERT_as_implemented_in_practice).

## PostgreSQL

The [`ON CONFLICT` syntax](https://www.postgresql.org/docs/9.5/static/sql-insert.html) was added in 9.5+. It doesn't look like there is an easy way to emulate this behavior pre 9.5, so possibly raise an error if not available.

There are two options when a unique constraint is violated: `DO NOTHING` or `DO UPDATE`.

```
INSERT INTO table_name ( column_name [, ...] ) VALUES ( values [, ...] )

  ON CONFLICT [ conflict_target ] DO NOTHING

-- or

  ON CONFLICT conflict_target DO UPDATE SET ...
```

### Conflict Target

PostgreSQL requires the `conflict_target` be set when performing an update. The `conflict_target` is used to determine which unique constraint to check when identifying a potential row to update. In addition to checking for columns that belong to a unique index, it also allows index expressions and predicates.

It seems that the simple case for an upsert would be to specify a list of columns to check, but there will also be times where you may need to specify an index expression or predicate as well.

```
conflict_target: [:col_a, :col_b]
conflict_target: "(lower(col_a), col_b) WHERE col_c = false"
```

### Conflict Action

Once a candidate row is found on conflict, it can either be ignored or updated. The update statement seems to allow for arbitrary fields to be modified, but it might be safe to assume we can just update the columns with the supplied values. The new values are available in the `EXCLUDED` table.

```
ON CONFLICT (col_a) DO UPDATE SET col_b = EXCLUDED.col_b, col_c = EXCLUDED.col_c
```

I think we'd have to specify the list of columns to update.

```
conflict_target: [:col_a], columns: [:col_b, :col_c]
```

There is also an optional `[ WHERE condition ]` which will only update conflicting rows that pass this expression. It may be safe to assume that the user wants to update every row that is in conflict.

## MySQL

[...]

PR #1687, opened 2016-09-13
Upsert #1526

alesasnouski:


integration_test/cases/repo.exs
```diff
@@ -999,4 +999,134 @@ defmodule Ecto.Integration.RepoTest do
...
+    post = TestRepo.one!(Post)
+    assert post.id == inserted.id + 1
+    assert post.title == "second"
+  end
+
+  @tag :upsert
+  test "upsert on conflict ignore" do
+    {:ok, inserted} = TestRepo.upsert(%Ecto.Integration.Post{title: "first", uuid: "6fa459ea-ee8a-3ca4-894e-db77e160355e"},
+      on_conflict: :nothing)
+
```
josevalim: Should we remove this new line?

integration_test/cases/repo.exs
```diff
@@ -999,4 +999,134 @@ defmodule Ecto.Integration.RepoTest do
...
+    assert inserted.id > 0
```
josevalim: Maybe `assert inserted.id`? There is no guarantee ID is an integer.

lib/ecto/changeset.ex
```diff
@@ -174,7 +174,7 @@ defmodule Ecto.Changeset do
                         types: nil | %{atom => Ecto.Type.t}}
 
   @type error :: {String.t, Keyword.t}
-  @type action :: nil | :insert | :update | :delete | :replace
+  @type action :: nil | :insert | :upsert | :update | :delete | :replace
```
josevalim: I don't think we need this change here?

josevalim:
@alesasnouski would it be possible to keep @atsheehan original's commit with tests? Also, could you please add unit tests in `test/ecto/adapters/postgres_test.exs` and friends for the upsert queries?

lib/ecto/adapters/mysql.ex
```diff
@@ -196,6 +196,34 @@ defmodule Ecto.Adapters.MySQL do
   end
 
   @doc false
+  def upsert(repo, %{source: {prefix, source}, autogenerate_id: {key, :id}}, params, on_conflict, conflict_target, update, returning, opts) do
```
josevalim: This line is too long. You can call the second argument `meta` and do the matching on the key and source on the next line.

lib/ecto/changeset.ex
josevalim: Oh, I see why you added this. I think we need to consider upsert either an insert or an update. Maybe it would be best to *not* add an `upsert` function in the repository and instead use `insert` and use the `:on_conflict` option to specify the conflict resolution?

josevalim:
@alesasnouski This looks excellent, thank you!

I have added some comments. The most important for now is to add unit tests for the adapter query generation. If we are going to use insert or upsert we can decide it later on. :)

lib/ecto/adapters/sql.ex
```diff
@@ -91,6 +91,34 @@ defmodule Ecto.Adapters.SQL do
...
+      def upsert(repo, %{source: {prefix, source}}, params, on_conflict, conflict_target, update, returning, opts) do
+        {fields, ins_values} = :lists.unzip(params)
+        values = case on_conflict do
+          :update ->
+            checked_update = Enum.filter(update, fn update_key -> Keyword.has_key?(params, update_key) end)
+            update_values = Enum.map(checked_update, fn update_key -> Keyword.get(params, update_key) end)
+            ins_values ++ update_values
+          _ -> ins_values
+        end
+        sql = @conn.upsert(prefix, source, fields, [fields], on_conflict, conflict_target, update, returning)
```
atsheehan: If we're calling `upsert` on the connection, maybe we should also add `upsert` as a callback on the `Ecto.Adapters.SQL.Connection` behavior? Even if not a database doesn't supports upsert, they could still add a function to raise with an appropriate error message.

atsheehan:
Hey @alesasnouski, thanks for implementing this! Very much looking forward to upserts in Ecto.

One thing that came up in #1541 about the interface was supporting arbitrary update statements. So in addition to `on_conflict: :update`, also accepting options like `on_conflict: [inc: [count: 1]]` or `on_conflict: from(p in Post, where: p.archived == false, update: [inc: [count: 1]])`. The latter two are supported by [`update_all`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:update_all/3), so it may be possible to leverage the existing code from [Ecto.Repo.Queryable](https://github.com/elixir-ecto/ecto/blob/1cde33a9464e1c5d725f6118e0ccbed4720583fb/lib/ecto/repo/queryable.ex#L80-L97) to generate the UPDATE clause.

I've been trying to think of what information the database adapter needs to generate these upsert statements for the `Ecto.Adapters.SQL.Connection`. I think there are three scenarios:

* Generate SQL that ignores the conflict.

```elixir
def upsert(prefix, table, header, rows, :ignore, conflict_target, returning)

## Example

Ecto.Adapters.Postgres.Connection.upsert(
  nil, "posts", [:title, :uuid], [["foo", "bar"]], :ignore, [:uuid], false)
# INSERT INTO posts (title, uuid) VALUES ($1, $2) ON CONFLICT (uuid) DO NOTHING

Ecto.Adapters.MySQL.Connection.upsert(
  nil, "posts", [:title, :uuid], [["foo", "bar"]], :ignore, [:uuid], false)
# INSERT IGNORE INTO posts (title, uuid) VALUES ($1, $2)
"""
```

* Generate SQL that updates a set of columns on conflict.

```elixir
def upsert(prefix, table, header, rows, on_conflict, conflict_target, returning) when is_list(on_conflict)

## Example

Ecto.Adapters.Postgres.Connection.upsert(
  nil, "posts", [:title, :uuid], [["foo", "bar"]], [:title], [:uuid], false)
# INSERT INTO posts (title, uuid) VALUES ($1, $2)
#   ON CONFLICT (uuid) DO UPDATE SET title = EXCLUDED.title

Ecto.Adapters.MySQL.Connection.upsert(
  nil, "posts", [:title, :uuid], [["foo", "bar"]], [:title], [:uuid], false)
# INSERT INTO posts (title, uuid) VALUES ($1, $2)
#   ON DUPLICATE KEY UPDATE title = VALUES(title)
```

* Generate SQL that runs an arbitrary update statement on conflict.

```elixir
def upsert(prefix, table, header, rows, %Ecto.Query{} = on_conflict, conflict_target, returning)

## Example

Ecto.Adapters.Postgres.Connection.upsert(
  nil, "posts", [:tiitle, :uuid], [["foo", "bar"]],
  %Ecto.Query{updates: [inc: [count: 1]]}, [:uuid], false)
# INSERT INTO posts (title, uuid) VALUES ($1, $2)
#   ON CONFLICT (uuid) DO UPDATE SET count = count + 1

Ecto.Adapters.MySQL.Connection.upsert(
  nil, "posts", [:tiitle, :uuid], [["foo", "bar"]],
  %Ecto.Query{updates: [inc: [count: 1]]}, [:uuid], false)
# INSERT INTO posts (title, uuid) VALUES ($1, $2)
#   ON DUPLICATE KEY UPDATE count = count + 1
```

Not sure if the examples above are entirely correct, just wanted to get your thoughts on if you think this might be a valid approach or if I'm missing/overcomplicating something. Thanks!

integration_test/cases/repo.exs
```diff
@@ -999,4 +999,134 @@ defmodule Ecto.Integration.RepoTest do
...
+
+    {:ok, not_inserted} = TestRepo.upsert(%Ecto.Integration.Post{title: "first", uuid: "6fa459ea-ee8a-3ca4-894e-db77e160355e"},
+      on_conflict: :nothing, update: [:id]) # Update field specified for mysql resulting in ON DUPLICATE KEY `id` = `id`
+    assert not_inserted.id in [0, nil]
```
alesasnouski: What `id` should we get in case of duplicate record and do :nothing ? Now we are getting 0 from postgres and nil from mysql ...

josevalim:
> One thing that came up in #1541 about the interface was supporting arbitrary update statements. So in addition to on_conflict: :update, also accepting options like on_conflict: [inc: [count: 1]] or on_conflict: from(p in Post, where: p.archived == false, update: [inc: [count: 1]])

We should support this but let's not worry about this for now. Let's get upsert as a whole in and then we can make it more complex later on.

integration_test/cases/repo.exs
josevalim: In this case, is it possible for postgres/mysql to know the insert did not change anything to the table? Because if we can know that, then we can make it so none of them change anything at all (i.e. both are nil).

alesasnouski:
would it be possible to keep @atsheehan original's commit with tests?
I removed @atsheehan code from my branch. @atsheehan , can you update your tests on the base of my branch ? And docs :) I will review your PR.

atsheehan:
Hi @alesasnouski, I rebased your changes on my branch to preserve the ordering, I'm not sure if there's a way to rebase into your fork with a PR though. The changes are pushed up to #1541 and I added you as a collaborator if it would be easier to work off one repo/PR.

PR #1692, opened 2016-09-14
Fix `put_assoc` with schema_prefix

foeken:
# Inserting/Updating/Deleting associations through a changeset does not respect `schema_prefix`

```elixir
defmodule MyApp.Person do
  schema "people" do
    has_many :cats, MyApp.Cat
  end

  def changeset(model, params \\ %{}) do
    model
    |> cast(params, [:name], [])
    |> cast_assoc(:cats) # Is this needed? What does it even do?
  end
end

model = %MyApp.Person{}
|> Ecto.put_meta(prefix: "my_tenant_database")

MyApp.Person.changeset(model, %{name: "José"})
|> put_assoc(:cats, [%{name: "Fluffy"}])
|> MyApp.Repo.insert!
# => ERROR: Unknown table `cats`
```

# Expected result

Both a new Person and Cat get inserted into `my_tenant_database`.

# Actual result

An error message that the cats table does not exist on the global database.

# Solution

Copy the prefix from the parent changeset when it is set.

# Question

- I would love to add a test for this but I might need some pointers on where to start.
- I don't know the use of the `cast_assoc` inside the changeset, it seems to work without it, but the docs are quite vague as to what it is used for...

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -179,6 +179,16 @@ defmodule Ecto.Association do
     struct
   end
 
+  @doc false
+  def update_parent_prefix(changeset, parent) do
+    case parent do
+      %{__meta__: %{source: {prefix, _}}} ->
+        update_in changeset.data, &Ecto.put_meta(&1, prefix: prefix)
+      _ ->
+        changeset
+    end
+  end
+
   @doc """
   Performs the repository action in the related changeset,
   returning `{:ok, data}` or `{:error, changes}`.
@@ -412,12 +422,14 @@ defmodule Ecto.Association.Has do
   end
 
   @doc false
-  def on_repo_change(%{on_replace: on_replace} = refl, parent_changeset,
+  def on_repo_change(%{on_replace: on_replace} = refl, %{data: parent} = parent_changeset,
                      %{action: :replace} = changeset, opts) do
     changeset = case on_replace do
       :nilify -> %{changeset | action: :update}
       :delete -> %{changeset | action: :delete}
     end
+
+    changeset = Ecto.Association.update_parent_prefix(changeset, parent)
 
     case on_repo_change(refl, %{parent_changeset | data: nil}, changeset, opts) do
       {:ok, _} -> {:ok, nil}
@@ -431,6 +443,7 @@ defmodule Ecto.Association.Has do
 
     {key, value} = parent_key(assoc, parent)
     changeset = update_parent_key(changeset, action, key, value)
+    changeset = Ecto.Association.update_parent_prefix(changeset, parent)
 
     case apply(repo, action, [changeset, opts]) do
       {:ok, _} = ok ->
@@ -767,7 +780,9 @@ defmodule Ecto.Association.BelongsTo do
     on_repo_change(refl, parent_changeset, %{changeset | action: :delete}, opts)
   end
 
-  def on_repo_change(_refl, %{repo: repo}, %{action: action} = changeset, opts) do
+  def on_repo_change(_refl, %{data: parent, repo: repo}, %{action: action} = changeset, opts) do
+    changeset = Ecto.Association.update_parent_prefix(changeset, parent)
+
     case apply(repo, action, [changeset, opts]) do
       {:ok, _} = ok ->
         if action == :delete, do: {:ok, nil}, else: ok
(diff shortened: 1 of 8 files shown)


josevalim:
@foeken I am torn on this one because it removes the ability of people customizing it per association basis. Why don't you set it for your associations as well?

foeken:
Where would I do that? Do you mean that I could replace passing a map to `put_assoc`with passing a struct with the meta prefix? I'll see if that works.

foeken:
Yes, that works but it adds a lot of boilerplate code. I remember having this same discussion from the other perspective with you last time but then you were arguing *not* to allow different prefixes in the same query (for preloading/joining) since prefixes should be use for tenants, why would you allow mixing them for inserting stuff?

foeken:
```elixir
model = %MyApp.Person{}
|> Ecto.put_meta(prefix: "my_tenant_database")

MyApp.Person.changeset(model, %{name: "José"})
|> put_assoc(:cats, [%{name: "Fluffy"}])
|> MyApp.Repo.insert!
```

changes into:

```elixir
model = %MyApp.Person{}
|> Ecto.put_meta(prefix: "my_tenant_database")

cats = Enum.map([%{name: "Fluffy"}], fn(c) ->
  model = %MyApp.Cat{}
  |> Ecto.put_meta(prefix: "my_tenant_database")
  MyApp.Cat.changeset(model, c)
end)

MyApp.Person.changeset(model, %{name: "José"})
|> put_assoc(:cats, cats)
|> MyApp.Repo.insert!
```

lib/ecto/association.ex
```diff
@@ -432,9 +432,13 @@ defmodule Ecto.Association.Has do
     {key, value} = parent_key(assoc, parent)
     changeset = update_parent_key(changeset, action, key, value)
 
-    changeset = case parent.__meta__.source do
-      { nil, _ } -> changeset
-      { prefix, _ } -> %{changeset | data: Ecto.put_meta(changeset.data, prefix: prefix)}
+    changeset = if parent do
```
josevalim: This would be cleaner:

```elixir
changeset =
  case parent do
    %{__meta__: %{source: {prefix, _}}} -> update_in changeset.data, &Ecto.put_meta(&1, prefix: prefix)
    _ -> changeset
  end

**Comment by josevalim** [comment]
> I remember having this same discussion from the other perspective with you last time but then you were arguing not to allow different prefixes in the same query (for preloading/joining) since prefixes should be use for tenants, why would you allow mixing them for inserting stuff?

Yup. I was thinking the same thing after replying. Please give me a day or two to think about this. For the queries, this would certainly make it consistent with preloading.

**Comment by josevalim** [comment]
Regarding the tests, the best would be a unit test that checks later on that the prefix in the associations were propagated.

**Comment by foeken** [comment]
I will work on it today to finish the PR, take any time you need!

**Comment by foeken** [comment]
I finished my changes. In order to get them to test properly I had to refactor some stuff, let me know if you agree. The TL;DR:

 - The TestRepo now sends messages with a consistent and *full* source (`{prefix, table}`) instead of just the table name for the *_all cases. This has been refactored in all tests cases.

- I added three new test cases (update, insert, delete) for each relation (Has, BelongsTo, ManyToMany) to ensure the prefix is preserved

- I added a helper method (`update_parent_prefix`) to prevent too much duplication, but I am unsure if it's in the best location. I added it in `Ecto.Association`

File: lib/ecto/association.ex [hunk 2]
```diff
@@ -412,13 +422,15 @@ defmodule Ecto.Association.Has do
...
   @doc false
-  def on_repo_change(%{on_replace: on_replace} = refl, parent_changeset,
+  def on_repo_change(%{on_replace: on_replace} = refl, %{data: parent} = parent_changeset,
                      %{action: :replace} = changeset, opts) do
     changeset = case on_replace do
       :nilify -> %{changeset | action: :update}
       :delete -> %{changeset | action: :delete}
     end
 
+    changeset = Ecto.Association.update_parent_prefix(changeset, parent)
```
foeken: This is required here since the parent is nullified on line 434 and thus looses its prefix. I was unsure if you would normally add a comment here to reflect that.

lib/ecto/association.ex
foeken: Thanks for this, the `update_in` method is awesome. I learned something 😺

josevalim:

PR #1693, opened 2016-09-15
Update schema.ex

jrnt30:
Updating example about belongs_to association.  It was difficult to reason for me how `company` would `belong_to` a `manager`.  Instead used what I believe is a more natural association of `employee <-> company`


josevalim:

Issue #1694, opened 2016-09-15
Ecto.Adapters.MySQL shouldn't use UNIX socket when hostname is localhost


shinta-h:
Currently, Ecto.Adapters.MySQL use UNIX socket when hostname is localhost
because it execute mysql command internally.
https://dev.mysql.com/doc/refman/5.7/en/connecting.html
https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/mysql.ex

So, it require both UNIX socket and TCP socket for Mariaex ( not yet support UNIX socket https://github.com/xerions/mariaex/issues/66 ) when hostname is localhost.

The behavior can be unexpected in the cases that both TCP socket and UNIX socket are not listened. 
For example,
- mysqld is running in Docker.
- Google Cloud SQL Proxy is running on localhost.

IMO, Ecto.Adapters.MySQL should always use only TCP socket until both Ecto and Mariaex support UNIX socket.

PR #1695, opened 2016-09-15
Force to use TCP socket when hostname is localhost

shinta-h:
It is PR to force to use only TCP socket in Ecto.Adapters.MySQL when hostname is localhost.
https://github.com/elixir-ecto/ecto/issues/1694


josevalim:

shinta-h:
Thanks!

PR #1702, opened 2016-09-18
El add comment support on pgsql

eniolopes:


integration_test/pg/constraints_test.exs
```diff
@@ -9,16 +9,16 @@ defmodule Ecto.Integration.ConstraintsTest do
   defmodule ConstraintMigration do
     use Ecto.Migration
 
-    @table table(:constraints_test)
+    @table table(:constraints_test, comment: "table")
```
josevalim: I would move those comments to the actual migration file in `integration_test/support/migrations.exs` since then it could be useful for all databases that support migrations. :)

eniolopes: done

eniolopes:
@josevalim Now tests are failing because MySQL doesn't support constraints, should I just remove it from the test or can we do a better work around?

integration_test/support/migration.exs
```diff
@@ -39,7 +39,9 @@ defmodule Ecto.Integration.Migration do
     # Add a unique index on uuid. We use this
     # to verify the behaviour that the index
     # only matters if the UUID column is not NULL.
-    create unique_index(:posts, [:uuid])
+    create unique_index(:posts, [:uuid], comment: "posts index")
+
+    create constraint(:posts, :cost_must_be_positive, check: "cost > 0")
```
josevalim: Just remove this one, it is fine, yes.

eniolopes: :+1:

eniolopes:
@josevalim ping

josevalim:

PR #1704, opened 2016-09-20
Ecto is not able to handle malformed UUID.

noma4i:
ordinary ID - can.

I think simple `Ecto.UUID.cast` should happen on input param to return an `:error` other then raising `cannot be cast to type`

Idea is to utilise existing `preprocess` in `lib/ecto/repo/queryable.ex` to perform validation/casting w/o raising error hence returning it to user. This will follow general logic of Ecto - return errors to use them in cond, case, and conditions.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -519,6 +519,25 @@ defmodule Ecto.Integration.RepoTest do
     assert_raise Ecto.NoResultsError, fn ->
       TestRepo.get!(Post, post1.id)
     end
+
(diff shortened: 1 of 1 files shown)


josevalim:
They will both fail, the issue is that you gave a cast-able ID to the integer one.

noma4i:
I see. Letters passed to integer id will lead to fail as well. What about general idea of filtering out such cases to return error instead of raise?

josevalim:
See the linked issue #1705 where we are discussing an API to allow that. I am closing this for now so we can continue it there. :)

PR #1710, opened 2016-09-22
Allow nil as first argument to Repo.preload

jfrolich:


shhavel:
@jfrolich  I think also a corresponding test should be added. Thanks

jfrolich:
Thanks for the feedback! I am not very comfortable around the source code yet, so let me know if the location of the test is right!

shhavel:
@jfrolich I think the location of the test is perfect.

josevalim:

PR #1722, opened 2016-09-30
Ecto dump models

alexandrubagu:
added ecto.dump.models tasks for generating Models from command line


josevalim:
Thank you @alexandrubagu, however we have no plans to support those in ecto. Maybe you should do it as a separate project? Also, keep in mind we no longer use the word model in Ecto. You probably want to call that schemas or something. :)

alexandrubagu:
Thanks, @josevalim. I will start a brand new project maybe will be people that will search for this. I was thinking that will be useful if you start a phoenix project with an existing database.

Issue #1107, opened 2015-11-29
Ecto.Changeset.validate_change/3 should check field


michalmuskala:
It's an issue I see often repeated - someone adding a validation on a field, but forgetting to add it to required or optional fields in cast.

I propose that we raise in `validate_change/3` if the field is not in the changes map, and is neither in `required` nor `optional` fields.

josevalim:
Yes, that's one way to do it. Another option is to warn if a field is given to cast parameters but never used. If we warn with Logger.debug, we should be fine.

tony612:
Yeah, this is a problem indeed. And I prefer warning in log.

josevalim:
This will be for 2.1 or 2.2 because I would like to be able to disable the log and the API for such in 2.0 would be confusing.

aphillipo:
I too had this problem.

Could it be even better to simply not have an optional fields key and just calculate them somehow, maybe with a macro?

michalmuskala:
@josevalim I was looking into this, and I'm not sure what you meant by a field "not being used". Could you clarify that?

I still thing it would be totally ok, to raise if you give a field that does not exist in the schema to `validate_change/3`.

josevalim:
> I still thing it would be totally ok, to raise if you give a field that does not exist in the schema to validate_change/3.

But that's not the issue here right? The field exists in the schema... but it was not cast as a change. And that's OK... specially if you have validations being shared by two different cast/2 calls. The best we can do is log, I would't raise.

michalmuskala:
The initial issue I reported is validation used on a field that was never casted - it doesn't matter if it exists in the schema. So the problematic thing is:

```elixir
schema
|> cast(params, [:foo])
|> validate_change(:bar, &validate_bar/2)
```

The validation function will never run, thus it's safe to raise - the call is either useless and can be removed, or erroneous and should be corrected.

josevalim:
> The validation function will never run, thus it's safe to raise - the call is either useless and can be removed, or erroneous and should be corrected.

I disagree. The current code allows two changeset functions to share a validation path even if they don't share all of validations. Maybe we want to do this change anyway, I am just pointing out there is a scenario where the current code is valid and that's why we should start with at least a warning.

michalmuskala:
I see your point. So maybe we should limit ourselves to situations when the field is not in changes and not in the schema?

PR #1725, opened 2016-10-02
Raise error when non-existing field is being validated

wojtekmach:
Solves part of #1107.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1234,7 +1234,7 @@ defmodule Ecto.Changeset do
   end
 
   defp ensure_field_exists!(%Changeset{types: types}, field) do
-    unless field in Map.keys(types) do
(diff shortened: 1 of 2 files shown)


josevalim:
Sounds good to me. @michalmuskala?

michalmuskala:
Are we sure this should be an argument error? 

Other than that a :+1: from me.

josevalim:
ArgumentError sounds good. Let's give this patch a try. @wojtekmach should we add a bit more information? "unknown field #{inspect field} for changeset on #{inspect changeset.data}"?

wojtekmach:
Adding more information sounds good, I'll get on it this evening. I was initially thinking about listing all available fields in the error message, but inspecting the data sounds good too.

wojtekmach:
@josevalim updated, this is how it looks like now:

preview

PR #1731, opened 2016-10-05
Fix Changeset.types type

wojtekmach:
Looks like it never used Keyword.t. I verified in the commit that
introduced schemaless changesets (019e2bc)


josevalim:

PR #1732, opened 2016-10-06
WIP Validate unique embeds primary key

### LINKED ISSUE #1729: Embeds with the same primary key can be inserted
### Environment

* Elixir 1.3.3
* PostgreSQL 9.5.4
* Ecto 2.1.0-rc.1
* postgrex 0.12.1
* Arch Linux

### Current behavior

I have the ff. `Parent` schema:

```elixir
defmodule MyApp.Parent do
  use MyApp.Web, :model

  schema "parent" do
    field :name, :string
    embeds_many :children, MyApp.Child

    timestamps()
  end

    struct
    |> cast(params, [:name])
    |> cast_embed(:children)
    |> validate_required([:name])
  end
end
```

I want to ensure that children's names are unique for a given parent. I'm trying to do this by making the children's `name` field the primary key:

```elixir
defmodule MyApp.Child do
  use MyApp.Web, :model

  @primary_key {:name, :string, autogenerate: false}

  embedded_schema do
    field :age, :integer
  end

    struct
    |> cast(params, [:name, :age])
    |> validate_required([:name])
  end
end
```

However, when I try to add children with the same name, both children are added successfully:

```
iex(2)> Repo.insert! Parent.changeset(%Parent{}, %{name: "Alice", children: [%{name: "Bob"}, %{name: "Bob"}]})

[debug] QUERY OK db=6.0ms
INSERT INTO "parent" ("children","name","inserted_at","updated_at") VALUES ($1,$2,$3,$4) RETURNING "id" [[%{age: nil, name: "Bob"}, %{age: nil, name: "Bob"}], "Alice", {{2016, 10, 5}, {7, 34, 33, 658211}}, {{2016, 10, 5}, {7, 34, 33, 660880}}]

%MyApp.Parent{__meta__: #Ecto.Schema.Metadata<:loaded, "parent">,
 children: [%MyApp.Child{age: nil, name: "Bob"},
  %MyApp.Child{age: nil, name: "Bob"}], id: 1,
 inserted_at: ~N[2016-10-05 07:34:33.658211], name: "Alice",
 updated_at: ~N[2016-10-05 07:34:33.660880]}
```

### Expected behavior

Being the primary key, names should be unique.

shhavel:
Fix #1729

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -204,9 +204,10 @@ defmodule Ecto.Changeset.Relation do
     {:ok, [], true, false}
   end
 
-  defp cast_or_change(%{cardinality: :many}, value, current, current_pks,
+  defp cast_or_change(%{cardinality: :many, related: mod}, value, current, current_pks,
                       new_pks, fun) when is_list(value) do
-    map_changes(value, new_pks, fun, process_current(current, current_pks), [], true, true)
+    pks = mod.__schema__(:primary_key)
+    map_changes(value, new_pks, fun, process_current(current, current_pks), [], true, true, pks, [])
   end
 
   defp cast_or_change(_, _, _, _, _, _), do: :error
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @shhavel! :heart:

I was thinking more about this and I think we need to validate it inside the repository callback: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/embedded.ex#L49

Otherwise you can still do this:

    Repo.insert %Post{comments: [%Embed{name: "foo"}, %Embed{name: "foo"}]}

and we won't abort.

What do you think?

shhavel:
@josevalim Yes, sure, I will update pull request.

josevalim:

aptinio:
Thank you! :-)

PR #1735, opened 2016-10-08
Rename example migration module

jbranchaud:
The example is about a Weather table and a few lines further down the
docs is a sample `mix ecto.gen.migration` that uses `add_weather_table`,
so it makes sense to change `CreatePosts` to `AddWeatherTable`.


josevalim:

PR #1736, opened 2016-10-09
Support passing `:ownership_timeout` when checking out a connection.

myronmarston:
Fixes elixir-ecto/db_connection#62.

I wasn't sure how to test this so I'm hoping one of the Ecto maintainers can suggest a good way to do that.

Also, I wasn't sure if there were any other options that we should let be overriden by the `checkout` call.  For now it's just `:ownership_timeout` but the list can easily be expanded.

/cc @fishcakez

### DIFF
--- a/integration_test/sql/sandbox.exs
+++ b/integration_test/sql/sandbox.exs
@@ -127,4 +127,18 @@ defmodule Ecto.Integration.SandboxTest do
 
     assert_receive :success
   end
+
+  test "allows an ownership timeout to be passed for an individual `checkout` call" do
+    log = capture_log fn ->
+      :ok = Sandbox.checkout(TestRepo, ownership_timeout: 20)
+
+      Process.sleep(1000)
+
+      assert_raise DBConnection.OwnershipError, fn ->
+        TestRepo.all(Post)
+      end
+    end
+
+    assert log =~ ~r/timed out.*20ms/
+  end
 end
--- a/lib/ecto/adapters/sql/sandbox.ex
+++ b/lib/ecto/adapters/sql/sandbox.ex
@@ -232,6 +232,11 @@ defmodule Ecto.Adapters.SQL.Sandbox do
   [`DBConnection.Ownership`](https://hexdocs.pm/db_connection/DBConnection.Ownership.html)
   and defaults to 15000ms. Timeouts are given as integers in milliseconds.
 
+  Alternately, if this is an issue for only a handful of long-running tests,
+  you can pass an `:ownership_timeout` option when calling
+  `Ecto.Adapters.SQL.Sandbox.checkout/2` instead of setting a longer timeout
+  globally in your config.
+
   ### Database locks and deadlocks
 
   Since the sandbox relies on concurrent transactional tests, there is
@@ -442,6 +447,11 @@ defmodule Ecto.Adapters.SQL.Sandbox do
       a transaction. Defaults to true.
 
     * `:isolation` - set the query to the given isolation level
+
+    * `:ownership_timeout` - limits how long the connection can be
+      owned. Defaults to the compiled value from your repo config in
+      `config/config.exs` (or preferably in `config/test.exs`), or
+      15000 ms if not set.
   """
   def checkout(repo, opts \\ []) do
     {name, pool_opts} =
@@ -450,6 +460,9 @@ defmodule Ecto.Adapters.SQL.Sandbox do
       else
         repo.__pool__
       end
+
+    pool_opts_overrides = Keyword.take(opts, [:ownership_timeout])
+    pool_opts = Keyword.merge(pool_opts, pool_opts_overrides)
 
     case DBConnection.Ownership.ownership_checkout(name, pool_opts) do
       :ok ->


josevalim:
Maybe one option to test it is to call `Process.flag(:trap_exit, true)` in the test process, checkout a connection with a very low timeout value (like 10ms) and then assert you receive the timeout message on your inbox.

@fishcakez will be able to confirm if this is a good or bad idea™. :)

fishcakez:
I think it might be easier to assert an ownership exception is raised on query after the ownership timeout expires. For example we could do a (manual) checkout with ownership_timeout set to a tiny ms and then sleep for a slightly longer time, then attempt a test query and assert it raises `DBConnection.OwnershipError` because the proxy has timed out. We would also capture log to make sure we have a nice log message:

```elixir
assert capture_log(fn() ->
  :ok = Ecto.Adapters.Sandbox.checkout(TestRepo,  [ownership_timeout: 10])
  :timer.sleep(1000)
  assert_raise DBConnection.OwnershipError, fn() -> TestRepo.all() end
end) =~ "timed out"
```

myronmarston:
Thanks for the suggestions!  I just pushed a test that's similar to @fishcakez's test.

It usually passes, but occasionally I'm seeing some weird behavior from it.

Once, it failed due to the log being empty, but the console showed that it logged the expected message:

```
➜  ecto git:(myron/support-pool-timeout) ✗ mix test.adapters
==> Running tests for MIX_ENV=mysql mix test
Excluding tags: [:array_type, :read_after_writes, :uses_usec, :uses_msec, :returning, :strict_savepoint, :create_index_if_not_exists, :modify_column, :transaction_isolation, :rename_column, :upsert_all, :with_conflict_target]

..............................................................................................................................................................................................................
18:45:14.768 [error] Mariaex.Protocol (#PID<0.212.0>) disconnected: ** (DBConnection.ConnectionError) owner #PID<0.963.0> timed out because it owned the connection for longer than 20ms


  1) test allows an alternate timeout to be passed for an individual `checkout` call (Ecto.Integration.SandboxTest)
     integration_test/sql/sandbox.exs:131
     Assertion with =~ failed
     code: log =~ ~r"timed out.*20ms"
     lhs:  ""
     rhs:  ~r/timed out.*20ms/
     stacktrace:
       integration_test/sql/sandbox.exs:143: (test)

...................

Finished in 11.2 seconds
256 tests, 1 failure, 30 skipped
```

It also looks like this happens on Travis, too.

I'm guessing it raises the `DBConnection.OwnershipError` before logging the message so there's a race condition unless we synchronize in the fun passed to `capture_log`?

Another time, I got this failure:

```
➜  ecto git:(myron/support-pool-timeout) ✗ mix test.adapters
==> Running tests for MIX_ENV=mysql mix test
Excluding tags: [:array_type, :read_after_writes, :uses_usec, :uses_msec, :returning, :strict_savepoint, :create_index_if_not_exists, :modify_column, :transaction_isolation, :rename_column, :upsert_all, :with_conflict_target]

...........................................................................................................................................................................................................
18:47:05.315 [error] Mariaex.Protocol (#PID<0.204.0>) disconnected: ** (DBConnection.ConnectionError) owner #PID<0.956.0> timed out because it owned the connection for longer than 20ms


  1) test allows an ownership timeout to be passed for an individual `checkout` call (Ecto.Integration.SandboxTest)
     integration_test/sql/sandbox.exs:131
     ** (exit) exited in: GenServer.call(#PID<0.957.0>, {:checkout, #Reference<0.0.3.2164>, true, 15000}, 5000)
         ** (EXIT) no process
     stacktrace:
       (db_connection) lib/db_connection/ownership/proxy.ex:32: DBConnection.Ownership.Proxy.checkout/2
       (db_connection) lib/db_connection.ex:712: DBConnection.checkout/2
       (db_connection) lib/db_connection.ex:619: DBConnection.run/3
       (db_connection) lib/db_connection.ex:515: DBConnection.execute/4
       (ecto) lib/ecto/adapters/sql.ex:244: Ecto.Adapters.SQL.sql_call/6
       (ecto) lib/ecto/adapters/sql.ex:420: Ecto.Adapters.SQL.execute_or_reset/7
       (ecto) lib/ecto/repo/queryable.ex:121: Ecto.Repo.Queryable.execute/5
       (ecto) lib/ecto/repo/queryable.ex:35: Ecto.Repo.Queryable.all/4
       (ex_unit) lib/ex_unit/capture_log.ex:66: ExUnit.CaptureLog.capture_log/2
       integration_test/sql/sandbox.exs:137: (test)

......................

Finished in 11.1 seconds
256 tests, 1 failure, 30 skipped

Randomized with seed 837370
```

I'm puzzled by this one--I don't have a theory about it yet.

Another time, it passed, but oddly enough, the captured log was printed to the console, too:

```
➜  ecto git:(myron/support-pool-timeout) ✗ mix test.adapters
==> Running tests for MIX_ENV=mysql mix test
Excluding tags: [:array_type, :read_after_writes, :uses_usec, :uses_msec, :returning, :strict_savepoint, :create_index_if_not_exists, :modify_column, :transaction_isolation, :rename_column, :upsert_all, :with_conflict_target]

...............................................................................................................................................................................................................
18:50:08.760 [error] Mariaex.Protocol (#PID<0.209.0>) disconnected: ** (DBConnection.ConnectionError) owner #PID<0.967.0> timed out because it owned the connection for longer than 20ms
...................

Finished in 11.0 seconds
256 tests, 0 failures, 30 skipped

Randomized with seed 473633
```

Suggestions?

myronmarston:
> The sleep needs to be inside the capture_log because the timeout occurs in another process, and that other process sends the log message. Hence the race condition where capture_log can be empty because the timeout/log happened before we entered capture_log.

Does this mean that the message is logged when `ownership_timeout` ms elapses, instead of it happening when the connection is attempted to be used?  I was assuming that it was like the error and would be triggered by the `TestRepo.all` call, but what you're saying makes sense.

> We can avoid this by sleeping for longer, see another example in this file where we timeout transactions with a `timer:sleep(1000)`.

I've done this, but it kinda bothers me to write such a slow test (due to the one second sleep) for something that is a relatively trivial thing.  When I maintain projects I care a lot about how fast the test suite runs and would try to find a way to avoid sleeping for so long, but obviously it's more important to not have flickering tests and here I'm happy to do whatever you prefer :).

myronmarston:
I've pushed an update reflecting your suggested changes, @fishcakez.

fishcakez:
@myronmarston it is ok to reduce the sleep timeouts but preferable to the same value. We can try 500 instead if you like. Any test that has a non-infinity sleep is likely to have a race condition. Unfortunately we cant remove sleeping here as even if we could monitor the sandbox process it would not mean the log is sent when it goes down.

myronmarston:
> @myronmarston it is ok to reduce the sleep timeouts but preferable to the same value. We can try 500 instead if you like.

Let's go with what I have for now.  If the timeout is going to be reduced in multiple tests it probably make sense to do that separate from this PR, anyway.

> Unfortunately we cant remove sleeping here as even if we could monitor the sandbox process it would not mean the log is sent when it goes down.

This kind of situation makes me wish there was a `capture_log/3` like so:

``` elixir
log = capture_log ~r/timed out/, 500, fn ->
  do_stuff()
end
```

The idea being that by passing a regex and timeout to `capture_log`, it would wait up to 500ms after `do_stuff` completes until `~r/timed out/` matches what it has captured in the logs.  Such an approach would obviate the need to sleep in a test like this, letting `capture_log` take care of monitoring what it has captured so you can more easily write tests for this kind of situation.

@josevalim would you consider that for inclusion in ExUnit?

fishcakez:
Thank you :heart:

PR #1741, opened 2016-10-12
Do not run transaction for an empty Multi

andrewtimberlake:
This change bypasses the transaction for a `Multi` that has no operations.

### DIFF
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -14,6 +14,8 @@ defmodule Ecto.Repo.Queryable do
     adapter.transaction(repo, opts, fun)
   end
 
+  def transaction(_adapter, _repo, %Ecto.Multi{operations:  []}, _opts),
+    do: {:ok, %{}}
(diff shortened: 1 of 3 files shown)


josevalim:
@michalmuskala please merge after you review it. :)

lib/ecto/repo/queryable.ex
```diff
@@ -14,6 +14,8 @@ defmodule Ecto.Repo.Queryable do
     adapter.transaction(repo, opts, fun)
   end
 
+  def transaction(_adapter, _repo, %Ecto.Multi{operations:  []}, _opts),
+    do: {:ok, %{}}
```
michalmuskala: I would prefer we do this in the `Ecto.Multi` module instead of here.

The `Multi` struct is defined as opaque, so we shouldn't destructure it outside the defining module.

Appropriate place would be here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/multi.ex#L378

josevalim:

michalmuskala:
Thanks @andrewtimberlake

PR #1742, opened 2016-10-12
Add Ecto.Multi.empty?/1

andrewtimberlake:
This adds a simple check for a `Multi` with no operations.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -353,6 +353,15 @@ defmodule Ecto.Multi do
   defp format_operation(other),
     do: other
 
+  @doc """
+  Determines if the `Ecto.Multi` is empty.
+
+  Returns `true` if `multi` is empty, otherwise `false`.
+  """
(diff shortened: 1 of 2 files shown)


michalmuskala:
I'm not really sure I like this. We should make sure multi doesn't run if empty (which is done by the other PR).

I'm afraid that this function will encourage code like:
```elixir
if Multi.empty?(multi) do
  {:ok, %{}}
else
  Repo.transaction(multi)
end
```

Which is bad, since it's completely meaningless (with the other PR merged). I struggle to find a use case that would go beyond that example.

andrewtimberlake:
That makes sense. I’m actually only using a check like that in my tests to make sure that a function that generates none or more commands actually generates none.

josevalim:
@andrewtimberlake then maybe `Ecto.Multi.to_list(multi) == []` is best?

andrewtimberlake:
Thanks

PR #1743, opened 2016-10-12
Correct configuration typo

MartinElvar:
So this doesn't seem to be a issue when running on local, as Ecto will fallback to localhost, if hostname is nil. But hopefully, now there won't be connection confusing, when using Ecto in docker or production, for people who copy pasted the configuration from the readme, as i did 😄 .


michalmuskala:
Thank you! ❤️

PR #1747, opened 2016-10-13
Fix assoc schema_prefix on through-associations

### LINKED ISSUE #1746: assoc doesn't respect @schema_prefix on through-associations
While **preloading** "through-associations" on a prefixed schema works fine, requesting them with **assoc** doesn't respect the @schema_prefix.

### Environment

* Elixir version (elixir -v): 1.3.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): MariaDB 
* Ecto version (mix deps): 2.0.5
* Database adapter and version (mix deps): mariaex 0.7.8
* Operating system: arch linux

### Current behavior
```elixir
defmodule Hotel do
  use Ecto.Schema
  @schema_prefix "br_dev"

  schema "hotel_hotels" do
    has_many :room_categories, Hotel.RoomCategory
    has_many :rooms, through: [:room_categories, :rooms]
  end
end

defmodule Hotel.RoomCategory do
  use Ecto.Schema
  @schema_prefix "br_dev"

  schema "hotel_room_categories" do
    belongs_to :hotel, Hotel
    has_many :rooms, Hotel.Room
  end
end

defmodule Hotel.Room do
  use Ecto.Schema
  @schema_prefix "br_dev"

  schema "hotel_rooms" do
    belongs_to :room_category, Hotel.RoomCategory
    has_one :hotel, through: [:room_category, :hotel]
  end
end

h = Repo.get Hotel, 1
Repo.preload(h, :rooms) # works fine, respects @schema_prefix
# [debug] QUERY OK source="hotel_room_categories" db=0.5ms SELECT h0.`id`, h0.`hotel_id`, h0.`hotel_id` FROM `br_dev`.`hotel_room_categories` AS h0 WHERE (h0.`hotel_id` = ?) ORDER BY h0.`hotel_id` [1]
# [debug] QUERY OK source="hotel_rooms" db=0.5ms decode=0.1ms SELECT h0.`id`, h0.`room_category_id`, h0.`room_category_id` FROM `br_dev`.`hotel_rooms` AS h0 WHERE (h0.`room_category_id` IN (?,?,?,?,?,?,?,?)) ORDER BY h0.`room_category_id` [4, 5, 6, 163, 164, 165, 166, 167]

Repo.all Ecto.assoc(h, :rooms) # fails for ignoring @schema_prefix on through-association
# [debug] QUERY ERROR source="hotel_rooms" db=0.1ms SELECT DISTINCT h0.`id`, h0.`room_category_id` FROM `hotel_rooms` AS h0 INNER JOIN `hotel_room_categories` AS h1 ON h1.`hotel_id` = ? WHERE (h0.`room_category_id` = h1.`id`) [1]
# ** (Mariaex.Error) (1146): Table 'br.hotel_rooms' doesn't exist

Repo.all Ecto.assoc(h, :room_categories) # but direct associations work fine with @schema_prefix
# [debug] QUERY OK source="hotel_room_categories" db=0.5ms SELECT h0.`id`, h0.`hotel_id` FROM `br_dev`.`hotel_room_categories` AS h0 WHERE (h0.`hotel_id` = ?) [1]

```
### Expected behavior
`Repo.all Ecto.assoc(h, :rooms)` should respect the "br_dev" schema_prefix.

shhavel:
Fix #1746
Added two schemes in integration tests: `PostWithPrefix`, `CommentWithPrefix` to test through association.
Tests contain generated SQL expectations for clarity.

### DIFF
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -153,6 +153,8 @@ defmodule Ecto.Integration.User do
     belongs_to :custom, Ecto.Integration.Custom, references: :bid, type: :binary_id
     many_to_many :schema_posts, Ecto.Integration.Post, join_through: Ecto.Integration.PostUser
     many_to_many :unique_posts, Ecto.Integration.Post, join_through: Ecto.Integration.PostUserCompositePk
+    has_many :posts_with_prefix, Ecto.Integration.PostWithPrefix
+    has_many :comments_with_prefix, through: [:posts_with_prefix, :comments_with_prefix]
     timestamps(type: :utc_datetime)
   end
 end
@@ -269,5 +271,36 @@ defmodule Ecto.Integration.PostUserCompositePk do
     belongs_to :user, Ecto.Integration.User, primary_key: true
     belongs_to :post, Ecto.Integration.Post, primary_key: true
     timestamps()
+  end
+end
+
+defmodule Ecto.Integration.PostWithPrefix do
+  @moduledoc """
+  This module is used to test:
+
+    * Ecto.assoc schema_prefix on through-associations
+
+  """
+  use Ecto.Schema
+  @schema_prefix "my_prefix"
+
+  schema "posts_with_prefix" do
+    belongs_to :user, Ecto.Integration.User
+    has_many :comments_with_prefix, Ecto.Integration.CommentWithPrefix
+  end
+end
+
+defmodule Ecto.Integration.CommentWithPrefix do
+  @moduledoc """
+  This module is used to test:
+
+    * Ecto.assoc schema_prefix on through-associations
+
+  """
+  use Ecto.Schema
+  @schema_prefix "my_prefix"
+
+  schema "comments_with_prefix" do
+    belongs_to :posts_with_prefix, Ecto.Integration.Post, foreign_key: :post_with_prefix_id
   end
 end
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -160,7 +160,7 @@ defmodule Ecto.Association do
   end
 
   def assoc_query(refl, t, query, values) do
-    query = query || %Ecto.Query{from: {"join expression", nil}}
+    query = query || %Ecto.Query{from: {"join expression", nil}, prefix: refl.queryable.__schema__(:prefix)}
 
     # Find the position for upcoming joins
     position = length(query.joins) + 1
--- a/integration_test/cases/assoc.exs
+++ b/integration_test/cases/assoc.exs
@@ -140,6 +140,25 @@ defmodule Ecto.Integration.AssocTest do
     assert u2.id == uid2
   end
 
+  test "assoc has_many with schema_prefix" do
+    u = TestRepo.insert!(%User{})
+    q = Ecto.assoc(u, :posts_with_prefix)
+    {sql, _} = Ecto.Adapters.SQL.to_sql(:all, TestRepo, q)
+
+    assert sql =~ ~r/ FROM ["`]my_prefix["`]\.["`]posts_with_prefix["`] /
+    assert [] == TestRepo.all(q)
+  end
+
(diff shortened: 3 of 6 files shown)


integration_test/support/schemas.exs
```diff
@@ -271,3 +271,50 @@ defmodule Ecto.Integration.PostUserCompositePk do
...
 end
+
+defmodule Ecto.Integration.Hotel do
+  @moduledoc """
+  This module is used to test:
+
+    * Ecto.assoc schema_prefix on through-associations
+
+  """
+  use Ecto.Integration.Schema
```
josevalim: I would like to avoid adding multiple schemas more. Could we do a single new schema? Maybe call it PostWithPrefix?

lib/ecto/association.ex
```diff
@@ -189,8 +189,9 @@ defmodule Ecto.Association do
     rewrite_ix = assoc.ix
     [assoc|joins] = Enum.map([assoc|joins], &rewrite_join(&1, rewrite_ix))
 
+    {_, schema} = assoc.source
     %{query | wheres: [assoc_to_where(assoc)|query.wheres], joins: joins,
-              from: merge_from(query.from, assoc.source), sources: nil}
+              from: merge_from(query.from, assoc.source), sources: nil, prefix: schema.__schema__(:prefix)}
```
josevalim: I think the prefix should be set here: https://github.com/shhavel/ecto/blob/51e69a35b036ebf3efc4abe625b8d829f86b5011/lib/ecto/association.ex#L163

Only on the right side of `||`. If a query is given, we should likely respect the prefix in the query because we are going to respect the from in that query too.

josevalim:
Thank you :heart: I have added some comments!

shhavel:
@josevalim Please review.

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -495,7 +495,7 @@ if Code.ensure_loaded?(Mariaex) do
         case elem(sources, pos) do
           {table, schema} ->
             name = String.first(table) <> Integer.to_string(pos)
-            {quote_table(prefix, table), name, schema}
+            {quote_table(prefix || (schema && schema.__schema__(:prefix)), table), name, schema}
```
josevalim: Generally we avoid calling schema functions in the adapter. The idea is that the adapter already receives all the data pre-processed. This would also break cases where the prefix is explicitly set to nil.

Is there any reason why you preferred to change it here instead of setting in the ecto/association.ex file when the query is built as previously mentioned?

shhavel: The query has only one prefix that is applied to all tables in SQL query,
this does not work for cross-schema queries, e.g.

```sql
SELECT * FROM "my_prefix"."comments" INNER JOIN "posts" ...
```
In this example `my_prefix` should be applied only to `comments`,
but if it will be set to query it will be set to both `comments` and `posts`.

I've create a test for this case - https://github.com/shhavel/ecto/blob/bdad903e3ad09458099b4a13d53ea4f92393acc8/integration_test/cases/assoc.exs#L162

josevalim: We don't support queries across schemas by design. If we are going to support it, we need to revisit all cases and scenarios, it should not be handled in the adapter because the adapter may not have enough information to decide which one to use.

shhavel: Ok, will do. Thanks

shhavel: Changed back to set prefix in `ecto/association.ex`.

josevalim:

josevalim:
@shhavel I had an idea of how we could use unit tests instead and I moved the schemas you wrote to unit tests in 57eb1c290c6f2491c7b5970c8be1d3edc67d9f8e.

PR #1752, opened 2016-10-19
Pass validators data to traverse_errors/3 in Ecto.Changeset

### LINKED ISSUE #1751: Additional options for all Ecto.Changeset validation errors
Hi guys,

[We](https://github.com/Nebo15/) love to provide our users very descriptive validation errors on Front-End of our projects, so our error structure [is complex](http://docs.apimanifest.apiary.io/#introduction/interacting-with-api/errors):
```
{
  "error": {
    "type": "validation_failed",
    "invalid": [
      {
        "entry_type": "json_data_proprty",
        "entry": "#/product_type",
        "rules": [
          {
            "rule": "string",
            "params": []
          },
          {
            "rule": "min:2",
            "params": [{"min": 2}]
          }
        ]
      }
    ],
    "message": "Validation failed. Return human-readable error message. You find all possible validation rules at https://docs.qbill.ly/#request-validators."
  }
}
```

And `Ecto.Changeset` doesn't provide sufficient data for such response structure, and setting `[message:..]` option for each changeset in each project is will mess everything (for example, you can forget to set it, and it's hard to refactor messages across all projects).

Our core idea is to provide PR that will add additional data to options (validation function and all validation options), that will be accessible in `traverse_errors/1` function and it should not break backwards compatibility.

Is it ok to work on this PR (would you like to see it?) or we should look for another way to solve our issue?

AndrewDryga:
Fixes #1751.

Open questions:
1. Should we add `include_validations?` option to `traverse_errors/3`? It can be better for backwards compatibility, but boolean flags is ugly. Also I can make another method for this purpose, but then lets figure out a name for it.
2. Any suggestions to validators naming or it's metadata?
3. Is it ok to add `data` and `format` to message options?
4. My PR doesn't look elegant, maybe there is a better way to do this?

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2084,15 +2084,16 @@ defmodule Ecto.Changeset do
       ...> end)
       %{title: ["should be at least 3 characters"]}
 
-  Optionally, you can pass function that accepts second argument `field` that will contain all
-  validation rules applied to it.
+  Optionally function can accept three arguments: `changeset`, `field` and `{msg, opts}`.
+  It is useful whenever you want to extract validations rules from `changeset.validations`
+  to build detailed error description.
   """
-  @spec traverse_errors(t, (error -> String.t) | (error, Keyword.t -> String.t)) :: %{atom => [String.t]}
-  def traverse_errors(%Changeset{errors: errors, changes: changes, types: types, validations: validations}, msg_func)
-      when is_function(msg_func, 1) or is_function(msg_func, 2) do
+  @spec traverse_errors(t, (error -> String.t) | (Changeset.t, atom, error -> String.t)) :: %{atom => [String.t]}
+  def traverse_errors(%Changeset{errors: errors, changes: changes, types: types} = changeset, msg_func)
+      when is_function(msg_func, 1) or is_function(msg_func, 3) do
     errors
     |> Enum.reverse()
-    |> merge_error_keys(msg_func, validations)
+    |> merge_error_keys(msg_func, changeset)
     |> merge_related_keys(changes, types, msg_func)
   end
 
@@ -2103,10 +2104,9 @@ defmodule Ecto.Changeset do
     end)
   end
 
-  defp merge_error_keys(errors, msg_func, validations) when is_function(msg_func, 2)  do
+  defp merge_error_keys(errors, msg_func, changeset) when is_function(msg_func, 3)  do
     Enum.reduce(errors, %{}, fn({key, val}, acc) ->
-      field = Keyword.get_values(validations, key)
-      val = msg_func.(val, field)
+      val = msg_func.(changeset, key, val)
       Map.update(acc, key, [val], &[val|&1])
     end)
   end
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -249,6 +249,7 @@ defmodule Ecto.Changeset do
 
   @relations [:embed, :assoc]
   @match_types [:exact, :suffix]
+  @validation_opts [:validation, :format, :data]
 
   @doc """
   Wraps the given data in a changeset or adds changes to a changeset.
@@ -487,7 +488,7 @@ defmodule Ecto.Changeset do
          {errors, valid?} = error_on_nil(kind, key, Map.get(changes, key, current), errors, valid?)
          {changes, errors, valid?}
        :invalid ->
-         {changes, [{key, {"is invalid", [type: type]}} | errors], false}
+         {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
      end}
   end
 
@@ -725,7 +726,7 @@ defmodule Ecto.Changeset do
                         name, current, required?, relation, opts) do
     current_changes = Map.get(changes, name, current)
     if required? and Relation.empty?(relation, current_changes) do
-      errors = [{name, {message(opts, :required_message, "can't be blank"), []}} | errors]
+      errors = [{name, {message(opts, :required_message, "can't be blank"), [validation: :required]}} | errors]
       %{changeset | errors: errors, valid?: false}
     else
       changeset
@@ -1297,7 +1298,7 @@ defmodule Ecto.Changeset do
           ensure_field_exists!(changeset, field),
           missing?(changeset, field),
           is_nil(errors[field]),
-          do: {field, {message, []}}
+          do: {field, {message, [validation: :required]}}
 
     case new_errors do
       [] -> %{changeset | required: fields ++ required}
@@ -1347,7 +1348,7 @@ defmodule Ecto.Changeset do
   @spec validate_format(t, atom, Regex.t, Keyword.t) :: t
   def validate_format(changeset, field, format, opts \\ []) do
     validate_change changeset, field, {:format, format}, fn _, value ->
-      if value =~ format, do: [], else: [{field, {message(opts, "has invalid format"), []}}]
+      if value =~ format, do: [], else: [{field, {message(opts, "has invalid format"), [validation: :format, format: format]}}]
     end
   end
 
@@ -1367,7 +1368,9 @@ defmodule Ecto.Changeset do
   @spec validate_inclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_inclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:inclusion, data}, fn _, value ->
-      if value in data, do: [], else: [{field, {message(opts, "is invalid"), []}}]
+      if value in data,
+        do: [],
+        else: [{field, {message(opts, "is invalid"), [validation: :inclusion, data: data]}}]
     end
   end
 
@@ -1389,7 +1392,7 @@ defmodule Ecto.Changeset do
   def validate_subset(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:subset, data}, fn _, value ->
       case Enum.any?(value, fn(x) -> not x in data end) do
-        true -> [{field, {message(opts, "has an invalid entry"), []}}]
+        true -> [{field, {message(opts, "has an invalid entry"), [validation: :subset, data: data]}}]
         false -> []
       end
     end
@@ -1410,7 +1413,8 @@ defmodule Ecto.Changeset do
   @spec validate_exclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_exclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:exclusion, data}, fn _, value ->
-      if value in data, do: [{field, {message(opts, "is reserved"), []}}], else: []
+      if value in data, do:
+        [{field, {message(opts, "is reserved"), [validation: :exclusion, data: data]}}], else: []
     end
   end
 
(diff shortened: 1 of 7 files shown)


lib/ecto/changeset.ex
```diff
@@ -487,7 +488,7 @@ defmodule Ecto.Changeset do
          {errors, valid?} = error_on_nil(kind, key, Map.get(changes, key, current), errors, valid?)
          {changes, errors, valid?}
        :invalid ->
-         {changes, [{key, {"is invalid", [type: type]}} | errors], false}
+         {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
```
AndrewDryga: Replace to "type"?

lib/ecto/changeset.ex
```diff
@@ -2076,33 +2084,36 @@ defmodule Ecto.Changeset do
...
       ...> end)
       %{title: ["should be at least 3 characters"]}
+
+  You can pass `true` to `include_validations?` argument if you want to receive validation rules in `opts`.
   """
-  @spec traverse_errors(t, (error -> String.t)) :: %{atom => [String.t]}
-  def traverse_errors(%Changeset{errors: errors, changes: changes, types: types}, msg_func)
-      when is_function(msg_func, 1) do
+  @spec traverse_errors(t, (error -> String.t), boolean) :: %{atom => [String.t]}
+  def traverse_errors(%Changeset{errors: errors, changes: changes, types: types}, msg_func, include_validations? \\ false)
```
AndrewDryga: I don't like this flag, but I need your approve that it won't break anything :)

lib/ecto/changeset.ex
```diff
@@ -1347,7 +1348,7 @@ defmodule Ecto.Changeset do
   @spec validate_format(t, atom, Regex.t, Keyword.t) :: t
   def validate_format(changeset, field, format, opts \\ []) do
     validate_change changeset, field, {:format, format}, fn _, value ->
-      if value =~ format, do: [], else: [{field, {message(opts, "has invalid format"), []}}]
+      if value =~ format, do: [], else: [{field, {message(opts, "has invalid format"), [validation: :format, format: format]}}]
```
josevalim: So adding the format here is going to pose problems. That's going to break most traverse_errors implementations out there because it expects something that can be converted to a string (and a regex cannot).

lib/ecto/changeset.ex
```diff
@@ -1367,7 +1368,9 @@ defmodule Ecto.Changeset do
   @spec validate_inclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_inclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:inclusion, data}, fn _, value ->
-      if value in data, do: [], else: [{field, {message(opts, "is invalid"), []}}]
+      if value in data,
+        do: [],
+        else: [{field, {message(opts, "is invalid"), [validation: :inclusion, data: data]}}]
```
josevalim: Same here.

lib/ecto/changeset.ex
```diff
@@ -1389,7 +1392,7 @@ defmodule Ecto.Changeset do
   def validate_subset(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:subset, data}, fn _, value ->
       case Enum.any?(value, fn(x) -> not x in data end) do
-        true -> [{field, {message(opts, "has an invalid entry"), []}}]
+        true -> [{field, {message(opts, "has an invalid entry"), [validation: :subset, data: data]}}]
```
josevalim: Same here.

lib/ecto/changeset.ex
```diff
@@ -1410,7 +1413,8 @@ defmodule Ecto.Changeset do
   @spec validate_exclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_exclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:exclusion, data}, fn _, value ->
-      if value in data, do: [{field, {message(opts, "is reserved"), []}}], else: []
+      if value in data, do:
+        [{field, {message(opts, "is reserved"), [validation: :exclusion, data: data]}}], else: []
```
josevalim: Same here.

lib/ecto/changeset.ex
```diff
@@ -1462,21 +1466,21 @@ defmodule Ecto.Changeset do
...
-    {message(opts, "should be %{count} character(s)"), count: value}
+    {message(opts, "should be %{count} character(s)"), count: value, validation: :length}
   defp wrong_length(:list, _length, value, opts), do:
-    {message(opts, "should have %{count} item(s)"), count: value}
+    {message(opts, "should have %{count} item(s)"), count: value, validation: :length}
 
   defp too_short(_type, length, value, _opts) when length >= value, do: nil
   defp too_short(:string, _length, value, opts), do:
-    {message(opts, "should be at least %{count} character(s)"), count: value}
+    {message(opts, "should be at least %{count} character(s)"), count: value, validation: :length_min}
```
josevalim: I think `:min_length` would be grammatically better? After all we say "minimum length"?

lib/ecto/changeset.ex
```diff
@@ -1576,9 +1580,13 @@ defmodule Ecto.Changeset do
...
     errors =
       case Map.fetch(changeset.params, error_param) do
-        {:ok, ^value} -> []
-        {:ok, _}      -> [{error_field, {message(opts, "does not match confirmation"), []}}]
-        :error        -> confirmation_missing(opts, error_field)
+        {:ok, ^value} ->
+          []
+        {:ok, _} ->
+          [{error_field,
+           {message(opts, "does not match confirmation"), [validation: :confirmation_doesnt_match]}}]
```
josevalim: I think this validation should be simply `:confirmation` and the one below should be `:required`?

josevalim:
Thank you @AndrewDryga for the quick turn around on this!

The pull request is mostly good but adding non-interpolatable data in the error message is a no-go from my side. I would be OK with excluding only the `:validation` key (which turns out we don't need to exclude anyway) but excluding keys like `:format` and `:data` can lead us to accidentally exclude valid user input and that can be pretty confusing.

I think the best way is for us to include the `:validation` type so you can at least cross-reference data with `changeset.validations`.

AndrewDryga:
@josevalim I guess we can accept `msg_fun` with different arity to make it easier to traverse `changeset.validations`. Quick sample:
```
  def traverse_errors(%Changeset{errors: errors, changes: changes, types: types, validations: validations}, msg_func)
      when is_function(msg_func, 1) or is_function(msg_func, 2) do
    errors
    |> Enum.reverse()
    |> merge_error_keys(msg_func, validations)
    |> merge_related_keys(changes, types, msg_func)
  end

  defp merge_error_keys(errors, msg_func, _) when is_function(msg_func, 1)  do
    Enum.reduce(errors, %{}, fn({key, val}, acc) ->
      val = msg_func.(val)
      Map.update(acc, key, [val], &[val|&1])
    end)
  end

  defp merge_error_keys(errors, msg_func, validations) when is_function(msg_func, 2)  do
    Enum.reduce(errors, %{}, fn({key, val}, acc) ->
      val = msg_func.(val, validations)
      Map.update(acc, key, [val], &[val|&1])
    end)
  end
```

josevalim:
@AndrewDryga we can do that but I would prefer to pass the field. The field is generally more useful and you will have access to the validations anyway as long as you have access to the changeset (which you must have prior to calling traverse_errors).

AndrewDryga:
@josevalim I've already got rid off any additions except `validation` field. But I propose to give a easy way to receive `changeset.validations` inside `msg_func/2`.

josevalim:
> But also I propose to give an easy way to receive changeset.validations inside msg_func/2.

Can you explain why? If we pass the field to `msg_func/2`, we pass information that is not available today and that will allow you to retrieve the validations easily, basically by calling `changeset.validations[field]`. That seems to be better than passing validations which is specific to this use case. What if someone needs the field in the future? Should we have `msg_func/3` then?

josevalim:
@AndrewDryga so just to be clear. I agree with you we need to change traverse_errors/2 to call msg_func/2, but I think we should change it to receive the `field` and not `validations`.

AndrewDryga:
Sorry for misunderstanding. I'll implement it.

AndrewDryga:
@josevalim Done.

But there is still an edge case, it would be impossible to tell which validation triggered error. Example:
```
changeset
|> validate_inclusion(:body, ["hola", "bonjour", "hallo"])
|> validate_inclusion(:body, ["hola", "bonjour"])
```
Validations will contain: ```inclusion: ["hola", "bonjour"], inclusion: ["hola", "bonjour", "hallo"]```. 
So whenever you have duplicate rules it would be hard to tell which one triggered error.

Should we deal with it?

josevalim:
@AndrewDryga yes, I am aware of such, although I couldn't think of a scenario where keeping it apart would be relevant? For inclusion, for example, they must be equal otherwise there are values that will never pass. If you reported "hallo" is allowed, it wouldn't actually be true.

lib/ecto/changeset.ex
```diff
@@ -1462,21 +1465,21 @@ defmodule Ecto.Changeset do
...
-    {message(opts, "should be %{count} character(s)"), count: value}
+    {message(opts, "should be %{count} character(s)"), count: value, validation: :length}
   defp wrong_length(:list, _length, value, opts), do:
-    {message(opts, "should have %{count} item(s)"), count: value}
+    {message(opts, "should have %{count} item(s)"), count: value, validation: :length}
 
   defp too_short(_type, length, value, _opts) when length >= value, do: nil
   defp too_short(:string, _length, value, opts), do:
-    {message(opts, "should be at least %{count} character(s)"), count: value}
+    {message(opts, "should be at least %{count} character(s)"), count: value, validation: :min_length}
```
josevalim: For this one and the following what if we do this: `validation: :length, min: min`?

I am thinking if you call it `validation: :min_length`, it will make your life harder later on when putting the validations together. Also, `:min` is an integer, which means we can interpolate it in the error message.

lib/ecto/changeset.ex
```diff
@@ -2076,23 +2083,34 @@ defmodule Ecto.Changeset do
...
     Enum.reduce(errors, %{}, fn({key, val}, acc) ->
       val = msg_func.(val)
       Map.update(acc, key, [val], &[val|&1])
     end)
   end
 
+  defp merge_error_keys(errors, msg_func, validations) when is_function(msg_func, 2)  do
+    Enum.reduce(errors, %{}, fn({key, val}, acc) ->
+      field = Keyword.get_values(validations, key)
+      val = msg_func.(val, field)
```
josevalim: This should be `msg_func.(key, val)`. You can call `Keyword.get_values(validations, key)` in the callback yourself.

josevalim: @AndrewDryga oh, that's true. Good call.

We should pass then `msg_func.(changeset, key, value)`. This way you should have everything you need to retrieve it. I don't want to pass validations because if people need more information in the future, like the changes field, we will end-up with `changes, validations, value` and whatever else they may need.

josevalim: So we could have msg_func/1 and msg_func/3.

lib/ecto/changeset.ex
```diff
@@ -2066,7 +2073,7 @@ defmodule Ecto.Changeset do
   error message as the changeset is traversed. The error message
   function receives an error tuple `{msg, opts}`, for example:
 
-      {"should be at least %{count} characters", [count: 3]}
+      {"should be at least %{count} characters", [count: 3, validation: :min_length]}
```
josevalim: This example is outdated now. :D

josevalim:
@AndrewDryga just one last tiny change and we are good to go! 🎉

AndrewDryga:
@josevalim Done 🎉 .

josevalim:

PR #1755, opened 2016-10-21
Typo in traverse_errors/2 example

AndrewDryga:
I think this is a long typo. Example should use accumulator to replace one all entries of options, and not only on the last one. And `acc` should throw warning that is not used in old example.


josevalim:

PR #1756, opened 2016-10-21
Teach `Ecto.DateTime.Utils.compare/2` about Elixir date types.

lmarlow:
Now works for `Date`, `DateTime`, `NaiveDateTime`, and `Time`. Also,
comparisons can be made across similar types:
- `Date` and `Ecto.Date`
- `DateTime`, `NaiveDateTime` and `Ecto.DateTime`
- `Time` and `Ecto.Time`

### DIFF
--- a/lib/ecto/date_time.ex
+++ b/lib/ecto/date_time.ex
@@ -71,9 +71,13 @@ defmodule Ecto.DateTime.Utils do
   Receives two datetimes and compares the `t1`
   against `t2` and returns `:lt`, `:eq` or `:gt`.
   """
-  def compare(%{__struct__: module} = t1, %{__struct__: module} = t2) do
-    {:ok, t1} = module.dump(t1)
-    {:ok, t2} = module.dump(t2)
+  def compare(t1, t2)
+  def compare(%{__struct__: module1} = t1, %{__struct__: module2} = t2)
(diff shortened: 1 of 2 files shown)


lmarlow:
I wasn't sure whether to go back and forth between `DateTime` and `NaiveDateTime`, but since `DateTime`s are always UTC at the moment I went for it.

Should `Ecto.DateTime.Utils.compare/2` handle `Date`, `Time`, `DateTime`, and `NaiveDateTime` at all? I added it because I upgraded to ecto 2.1.0-rc2 and switched my `Ecto.DateTime` fields to `:utc_datetime` and wanted an easy way to compare them without pulling in another library.

josevalim:
> since DateTimes are always UTC at the moment I went for it.

That's not true though if you are using an external library, so it is a dangerous precedent to set in Ecto.

Anyway, since Ecto 2.1 is migrating to the new types (which is exactly why you are having those problems), we don't plan on adding or changing functionality in Ecto.Date/Time/DateTime. If you need those functions for now and you don't want an external library, I would suggest to define them in your app. You will be able to remove them when a new Elixir version is out. :) Thank you.

lmarlow:
Sounds like a plan. Thanks for taking a look.

PR #1758, opened 2016-10-23
Add documentation and tests to use custom data type options in Ecto.Migration

ishikawa:
In [this topic](https://groups.google.com/forum/#!topic/elixir-ecto/uv73OIukcok), I found that we can use database specific options of field types in Ecto.Migration.

```elixir
create table(:posts) do
  add :title, :"VARCHAR(255) CHARACTER SET utf8mb4", null: false
  add :pub_time, :"timestamp WITH TIME ZONE"
end
```

If this behavior is not a kind of bug, I think it is helpful to add documentation and tests.


josevalim:

PR #1764, opened 2016-10-27
Add docs for composite primary_key

ssboisen:
`Ecto.Schema` was missing documentation for composite primary keys. I only found out about this feature because I searched google and eventually ended up on the [PR](https://github.com/elixir-ecto/ecto/pull/1210).


josevalim:

PR #1770, opened 2016-10-30
Fix cast_relation loaded struct when no association changes

### LINKED ISSUE #1761: cast_assoc has unexpected behavior when updating unrelated fields
### Precheck

* Do not use the issues tracker for help or support requests (try Stack Overflow, IRC or mailing lists, etc).
* For proposing a new feature, please start a discussion on [elixir-ecto](https://groups.google.com/forum/#!forum/elixir-ecto).
* For bugs, do a quick search and make sure the bug has not yet been reported.
* Finally, be nice and have fun!

### Environment

* Elixir version (elixir -v):1.3.2
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgres 9.5.0
* Ecto version (mix deps):2.0.4
* Database adapter and version (mix deps): postgrex 0.11.2
* Operating system: OSX

### Current behavior

Assume I have a schema like this:

```
  schema "foo" do
    field :myprop, :string
    has_many :bars, Bar, on_delete: :delete_all, on_replace: :delete

    timestamps
  end
```

And a changeset function that looks like this:

```
  def changeset(model, params) do
    model
    |> cast(params, @required_fields)
    |> cast_assoc(:bars)
  end
```

When I get a Foo without preloading `bars` and try to update only `myprop` it fails with a "Please preload your association error" as described below.

#### Command
```
Repo.get(Foo, id)
|> changeset(%{ "myprop" => "awesome" })
|> Repo.update!
``` 

#### Error

```
     ** (RuntimeError) attempting to cast or change association `bars` from `MyApp.Foo` that was not loaded. Please preload your associations before manipulating them through changesets
```

### Expected behavior

This error makes complete sense if I was trying to change `bars`. But if I'm not changing `bars` why should I still need to preload the association that I don't plan on changing?

I have worked around this by only calling `cast_assoc` if the changeset has a `bars` param, but I think the right long term answer here is for Ecto to only throw this error if the association parameter is actually set.

shhavel:
Fix #1761

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -684,11 +684,11 @@ defmodule Ecto.Changeset do
 
     on_cast  = Keyword.get_lazy(opts, :with, fn -> on_cast_default(type, related) end)
     original = Map.get(data, key)
-    current  = Relation.load!(data, original)
 
     changeset =
       case Map.fetch(params, param_key) do
         {:ok, value} ->
+          current  = Relation.load!(data, original)
           case Relation.cast(relation, value, current, on_cast) do
             {:ok, change, relation_valid?, false} when change != original ->
               missing_relation(%{changeset | changes: Map.put(changes, key, change),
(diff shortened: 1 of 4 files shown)


josevalim:

PR #1772, opened 2016-11-01
Convert map keys to int before sorting

### LINKED ISSUE #1771: `cast_embed` / `cast_assoc` sorts list elements wrong when given a map
As discussed with @ericmj on IRC this is probably, maybe a bug.

Given the following schema:

```elixir
defmodule Post do
  schema "posts" do
    embeds_many :categories, Category
  end

    struct
    |> cast_embed(:categories)
  end
end
```

With these params

```
posts[categories][1][title] = 'thing'
posts[categories][2][title] = 'other thing'
# ...
posts[categories][10][title] = 'totally different thing'
```

Correctly turns the categories into a list but it sorts them wrong. The indexes are strings and (of course) sorted like that. So it ends up like `1, 10, 11, 2, 3, ...` and so on.

In my app I manually convert the indexes to ints before casting but maybe ecto should do this by itself as it is already _trying_ to sort them (see https://github.com/elixir-ecto/ecto/blob/v2.0.5/lib/ecto/changeset/relation.ex#L84-L87)

mikker:
Fix #1771

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -83,7 +83,8 @@ defmodule Ecto.Changeset.Relation do
   def cast(%{cardinality: :many} = relation, params, current, on_cast) when is_map(params) do
     params =
       params
-      |> Enum.sort_by(&elem(&1, 0))
+      |> Enum.map(&key_as_int/1)
+      |> Enum.sort
       |> Enum.map(&elem(&1, 1))
     cast(relation, params, current, on_cast)
   end
@@ -323,6 +324,14 @@ defmodule Ecto.Changeset.Relation do
               "the foreign key value accordingly"
     end
   end
+
+  defp key_as_int({key, val}) when is_binary(key) do
+    case Integer.parse(key) do
+      {key, ""} -> {key, val}
+      _ -> {key, val}
+    end
+  end
+  defp key_as_int(key_val), do: key_val
 
   defp process_current(nil, _get_pks),
     do: %{}
--- a/test/ecto/changeset/embedded_no_pk_test.exs
+++ b/test/ecto/changeset/embedded_no_pk_test.exs
@@ -265,6 +265,23 @@ defmodule Ecto.Changeset.EmbeddedNoPkTest do
     assert changeset.valid?
   end
 
+  test "cast embeds_many with map with numbered string keys" do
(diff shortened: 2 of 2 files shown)


lib/ecto/changeset/relation.ex
```diff
@@ -83,11 +83,20 @@ defmodule Ecto.Changeset.Relation do
...
       params
-      |> Enum.sort_by(&elem(&1, 0))
+      |> Enum.sort_by(&key_as_int/1)
       |> Enum.map(&elem(&1, 1))
     cast(relation, params, current, on_cast)
   end
 
+  defp key_as_int({key, _}) when is_binary(key) do
+    try do
+      String.to_integer(key)
```
josevalim: We should use `Integer.parse` and return the key unchanged if we can't parse it. That's because there is no requirement for string keys to represent integers.

mikker: People on IRC suggested using something faster than `Integer.parse`. Is that a valid concern?

lib/ecto/changeset/relation.ex
```diff
@@ -324,6 +325,14 @@ defmodule Ecto.Changeset.Relation do
     end
   end
 
+  defp key_as_int({key, val}) when is_binary(key) do
+    case Integer.parse(key) do
+      {key, _} -> {key, val}
```
josevalim: It should be rather:

```elixir
case Integer.parse(key) do
  {key, ""} -> {key, val}
  _ -> {key, val}
end
```

Otherwise we will convert strings starting with numbers and discard the rest of the string.

lib/ecto/changeset/relation.ex
```diff
@@ -83,7 +83,8 @@ defmodule Ecto.Changeset.Relation do
   def cast(%{cardinality: :many} = relation, params, current, on_cast) when is_map(params) do
     params =
       params
-      |> Enum.sort_by(&elem(&1, 0))
+      |> Enum.map(&key_as_int/1)
+      |> Enum.sort
```
shhavel: I think it should be `Enum.sort_by(&elem(&1, 0))` istead of `Enum.sort`.
otherwise it will sort by values as well which is unnecessary (and slower), e.g.:

```elixir
iex> Enum.sort([%{"1" => "b"}, %{"1" => "a"}])
[%{"1" => "a"}, %{"1" => "b"}]
```

josevalim: Yup, `Enum.sort` is fine. Plus it will only sort by value if the keys are different.

josevalim:

mikker:
# Bucket list:

- [x] Got the hearts from Jose.

PR #1777, opened 2016-11-02
add options field to Index struct

vintikzzz:
I am currently developing adapter for [Cassandra DB](http://cassandra.apache.org/).
It may have numerous different options for index according to
its type. Here you can find an [example](https://docs.datastax.com/en/cql/3.3/cql/cql_using/useSASIIndex.html).
And I think it might be useful for other adapters.


josevalim:

PR #1782, opened 2016-11-05
add complex subtypes support

vintikzzz:
validation. So migrations could looks like:
```elixir
  add :value1, {:tuple, {:integer, :string, :float}}
  add :value2, {:tuple, {:integer, :string, {:tuple, {:integer, :integer}}}}
end
```

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -801,6 +801,10 @@ defmodule Ecto.Migration do
     validate_type!(subtype)
   end
 
+  defp validate_type!({type, subtype}) when is_atom(type) and is_tuple(subtype) do
+    for t <- Tuple.to_list(subtype), do: validate_type!(t)
+  end
+
   defp validate_type!(%Reference{} = reference) do
     reference
   end


josevalim:
@vintikzzz if the `{:tuple, {:integer, :string, :float}}` is meant to denote a tuple of integer, strings or floats, I would prefer for the second element to be a list because it may have N elements. So both `{:tuple, :integer}` and `{:tuple, [:integer, :string]}` would be supported perhaps?

vintikzzz:
@josevalim tuples by it's nature shouldn't have more then several elements, but technically you are right that it may have N elements. I have another example with maps:
```elixir
  add :value1, {:map, {:integer, :integer}}
end
```
In this case number of elements couldn't be more than two.
So I think maybe be it will be a good idea to pass both variants (tuple and list) through validation?

josevalim:
@vintikzzz sounds like an option, yes. Try to figure out what you want to accept on your side of things and then change Ecto accordingly. :)

josevalim:
To be clear, let's make Ecto only as extensible as it needs to be but not more. So if you need to support tuples with two elements only and lists, then that should be all.

vintikzzz:
@josevalim Thank you for clarification. I will be there with new pull request shortly.

PR #1783, opened 2016-11-05
add complex subtypes support

vintikzzz:
Just sequel of [1782](https://github.com/elixir-ecto/ecto/pull/1782)
validation. So migrations could look like:
```elixir
  add :value1, {:tuple, [:integer, :string, :float]}
  add :value2, {:tuple, [:integer, :string, {:tuple, [:integer, :integer]}]}
  add :value3, {:map, {:integer, :integer}}
end
```
For types with defined number of subtypes we use tuples (example `{:map, {:integer, :integer}}`)
For types with undefined number of subtypes we use lists (example `{:tuple, [:integer, :string, :float]}`)
If we want to make tuple with only one element we use just single atom (example `{:tuple, :integer}`

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -801,6 +801,14 @@ defmodule Ecto.Migration do
     validate_type!(subtype)
   end
 
+  defp validate_type!({type, subtype}) when is_atom(type) and is_list(subtype) do
+    for t <- subtype, do: validate_type!(t)
+  end
+
+  defp validate_type!({type, subtype}) when is_atom(type) and is_tuple(subtype) do
+    validate_type!({type, Tuple.to_list(subtype)})
+  end
+
   defp validate_type!(%Reference{} = reference) do
     reference
   end


lib/ecto/migration.ex
```diff
@@ -801,6 +801,14 @@ defmodule Ecto.Migration do
     validate_type!(subtype)
   end
 
+  defp validate_type!({type, subtype}) when is_atom(type) and is_list(subtype) do
+    for t <- subtype, do: validate_type!(t)
+  end
+
+  defp validate_type!({type, subtype}) when is_atom(type) and is_tuple(subtype) do
```
josevalim: Do you need to support tuples with more than two elements? If not, I would prefer to strictly check for that. We also need unit tests for this.

vintikzzz: Example migration for storing vector data requires 3 elements in a tuple:
```elixir
create table(:vector, primary_key: false) do
  add :start_coords, {:tuple, [:integer, :integer, :integer]}
  add :end_coords, {:tuple, [:integer, :integer, :integer]}
end
```

josevalim: That's a list though? And is that always 3 elements or it may be 2, 3, 4 and more?

vintikzzz: Yes, it is like a list. But it has always 3 elements and every element must has proper type (in my example `integer`).

vintikzzz: I think it might be a good option.

vintikzzz: @josevalim any ideas?

josevalim: @vintikzzz for all purposes it seems you don't have a use case for a list so I wouldn't support it until there is a need.

vintikzzz: @josevalim so we can switch back to variant introduced in [1782](https://github.com/elixir-ecto/ecto/pull/1782). Am I right?

josevalim: Yes, precisely. Sorry for the back and forth. :S

PR #1784, opened 2016-11-07
add complex subtypes support

vintikzzz:
Just sequel of [1782](https://github.com/elixir-ecto/ecto/pull/1782) and [1783](https://github.com/elixir-ecto/ecto/pull/1783)
validation. So migrations could look like:
```elixir
  add :value1, {:tuple, {:integer, :string, :float}}
  add :value2, {:tuple, {:integer, :string, {:tuple, {:integer, :integer}}}}
  add :value3, {:map, {:integer, :integer}}
end
```
If we want to make tuple with only one element we use just single atom (example `{:tuple, :integer}`

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -801,6 +801,10 @@ defmodule Ecto.Migration do
     validate_type!(subtype)
   end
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1785, opened 2016-11-07
Add cursor support

fishcakez:
PR to start discussion of streaming query results.

So far this branch implements `Repo.stream_all` that is a lazy version of `Repo.all`. `stream_all/2` returns a stream that can be reduced later inside a transaction that fetches rows in chunks so that not all rows need to be kept in memory at a time. There is only a simple test so far while we discuss API and behaviour.


fishcakez:
Note that we can't use named queries here because we can't reset the query cache if a column type is altered.

fishcakez:
Brought back MySQL with support for streaming.

josevalim:
It looks good to me. My only question is: do we want to call it stream or stream_all? Do we plan to support streaming updates or another streaming operation? That would help decide the API. :)

fishcakez:
@josevalim I am not sure to be honest, I would like to know also. I thought "stream all" would be the most feasible to implement for external adapters. I think if we go with `stream` we can go either way because `*_stream` would match the `*_all` API.

josevalim:
@fishcakez alright, let's go with `stream` because later we can go with either `update_stream` or `stream_update`.

fishcakez:
Renamed to `stream`, added and then reverted `delete/update_stream`, added some tests, callbacks, docs, rebased and squashed.

lib/ecto/adapters/sql.ex
```diff
@@ -78,6 +78,18 @@ defmodule Ecto.Adapters.SQL do
       end
 
       @doc false
+      def stream(repo, meta, query, params, process, opts) do
+        opts =
```
josevalim: Given this `opts` update is the same as in `execute`, I would probably move it to Ecto.Adapters.SQL and make sure it is a private function that can be shared between `execute_and_cache`, `execute_or_reset` and `prepare_stream`. I would also use pattern matching on maps instead of `Map.fetch`, something like: `%{sources: tuple} when tuple_size(...)`.

lib/ecto/repo.ex
```diff
@@ -428,6 +432,32 @@ defmodule Ecto.Repo do
...
+
+  See the "Shared options" section at the module documentation.
+
+  ## Example
+
+      # Fetch all post titles
+      query = from p in Post,
+           select: p.title
+      stream = MyRepo.stream(query)
+      MyRepo.transaction(fn() ->
```
josevalim: Is it required to run inside a transaction? If so, we should probably explicitly mention it in the documentation.

josevalim:
@fishcakez two comments and this is probably good to go! ❤️

lib/ecto/repo.ex
fishcakez: It is required for SQL adapters but others may not required it. I will add that to the docs though. We require it inside a transaction so we run the same code and get the same behaviour in and outside the sandbox. If we didn't use transaction it may be hard to understand when a connection is in use. It also means we can keep pooling completely separate from streaming.

lib/ecto/adapters/sql.ex
fishcakez: :+1:

fishcakez:
Handled comments and pointing at DBConnection package. Postgrex branch as been merged into master but a bootstrapping issue prevents a new release. Mariaex PR awaiting review.

fishcakez:
Fixed a bug when using drivers stream directly. Also mariaex cursor support is now in master. Just need to wait for the next release for mariaex and postgrex.

josevalim:
:heart:

michalmuskala:
As discussed over on IRC. Since we've introduced `Repo.query`, I think it would be appropriate to introduce `Repo.query_stream` for the raw SQL stream. 
That way, we can save users from learning about one more module - `Ecto.Adapters.SQL` - there's already a lot of those one has to look through for documentation of some Ecto things.

josevalim:
> As discussed over on IRC. Since we've introduced Repo.query, I think it would be appropriate to introduce Repo.query_stream for the raw SQL stream

We can wait. We introduced Repo.query because people were asking for it over and over again. Let's wait for the stream feature to get more mileage before we introduce a higher level API.

rhetzler:
https://github.com/elixir-ecto/ecto/issues/2274

PR #1792, opened 2016-11-11
WIP: Fix date casting from other structs

arcz:
Previously casting any struct to date resulted in no processing at all, throwing exceptions when making queries to database. The problem still exists for time, naive_datetime and utc_datetime. I can work on them as well, but first wanted to be sure that I'm not missing something.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -671,8 +671,6 @@ defmodule Ecto.Type do
       {:error, _} -> :error
     end
   end
-  defp cast_date(%{__struct__: _} = struct),
-    do: {:ok, struct}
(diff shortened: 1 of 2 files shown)


josevalim:
@arcz great catch. We should do this for the other types too.

arcz:
@josevalim I have fixed rest of types. `:time` has been changed to handle properly both microsecond as a single integer and as a tuple with precision. `:naive_datetime` is now using `cast_date` and `cast_time` for casting. `:utc_datetime` got fixed automatically after fixing `:naive_datetime` because it is using `cast_naive_datetime`.

josevalim:

PR #1793, opened 2016-11-12
Add type alter query cache tests

fishcakez:
This is work in progress to ensure our query cache recovers if a migration is done live. The tests are designed around how postgresql behaves but mysql may not have matching behaviour. However we need to investigate when or if the tests fail for mysql. Given that mariaex ignores all type information saved in the query struct it might be that migrations do not break the query cache and that it should fine to no longer store this information.

### DIFF
--- a/integration_test/sql/alter.exs
+++ b/integration_test/sql/alter.exs
@@ -45,15 +45,36 @@ defmodule Ecto.Integration.AlterTest do
     assert PoolRepo.all(values) == [1]
 
     assert :ok == up(PoolRepo, 20161112130000, AlterMigrationTwo, log: false)
-    assert_raise ArgumentError, ~r"stale type",
-      fn() -> PoolRepo.all(values) end
+
+    # optionally fail once with ArgumentError when preparing query prepared on
+    # another connection (and clear cache)
+    try do
+      PoolRepo.all(values)
+    rescue
+      err in [ArgumentError] ->
+        assert Exception.message(err) =~ "stale type"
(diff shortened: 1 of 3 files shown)


josevalim:
Beautiful! :heart:

integration_test/sql/alter.exs
```diff
@@ -0,0 +1,116 @@
...
+    # another connection (and clear cache)
+    try do
+      PoolRepo.update_all(values, [set: [value: 3]])
+    rescue
+      err in [ArgumentError] ->
+        assert Exception.message(err) =~ "stale type"
+        assert PoolRepo.update_all(values, [set: [value: 4]]) == {1, nil}
+    else
+      result ->
+        assert result == {1, nil}
```
fishcakez: If we hit this path it means the SQL adapter is encoding based on the Elixir type, and not the column type. This means the database will be casting from the type it receives to the type of the column. The MySQL adapter hits this path. Is this a bug or a feature?

fishcakez: If an alter column migration occurs and query is already prepared on the connection then MySQL will cast from the old type to the new type assuming it can. If we send an out of range integer for example an error will be returned by mariaex. Therefore I think we should remove the types from the mariaex query structs. This will also mean we never error on an alter column migration unless types are bad and then we want to error.

On maraiex master, but not any released version, we will error. I added this when doing cursor support. I'll merge this and track that issue in mariaex.

PR #1796, opened 2016-11-13
Improve Ecto.InvalidChangesetError message for embeds

wojtekmach:
Discussed on ML: https://groups.google.com/forum/#!topic/elixir-ecto/dB8Iu02YQO4

Before:

```
could not perform insert because changeset is invalid.

* Changeset changes

%{posts: [#Ecto.Changeset<action: :insert, changes: %{}, errors: [title: {"can't be blank", [validation: :required]}], data: #Ecto.ErrorTest.Post<>, valid?: false>, #Ecto.Changeset<action: :insert, changes: %{comments: [#Ecto.Changeset<action: :insert, changes: %{}, errors: [text: {"can't be blank", [validation: :required]}], data: #Ecto.ErrorTest.Comment<>, valid?: false>]}, errors: [title: {"can't be blank", [validation: :required]}], data: #Ecto.ErrorTest.Post<>, valid?: false>]}

* Changeset params

%{"name" => "", "posts" => [%{}, %{"comments" => [%{"text" => ""}], "title" => ""}]}

* Changeset errors

[name: {"can't be blank", [validation: :required]}]
```

After:

```
could not perform insert because changeset is invalid.

Changeset changes

    %{posts: [%{comments: [%{}]}]}

Changeset params

    %{"name" => "", "posts" => [%{"comments" => [%{"text" => ""}], "title" => ""}]}

Changeset errors

    %{name: [{"can't be blank", [validation: :required]}], posts: [%{title: [{"can't be blank", [validation: :required]}]}, %{comments: [%{text: [{"can't be blank", [validation: :required]}]}], title: [{"can't be blank", [validation: :required]}]}]}
```

(see an included temporary test that generates this message)

Note: I've initially used `traverse_error/2` to showcase deeply nested changesets. Problem with just using traverse_error that is we'd lose distinction between `changeset.errors` and `changeset.changes.foo.errors` which may or may not be a good thing. Alternatively, something close to what was originally proposed on ML could be used, e.g.: (not sure about the exact format)

```
Changeset errors

    [name: {"can't be blank", [validation: :required]}]
    posts:
      [title: {"can't be blank", [validation: :required]}] # changeset.changes.posts[0]

      [title: {"can't be blank", [validation: :required]}] # changeset.changes.posts[1]
      comments:
        [title: {"can't be blank", [validation: :required]}]
```

### DIFF
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -87,18 +87,28 @@ defmodule Ecto.InvalidChangesetError do
     """
     could not perform #{action} because changeset is invalid.
 
-    Changeset changes
+    Changes
 
-        #{inspect changes}
+    #{pretty changes}
 
-    Changeset params
+    Params
 
-        #{inspect changeset.params}
+    #{pretty changeset.params}
 
-    Changeset errors
+    Errors
 
-        #{inspect errors}
+    #{pretty errors}
+
+    Changeset
+
+    #{pretty changeset}
     """
+  end
+
+  defp pretty(term) do
+    inspect(term, pretty: true)
+    |> String.split("\n")
+    |> Enum.map_join("\n", &"    " <> &1)
   end
 
   defp extract_changes(%Ecto.Changeset{changes: changes}) do
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -81,22 +81,35 @@ defmodule Ecto.InvalidChangesetError do
   defexception [:action, :changeset]
 
   def message(%{action: action, changeset: changeset}) do
+    changes = extract_changes(changeset)
+    errors = Ecto.Changeset.traverse_errors(changeset, & &1)
+
     """
     could not perform #{action} because changeset is invalid.
 
-    * Changeset changes
+    Changeset changes
 
-    #{inspect changeset.changes}
+        #{inspect changes}
 
-    * Changeset params
+    Changeset params
 
-    #{inspect changeset.params}
+        #{inspect changeset.params}
 
-    * Changeset errors
+    Changeset errors
 
-    #{inspect changeset.errors}
+        #{inspect errors}
     """
   end
+
+  defp extract_changes(%Ecto.Changeset{changes: changes}) do
+    Enum.reduce(changes, %{}, fn({key, value}, acc) ->
+      Map.put(acc, key, extract_changes(value))
+    end)
+  end
+  defp extract_changes([%Ecto.Changeset{} = changeset | tail]),
+    do: [extract_changes(changeset) | extract_changes(tail)]
+  defp extract_changes(other),
+    do: other
 end
 
 defmodule Ecto.CastError do
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -120,7 +120,7 @@ defmodule Ecto.InvalidChangesetError do
     end)
   end
   defp extract_changes([%Ecto.Changeset{action: :delete} | tail]),
-    do: [extract_changes(tail)]
+    do: extract_changes(tail)
   defp extract_changes([%Ecto.Changeset{} = changeset | tail]),
     do: [extract_changes(changeset) | extract_changes(tail)]
   defp extract_changes(other),
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -113,9 +113,14 @@ defmodule Ecto.InvalidChangesetError do
 
   defp extract_changes(%Ecto.Changeset{changes: changes}) do
     Enum.reduce(changes, %{}, fn({key, value}, acc) ->
-      Map.put(acc, key, extract_changes(value))
+      case value do
+        %Ecto.Changeset{action: :delete} -> acc
+        _ -> Map.put(acc, key, extract_changes(value))
+      end
     end)
   end
+  defp extract_changes([%Ecto.Changeset{action: :delete} | tail]),
+    do: [extract_changes(tail)]
   defp extract_changes([%Ecto.Changeset{} = changeset | tail]),
     do: [extract_changes(changeset) | extract_changes(tail)]
   defp extract_changes(other),
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -87,7 +87,7 @@ defmodule Ecto.InvalidChangesetError do
     """
     could not perform #{action} because changeset is invalid.
 
-    Changes
+    Applied changes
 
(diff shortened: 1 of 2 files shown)


michalmuskala:
My only problem with this is that the values we show here are different from the fields in the changeset itself. I worry that this could be confusing. Maybe we should somehow make it more obvious that it's not what is actually stored in the changeset as the data, but rather it's logical value/interpretation.

ericmj:
It's a bit redundant to include "Changeset". What about removing it from each heading like this:

```
could not perform insert because changeset is invalid.

Changes

    %{posts: [%{comments: [%{}]}]}

Params

    %{"name" => "", "posts" => [%{"comments" => [%{"text" => ""}], "title" => ""}]}

Errors

    %{name: [{"can't be blank", [validation: :required]}], posts: [%{title: [{"can't be blank", [validation: :required]}]}, %{comments: [%{text: [{"can't be blank", [validation: :required]}]}], title: [{"can't be blank", [validation: :required]}]}]}
```

This way I think it will also be less associated with the exact changeset fields. Do you think that helps @michalmuskala?

wojtekmach:
Another solution is to include the Changeset in the exception message (although we rarely if ever do that in other exceptions) although that might be a bit too redundant:

preview

josevalim:
@michalmuskala @wojtekmach how should we proceed here then? :)

wojtekmach:
@michalmuskala ping :) I like @ericmj proposal the most, I also provided an alternative (that would be my 2nd choice)

michalmuskala:
I think I like the solution with printing the whole changeset too. It might be helpful in some occasions to see the whole, unprocessed data, that we show currently.

ericmj:
> It might be helpful in some occasions to see the whole, unprocessed data, that we show currently.

In those cases it is easy to inspect the changeset. I think in the majority of cases the changeset is not interesting.

michalmuskala:
Yeah, that's true. I'm still worried that printing something under `Changes:` that is different from what is in `changeset.changes` will be confusing. I'm not sure how to make it "not confusing".

josevalim:
@michalmuskala that's a good point. The changeset may also have important information, such as which action is going to be used. Maybe what we could do to improve this is by adding `width: 80` when inspecting the changesets so at least we don't  have a huge garbled message?

wojtekmach:
@josevalim @ericmj @michalmuskala  wdyt about the following (I've added a commit)? The top level changeset.action was always displayed in the first sentence of the error, but for embedded changesets I can see how it could be useful to see the actions at the bottom:

zrzut ekranu 2016-11-25 o 10 11 29

fishcakez:
Not sure about this but could use :erlang.error/2 to put the changeset in
the stacktrace.

wojtekmach:
@fishcakez I tried that but it looks like it's not converting `\n` in exception message, and the changeset (without pretty printing) suffers from the same "garbled message" problem.

(thanks for the tip about `:erlang.error/2` though, I did not know about it! :))


zrzut ekranu 2016-12-01 o 00 19 54

fishcakez:
@wojtekmach I meant to call it with the InvalidChangeSet exception as first argument, not the message,

wojtekmach:
ugh, doh. let me double check

wojtekmach:
hey everyone, just checking-in (sorry for nagging!): we clearly have a few different ways to display this as shown in the screenshots above, is there one way we could all agree on? Personally I think I like the 2nd screenshot from the top the most (pretty printing data and changeset), but I think any of the ones shown above would be an improvement over the original message.

ericmj:
I agree with @wojtekmach, I prefer the 2nd screenshot.

josevalim:
I am ok with the second screenshot as long as we call the first entry "Applied changes".

lib/ecto/exceptions.ex
```diff
@@ -81,22 +81,45 @@ defmodule Ecto.InvalidChangesetError do
...
     """
   end
+
+  defp pretty(term) do
+    inspect(term, pretty: true)
+    |> String.split("\n")
+    |> Enum.map_join("\n", &"    " <> &1)
+  end
+
+  defp extract_changes(%Ecto.Changeset{changes: changes}) do
```
josevalim: If the changeset is marked as `action: :delete`, should we maybe remove it from showing?

michalmuskala:
I'm fine with going with any of the proposed solutions. They are all better from what we have right now.

lib/ecto/exceptions.ex
wojtekmach: sounds good :+1:

wojtekmach: @josevalim do you mean the "main" changeset is marked as delete or one of the embeds changesets? If the main changeset is marked as default, do you mean that we wouldn't show the last part:

```
     ** (Ecto.InvalidChangesetError) could not perform delete because changeset is invalid.

     Applied changes

         %{posts: [nil, %{}, %{title: "new"}, %{title: nil}, %{title: "new name"}]}

     Params

         %{"posts" => [%{"title" => "new"}, %{"id" => "2", "title" => nil},
            %{"id" => "3", "title" => "new name"}]}

     Errors

         %{posts: [%{}, %{}, %{},
            %{title: [{"can't be blank", [validation: :required]}]}, %{}]}

     (Changeset used to be here)
```

or something else?

josevalim: The problem is that, when we hide the changeset, we can no longer know if the changeset is going to be inserted/updated/deleted. This way, if something was meant to be deleted but we still include it, it can be confusing. So I thought we should hide the changes not for the main changeset but any of the children changeset that is marked as delete.

wojtekmach: @josevalim okay thanks for clarifying it

wojtekmach: fixed in ebdf13c4727a7cb00a3dd6be28d5cfd13997e3dd

lib/ecto/exceptions.ex
```diff
@@ -113,9 +113,14 @@ defmodule Ecto.InvalidChangesetError do
...
     Enum.reduce(changes, %{}, fn({key, value}, acc) ->
-      Map.put(acc, key, extract_changes(value))
+      case value do
+        %Ecto.Changeset{action: :delete} -> acc
+        _ -> Map.put(acc, key, extract_changes(value))
+      end
     end)
   end
+  defp extract_changes([%Ecto.Changeset{action: :delete} | tail]),
+    do: [extract_changes(tail)]
```
michalmuskala: This shouldn't be wrapped in a list.

wojtekmach: good catch, fixed

josevalim:

PR #1800, opened 2016-11-14
Update decimal to 1.3.1

arcz:
As of version 1.3 decimal includes [Decimal.to_float/1](https://hexdocs.pm/decimal/Decimal.html#to_float/1) which helps for example to serialize decimals as floats in json.


josevalim:

michalmuskala:
Having `~> 1.2` as a requirement should match 1.3 as well. Is there a reason 1.2 can no longer be used that we force the upgrade?

Issue #1799, opened 2016-11-14
Use iodata when building SQL queries


josevalim:
Let's do it by parts so we don't have one big PR. :)

I believe the adapter should still return a binary (because of query caching) but the building can be done with iodata and then we can call IO.iodata_to_binary once at the end.

michalmuskala:
It's funny that you mention it, since I was working on it some time ago, but didn't finish. I have regular query generation for postgres (excluding ddl).

Through my measurements I didn't notice any significant speed difference, though I assume the GC pressure would be much lower since it would be producing less garbage. It might be that I messed up with the measurements.

squaresurf:
Shoulds this still be open since #1802 is merged?

josevalim:
This is pending removing the IO data conversion from adapters and doing it in the connection. This will benefit particularly insert_all.

PR #1802, opened 2016-11-15
Use iolists to build queries

michalmuskala:
Related to https://github.com/elixir-ecto/ecto/issues/1799

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -80,7 +80,7 @@ if Code.ensure_loaded?(Postgrex) do
       end
     end
 
-    def execute(conn, sql, params, opts) when is_binary(sql) do
+    def execute(conn, sql, params, opts) when is_binary(sql) or is_list(sql) do
       query = %Postgrex.Query{name: "", statement: sql}
       opts  = [function: :prepare_execute] ++ opts
       case DBConnection.prepare_execute(conn, query, params, opts) do
@@ -585,7 +585,7 @@ if Code.ensure_loaded?(Postgrex) do
       current =
         case elem(sources, pos) do
           {table, schema} ->
-            name = String.first(table) <> Integer.to_string(pos)
+            name = [String.first(table), Integer.to_string(pos)]
             {quote_table(prefix, table), name, schema}
           {:fragment, _, _} ->
             {nil, [?f, Integer.to_string(pos)], nil}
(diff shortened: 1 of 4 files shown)


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -910,6 +910,24 @@ if Code.ensure_loaded?(Postgrex) do
       |> Enum.join(" ")
     end
 
+    defp intersperse([], _separator, _mapper), do: []
+    defp intersperse(list, separator, mapper) do
+      [^separator | list] = Enum.reduce(list, [], fn x, acc ->
+        [separator, mapper.(x) | acc]
+      end)
+      :lists.reverse(list)
```
fishcakez: Should not need to reverse when using iodata.

fishcakez:
Do we only want to call IO.iodata_to_binary when caching? We should benchmark but something like insert_all seems to gain the most from this change and is not cached.

lib/ecto/adapters/postgres/connection.ex
michalmuskala: That's true. It might be hard to remove the extra separator in that case, though since it will be buried deep inside the list (if we put it on the front) or at the very end if we put it on the back.

I presume manual recursion is the best solution here, isn't it?

michalmuskala:
I've updated both intersperse functions so they don't need reversing.

Probably the best solution would be to always return iodata and let the caller convert it to binary, if necessary. I didn't do this, as the tests rely on functions from `Connection` returning strings, and it's some work to rewrite them, so I let it be for now.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -178,60 +177,62 @@ if Code.ensure_loaded?(Postgrex) do
...
-        {"#{quote_name(field)} = $#{acc}", acc + 1}
-      end
+      {filters, _} = intersperse_reduce(filters, " AND ", 1, fn field, acc ->
+        {[quote_name(field), " = $", Integer.to_string(acc)], acc + 1}
+      end)
 
-      assemble(["DELETE FROM #{quote_table(prefix, table)} WHERE " <> Enum.join(filters, " AND "),
-                returning(returning)])
+      IO.iodata_to_binary(["DELETE FROM ", quote_table(prefix, table), " WHERE ",
+       filters, returning(returning)])
```
josevalim: Can we please aligning this under the `"` as before? This and the IO.iodata_to_binary calls above. :)

josevalim:
Just one tiny comment and it looks good to go. I believe that we should definitely consider not converting it to a binary in the connection now that @fishcakez mentioned insert_all. But before we tackle that, we probably need to bring MySQL up to date too.

michalmuskala:
Updated. 

It should be fine to do the iodata conversion outside of connection even if we don't bring MySQL up to date on this change - `IO.iodata_to_binary/1` is a noop if given already a binary.

josevalim:
> It should be fine to do the iodata conversion outside of connection even if we don't bring MySQL up to date on this change - IO.iodata_to_binary/1 is a noop if given already a binary.

I know but I would like to make the adapters have feature parity, otherwise only Postgres would benefit from the insert_all improvements while there is no need to.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -910,6 +910,26 @@ if Code.ensure_loaded?(Postgrex) do
       |> Enum.join(" ")
     end
 
+    defp intersperse(list, separator, mapper, acc \\ [])
```
josevalim: Should we call this one `intersperse_map`?

josevalim:
@michalmuskala is this ready for merging? :)

michalmuskala:
I'd like to do a bit more refactoring in mysql adapter. This should be finished today.

michalmuskala:
@josevalim I think this is ready

michalmuskala:
Let's merge this after https://github.com/elixir-ecto/ecto/pull/1846, I will rebase and update

josevalim:
@michalmuskala #1846 has been merged!

michalmuskala:
@josevalim rebased and updated

josevalim:
@michalmuskala please ship it :heart:

josevalim:
Yes because we are not leveraging that yet in all places we could.

PR #1804, opened 2016-11-16
Improve doc of example of Ecto.Schema

sotte:
The example did not really correspond with the explanation.
There was no `belongs_to` in the example. The wording was off a bit.
It was not clear which settings were referenced.

This PR should fix the issues.


josevalim:

PR #1806, opened 2016-11-17
Minor "Getting Started" docs fixups

asummers:
1) Removes some leading and trailing whitespace
2) Fixes a bug in the example where the code as written would only return the last error in the errors in the case of errors length > 1.


josevalim:

PR #1812, opened 2016-11-19
Fix creating table with just comment on columns

arcz:
Previously, exception was raised when putting comment just on columns without specifying table comment. Commenting just on table worked, but I added a test for that.

```
17:41:21.417 [info]  create table example
    :erlang.++("CREATE TABLE \"example\" (\"id\" serial, \"foo\" decimal, PRIMARY KEY (\"id\"))", ["COMMENT ON COLUMN \"example\".\"foo\" IS 'bar'"])
    (ecto) lib/ecto/adapters/postgres/connection.ex:717: Ecto.Adapters.Postgres.Connection.add_comments_for_columns/2
    (ecto) lib/ecto/adapters/postgres.ex:71: Ecto.Adapters.Postgres.execute_ddl/3
    (ecto) lib/ecto/migration/runner.ex:98: anonymous fn/2 in Ecto.Migration.Runner.flush/0
    (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto) lib/ecto/migration/runner.ex:96: Ecto.Migration.Runner.flush/0
    (stdlib) timer.erl:181: :timer.tc/2
    (ecto) lib/ecto/migration/runner.ex:27: Ecto.Migration.Runner.run/6
```

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -699,7 +699,6 @@ if Code.ensure_loaded?(Postgrex) do
     defp add_comment([query], ""), do: query
     defp add_comment(queries, comment), do: queries ++ [comment]
 
-
     defp comment_on(_database_object, _name, nil), do:  ""
     defp comment_on(:column, {table_name, column_name}, comment) do
       column_name = quote_table(table_name, column_name)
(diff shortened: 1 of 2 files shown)


josevalim:

PR #1818, opened 2016-11-22
Handle `:invalid` params in validate_acceptance

wojtekmach:
Previously, an error was raised:

     ** (BadMapError) expected a map, got: nil
     stacktrace:
       (stdlib) :maps.find("terms_of_service", nil)
       (elixir) lib/map.ex:145: Map.get/3
       (ecto) lib/ecto/changeset.ex:1467: Ecto.Changeset.validate_acceptance/3
       test/ecto/changeset_test.exs:897: (test)

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -879,6 +879,7 @@ defmodule Ecto.ChangesetTest do
   end
 
   test "validate_acceptance/3" do
+    # accepted
     changeset = changeset(%{"terms_of_service" => "true"})
                 |> validate_acceptance(:terms_of_service)
     assert changeset.valid?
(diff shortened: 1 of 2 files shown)


test/ecto/changeset_test.exs
```diff
@@ -874,13 +875,32 @@ defmodule Ecto.ChangesetTest do
...
+    changeset = changeset(%{})
+                |> validate_acceptance(:terms_of_service)
+    refute changeset.valid?
+    assert changeset.errors == [terms_of_service: {"must be accepted", []}]
+
+    # invalid params
+    changeset = changeset(:invalid)
+                |> validate_acceptance(:terms_of_service)
+    refute changeset.valid?
+    assert changeset.errors == []
```
wojtekmach: doube-checking: is skipping validation fine here, when params is already `:invalid`? this matches behavior of validate_change

wojtekmach:
if this looks good, a backport to 2.0.6 would be very appreciated :)

test/ecto/changeset_test.exs
michalmuskala: Yes, this means we're in `new` or `edit` so it's fine.

josevalim:

PR #1827, opened 2016-11-29
[WIP] Add support to views in migrations

eniolopes:
- [ ] Write integration tests
- [ ] Support different params used when creating views
- [ ] Support drop
- [ ] Support create_or_replace

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -683,10 +683,10 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     def execute_ddl({:create, %View{} = view}) do
-      {query, _params, _key} = Ecto.Query.Planner.prepare(view.query, :all, Ecto.Adapters.Postgres, 0)
+      {query, params, _key} = Ecto.Query.Planner.prepare(view.query, :all, Ecto.Adapters.Postgres, 0)
       query = Ecto.Query.Planner.normalize(query, :all, Ecto.Adapters.Postgres, 0)
               |> all
-      ["CREATE VIEW #{view.name} AS #{query}"]
+      {"CREATE VIEW #{view.name} AS #{query}", params}
     end
 
     def execute_ddl(string) when is_binary(string), do: string
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -589,7 +589,7 @@ if Code.ensure_loaded?(Postgrex) do
 
     # DDL
 
-    alias Ecto.Migration.{Table, Index, Reference, Constraint}
+    alias Ecto.Migration.{Table, Index, Reference, Constraint, View}
 
     @drops [:drop, :drop_if_exists]
 
@@ -680,6 +680,13 @@ if Code.ensure_loaded?(Postgrex) do
 
     def execute_ddl({:drop, %Constraint{} = constraint}) do
       "ALTER TABLE #{quote_table(constraint.prefix, constraint.table)} DROP CONSTRAINT #{quote_name(constraint.name)}"
+    end
+
+    def execute_ddl({:create, %View{} = view}) do
+      {query, _params, _key} = Ecto.Query.Planner.prepare(view.query, :all, Ecto.Adapters.Postgres, 0)
+      query = Ecto.Query.Planner.normalize(query, :all, Ecto.Adapters.Postgres, 0)
+              |> all
+      ["CREATE VIEW #{view.name} AS #{query}"]
     end
 
     def execute_ddl(string) when is_binary(string), do: string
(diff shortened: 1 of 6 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -128,6 +130,13 @@ defmodule Ecto.Adapters.SQL do
       ## Migration
 
       @doc false
+      def execute_ddl(repo, %View{}=definition, opts) do
```
eniolopes: I'm not sure this is desired, maybe there is a better way to handle this?

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -682,6 +682,13 @@ if Code.ensure_loaded?(Postgrex) do
       "ALTER TABLE #{quote_table(constraint.prefix, constraint.table)} DROP CONSTRAINT #{quote_name(constraint.name)}"
     end
 
+    def execute_ddl({:create, %View{} = view}) do
+      {query, params, _key} = Ecto.Query.Planner.prepare(view.query, :all, Ecto.Adapters.Postgres, 0)
```
josevalim: Maybe we can use `Ecto.Adapters.SQL.to_sql(:all, query)`?

eniolopes: @josevalim I wanted to use that initially, but that requires the application repo, which is not available here.

lib/ecto/migration/runner.ex
```diff
@@ -260,6 +261,19 @@ defmodule Ecto.Migration.Runner do
   defp command({:drop, %Constraint{} = constraint}),
     do: "drop constraint #{constraint.name} from table #{quote_name(constraint.prefix, constraint.table)}"
 
+  defp command({:create, %View{name: nil, query: nil}}),
+    do: raise ArgumentError, "a view must have a name and a query"
```
josevalim: There is no need for those checks. If the name or query is nil, you should check when you create the `View` struct in Ecto.Migration. From there on, you can assume both name and query are valid.

lib/ecto/adapters/sql.ex
eniolopes: The reason this is present here, is in case the query is using `^` to reference a variable outside of context, so I need to get params and pas it to `Ecto.Adapters.SQL.query!`

lib/ecto/adapters/postgres/connection.ex
josevalim: I see. I think the planner API is not public, so we cannot access it from adapters. We will need to call those functions at some point before from Ecto, probably in Ecto.Migration, since it would give the best stacktrace in case something goes wrong.

lib/ecto/adapters/sql.ex
josevalim: We will need to change `@conn.execute_ddl` contract then altogether. Something like:

```elixir
# TODO: Enforce the {sql, params} on Ecto 3.0
for entry <- List.wrap(@conn.execute_ddl(definition)) do
  case entry do
    {sql, params} -> Ecto.Adapters.SQL.query!(repo, sql, params, opts)
    sql -> Ecto.Adapters.SQL.query!(repo, sql, params, opts)
  end
end
```

lib/ecto/migration.ex
```diff
@@ -781,6 +797,13 @@ defmodule Ecto.Migration do
     Runner.flush
   end
 
+  @doc """
+  """
+  def view(view_name), do: %View{name: view_name}
```
josevalim: We should probably remove this clause if we don't don't support views without queries.

lib/ecto/adapters/postgres/connection.ex
eniolopes: Yeah, also wanted to that from there initially, but application repo isn't available there either. The other point here is that repo is used by `Ecto.Adapters.SQL.to_sql` is when the query is cached, which isn't really a use case for migrations, so maybe we can try to simplify the `to_sql` API.

Aside from that the only place in the stack that has repo available, is the `Ecto.Adapters.SQL.execute_ddl`.

josevalim: Makes sense. To be clear, we should keep the code as it is today, we just need to move it to outside of the adapters. The adapters should receive the already normalized query. The normalization should happen either in Ecto.Migration or Ecto.Migration.Runner. :)

lib/ecto/adapters/sql.ex
eniolopes: cool, will add it

lib/ecto/adapters/postgres/connection.ex
eniolopes: Great, totally agree. Will try to get `Ecto.Adapters.SQL.to_sql` to work without an application repo, and move this piece of logic to the `Ecto.Migration`

josevalim:
Ping. :)

josevalim:
Closing this. Please let us know if work is resumed on this branch, thank you.

PR #1833, opened 2016-12-05
[WIP] Allow cast_assoc to work inside embeds

michalmuskala:
The issue is that embeds don't have the `__meta__` field, so we want to
raise only for regular schemas.
This, unfortunately, creates another problem. We don't know if the embed
was just created or loaded from database and needs preloading the
association data. This is very risky to use as it may cause data loss
if used improperly.

---

The use case for associations inside embeds is NoSQL databases, where this seems like a reasonable thing to do.

If we're going to merge this, we need more tests around this functionality.

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -62,10 +62,14 @@ defmodule Ecto.Changeset.Relation do
     cardinality_to_empty(cardinality)
   end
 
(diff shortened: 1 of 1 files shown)


josevalim:
Sounds good to me.

michalmuskala:
Closing in favour of the more comprehensible https://github.com/elixir-ecto/ecto/issues/1837

PR #1840, opened 2016-12-08
Don't generate transaction functions if not supported

### LINKED ISSUE #1530: Don't generate Repo.transaction when not supported
Today ecto always generates the `transaction/2`, `in_transaction?/0` and `rollback/1` functions on repo, even when the adapter does not support them and does not implement the `Ecto.Adapter.Transaction` behaviour.

This has a very unfortunate effect when combined with the new xref warnings from 1.3, because it tells the user that the appropriate adapter functions are not defined, which is correct and expected.

We have two options:
- don't generate those functions at all (might be problematic because of `Ecto.Repo` behaviour
- generate functions that raise an error when the adapter does not support those.

When fixed, I would ask to backport this to the 1.0 branch, since this affects mostly MongoDB users (and that's the last version supported).

michalmuskala:
If the adapter doesn't define transaction/3 function, we won't generate any
transaction related functions in the repo module. This also marks the repo
transaction callbacks as optional.

Closes #1530

---

Should we test this? I'm not exactly sure how that would look like.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -121,17 +121,19 @@ defmodule Ecto.Repo do
         Supervisor.stop(pid, :normal, timeout)
       end
 
-      def transaction(fun_or_multi, opts \\ []) do
-        Ecto.Repo.Queryable.transaction(@adapter, __MODULE__, fun_or_multi, opts)
-      end
+      if function_exported?(@adapter, :transaction, 3) do
+        def transaction(fun_or_multi, opts \\ []) do
+          Ecto.Repo.Queryable.transaction(@adapter, __MODULE__, fun_or_multi, opts)
+        end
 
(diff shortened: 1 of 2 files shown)


josevalim:
I would add a test to test/ecto/repo_test.exs. You will have to build a fake minimum adapter without the transaction function. Btw, do you think we should change the place we check for `function_exported?(adapter, :transaction, 3)` to check that in the repo now?

michalmuskala:
@josevalim I'm not sure I understand:
> Btw, do you think we should change the place we check for function_exported?(adapter, :transaction, 3) to check that in the repo now?

Could you explain?

josevalim:
@michalmuskala nevermind. :) Just the test and we should be good to go!

josevalim:

PR #1843, opened 2016-12-09
Squashing an credo issue around using a cond with only true and 1 condition.

shdblowers:
This is more subjective than other credo issues, but I think it help improve readability of the logic, on the downside it decreases how modifiable the code is if more conditions were to be added in the future.


shdblowers:
There is code coverage around this function and tests are passing, so I'm more confident about this one than the parentheses on the if statement one :laughing:

josevalim:
And I like this too! :+1:

josevalim:

PR #1847, opened 2016-12-10
Extracting part of the logic of association struct to private function

shdblowers:
This function was flagged up by credo as being too complex, so I pulled some of the logic out into a private function.

I think this function could be broken up further, but I thought this was a good start.

I wasn't sure about the naming of the function, whether to name it `get_ref` / `get_reference` / `get_owner_key`, so just went with what the variable is called. Personally I prefer not to use abbreviations, so would go with one of the other two.

_**Note:**_ there are no tests covering this logic, I had a go at adding them, but didn't get very far so will need a helping hand on that.


shdblowers:
I think that ebert thing isn't 100% because the [report](https://ebertapp.io/github/elixir-ecto/ecto/reviews/109l) for this PR has 42 issues, 1 less than master.

josevalim:

PR #1851, opened 2016-12-13
Extract setting transaction isolation level logic to named function

shdblowers:
This change will squash a credo warning around nesting.

I chose not to bring in all of `opts` in the function as it is only being pased to `checkin/2` which ignores it anyway.

No tests covering this logic, but I am confident the change is safe.


josevalim:

PR #1854, opened 2016-12-14
Add new changeset action :ignore

shhavel:
Ecto mailing list discussion - https://groups.google.com/forum/#!topic/elixir-ecto/ScPo16tfllc

Workaround with `cast_assoc/3` to ignore association with empty parameters:

```elixir
defmodule Post do
  use Ecto.Schema
  import Ecto.Changeset

  schema "posts" do
    field :title, :string
    field :body, :string
    has_many :comments, Comment
  end

  def changeset(post, params) do
    cast(post, params, [:title, :body])
    |> cast_assoc(:comments, with: &Comment.changeset_ignored_without_required_fields/2)
  end
end

defmodule Comment do
  use Ecto.Schema
  import Ecto.Changeset

  schema "comments" do
    field :body, :string
    belongs_to :post, Post
  end

  def changeset_ignored_without_required_fields(comment, params) do
    cast(comment, params, [:body])
    |> maybe_ignore
  end

  defp maybe_ignore(%{changes: %{body: body}} = changeset) when is_nil(body) or body == "" do
    %{changeset | action: :ignored}
  end
  defp maybe_ignore(changeset) do
    changeset
  end
end
```

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -70,6 +70,25 @@ defmodule Ecto.Integration.RepoTest do
     assert post.inserted_at
   end
 
+  test "insert, update and delete with action :ignore" do
+    changeset = %{Post.changeset(%Post{}, %{title: "hello"}) | action: :ignore}
+    TestRepo.insert! changeset
+    assert [] == TestRepo.all(Post)
+
+    changeset = %{Post.changeset(%Post{}, %{title: "hello"}) | action: :ignore, valid?: false}
+    {:error, ^changeset} = TestRepo.insert changeset
+
+    post = TestRepo.insert!(%Post{title: "hello"})
+
+    changeset = %{Post.changeset(post, %{title: "title1"}) | action: :ignore}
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you @shhavel.

I am not sure if we want to support `action: :ignore` at the repository level or rather at the association level. Given we want to discard the changeset altogether in the mailing list example, I think this should be pushed to association/embed casting exclusively?

michalmuskala:
Yes, I agree with @josevalim. We should handle the `ignore` action only for associations. I believe this should be done in `Ecto.Changeset.Relation`, similar to the `replace` action.

shhavel:
OK, sorry, I need some time to understand how option `:on_replace` in `Ecto.Changeset.Relation` works to implement `:ignored` by analogy with it.

josevalim:
@shhavel no worries. :) Take your time and let us know if you have any questions. I will close this in case you still want to leave this branch around and start work on a new one (advised). :)

PR #1859, opened 2016-12-16
Remove broken example from Ecto.Query.select docs

wojtekmach:
Previously, an error was raised:

    ** (Postgrex.Error) ERROR 42P18 (indeterminate_datatype): could not
    determine data type of parameter $1

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -683,7 +683,7 @@ defmodule Ecto.Query do
       from(c in City, select: c) # returns the schema as a struct
       from(c in City, select: {c.name, c.population})
       from(c in City, select: [c.name, c.county])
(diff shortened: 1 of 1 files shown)


michalmuskala:
I winder - do we want to change the example or actually make it run, and it's some regression it doesn't.

josevalim:
I would remove the example because it is quite meaningless as is. :)

josevalim:
And as is, I mean, before this PR. :)

wojtekmach:
I thought it was kinda cool that you can invoke outside functions in the query, but I suppose the same can be achieved with assigning result to a variable and pining that variable :) I've also determined that the example was broken at v1.0.0. I'm gonna go ahead and remove this example in this PR.

josevalim:

PR #1860, opened 2016-12-17
Stop producing warning when DBConnection is not included in mix.exs

AndrewDryga:
The only changes is that I wrapped Sandbox code in `if Code.ensure_loaded?(DBConnection) do .. end` block, because DBConnection is an optional dependency and it can be not included in end-user app, that produces two of warnings (because behaviours is undefined):
```
==> ecto
Compiling 69 files (.ex)
warning: behaviour DBConnection undefined
  lib/ecto/adapters/sql/sandbox.ex:296

warning: behaviour DBConnection.Pool undefined
  lib/ecto/adapters/sql/sandbox.ex:363

Generated ecto app
```

### DIFF
--- a/lib/ecto/adapters/sql/sandbox.ex
+++ b/lib/ecto/adapters/sql/sandbox.ex
@@ -1,516 +1,518 @@
-defmodule Ecto.Adapters.SQL.Sandbox do
-  @moduledoc ~S"""
-  A pool for concurrent transactional tests.
+if Code.ensure_loaded?(DBConnection) do
+  defmodule Ecto.Adapters.SQL.Sandbox do
+    @moduledoc ~S"""
+    A pool for concurrent transactional tests.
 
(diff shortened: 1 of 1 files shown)


josevalim:
@AndrewDryga instead of wrapping the whole chunk, let's please just wrap the behaviour calls.

AndrewDryga:
Sure, give me a second.

lib/ecto/adapters/sql/sandbox.ex
```diff
@@ -409,7 +413,7 @@ defmodule Ecto.Adapters.SQL.Sandbox do
     def stop(owner, reason, {_conn_mod, conn_state, _in_transaction?}, opts) do
       opts[:sandbox_pool].stop(owner, reason, conn_state, opts)
     end
- end
+  end
```
AndrewDryga: Here was formatting typo, so I fixed it.

josevalim:

PR #1865, opened 2016-12-17
Promote Mnesia adapter

AndrewDryga:
Hello,

I've spent some time to create [Ecto 2.0 compatible adapter for Mnesia Erlang term database](https://github.com/Nebo15/ecto_mnesia). Right now it can be used as drop-in replacement with few drawbacks:
- Joins and some other SQL-stuff is not supported. It can be emulated, but emulating them may look like a shot in the foot.
- SQL sandbox is not supported (because there are no connection pool that is needed to work with it).
- No type casting (and probably we don't need it at all).
- Other small stuff listed in readme.md.

Code may be dirty, any help, suggestions or pedantic code review is greatly appreciated 💯 👍 .

PR can be merged whenever community decides that adapter is ready to be promoted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Database   | Ecto Adapter           | Dependency                   | Ecto 2.0 co
 :----------| :--------------------- | :----------------------------| :-------------------
 PostgreSQL | Ecto.Adapters.Postgres | [postgrex][postgrex]         | Yes
 MySQL      | Ecto.Adapters.MySQL    | [mariaex][mariaex]           | Yes
-Mnesia     | Ecto.Adapters.Mnesia   | [ecto_mnesia][ecto_mnesia]   | Yes
+Mnesia     | EctoMnesia.Adapter     | [ecto_mnesia][ecto_mnesia]   | Yes
 MSSQL      | Tds.Ecto               | [tds_ecto][tds_ecto]         | No
 SQLite3    | Sqlite.Ecto            | [sqlite_ecto][sqlite_ecto]   | No
 MongoDB    | Mongo.Ecto             | [mongodb_ecto][mongodb_ecto] | No
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ Database   | Ecto Adapter           | Dependency                   | Ecto 2.0 co
 :----------| :--------------------- | :----------------------------| :-------------------
 PostgreSQL | Ecto.Adapters.Postgres | [postgrex][postgrex]         | Yes
 MySQL      | Ecto.Adapters.MySQL    | [mariaex][mariaex]           | Yes
+Mnesia     | Ecto.Adapters.Mnesia   | [ecto_mnesia][ecto_mnesia]   | Yes
 MSSQL      | Tds.Ecto               | [tds_ecto][tds_ecto]         | No
 SQLite3    | Sqlite.Ecto            | [sqlite_ecto][sqlite_ecto]   | No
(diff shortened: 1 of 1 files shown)


michalmuskala:
That's a great work! 💯 👍 

I took the liberty to look through the library a bit and here are some quick thoughts:
* it doesn't seem you use the ecto integration tests - Ecto comes with an extensive tests suite for adapters to ensure they work correctly with various features. They are split into several parts so you can choose to only run those that you support (for example, no joins). Using the tests may be extremely helpful in finding some corner cases that don't work correctly.
* The modules are places is the `Ecto.Mnesia` namespace and the adapter is called `Ecto.Adapters.Mnesia` - it's a general convention in the community not to pollute namespace of other packages, even related. It saves trouble when the base package adds a module. I think using `EctoMnesia` as the base namespace, and probably the "raw" `EctoMnesia` module as the adapter would be a better choice.

josevalim:
Great job @AndrewDryga! I will be glad to merge this as soon as you get some integration tests going, to help ensure proper compatibility. :)

AndrewDryga:
@michalmuskala  Can you suggest any place where I can see how integrations tests are used with other adapters? Looks like it's not that easy to run them outside of Ecto repo.

michalmuskala:
`integration_tests/cases` are the generic code that does not rely on any adapter. 

It's later used to test both mysql and pg, you can use it the same way for your adapter. You can take a look at how `mongodb_ecto` uses the test (the exact test cases may be different, since there's a version difference, but the idea stays the same):
* configuration of the tests: https://github.com/michalmuskala/mongodb_ecto/blob/master/test/test_helper.exs
* running the selected tests from the integration suite: https://github.com/michalmuskala/mongodb_ecto/blob/master/test/ecto_test.exs

You can see something very similar for the pg adapter:
* configuration of the tests: https://github.com/elixir-ecto/ecto/blob/master/integration_test/pg/test_helper.exs
* running selected tests from the integration suite: https://github.com/elixir-ecto/ecto/blob/master/integration_test/pg/all_test.exs

AndrewDryga:
Situation is pretty bad. Tests are full of different primary keys cases, and `ecto_mnesia` supports only simplest case - primary key on first field.

josevalim:
@AndrewDryga we are also open to doing changes to the Ecto's test suite to make it more compatible with other adapters. For example, if we can normalize our tests to always have the primary key as first element except a handful of cases, that would be welcome.

For the other tests, they should either be in separate files or have tags, so you don't need to load the file or you can skip them with tags.

AndrewDryga:
To keep this PR up to date. Right most of Repo integration tests are passing and merged into muster, however there are plenty of important features (like schemaless queries, constraints and tables without PK) that needs to be implemented.

Test results can be found here: https://travis-ci.org/Nebo15/ecto_mnesia/jobs/193757914#L309

josevalim:
We will be glad to merge this when you believe it is ready. Just let us know! You can also list on EctoMnesia README what is supported and what isn't to avoid confusion.

josevalim:
Ping. Let us know if we should move forward or not. :)

AndrewDryga:
Pong. I'm still in process of passing more integration tests and refactoring, I will tell whenever I will feel like adapter is ready to be merged.

AndrewDryga:
I think now I'm ready. Many issues has been fixed (some reported by other developers), also I renamed adapter to stop violating Ecto module space.

Passing integration tests is still hard (preloads will be my next feature to support).

I hope adding it to Ecto docs will drive attention, and maybe some contributors :).

josevalim:

PR #1869, opened 2016-12-20
Fix loading when field is given with different representation

arcz:
Issue happens when type supports non strict dump like naive_datetime which can be passed without ms, changes are then returned as is instead being loaded to correct form with ms in this case. This results in mismatch when returned from insert/update and get for example. Included test illustrates that (previously would fail).


josevalim:
Thank you @arcz although I am not sure we should merge this. There is no guarantee the database actually stores microseconds and we will have to investigate why Ecto is adding microseconds when none were specified.

arcz:
Thanks for a quick response! Yeah, it might be that the problem is somewhere else. [Here](https://github.com/elixir-ecto/ecto/blob/d8efe1e92907b3fbaf9c26222e591a1e4de4af3b/lib/ecto/type.ex#L783) `naive_datetime` is always dumped to format with microseconds, regardless if they are set or not it matches. The same goes for `utc_datetime`.

josevalim:
Closing this in favor of general discussion in #1935.

Issue #1876, opened 2016-12-27
Incorrect return value on select when database column has a NULL value


venkatd:
### Environment
* Elixir version 1.4.0-rc.1
* Database PostgreSQL 9.5.2.0
* Ecto version 2.1
* Database adapter postgrex 0.13.0
* Operating system: OSX El Capitan 10.11.6

### Current behavior
I have a helper method for getting the value of a specific column.
```elixir
  def get_prop(repo, type, conditions, column) when is_list(conditions) do
    import Ecto.Query
    query = from type, where: ^conditions, select: ^[column]
    case repo.all(query)  do
      [] -> nil
      [nil] -> nil # @TODO: figure out why this is needed
      [%{^column => value}] -> value
      [_rec | _rest] -> raise "Ambiguous result for #{type} #{conditions}. More than 1 value found."
    end
  end
```

If I call `get_prop(MyRepo, User, :timezone)`, `repo.all` will return `[nil]`. The incorrect behavior happens when the value is actually `NULL` in the database. It works just fine if the value is present.

### Expected behavior
It should return a struct with a `nil` value like `[%{^column => nil}]` so that the 3rd pattern in the code can be matched.

If I instead use the `field` macro, this works properly:
```elixir
  def get_prop(repo, type, conditions, column) when is_list(conditions) do
    import Ecto.Query
    query = from row in type, where: ^conditions, select: field(row, ^column)
    case repo.all(query)  do
      [] -> nil
      [val] -> val
      [_val | _rest] -> raise "Ambiguous result for #{type} #{conditions}. More than 1 value found."
    end
  end
```

michalmuskala:
I tracked down the issue to this code https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/sql.ex#L576-L586

It collapses the `nil` values returned by the database, but it's not entirely clear to me what's the purpose of that.

\cc @josevalim

josevalim:
@michalmuskala if you are using joins, multiple columns being nil collapses to a nil record. We should maybe have different ASTs for the join case.

PR #1877, opened 2016-12-27
Add test for null return when selecting field

michalmuskala:
This is a failing test case for https://github.com/elixir-ecto/ecto/issues/1876

\cc @josevalim


josevalim:
Thank you! I have included this on the fix!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1883, opened 2016-12-30
Validate fields passed to Ecto.Changeset.change/2

### LINKED ISSUE #1856: Validate fields passed to Ecto.Changeset.change/2
Branching from https://github.com/elixir-ecto/ecto/issues/1855#issuecomment-267157584

We should check against the `types` field and raise on any fields that are not in types.

wojtekmach:
Fixes #1856

### DIFF
--- a/test/ecto/repo/autogenerate_test.exs
+++ b/test/ecto/repo/autogenerate_test.exs
@@ -8,6 +8,7 @@ defmodule Ecto.Repo.AutogenerateTest do
 
     @timestamps_opts [inserted_at: :created_on]
     schema "manager" do
+      belongs_to :company, Company
       timestamps updated_at: :updated_on, type: :utc_datetime
     end
   end
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -330,6 +330,7 @@ defmodule Ecto.Changeset do
   defp get_changed(data, types, old_changes, new_changes, errors, valid?) do
     Enum.reduce(new_changes, {old_changes, errors, valid?}, fn
       {key, value}, {changes, errors, valid?} ->
+        ensure_field_exists!(data, types, key)
         put_change(data, changes, errors, valid?, key, value, Map.get(types, key))
     end)
   end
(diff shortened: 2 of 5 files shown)


michalmuskala:
Maybe we should move the check to `put_change/7`? That way it would work for both `change/2` and `put_change/3` public functions. It seems strange to me one would check and the other wouldn't.

The `put_change/7` already receives the type, we could match on `nil` or use `Map.fetch/1` instead and match on the ok/error.

wojtekmach:
good point, I forgot about the `put_change`.

> we could match on nil or use Map.fetch/1 instead and match on the ok/error.

I wanted to reuse the `ensure_field_exists!` helper function to use the same error message (as it seems appropriate in this case). WDYT?

wojtekmach:
@michalmuskala should we add the same validation to `force_change` too?

michalmuskala:
Yes, we should change `force_change` too.

It's a good point about raising the same error in all places.

test/ecto/repo/autogenerate_test.exs
```diff
@@ -8,6 +8,7 @@ defmodule Ecto.Repo.AutogenerateTest do
 
     @timestamps_opts [inserted_at: :created_on]
     schema "manager" do
+      belongs_to :company, Company
```
wojtekmach: FYI: without it I got a test failure:

```
** (ArgumentError) unknown field :company_id for changeset on %Ecto.Repo.AutogenerateTest.Manager{__meta__: #Ecto.Schema.Metadata<:built, "manager">, created_on: nil, id: nil, updated_on: nil}
```

wojtekmach:
Ok, I think it's ready to go - let me know if anything else is missing!

test/ecto/repo/autogenerate_test.exs
wojtekmach: btw, looking at docs for [`has_one`](https://github.com/elixir-ecto/ecto/blob/v2.1.1/lib/ecto/schema.ex#L634:L635):

> The other schema often has a `belongs_to` field with the reverse association.

as in this case, if the other schema does not have the belongs_to, trying to `Changeset.change` FK will fail. Perhaps it's obvious but maybe additional docs would be nice (not sure how they'd look like though)

josevalim:
After reviewing this, I do not agree with this change. Every time we add those checks, we are removing legitimate use cases, such as storing related information in changes field to be used later on.

It is a tight rope to walk on but I think this one is too harmful. In cast it is less of an issue since we need the type anyway... but changes are just a map.

wojtekmach:
Fair enough, let me know if you ever want to revisit this addition. Happy new year! :)

PR #1884, opened 2016-12-30
Correctly list the available options for creating constraints

nathanl:
`:name` isn't really an option, and `:exclude` is, as shown in the examples. This was my mistake.


josevalim:

PR #1885, opened 2017-01-01
Allow migrations to log raw SQL if required

bradleybeddoes:
A small change but access to the raw SQL can be quite useful.

### DIFF
--- a/lib/ecto/migration/runner.ex
+++ b/lib/ecto/migration/runner.ex
@@ -15,8 +15,9 @@ defmodule Ecto.Migration.Runner do
   """
   def run(repo, module, direction, operation, migrator_direction, opts) do
     level = Keyword.get(opts, :log, :info)
-    log_sql = Keyword.get(opts, :log_sql, false)
-    args  = [self(), repo, direction, migrator_direction, level, log_sql]
+    sql = Keyword.get(opts, :log_sql, false)
+    log = [level: level, sql: sql]
+    args  = [self(), repo, direction, migrator_direction, log]
 
     {:ok, runner} = Supervisor.start_child(Ecto.Migration.Supervisor, args)
     metadata(runner, opts)
@@ -41,11 +42,11 @@ defmodule Ecto.Migration.Runner do
   @doc """
   Starts the runner for the specified repo.
   """
-  def start_link(parent, repo, direction, migrator_direction, level, log_sql) do
+  def start_link(parent, repo, direction, migrator_direction, log) do
     Agent.start_link(fn ->
       Process.link(parent)
       %{direction: direction, repo: repo, migrator_direction: migrator_direction,
-        command: nil, subcommands: [], level: level, log_sql: log_sql, commands: []}
+        command: nil, subcommands: [], log: log, commands: []}
     end)
   end
 
@@ -93,8 +94,8 @@ defmodule Ecto.Migration.Runner do
     commands  = if direction == :backward, do: commands, else: Enum.reverse(commands)
 
     for command <- commands do
-      {repo, direction, level} = repo_and_direction_and_level()
-      execute_in_direction(repo, direction, level, command)
+      {repo, direction, log} = runner_config
+      execute_in_direction(repo, direction, log, command)
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/mix/tasks/ecto.migrate.ex
```diff
@@ -66,6 +67,11 @@ defmodule Mix.Tasks.Ecto.Migrate do
         do: Keyword.put(opts, :log, false),
         else: opts
 
+    opts =
+      if opts[:quiet],
+        do: Keyword.put(opts, :log_sql, false),
+        else: opts
```
michalmuskala: Let's combine the two:
```elixir
    opts =
      if opts[:quiet],
        do: Keyword.merge(opts, [log: false, log_sql: false]),
        else: opts
```

lib/mix/tasks/ecto.rollback.ex
```diff
@@ -66,6 +67,11 @@ defmodule Mix.Tasks.Ecto.Rollback do
         do: Keyword.put(opts, :log, false),
         else: opts
 
+    opts =
+      if opts[:quiet],
+        do: Keyword.put(opts, :log_sql, false),
+        else: opts
```
michalmuskala: Similar here

lib/ecto/migration/runner.ex
```diff
@@ -214,9 +213,14 @@ defmodule Ecto.Migration.Runner do
...
   end
 
+  defp log_sql do
+    Agent.get(runner(), fn %{log_sql: log_sql} -> log_sql end)
+  end
+
   defp log_and_execute_ddl(repo, level, command) do
     log(level, command(command))
-    repo.__adapter__.execute_ddl(repo, command, @opts)
+    opts = [timeout: :infinity, log: log_sql]
```
michalmuskala: Let's read the `log_sql` value in `repo_and_direction_and_level/0` function and rename it to `runner_config()` and pass the configuration as arguments to this function.

lib/ecto/migration/runner.ex
```diff
@@ -42,11 +41,11 @@ defmodule Ecto.Migration.Runner do
   @doc """
   Starts the runner for the specified repo.
   """
-  def start_link(parent, repo, direction, migrator_direction, level) do
+  def start_link(parent, repo, direction, migrator_direction, level, log_sql) do
```
michalmuskala: Let's change the last argument to be `log` and contain a keyword `[level: level, sql: log_sql]`, I'm not a huge fan of multiple boolean flags passed around.

bradleybeddoes:
No worries, will get around to these fixes in the next day or so.

bradleybeddoes:
Back to you @michalmuskala

bradleybeddoes:
Any further thoughts on this PR?

lib/ecto/migration/runner.ex
```diff
@@ -94,8 +94,8 @@ defmodule Ecto.Migration.Runner do
     commands  = if direction == :backward, do: commands, else: Enum.reverse(commands)
 
     for command <- commands do
-      {repo, direction, level} = repo_and_direction_and_level()
-      execute_in_direction(repo, direction, level, command)
+      {repo, direction, log} = runner_config
```
josevalim: It must be `runner_config()` to avoid warnings on 1.4.

lib/ecto/migration/runner.ex
```diff
@@ -208,15 +208,15 @@ defmodule Ecto.Migration.Runner do
...
-    Agent.get(runner(), fn %{repo: repo, direction: direction, level: level} ->
-      {repo, direction, level}
+  defp runner_config do
+    Agent.get(runner(), fn %{repo: repo, direction: direction, log: log} ->
+      {repo, direction, log}
     end)
   end
 
-  defp log_and_execute_ddl(repo, level, command) do
+  defp log_and_execute_ddl(repo, [level: level, sql: sql], command) do
```
josevalim: This should either be a tuple or a map. We should not use lists and then pattern match on them with specific ordering because this means the keyword list `[sql: sql, level: level]` would not match and that would be counter-intuitive.

josevalim:
I have added some comments, sorry for the delay and thank you!

josevalim:

PR #1890, opened 2017-01-03
Patch 4

toraritte:
Thank you for all the great work!

### DIFF
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -389,7 +389,7 @@ defmodule Ecto do
   Ecto requires developers to specify the key `:ecto_repos` in their application
   configuration before using tasks like `ecto.create` and `ecto.migrate`. For example:
 
-      config :my_app, :ecto_repos, [MyApp.Repo]
+      config :my_app, ecto_repos: [MyApp.Repo]
 
       config :my_app, MyApp.Repo,
         adapter: Ecto.Adapters.Postgres,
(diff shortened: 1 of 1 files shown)


lib/ecto.ex
```diff
@@ -355,7 +355,7 @@ defmodule Ecto do
 
   ### Embeds
 
-  Ecto also supports embeds. While associations keep parent and child
+  Ecto also supports [embeds](https://docs.mongodb.com/manual/core/data-model-design/#data-modeling-embedding). While associations keep parent and child
```
josevalim: Embeds are not specific to mongo, so I don't think we should link exclusively to mongo. :)

josevalim:
I have added one comment, thank you!

lib/ecto.ex
toraritte: Yeah, I had my concerns about that too but I couldn't find any DB-agnostic description of it such as a wikipedia entry. Thanks again and happy new year!

lib/ecto.ex
```diff
@@ -389,7 +389,7 @@ defmodule Ecto do
   Ecto requires developers to specify the key `:ecto_repos` in their application
   configuration before using tasks like `ecto.create` and `ecto.migrate`. For example:
 
-      config :my_app, :ecto_repos, [MyApp.Repo]
+      config :my_app, ecto_repos: [MyApp.Repo]
```
josevalim: Btw, those two calls are equivalent. Both would work. :)

josevalim:
Merged one commit manually, thank you!

PR #1893, opened 2017-01-06
put_assoc with params given as string keyed maps does behave correctly

jfrolich:
When you provide put_assoc with a map that is given with string keys, the changes are not resolved when converting the changeset to a struct.

See provided test case.

### DIFF
--- a/test/ecto/changeset/has_assoc_test.exs
+++ b/test/ecto/changeset/has_assoc_test.exs
@@ -804,6 +804,21 @@ defmodule Ecto.Changeset.HasAssocTest do
     refute Map.has_key?(changeset.changes, :profile)
   end
 
+  test "put_assoc/4 with string map" do
+    ## test maps
+    base_changeset = Changeset.change(%Author{})
+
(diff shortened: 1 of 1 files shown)


michalmuskala:
I'm not sure I understand the purpose of this PR. Is this a feature request?

jfrolich:
Hey @michalmuskala, I thought this was a bug, are you not allowed to pass string keys to `put_assoc`?

jfrolich:
Hmm I think I know why this does not work, I should probably use cast_assoc. In that case would it be helpful to raise an error when a map is passed that is not a struct?

michalmuskala:
The problem is what should happen there? String keys would suggest the data is unprocessed - should we cast it? If yes, what changeset should be used?

`put_assoc` already accepts changesets that allows you to fully control it.

Maybe we should update the docs to make it clearer?

josevalim:
We definitely cannot support this. Our options is to either raise (which may be tricky) and definitely improve the docs.

Issue #1790, opened 2016-11-10
Changeset.Relation raises runtime error rather than InvalidChangesetError


adamboas:
### Environment

* Elixir version: 1.3.2:
* Database and version: PostgreSQL 9.5.4
* Ecto version (mix deps):2.0.2
* Database adapter and version (mix deps): postgrex 0.11.2
* Operating system: OS X Sierra

### Current behavior

If you provide invalid joins in your Changeset for a many_to_many relation it raises a RuntimeError

### Expected behavior

I would expect it to raise an Ecto.InvalidChangesetError like any other invalid Changeset. Perhaps I am missing something? Unfortunately this requires me to catch any runtime error in my processing stream rather than being able to deal with invalid data specifically.

josevalim:
Can you please include the full error, stacktrace and an example of the
changeset fields being set?



-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

adamboas:
The stacktrace is below. Unfortunately it doesn't dump the Changeset as it is just a RuntimeError.

The Question schema looks like this:
```
 schema "questions" do
    field :uuid, :string
    field :name, :string

    belongs_to :survey, WaffleEventImporter.Survey
    has_many :answers, Answer
    many_to_many :factors, WaffleEventImporter.Factor, join_through: "factors_questions"

    timestamps
  end
```

I took a look at the Changeset.Relation#on_replace where the below error message originates and it appears that this is intended behaviour.  I might be missing something but it seems like failing to save because of an invalid changeset should consistently raise Ecto.InvalidChangesetError and dump the changeset to assist with debugging.

```
 ** (RuntimeError) you are attempting to change relation :factors of
 WaffleEventImporter.Question, but there is missing data.
 
 If you are attempting to update an existing entry, please make sure
 you include the entry primary key (ID) alongside the data.
 
 If you have a relationship with many children, at least the same N
 children must be given on update. By default it is not possible to
 orphan embed nor associated records, attempting to do so results in
 this error message.
 
 It is possible to change this behaviour by setting `:on_replace` when
 defining the relation. See `Ecto.Changeset`'s section on related data
 for more info.
 
     (ecto) lib/ecto/changeset/relation.ex:193: Ecto.Changeset.Relation.on_replace/2
     (ecto) lib/ecto/changeset/relation.ex:302: Ecto.Changeset.Relation.reduce_delete_changesets/5
[...]

PR #1896, opened 2017-01-08
Improve error message for on_replace: :raise

bobbypriambodo:
There are two cases in my attention that is caused by the confusing error message: [this thread in Elixir Forum](https://elixirforum.com/t/updating-many-to-many-relation/3172) and #1790.

Using the [Assocs, embeds, and on replace](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-assocs-embeds-and-on-replace) section of `Ecto.Changeset` and this repo's tests as the basis, it seems to me that the only thing that causes this error is by either not specifying `:on_replace` (therefore defaulting to `:raise`) or explicitly setting `on_replace: :raise`.

This PR removes what I see as irrelevant hints and points out the fact that `:on_replace` is currently set to `:raise`. I might be wrong and missing something, so feedbacks are welcome! Also do tell if there are grammatical errors, English is not my first language 🙂

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -175,20 +175,16 @@ defmodule Ecto.Changeset.Relation do
   def on_replace(%{on_replace: :raise, field: name, owner: owner}, _) do
     raise """
     you are attempting to change relation #{inspect name} of
-    #{inspect owner}, but there is missing data.
+    #{inspect owner}, but the `:on_replace` option of this
+    relation is set to `:raise`.
 
-    If you are attempting to update an existing entry, please make sure
-    you include the entry primary key (ID) alongside the data.
+    By default, if you don't specify it, Ecto sets this option to
+    `:raise`, therefore not allowing you to change relations or
+    embeds through the parent record's changeset.
 
-    If you have a relationship with many children, at least the same N
-    children must be given on update. By default it is not possible to
-    orphan embed nor associated records, attempting to do so results in
-    this error message.
-
-    If you don't desire the current behavior or if you are using embeds
-    without a primary key, it is possible to change this behaviour by
-    setting `:on_replace` when defining the relation. See `Ecto.Changeset`'s
-    section on related data for more info.
+    It is possible to change this behaviour by setting `:on_replace`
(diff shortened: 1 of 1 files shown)


bobbypriambodo:
I just realized that @expilo (starter of the aforementioned Elixir Forum thread) has opened a similar PR: #1895. Let me know if I should close this.

lib/ecto/changeset/relation.ex
```diff
@@ -175,20 +175,16 @@ defmodule Ecto.Changeset.Relation do
...
     you are attempting to change relation #{inspect name} of
-    #{inspect owner}, but there is missing data.
+    #{inspect owner}, but the `:on_replace` option of this
+    relation is set to `:raise`.
 
-    If you are attempting to update an existing entry, please make sure
-    you include the entry primary key (ID) alongside the data.
+    By default, if you don't specify it, Ecto sets this option to
+    `:raise`, therefore not allowing you to change relations or
+    embeds through the parent record's changeset.
```
michalmuskala: This is not true. 
The error is only raised if data is missing, i.e. some of the objects that are currently associated were not accounted for in the params. It's not raised on every change.

bobbypriambodo: Do you mean on every change of the parent's changeset?

With this message, I meant to say that if you build a changeset of a record that is also changing its association/relation, then it is not allowed if you don't specify `:on_replace`. Is that incorrect? I might misuse some terms here, so pardon me if that's the case. Do you have suggestions on how to improve it?

lib/ecto/changeset/relation.ex
```diff
@@ -175,20 +175,16 @@ defmodule Ecto.Changeset.Relation do
   def on_replace(%{on_replace: :raise, field: name, owner: owner}, _) do
     raise """
     you are attempting to change relation #{inspect name} of
-    #{inspect owner}, but there is missing data.
+    #{inspect owner}, but the `:on_replace` option of this
+    relation is set to `:raise`.
```
josevalim: I think this is the essential change we should to this message, i will push something similar based on both PRs. Thank you! :heart:

PR #1911, opened 2017-01-14
Added tag for test that uses SQL.query call

AndrewDryga:
There are test that uses raw SQL call that is not suitable for non-sql based adapters. 

Tag will help to exclude it when running integration tests.


josevalim:
This test is in the wrong file. We should move it to https://github.com/elixir-ecto/ecto/blob/master/integration_test/sql/sql.exs. Can you please do it?

AndrewDryga:
Yes, sure.

AndrewDryga:
@josevalim done 👍

josevalim:

PR #1917, opened 2017-01-19
Tag test that requires support for schemas without PK

AndrewDryga:
There are some databases that require schemas to have a primary key. 

For this cases it’s useful to allow developers to skip this test.


josevalim:

PR #1919, opened 2017-01-20
Fix Ecto.Migration.drop documentation

mememori:
`Ecto.Migration.constraint/{2,3}` uses the second argument as the name of the constraint, so, the code that was in the `drop/1` documentation would return a `%Constraint{name: [name: "price_must_be_positive"]}` struct that would fail the operation

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -378,7 +378,7 @@ defmodule Ecto.Migration do
 
       drop index(:posts, [:name])
       drop table(:posts)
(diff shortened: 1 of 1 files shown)


mememori:
Since this is a simple documentation change, i did not bother to add a test to ensure that trying to use a proplist as the second arg for `{add,drop} constraint` raises an `ArgumentError`, but if you find such test to be of any use, i can update this PR with that test :grin:

josevalim:

PR #1926, opened 2017-01-24
Use the "postgres" database for create/drop database commands

### LINKED ISSUE #1925: Multiple concurrent migrations don't work due to "template1" database being used
### Environment

* Elixir version (elixir -v): 1.4.0
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6.1
* Ecto version (mix deps): 2.1.3
* Database adapter and version (mix deps): postgrex 0.13.0
* Operating system: OS X 10.12.2

### Current behavior

When doing `mix ecto.create`, a connection is opened to the `template1` database. However, if there is already another connection to `template1`, the database creation will fail:

```
** (Mix) The database for Envoy.Repo couldn't be created: ERROR 55006 (object_in_use): source database "template1" is being accessed by other users

There is 1 other session using the database.
```

To reproduce this, just do `psql -d template1` and leave that prompt open while doing `mix ecto.create` in another terminal window.

### Expected behavior

Creating the database should succeed no matter how many connections there are to `template1`. This is especially true for running multiple `mix ecto.create` tasks concurrently.

### Additional info

This is documented PostgreSQL behavior, as you can read here: https://www.postgresql.org/docs/current/static/manage-ag-templatedbs.html

The problem with this is, if you have multiple applications that have their database on the same database server, creating the databases will fail if you happen to do that concurrently. I have never seen this happen in production, but in theory it could.

We also have a Docker environment for running our microservices landscape on our local development machines. There, we execute migrations and database creation on startup. In that environment, we can see this behavior rather often, which is not ideal.

mmainz:
Formerly, `mix ecto.create` or `mix ecto.drop` used the "template1" database.
However, other processes are not able to create new databases as long as there
is someone connected to the "template1" database. To fix that, we are now
connecting to the "postgres" database, which is also what PostgreSQL tools like
`createdb` do.

Fixes #1925

### DIFF
--- a/lib/ecto/adapters/postgres.ex
+++ b/lib/ecto/adapters/postgres.ex
@@ -110,7 +110,7 @@ defmodule Ecto.Adapters.Postgres do
   def storage_up(opts) do
     database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
     encoding = opts[:encoding] || "UTF8"
-    opts     = Keyword.put(opts, :database, "template1")
+    opts     = Keyword.put(opts, :database, "postgres")
 
     command =
       ~s(CREATE DATABASE "#{database}" ENCODING '#{encoding}')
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1929, opened 2017-01-30
Add apply_action to changeset

chrismccord:
Useful for composition in Ecto.Multi's to apply
changes only when valid and return an error
tuple in the event of invalid changes.
Additionally, an action may also be provied to
represent the action being peformed from the
changes. This is useful when working with embedded
schemas where operations on the changeset are
committed manually without relying on a Repo to
reflect failed operations on the changeset by
updating the action field.

/cc @josevalim

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -251,6 +251,7 @@ defmodule Ecto.Changeset do
 
   @relations [:embed, :assoc]
   @match_types [:exact, :suffix]
+  @actions [:insert, :update, :delete, :replace]
 
   @doc """
   Wraps the given data in a changeset or adds changes to a changeset.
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @chrismccord! I am not a big fan of the name but :+1: for the idea. The reason the name doesn't excite me is because the current `apply_changes` returns a struct (always) and now we have a function with very similar structure but a different return type. Suggestions?

Also, shouldn't we expect the action as argument and add it to the changeset? If so, shouldn't we have something with action in the name? /cc @michalmuskala

chrismccord:
Alternative name proposal, `verify_changes` ? 

> Also, shouldn't we expect the action as argument and add it to the changeset?

The function already optionally accepts the action. Are you saying it should always be there?

josevalim:
Yes, it should always be there. We can add another API if we need it. I think I am ok with `verify` as an API but I would like @michalmuskala's opinion.

michalmuskala:
I have this exact function in my application, so it's definitely useful - especially when using ecto schemas for form-object-style thingy (we need a good name for the use case btw). 

I was thinking about this one the whole day and I'm torn on the name. On one hand I agree with the potential confusion concerns, that @josevalim raised, on the other hand the more I think about potential names, the more the name `apply_valid_changes` grows on me. I think `verify_changes` is a good compromise.

As to the action, I think we should enforce passing it. I see this function as a replacement for a Repo call when you're not passing your schema to the database. It will make things much simpler for people using it for forms - the `Phoenix.Form` implementation for changesets will only copy the errors over if the action is set - that's just one place that relies on this behaviour.

We should also probably enforce that the action is one of the allowed ones - as specified by the typespec.

josevalim:
The problem with `verify` though, now that I think about it, is that I don't expect it to return a struct out of it. We need something with the notion of applying or merging.

chrismccord:
`apply_action/2` ? Or would two `apply_*` function w/ different signatures be confusing? Maybe Another option is to introduce `apply_changes/2` which returns the 2-tuple, but then we have the same function returning different signatures, so 🤷‍♂️

josevalim:
I am actually :+1: for `apply_action/2`.

chrismccord:
Updated with `apply_action/2` and strict requirement of valid actions in `:insert, :update, :delete, :replace`.

lib/ecto/changeset.ex
```diff
@@ -1136,6 +1137,35 @@ defmodule Ecto.Changeset do
...
+  The action may be one of #{Enum.map_join(@actions, ", ", &"`#{inspect &1}`")}.
+
+  ## Examples
+
+      iex> {:ok, data} = apply_action(changeset, :update)
+
+      iex> {:error, changeset} = apply_action(changeset, :update)
+      %Ecto.Changeset{action: :update}
+  """
+  @spec apply_action(t, action) :: {:ok, data} | {:error, Ecto.Schema.t}
```
wojtekmach: I think the type spec should be:

```elixir
 :: {:ok, data} | {:error, t}
```

or (similar to apply_changes)

```elixir
 :: {:ok, Ecto.Schema.t | data} | {:error, t}
```

PS awesome addition!

chrismccord: good catch! Fixed!

josevalim:

PR #1938, opened 2017-02-04
Update examples for like & ilike

katbow:
The string being searched for using `like` or `ilike` must start and end with `%`.
For example, `"%Chapter%"` will work, but not `"Chapter%"` or `"%Chapter"`.

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -80,7 +80,7 @@ defmodule Ecto.Query.API do
   @doc """
   Searches for `search` in `string`.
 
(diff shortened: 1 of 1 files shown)


michalmuskala:
All three examples you posted are valid and have different semantics. The `%` character in `LIKE` expression can stand for arbitrary characters. So having 3 strings: `"Chapter 1."`, `"1. Chapter"` and `"1. Chapter 1."`, the first will match all of those, the second, the first one, and the last will match the second one.

josevalim:
Yup! Check the database docs for more information on the semantics, which changes between databases. Thank you!

Issue #1776, opened 2016-11-02
Tasks ran with MySQL adapter relies on mysql client binary


rbishop:
### Environment

* Elixir version (elixir -v): 1.3.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): MySQL 5.7
* Ecto version (mix deps): 2.0
* Database adapter and version (mix deps): Mariaex
* Operating system: Heroku Cedar-14 (applies to any OS though)

### Current behavior

When running tasks such as `ecto.create` and `ecto.migrate` an exception occurs. Following the stacktrace leads to `Ecto.Adapters.MySQL.storage_up/1` and `Ecto.Adapters.MySQL.storage_down/2`, which are delegating to `Ecto.Adapters.MySQL.run_with_mysql/2`.

`Ecto.Adapters.MySQL.run_with_mysql/2` shells out to the `mysql` client binary. The problem is that in production environments you typically won't have the MySQL client binary installed on your servers that Ecto will be running tasks from. The common case is that you will have the MySQL client development headers installed (this doesn't matter for us because Mariaex is an Elixir implementation of the MySQL protocol).

### Expected behavior

I would expect Ecto to use the Mariaex adapter to run these queries rather than shelling out to the `mysql` binary.

josevalim:
Yes, if someone wants to send a pull request, it is welcome. That's what we do for Postgres: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/postgres.ex#L189-L223

rbishop:
I'll take a stab at it. I just wanted to confirm this wasn't the preferred behavior first.

sstine:
asdasd
asdasd
asdasd

fishcakez:
@rbishop this would be helpful for me when working on Ecto. If you could find some time to pick this up I would be very grateful.

rbishop:
@fishcakez We just stopped using MySQL since Postgres is HIPAA certified on AWS RDS now, but I'll try and take care of this anyway. I will try my best in the coming weeks!

PR #1939, opened 2017-02-05
Use Mariaex instead of MySQL client to create & drop database (#1776)

jayjun:
Creating and dropping MySQL databases now use Mariaex instead of shelling out to `mysql`. See #1776 and MySQL error codes [here](https://dev.mysql.com/doc/refman/5.7/en/error-messages-server.html).

MySQL doesn't need a database specified in order to connect, nor comes installed with a default database. So it's helpful to pass `skip_database: true` when creating databases. The `:skip_database` flag [was added](https://github.com/xerions/mariaex/commit/650df25e1017803dd41baaf8e79037071613abe4#diff-6ab42e58f582f0431c9882ef92faba0aR31) in Mariaex v0.4.3.

As a sidenote, [`structure_load/2`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/mysql.ex#L252-L255) still shells out to `mysql` but the Postgres adapter [does the same](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/postgres.ex#L204-L207).

### DIFF
--- a/lib/ecto/adapters/mysql.ex
+++ b/lib/ecto/adapters/mysql.ex
@@ -131,17 +131,19 @@ defmodule Ecto.Adapters.MySQL do
   def storage_up(opts) do
     database = Keyword.fetch!(opts, :database) || raise ":database is nil in repository configuration"
     charset  = opts[:charset] || "utf8"
+    opts     = Keyword.put(opts, :skip_database, true)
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #1943, opened 2017-02-10
Fixup docs for schemaless changesets

jonleighton:
"However, changesets also run without a changeset" is a bit self-referential, so I've tried to distinguish between data in a struct and data in a map (which has no implied schema). Also, the "as a tuple instead of a struct" bit seems superfluous (it's like saying "pass a tuple as a tuple").


josevalim:
@jonleighton ohai!

Oops, the second "changeset" was meant to be a "schema". Your explanation looks great though! 

jonleighton:
ohai! :D I'm enjoying playing with Elixir and Ecto <3 <3

PR #1947, opened 2017-02-13
Support selecting into structs

michalmuskala:
This enables queries like `from(c in City, select: %Data{name: c.name})` to work, which might be especially interesting in case of complex joins or schemaless queries.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -892,6 +892,10 @@ defmodule Ecto.Integration.RepoTest do
 
   ## Query syntax
 
+  defmodule Foo do
+    defstruct [:title]
+  end
+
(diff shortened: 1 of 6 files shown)


integration_test/cases/repo.exs
```diff
@@ -892,6 +892,8 @@ defmodule Ecto.Integration.RepoTest do
 
   ## Query syntax
 
+  defstruct [:title]
```
josevalim: Let's please define a module because this is a bit "lost" in the middle of the file.

josevalim:
Just one tiny comment and this is good to go!

michalmuskala:
@josevalim updated

josevalim:

PR #1950, opened 2017-02-17
Raise error when precision is missing

shulhi:
Raise `ArgumentError` when `precision` is missing for `numeric` type.

Fix for #1946

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -601,7 +601,7 @@ defmodule Ecto.Migration do
     * `:null` - when `false`, the column does not allow null values
     * `:size` - the size of the type (for example the numbers of characters).
       Default is no size, except for `:string` that defaults to 255.
-    * `:precision` - the precision for a numeric type. Default is no precision
+    * `:precision` - the precision for a numeric type. Required when `scale` is specified.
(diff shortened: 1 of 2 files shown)


josevalim:
@shulhi instead of checking it on every adapter, what if we check this on `Ecto.Migration` when adding the column information? Thanks!

shulhi:
I fixed it so that the checking is done in `Ecto.Migration`. Thanks for the pointing it out.

lib/ecto/migration.ex
```diff
@@ -613,7 +613,20 @@ defmodule Ecto.Migration do
     add(column, :naive_datetime, opts)
   end
 
-  def add(column, type, opts) when is_atom(column) do
+  def add(column, type, opts) when is_atom(column) and type in [:decimal, :numeric] do
```
josevalim: I don't think we need to do it only for those types, it can be for any type. :) Before `validate_type!(type)` you could do this:

```elixir
if opts[:scale] && !opts[:precision] do
  raise ...
end
```

shulhi: Right, I thought `precision` and `scale` only make sense for numerical type. Let me fix it real quick.

josevalim:

Issue #1949, opened 2017-02-17
Ecto.Multi missing insert_or_update


smeevil:
Hello,

I just wanted to do an insert_or_update call in an Echo.Multi, but found it missing in action.
I was wondering if there is a specific reason for this or if it's just not implemented :)

Gerard.

josevalim:
No particular reason. A PR is welcome!

PR #1951, opened 2017-02-17
added insert_or_update/3 to Ecto.Multi

smeevil:
referencing https://github.com/elixir-ecto/ecto/issues/1949

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -239,6 +239,20 @@ defmodule Ecto.Multi do
   def update(multi, name, %Changeset{} = changeset, opts \\ []) do
     add_changeset(multi, :update, name, changeset, opts)
   end
+  
+  @doc """
+  Inserts or updates a changeset depending on whether the struct is persisted or not.
+
+  Accepts the same arguments and options as `c:Ecto.Repo.insert_or_update/2` does.
+  """
+  @spec insert_or_update(t, name, Changeset.t | Ecto.Schema.t, Keyword.t) :: t
+  def insert_or_update(multi, name, changeset_or_struct, opts \\ [])
+  def insert_or_update(multi, name, %Changeset{data: %{__meta__: %{state: :loaded}}} = changeset, opts) do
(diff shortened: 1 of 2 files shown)


lib/ecto/multi.ex
```diff
@@ -239,6 +239,20 @@ defmodule Ecto.Multi do
...
   end
+  
+  @doc """
+  Inserts or updates a changeset depending on whether the struct is persisted or not.
+
+  Accepts the same arguments and options as `c:Ecto.Repo.insert_or_update/2` does.
+  """
+  @spec insert_or_update(t, name, Changeset.t | Ecto.Schema.t, Keyword.t) :: t
+  def insert_or_update(multi, name, changeset_or_struct, opts \\ [])
+  def insert_or_update(multi, name, %Changeset{data: %{__meta__: %{state: :loaded}}} = changeset, opts) do
```
josevalim: This does not take into account when a struct is given. I think it would be best for us to simply delegate to Repo.insert_and_update as the logic is already implemented there.

michalmuskala: `insert_or_update` is only allowed to work with changesets https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L318-L322

josevalim: So we need to pattern match on `Changeset` on the clause below. :)

smeevil: ok will update, just a moment

josevalim:

neerfri:
I believe this change is missing on `master`.
I see the commit exists in `v2.1` branch but it was not released as part of `2.1.4`
Is there a reason for this?

/cc @josevalim

josevalim:
This commit is definitely on master:

    $ git branch --contains 3701f6eaefbffb91b53841082364e6818ab4d459
    * master

It is not released in 2.1.4 because the commit was backported to v2.1 AFTER v2.1.4 was released. It should be in the next 2.1 release.

neerfri:
👍 
I was looking for the wrong commit, sorry

Thanks!

PR #1960, opened 2017-02-23
Add queryable field to %Ecto.Association.NotLoaded

aditya7iyengar:
- This could be helpful when trying to get the queryable.struct type without hitting the database.
*It feels weird that we have to do a ```owner.preload(:association).assoc``` to get the struct_type as it is defined in `Ecto.Schema` in the application layer.*
Thoughts?

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -7,11 +7,12 @@ defmodule Ecto.Association.NotLoaded do
   The fields are:
 
     * `__field__` - the association field in `owner`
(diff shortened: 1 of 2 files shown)


michalmuskala:
Could you expand on your use case? I'm not sure I understand that?

Would it be solved by using the reflection API?
```elixir
Owner.__schema__(:assoc, :association).related
#=> struct of the association
```

aditya7iyengar:
It sure would be! Thanks @michalmuskala. Don't know how I missed that! 😅

PR #1963, opened 2017-02-24
make the `:otp_app` argument to a new Repo optional

talentdeficit:
if declaring repo options inline (via `using Ecto.Repo opts`) the
`:otp_app` option is unesscessary. using `Keyword.get/2` instead of
`Keyword.fetch!/2` returns `nil` if `:otp_app` is missing which
results in `[]` being returned from the `Application.get_env` call,
the same result as if `:otp_app` pointed to a missing config


josevalim:
This unfortunately has the downside we are no longer giving feedback to the ones that forget the key. So I would prefer to have an explicit why to opt-out. Why don't you have a name in your case?

talentdeficit:
closing this, see #1964

PR #1970, opened 2017-03-01
Support repo opts for Changeset.prepare_changes/2

mkarnebeek:
In the example at https://hexdocs.pm/ecto/Ecto.Changeset.html#prepare_changes/2 it executes an additional query for updating the comment count. In my case this repo is a multi-tenant database using the `prefix` option when calling `Repo` functions.

When building models, it feels incorrect to have the changeset method require a prefix:

```Elixir
defmodule TestApp.Comment do
  ...
  def changeset(comment, params, prefix) do
    comment
    |> cast(params, [:body, :post_id])
    |> prepare_changes(fn changeset ->
      assoc(changeset.data, :post)
      |> changeset.repo.update_all([inc: [comment_count: 1]], prefix: prefix)
      changeset
    end)
  end
end

%TestApp.Comment{}
|> TestApp.Comment.changeset(..., prefix)
|> TestApp.Repo.insert(prefix: prefix)

```

Instead, I propose:

```Elixir
defmodule TestApp.Comment do
  ...
  def changeset(comment, params) do
    comment
    |> cast(params, [:body, :post_id])
    |> prepare_changes(fn changeset ->
      assoc(changeset.data, :post)
      |> changeset.repo.update_all([inc: [comment_count: 1]], changeset.repo.opts)
      changeset
    end)
  end
end

%TestApp.Comment{}
|> TestApp.Comment.changeset(...)
|> TestApp.Repo.insert(prefix: prefix)

```

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -204,6 +204,7 @@ defmodule Ecto.Changeset do
     * `action`       - The action to be performed with the changeset
     * `types`        - Cache of the data's field types
     * `repo`         - The repository applying the changeset (only set after a Repo function is called)
+    * `opts`         - The opts given to the repo function applying the changeset (only set after a Repo function is called)
     * `empty_values` - A list of values to be considered empty
 
   """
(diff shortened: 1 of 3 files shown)


josevalim:
The best would be to use Ecto.Multi to build the operations you want to perform, this way you have all of the prefix information upfront and just call Repo.transaction at the end. I don't think we should add more repo specific options to the changeset. Thank you.

Issue #1959, opened 2017-02-22
Sql sandbox without leveraging savepoints


tlvenn:
Hi all,

Since quite some time, I am trying to help with plugging [CoackroachDB](www.cockroachdb.com) into Ecto. This database implements the postgres protocol so on theory, it's the same as integrating with Postgres.

Of course, the devil is in the detail but with the help of @fishcakez and the CockroachDB team, we have made some pretty [good progress ](https://github.com/cockroachdb/cockroach/issues/5582) and right now it pretty much works.

We have to use my fork of Postgrex for the time being but that should be solved pretty soon and apart from not supporting jsonb and arrays types, I believe pretty much else is working fine.

The big remaining caveat however is that CockroachDB does not support savepoints and after [discussing](https://github.com/cockroachdb/cockroach/issues/10735) it a bit with them, it does not look like they will be able to provide that feature any time soon.

I am honestly not very clear in which scenario savepoints help beyond what we get by opening and rollbacking a transaction for each test but even if it's a little bit more limited than the current Sandbox, I would like to explore how to provide an alternate version that does not use savepoints.

Would there be an easy way to pass a flag to the current Sandbox to do so or does this warrant a brand new Sandbox implementation ?

Can someone familiar with the inner working of the sandbox provides some pointers as how feasible this would be ?

Thanks in avance !

josevalim:
@tlvenn if you don't support savepoints, then you are unable to test anything that uses transactions inside the sql sandbox. How big of a problem do you think this is?

tlvenn:
I am not sure I follow what you are saying @josevalim. I am obviously missing something here but it looks like I gonna need a little bit more help, so please bear with me.

When a test run, the sandbox would begin a transaction (the outer transaction). Multi that are executed within that test will detect that a transaction has been started already and simply execute. If a transaction is requested by the service being tested, given that the sandbox is proxying the commands and an outer transaction already exist, it can noop on it.

At this stage, either all things work, the assertion are ok, the test is passing and therefore the sandbox rollbacks its outer transaction in some kind of teardown phase to leave the database in a clean state or an error is emitted at some point and the transaction is rollbacked just the same and consequently the test fails.

Nested transactions would require savepoints but otherwise I dont understand why we need them.

josevalim:
[...]

PR #1971, opened 2017-03-01
Make the Sandbox truly pluggable

tlvenn:
As discussed with @josevalim in #1959, this small PR removes the hardcoded references on `Ecto.Adapters.SQL.Sandbox` so that one can plug his own version of the sandbox without overriding the existing `Ecto.Adapters.SQL.Sandbox`.


lib/ecto/adapters/sql.ex
```diff
@@ -317,10 +317,9 @@ defmodule Ecto.Adapters.SQL do
     |> Keyword.put_new(:pool_timeout, @pool_timeout)
   end
 
-  defp normalize_pool(Ecto.Adapters.SQL.Sandbox),
-    do: DBConnection.Ownership
-  defp normalize_pool(pool),
-    do: pool
+  defp normalize_pool(pool) do
+    if function_exported?(pool, :unboxed_run, 2), do: DBConnection.Ownership, else: pool
```
josevalim: I don't think we should do this change. We only do it for backwards compatibility, your sandbox should require DBConnection.Ownership to be explicitly passed instead.

tlvenn: Ha I didn't know about that, I thought it was some implicit contract, reverted to the original code.

lib/ecto/adapters/sql.ex
```diff
@@ -348,11 +348,10 @@ defmodule Ecto.Adapters.SQL do
 
     # Check if the pool options should overridden
     {pool_name, pool_opts} =
-      case Keyword.fetch(opts, :pool) do
```
josevalim: I believe this case is outdated. I have remove it from master.

tlvenn: Ha thanks fixed.

tlvenn:
I have pushed a branch for my sandbox to test out how it works on its own.
> https://github.com/jumpn/cockroachdb_sandbox/tree/no-override

I had to provide `ensure_all_started` and `child_spec` and I am not entirely convinced I am doing it properly. Would very much appreciate some guidance there.

And beyond that, my sandbox does not work any more:

```
** (UndefinedFunctionError) function Jumpn.Repo.Pool.__pool__/0 is undefined (module Jumpn.Repo.Pool is not available)
    Jumpn.Repo.Pool.__pool__()
    lib/cockroachdb_sandbox.ex:586: CockroachDBSandbox.proxy_pool/1
    lib/cockroachdb_sandbox.ex:518: CockroachDBSandbox.checkout/2
    (db_connection) lib/db_connection.ex:919: DBConnection.checkout/2
    (db_connection) lib/db_connection.ex:741: DBConnection.run/3
    (db_connection) lib/db_connection.ex:1132: DBConnection.run_meter/3
    (db_connection) lib/db_connection.ex:584: DBConnection.prepare_execute/4
    (ecto) lib/ecto/adapters/postgres/connection.ex:86: Ecto.Adapters.Postgres.Connection.execute/4
```

For some reason, the repo arg being passed to `checkout` is `Jumpn.Repo.Pool` all of the sudden instead of the expected `Jumpn.Repo`

josevalim:
@tlvenn your pool should be `pool: DBConnection.Ownership` and set `ownership_pool: ToYourCurrentPool`.

tlvenn:
Ha make sense, i was a puzzled by that and decided to ignore it, I guess I should not have :)

Also i am curious as to how `__pool__` got injected into my repo, I dont understand how this is happening.

josevalim:
@tlvenn your pool should be `pool: DBConnection.Ownership` and your module is just conveniences for working with the pool, nothing more. :)

josevalim:
@tlvenn the `__pool__` configuration is injected into the repository by the adapter. There is an open issue to make the `__pool__` configuration dynamic and get rid of the `__pool__` compile-time injection.

tlvenn:
Regarding the `__pool__`, it is injected when you declare the following?
```
def __adapter__ do
        @adapter
      end
```

As for the sandbox, i am not sure I actually understand what you are suggesting. If I define the pool to use `DBConnection.Ownership` then all the checks we do in the PR will fail to detect we are using a sandbox. Am i missing something ?

tlvenn:
Solved the issue by changing back the `normalize_pool/1`  to:
```
defp normalize_pool(pool) do
   if function_exported?(pool, :unboxed_run, 2), do: DBConnection.Ownership, else: pool
end
```

Now my sandbox works properly.
So it seems this is needed for more than backward compatibility.

josevalim:
@tlvenn you are right. We need to figure out a better way to handle this then. I am going to merge your PR as is and then try to come up with a better approach than checking for unboxed_run on normalize_config.

tlvenn:
Thanks for the review @josevalim and have a great WE.

tlvenn:
The tests are failing with Travis but I cant seem to reproduce it, all the tests are passing on my computer. I have seen that error before but it disappeared after a `mix clean`. I dont really understand how this is happening given the scope of the changes I made.

The error:
```
** (UndefinedFunctionError) function Ecto.Adapters.SQL.Sandbox.ensure_all_started/2 is undefined or private
    (ecto) Ecto.Adapters.SQL.Sandbox.ensure_all_started([pool_timeout: 5000, timeout: 15000, loggers: [Ecto.LogEntry, {Ecto.Integration.Repo, :log, [:on_log]}], adapter: Ecto.Adapters.Postgres, url: "ecto://postgres:postgres@localhost/ecto_test", pool: Ecto.Adapters.SQL.Sandbox, ownership_pool: DBConnection.Sojourn], :temporary)
    (ecto) lib/ecto/adapters/postgres.ex:86: Ecto.Adapters.Postgres.ensure_all_started/2
    integration_test/pg/test_helper.exs:68: (file)
    (elixir) lib/code.ex:363: Code.require_file/2
    (elixir) lib/enum.ex:651: Enum."-each/2-lists^foreach/1-0-"/2
```

josevalim:
It is happening because `function_exported?` does not load the module. That said, the check you added to normalize_config will never pass, even for the sandbox, because the module has not been loaded yet. So we never rewrite the sandbox pool entry to DBConnection.Ownership.

tlvenn:
Hi @josevalim, can you help out on this one, I am not so intimate with how/when module loading works and how to best address this issue. Thanks in advance.

tlvenn:
I guess a `Code.ensure_loaded(pool)` does the trick

josevalim:

Issue #1832, opened 2016-12-05
Allow enabling extenstions using migrations


hauleth:
It would be nice to have `Ecto.Migration.extension` method which would allow us to run:

```elixir
defmodule MyApp.Migrations.EnableExtensions do
  use Ecto.Migration

  def change do
    create_if_not_exist extension("uuid-ossp")
  end
end
```

Or similar code. Similar to #1805

michalmuskala:
Extensions are very postgresql specific. I would personally rather see them in a separate package, something like `PostgresEcto` that could also include other postgresql specific extensions - macros wrapping fragments for some common functions, etc.

hauleth:
I can try to produce something like that.

josevalim:
Unfortunately it is not possible to implement something like the above in a third party project. In any case, I don't mind having Postgres specific behaviour in Ecto most times, specially this one which is a very minor feature.

hauleth:
@josevalim I can accept that challenge.

michalmuskala:
Maybe we should look into ways of making it possible for extensions to provide such behaviour?

josevalim:
@michalmuskala I would love to and tried a couple times in the past. The best solution I had in mind is to define one protocol per adapter and I am not confident it is the best way to go.

PR #1975, opened 2017-03-02
Add support for PostgreSQL create extension

roehst:
Hi, this is a pull request for implementing #1832 

This is my first PR to an open source project, so forgive me if the code is sub-standard. I tried to implement the same tests and functionalities as in other CREATE commands, but I am new to the code base so I might be missing something. Please give feedback!

Also, PostgreSQL supports two options in the CREATE EXTENSION command, namely version and old_version, which are not implemented here yet (but this is straight forward to do).

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -584,7 +584,7 @@ if Code.ensure_loaded?(Postgrex) do
 
     # DDL
 
-    alias Ecto.Migration.{Table, Index, Reference, Constraint}
+    alias Ecto.Migration.{Table, Index, Reference, Constraint, Extension}
 
     @drops [:drop, :drop_if_exists]
 
@@ -603,6 +603,18 @@ if Code.ensure_loaded?(Postgrex) do
     def execute_ddl({command, %Table{} = table}) when command in @drops do
       [["DROP TABLE ", if_do(command == :drop_if_exists, "IF EXISTS "),
         quote_table(table.prefix, table.name)]]
+    end
+
+    def execute_ddl({:drop, %Extension{} = extension}) do
+      [["DROP EXTENSION ", quote_table(extension.name)]]
+    end
+
+    def execute_ddl({:create, %Extension{prefix: prefix} = extension}) when not is_nil(prefix) do
+      [["CREATE EXTENSION ", quote_name(extension.name), " SCHEMA ", quote_name(prefix)]]
+    end
+
+    def execute_ddl({:create, %Extension{} = extension}) do
+      [["CREATE EXTENSION ", quote_table(extension.name)]]
     end
 
     def execute_ddl({:alter, %Table{} = table, changes}) do
--- a/lib/ecto/migration/runner.ex
+++ b/lib/ecto/migration/runner.ex
@@ -9,6 +9,7 @@ defmodule Ecto.Migration.Runner do
   alias Ecto.Migration.Table
   alias Ecto.Migration.Index
   alias Ecto.Migration.Constraint
+  alias Ecto.Migration.Extension
 
   @doc """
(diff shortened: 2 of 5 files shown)


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -605,6 +605,18 @@ if Code.ensure_loaded?(Postgrex) do
         quote_table(table.prefix, table.name)]]
     end
 
+    def execute_ddl({:drop, %Extension{} = extension}) do
+      [["DROP EXTENSION ", quote_table(extension.name)]]
+    end
+
+    def execute_ddl({:create, %Extension{prefix: prefix} = extension}) when not is_nil(prefix) do
```
michalmuskala: Let's match the `nil` case first - I think it will make the code clearer.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -605,6 +605,18 @@ if Code.ensure_loaded?(Postgrex) do
         quote_table(table.prefix, table.name)]]
     end
 
+    def execute_ddl({:drop, %Extension{} = extension}) do
+      [["DROP EXTENSION ", quote_table(extension.name)]]
```
michalmuskala: Let's use `quote_name` consistently everywhere

lib/ecto/migration/runner.ex
```diff
@@ -164,6 +165,10 @@ defmodule Ecto.Migration.Runner do
     log_and_execute_ddl(repo, log, {:create, index})
   end
 
+  defp execute_in_direction(repo, :backward, log, {:create, %Extension{} = extension}) do
+    log_and_execute_ddl(repo, log, {:drop, extension})
+  end
```
michalmuskala: I think we should be able to handle the extension with the generic clause here, that dispatches to `reverse/1`.

josevalim:
Hi @roehst, this PR looks excellent!

One of the reasons we decided to support `extension` in Ecto was to make sure the bidirectional migrations, the ones that uses `change` instead of `up`/`down`, work properly.

However, I have just realized that, instead of supporting `extension`, it would be better if we changed the `execute` command in migrations to accept two commands: the forward command and the backward command:

    execute "create extension foo",
            "drop extension foo"

This way we no longer need to support extensions and we generalize the feature. The way it works is that we can only reverse the execute command if the second argument is given. If the second argument is not given, we should continue raising as today.

That said, I think we should rather improve the execute statement above instead of directly supporting extensions (which are PG specific). Could you instead send a PR that improves `execute` then? I apologize for not being able to see the improvements to `execute` sooner, but I only realized this as I reviewed your PR.

Thank you!

roehst:
@josevalim and @michalmuskala  Thanks for your prompt feedback!

This feature actually did seem odd to me because extensions are postgresql specific while the other migration commands (table, index, ...) are more general. Anyways, it was a good exercise to get know a bit more about the codebase!

If I understand correctly, the revised proposal is to make execute support two command strings, one for going forward and one for going backward.

Now, just as a constructive reflection. As a developer I would feel fine about using the execute command with two strings inside the change method. But is this substantially better than writing an up/down method in each specific migration?

Nevertheless, I need a couple days to study the execute command, but I can submit a PR on this. It will be good exercise again even if reasoning on the question above it looks like it is better to stick to using UP/DOWN methods vs. "reversible executes".

Thanks again!

josevalim:
It is better than writing up and down because sometimes you have multiple
commands and a single non reversible command may force you to rewrite
everything with up/down.

-- 


roehst:
@josevalim It makes a lot of sense! I will work on it. Thanks

josevalim:
Awesome, thank you @roehst!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Issue #1964, opened 2017-02-24
Dynamic Repo configuration


talentdeficit:
i have a use case where i'm unable to configure my database ahead of time via creating a Repo module. i know it's possible to compile at runtime but i'm uneasy about the possible security and performance implications of doing so. what i'd like is an alternative to `Repo` that accepts configuration at runtime

```elixir
repo = [
  adapter: Ecto.Adapters.Postgres,
  database: "ecto",
  ...
]
DynamicRepo.all(repo, query, opts)
```

`DynamicRepo` would be identical to modules generated via `using Ecto.Repo` apart from the arity change of functions to accept repo options

this is doable now as a library but i'm unsure how much of a `Repo`'s api is considered public and how much is considered private. if private, it may be beneficial to make the functionality part of ecto itself so the api's can be updated in tandem

i'm also open to alternatives if there is a better path forward

josevalim:
@talentdeficit have you tried the `init` callback we have recently added in Ecto 2.1? I believe from that callback you should be able to retrieve the configuration dynamically. So you configure the bare minimum in the config/config.exs file and the remaining options are configured on start.

There are probably some configurations that we are still using statically, such as the pool configuration in #1953 and there are probably things that are based on the name, such as ets tables that would avoid multiple starts, but with your help we can identify those and improve the code today.

What do you think?

michalmuskala:
I was thinking about this as well. What if we had the `Ecto.Repo` (or some other module) export functions exactly as those generated in user repos, but accepting an additional argument - a config struct. For the regular repos, we could generate this config struct based on the env and delegate the implementation, similar to what we're already doing today with multiple modules in the `Ecto.Repo.*` namespace. If somebody needs dynamic access to the repo, they could build this struct manually and call the extended functions directly.

This would make the ecto interface function-first with the macro generation of `use Ecto.Repo` a mere convenience for removing the boilerplate of passing the config struct everywhere manually.

josevalim:
@michalmuskala it is not that straight-forward because you still need to start a supervision tree, the pool and the processes that connect to the database. You cannot simply pass a config to Ecto.Repo.insert and have it connect to the database, that would be inefficient and its processes wouldn't be properly supervised.

josevalim:
@michalmuskala although your description would be correct if, instead of passing of a config, we pass a pid of a previously started repo.

michalmuskala:
[...]

PR #1977, opened 2017-03-04
[WIP] Dynamic pool configuration

michalmuskala:
This is a proof of concept of how we could allow for more dynamic configuration of repo.

Here's what's happening:
* A new `Ecto.Registry` module for a simple process-attached key-value store.
* A new adapter callback `init/2`, where adapter returns "runtime data", that is later stored in the registry
* Instead of passing the repo module, all adapter callbacks receive this "runtime data"
* The `__pool__` definition is removed
* The `repo` passed to various functions in the `Ecto.Repo.*` modules is treated as a process or a process name. They use it to retrieve data from the registry and pass to the adapter callbacks in place of repo.

This still needs work to truly allow dynamic repos, but should server as a good starting point for further work. Only tested with postgres for now.

Closes https://github.com/elixir-ecto/ecto/issues/1964

\cc @josevalim @talentdeficit


lib/ecto/adapter.ex
```diff
@@ -45,6 +45,11 @@ defmodule Ecto.Adapter do
   """
   @callback child_spec(repo, options) :: Supervisor.Spec.spec
 
+  @doc """
+  Initializes the adapter
+  """
+  @callback init(repo, options) :: {:ok, state :: term} | {:error, reason :: term}
```
josevalim: There is no need for an init callback. If the adapter needs to do extra work, then it needs to wrap the child_spec or the function in the child_spec returned by DBConnection by its function/child_spec and then execute any custom action that is necessary.

lib/ecto/repo/queryable.ex
```diff
@@ -11,11 +11,13 @@ defmodule Ecto.Repo.Queryable do
   require Ecto.Query
 
   def transaction(adapter, repo, fun, opts) when is_function(fun, 0) do
-    adapter.transaction(repo, opts, fun)
+    data = Ecto.Registry.lookup(repo)
```
josevalim: I am not sure this is the way we want to go. This is imposing all adapters to go through the Ecto.Registry, which means that now all adapters need to register something when they are initialized. I would prefer if, as this change relates only to pools, we did not change the adapter API at all, but only the SQL adapters which are the ones who effectively use a pool.

josevalim:
@michalmuskala I believe this PR is trying to do too much at once. I would prefer if we originally tackled the `__pool__` issue separately from the dynamic repositories. That's because pools are not a requirement for every Ecto adapter and if we put those together we will implicitly require all adapters to use a pool (which this PR seems to be accidentally doing).

In as few steps as possible, here is how I would see this PR working:

  1. It uses a registry
  2. It overrides the child_spec to invoke a function that would allow the pool options to be registered on start up
  3. We change __pool__ to lookup the pool information in the registry instead of being compiled in memory

This way we keep the existing APIs and make almost everything dynamic. The only non-dynamic piece is the query cache table which we can then discuss how to tackle next.

I would also suggest for us to use Elixir's v1.4 registry. Ecto 2.2 is not coming out soon and by the time it is out, I assume requiring Elixir v1.4 won't be a huge issue.

I would personally prefer if we don't do the lookup and let adapters be the ones responsible for doing any registry lookup. All we guarantee is that the adapter will either receive the repo name or a pid. We can then provide an up and running registry as a convenience (so an adapter doesn't need to start one).

michalmuskala:
This is exactly the way I tried to do it initially, but it presented a lot of issues. Most notably, we need to associate the top-level repo supervisor to the pool config, and not the pool process itself - if we had the pid of the pool process, we wouldn't need any of this.

josevalim:
Good call.

You should be able to get the pid though if you do the child_spec trick I
mentioned, because you will be the one calling DBConnection's start_link.

However this means you won't be able to use Elixir registry. So your
registry implemention will still be necessary.
-- 


michalmuskala:
@josevalim updated. The pool handling is now entirely on the adapter side - this actually simplified the code.

The only thing, I'm not 100% convinced about is that we assume in the `child_spec` function of the adapter that it will be called from within the main repo supervisor. I guess this is fine, but wanted to point it out. Ideally the supervisor would pass the pid/name to the `child_spec` function, but this would require changing the adapter API.

The ets table shouldn't be a problem - we can always name it as the top level repo supervisor, and require it to be named with an atom (could be called differently that the repo module itself - by default it wouldn't be called differently).

I was wondering how to provide the dynamic repo interface - I see two ways:
* add a module `Ecto.Repo.Dynamic` that exports all the functions `Ecto.Repo` does, but adds additional two arguments - for the name/pid of the dynamic repo.
* make all functions in the regular repos, accept an additional argument/option for the repo process.

I lean towards the first solution.

lib/ecto/adapters/sql.ex
```diff
@@ -525,26 +523,26 @@ defmodule Ecto.Adapters.SQL do
 
   @doc false
   def reduce(repo, statement, params, mapper, opts, acc, fun) do
-    {pool, default_opts} = repo.__pool__
+    {repo_mod, pool, default_opts} = Ecto.Registry.lookup(repo)
```
josevalim: We should probably move `Ecto.Registry.lookup(repo)` to a private function (given you had to repeat it in many places in this PR).

lib/ecto/repo/supervisor.ex
```diff
@@ -111,8 +111,8 @@ defmodule Ecto.Repo.Supervisor do
           :ets.new(repo, [:set, :public, :named_table, read_concurrency: true])
         end
         supervise(children, strategy: :one_for_one)
-      :ignore ->
```
josevalim: Why this change?

test/ecto/adapters/sql_test.exs
```diff
@@ -1,30 +1,30 @@
-defmodule Ecto.Adapters.SQLTest do
```
josevalim: We can delete this file but we need to add an integration test that shows we can pass the pool options on `start_link`.

josevalim: One good option to test is the `:pool_name` option.

josevalim:
> The only thing, I'm not 100% convinced about is that we assume in the child_spec function of the adapter will be called from within the main repo supervisor.

That's a reasonable assumption to make. In our proposal in the elixir-lang-core mailing list we are pushing towards the same direction.

> add a module Ecto.Repo.Dynamic that exports all the functions Ecto.Repo does, but adds additional argument - for the name/pid of the dynamic repo.

This is my preference too. We can also consider adding those functions directly to `Ecto.Repo` but it may end-up cluttering the API. If not `Ecto.Repo`, I would also prefer to call it `Ecto.DynamicRepo`.

josevalim:
@michalmuskala I pushed a commit that slightly renames normalize_config to pool_config and cleans up some of the indirection.

josevalim:
I went ahead and fixed the last issues beacuse I want to do more work on top of this. Thanks @michalmuskala!

michalmuskala:
:+1: sorry for the delay

PR #1984, opened 2017-03-10
[WIP] Array based relations

ankhers:
This PR contains the following relations

## ArrayBelongsTo

`array_belongs_to` is meant to be a many-to-one or many-to-many style relationship where instead of having a single `foo_id` field, you would have a `foo_ids` field that is an array of primary keys.

## ArrayHas

`array_has_one` and `array_has_many` is the opposite side of the `array_belongs_to` relation. The primary key for these records will be going into the `foo_ids` field.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -1038,3 +1038,237 @@ defmodule Ecto.Association.ManyToMany do
     end
   end
 end
+
+defmodule Ecto.Association.ArrayBelongsTo do
+  @moduledoc """
+  The association struct for `array_belongs_to` associations.
+
+  Its fields are:
+
+    * `cardinality` - The association cardinality
+    * `field` - The name of the association field on the schema
+    * `owner` - The schema where the association was defined
+    * `owner_key` - The key on the `owner` schema used for the association
+    * `related` - The schema that is associated
+    * `related_key` - The key on the `related` schema used for the association
+    * `queryable` - The real query to use for querying association
+    * `defaults` - Default fields used when building the association
+    * `relationship` - The relationship to the specified schema, default is `:child`
+    * `on_replace` - The action taken on associations when schema is replaced
+  """
+
+  @behaviour Ecto.Association
+  @on_replace_opts [:raise, :delete, :remove, :update]
+  defstruct [:field, :owner, :owner_key, :related, :related_key, :queryable,
+             :on_cast, defaults: [], cardinality: :many, relationship: :child,
+             unique: true]
+
+  @doc false
+  def struct(module, name, opts) do
+    ref       = if ref = opts[:references], do: ref, else: :id
+    queryable = Keyword.fetch!(opts, :queryable)
+    related   = Ecto.Association.related_from_query(queryable)
+
+    unless is_atom(related) do
+      raise ArgumentError, "association queryable must be a schema, got: #{inspect related}"
+    end
+
+    on_replace = Keyword.get(opts, :on_replace, :raise)
+
+    unless on_replace in @on_replace_opts do
+      raise ArgumentError, "invalid `:on_replace` option for #{inspect name}. " <>
+        "The only valid options are : " <>
+        Enum.map_join(@on_replace_opts, ", ", &"`#{inspect &1}`")
+    end
+
+    %__MODULE__{
+      field: name,
+      owner: module,
+      related: related,
+      owner_key: Keyword.fetch!(opts, :foreign_key),
+      related_key: ref,
+      queryable: queryable,
+      defaults: opts[:defaults] || []
+    }
+  end
+
+  @doc false
+  def build(refl, _, attributes) do
+    refl
+    |> build()
+    |> struct(attributes)
+  end
+
+  @doc false
+  def joins_query(%{queryable: queryable, related_key: related_key,
+                    owner: owner, owner_key: owner_key}) do
+    from o in owner,
+      join: q in ^queryable,
+      on: field(o, ^owner_key) in field(q, ^related_key)
+  end
+
+  @doc false
+  def assoc_query(%{queryable: queryable, related_key: related_key}, query, values) do
+  end
+
+  @doc false
+  def preload_info(%{related_key: related_key} = refl) do
+    {:assoc, refl, {0, related_key}}
+  end
+
+  @doc false
+  def on_repo_change() do
+  end
+
+  @behaviour Ecto.Changeset.Relation
+
+  @doc false
+  def build(%{related: related, queryable: queryable, defaults: defaults}) do
+    related
+    |> struct(defaults)
+    |> Ecto.Association.merge_source(queryable)
+  end
+end
+
+defmodule Ecto.Association.ArrayHas do
+  @moduledoc """
+  The association struct for `array_has_one` and `array_has_many` associations.
+
+  Its fields are:
+
+    * `cardinality` - The association cardinality
+    * `field` - The name of the association field on the schema
+    * `owner` - The schema where the association was defined
+    * `owner_key` - The key on the `owner` schema used for the association
+    * `related` - The schema that is associated
+    * `related_key` - The key on the `related` schema used for the association
+    * `queryable` - The real query to use for querying the association
+    * `on_delete` - The action taken on associations when schema is deleted
+    * `on_replace` - The action taken on associations when schema is replaced
+    * `defaults` - Default fields used when building the association
+    * `relationship` - The relationship to the specified schema, default is `:child`
+  """
+
+  @behaviour Ecto.Association
+  @on_delete_opts [:nothing, :remove_all, :delete_all]
+  @on_replace_opts [:raise, :mark_as_invalid, :delete, :remove]
+  @array_has_one_on_replace_opts @on_replace_opts ++ [:update]
+  defstruct [:cardinality, :field, :owner, :owner_key, :related, :related_key,
+             :on_cast, :queryable, :on_delete, :on_replace, unique: true,
+             defaults: [], relationship: :child]
+
+  @doc false
+  def struct(module, name, opts) do
+    ref =
+      module
+      |> Module.get_attribute(:primary_key)
+      |> get_ref(opts[:references], name)
+
+    unless Module.get_attribute(module, :ecto_fields)[ref] do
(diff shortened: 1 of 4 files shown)


ankhers:
This stemmed from michalmuskala/mongodb_ecto#113

johnnyshields:
@ankhers thanks for the awesome start on this. Did you see my post here: https://groups.google.com/forum/#!topic/elixir-ecto/bqf5LrSmEQM

ankhers:
@johnnyshields I had not until just now. If you would like to help with this PR, feel free to create a PR within my repository.

maxnordlund:
One think to note about arrays in Postgres is that you can't use foreign keys. This is not a problem for NoSQL databases, but I think it should be made clear in the docs that this is handled purely in Ecto.

Well, that or shipping some sort of trigger function that does the check alternatively using a materialised view over a normal join table (which would have to shadow the real table).

https://dba.stackexchange.com/questions/60132/foreign-key-constraint-on-array-member
https://www.postgresql.org/message-id/1343842863.5162.4.camel@greygoo.devise-it.lan

ankhers:
@maxnordlund I will add something in the documentation for the relations.

@josevalim @michalmuskala Would either of you (or anyone else really) mind taking a look at what I currently have and let me know what you think. I know I still need to implement some callbacks, but if I could get some guidance on at least `on_repo_change/4`, it would be appreciated.

johnnyshields:
@ankhers can you support the cases where you want to persist the array keys on BOTH sides of a many-to-many relation (and keep it in-sync)? e.g.:

```elixir
Student.array_belongs_to :course    # => stores data in Student.course_ids
Course.array_belongs_to :student    # => stores data in Course.student_ids
```

Bharat311:
@ankhers - I did try using this branch for `array_belongs_to` and it didn't seem to work for me. 

I got it working by making the following fixes: [https://github.com/kkvesper/ecto/commit/551bf5b5f298534d7c19c60249c6868697f74a31](https://github.com/kkvesper/ecto/commit/551bf5b5f298534d7c19c60249c6868697f74a31)

But for some reason GH is not allowing me to raise a PR against your branch. Could you please pull in these fixes in this PR ?

Great job on the PR 👍

ankhers:
@johnnyshields There are no plans to support setting the keys on both sides of the relationship in a single operation. It is unsafe to do that since Mongo does not support transactions. Ecto does not want to do anything unsafe by default.

@Bharat311 Thanks. I will take a look at the changes in a bit.

johnnyshields:
@ankhers Ruby's Mongoid library supports it, it is NOT unsafe since you can set the opposite side using the atomic `$push` and `$pull` modifiers. (I've used it in production for 4 years on a application with hundreds of millions of records with no issue)

ankhers:
@johnnyshields It is possible that I am wrong. But my understanding is that you can edit any single document atomically. Unfortunately, because Mongo does not have transactions, you cannot atomically edit 2 or more documents with a single query.

If you can show me documentation that states otherwise, I would be happy to support that functionality. However, after reading through [this](https://docs.mongodb.com/manual/core/write-operations-atomicity/) article, it does not seem like it is possible.

johnnyshields:
@ankhers, you can do with with two successive atomic queries ($addToSet / $pull). Even though there are two queries, it is safe (i.e. eventually consistent) because the queries are atomic.

For example, suppose I have models Student and Courses, and do the following actions:

```
student1.course_ids = [course1._id]
student1.save

# generates two queries:
# students (where _id: student1._id)  $addToSet: { course_ids: course1._id }
# courses (where _id: course1._id)  $addToSet: { student_ids: student1._id }

student1.course_ids = [course1._id, course2._id, course3._id]
student1.name = "Bob"
student1.save

# generates two queries -- notice that $set is used for non-relation attributes.
# students (where _id: student1._id)  $set: {name: "Bob"}, $addToSet: { course_ids: course1._id }
# courses (where _id: course1._id, course2._id, course3._id)  $addToSet: { student_ids: student1._id }

student1.course_ids = [course1._id, course2._id, course4._id]
student1.save

# generates three queries:
# students (where _id: student1._id)  $pull: { course_ids: course3._id }, $addToSet: { course_ids: course4._id }
# courses (where _id: course3._id)  $pull: { student_ids: student1._id }
# courses (where _id: course4._id)  $addToSet: { student_ids: student1._id }
```

I highly, highly recommend you study Mongoid. It is a *near-perfect* reference implementation of how to implement a model/relation system with MongoDB. Ecto for MongoDB essentially clone what they have.

johnnyshields:
(By the way, **eventually consistent** is the name of the game for Mongoid. It is unlike SQL databases where you try to go from one transactional state from another--if you want this sort of thing, you shouldn't use Mongoid or you're going to be in for constant heartache.)

ankhers:
That is the problem though. You are executing two statements.

What would happen in mongoid if the first query was successful, but the second query failed? It does not matter why it failed, but it did. There is not way, you could guarantee to be in a good state. I see two options to what you could do.

* Try the same failing query over and over again. 
* Attempt to revert the previous query in order to be in a stable state.

Neither of these things seem like good solutions. At the heart of it, you cannot guarantee that 2 or more operations are going to be successful within Mongo.

josevalim:
> That is the problem though. You are executing two statements.

Exactly. You don't have transactions and there is no guarantee both statements will fail or succeed at once.

johnnyshields:
@ankhers @josevalim there are two points to consider:

1) Assuming that MongoDB processes all transactions **without error**, but processes them potentially multiple updates on various object in various orders, if we use the `$addToSet` approach I have outlined the data will maintain integrity (i.e. it is free of race-conditions, one update will not overwrite another)

2) Single queries generally do not "fail" in my experience unless the entire database fails and starts dropping transactions. If this happens, you are probably in deeper shit than just a few out-of-sync relations, and in fact having a two-sided relation actually **helps** this recovery as you can compare the two sides and reconcile the data--I had data-corruption a long time ago and the two-sided relation was a life-saver.

This two-sided array relation would be **optional**; the implementer would need to be aware of the limitations in case of DB failure. As stated I've used it successfully for 4 year in production.

Lastly, I assert it is incorrect to design Ecto for MongoDB as it if MongoDB were a transactional, relational-constrained DB. By design, MongoDB is NOT such a database.

josevalim:
The point of eventual consistency is exactly about what happens to your system *in face of failures*. If you remove network partitions, hiccups, nodes going down, then eventual consistency becomes really straight-forward.

> Single queries generally do not "fail" in my experience unless the entire database fails and starts dropping queries. 

Maybe a linked process crashed bringing the one running the query down. Maybe the node is being shut down abruptly because another application misbehaved. Maybe there was a network partition right after the first call. Maybe mongo is overloaded and refused a request.

> in fact having a two-sided relation actually helps this recovery as you can compare the two sides and reconcile the data

What would really help recovery is if the data could not get corrupted in the first place. :)

> Lastly, I assert it is incorrect to design Ecto for MongoDB as it if MongoDB were a transactional, relationally-constrained DB. By design, MongoDB is NOT such a database.

We agree but I don't think we should promote bad practices. Exactly because the abstractions we provide will be available to all databases.

maxnordlund:
I think the point the good people here are trying to make is that Ecto shouldn't ship anything remotely dangerous, aka no caveat emptor to beware of.

Their [documentation](https://docs.mongodb.com/manual/core/write-operations-atomicity/
) clearly states that Mongo does not have support for multi-document transactions, and instead encourage the reader to use nested documents since operations on _individual_ documents are guaranteed to be atomic.

But they do provide an `$isolated` operator that locks the entire collection (table), which can be used to [perform a two-phased commit]()https://docs.mongodb.com/manual/tutorial/perform-two-phase-commits/. The docs clear state that this is just transaction-*like* semantics, per the document level isolation.

Finally a good read about what happens to a Mongo database when things _do_ break, be sure to read the follow up posts for later versions. https://aphyr.com/posts/284-jepsen-mongodb

johnnyshields:
> Ecto shouldn't ship anything remotely dangerous, aka no caveat emptor to beware of.

I do not consider this "remotely dangerous". The **MongoDB team themselves is supporting via the Mongoid Ruby library** which they own--in fact it's the **default behavior** for many-to-many relationships!

> $isolated / two-phased commits

These do not address the problem at hand (close but no cigar.)

ankhers:
It may be worth doing what michalmuskala said in his comment [here](https://github.com/michalmuskala/mongodb_ecto/issues/113#issuecomment-284804886). The important part:

> Providing support for this, would first need work to be done on ecto, to make the association API extensible (it's not far from being that, but it's not exactly it yet). Once that is done, custom association types could be provided by `mongodb_ecto`.

I think if we keep the current direction in this PR, make the required modifications to the association API in a different PR, then we can build what you are looking for in either the `mongodb_ecto` adapter, or in a separate library altogether.

Or would it be better to forgo this PR and just focus on making changes to the association API?

josevalim:
> Or would it be better to forgo this PR (since it is currently Mongo centric) and just focus on making changes to the association API?

Let's keep this PR as is right now because we may just merge the extension points and then you can simply move the rest of the code to `mongodb_ecto`. Keeping it altogether for now will the best in terms of discussion and review.

johnnyshields:
@ankhers I'm all in favor to move this behavior to mongodb_ecto.

I cannot say more--for anyone working on MongoDB in Ecto, please, please study Mongoid.

josevalim:
I will be reviewing this PR this week and hopefully either merge it or provide the extension points for MongoEcto (or both).

In any case, the two-sided array_belongs_to feature is UNSAFE, even for Mongo. The fact Mongoid implements it does not mean it is safe, it just means they accepted the trade-offs that come with its unsoundness.

However, given this is the modus operandi for Mongo, it probably makes sense for the MongoEcto team to support the feature regardless. Especially because there are parts of Ecto that is supported by MongoEcto and rely on transactions, which Mongo does not provide, and therefore is already unsafe. It may be a good idea to document the trade-offs nonetheless.

josevalim:
Hi @ankhers, this looks great so far! It seems the only feature you currently need to implement it from MongoEcto is an Ecto.Schema API that allows you to define the association. Is this correct? If so, I can work on it this week.

johnnyshields:
Just a note, MongoDB 4.0 is now supporting transactions, so it is now possible to implement a "safe" (transactional) two-sided array relation.

PR #1986, opened 2017-03-11
Fix confusing example index names

knewter:
These are indexes that constrain a user's email, but they are named
things like `posts_special_email_index`. Renamed to
`users_special_email_index` as I believe this was the original intent.


michalmuskala:
Thank you! :heart:

PR #1989, opened 2017-03-13
Improve changeset docs

michalmuskala:
I'm opening this as a PR to encourage discussion about those changes.

The key difference is introducing the concept of internal vs external data and advising that `*_assoc` functions work on entire associations and not single elements.

\cc @josevalim @yordis


josevalim:

PR #1990, opened 2017-03-13
Simplify embeds_one example to use embedded_schema

archSeer:
Seems like the examples were updated everywhere 2 years ago when the function was introduced, but this section was forgotten.


josevalim:

PR #1991, opened 2017-03-15
Validation rule: validate_required_without

PavelVesnin:
Hello!
I'd like to add new validation rule `validate_required_without\4`
It's useful for cases, when some field should be required when another fields is not present.
`%{"title" => "hello"} |> validate_required_without(:body, :title) # validation passed`
`%{"title" => "hello"} |> validate_required_without(:title, :body) # validation passed`
`%{"title" => "hello"} |> validate_required_without(:body, :decimal) # validation failed`

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1383,8 +1383,8 @@ defmodule Ecto.Changeset do
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @PavelVesnin but it is probably better to include this in your project that adding an ever growing list of validations to Ecto.

Issue #1978, opened 2017-03-04
Support forward and backward command in Ecto.Migration.execute


josevalim:
/cc @roehst

roehst:
I opened a PR with a minimal implementation to get feedback proceeding further. More details on the PR. If this is not the best workflow (if I should work on my fork) then sorry and I'll fix it.

#1993

PR #1993, opened 2017-03-16
Minimal implementation for reversible Ecto.Migration.execute

roehst:


roehst:
Addressing #1978 

This is a minimal implementation of a reversible Ecto.Migration.execute function.

I am submitting this to get feedback on the strategy I used to implement it, for the following reason.

The Ecto.Migration.Runner stores create/drop table/index/constraint commands as a struct and derives the forward/backward DDL statements. Reversing the command is as simple as switching a tuple from {:create, %Table{}} to {:drop, %Table{}}.

For the reversible execute, both directions must be stated and kept in a struct. 

So I added a %Command{up: String.t, down: String.t} struct to go along with %Table and %Index and %Constraint. 

This makes commands from execute/1 be treated differently from execute/2, because execute/1 will be represented by a String and execute/2 by %Command.

Is this approach ok? It feels a bit awkward to me.

lib/ecto/migration.ex
```diff
@@ -560,6 +565,13 @@ defmodule Ecto.Migration do
     Runner.execute command
   end
 
+  def execute(command_up, command_down)
+    when
+    (is_binary(command_up) or is_list(command_up)) and
+    (is_binary(command_down) or is_list(command_down)) do
+      Runner.execute %Command{up: command_up, down: command_down}
+  end
```
josevalim: Proper indentation will be:

```elixir
    def execute(command_up, command_down)
        when (is_binary(command_up) or is_list(command_up)) and
             (is_binary(command_down) or is_list(command_down)) do
      Runner.execute %Command{up: command_up, down: command_down}
    end
```

lib/ecto/migration/runner.ex
```diff
@@ -152,6 +153,14 @@ defmodule Ecto.Migration.Runner do
   ## Execute
   @creates [:create, :create_if_not_exists]
 
+  defp execute_in_direction(repo, :forward, log, %Command{up: up}) do
+    log_and_execute_ddl(repo, log, up)
+  end
+
+  defp execute_in_direction(repo, :backward, log, %Command{down: down}) do
```
josevalim: Please move this one clause below to group all `:backward` clauses together.

josevalim:
This looks great! Next steps is to address the comments, provide documentation and tests in "test/ecto/migration_test.exs" or similar.

michalmuskala:
Would it make sense to place the regular `execute` into the `Command` struct given we have it now? We could raise when trying to convert to backwards if the `down` field is `nil`.

josevalim:
@michalmuskala I thought about the same and a quick try showed it is more line of code with more cases to handle (because execute with a single argument still needs to work on down) so I got to the conclusion the current approach is the way to go.

roehst:
@michalmuskala @josevalim ok guys, I'll find some time to work on this. tks for the feedback =)

roehst:
Hi, I added documentation and test cases, does it look OK? Is there something missing?

As stated in the commit, the test strategy used for create/drop commands is not exactly appliable to the execute command, for the same reason discussed above (raw commands go straight to the runner). 

Please see the message on https://github.com/elixir-ecto/ecto/pull/1993/commits/b660fd61fa0b24dfafb15c90e494a4068f9e8245

josevalim:

PR #1996, opened 2017-03-17
Add new test tags to support upcoming release of sqlite_ecto2.

scouten:
Each of the tagged tests needs to be disabled for sqlite_ecto2 because it makes use of features that can not be supported by SQLite.

(Note: Yes, I'm intentionally targeting the 2.1 branch, though I'd be grateful to see this carried forward to the master branch as well.)

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -479,6 +479,7 @@ defmodule Ecto.Integration.RepoTest do
     assert changeset.errors == [permalink: {"is still associated with this entry", []}]
   end
 
+  @tag :foreign_key_constraint
   test "insert and update with failing child foreign key" do
     defmodule Order do
       use Ecto.Integration.Schema
@@ -632,6 +633,7 @@ defmodule Ecto.Integration.RepoTest do
     assert TestRepo.aggregate(query, :count, :visits) == 3
   end
 
+  @tag :insert_cell_wise_defaults
   test "insert all" do
     assert {2, nil} = TestRepo.insert_all("comments", [[text: "1"], %{text: "2", lock_version: 2}])
     assert {2, nil} = TestRepo.insert_all({"comments", Comment}, [[text: "3"], %{text: "4", lock_version: 2}])
@@ -653,6 +655,7 @@ defmodule Ecto.Integration.RepoTest do
   end
 
   @tag :returning
+  @tag :insert_cell_wise_defaults
   test "insert all with returning with schema" do
     assert {0, []} = TestRepo.insert_all(Comment, [], returning: true)
     assert {0, nil} = TestRepo.insert_all(Comment, [], returning: false)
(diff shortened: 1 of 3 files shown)


scouten:
I'm looking forward to retiring this awful hack 😄 :

https://github.com/scouten/sqlite_ecto2/blob/master/integration/hack_out_incompatible_tests.sh

integration_test/cases/repo.exs
```diff
@@ -632,6 +633,7 @@ defmodule Ecto.Integration.RepoTest do
     assert TestRepo.aggregate(query, :count, :visits) == 3
   end
 
+  @tag :insert_cell_wise_defaults
```
josevalim: Sorry but I don' understand exactly what this tag is about. Can you please explain why this does not work on sqlite_ecto2?

integration_test/sql/subquery.exs
```diff
@@ -18,6 +18,7 @@ defmodule Ecto.Integration.SubQueryTest do
            TestRepo.all(from p in subquery(query), select: p)
   end
 
+  @tag :map_boolean_in_subquery
```
josevalim: Sorry but I don' understand exactly what this tag is about. Can you please explain why this does not work on sqlite_ecto2?

Usually the tags are about the database features that we don't support but this is unfortunately phrased in terms of Ecto features.

josevalim:
@scouten thank you! I have added some comments!

integration_test/cases/repo.exs
scouten: Sure. In Postgres, you can write `INSERT` statements that use `DEFAULT` value for individual column values. If you are doing a multi-row insert, different columns can be defaulted for different rows.

This is not possible in SQLite. As [SQLite's documentation for `INSERT`](https://www.sqlite.org/lang_insert.html) shows, you can _either_ have a global `DEFAULT VALUES` clause or you can explicitly provide values. You can not use `DEFAULT` inside the `VALUES(...)` clause.

josevalim: Ok, thanks makes sense. This tag looks good to me!

integration_test/sql/subquery.exs
scouten: This might be reflective of something I didn't quite understand properly about how to write an Ecto adapter. The test in question relied on being able to pass values from a subquery through placeholder values `?nnn` to the database engine.

The problem, for SQLite, is that this test uses a boolean value for the `public` field that comes through in a subquery. I did not see the dumper function get called for that value, so the boolean got passed through directly to the database engine.

SQLite does not have an intrinsic boolean data type, so I rely on the trick of replacing boolean with numeric 1 and 0. If that doesn't happen, I get an error running the query.

It's certainly possible that I misunderstood how the dumper should be invoked or maybe there's a bug in Ecto. Open to further discussion/guidance on how to proceed here.

josevalim: @scouten MySQL has the same issue. They automatically convert `true` and `false` in the database driver into 1 and 0. If you don't want to do it on the adapter, you can do it on a pre-pass of the parameters on Ecto side, like Ecto's MySQL adapter does here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/mysql/connection.ex#L36-L45

We can also change the test to use another value rather boolean but I think the adapter would be benefit more if it could handle true and false.

We won't call the dumper because we consider true and false to be a native value.

scouten: OK, I'll review the MySQL adapter and see if that works. I'll withdraw this tag.

josevalim:

PR #2000, opened 2017-03-21
Improve validation performed by UUID.cast

### LINKED ISSUE #1999: UUID.cast should treat upper- and mixed-case UUIDs as invalid
### Environment

* Elixir version (elixir -v): 1.4.2
* Ecto version (mix deps): ecto 2.1.4
* Operating system: macOS 10.12.3

### Current behavior

```
UUID.cast("20d935d2-1630-487B-9b5d-38dDd3c5d88A")
```
returns
```
{:ok, "20d935d2-1630-487B-9b5d-38dDd3c5d88A"}
```

### Expected behavior

According to [Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format), only lowercase characters are allowed in the canonical format. I would expect the code above to return `:error`.
Perhaps normalization (automatically converting to lowercase) is the way to go? But then should the dashes be also automatically added if missing?

Dremora:
Closes #1999.

On top of what was reported, I've realised that the function will actually accept any characters in place of `a-z0-9`. This is now fixed.
I've also limited which characters can be allowed at two specific positions:
```
xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx
```
M is `[1-5]` and N is `[89ab]`. [Source](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format).

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -13,7 +13,25 @@ defmodule Ecto.UUID do
   @doc """
   Casts to UUID.
   """
-  def cast(<< _::64, ?-, _::32, ?-, _::32, ?-, _::32, ?-, _::96 >> = u), do: {:ok, u}
+  def cast(<< a1, a2, a3, a4, a5, a6, a7, a8, ?-,
+              b1, b2, b3, b4, ?-,
+              c1, c2, c3, c4, ?-,
+              d1, d2, d3, d4, ?-,
+              e1, e2, e3, e4, e5, e6, e7, e8, e9, e10, e11, e12 >>) do
+    << c(a1), c(a2), c(a3), c(a4),
+       c(a5), c(a6), c(a7), c(a8), ?-,
+       c(b1), c(b2), c(b3), c(b4), ?-,
+       cast_version(c1), c(c2), c(c3), c(c4), ?-,
+       cast_variant(d1), c(d2), c(d3), c(d4), ?-,
+       c(e1), c(e2), c(e3), c(e4),
+       c(e5), c(e6), c(e7), c(e8),
+       c(e9), c(e10), c(e11), c(e12) >>
+  catch
+    :error -> :error
+  else
+    casted ->
+      {:ok, casted}
+  end
   def cast(_), do: :error
 
   @doc """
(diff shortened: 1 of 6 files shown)


Dremora:
After [digging into](https://tools.ietf.org/html/rfc4122) [standards](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-X.667-201210-I!!PDF-E&type=items), I'm coming to the conclusion that we should parse both upper and lower-case UUIDs but always output lowercase. Since the function is called `case` (and not `validate`), I assume that conversion can be expected.
Also, it seems there's a few tests failing which were not using valid UUIDs, I'll fix those as well.

lib/ecto/uuid.ex
```diff
@@ -13,8 +13,14 @@ defmodule Ecto.UUID do
   @doc """
   Casts to UUID.
   """
-  def cast(<< _::64, ?-, _::32, ?-, _::32, ?-, _::32, ?-, _::96 >> = u), do: {:ok, u}
-  def cast(_), do: :error
+  def cast(uuid) do
+    regex = ~r/\A[a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}\z/i
```
josevalim: Please don't use a regular expression. :) See the function below on this module on how you can use pattern matching to solve this. I assume it will end-up looking like dump but instead of converting to the underlying binary, we will keep the normalized UUID.

Dremora: Good point! I didn't quite realise it's possible, but now I see that this is what `dump` is doing. I don't really like regular expressions anyway :) The build is green now.

lib/ecto/uuid.ex
```diff
@@ -13,7 +13,27 @@ defmodule Ecto.UUID do
   @doc """
   Casts to UUID.
   """
-  def cast(<< _::64, ?-, _::32, ?-, _::32, ?-, _::32, ?-, _::96 >> = u), do: {:ok, u}
+  def cast(<< a1, a2, a3, a4, a5, a6, a7, a8, ?-,
+              b1, b2, b3, b4, ?-,
+              c1, c2, c3, c4, ?-,
+              d1, d2, d3, d4, ?-,
+              e1, e2, e3, e4, e5, e6, e7, e8, e9, e10, e11, e12 >>) do
+    try do
```
Dremora: I'll fix this now.

lib/ecto/uuid.ex
```diff
@@ -26,6 +44,51 @@ defmodule Ecto.UUID do
...
+  defp c(?D), do: ?d
+  defp c(?E), do: ?e
+  defp c(?F), do: ?f
+  defp c(?a), do: ?a
+  defp c(?b), do: ?b
+  defp c(?c), do: ?c
+  defp c(?d), do: ?d
+  defp c(?e), do: ?e
+  defp c(?f), do: ?f
+  defp c(_),  do: throw(:error)
```
OvermindDL1: You can collapse all those `c/1`'s into:
```elixir
defp c(v) when (v >= ?0 and v<=?9) or (v >= ?a and v<= ?f), do: c
defp c(v) when v >= ?A and v <= ?9), do: v + (?a - ?A) #  Do you really want to allow uppercase even if normalized?
defp c(_), do: throw(:error) # Should probably be an exception
```
It might also be faster in matching, maybe (benchmark if anyone really cares?).

OvermindDL1: `cast_version/1` and `cast_variant/1` can have the same thing done as well, if it matters.  The longer lines might be easier to read though?

Dremora: Interesting point. I was mainly following the way `dump` is implemented. I'd say it would make sense to change both functions in one go if we decide to do so — but personally for me longer version is more readable.
In regards to uppercase, we should always convert it to lowercase. I've noticed `cast_variant` doesn't do it — it's fixed now.

Dremora: In regards to throwing exception, I don't know what the best practice here is. The only other place in the Ecto codebase also simply throws `:error`, so I've simply followed that.

josevalim: @OvermindDL1 the current versions are much faster. I have sent a PR to OTP that makes the comparisons faster but the explicitly pattern matching will almost always beat it, especially for small trees like this.

Also throw is better conceptually than an exception because *it is meant to be caught*. It should never leak the current module. And throw is faster on this case because we don't change the stacktrace (which exceptions do when rescued).

So please keep it as is. :)

Dremora: This is now ready to merge. Any other feedback?

josevalim:

PR #2005, opened 2017-03-23
Clarifies the documentation on the :on_conflict option

isaacsanders:
Specifies that the :replace_all value replaces "values" on an "entry", rather than "entries" in a "database".


michalmuskala:
Thank you! :heart:

PR #2008, opened 2017-03-29
Add option to generate changeset error on unknown fields in params

### LINKED ISSUE #2007: `Ecto.Changeset.cast` variant that results in an error on unexpected fields
i'd like a variant of `Ecto.Changeset.cast` (here named `restrict`) that instead of stripping fields from `data` produces an error like:
```elixir
data = %{}
types = %{ :allowed => :integer }
{ data, types } |> restrict(%{ :verboten => "this field is not allowed" }, [ :allowed ])
# %Ecto.Changeset{ errors: [ :verboten, { "is not a permitted field", [ validation: :restrict ] } ], ... }
```
is there a better method that i'm missing or is this something that would be a welcome addition?

talentdeficit:
closes #2007

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -232,6 +232,7 @@ defmodule Ecto.Changeset do
   alias Ecto.Changeset.Relation
 
   @empty_values [""]
+  @unknown_fields :ignore
 
   # If a new field is added here, def merge must be adapted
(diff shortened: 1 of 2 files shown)


josevalim:
Per #2007, doing this in `cast` won't play well with many features in Ecto, so we are closing this. Thank you!

PR #2010, opened 2017-03-30
new "force" parameter to ecto.drop task

epilgrim:
It allows to skip the shell confirmation for a database drop.

The usecase, is mainly for staging environments, that run in a production like setup.

The fact that the skip_safety_warnings? is environment dependent tripped me over. If we are in a mix task, why making this decision indirect? from my point of view it should be either

A_ Always ask for confirmation
B_ Forced drop

where the user that runs the command is directly in control, not depending on "indirect" configuration

In my opinion, the `skip_safety_warnings?` should not exist (or depend solely on the tasks arguments). But that is not done in this PR (could be done, if you agree)

### DIFF
--- a/lib/mix/tasks/ecto.drop.ex
+++ b/lib/mix/tasks/ecto.drop.ex
@@ -20,13 +20,14 @@ defmodule Mix.Tasks.Ecto.Drop do
 
     * `-r`, `--repo` - the repo to drop
     * `--no-compile` - do not compile before stopping
+    * `--force` - do not ask for confirmation
 
   """
 
   @doc false
   def run(args) do
     repos = parse_repo(args)
-    {opts, _, _} = OptionParser.parse args, switches: [quiet: :boolean]
+    {opts, _, _} = OptionParser.parse args, switches: [quiet: :boolean, force: :boolean]
 
     Enum.each repos, fn repo ->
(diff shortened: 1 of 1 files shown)


josevalim:
@epilgrim please list the option as a boolean here: https://github.com/epilgrim/ecto/blob/eeb4e3855dea4a7fb72dc9cd11473e70fa2180ae/lib/mix/tasks/ecto.drop.ex#L30

epilgrim:
right, added @josevalim 

can you clarify the rationale behind `skip_safety_warnings?` please?

josevalim:
@epilgrim we don't want to show the warning in development / test but it is required in production. So it does the right thing in 90% of the cases.

josevalim:
Of course the number of 90% has no scientific basis. :P But the idea is that it is a reasonable default.

epilgrim:
ok, get your point. I would argue that in those environments, user is in control and can add a --force (if the PR is accepted) without having to depend on other configurations.
But in the end, it's a usability feature, and your point is perfectly valid :)

josevalim:
Yes, we could stop supporting the skip safety warnings if we have force but it is not worth breaking people's builds, so this is fine. :)

josevalim:

PR #2011, opened 2017-03-30
Clarify 'from' macro expected arguments

tsubery:
It's quite confusing when you pass what will be a keyword list at runtime and get an error message saying it's not a keyword list.


josevalim:

PR #2016, opened 2017-04-01
only key needs to be passed for delete method

ananthakumaran:


josevalim:

Issue #2018, opened 2017-04-03
Introduce usec types


josevalim:
Since we migrated to Elixir built-in types, issues are starting to surface on how Ecto handles microseconds.

Those issues comes from the fact that database adapters do not cover the difference in precision that exists in Elixir. Leading to issues described in #1935.

Furthermore, databases may require different types for usec and non-usec, such as MySQL in #2012.

One solution to this problem is to introduce `:naive_datetime_usec` and `:utc_datetime_usec`. Those types can be used in migrations, explicitly tackling MySQL issues, and will guarantee on casting/dumping/loading that the microseconds precision is always 6, mirroring the database behaviour. The current `:naive_datetime` and `:utc_datetime` will discard the microsecond values, both on cast, dumping and loading.

The downside of this solution is the multiplicity of times but, given the difference in behaviour, it seems more explicit then passing `usec: true` everywhere (which even though is not enough to solving all issues above).

Thoughts?

/cc @michalmuskala @wojtekmach @lau

lau:
I like the idea. Isn't `usec: true` default now for :utc_datetime? How about instead of changing that, adding `:utc_datetime_no_subsecond` (just an example name) for datetimes with no sub second precision? Ie. instead of renaming the current type and adding a new one without subseconds, we add a new type with subseconds. I'm not 100% set on what I think would be best of the two, but want to hear opinions on it.

lau:
The shorter name (`utc_datetime`) is the default and will possibly seen as the default vs a new option with a longer name.

If you want to throw away as little data as possible by default, that would suggest having the default option saving the microseconds.

josevalim:
`usec: true` is not the default. Plus it also won't work in many databases.

lau:
I must have been reading the docs about the `timestamps_opts` where `usec: true` is the default. Was this a change of defaults? How is the  `timestamps_opts` with `usec: true` default working with MySQL? Does it work with MySQL 5.7?

For people using Postgres I think not throwing away the microseconds is the best default. But of course Ecto has to work with MySQL too. Could MySQL users using an old version of MySQL use a `:utc_datetime_no_subsecond` type after upgrading to a new minor release?

josevalim:
[...]

PR #2030, opened 2017-04-12
Add type checking to put_change/3 and update_change/3

wojtekmach:
This will be especially useful if we introduce `:naive_datetime_usec` (https://github.com/elixir-ecto/ecto/issues/2018), as doing e.g.: `Changeset.change(accepted_at: NaiveDateTime.utc)` would cast to the correct type (with or without microseconds)

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -287,6 +287,8 @@ defmodule Ecto.Changeset do
   The function is meant for working with data internal to the application.
   Because of that neither validation nor casting is performed. This means
   `change/2` expects the keys in the `changes` map or keyword to be atoms.
+  If any value in the given `changes` does not match the
+  type for that key in the schema (or `types` map), an error is raised.
 
   When a changeset is passed as the first argument, the changes passed as the
   second argument are merged over the changes already in the changeset if they
@@ -999,6 +1001,9 @@ defmodule Ecto.Changeset do
   in the changeset data, it is not added to the list of changes.
 
   The function is meant for working with data internal to the application.
+  Because of that neither validation nor casting is performed. If `value`
+  does not match the type for the given `key` in the schema (or `types` map),
+  an error is raised.
 
   ## Examples
 
@@ -1037,10 +1042,12 @@ defmodule Ecto.Changeset do
 
   defp put_change(data, changes, errors, valid?, key, value, type) do
     case Ecto.Type.cast(type, value) do
-      {:ok, value} ->
+      {:ok, ^value} ->
         put_change(data, changes, errors, valid?, key, value)
+      {:ok, _} ->
+        raise Ecto.CastError, type: type, value: value
       :error ->
-        raise Ecto.CastError, "cannot cast #{inspect value} to type #{inspect type}"
+        raise Ecto.CastError, type: type, value: value
     end
   end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -974,6 +974,11 @@ defmodule Ecto.Changeset do
   is a change for the given `key`. Note that the value of the change
   can still be `nil` (unless the field was marked as required on `cast/4`).
 
+  The function is meant for working with data internal to the application.
+  Because of that neither validation nor casting is performed. If the return
+  value of the `function` does not match the type for the given `key` in the
+  schema (or `types` map), an error is raised.
+
   ## Examples
 
       iex> changeset = change(%Post{}, %{impressions: 1})
@@ -986,8 +991,7 @@ defmodule Ecto.Changeset do
   def update_change(%Changeset{changes: changes} = changeset, key, function) when is_atom(key) do
     case Map.fetch(changes, key) do
       {:ok, value} ->
-        changes = Map.put(changes, key, function.(value))
-        %{changeset | changes: changes}
+        put_change(changeset, key, function.(value))
       :error ->
         changeset
     end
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1012,7 +1012,20 @@ defmodule Ecto.Changeset do
...
+  defp put_change(data, changes, errors, valid?, key, value, nil) do
+    put_change(data, changes, errors, valid?, key, value)
+  end
+
+  defp put_change(data, changes, errors, valid?, key, value, type) do
+    case Ecto.Type.cast(type, value) do
+      {:ok, value} ->
+        put_change(data, changes, errors, valid?, key, value)
+      :error ->
+        raise Ecto.CastError, "cannot cast #{inspect value} to type #{inspect type}"
```
wojtekmach: if accepted, we could also use Ecto.Type.Cast proposed in https://github.com/elixir-ecto/ecto/pull/2031/files#diff-20335f2fd54513fa94ed285d49d98147R130

lib/ecto/changeset.ex
```diff
@@ -1012,7 +1012,20 @@ defmodule Ecto.Changeset do
           "please use put_#{tag}/4 instead"
   end
 
-  defp put_change(data, changes, errors, valid?, key, value, _type) do
+  defp put_change(data, changes, errors, valid?, key, value, nil) do
```
michalmuskala: When would this happen?

wojtekmach: when a given `key` is not in the map of `types` (https://github.com/elixir-ecto/ecto/blob/v2.1.4/lib/ecto/changeset.ex#L1002). Without adding this clause, I got a bunch of failing tests

wojtekmach: (https://github.com/elixir-ecto/ecto/pull/2031#issuecomment-293809301)

> That's strange... As far as I'm aware we should always have types for all the possible fields, if there's no type, this should be an error

@michalmuskala I think you meant to make the comment here. We allow doing e.g. `|> put_change(:some_non_existing_field, 42)`, and so there won't be a corresponding type for this field. We discussed making `put_change` stricter with regards to fields it accepts here https://github.com/elixir-ecto/ecto/pull/1883 but it got rejected.

michalmuskala: Oh yeah, good job me, keeping up with the issues 😆 

You're right - this is fine.

josevalim:
My concern with this change is that we always required `put_change` to give us an exact value and now we are casting. We are allowing for an example a string to be given where an integer is expected. In certain ways, this has become `cast!` except we don't read it from the parameters but from a given value.

In order words, I think the value checking is probably a good change but I am worried about the casting aspects.

One thing we could do is to ensure that the value returned by Ecto.Type.cast is of the same type as the one given. This would alleviate my concerns but feels somewhat coupled.

lib/ecto/changeset.ex
```diff
@@ -287,6 +287,8 @@ defmodule Ecto.Changeset do
   The function is meant for working with data internal to the application.
   Because of that neither validation nor casting is performed. This means
   `change/2` expects the keys in the `changes` map or keyword to be atoms.
+  If any value in the given `changes` does not match the
+  type for that key in the schema (or `types` map), an error is raised.
```
wojtekmach: I repeated this section here (`change/2`) and `put_change` and `update_change` - is this OK? Also, I kind of wanted to mentioned that an arbitrary key can be passed (that's not in schema or types) and it won't be type checked and ignored when persisting. But not sure how to best write it.

lib/ecto/changeset.ex
```diff
@@ -1015,7 +1024,22 @@ defmodule Ecto.Changeset do
           "please use put_#{tag}/4 instead"
   end
 
-  defp put_change(data, changes, errors, valid?, key, value, _type) do
+  defp put_change(data, changes, errors, valid?, key, value, nil) do
+    put_change(data, changes, errors, valid?, key, value)
+  end
+
+  defp put_change(data, changes, errors, valid?, key, value, type) do
+    case Ecto.Type.cast(type, value) do
```
wojtekmach: looking at the clauses below we could use `Ecto.Type.dump` instead, because the given value must exactly match type type. WDYT?

lib/ecto/changeset.ex
```diff
@@ -1015,7 +1024,22 @@ defmodule Ecto.Changeset do
...
+      {:ok, ^value} ->
+        put_change(data, changes, errors, valid?, key, value)
+      {:ok, _} ->
+        raise Ecto.CastError, type: type, value: value
```
wojtekmach: as we discussed e.g. `"10"` can be put for `:integer` field. Which makes me think that `cannot cast "10" as :integer` can be a confusing error message, because casting in other contexts would allow that. Should we do e.g.:

```elixir
raise Ecto.CastError, type: type, value: value, message: "value #{inspect value} must exactly match #{inspect type}"
```

or something similar? not sure what's the best wording here.

lib/ecto/changeset.ex
michalmuskala: To be honest, I'm not sure about the "nor casting is performed" part of the docs now - they are not exactly true.

wojtekmach: I'm not sure either. FWIW we don't perform casting in the sense that we don't use the return value of `cast`, we just use cast to check types. but maybe treating it as essentially `cast!` is still considered casting.

lib/ecto/changeset.ex
kelvinst: Well, in my humble opinion this should not be a `CastError` and yes an `ArgumentError` or something similar. Since it's not a cast at all.

josevalim:
Sorry but I am still not convinced about this change. Maybe we need a new callback on Ecto.Type but I truly don't think we should cast unless explicitly asked to. Thank you.

PR #2031, opened 2017-04-12
Add :type and :value to Ecto.CastError

wojtekmach:


michalmuskala:
To be honest, it's not clear to me when we would use one over the other.

wojtekmach:
I started DRYing up Ecto.CastError usage (so we always pass `type` and `value`) and I hit this: https://github.com/elixir-ecto/ecto/blob/v2.1.4/lib/ecto/changeset.ex#L404 which didn't quite fit the other error messages as it mentions `params`

michalmuskala:
That's strange... As far as I'm aware we should always have types for all the possible fields, if there's no type, this should be an error. It might be that our tests are wrong and construct some changes manually without a regard for this. But going through `cast/4` or `change/2` should always contain type information for all the possible fields.

michalmuskala:
I guess the problematic error with wrong value passed as params could be simply made an `ArgumentError` - seems quite appropriate here.

wojtekmach:
that works for me. So we'll keep and update `Ecto.CastError` instead of introducing a new error type, right?

lib/ecto/uuid.ex
```diff
@@ -35,12 +35,12 @@ defmodule Ecto.UUID do
   def cast(_), do: :error
 
   @doc """
-  Same as `cast/1` but raises `Ecto.CastError` on invalid arguments.
+  Same as `cast/1` but raises `Ecto.Type.CastError` on invalid arguments.
   """
   def cast!(value) do
```
wojtekmach: while at it, I found myself re-implementing `cast!` on pretty much in all my custom ecto types, repeating the pattern:

```elixir
def cast!(value)
  case cast(value) do
    {:ok, value} -> value
    :error -> raise "cannot cast ... as ..."
  end
```

often. WDYT about introducing `Ecto.Type.cast!`? Then, we could have:

```elixir
defmodule Ecto.UUID do
  # ...

  def cast!(value), do: Ecto.Type.cast!(__MODULE__, value)
end
```

if sounds like worth pursuing, I can open up thread/issue/pr to flesh it out (there's a caveat that: `Ecto.Type.cast(Ecto.UUID, nil) => {:ok, nil}`, but `Ecto.UUID.cast(nil) => :error` so not sure if it can be easily made generic in Ecto.Type)

josevalim:
> I started DRYing up Ecto.CastError usage (so we always pass type and value) and I hit this: https://github.com/elixir-ecto/ecto/blob/v2.1.4/lib/ecto/changeset.ex#L404 which didn't quite fit the other error messages as it mentions params

To me that is also a casting error and the type is a map. Maybe we can keep using Ecto.CastError? and we allow both the type and value to be given but still support custom messages if desired.

wojtekmach:
> To me that is also a casting error and the type is a map. Maybe we can keep using Ecto.CastError? and we allow both the type and value to be given but still support custom messages if desired.

sounds good to me, I will look into it. It would also be backwards compatible (not sure if shape of error structs is considered part of public API, I suppose it is)

> Also note that it is preferred to build the message on the moment of raising.

you mean `raise SomeError, "foo"` is preferred than `raise SomeError, type: :foo`? As a convention? performance-wise? both?

josevalim:
> you mean raise SomeError, "foo" is preferred than raise SomeError, type: :foo? As a convention? performance-wise? both?

Sorry, I mean in `def exception` rather than in `def message`. That's because if we can't build the error, it happens at the time of building, and it is usually easier to spot.

wojtekmach:
@josevalim good call, updated!

josevalim:

PR #2044, opened 2017-04-26
Add :caller_pid to the Ecto.LogEntry struct.

asummers:
This will allow a way to reconnect the log entry with other already
accumulated state for use in e.g. metric collection of larger overall
request latency. Must check if `:caller` is already set in the case of
`preload`, which may load results in parallel and explicitly sets the
`:caller` opt.


josevalim:

PR #2047, opened 2017-04-28
suggesting valid associations when raising on non-existent association in get_assoc

mitchellhenke:
I noticed a few people asking questions in the Elixir slack recently about receiving the error for `cannot add constraint to changeset because association ``#{assoc}`` does not exist` when using `Ecto.Changeset.assoc_constraint`.  Most of the time, they were doing something like `assoc_constraint(:user_id)` when it should have been `assoc_constraint(:user)`.

It isn't necessarily clear in the error why `user` is correct and `user_id` isn't, and since the valid associations are available to suggest, I thought it could be helpful to suggest one.  I borrowed heavily from https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/exceptions.ex#L24-L34 to suggest a likely option.  It would also be possible and simpler to just list all of the associations, but I wasn't sure if that was preferable.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2127,8 +2127,15 @@ defmodule Ecto.Changeset do
     raise(ArgumentError, "cannot add constraint to changeset because it does not have a source, got: #{inspect data}")
 
   defp get_assoc(%{data: %{__struct__: schema}}, assoc) do
-    schema.__schema__(:association, assoc) ||
-      raise(ArgumentError, "cannot add constraint to changeset because association `#{assoc}` does not exist")
+    schema.__schema__(:association, assoc) || raise_invalid_assoc(schema, assoc)
+  end
+
+  defp raise_invalid_assoc(schema, assoc) do
+    associations = Enum.map_join(schema.__schema__(:associations), ", ", fn(association) ->
+      "`#{association}`"
+    end)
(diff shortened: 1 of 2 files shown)


michalmuskala:
This adds quite a lot of complexity when you usually only have just a handful of associations. Maybe we should "just" print all of them?

mitchellhenke:
@michalmuskala That makes sense to me!  I've added a commit to do that instead.

lib/ecto/changeset.ex
```diff
@@ -2127,8 +2127,16 @@ defmodule Ecto.Changeset do
...
-    schema.__schema__(:association, assoc) ||
-      raise(ArgumentError, "cannot add constraint to changeset because association `#{assoc}` does not exist")
+    schema.__schema__(:association, assoc) || raise_invalid_assoc(schema, assoc)
+  end
+
+  defp raise_invalid_assoc(schema, assoc) do
+    associations = Enum.map(schema.__schema__(:associations), fn(association) ->
+      "`#{association}`"
+    end)
+    |> Enum.join(", ")
```
josevalim: Maybe use `Enum.map_join`?

josevalim:
I have added one style related comment and this should be good to go. Thank you!

mitchellhenke:
@josevalim updated!

josevalim:

PR #2050, opened 2017-05-03
Doc addition. is_nil - show not is_nil

fmcgeough:
There have been a couple requests on Slack where people were confused about how to do IS NOT NULL. This adds to is_nil doc to point out how to do that.


josevalim:

PR #2056, opened 2017-05-10
Allow casting binary UUIDs

### LINKED ISSUE #2045: UUID.cast to allow binaries
### Environment

* Elixir version: 1.4.2
* Database and version: Postgres 9.6.1
* Ecto version: 2.1.4
* Database adapter and version: Postgrex 0.13.0
* Operating system: OSX 10.12.4

### Current behavior

I have a use case for manually generating binary UUIDs. Currently Ecto.UUID doesn't accept raw binaries even if they're valid 16 bytes. Is there any reason it shouldn't?

```elixir
defmodule File do
  use Ecto.Schema

  schema "files" do
    field :hash, Ecto.UUID, autogenerate: false
  end
end
```

```
> Repo.insert(%File{hash: :erlang.md5("file content")})
> ** (Ecto.ChangeError) value `<<209, 11, 76, 63, 241, 35, 178, 109, 192, 104, 212, 58, 139, 239, 45, 35>>` for `File.hash` in `insert` does not match type Ecto.UUID
      (ecto) lib/ecto/repo/schema.ex:699: Ecto.Repo.Schema.dump_field!/6
      (ecto) lib/ecto/repo/schema.ex:708: anonymous fn/6 in Ecto.Repo.Schema.dump_fields!/5
      (stdlib) lists.erl:1263: :lists.foldl/3
      (ecto) lib/ecto/repo/schema.ex:706: Ecto.Repo.Schema.dump_fields!/5
      (ecto) lib/ecto/repo/schema.ex:655: Ecto.Repo.Schema.dump_changes!/6
      (ecto) lib/ecto/repo/schema.ex:200: anonymous fn/13 in Ecto.Repo.Schema.do_insert/4
```

Digging into the UUID generate functions also yields some unexpected results:

```
> Ecto.UUID.generate() |> Ecto.UUID.cast()
{:ok, "ff7f95df-c6df-4743-a630-74069d82ac63"}
> Ecto.UUID.bingenerate() |> Ecto.UUID.cast()
:error
```

### Expected behavior

I'd expect Ecto.UUID to accept any valid 16 byte binary.

rockwood:
Currently, only dash-separated strings are allowed in `UUID.cast/1`. This adds the ability to pass 16 byte binaries. The same validation rules apply to both binaries and dash-separated strings.

Closes #2045

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -13,10 +13,28 @@ defmodule Ecto.UUID do
   @doc """
   Casts to UUID.
   """
-  def cast(<< a1, a2, a3, a4, a5, a6, a7, a8, ?-,
-              b1, b2, b3, b4, ?-,
-              c1, c2, c3, c4, ?-,
-              d1, d2, d3, d4, ?-,
+  def cast(<< a::64, ?-, b::32, ?-, c::32, ?-, d::32, ?-, e::96 >>) do
+    do_cast(<< a::64, b::32, c::32, d::32, e::96 >>)
+  end
+  def cast(<< _::128 >> = binary) do
+    binary |> Base.encode16(case: :lower) |> do_cast()
+  end
+  def cast(_), do: :error
+
+  @doc """
+  Same as `cast/1` but raises `Ecto.CastError` on invalid arguments.
+  """
+  def cast!(value) do
+    case cast(value) do
+      {:ok, uuid} -> uuid
+      :error -> raise Ecto.CastError, type: __MODULE__, value: value
+    end
+  end
+
+  defp do_cast(<< a1, a2, a3, a4, a5, a6, a7, a8,
+              b1, b2, b3, b4,
+              c1, c2, c3, c4,
+              d1, d2, d3, d4,
               e1, e2, e3, e4, e5, e6, e7, e8, e9, e10, e11, e12 >>) do
     << c(a1), c(a2), c(a3), c(a4),
        c(a5), c(a6), c(a7), c(a8), ?-,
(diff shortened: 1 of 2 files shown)


lib/ecto/uuid.ex
```diff
@@ -13,10 +13,28 @@ defmodule Ecto.UUID do
   @doc """
   Casts to UUID.
   """
-  def cast(<< a1, a2, a3, a4, a5, a6, a7, a8, ?-,
-              b1, b2, b3, b4, ?-,
-              c1, c2, c3, c4, ?-,
-              d1, d2, d3, d4, ?-,
+  def cast(<< a::64, ?-, b::32, ?-, c::32, ?-, d::32, ?-, e::96 >>) do
+    do_cast(<< a::64, b::32, c::32, d::32, e::96 >>)
```
josevalim: The code was written as before because it can be very performance sensitive and repacking the binary is not going to help. I would suggest keep the previous cast clause as is and introduce a new clause especially for handling raw binaries. One option is to call load() and handle the result of that.

rockwood: @josevalim Thanks for your feedback, that totally makes sense. Let me know if my update looks ok.

rockwood:
It seems that Postgres and Mysql don't actually validate that UUIDs conform to any particular spec, just that they are 16 bytes in total. If performance is a big deal here, do you think it's good enough to just validate length?

lib/ecto/uuid.ex
```diff
@@ -29,9 +29,9 @@ defmodule Ecto.UUID do
   catch
     :error -> :error
   else
-    casted ->
-      {:ok, casted}
+    casted -> {:ok, casted}
   end
+  def cast(<< _::128 >> = binary), do: binary |> encode() |> cast()
```
michalmuskala: It should be safe to do `{:ok, encode(binary)}` here - we know encode returns the proper format.

rockwood: @michalmuskala I agree, but doing that will bypass the validations from `cast_version` and `cast_variant`. Do you think that's ok?

michalmuskala: Can we introduce those checks to `encode/1`? This would also mean we'd need to change encode to return `{:ok, string} | :error`, but that shouldn't be a problem since it's a private function.

rockwood: @michalmuskala Done! Let me know if that looks ok.

josevalim:

PR #2065, opened 2017-05-18
Set schema prefix on insert, delete, and update

### LINKED ISSUE #2051: Set schema prefix on insert
Referenced from: https://github.com/phoenixframework/phoenix_ecto/issues/82

### Environment

Elixir: 1.4.2
PostgreSQL 9.6.2
Ecto: 2.1.4
Phoenix:  1.2.3
postgrex 0.13.2
OS: macOS Sierra 10.12.4

### Current behavior

Here is a factory module based off of what is in my app:

    defmodule MyApp.Factory do
      def build(:tenant) do 
         %MyApp.Tenant{name: "tenant"}
      end

      def build(:user) do
        %Scalpel.User{username: "example-#{System.unique_integer()}"}
      end
      
      def build(factory_name, attributes) do
        factory_name |> build() |> struct(attributes)
      end
      
      def insert!(:tenant) do 
        tenant = Repo.insert!(build(:tenant))
        # Runs all of the tenant specific migrations
        MyApp.TenantActions.create_tenant(Repo, tenant)
        tenant
      end

      def insert!(factory_name, attributes \\ [], prefix \\ "public") do
        Repo.insert! build(factory_name, attributes), prefix: prefix
      end
    end

    tenant = insert!(:tenant)
    user = insert(:user, [], build_prefix(tenant))

    let(:response) do 
      build_conn |> assign(:current_user, user)
    end

  In a controller we have a plug function that loads the user's profile

    plug :load_profile, "before show" when action in [:show]

    #  We get a failure here in our test because the profile table only 
    #  exists under the schema of the tenant.
    def load_profile(%{current_user: user }) do 
       user |> Ecto.assoc(:profile) |> Repo.one()
    end

    def load_profile(conn, _opts) do 
       cond do 
          conn.assigns[:profile] ->
               conn
          profile = load_profile(conn.assigns) -> 
            conn |> assign(:profile, profile)
          true -> 
              conn |> redirect(to: page_path(conn, :index)) |> halt()
       end 
    end

### Expected Behavior

My expectation is that if I have inserted the user with the tenant prefix, it should update the record with the schema it has been given.

securingsincity:
Closes #2051

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -199,7 +199,7 @@ defmodule Ecto.Repo.Schema do
         case apply(changeset, adapter, :insert, args) do
           {:ok, values} ->
             changeset
-            |> load_changes(:loaded, extra ++ values, autogen, adapter)
+            |> load_changes(:loaded, extra ++ values, autogen, adapter, opts)
             |> process_children(children, user_changeset, opts)
           {:error, _} = error ->
             error
@@ -270,7 +270,7 @@ defmodule Ecto.Repo.Schema do
           case apply(changeset, adapter, action, args) do
             {:ok, values} ->
               changeset
-              |> load_changes(:loaded, values, autogen, adapter)
+              |> load_changes(:loaded, values, autogen, adapter, opts)
               |> process_children(children, user_changeset, opts)
             {:error, _} = error ->
               error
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you!

I think a better fix would be to pass the prefix parameter directly to
load_changes. load_changes already change the metadata struct, so there is
no reason to extract it once again to handle the prefix changes. What do
you think?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

securingsincity:
@josevalim yes, that makes way more sense.

securingsincity:
Hmm it looks like only one of the builds failed but it was travis related. I can't rerun the build manually.

josevalim:

kelvinst:
I really need this change. :/

This kind of tests fails without it:

```elixir
{:ok, user} = Repo.insert(%User{name: "José"}, prefix: "auth")
assert user == Repo.get!(user.id, prefix: "auth")
```

Issue #2021, opened 2017-04-06
Column 'a_column' in field list is ambiguous when using update_all


gjaldon:
### Environment

* Elixir version (elixir -v): 1.4.2
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Ver 15.1 Distrib 10.1.22-MariaDB
* Ecto version (mix deps): 2.1.4
* Database adapter and version (mix deps): mariaex 0.8.2
* Operating system: Arch Linux

### Current behavior
I have the following code:
```elixir
Report
|> Report.by_content_type(params["type"])
|> Report.by_content_item(params["id"])
|> Repo.update_all(set: [archived_at: ^Timex.now()])
```

Running that returns the error `1052: Column 'archived_at' in field list is ambiguous`.
I checked the output query and it looks like below:
``` 
"UPDATE `reports` AS r0 INNER JOIN `contents` AS c1 ON c1.`id` = r0.`content_id` INNER JOIN `contents` AS c2 ON c2.`id` = r0.`content_id` SET `archived_at` = ? WHERE (c1.`type` = ?) AND (c2.`id` = ?)"
```

### Expected behavior
The problem is that `SET 'archived_at' =?` above doesn't specify the table. The fix is to change it to `SET r0.'archived_at' =?`. `update_all` should specify the table for the column.

josevalim:
I was going to say it should not be ambiguous because you can only update the current table... then I realized it is MySQL/MariaDB. :(

Can you please send a PR? It should be fixed in ecto/adapters/mysql/connection.ex. Thank you.

gjaldon:
Ok will submit a PR sometime next week.

josevalim:
Ping. :)

valyukov:
1 month it's a long time, so I fix this.

andrelip:
@valyukov Don't know how to set the table looking for your code. Can you provide an example?

I had to denormalize a key and in my case I solved the conflict creating a subquery.

```
    copies = from(c in "book_copies",
    select: %{id: c.id, bid: c.book_id})
    from(br in "book_rents",
    join: c in subquery(copies),
    on: br.book_copy_id == c.id,
    update: [set: [book_id: c.bid]])
```

valyukov:
@andrelip Hi there, my code just fix mysql/mariadb specific error.
Could you provide more information about what exactly you are trying to do?

andrelip:
[...]

PR #2066, opened 2017-05-18
Add table prefix for set field in update_all query (#2021)

valyukov:


josevalim:
Thank you! Can you please remove the integration test?thise are rather
expensive and I think the unit tests at SQL generation level are enough to
guarantee we are well covered. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

valyukov:
Sure, I used that for reproduce the original problem.

josevalim:

PR #2069, opened 2017-05-21
Fix Query.Builder.From.escape/1 typespec

camcaine:
Pretty sure the typespec return was back to front.


josevalim:

PR #2071, opened 2017-05-22
Add prefix constraint name checking

sb8244:
Feature request: https://groups.google.com/forum/#!msg/elixir-ecto/vwzgldHKNMM/TZzBYO9LBgAJ

Reasoning:
I am using Citus which is a Postgres extension that provides sharding capabilities. When indices are propagated to worker nodes, an integer is appended to the end of the index name. This means that I will have a normal name that Ecto can guess, but random integers at the end which differ between environments and runnings of the index being built. If Ecto supported prefix, I would set it to prefix mode and not have a custom name at all.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -269,7 +269,7 @@ defmodule Ecto.Changeset do
   }
 
   @relations [:embed, :assoc]
-  @match_types [:exact, :suffix]
+  @match_types [:exact, :suffix, :prefix]
   @actions [:insert, :update, :delete, :replace]
 
   @doc """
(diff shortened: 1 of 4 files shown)


sb8244:
To ensure sufficient coverage vs suffix:

```
➜  ecto git:(sb8244-prefix-constraints) grep -R "suffix" test
test/ecto/changeset_test.exs:    changeset = change(%Post{}) |> unique_constraint(:title, name: :whatever, match: :suffix, message: "is taken")
test/ecto/changeset_test.exs:           [%{type: :unique, field: :title, constraint: "whatever", match: :suffix, error: {"is taken", []}}]
test/ecto/repo_test.exs:    test "are mapped to repo constraint violation using suffix match" do
test/ecto/repo_test.exs:        |> Ecto.Changeset.unique_constraint(:foo, name: "custom_foo_index", match: :suffix)
```

josevalim:

sb8244:
Woot! Thanks @josevalim. Very painless process

PR #2072, opened 2017-05-22
Include case where database is missing.

schneiderderek:
I came across this error when a database had not yet been created. 

I figured adding some extra context to the message would be helpful.


josevalim:

PR #2073, opened 2017-05-23
Add functions that specify version of UUID being generated.

schneiderderek:
Since there are multiple methods/versions of UUID generation, I thought it would be useful to specify the version of UUID being generated. I kept the existing `generate` function for backwards compatibility.

Alternatively, this could also be implemented like so:
```elixir
def generate(version \\ :v4) do
  ...
end
```

Instead of creating an additional function, but I'm not sure which you guys would prefer.

Let me know if you think this is useful.

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -160,6 +160,13 @@ defmodule Ecto.UUID do
   Generates a version 4 (random) UUID.
   """
   def generate do
+    generate_v4()
+  end
+
+  @doc """
+  Generates a version 4 (random) UUID.
+  """
+  def generate_v4 do
     {:ok, uuid} = encode(bingenerate())
     uuid
   end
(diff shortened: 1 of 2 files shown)


josevalim:
I think we only need to make it clear in the docs, as right now in the code we only generate v4.

schneiderderek:
@josevalim Ok, fair enough. My thought was just to be more explicit.

michalmuskala:
Yeah, I don't think it makes sense to add the additional function given we don't intend to generate other types of UUID. Improvements to docs are welcome, though.

kelvinst:
Well, I'm with @michalmuskala in this one. But, if you still choose to make an additional function for the v4 version, I would stick with the other approach you mentioned.

josevalim:
Closing this as per the feedback.

PR #2075, opened 2017-05-25
added a pattern to schema_for_association_join! for map updates

zachdaniel:
Support was recently added for map updates in select statements from subqueries. In order to use `assoc/2` with the result of those subqueries, this pattern needs to be added to `schema_for_association_join!/3`

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -427,6 +427,8 @@ defmodule Ecto.Query.Planner do
         schema
       %Ecto.SubQuery{query: %{select: %{expr: {:&, _, [ix]}}, sources: sources}} when is_integer(ix) ->
         schema_for_association_join!(query, join, elem(sources, ix))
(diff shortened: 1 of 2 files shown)


zachdaniel:
I wasn't quite sure where to add the appropriate test case, so if someone can't point me to it I'll be more than happy to add a test case, and to update any relevant documentation.

josevalim:
@zachdaniel please check the commit that added update support and see where it added the test. I think it was in test/ecto/query/planner_test.exs. :)

zachdaniel:
This seemed kind of cross domain, so I wasn't sure if I should put this in the same place w/ the other subquery tests in `integration_test/sql/subquery.exs` or if I should put it in association tests or in the assoc tests in `integration_test/cases/assoc.exs`, or both. I'll write up what I think they would both look like and put them here.

josevalim:
I don't think we need an integration test for this change, so it needs to go in `test`. If planner is too large, then we can break it apart as part of another PR.

zachdaniel:
@josevalim Added a test case that illustrates the source being set appropriately.

josevalim:

PR #2084, opened 2017-06-02
Warn about use of match operator in the query DSL

xavier:
This change improves error handling when a match operator is used instead of a comparison operator.

See discussion:
https://groups.google.com/forum/#!topic/elixir-ecto/gYeF0jD8MPc

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -218,8 +218,8 @@ defmodule Ecto.Query.Builder do
   end
 
   def escape({:=, _, _}, _type, _params_acc, _vars, _env) do
-    error! "match operator is  not supported: `=`. " <>
-           "Instead use the comparison operator: `==`"
+    error! "match operator is not supported: `=`. " <>
+           "Did you mean to use `==` instead?"
   end
 
   def escape({op, _, _}, _type, _params_acc, _vars, _env) when op in ~w(|| && !)a do
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -217,6 +217,11 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:=, _, _}, _type, _params_acc, _vars, _env) do
+    error! "match operator is  not supported: `=`. " <>
+           "Instead use the comparison operator: `==`"
+  end
+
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder.ex
```diff
@@ -217,6 +217,11 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:=, _, _}, _type, _params_acc, _vars, _env) do
+    error! "match operator is  not supported: `=`. " <>
+           "Instead use the comparison operator: `==`"
```
josevalim: I would prefer to say: "Did you mean to use `==` instead?". That's because i don't want to assume that all cases of `=` is actually meant to be `==`. It is likely... but hard to say for sure across DBs. What do you think?

lib/ecto/query/builder.ex
```diff
@@ -217,6 +217,11 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:=, _, _}, _type, _params_acc, _vars, _env) do
+    error! "match operator is  not supported: `=`. " <>
```
josevalim: There are two spaces between "is" and "not".

lib/ecto/query/builder.ex
xavier: Sounds better indeed, it invites the developer to take a moment and think about what they're trying to achieve instead of blindly replacing `=` with `==`.

lib/ecto/query/builder.ex
```diff
@@ -217,6 +217,11 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:=, _, _}, _type, _params_acc, _vars, _env) do
+    error! "match operator is not supported: `=`. " <>
```
michalmuskala: The function error would add `Macro.to_string` representation of the whole offending part of the query. Should we do something similar here? It might be hard to find the exact point in some complex expression and this way we're able to show the surroundings of where's the issue.

xavier: How about this message?

```elixir
"`#{Macro.to_string(expr)}` uses the match operator which is not supported: `=`" <>
"Did you mean to use `==` instead?"
```

josevalim:

PR #2090, opened 2017-06-09
Support setting the default of a :map type with a Map

aseigo:
Currently, defining a :map type in a table can only have a default set with a literal string. Principle of least astonishment suggests that setting the default with an Elixir Map should be possible. With these changes this is now possible:

```
  def change do
    create table :testmap do
      add :map, :map, default: %{"foo" => "'bar'"}
    end
  end
```

### DIFF
--- a/lib/ecto/adapter.ex
+++ b/lib/ecto/adapter.ex
@@ -185,4 +185,9 @@ defmodule Ecto.Adapter do
   @callback delete(repo, schema_meta, filters, options) ::
                      {:ok, fields} | {:invalid, constraints} |
                      {:error, :stale} | no_return
+
+  @doc false
+  def json_library do
+    Application.get_env(:ecto, :json_library, Poison)
+  end
 end
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -642,6 +642,10 @@ if Code.ensure_loaded?(Mariaex) do
       do: [" DEFAULT '", escape_string(literal), ?']
     defp default_expr({:ok, literal}) when is_number(literal) or is_boolean(literal),
       do: [" DEFAULT ", to_string(literal)]
+    defp default_expr({:ok, %{} = map}) do
+      default = Ecto.Adapter.json_library().encode!(map)
+      [" DEFAULT ", [?', escape_string(default), ?']]
+    end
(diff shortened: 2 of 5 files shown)


josevalim:
@aseigo can you please add some unit tests to test/ecto/adapters/mysql_test.exs and test/ecto/adapters/postgres_test.exs? Thank you!

aseigo:
Will do! I looked at the test before submitting the patch but did not see a natural place for them (e.g. where the other code paths were already being tested) .. perhaps I missed it, so I will look again a bit more carefully this time ... otherwise I will write some tests that cover schema defaults in general. If you can point me towards existing tests for the schema default paths, that would be great of course :) 

Will try to get this over the weekend.

On June 14, 2017 5:57:20 PM GMT+02:00, "José Valim" <notifications@github.com> wrote:
>@aseigo can you please add some unit tests to
>test/ecto/adapters/mysql_test.exs and
>test/ecto/adapters/postgres_test.exs? Thank you!
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/elixir-ecto/ecto/pull/2090#issuecomment-308476724

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

josevalim:
Here is an example: https://github.com/elixir-ecto/ecto/blob/master/test/ecto/adapters/postgres_test.exs#L661-L678

You probably don't need the whole create table though,  just a new test with a create table and a single column.

aseigo:
awesome, thanks. will get right on this.

On June 15, 2017 9:41:51 AM GMT+02:00, "José Valim" <notifications@github.com> wrote:
>Here is an example:
>https://github.com/elixir-ecto/ecto/blob/master/test/ecto/adapters/postgres_test.exs#L661-L678
>
>You probably don't need the whole create table though,  just a new test
>with a create table and a single column.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/elixir-ecto/ecto/pull/2090#issuecomment-308654627

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

aseigo:
Tests are now there. They check for an empty map, a map with values, and a string default, for both postgresql and mysql.

josevalim:

Issue #1805, opened 2016-11-16
Support views in migrations


josevalim:
The idea is to introduce functions called `view/1` and `view/2` such that:

    create view(:my_view, from(u in "users", where: u.type == :student))
    drop view(:my_view)

`create` requires a query to be given to the view, `drop` does not.

We will also introduce a `create_or_replace/1` function, similar to `create/1`, but mean to be used with views.

OvermindDL1:
Definitely want that, I have a few view's in my system right now but I setup and tear them down via SQL.

lukerollans:
Will be keeping an eye on this. The frontend to one of our biggest apps relies fairly heavily on views, would be very nice to manage them natively

joshuataylor:
What about materialized views?

josevalim:
@joshuataylor we can support a `:materialized` option in `view/3` or a `materialized_view/2` function as well. I will need to study the behaviour in different databases before deciding which one works best.

eniolopes:
can i work on this one?

josevalim:
Yes, please!

hauleth:
I was just thinking about introducing it by myself. What would be useful with this one is:

* triggers
* functions

As it would allow us to create additional things that would be useful with views.

michalmuskala:
We should carefully consider if we're able in any way to provide something more usable than `execute(SQL)`. 
I'm not sure I can imagine providing something more than
```elixir
create function("""
some sql
""")
```

and I'm not sure I see advantage over

```elixir
execute """
CREATE FUNCTION some sql
"""
```

hauleth:
@michalmuskala there is one: autocreate `down` method for rollbacks. However there is only a little (if any) advantage of doing that. Maybe except allowing to reference function types using `Ecto.Type.type/0` which would be handy.

roehst:
Hi, I want to start contributing to Ecto (and Elixir in general) and I was about to suggest this feature. Is anyone working on it? Is it still in discussion? I am new to open source projects so forgive me if I don't understand part of the workflow on this. Thanks!

josevalim:
I don't think there is someone working on it right now. If you want to start contributing on this area, maybe #1832 is a better starting point, for you to get acquainted with the codebase. :)

roehst:
@josevalim Nice, I will try that. Tks

josevalim:
[...]

PR #2092, opened 2017-06-11
Support views in migrations #1805

alesasnouski:
Can we somehow rewrite update_view_sql function? We need to convert our query to raw SQL and I'm not sure that my solution is the best one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   * [Ecto.Adapters.MySQL] Use Mariaex instead of MySQL client to create & drop database
   * [Ecto.Changeset] Add `apply_action/2`
   * [Ecto.Migrations] Add reversible `execute/2` to migrations
+  * [Ecto.Migrations] Add SQL views support
   * [Ecto.Migrator] Allow migration/rollback to log SQL commands via the `--log-sql` flag
   * [Ecto.LogEntry] Add `:caller_pid` to the Ecto.LogEntry struct
   * [Ecto.Repo] Implement `:returning` option on insert
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -506,7 +506,10 @@ if Code.ensure_loaded?(Mariaex) do
 
     ## DDL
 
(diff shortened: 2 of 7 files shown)


mbuhot:
Is it possible to define the views in separate files from the migrations themselves?
This becomes necessary when views need to be dropped and re-created because an underlying table column is altered.

josevalim:
`execute File.read!("priv/migrations/view/foo.sql")`

wojtekmach:
bonus points for `foo_v1.sql`, `foo_v2.sql` etc to trivially distinguish which files goes to `up` and `down` as well as tracking how view changes over time.

mbuhot:
@josevalim Yeah that's pretty much how I manage views now. Is there an equivalent with views defined in `.exs` scripts using the `Migration` module and Ecto query syntax?

```elixir
defmodule MyMigration do
  use Ecto.Migration
  def change do
    Code.requre_file("priv/migrations/views/foo_v1.exs")
    create Foo.V1.view()
  end
end
```

with the view module defined like:

```elixir
defmodule Foo.V1 do
  use Ecto.Migration
  def view do
    Ecto.Migration.view(query: from(p in "posts", select: p.name), prefix: "baz")
  end
end
```

josevalim:
@mbuhot that will work too. You could define multiple views on the same file, so you don't need multiple requires. Versioning can be done with function names.

josevalim:
Thanks @alesasnouski. I have tried this patch on a side-project and I started to run into issues that show the approach of using queries is unfortunately flawed because Ecto may take some liberties when building queries, especially if subqueries are involved. To make matters worse if we start generating `AS` in `select` to internal purposes, that will affect views too.

I am considering going with a variant that is based on files, as we mentioned above. It is also worth noting that the new `execute/2` simplify those scenarios a bit. To create:

    execute File.read!("priv/repo/views/foo_v1.sql"), "DROP VIEW foo"

To update:

    execute File.read!("priv/repo/views/foo_v2.sql"), File.read!("priv/repo/views/foo_v1.sql")

It is a bit verbose but I am more inclined to build helpers on this direction than on top of Ecto.Query. Apologies for the false positive and thank you for your contributions!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #2101, opened 2017-06-22
[WIP] Always log from the main process in parallel preloads

michalmuskala:
This allows preserving process metadata when logging

---

This is just a proof of concept. It requires some tests and probably additional documentation. To be honest, I'm not 100% sure how to test this.

\cc @josevalim @fishcakez

### DIFF
--- a/lib/ecto/adapter.ex
+++ b/lib/ecto/adapter.ex
@@ -190,4 +190,13 @@ defmodule Ecto.Adapter do
   def json_library do
     Application.get_env(:ecto, :json_library, Poison)
   end
+
+  def log(repo, log, opts) do
+    case Keyword.fetch(opts, :forward_logs) do
+      {:ok, pid} when is_pid(pid) ->
+        send(pid, {:ecto_log, repo, log})
+      :error ->
+        repo.__log__(log)
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


fishcakez:
If I understand this correctly we want to log in the calling process to use the Logger metadata from the calling pid? I am not keen on this approach because it only works for parallel preloads, it does not work for those who require a similar mechanism, such as absinthe_ecto's batching. On master we have `caller_pid` so at least the caller pids is known :S, it is not ideal.

I think we might want a unique reference to make it easier to prevent message leaking. I wonder why we don't use Task.async_stream here...

I think it would be nice to wait until phoenix metrics is out, as it would convenient to be built into that approach. Thoughts?

michalmuskala:
Yes, it's a solution only for this case, so that's a bit unfortunate. On the other hand, I can't think of a generic solution that wouldn't involve changes to how Logger works. I think it's still worth considering it - it will, at least, make ecto itself consistent.

Messages should not leak - we're guaranteed to always get the task result after we get the log from the particular task.

We can definitely wait, if there's a possibility of having a better way to solve this. I think most metrics libraries have ways to work around this issue in place already.

fishcakez:
We can have situation where user traps exits and catches exit, then if
another task exits we leak a wierd message. I.e use {ref, _} format so that
it's known catch exit/leak message.

fishcakez:
Will also copy the result twice :(

michalmuskala:
Ah. Copying the result is bad

PR #2104, opened 2017-06-25
`execute/1` isn't specific to DB type

toraritte:
This way the docs for `execute/1` is consistent with `execute/2` where DB type is also not mentioned.


josevalim:

PR #2107, opened 2017-06-27
multi-column unique index doc updated

murb:
A name was used that came out of the blue. Since there is a convention ( “#{table}_#{Enum.join(columns, "_")}_index” ), this is used in this updated example.


michalmuskala:
Thank you! :heart:

PR #2108, opened 2017-06-27
Ensure that Ecto.Repo.Queryable.transaction accepts function or multi

wojtekmach:
I accidentally passed changeset to Repo.transaction and got a weird error:

     ** (FunctionClauseError) no function clause matching in Ecto.Multi.__apply__/4
     stacktrace:
       (ecto) lib/ecto/multi.ex:392: Ecto.Multi.__apply__(#Ecto.Changeset<action: nil, ...>)
       (ecto) lib/ecto/repo/queryable.ex:21: Ecto.Repo.Queryable.transaction/4
       ...

FunctionClauseError on `Queryable.transaction/4` seems more appropriate (especially later with Exception.blame)?

### DIFF
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -14,7 +14,7 @@ defmodule Ecto.Repo.Queryable do
     adapter.transaction(repo, opts, fun)
   end
 
-  def transaction(adapter, repo, multi, opts) do
+  def transaction(adapter, repo, %Ecto.Multi{} = multi, opts) do
     wrap   = &adapter.transaction(repo, opts, &1)
     return = &adapter.rollback(repo, &1)
 


josevalim:

michalmuskala:
This breaks dialyzer since multi is an opaque struct :( We should either revert this and make sure `Ecto.Multi.__apply__` has a nicer error or add a dialyzer exception.

wojtekmach:
ugh, I know see I basically reverted your change [1], sorry for that!

[1] https://github.com/elixir-ecto/ecto/commit/12f27d9ec25d035fc2f71c25052c2cae6c5eb781?diff=unified#diff-ed753f133e94caf75aab64df9433dbb8L17

michalmuskala:
I think adding a dialyzer exception is going to be the best way to solve this without sacrificing good error messages

PR #2110, opened 2017-06-29
Minor refactoring in Ecto.Schema.__timestamps__/2

wojtekmach:
Ecto 2.2 depends on Elixir v1.4 so it's safe to do.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1388,13 +1388,11 @@ defmodule Ecto.Schema do
   ## Callbacks
 
   @doc false
-  # TODO: Use utc_now instead.
   def __timestamps__(:naive_datetime, :seconds) do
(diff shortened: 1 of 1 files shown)


lib/ecto/schema.ex
```diff
@@ -1388,13 +1388,11 @@ defmodule Ecto.Schema do
   ## Callbacks
 
   @doc false
-  # TODO: Use utc_now instead.
   def __timestamps__(:naive_datetime, :seconds) do
-    NaiveDateTime.from_erl!(:erlang.universaltime, 0)
+    %{NaiveDateTime.utc_now() | microsecond: {0, 6}}
```
wojtekmach: setting it as `{0, 0}` would break some pg integration tests: https://gist.github.com/wojtekmach/7e848d91881cb457395c469f6b14850a

josevalim:

Issue #2091, opened 2017-06-10
Incorrect documentation for Ecto.Changeset.cast/3


shankardevy:
The documentation for `Ecto.Changeset.cast/3` says

> The given data may be either a changeset, a struct or a {data, types} tuple. The second argument is a map of params that are cast according to the type information from data. params is a map with string keys or a map with atom keys containing potentially unsafe data.

However, the function doesn't accept _struct_ as input.  It should be _schema_ instead. 

```elixir

iex> defmodule Post do        
...    defstruct [:title, :body]  
...  end
iex> Ecto.Changeset.cast %Post{}, %{title: "hello", body: ""}, [:title]
** (UndefinedFunctionError) function Post.__changeset__/0 is undefined or private
           Post.__changeset__()
    (ecto) lib/ecto/changeset.ex:422: Ecto.Changeset.do_cast/4
```
The offending code is https://github.com/elixir-ecto/ecto/blob/v2.1.4/lib/ecto/changeset.ex#L422

When a struct is given, the function expects the struct's module to implement a `__changeset__` function which is defined when using `Ecto.Schema`. 

Can I submit a PR for the docs?

michalmuskala:
Yes, I think the corrected text should read "a struct of an ecto schema" instead of "a struct"

shankardevy:
It looks this is more complicated than a simple text change here. I noticed in several places that Ecto documentation is using the term _struct_ when it is actually a module that uses `schema` macro.  Technically having _struct_ in the docs is incorrect and is misleading. For eg., `Repo.insert/2` says 

> Inserts a struct or a changeset.

It cannot insert a struct. Struct has no information about the table name to store data.

However, having to replace _struct_ everywhere with  _struct created by schema macro_  is too verbose and repetitive. Other option is to use `schema` but again it's technically incorrect, though it is short. It can be `ecto schema module`, though. Any thoughts?

josevalim:
I would say `struct` is fine because in the context it makes sense to be a struct backed by a schema. But being clearer is always welcome, so we can also say `schema struct`. I would agree that calling it `schema` is confusing.

josevalim:
PR please? :)

josevalim:
Closing in favor of the PR.

PR #2112, opened 2017-06-30
WIP: Rename struct to schema struct

shankardevy:
WIP for issue https://github.com/elixir-ecto/ecto/issues/2091

I am not sure if I should go ahead and rename every instance of `struct` where it technically means a `schema struct`?  In this PR I have replaced several instances of `struct` to `schema struct`. 
However, I haven't touched the variable names used in code. Should I this be also changed? I feel that would make it too verbose but not sure.

Any decision made here needs to be carried to other modules such as `Ecto.Repo` which also has references to `struct`.

Please advise.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -359,7 +359,7 @@ defmodule Ecto.Changeset do
   Applies the given `params` as changes for the given `data` according to
   the given set of keys. Returns a changeset.
 
-  The given `data` may be either a changeset, a struct or a `{data, types}`
+  The given `data` may be either a changeset, a schema struct or a `{data, types}`
   tuple. The second argument is a map of `params` that are cast according
   to the type information from `data`. `params` is a map with string keys
   or a map with atom keys containing potentially unsafe data.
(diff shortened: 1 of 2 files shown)


josevalim:
The changeset is not a good place for doing this change because it actually allows regular structs to be used. Ecto.Repo would be a good place for this though.

josevalim:
Also, we don't need to change all mentions of structs. We just need to have a disclaimer at some points that says something like: "This module (or this function) expects a schema struct, i.e. a struct that is defined via Ecto.Schema".

shankardevy:
@josevalim Changeset still needs a schema struct for `cast` function, IMHO. So I have changed the documentation in 3 places to keep the changes to minimum. Please check.

lib/ecto/changeset.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Ecto.Changeset do
   @moduledoc ~S"""
   Changesets allow filtering, casting, validation and
-  definition of constraints when manipulating structs.
+  definition of constraints when manipulating structs that are defined via `Ecto.Schema`.
```
josevalim: Can we please remove this mention? As we also support regular structs as long as you give the types.

josevalim:
One last comment and we should be good to go!

shankardevy:
@josevalim I have modified the commit as you have mentioned. Please check now.

josevalim:

Issue #1763, opened 2016-10-26
`cast_embed` uses schema types, not a changeset types


AndrewDryga:
Hello,

There are method [`Ecto.Relation.do_cast/4`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset/relation.ex#L92) that uses `__schema__` metadata, but I think it would be a good idea to extract type data from `changeset.types`. 

Thus we will make it possible to hack type system and, if developer want, replace type in `changeset/2` functions. It it suitable when there are many different embeds in a same schema structure. (It would require many copy-paste schemas.)

I have real world example: 
1. Different peer-to-peer transfer peers is defined as custom type: https://github.com/Nebo15/tokenizer.api/blob/master/lib/repo/schemas/peer.ex#L13
2. Type itself can resolve what kind of data in it and cast/dump it to map: https://github.com/Nebo15/tokenizer.api/blob/master/lib/repo/types/peer_credential.ex#L10
3. Custom cast for this types: https://github.com/Nebo15/tokenizer.api/blob/master/lib/repo/changeset/dynamic_embed.ex#L42

Right now only issue left is that when saving to DB this data doesn't get dumped from Changeset struct to map (`changeset.changes.credential` stores changeset that is ignored by Ecto, because its defined as embed type only in changeset.)

```
"encode map"
%{credential: #Ecto.Changeset<action: nil,
   changes: %{number: "5473959513413611"}, errors: [],
   data: #Tokenizer.DB.Models.CardNumber<>, valid?: true>,
  email: "ivan@example.com", phone: "+380631112233"}
23:17:02.283 [error] #PID<0.466.0> running Tokenizer.HTTP.Endpoint terminated
Server: localhost:4001 (http)
Request: POST /payments
** (exit) an exception was raised:
    ** (Poison.EncodeError) unable to encode value: {:number, :credit_card}
...
```

josevalim:
Yes, we should always use the types information if one is available. A PR is welcome.

AndrewDryga:
Ok, I'll look into it in few days :).

*For people that found this issue in Google*: There are easy way how to deal with this issue, you simply need to call `Ecto.Changeset.apply_changes/1` before saving data to database, and all castings will be done with extraction of this data from `changeset.types`.

josevalim:
Ping? :)

AndrewDryga:
Pong :).

[...]

PR #2113, opened 2017-07-01
Stop using `__schema__` to resolve assocs in Ecto.Changeset

AndrewDryga:
Ecto.Changeset should use information from type when available.

This is a first step in implementing Dynamic Embeds as discussed in #1763. I'm splitting all work to have many small changes, rather than one big.

In next PR I'll try to reduce the number of `__schema__` calls in Ecto.Repo.Schema, so Changesets that were dynamically created will be able to be inserted into DB.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2154,15 +2154,10 @@ defmodule Ecto.Changeset do
   end
 
   defp raise_invalid_assoc(types, assoc) do
-    associations =
-      Enum.reduce(types, "", fn
-        {_key, {:assoc, %{field: field}}}, "" -> "`#{field}`"
-        {_key, {:assoc, %{field: field}}}, acc -> acc <> ", `#{field}`"
-        _, acc -> acc
-      end)
 
+    associations = for {_key, {:assoc, %{field: field}}} <- types, do: field
     raise ArgumentError, "cannot add constraint to changeset because association `#{assoc}` does not exist. " <>
-                         "Did you mean one of #{associations}?"
+                         "Did you mean one of `#{Enum.join(associations, "`, `")}`?"
   end
 
   @doc ~S"""
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1902,7 +1902,7 @@ defmodule Ecto.Changeset do
       create unique_index(:users, [:email, :company_id])
 
   Because such indexes have usually more complex names, we need
-  to explicitly tell the changeset which constraint name to use (here we're 
+  to explicitly tell the changeset which constraint name to use (here we're
   using the naming convention that `unique_index` uses):
 
       cast(user, params, [:email])
@@ -2132,7 +2132,6 @@ defmodule Ecto.Changeset do
   defp no_assoc_message(:one), do: "is still associated with this entry"
   defp no_assoc_message(:many), do: "are still associated with this entry"
 
-
   defp add_constraint(changeset, type, constraint, match, field, error)
        when is_binary(constraint) and is_atom(field) and is_tuple(error) and is_atom(match)  do
     unless match in @match_types, do: raise(ArgumentError, "invalid match type: #{inspect match}. Allowed match types: #{inspect @match_types}")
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -2145,16 +2144,25 @@ defmodule Ecto.Changeset do
...
-    end)
-    msg = "cannot add constraint to changeset because association `#{assoc}` does not exist. Did you mean one of #{associations}?"
-    raise(ArgumentError, msg)
+  defp raise_invalid_assoc(types, assoc) do
+    associations =
+      Enum.reduce(types, "", fn
+        {_key, {:assoc, %{field: field}}}, "" -> "`#{field}`"
+        {_key, {:assoc, %{field: field}}}, acc -> acc <> ", `#{field}`"
+        _, acc -> acc
+      end)
```
michalmuskala: Let's use something like the following here.
```elixir
associations = for {_key, {:assoc, %{field: field}}} <- types, do: field
raise ArgumentError, "..." <>
  "Did you mean one of #{Enum.join(associations, ", ")}"
```

AndrewDryga:
Should we also use information from changeset for [`Ecto.Repo.Schema.do_insert/4`](https://github.com/AndrewDryga/ecto/blob/60cd576e8410760c0b208ad80a8b6e689f4d2787/lib/ecto/repo/schema.ex#L170) and `Ecto.Repo.Schema.do_update/4`?

Most of type information is already in changeset.

Also, we can get rid from more `__schema__` calls by adding `:relation_primary_keys` field to the `Ecto.Changeset.Relation`. It should carry list of relation primary keys and their type information. Here are calls that can be replacaed:
1 . https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset/relation.ex#L93 
2. https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset/relation.ex#L127
3. https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset/relation.ex#L282
4. https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset/relation.ex#L373

I did not find any good ways to insert them, because it is not possbile to call `related.__schema__(:primary_key)` in `Ecto.Association.Has.struct/3`. Because related schema sometimes is not compiled when this code is executed. Any ideas?

josevalim:
> I did not find any good ways to insert them. Looks like it is not possbile to call related.__schema__(:primary_key) in Ecto.Association.Has.struct/3, because related schema sometimes is not compiled when this code is executed.

If we do that, we will require ordering between schemas and we may also run into deadlocks when compiling code. So unfortunately, it cannot be done. I think we should aim to remove `__schema__` from changeset but when it goes into associations and repo, there is no need to worry about removing it, since those certainly rely on a schema.

AndrewDryga:
@josevalim this PR removes all `__schema__` calls that can be replaced with types information, rest of them require us to add more data to the Changeset struct, especially `Changeset.Relation` struct (you can see links above).

Can you give more ideas what's left to support schemaless changesets? I've written a simple test and looks loke they work, even with dynamic embeds. (But they can not be inserted into DB because of `Repo.Schema` `__schema__` calls.)

josevalim:
> Can you give more ideas what's left to support schemaless changesets?

Just to clarify: we already support schemaless changesets but they cannot be used against the database. So what you are describing is how it is intended to work. If your dynamic embeds tests work, then I would just contribute them to Ecto's test suite and call it a day. :)

josevalim:

1player:
@AndrewDryga is the "next PR that supports inserting dynamic embeds" you reference ready yet?

This PR works great when casting a dynamic embed, but Repo.insert / Multi breaks when trying to insert the dynamically casted embed, saying:

```
** (Ecto.ChangeError) value `#Ecto.Changeset<action: :insert, changes: %{url: "http://example.com/feed.xml"}, errors: [], data: #DynamicEmbed<>, valid?: true>` for `OC.Feeds.Feed.dynamic_field` in `insert` does not match type :map
```

Basically, Repo.insert is not converting the dynamic changeset into an actual map (even if Ecto.Changeset.apply_changes does correctly) before writing to database.

I'll try and work around it by calling apply_changes myself before Repo.insert

AndrewDryga:
@1player unfortunately I ended up understanding that it's can't be done without carrying to many knowledge in the Changeset, so PR was just closed. But mentioned workaround worked for me.

axelson:
How would you handle errors with the workaround? It seems like you'd need to collapse them into a single string which would be very difficult to parse back out.

1player:
@axelson This is how I'm handling it in my project.

On the schema (called Source), where `type` decides the inner type on the `options` map:
```
  def create_changeset(params) do
    changeset =
      %__MODULE__{}
      |> cast(params, [:type])
      |> validate_required([:type])
      |> apply_options_schema_type()

    if changeset.valid? do
      cast_embed(changeset, :options, required: true)
    else
      changeset
    end
  end

 # Update the :options field type to be an embedded changeset,
  # and cast using the options_schema changeset function
  # From https://github.com/elixir-ecto/ecto/blob/7a165571bcb19c701e573eb1f4eb0820f2b9695b/test/ecto/changeset_test.exs#L129
  defp apply_options_schema_type(changeset) do
    type = get_field(changeset, :type)

    case Sources.options_schema(type) do
      {:ok, options_schema} ->
        Map.update!(changeset, :types, &Map.put(&1, :options, options_type(options_schema)))

      :error ->
        add_error(changeset, :type, "Unknown source type")
    end
  end

  defp options_type(options_schema) do
    {:embed,
     %Ecto.Embedded{
       cardinality: :one,
       field: :options,
       on_cast: &options_schema.changeset(&1, &2),
       on_replace: :update,
       owner: nil,
       related: options_schema,
       unique: true
     }}
  end
```

And on the context:
```
def create_source(params) do
    changeset = Source.create_changeset(params)

    # Work around for https://github.com/elixir-ecto/ecto/pull/2113#issuecomment-361577933
    with {:ok, source_from_changeset} <- Ecto.Changeset.apply_action(changeset, :insert) do
      workaround_changeset =
        case source_from_changeset.options do
          nil ->
            changeset

          _ ->
            options_as_map = Map.from_struct(source_from_changeset.options)
            %{source_from_changeset | options: options_as_map}
        end
      Repo.insert(workaround_changeset)
    end
```

Hope this helps. I've adapted the code a bit for this comment, so sorry if I've left something off or there's any mistake.

PR #2114, opened 2017-07-01
updates the documentation of load and dump with shell command deps

DianaOlympos:
Following https://github.com/elixir-ecto/ecto/pull/1939, the only use of the shell client is for the Mix Tasks `ecto.load` and `ecto.dump`. This was no documented, leading to some problems. This updates these Tasks with information about this dependency.

I am open to any idea or change for the wording of course 😃


josevalim:

PR #2115, opened 2017-07-02
Add unlogged option to create table

IanVaughan:
This is only applicable to Postgres adaptor.
UNLOGGED optiona prevents Postgres using the WAL :

> Data written to unlogged tables is not written to the write-ahead log
> which makes them considerably faster than ordinary tables.
-- https://www.postgresql.org/docs/9.1/static/sql-createtable.html

----

Problem with this commit is that by adding to Ecto.Migration.Table
means it needs to work on all adapters (I would guess).
We cannot use `options` either as that raises if keyword list is used
(as its used for generic text to add to column SQL commands).

Ideally this would work :

    create table(:test_unlogged, options: [unlogged: true])

I welcome feedback on direction this could go in to be accepted.

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -233,6 +233,16 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule UnloggedMigration do
+    use Ecto.Migration
+
+    def change do
+      create table(:default)
+      create table(:unlogged_true, unlogged: true)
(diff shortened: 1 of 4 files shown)


IanVaughan:
I cannot think of a good way to fit this in, and as its quite specific, and just as easy to update after, that maybe is the better solution!
```
execute("ALTER TABLE name SET UNLOGGED")
```

PR #2118, opened 2017-07-06
Fix spec for Ecto.Adapters.SQL.query/4 etc

### LINKED ISSUE #2117: Spec for Ecto.Adapters.SQL.query incorrect for postgrex
### Environment

* Elixir version (elixir -v):  1.4.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):  PostgreSQL 9.6.1
* Ecto version (mix deps): 2.1.4
* Database adapter and version (mix deps): postgrex 0.13.2
* Operating system: mac OS

### Current behavior

The spec is:

```
query(Ecto.Repo.t, String.t, [term], Keyword.t) ::
  {:ok, %{rows: nil | [tuple], num_rows: non_neg_integer}} |
  {:error, Exception.t}
```

However, postgrex queries return a `Postgrex.Result` struct, e.g.,

```
{:ok,
 %Postgrex.Result{columns: ["max"], command: :select, connection_id: 72712,
  num_rows: 1, rows: [["20170516150906"]]}}
```

Notably, the `rows` value is a list of lists rather than a list of tuples.  This causes dialyzer to fail any match on the returned rows.

### Expected behavior

Without digging, I'm not sure if just the spec needs to be relaxed (`rows: nil | [term]` would work, I think) or if the code needs to be modified to convert the result to a list of tuples.   If the former, I would be glad to submit a PR.

I have the postgrex version overridden to 0.13.2 b/c I need features in the newer version.  The ecto mix.exs wants 0.13.0.  I checked the docs for 0.13.0, 0.13.2, and 0.13.3, and they all say that they return a list for `rows`.

dantswain:
This should fix #2117

I took the actual spec for rows from Postgrex.Result.  I've verified
that this change solves the problem I ran into.

I tried to update the documentation where relevant.  Let me know if I
missed anything.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -181,7 +181,8 @@ defmodule Ecto.Adapters.SQL do
   Same as `query/4` but raises on invalid queries.
   """
   @spec query!(Ecto.Repo.t, String.t, [term], Keyword.t) ::
-               %{rows: nil | [tuple], num_rows: non_neg_integer} | no_return
+               %{rows: nil | [[term] | binary], num_rows: non_neg_integer}
+               | no_return
   def query!(repo, sql, params \\ [], opts \\ []) do
     query!(repo, sql, map_params(params), fn x -> x end, opts)
   end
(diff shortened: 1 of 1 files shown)


josevalim:

PR #2122, opened 2017-07-11
Custom Associations

mgwidmann:
Ecto has several association types which all implement via the `Ecto.Association` behavior. I'm not sure if this is intended to be something that external libraries can implement, but I've been trying.

My goal is to support a legacy Rails database with polymorphic relationships. I know, ecto has chosen not to support polymorphism and I'm fine with that, I agree its a problematic structure. However, I do believe that even if Ecto doesn't support it out of the box, **it should be supportable via a 3rd party library or a gist**. However, my attempts at doing so have proven to be problematic.

Following is my gist which with this pull request makes `Repo.one/1` fetches w/ preloading works with polymorphism, however `Repo.all/1` calls with polymorphism don't work since the code is expecting to execute a single query to get a single relationship (when polymorphic relationships can execute 1 or more queries). If we can make a list of queries get executed returned from `assoc_query/3` that will solve this issue.

https://gist.github.com/mgwidmann/ad2ef5138378dc3772a321567610520d

If you have any better ideas how to support this, I'm all ears. Again, I know Ecto does not intend to ever support polymorphism, I'm only asking for compatibility to allow it to be extended to do so. The purpose for this is only to support legacy systems.

### DIFF
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -497,14 +497,17 @@ defmodule Ecto do
     end
 
     schema = hd(structs).__struct__
-    assoc = %{owner_key: owner_key} =
+    assoc = %{owner_key: owner_key, assoc_query_receives_structs: structs?} =
       Ecto.Association.association_from_schema!(schema, assoc)
 
-    values =
+    values = if structs? do
+      structs
+    else
       Enum.uniq for(struct <- structs,
         assert_struct!(schema, struct),
         key = Map.fetch!(struct, owner_key),
         do: key)
+    end
 
     Ecto.Association.assoc_query(assoc, assocs, nil, values)
   end
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -464,7 +464,7 @@ defmodule Ecto do
   defp drop_meta([_|_] = attrs), do: Keyword.drop(attrs, [:__struct__, :__meta__])
 
   @doc """
-  Builds a query for the association in the given struct or structs.
+  Builds query/queries for the association in the given struct or structs.
 
   ## Examples
 
@@ -503,8 +503,8 @@ defmodule Ecto do
     values =
       Enum.uniq for(struct <- structs,
         assert_struct!(schema, struct),
-        key = Map.fetch!(struct, owner_key),
-        do: key)
+        _id = Map.get(struct, owner_key), # Filter out structs without owner_key data
+        do: struct)
 
     Ecto.Association.assoc_query(assoc, assocs, nil, values)
   end
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -497,17 +497,14 @@ defmodule Ecto do
     end
 
     schema = hd(structs).__struct__
-    assoc = %{owner_key: owner_key, assoc_query_receives_structs: structs?} =
+    assoc = %{owner_key: owner_key} =
       Ecto.Association.association_from_schema!(schema, assoc)
 
-    values = if structs? do
-      structs
-    else
+    values =
       Enum.uniq for(struct <- structs,
         assert_struct!(schema, struct),
         key = Map.fetch!(struct, owner_key),
         do: key)
-    end
 
     Ecto.Association.assoc_query(assoc, assocs, nil, values)
   end
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -490,13 +490,15 @@ defmodule Ecto.Association.Has do
   end
 
   @doc false
-  def assoc_query(%{queryable: queryable, related_key: related_key}, query, [value]) do
+  def assoc_query(%{queryable: queryable, related_key: related_key, owner_key: owner_key}, query, [value]) do
+    value = Map.get(value, owner_key)
     from x in (query || queryable),
       where: field(x, ^related_key) == ^value
   end
 
   @doc false
-  def assoc_query(%{queryable: queryable, related_key: related_key}, query, values) do
+  def assoc_query(%{queryable: queryable, related_key: related_key, owner_key: owner_key}, query, values) do
+    values = Enum.map(values, &(Map.get(&1, owner_key)))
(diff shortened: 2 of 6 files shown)


michalmuskala:
Right now the `Ecto.Association` module is private to ecto and we don't guarantee it won't change (in fact it already has between 2.1 and current master). I'm not sure how I feel about exposing this.

mgwidmann:
🤔 hmm I see. Any suggestions for supporting a legacy database w/ polymorphism? The plan would be to move away from polymorphism but its just a large effort to do it all at once...

michalmuskala:
One way I could see doing it would be through views. You could create a view per each association type that you have with a constraint on the type. Using Postgres such views are even writable. With those, you could use the elixir polymorphic features where you have a separate table per type as described in the docs.

mgwidmann:
I'd really prefer not to alter the database schema if possible... So is this PR dead then? not possible to support polymorphism in a 3rd party gist or package?

michalmuskala:
I think exposing those internals will be limiting for Ecto, especially that I feel like this area could really use some love & potentially a rewrite.

I'm also not sure the code changes you provided here would be actually enough - one area I'm really worried about is changesets and nested associations. If we are going to introduce this, we'd probably need a concrete use case & some tests.

mgwidmann:
Well I think the concrete use case is supporting legacy schemas w/ polymorphism. While I think its fine ecto chooses not to support it out of the box, I'd like a third party library or gist to be able to enable it.

Is this something ecto would prefer to just close the door on completely or can it be rebuilt to be more extendable?

josevalim:
I don't think Ecto.Associations need to rebuilt. The API was designed from scratch to be extensible. The difficulty, however, is knowing when it becomes stable, as it is driven by internal needs.

However, I am completely ok with making the Ecto.Associations public with the note that it is "unstable" until we feel comfortable enough to freeze it.

Therefore, this change is welcome, although I am not happy with the current implementation / direction. It feels we are adding a specific option only to support polymorphic belongs to, instead of a generic API that would apply to other cases. I would simply change `assoc_query` to always receive structs. The association can then extract the information it needs.

This would be the best change and it brings us back exactly to the point I was trying to make above. New knowledge will force us to break the API as in cases like this, which makes me uncomfortable to say it is public and stable.

josevalim:
However, to be fair, I am not even sure the change above would be enough. Don't you need to perform multiple queries when preload polymorphic belongs to? If that's the case, then the best option is to add a `preload` function to Ecto.Association that does the actual preloading and move the existing `fetch_query` implementation that calls assoc_query to the association (except the unzip_ids call).

mgwidmann:
@josevalim yeah thats where I got stuck and opened the PR... it would mean the code which called `assoc_query` would need to handle running a list of queries instead of just a single one, something that sounds much simpler than it actually is in practice.

With your idea that would push preloading completely to the association is that correct? I'll give this another look with that direction and see where that lands us...

josevalim:
assoc_query just cannot work on polymorphic belongs to. Even Rails fails on
the equivalent behavior. So we need to abstract it on the preload part,
yeah.
-- 


mgwidmann:
I'm not sure I understand fully you're direction. I've reverted my changes and started over so the `Ecto.Association` type is not modified. I'll rebase and merge the commits into one once everything is complete.

I've changed `assoc_query` to always return a list of queries and allowed `Repo.all` and `Repo.one` to run a list of queries (for `Repo.one` this is just for compatibility, it still requires one item to be returned so it should always be a list of one item).
Additionally, `assoc_query` always receives the structs directly now instead of the IDs.

I'm working on `preload_info` now as it seems its not called unless `all` is used. It too, will need to receive the list of structs as arguments to be able to determine which relation (or many) to return. I don't have this part working as of yet, but it needs to also return a list of `assoc`s and the downstream code needs to handle that (which is what I'm working on now).

Can you explain in a bit more detail how you think it this should work? Am I headed in the right direction?

josevalim:
@mgwidmann assoc_query should be a single query because you should be able to compose on it: `Ecto.assoc(post, :users) |> where(...)`. Returning a list is a breaking a change and we definitely wouldn't accept it.

My point was actually the opposite, assoc_query won't work for polymorphic associations, and they don't even work in Rails. If you try to use `joins` in Rails with a polymorphic association, which is what assoc_query effectively is, you will get an error.

josevalim:
If you are unclear on what are the next steps or the proper APIs, I can suggest to tackle the problem the way I usually do. First I would add a bunch of conditionals to the code that explicitly check for the polymorphic associations and leave a bunch of TODOs. Once the code works, then you look at the points you added conditionals and decide what are the best contracts.

I prefer to do that because, If you simply try to extend the existing contracts at a meal peace fashion, you may end-up with a solution that is inconsistent or too fragmented.

josevalim:
Hi @mgwidmann! Any news on this front?

mgwidmann:
Yeah I haven't had come back to it, though I still think its something that needs to be addressed. I guess I'm not entirely clear on what the expected API from a user standpoint would be still. I'm a bit confused about how support for a polymorphic relationship would work when functions like (as you explained `Ecto.assoc(post, :users) |> where(...)` has to return a single query, but how could it if the user relationship were split between a `users` table and an `admins` table for example.

josevalim:
> I'm a bit confused about how support for a polymorphic relationship would work when functions like (as you explained Ecto.assoc(post, :users) |> where(...) has to return a single query

From what you described, you are porting a legacy Rails application. The construct above is not possible in Rails as well so there is probably no reason to bring it to your port of polymorphic associations.

mgwidmann:
Right but since `preload` uses `Ecto.assoc/2` under the hood, how would `from p in Post, preload: [:user]` be expected to work then?

josevalim:
We will need to stop using Ecto.assoc :)

mgwidmann:
Closing in favor of new direction in #2293

PR #2125, opened 2017-07-12
Add support for clauses with nil values in `Repo.get_by(!)/2`

tlux:
Hello Everyone!

I've just been trying to use `Repo.get_by/2` with a `nil` value in one of the clauses and encountered the following error:

```elixir
Repo.get_by(Price, %{organization_id: nil, type: "default"})
```

```
** (ArgumentError) nil given for :organization_id. Comparison with nil is forbidden as it is unsafe. Instead write a query with is_nil/1, for example: is_nil(s.organization_id)
```

However, this PR solves the issue. Any feedback appreciated! Thanks for all the great work with Ecto! :+1:

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -563,6 +563,7 @@ defmodule Ecto.Integration.RepoTest do
   test "get_by(!)" do
     post1 = TestRepo.insert!(%Post{title: "1", text: "hai"})
     post2 = TestRepo.insert!(%Post{title: "2", text: "hello"})
+    post3 = TestRepo.insert!(%Post{title: "3", text: nil})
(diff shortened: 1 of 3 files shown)


josevalim:

josevalim:
This unfortunately has security implications I did not consider when I merged the pull request, apologies. This will be reverted in the upcoming 2.2.1 release.

tlux:
Could you please give some details on that?

josevalim:
It is the whole reason why we ask `nil` to be explicitly checked, otherwise someone requesting your API using JSON can pass a field as nil and force your application to do a query it was not intended to. For example, someone can force a token field to be nil.

Issue #1874, opened 2016-12-27
update_change behavior on required field


almassapargali:
Hello

### Environment

* Elixir version (elixir -v): 1.3.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.4.5
* Ecto version (mix deps): 2.1.0-rc.1
* Database adapter and version (mix deps): postgrex 0.12.1
* Operating system: macOS

Currently, docs for `update_change` stats:

```
The given `function` is invoked with the change value only if there
  is a change for the given `key`. Note that the value of the change
  can still be `nil` (unless the field was marked as required on `cast/3`).
```

1. We are not passing required through `cast` functions anymore.
2. Even with `validate_required` before `update_change` I'm getting nil value in update function.

I think we either should update docs that it can be `nil` even for required fields, or make function passed to `update_change` run only when value is not nil. Let me know what should we do, and I'll make a PR with changes.

josevalim:
I think we should change `validate_required` and `validate_change` to remove the field from `changes` if it does not pass validation. It may solve many of the problems with checking for `valid?: true` in changeset. @michalmuskala, @wojtekmach, what do you think?

wojtekmach:
+1 for removing field from `changes` when it doesn't pass validation. The incorrect (blank) field value will be still in `params` so it won't break e.g. forms. The only other change I can think of is to update wording for https://github.com/elixir-ecto/ecto/blob/v2.1.1/lib/ecto/changeset.ex#L199: to be something along the lines of:

```markdown
* `changes`      -  The `changes` from parameters that were approved in casting and passed validations
```

michalmuskala:
I'm :-1: for removing from `changes` on validation failures. This would make it impossible to catch multiple validation errors on a single field - it's not inconceivable that could happen.

`validate_required` is a different case, I think it might make sense to remove `nil` values from changes there, provided the value in `data` is `nil` too.

josevalim:
Good call. Restricting it to validate_required sounds good then.

PR #2129, opened 2017-07-17
Remove the field from changes if it does not pass validate_required

slashmili:
based on the discussion in #1874

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1370,14 +1370,10 @@ defmodule Ecto.Changeset do
           is_nil(errors[field]),
           do: field
 
-    new_errors = fields_with_errors
-                 |> Enum.map(fn field ->
-                   {field, {message, [validation: :required]}}
-                 end)
-
-    case new_errors do
+    case fields_with_errors do
       [] -> %{changeset | required: fields ++ required}
       _  ->
+        new_errors = Enum.map(fields_with_errors, &{&1, {message, [validation: :required]}})
         changes = Map.drop(changes, fields_with_errors)
         %{changeset | changes: changes, required: fields ++ required, errors: new_errors ++ errors, valid?: false}
     end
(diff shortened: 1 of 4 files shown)


lib/ecto/changeset.ex
```diff
@@ -1372,7 +1372,10 @@ defmodule Ecto.Changeset do
 
     case new_errors do
       [] -> %{changeset | required: fields ++ required}
-      _  -> %{changeset | required: fields ++ required, errors: new_errors ++ errors, valid?: false}
+      _  ->
+        new_errors_keys = Keyword.keys(new_errors)
```
josevalim: What if we change the `for` that computes `new_errors` above to simply return `fields_with_errors` and then we build the messages by doing another pass?

Another idea is to use a reduce and build the errors and changes at once.

josevalim:
Thank you! I have added one comment!

slashmili:
@josevalim is it good enough?

Not sure if you are ok with this pipe operator in middle of the code 😬 
```diff
+    new_errors = fields_with_errors
+                 |> Enum.map(&({&1, {message, [validation: :required]}}))
```

lib/ecto/changeset.ex
```diff
@@ -1359,20 +1359,27 @@ defmodule Ecto.Changeset do
...
-    new_errors =
+    fields_with_errors =
       for field <- fields,
           missing?(changeset, field),
           ensure_field_exists!(changeset, field),
           is_nil(errors[field]),
-          do: {field, {message, [validation: :required]}}
+          do: field
+
+    new_errors = fields_with_errors
```
josevalim: You can move this inside the case, since it is needed only when fields_with_errors is not empty. Also, don't pipe only once, you can write this as:

    new_errors = Enum.map(fields_with_error, &{&1, {message, [validation: :required]}})

josevalim:
Beautiful, as soon as the build passes I will merge it. Thank you! 

PR #2131, opened 2017-07-17
cast_assoc for required has_many allows empty list if assoc set to on_replace: :delete

novaugust:
When

1. a struct's has_many assoc is required
2. that assoc is marked `on_replace: :delete`
3. that assoc is cast with an empty list

then: the `required` validation still passes, even though an empty list is invalid in other situations (`def empty?(%{cardinality: :many}, []), do: true`)

This PR adds a failing test for this as well as a fix.

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -29,6 +29,7 @@ defmodule Ecto.Changeset.Relation do
   """
   def empty?(%{cardinality: _}, %NotLoaded{}), do: true
   def empty?(%{cardinality: :many}, []), do: true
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset/relation.ex
```diff
@@ -29,6 +29,7 @@ defmodule Ecto.Changeset.Relation do
   """
   def empty?(%{cardinality: _}, %NotLoaded{}), do: true
   def empty?(%{cardinality: :many}, []), do: true
+  def empty?(%{cardinality: :many}, changes), do: Enum.all?(changes, & &1.action == :replace)
```
josevalim: Should we do: `& &1.action in [:replace, :delete]`? Since `:delete` can also be set manually in the changeset function?

josevalim:

Issue #2128, opened 2017-07-17
Incorrect parameter numbering on self join with fragment


bernardd:
* Elixir version (elixir -v): 1.4.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL
* Ecto version (mix deps): master
* Database adapter and version (mix deps): postgrex 0.13.2

The SQL generated when a self-JOIN uses a `fragment` in the subquery has incorrect numbering on the parameters, which causes Postgrex to throw the following error:

`(Postgrex.Error) ERROR 42P18 (indeterminate_datatype): could not determine data type of parameter $2`

(where $2 is the missing parameter).
The following tests, when placed in `postgres_test.exs` demonstrate the problem:

```elixir
  test "self join on subquery" do
    subquery = Schema |> select([r], %{x: r.x, y: r.y})
    query = subquery |> join(:inner, [c], p in subquery(subquery), true) |> normalize
    assert SQL.all(query) ==
           ~s{SELECT s0."x", s0."y" FROM "schema" AS s0 INNER JOIN } <>
           ~s{(SELECT s0."x" AS "x", s0."y" AS "y" FROM "schema" AS s0) } <>
           ~s{AS s1 ON TRUE}
  end

  test "self join on subquery with fragment" do
    subquery = Schema |> select([r], %{string: fragment("downcase(?)", ^"string")})
    query = subquery |> join(:inner, [c], p in subquery(subquery), true) |> normalize
    assert SQL.all(query) ==
           ~s{SELECT downcase($1) FROM "schema" AS s0 INNER JOIN } <>
           ~s{(SELECT downcase($2) AS "string" FROM "schema" AS s0) } <>
           ~s{AS s1 ON TRUE}
  end
```

The first test passes, however the second test ends up with `downcase($1)` twice and no `downcase($2)`.

bernardd:
Actually, on further inspection, you don't need `fragment` to hit this error - it looks like any expression that pulls in an external value using `^` hits the same problem:
```elixir
    subquery = Schema |> select([r], %{x: ^999})
    query = subquery |> join(:inner, [c], p in subquery(subquery), true) |> normalize
```
gives:
`SELECT $1 FROM "schema" AS s0 INNER JOIN (SELECT $1 AS "x" FROM "schema" AS s0) AS s1 ON TRUE`

josevalim:
Thanks, it seems we are not counting the entries in the select inside the subquery.

[...]

PR #2133, opened 2017-07-20
Partial fix for subquery select parameter count

bernardd:
This PR addresses the issue in #2128, at least as far as the example test cases therein go. It still can't deal with the following case correctly:

```elixir
  test "join on subquery with simple select and multi element outer select" do
    subquery = Schema |> select([r], %{x: ^999})
    query = Schema
            |> select([r], %{y: ^666, w: ^777})
            |> join(:inner, [c], p in subquery(subquery), true)
            |> where([a, b], a.x == ^111)
            |> normalize

    assert SQL.all(query) ==
           ~s{SELECT $1, $2 FROM "schema" AS s0 INNER JOIN } <>
           ~s{(SELECT $3 AS "x" FROM "schema" AS s0) AS s1 ON TRUE } <>
           ~s{WHERE (s0."x" = $4)}
  end
```

As such I'm not actually sure if it's a) the right approach and I just need to figure out how to extend it a bit, or b) entirely the wrong approach. So while it solves my immediate issue, I'm happy to take pointers on how to fix it (more) properly.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -233,10 +233,11 @@ defmodule Ecto.Query.Planner do
               sources: (tail_sources ++ sources) |> Enum.reverse |> List.to_tuple()}
   end
 
-  defp prepare_source(query, %Ecto.SubQuery{query: inner_query} = subquery, adapter) do
+  defp prepare_source(query, subquery, adapter, counter \\ 0)
+  defp prepare_source(query, %Ecto.SubQuery{query: inner_query} = subquery, adapter, counter) do
     try do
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @bernardd. I would need to jump into the code to let you know exactly what is left, I will let you know when I do that. The test cases are also extremely helpful.

josevalim:
Properly fixing this sent me down to a rabbit hole and it turned out rewriting how our planner prepares selects was the best fix. This led to PR #2155 which simplifies adapters, uses less memory and performs faster.

Thank you for the test case @bernardd and for the initial solution!

PR #2141, opened 2017-07-27
Broaden typespecs for query and query!

bernardd:
`Ecto.Adapters.SQL.query/4` and `query!/4` can, at least for some adaptors, return more fields than just `rows` and `num_columns` (the example of `columns` appears in the docs). If you're relying on those fields, you'll get a dialyzer error because the typespec for the returned value doesn't allow for them. This patch fixes that.


josevalim:

PR #2153, opened 2017-08-04
convert subqueries to queries in preloader

zachdaniel:
This (perhaps too naively?) converts a passed in subquery to its query representation when its used in a preload.

### DIFF
--- a/lib/ecto/repo/preloader.ex
+++ b/lib/ecto/repo/preloader.ex
@@ -314,6 +314,7 @@ defmodule Ecto.Repo.Preloader do
 
   defp query!(query) when is_function(query, 1), do: query
   defp query!(%Ecto.Query{} = query), do: query
+  defp query!(%Ecto.SubQuery{} = query), do: Ecto.Queryable.to_query(query)
 
   defp take(take, field) do
     case Access.fetch(take, field) do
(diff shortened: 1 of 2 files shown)


zachdaniel:
Going to add a test for this as well.

michalmuskala:
To be honest, I'm not sure I understand this behaviour... Isn't it weird that I pass a subquery, but then two queries are issued?

zachdaniel:
@michalmuskala What do you mean two queries are issued?

michalmuskala:
Preload does two queries - first loads the main resource, second the subresource.

zachdaniel:
Yeah, okay I thought you meant it would do two queries in addition to the primary query. We just build a whole bunch of queries dynamically all over the place, and we subquery them often to allow for filtering/sorting on calculated fields, but if, for instance, we don't end up calling `from/2` on that subquery with more filters/updating the select statement/sorting, then we can end up passing that query as an include to another query. Its trivial for us to do something about it as well (just make sure its a regular Ecto.Query before putting it into the preload), but I figured others may want that behavior by default.

josevalim:
I see now what you are trying to achieve @zachdaniel. I understood something different (and much more complex to implement)  from the issue report. :)

I don't think we should allow this because in this case you shouldn't be wrapping in a subquery in the first place.

I understand you have reasons for doing this but I think it is best to raise and let the user decide how to proceed.

Thank you!

PR #2155, opened 2017-08-04
Introduce an AST for preprocessing + postprocessing

josevalim:
This reduces the complexity in the adapters, reduces
the caching cost per entry and speeds up deserialization.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -338,13 +338,10 @@ if Code.ensure_loaded?(Mariaex) do
       [name, ?. | quote_name(field)]
     end
 
-    defp expr({:&, _, [idx, fields, _counter]}, sources, query) do
-      {source, name, schema} = elem(sources, idx)
-      if is_nil(schema) and is_nil(fields) do
-        error!(query, "MySQL does not support selecting all fields from #{source} without a schema. " <>
-                      "Please specify a schema or specify exactly which fields you want to select")
-      end
-      intersperse_map(fields, ", ", &[name, ?. | quote_name(&1)])
+    defp expr({:&, _, [idx]}, sources, query) do
+      {source, _name, _schema} = elem(sources, idx)
+      error!(query, "MySQL does not support selecting all fields from #{source} without a schema. " <>
+                    "Please specify a schema or specify exactly which fields you want to select")
     end
 
     defp expr({:in, _, [_left, []]}, _sources, _query) do
@@ -377,8 +374,8 @@ if Code.ensure_loaded?(Mariaex) do
       ["NOT (", expr(expr, sources, query), ?)]
     end
 
-    defp expr(%Ecto.SubQuery{query: query, fields: fields}, _sources, _query) do
-      query.select.fields |> put_in(fields) |> all()
+    defp expr(%Ecto.SubQuery{query: query}, _sources, _query) do
+      all(query)
     end
 
     defp expr({:fragment, _, [kw]}, _sources, query) when is_list(kw) or tuple_size(kw) == 3 do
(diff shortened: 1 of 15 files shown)


josevalim:
Hi @shdblowers, @jbachhardie, @AndrewDryga!

If this pull request is merged, it will require changes in adapters in order to support v2.2. The good news are: those changes can be done in a way to remain compatible with earlier codes and they all simplify the adapter code!

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -338,13 +338,10 @@ if Code.ensure_loaded?(Mariaex) do
...
     end
 
-    defp expr({:&, _, [idx, fields, _counter]}, sources, query) do
-      {source, name, schema} = elem(sources, idx)
-      if is_nil(schema) and is_nil(fields) do
-        error!(query, "MySQL does not support selecting all fields from #{source} without a schema. " <>
-                      "Please specify a schema or specify exactly which fields you want to select")
-      end
-      intersperse_map(fields, ", ", &[name, ?. | quote_name(&1)])
+    defp expr({:&, _, [idx]}, sources, query) do
```
josevalim: Ecto 2.2 will no longer pass those weird and invalid `{:&, _, list_with_three_elements}`. It will pass `{:&, [], [counter]}` which should behave as the old `{:&, _, [counter, nil, nil]}`. If you want to remain backwards compatible, you just need to add this new clause and keep the old one around.

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -377,8 +374,8 @@ if Code.ensure_loaded?(Mariaex) do
       ["NOT (", expr(expr, sources, query), ?)]
     end
 
-    defp expr(%Ecto.SubQuery{query: query, fields: fields}, _sources, _query) do
-      query.select.fields |> put_in(fields) |> all()
+    defp expr(%Ecto.SubQuery{query: query}, _sources, _query) do
```
josevalim: If you want to remain backwards compatible, use this implementation for subqueries:

```
defp expr(%Ecto.SubQuery{query: query} = subquery, _sources, _query) do
  if Map.has_key?(subquery, :fields) do
    query.select.fields |> put_in(subquery.fields) |> all()
  else
    all(query)
  end
end
```

lib/ecto/adapters/sql.ex
```diff
@@ -570,33 +538,6 @@ defmodule Ecto.Adapters.SQL do
     end
   end
 
-  defp process_row(row, process, fields, sources) do
```
josevalim: The difference in this code and the previous code is that adapters are no longer responsible for the whole logic below. To handle this in a backwards compatible fashion, you can change this code in v2.1:

```
mapper = &process_row(&1, process, fields, sources)
```

to something like this:

```
mapper = processor(process, fields, sources)
```

where

```
defp processor(process, fields, source) when is_function(process, 3) do
  &process_row(&1, process, fields, sources)
end
defp processor(process, _fields, _source) when is_function(process, 1) do
  process
end
defp processor(nil, _fields, _source) do
  nil
end
```

josevalim:

josevalim:
Hi @shdblowers, @jbachhardie, @AndrewDryga, @scouten!

Just a heads up that Ecto 2.2 will be released tomorrow (this monday). This issue should cover the new formats added to the query AST.

I have also invited all of you to the @elixir-ecto/adapters team so it is easier to discuss such changes in the future.

AndrewDryga:
Hello, @josevalim. 

I will try to find some time over the next week or two to update Mensia adapter and make it compatible with Ecto 2.2. Thanks for this changes. 

I really feel that there is a huge space for improvements in Adapter behavior to make it easy to build new ones.

PR #2162, opened 2017-08-09
Add validate_unique_tentatively/3

nathanl:
Provide a way to check for uniqueness in the *validation* phase.

This PR comes after much discussion - see https://github.com/elixir-ecto/ecto/pull/2006 - but that doesn't mean that the implementation (or the tests) are perfect.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -31,10 +31,10 @@ defmodule Ecto.Changeset do
   Ecto changesets provide both validations and constraints which
   are ultimately turned into errors in case something goes wrong.
 
-  The difference between them is that validations can be executed
-  without a need to interact with the database and, therefore, are
-  always executed before attempting to insert or update the entry
-  in the database.
+  The difference between them is that validations (with the exception of
+  `validate_unique_tentatively/3`) can be executed without a need to interact
+  with the database and, therefore, are always executed before attempting to
+  insert or update the entry in the database.
 
   However, constraints can only be checked in a safe way when
   performing the operation in the database. As a consequence,
@@ -219,6 +219,7 @@ defmodule Ecto.Changeset do
 
   """
 
+  require Ecto.Query
   alias __MODULE__
   alias Ecto.Changeset.Relation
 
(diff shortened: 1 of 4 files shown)


test/ecto/changeset_test.exs
```diff
@@ -1044,6 +1045,58 @@ defmodule Ecto.ChangesetTest do
     assert changeset.errors == [terms_of_service: {"must be abided", [validation: :acceptance]}]
   end
 
+  test "validate_unique_tentatively/3" do
+    defmodule FakeRepoWithDup do
```
nathanl: @josevalim I couldn't find (or maybe didn't recognize) what you were referring to here:

> One way to solve it is to use the process inbox or the process dictionary to instruct the adapter to return a certain value. We do so currently only on the migration tests.

... so this is my somewhat dumb solution. 😊  I'm happy to do it differently if someone has a better way.

nathanl:
Working on fixing test failures

michalmuskala:
I wonder about the name. Often something that bypasses some rules is called "dirty". Would it make sense to call the function `validate_dirty_uniqueness` or `dirty_validate_uniqueness`?

test/ecto/changeset_test.exs
michalmuskala: I think @josevalim had an idea to modify the test adapter. In particular the function here: https://github.com/elixir-ecto/ecto/blob/master/test/support/test_repo.exs#L43-L45

We could make it use something like:
```
Process.get(:test_repo_all_result) || {1, [[1]]}
```

And then in the test case, we could set the result with `Process.put(:test_repo, all_result, ...)`.

johnnyshields:
@michalmuskala "dirty" implies dirty tracking (i.e. check if an object that was touched)

I prefer the name "validate_unique_optimistically"

nathanl:
@michalmuskala I tried the process dictionary approach - see latest commit.

Personally I like this approach less than the previous one. I had to guess to discover the values `{1, [true]}` and `{0, []}` that the adapter returns, and relying on side effects elsewhere doesn't seem as clean as passing something like `FakeRepoWithDup` to the validation. But I'm happy to use this approach if it's more in line with Ecto test conventions.

Regarding the name: all the existing validation names are of the pattern `validate_{some property of the changeset}`, so I like the idea of having  this one start with `validate_unique`, then adding an adverb. `tentatively` seems to fit well; the definition is "subject to further confirmation", which is exactly what we want to convey: this needs to be followed up with a constraint check.

`optimistically` also seems OK given how the word is used in programming contexts, but to me it's less obvious what it means here; the point is not that we have a sunny outlook on life 😆 , but that we already know we'll need to do some further checking.

nathanl:
I should say - I'm not adamant about the name. Happy to call it whatever y'all think best.

michalmuskala:
My primary concern with "tentatively" is that it's not a very common word and might be confusing for non-native speakers.

chrismcg:
Perhaps `validate_unique_naively` to fit in with `NaiveDateTime`

nathanl:
> Perhaps validate_unique_naively to fit in with NaiveDateTime

That could work. But it might convey "this approach isn't very sophisticated" as opposed to "this is intended to be the first check, to be followed up somehow with a final check." "Tentative" has the built-in idea of "needs confirmation".

> My primary concern with "tentatively" is that it's not a very common word and might be confusing for non-native speakers.

Ah. Maybe so. It's also not super easy to type. 😆

lib/ecto/changeset.ex
```diff
@@ -1392,6 +1393,72 @@ defmodule Ecto.Changeset do
...
+  """
+  def validate_unique_tentatively(changeset, field_names, repo, error_message \\ "has already been taken") do
+    field_names = List.wrap(field_names)
+    where_clause = Enum.map(field_names, fn (field_name) ->
+      {field_name, get_field(changeset, field_name)}
+    end)
+
+    # If we don't have values for all fields, we can't query for uniqueness
+    if Enum.any?(where_clause, fn (tuple) -> is_nil(elem(tuple, 1)) end) do
+      changeset
```
michalmuskala: We should still add something to the `validations` key here

johnnyshields:
I still believe "optimistic" is the correct word in the computer engineering context. https://en.wikipedia.org/wiki/Optimistic_concurrency_control

nathanl:
@johnnyshields I see what you're saying, but I think a unique constraint in the db is the optimistic check here - we don't lock the table before inserting, but if, at the last moment, it turns out that there's a duplicate, we roll back the transaction. The check is optimistic (no up-front locks), but it's reliable.

Validating uniqueness doesn't offer any guarantees, just a better user experience in the 99.99% case where the duplicate already exists at the time of validation.

josevalim:

lib/ecto/changeset.ex
nathanl: @michalmuskala This PR was merged, but I can do another to add to `validations`. I was looking for an example of a validation that explicitly does that, but had to stop working before I found one.

josevalim: No worries, I have added this to master and also renamed it to unsafe_validate_unique. :)

PR #2175, opened 2017-08-22
Parse URL query params

AndrewDryga:
It is common to use params in PostgreSQL connection URIs
(https://www.postgresql.org/docs/9.6/static/libpq-connect.html#AEN45575)
 and we can leverage them to set our shared params.

Change is inspired by our production issue, we had to enable SSL after upgrading database on Heroku without rebuilding the release artifact.

### DIFF
--- a/lib/ecto/repo/supervisor.ex
+++ b/lib/ecto/repo/supervisor.ex
@@ -107,13 +107,15 @@ defmodule Ecto.Repo.Supervisor do
     destructure [username, password], info.userinfo && String.split(info.userinfo, ":")
     "/" <> database = info.path
 
-    opts = [username: username,
-            password: password,
-            database: database,
-            hostname: info.host,
-            port:     info.port] ++ parse_uri_query(info)
+    url_opts = [username: username,
+                password: password,
+                database: database,
+                hostname: info.host,
+                port:     info.port]
 
-    for {k, v} <- opts, v, do: {k, if(is_binary(v), do: URI.decode(v), else: v)}
+    query_opts = parse_uri_query(info)
+
+    for {k, v} <- url_opts ++ query_opts, v, do: {k, if(is_binary(v), do: URI.decode(v), else: v)}
   end
 
   defp parse_uri_query(%URI{query: nil}),
--- a/lib/ecto/repo/supervisor.ex
+++ b/lib/ecto/repo/supervisor.ex
@@ -134,7 +134,7 @@ defmodule Ecto.Repo.Supervisor do
         [{String.to_atom(key), parse_integer!(key, value, url)}] ++ acc
 
       {key, _value}, _acc ->
-        raise Ecto.InvalidURLError, url: url, message: "unsupported query parameter #{key}"
+        raise Ecto.InvalidURLError, url: url, message: "unsupported query parameter `#{key}`"
     end)
   end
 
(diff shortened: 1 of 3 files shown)


lib/ecto/repo/supervisor.ex
```diff
@@ -110,11 +110,33 @@ defmodule Ecto.Repo.Supervisor do
...
     for {k, v} <- opts, v, do: {k, if(is_binary(v), do: URI.decode(v), else: v)}
   end
 
+  defp parse_uri_query(%URI{query: nil}),
+    do: []
+  defp parse_uri_query(%URI{query: query} = url) do
+    query
+    |> URI.query_decoder()
+    |> Enum.reduce([], fn
+      {"ssl", "true"}, acc ->
```
AndrewDryga: This is a subject of discussion, there are many ways to implement it. For example, maybe it's better to move all this stuff to a separate private function.

lib/ecto/repo/supervisor.ex
```diff
@@ -110,11 +110,33 @@ defmodule Ecto.Repo.Supervisor do
             password: password,
             database: database,
             hostname: info.host,
-            port:     info.port]
+            port:     info.port] ++ parse_uri_query(info)
```
AndrewDryga: Optionally, we can get rid of `++` by passing opts as accumulator to the `parse_uri_query/1`.

josevalim:
@michalmuskala thoughts?

AndrewDryga:
Just in case adapter developers want to leverage this feature, we can extend adapter behavior with optional callbacks that resolve query params or return us list of supported options and their types.

michalmuskala:
Many of the postgres libraries support URL params - JDBC seems to have couple pages of them, libpq has a couple. I guess we could do it too. I don't see a reason not to do it.

AndrewDryga:
@josevalim maybe you can guide me to the next steps, do we want to add this to the adapter behaviour.

josevalim:
I would prefer to start small. So this looks good. We just need documentation!

AndrewDryga:
@josevalim I've added docs and made some other sane changes, can you please check it again?

josevalim:

wojtekmach:
I'm working on a sample app that uses postgrex directly and I expected it would allow me to do: `Postgrex.start_link(url: "...")` but it doesn't; would you be interested in a PR that adds it?

AndrewDryga:
@wojtekmach you didn't mention, but I guess you not asking Jose for that 🤣. I'm open to doing that, however, can't tell when because of I'm on a day job :). Maybe it's a good idea to open issue in Postgrex, you can even mention/assign me there and I'll take care of it when it would be possible.

wojtekmach:
oops, it was a question to @josevalim and @michalmuskala whether they'd take a PR, and I wanted to work on it myself!

josevalim:
I am not a big fan of further pushing Ecto concerns into adapters. :S

michalmuskala:
As far as I remember, we had it in the drivers initially. The problem was the implementations weren't consistent and they had to be re-done in every driver/adapter, so we decided to move it to ecto instead.

I'm worried if we just go back to adapters/drivers handling this, we'll end up in the same place. Are there any other options?

wojtekmach:
what if we pretend that Ecto doesn't exist: does parsing URLs makes sense as a feature to Postgrex on it's own? 
One small data point is that, as mentioned at the top of the PR, libpq does support connection strings like: `postgresql://user@localhost?connect_timeout=10`. Even if Postgrex supports such feature it doesn't necessarily mean that Ecto.Adapters.Postgres or other adapters need to use it (but they could).

That being said I totally understand increased maintenance burden of a) allowing more configuration and b) maintaining essentially the same URL parsing code in at least two places.

I only thought about this feature for a sample app that I'm building so it's by no means something I truly need. Hope this wasn't too counterproductive!

PR #2176, opened 2017-08-22
Optimise Changeset.validate_inclusion/4 for MapSet

lpil:
In the even that Changeset.validate_inclusion/4 is given a MapSet of
permitted values we can optimise the lookup to be less than O(n) for
larger collections.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1529,7 +1529,17 @@ defmodule Ecto.Changeset do
 
   """
   @spec validate_inclusion(t, atom, Enum.t, Keyword.t) :: t
-  def validate_inclusion(changeset, field, data, opts \\ []) do
+  def validate_inclusion(changeset, field, data, opts \\ [])
+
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1529,7 +1529,17 @@ defmodule Ecto.Changeset do
...
+    validate_change changeset, field, {:inclusion, set}, fn _, value ->
+      if MapSet.member?(set, value),
+        do: [],
+        else: [{field, {message(opts, "is invalid"), [validation: :inclusion]}}]
+    end
+  end
+
+  def validate_inclusion(changeset, field, data, opts) do
     validate_change changeset, field, {:inclusion, data}, fn _, value ->
       if value in data,
```
josevalim: I believe `value in data` will delegate to `Enum.member?/2`  which will already perform the fast operation for MapSet. :)

lpil: Fantastic, I'd no idea :)

lpil:
Thanks gang :D

PR #2177, opened 2017-08-22
Add key to casting error msg

sztosz:
I've had a problem with identifying  cast which schema field in my changeset was problematic few days ago. This simple change shows problematic field.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -496,7 +496,7 @@ defmodule Ecto.Changeset do
   defp type!(types, key) do
     case Map.fetch(types, key) do
       {:ok, {tag, _}} when tag in @relations ->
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -496,7 +496,7 @@ defmodule Ecto.Changeset do
   defp type!(types, key) do
     case Map.fetch(types, key) do
       {:ok, {tag, _}} when tag in @relations ->
-        raise "casting #{tag}s with cast/3 is not supported, use cast_#{tag}/3 instead"
+        raise "casting #{tag}s with cast/3 for :#{key} is not supported, use cast_#{tag}/3 instead"
```
michalmuskala: let's use `#{inspect key}`

michalmuskala:
Thank you! :heart:

PR #2183, opened 2017-08-25
Use MariaEx typed binary and string if supported

kzemek:
By default, MariaEx doesn't distinguish between raw binaries and text strings. MariaEx `query` features an option `binary_as` that makes the library either treat *all* Elixir strings as raw binaries or as text. This option, however, is not easily accessible from Ecto and is not a solution when a query features both a raw binary and text string.

I've created a pull request https://github.com/xerions/mariaex/pull/198 that allows passing Elixir string parameters wrapped in `%Mariaex.TypedValue` with an explicit `type` field (with a value of either `:binary` or `:string`) to remove ambiguity.

This PR leverages Ecto's knowledge of the model to inform MariaEx of the exact types of Elixir string parameters. The change is made in a backward-compatible way; Travis tests using updated MariaEx can be found [here](https://travis-ci.org/kzemek/ecto/builds/268303373).

### DIFF
--- a/lib/ecto/adapters/mysql.ex
+++ b/lib/ecto/adapters/mysql.ex
@@ -110,6 +110,32 @@ defmodule Ecto.Adapters.MySQL do
   @behaviour Ecto.Adapter.Storage
   @behaviour Ecto.Adapter.Structure
 
+  # Add dumpers functions if tagged values are supported by the driver
+  try do
+    DBConnection.Query.encode(
+      %Mariaex.Query{type: :binary, num_params: 1},
+      [%{__struct__: Mariaex.TypedValue, type: :binary, value: <<3, 2, 1>>}], [])
+
+    @doc false
+    def dumpers(:binary, type),  do: [type, &encode_binary/1]
+    def dumpers(:string, type),  do: [type, &encode_string/1]
+    def dumpers(type_a, type_b), do: super(type_a, type_b)
+
+    defp additional_loaders(:binary, type), do: [&decode_binary/1, type]
+    defp additional_loaders(:string, type), do: [&decode_binary/1, type]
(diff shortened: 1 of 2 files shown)


lib/ecto/adapters/mysql.ex
```diff
@@ -119,7 +145,7 @@ defmodule Ecto.Adapters.MySQL do
   def loaders(:float, type),          do: [&float_decode/1, type]
   def loaders(:binary_id, type),      do: [Ecto.UUID, type]
   def loaders({:embed, _} = type, _), do: [&json_decode/1, &Ecto.Adapters.SQL.load_embed(type, &1)]
-  def loaders(_, type),               do: [type]
+  def loaders(type_a, type_b),        do: additional_loaders(type_a, type_b)
```
michalmuskala: All the loaders are inlined for the reasons of efficiency. This code is called very often, even a function call is a big overhead here - we should avoid it, if we can.

lib/ecto/adapters/mysql.ex
```diff
@@ -110,6 +110,32 @@ defmodule Ecto.Adapters.MySQL do
   @behaviour Ecto.Adapter.Storage
   @behaviour Ecto.Adapter.Structure
 
+  # Add dumpers functions if tagged values are supported by the driver
```
josevalim: We don't need the conditional here, instead we should change "mix.exs" to the minimum Mariaex version that supports this feature.

josevalim:
We have added some comments, thank you!

fishcakez:
Please see my comment in the Mariaex issue, we should be using the type information sent back by the database to encode parameters instead of doing implicit casting. This should mean we can remove the `binary_as` option in Mariaex.

PR #2184, opened 2017-08-25
Allow validation of codepoint count in validate_length

alvinlindstam:
A common use case of validate_length is probably to check
that the given text will fit into a text field with a limited
size.

Most databases ignore graphemes in the size limit for char fields,
and only care about the number of code points.

This commit introduces the `:count` option for `validate_length`,
which will count codepoints instead of graphemes.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1622,7 +1622,7 @@ defmodule Ecto.Changeset do
         count_type = opts[:count] || :graphemes
         {type, length} = case {value, count_type} do
           {value, :codepoints} when is_binary(value) ->
-            {:string, length(String.codepoints(value))}
+            {:string, codepoints_length(value, 0)}
           {value, :graphemes} when is_binary(value) ->
             {:string, String.length(value)}
           {value, _} when is_list(value) ->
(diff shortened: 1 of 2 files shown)


michalmuskala:
If we're checking for "field width" wouldn't bytes be more appropriate then codepoints?

josevalim:
Counting codepoints is ultimately flawed for languages like korean. Can we please try to double check the database behaviour?

alvinlindstam:
Yes, counting unicode codepoints is flawed, but it's what many backend systems do, and if we want to check if a string fits in a CHAR/VARCHAR field it's what you have to do. I might not care about the number of user perceived characters, I just want to make sure that it fits in the column.

When running these queries in a MySQL database with table where name is limited to 50 chars, the second query fails with 
```
UPDATE tag_tag SET name = '12345678901234567890123456789012345678901234567890' WHERE `id` = 1;
UPDATE tag_tag SET name = 'éééééééééééééééééééééééééééééééééééééééééééééééééé' WHERE `id` = 1;
```

The same goes for a test with postgres (test table with name defined as `name character varying(5)`):

```
INSERT INTO test(name) VALUES ('12345');
INSERT INTO test(name) VALUES ('ééééé');
```

alvinlindstam:
@michalmuskala: Yes, bytes could be appropriate as well, if we know of any need for it (a datastore that limits the bytesize). It would also depend on the backend using utf-8. For MySQL and postgres, the limit for (var)char fields is in characters (codepoints), not bytes or graphemes.

josevalim:
@alvinlindstam looks good to me. However, if we are going to count codepoints, we should do it like this:

```elixir
def codepoints_length(<<_::utf8, rest::binary>>, acc), do: codepoints_length(rest, acc + 1)
def codepoints_length(<<_, rest::binary>>, acc), do: codepoints_length(rest, acc + 1)
def codepoints_length(<<>>, acc), do: acc
```

The current implementation is not performant.

josevalim:

PR #2187, opened 2017-08-25
Correct the description of how many_to_many works closes #2186

jmarca:
put_assoc raises a runtime error when changing many to many relationships.  In the example provided in the docs, removing the "clickbait" tag will raise an error.  

To make the removal work as originally presented in the documentation, one must first change the on_replace behavior from its default of :raise to :delete.  

I wrote up some description of this necessary change.  Hope it fits nicely with the existing doc.  It may have typographical errors, as I haven't proofread that well, aside from reading the markdown.

PR #2188, opened 2017-08-25
Correct the description of how many_to_many works

jmarca:
put_assoc raises a runtime error when changing many to many relationships.  In the example provided in the docs, removing the "clickbait" tag will raise an error.

To make the removal work as originally presented in the documentation, one must first change the on_replace behavior from its default of :raise to :delete.

I wrote up some description of this necessary change.  Hope it fits nicely with the existing doc.  It may have typographical errors, as I haven't proofread that well, aside from reading the markdown.

(also, removed stray edit in PR #2187 )


josevalim:

PR #2189, opened 2017-08-27
Implement hint option to join for mysql

matsu911:
Sometimes MySQL needs to specify which index to use.
However, current ecto does not option to specify such hints when we use join.
A user can specify any option with `hint` by hand, so not secure but convenient anyway.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -247,9 +247,9 @@ if Code.ensure_loaded?(Mariaex) do
     defp join(%Query{joins: []}, _sources), do: []
     defp join(%Query{joins: joins} = query, sources) do
       Enum.map(joins, fn
-        %JoinExpr{on: %QueryExpr{expr: expr}, qual: qual, ix: ix, source: source} ->
+        %JoinExpr{on: %QueryExpr{expr: expr}, qual: qual, ix: ix, source: source, hint: hint} ->
           {join, name} = get_source(query, sources, ix, source)
-          [join_qual(qual, query), join, " AS ", name, " ON " | expr(expr, sources, query)]
+          [join_qual(qual, query), join, " AS ", name, hint || "", " ON " | expr(expr, sources, query)]
       end)
     end
 
--- a/lib/ecto/query/builder/join.ex
+++ b/lib/ecto/query/builder/join.ex
@@ -108,9 +108,9 @@ defmodule Ecto.Query.Builder.Join do
   If possible, it does all calculations at compile time to avoid
   runtime work.
   """
-  @spec build(Macro.t, atom, [Macro.t], Macro.t, Macro.t, Macro.t, Macro.Env.t) ::
+  @spec build(Macro.t, atom, [Macro.t], Macro.t, Macro.t, Macro.t, Macro.t, Macro.Env.t) ::
               {Macro.t, Keyword.t, non_neg_integer | nil}
-  def build(query, qual, binding, expr, on, count_bind, env) do
+  def build(query, qual, binding, expr, on, hint, count_bind, env) do
     {query, binding} = Builder.escape_binding(query, binding)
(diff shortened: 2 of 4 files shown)


lib/ecto/query/builder/join.ex
```diff
@@ -187,12 +193,13 @@ defmodule Ecto.Query.Builder.Join do
...
   Called at runtime to build a join.
   """
-  def join!(query, expr, count_bind, join_qual, join_source, join_assoc, join_params, file, line) do
+  def join!(query, expr, count_bind, join_qual, join_source, join_assoc, join_params, hint, file, line) do
     {on_expr, on_params, on_file, on_line} =
       Ecto.Query.Builder.Filter.filter!(:on, query, expr, count_bind, file, line)
 
     join = %JoinExpr{qual: join_qual, source: join_source, assoc: join_assoc,
                      file: file, line: line, params: join_params,
+                     hint: hint,
```
josevalim: Can you please put it in the previous line?

lib/ecto/query/builder/join.ex
```diff
@@ -167,13 +168,18 @@ defmodule Ecto.Query.Builder.Join do
...
                   line: unquote(env.line), params: unquote(join_params),
+                  hint: unquote(hint),
                   on: %QueryExpr{expr: unquote(on_expr), params: unquote(on_params),
                                  line: unquote(env.line), file: unquote(env.file)}}
       end
 
     Builder.apply_query(query, __MODULE__, [join], env)
   end
 
+  defmacro hint(expr) do
```
josevalim: Where is this macro used?

josevalim: To be clear, the formatting of the hint should not be done here in any level but exclusively on the adapter.

test/ecto/query/builder/join_test.exs
```diff
@@ -61,6 +61,12 @@ defmodule Ecto.Query.Builder.JoinTest do
     assert join.on.params == [{true, {1, :public}}]
   end
 
+  test "accepts hint option" do
+    assert %{joins: [join]} =
+      join("posts", :inner, [p], c in "comments", ^[post_id: 1, public: true], hint("force index (id)"))
```
josevalim: We probably don't need the `hint` macro. Just pass a string directly.

test/ecto/query/builder/join_test.exs
```diff
@@ -61,6 +61,12 @@ defmodule Ecto.Query.Builder.JoinTest do
     assert join.on.params == [{true, {1, :public}}]
   end
 
+  test "accepts hint option" do
```
josevalim: We also need tests for the keywords list version.

josevalim: Not sure if it is in this file or in "planner_test.exs".

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -247,9 +247,9 @@ if Code.ensure_loaded?(Mariaex) do
     defp join(%Query{joins: []}, _sources), do: []
     defp join(%Query{joins: joins} = query, sources) do
       Enum.map(joins, fn
-        %JoinExpr{on: %QueryExpr{expr: expr}, qual: qual, ix: ix, source: source} ->
```
josevalim: We need tests for the generated query in "test/ecto/adapters/mysql_test.exs".

josevalim:
Thank you @matsu911, I have added some comments. :heart:

The main concern right now is that this could be a source of SQL injection attacks, since it is not escaped in anyway, therefore we need to guarantee the hint is an index at compile time and add a test for that too.

We also need to study how this works with other adapters before officially merging it.

josevalim:
Ping. :)

josevalim:
I will close this since we didn't have some progress for a while. If there is still an interest on this being merged, please let us know and we will gladly reopen (or resubmit).

PR #2191, opened 2017-08-27
Update documentation for Ecto.Repo.preload/3

LostKobrakai:
Does fix an error in the previous examples and add's descriptive comments about all the different preload definition syntaxes. This should make it easier for people to differentiate the options they have and use the appropriate ones.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -443,12 +443,23 @@ defmodule Ecto.Repo do
 
   ## Examples
 
+      # Use a single atom to preload an association
       posts = Repo.preload posts, :comments
-      posts = Repo.preload posts, comments: :permalinks
-      posts = Repo.preload posts, [:comments, :authors]
-      posts = Repo.preload posts, [comments: [:replies, :likes], authors: []]
-      posts = Repo.preload posts, comments: from(c in Comment, order_by: c.published_at)
 
+      # Use a list of atoms to preload multiple associations
+      posts = Repo.preload posts, [:comments, :authors]
+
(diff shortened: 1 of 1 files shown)


lib/ecto/repo.ex
```diff
@@ -443,12 +443,23 @@ defmodule Ecto.Repo do
...
+      posts = Repo.preload posts, [comments: [:replies, :likes], authors: []]
+
+      # Use a keyword list to customize how associations are queried
+      posts = Repo.preload posts, [comments: from(c in Comment, order_by: c.published_at)]
+
+      # Use a two-element tuple for a custom query and nested association definition
+      query = from c in Comment, order_by: c.published_at
+      posts = Repo.preload posts, [comments: {query, [:replies, :likes]}]
+
+  Note: Preloading nested associations in a custom query definition does work as well.
```
josevalim: What about just saying:

> The query given to preload may also preload its own entries.

LostKobrakai: That's sounds better. Just maybe use `its own associations.`? To me this would be more consistent with the terminology used in the code comments.

josevalim:
Yup, perfect.
-- 


LostKobrakai:
It seems like the changes of this PR never made it into the 2.2.0 branches, so the docs still have incorrect code examples and I'm not sure how useful it'll be in master if you're starting to develop for a v3 of ecto.

josevalim:
@LostKobrakai I will backport to v2.2 but v3.0 is supposed to have very few backwards incompatible changes. The only one that really stands out is the removal of the old datetime types.

Issue #2036, opened 2017-04-20
Prefix option in `Repo.aggregate`?


brandonparsons:
Hi there,

It appears that there is no `prefix` option that can be passed to `Repo.aggregate`. Is this by design?

Is there a way to work around this, if I'm trying to use `Repo.aggregate` to realize the results of a query that requires searching within a given prefix?

**EDIT**

As a follow-up, it also seems I can't add a prefix in `Ecto.Query.subquery`, which is the first thing I tried to do to work around this :)

michalmuskala:
This can definitely be supported. PR will be welcome :)

brandonparsons:
Thanks. I was hoping to make sure I wasn't missing something completely obvious :)

I will have a look at the source to see if it's something I could take a crack at. If it involves actually hitting the SQL queries it will be outside of my ability. If it's munging & passing around options might be something I could do :)

brandonparsons:
Turns out there was a better way to do this (in my case), so I'm going to close. Assuming nobody else is asking for this at the time being.

PR #2200, opened 2017-09-01
Support `prefix` in subqueries.

brandonparsons:
I raised an issue a while back (https://github.com/elixir-ecto/ecto/issues/2036) regarding a lack of a `prefix` option in subqueries. I closed it at the time because I found a workaround, and assumed there was no good reason to include it.

I ran into another case where this was required, so attempted to add the required functionality.

Please review that this makes sense, I'm not very familiar with the rest of the code. All tests do still pass though.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -443,29 +443,17 @@ defmodule Ecto.Query do
       end)
 
   """
-  def subquery(query_source, opts \\ []) do
-    query_source = if Keyword.has_key?(opts, :prefix) do
-      prefix = Keyword.fetch!(opts, :prefix)
-      add_prefix_to(query_source, prefix)
-    else
-      query_source
+  def subquery(query, opts \\ []) do
+    subquery = wrap_in_subquery(query)
+    case Keyword.fetch(opts, :prefix) do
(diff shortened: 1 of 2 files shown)


lib/ecto/query.ex
```diff
@@ -438,12 +443,29 @@ defmodule Ecto.Query do
...
 
   """
-  def subquery(%Ecto.SubQuery{} = subquery),
-    do: subquery
-  def subquery(%Ecto.Query{} = query),
-    do: %Ecto.SubQuery{query: query}
-  def subquery(queryable),
-    do: %Ecto.SubQuery{query: Ecto.Queryable.to_query(queryable)}
+  def subquery(query_source, opts \\ []) do
+    query_source = if Keyword.has_key?(opts, :prefix) do
```
josevalim: Thank you! I think this code could be slightly cleaner though. Something like this:

```elixir
def subquery(query, opts \\ []) do
  subquery = wrap_query(query)
    {:ok, prefix} when is_binary(prefix) -> put_in subquery.query.prefix, prefix
    :error -> subquery
  end
end
```

brandonparsons: @josevalim Updated per your view and pushed. At one point I did originally have the `Keyword.fetch` style, but the `put_in` part was great! Learning for me :)

josevalim:

PR #2201, opened 2017-09-02
Support `prefix` in subqueries (#2200)

brandonparsons:
Just back-porting https://github.com/elixir-ecto/ecto/pull/2200 so that hopefully a `2.2.2` can go out at some point and I can get back onto the hex version :)


josevalim:

PR #2204, opened 2017-09-04
Set foreign key type to same used on primary keys

aeliton:
User defined primary key types set on configuration also defines the
type for foreign keys.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -834,6 +834,8 @@ defmodule Ecto.Migration do
   end
 
   def references(table, opts) when is_binary(table) and is_list(opts) do
+    repo_opts = Keyword.take(Runner.repo_config(:migration_primary_key, []), [:type])
+    opts = Keyword.merge(repo_opts, opts)
     reference = struct(%Reference{table: table}, opts)
 
     unless reference.on_delete in [:nothing, :delete_all, :nilify_all] do
--- a/test/ecto/adapters/mysql_test.exs
+++ b/test/ecto/adapters/mysql_test.exs
@@ -7,6 +7,7 @@ defmodule Ecto.Adapters.MySQLTest do
 
   alias Ecto.Queryable
   alias Ecto.Adapters.MySQL.Connection, as: SQL
+  alias Ecto.Migration.Reference
 
   defmodule Schema do
     use Ecto.Schema
@@ -546,7 +547,7 @@ defmodule Ecto.Adapters.MySQLTest do
   # DDL
 
   import Ecto.Migration, only: [table: 1, table: 2, index: 2, index: 3,
-                                references: 1, references: 2, constraint: 3]
+                                constraint: 3]
 
   test "executing a string during migration" do
     assert execute_ddl("example") == ["example"]
@@ -581,7 +582,7 @@ defmodule Ecto.Adapters.MySQLTest do
 
   test "create table with prefix" do
     create = {:create, table(:posts, prefix: :foo),
-               [{:add, :category_0, references(:categories, prefix: :foo), []}]}
+               [{:add, :category_0, %Reference{table: :categories}, []}]}
 
     assert execute_ddl(create) == ["""
     CREATE TABLE `foo`.`posts` (`category_0` BIGINT UNSIGNED,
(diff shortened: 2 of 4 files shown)


lib/ecto/migration.ex
```diff
@@ -834,6 +834,7 @@ defmodule Ecto.Migration do
   end
 
   def references(table, opts) when is_binary(table) and is_list(opts) do
+    opts = opts ++ Runner.repo_config(:migration_primary_key, [])
```
josevalim: This will silently set the name of all references without name to the name of the primary key. We probably want to do this:

```elixir
repo_opts = Keyword.take(Runner.repo_config(:migration_primary_key, []), [:type])
opts = Keyword.merge(repo_opts, opts)
```

test/ecto/adapters/mysql_test.exs
```diff
@@ -66,6 +68,19 @@ defmodule Ecto.Adapters.MySQLTest do
     IO.iodata_to_binary SQL.delete(prefx, table, filter, returning)
   end
 
+  setup meta do
```
josevalim: We don't need those tests at the adapter level. The unit tests you added in migration_test is enough.

test/ecto/adapters/postgres_test.exs
```diff
@@ -69,6 +71,19 @@ defmodule Ecto.Adapters.PostgresTest do
     IO.iodata_to_binary SQL.delete(prefx, table, filter, returning)
   end
 
+  setup meta do
```
josevalim: We don't need those tests at the adapter level. The unit tests you added in migration_test is enough.

josevalim:
Hi @aeliton! I have added some comments and we are good to go. Thank you!

test/ecto/adapters/mysql_test.exs
aeliton: Without this the adapters tests will be broken. Ex:
`test create table with prefix (Ecto.Adapters.PostgresTest)
     test/ecto/adapters/postgres_test.exs:737
     ** (RuntimeError) could not find migration runner process for #PID<0.572.0>
     code: [{:add, :category_0, references(:categories), []}]}
     stacktrace:
       (ecto) lib/ecto/migration/runner.ex:221: Ecto.Migration.Runner.runner/0
       (ecto) lib/ecto/migration/runner.ex:65: Ecto.Migration.Runner.repo_config/2
       (ecto) lib/ecto/migration.ex:837: Ecto.Migration.references/2
       test/ecto/adapters/postgres_test.exs:739: (test)
`

I have just copied code from the migration test to make them pass, but I have probably added unnecessary code there. How should I proceed?

test/ecto/adapters/postgres_test.exs
aeliton: Please see the comment reply for mysql_test.exs above.

lib/ecto/migration.ex
aeliton: Done.

test/ecto/adapters/mysql_test.exs
josevalim: I see. Maybe we should change those tests to, instead of using `reference`, they build the `%Reference{}` struct directly?

aeliton: Sure, I can try that.

aeliton: Hi @josevalim, I have changed to used %Reference{} and all tests are passing. Can you please check if it's all ok?

test/ecto/adapters/mysql_test.exs
```diff
@@ -599,11 +600,11 @@ defmodule Ecto.Adapters.MySQLTest do
   test "create table with references" do
     create = {:create, table(:posts),
                [{:add, :id, :serial, [primary_key: true]},
-                {:add, :category_0, references(:categories), []},
-                {:add, :category_1, references(:categories, name: :foo_bar), []},
-                {:add, :category_2, references(:categories, on_delete: :nothing), []},
-                {:add, :category_3, references(:categories, on_delete: :delete_all), [null: false]},
-                {:add, :category_4, references(:categories, on_delete: :nilify_all), []}]}
+                {:add, :category_0, struct(%Reference{table: :categories}, []), []},
```
josevalim: You don't need to use `struct(%Reference{table: :categories}, name: :foo_bar)`, instead you can write it as `%Reference{table: :categories, name: :foo_bar}` and it will be a bit more concise. :)

josevalim:
@aeliton that's great! I have added one last comment and we should be good to go!

test/ecto/adapters/postgres_test.exs
```diff
@@ -919,8 +920,8 @@ defmodule Ecto.Adapters.PostgresTest do
 
   test "alter table with prefix" do
     alter = {:alter, table(:posts, prefix: :foo),
-             [{:add, :author_id, references(:author, prefix: :foo), []},
-              {:modify, :permalink_id, references(:permalinks, prefix: :foo), null: false}]}
+             [{:add, :author_id, %Reference{table: :author}, []},
```
aeliton: @josevalim, I have removed the prefix here, since Reference doesn't have it and it wouldn't be set anyway. (same thing for mysql adapter test.

josevalim:

PR #2210, opened 2017-09-07
Additional test tags to support SQLite adapter.

scouten:


scouten:
cc @fishcakez per our discussion earlier today

integration_test/sql/subquery.exs
```diff
@@ -18,6 +18,7 @@ defmodule Ecto.Integration.SubQueryTest do
            TestRepo.all(from p in subquery(query), select: p)
   end
 
+  @tag :map_boolean_in_expression
```
josevalim: We don't need to use a boolean field on those, so maybe it is better to update the tests to use another field and you can get better coverage in sqlite?

integration_test/sql/subquery.exs
```diff
@@ -32,6 +33,7 @@ defmodule Ecto.Integration.SubQueryTest do
            TestRepo.all(from p in subquery(query), select: {p, p.pub})
   end
 
+  @tag :map_boolean_in_expression
```
josevalim: Same here.

integration_test/sql/subquery.exs
scouten: I think the main purpose of this test is to verify handling of expressions in subqueries. I'm having trouble finding a replacement expression to use here without resorting to fragments, which I think would be even more fragile.

Can you suggest an alternative? Otherwise, I think it might be best to leave this as is and accept the loss in coverage in the SQLite case.

josevalim: I see, almost all expressions return booleans, so it is hard to write the test otherwise. Thanks.

josevalim:

Issue #2211, opened 2017-09-07
many to many association with length validation not consistent on create and update.


bartekupartek:
### Environment

* Elixir version (elixir -v): Elixir 1.5
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): 9.6.3
* Ecto version (mix deps): 2.2.1 and current master
* Database adapter and version (mix deps): 0.13.3
* Operating system: macOS Sierra

### Current behavior

Room model with `many_to_many :tags, App.Web.Tag, join_through: "rooms_tags", on_delete: :delete_all, on_replace: :delete` association. On creation `validate_length` works as expected, but on update action when I try to put empty `tag_ids` list `validate_length` ignores that I want to insert empty associations.

    %App.Web.Room{name: "test"} 
    |> Ecto.Changeset.change() 
    |> Ecto.Changeset.put_assoc(:tags, []) 
    |> Ecto.Changeset.validate_length(:tags, min: 1, max: 5) 
    |> App.Repo.insert
    {:error,
     #Ecto.Changeset<action: :insert, changes: %{tags: []},
      errors: [tags: {"should have at least %{count} item(s)",
        [count: 1, validation: :length, min: 1]}], data: #App.Web.Room<>,
      valid?: false>}

    tag = %App.Web.Tag{name: %{"en" => "bike"}} |> App.Repo.insert!
    {:ok, room} = %App.Web.Room{name: "test", tags: [tag]} |> App.Repo.insert
  
    %App.Web.Room{__meta__: #Ecto.Schema.Metadata<:loaded, "rooms">, id: 4, name: "test",
     tags: [%App.Web.Tag{__meta__: #Ecto.Schema.Metadata<:loaded, "tags">,
       id: 31, name: %{"en" => "bike"},
       rooms: #Ecto.Association.NotLoaded<association :rooms is not loaded>}]}

    cs = 
      room 
      |> Ecto.Changeset.change() 
      |> Ecto.Changeset.put_assoc(:tags, []) 
      |> Ecto.Changeset.validate_length(:tags, min: 1, max: 5)
    #Ecto.Changeset<action: nil,
     changes: %{tags: [#Ecto.Changeset<action: :replace, changes: %{}, errors: [],
         data: #App.Web.Tag<>, valid?: true>]}, errors: [],
     data: #App.Web.Room<>, valid?: true>

    cs |> Repo.update
    {:ok,
[...]

PR #2212, opened 2017-09-08
Validate assocs length #2211

bartekupartek:


josevalim:
Thanks @bartekupartek! I wonder if it would be better to set those on the cast_assoc/cast_embed functions instead of adding a new function to the API?

bartekupartek:
I need it in `put_assoc` with `validate_length` cause I have to insert only existing tags, but I will check how it will behave in`cast_assoc`.

josevalim:
> I need it in put_assoc with validate_length cause I have to insert only existing tags, but I will check how it will behave incast_assoc.

Hrm, that's a good point. I think what you have then is the way to go.

lib/ecto/changeset.ex
```diff
@@ -1641,6 +1641,15 @@ defmodule Ecto.Changeset do
   defp codepoints_length(<<_, rest::binary>>, acc), do: codepoints_length(rest, acc + 1)
   defp codepoints_length(<<>>, acc), do: acc
 
+  defp list_length(%{types: types}, field, value) do
+    case Map.fetch(types, field) do
```
josevalim: This is going to break in a couple situations:

1. When the first element of the tuple is `:embed`
2. When the user has a custom type that works as a list

So I think we should pattern match rather on this:

```elixir
{:ok, {tag, relation}} when tag in [:embed, :assoc] ->
  ...
_ ->
  length(value)
```

lib/ecto/changeset/relation.ex
```diff
@@ -39,6 +39,13 @@ defmodule Ecto.Changeset.Relation do
   def empty?(%{}, _), do: false
 
   @doc """
+  Filter empty changes
+  """
+  def filter_empty(%{cardinality: :many}, changes) do
```
josevalim: Can you please make `empty?` use this function for the `cardinality: :many` case and then compare against an empty list?

bartekupartek:
Ok thanks, I didn't considered `:embed` relation, I will try to correct this and `filter_empty`.

josevalim:

PR #2214, opened 2017-09-09
Corrected instruction for supervision tree under Ecto 1.5

earthrid:
“import Supervisor.Spec” is not applicable when using Ecto 1.5.


josevalim:

PR #2215, opened 2017-09-10
Lock migrations to support multiple concurrent migrators

blatyo:
This is still a work in progress. I wanted to put this up for feedback.

* Locks migrations table in a transaction
* Wraps each migration in a Task, which may have its own transaction
* Propogates errors up, while ensure transactions are committed

### DIFF
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -50,4 +50,6 @@ defmodule Ecto.Adapter.Migration  do
     * `:log` - When false, does not log begin/commit/rollback queries
   """
   @callback execute_ddl(repo :: Ecto.Repo.t, command, options :: Keyword.t) :: :ok | no_return
+
+  @callback lock_for_migrations(repo :: Ecto.Repo.t, options :: Keyword.t) :: :ok | no_return
 end
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -51,5 +51,6 @@ defmodule Ecto.Adapter.Migration  do
   """
   @callback execute_ddl(repo :: Ecto.Repo.t, command, options :: Keyword.t) :: :ok | no_return
 
-  @callback lock_for_migrations(repo :: Ecto.Repo.t, options :: Keyword.t, function :: fun) :: {:ok, any} | {:error, any}
+  @callback lock_for_migrations(repo :: Ecto.Repo.t, table :: String.t, options :: Keyword.t, function :: fun) ::
+    {:ok, any} | {:error, any}
 end
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -51,5 +51,5 @@ defmodule Ecto.Adapter.Migration  do
   """
   @callback execute_ddl(repo :: Ecto.Repo.t, command, options :: Keyword.t) :: :ok | no_return
 
-  @callback lock_for_migrations(repo :: Ecto.Repo.t, options :: Keyword.t) :: :ok | no_return
+  @callback lock_for_migrations(repo :: Ecto.Repo.t, options :: Keyword.t, function :: fun) :: {:ok, any} | {:error, any}
 end
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -136,8 +136,14 @@ defmodule Ecto.Adapters.SQL do
         :ok
       end
 
+      @doc false
+      def lock_for_migrations(repo, opts) do
+        Ecto.Adapters.SQL.lock_for_migrations(repo, opts)
+      end
+
       defoverridable [prepare: 2, execute: 6, insert: 6, update: 6, delete: 4, insert_all: 7,
-                      execute_ddl: 3, loaders: 2, dumpers: 2, autogenerate: 1, ensure_all_started: 2]
+                      execute_ddl: 3, loaders: 2, dumpers: 2, autogenerate: 1, ensure_all_started: 2,
+                      lock_for_migrations: 2]
     end
   end
 
@@ -594,6 +600,13 @@ defmodule Ecto.Adapters.SQL do
       nil  -> raise "cannot call rollback outside of transaction"
       conn -> DBConnection.rollback(conn, value)
     end
+  end
+
+  ## Migrations
+
+  @doc false
+  def lock_for_migrations(repo, opts) do
+    # TODO: Figure out what should go here
   end
 
   ## Log
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -608,7 +608,7 @@ defmodule Ecto.Adapters.SQL do
   def lock_for_migrations(repo, table, opts, fun) do
     import Ecto.Query, only: [from: 2]
 
-    transaction(repo, [log: false, timeout: :infinity] ++ [opts], fn ->
+    transaction(repo, opts ++ [log: false, timeout: :infinity], fn ->
       versions =
         from(p in table, select: type(p.version, :integer))
         |> Map.put(:prefix, opts[:prefix])
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -137,13 +137,13 @@ defmodule Ecto.Adapters.SQL do
       end
 
       @doc false
-      def lock_for_migrations(repo, opts) do
-        Ecto.Adapters.SQL.lock_for_migrations(repo, opts)
+      def lock_for_migrations(repo, opts, fun) do
+        Ecto.Adapters.SQL.lock_for_migrations(repo, opts, fun)
       end
 
       defoverridable [prepare: 2, execute: 6, insert: 6, update: 6, delete: 4, insert_all: 7,
                       execute_ddl: 3, loaders: 2, dumpers: 2, autogenerate: 1, ensure_all_started: 2,
-                      lock_for_migrations: 2]
+                      lock_for_migrations: 3]
     end
   end
 
@@ -605,8 +605,18 @@ defmodule Ecto.Adapters.SQL do
   ## Migrations
 
   @doc false
-  def lock_for_migrations(repo, opts) do
-    # TODO: Figure out what should go here
+  def lock_for_migrations(repo, opts, fun) do
+    alias Ecto.Migration.SchemaMigration
+    import Ecto.Query, only: [from: 2]
+
+    transaction(repo, opts, fn ->
+      from(p in {SchemaMigration.get_source(repo), SchemaMigration}, select: p.version)
+      |> Map.put(:prefix, opts[:prefix])
+      |> Map.put(:lock, "FOR UPDATE")
+      |> repo.all(timeout: :infinity, log: false)
+
+      fun.()
+    end)
   end
 
   ## Log
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -608,7 +608,7 @@ defmodule Ecto.Adapters.SQL do
   def lock_for_migrations(repo, table, opts, fun) do
     import Ecto.Query, only: [from: 2]
 
-    transaction(repo, opts, fn ->
+    transaction(repo, [log: false, timeout: :infinity | opts], fn ->
       versions =
         from(p in table, select: type(p.version, :integer))
         |> Map.put(:prefix, opts[:prefix])
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -137,13 +137,13 @@ defmodule Ecto.Adapters.SQL do
       end
 
       @doc false
-      def lock_for_migrations(repo, opts, fun) do
-        Ecto.Adapters.SQL.lock_for_migrations(repo, opts, fun)
+      def lock_for_migrations(repo, table, opts, fun) do
+        Ecto.Adapters.SQL.lock_for_migrations(repo, table, opts, fun)
       end
 
       defoverridable [prepare: 2, execute: 6, insert: 6, update: 6, delete: 4, insert_all: 7,
                       execute_ddl: 3, loaders: 2, dumpers: 2, autogenerate: 1, ensure_all_started: 2,
-                      lock_for_migrations: 3]
+                      lock_for_migrations: 4]
     end
   end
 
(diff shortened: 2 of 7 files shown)


lib/ecto/migrator.ex
```diff
@@ -105,13 +120,27 @@ defmodule Ecto.Migrator do
...
     end
   end
 
   defp run_maybe_in_transaction(repo, module, fun) do
+    Task.async(fn ->
+      do_run_maybe_in_transaction(repo, module, fun)
+    end)
+    |> Task.await(:infinity)
+    |> case do
+      {:error, error, stacktrace} ->
```
josevalim: A task would never return this. Instead it would cause the linked process to fail.

lib/ecto/migrator.ex
```diff
@@ -105,13 +120,27 @@ defmodule Ecto.Migrator do
...
+    verbose_schema_migration repo, "update schema migrations", fn ->
+      SchemaMigration.down(repo, version, opts[:prefix])
     end
   end
 
   defp run_maybe_in_transaction(repo, module, fun) do
+    Task.async(fn ->
+      do_run_maybe_in_transaction(repo, module, fun)
+    end)
+    |> Task.await(:infinity)
```
josevalim: It is better to use `Task.yield` instead of rescuing inside the Task but you would probably need a Task.Supervisor so you can use `async_nolink`.

lib/ecto/migrator.ex
```diff
@@ -177,14 +214,36 @@ defmodule Ecto.Migrator do
...
   end
 
+  defp lock_schema_migrations(repo, opts, fun) do
+    repo.transaction(fn ->
+      # TODO: Move to adapter
+      Ecto.Query.from(p in {SchemaMigration.get_source(repo), SchemaMigration}, select: p.version, lock: "FOR UPDATE")
+      |> Map.put(:prefix, opts[:prefix])
+      |> repo.all(timeout: :infinity, log: false)
+
+      try do
```
josevalim: It seems this try catch is also not necessary.

josevalim:
Thanks @blatyo!

I think we can directly change the migrated_versions function to include the lock:

```elixir
  def migrated_versions(repo, prefix) do
    from(p in {get_source(repo), __MODULE__}, select: p.version)
    |> Map.put(:lock, ^repo.__adapter__.lock_for_migration)
    |> Map.put(:prefix, prefix)
    |> repo.all(@opts)
  end
```

This way all the function the adapter function simply returns the lock string and the `lock_for_migration` function only needs to start the transaction.

lib/ecto/migrator.ex
josevalim: Or alternatively we just let the task crash and bring the link down. The only concern is if the error report will be too verbose or not.

josevalim: Let's go with let it crash first and then we can review the logged output after this is merged. :)

lib/ecto/migrator.ex
blatyo: The reason for this try catch is so that if an exception occurs on the 5th migration, we still commit the schema migrations transaction, which records that we did the first 4.

josevalim:
I see. I am not a fan of the excessive use of rescue and raise. How tricky
would be to change the underlying migration code to return :ok | {:error,
_} some we just reraise at the very end if necessary?
-- 


blatyo:
Is it expected that adapters are suppose to be able to understand the lock piece of the query and turn that into this, for databases like db2, derby, and oracle?

``` sql
LOCK TABLE schema_migrations IN EXCLUSIVE MODE;
SELECT version FROM schema_migrations;
```

I was assuming they weren't, which is why I was doing the lock piece separately. If we're only worried about existing adapters, I believe what you suggest is possible with `lock: "WITH (TABLOCKX)"` for the MSSQL adapter.

It also seems like the mnesia adapter could theoretically support this with [lock/2](http://erlang.org/doc/man/mnesia.html#lock-2).

josevalim:
@blatyo  a good point, so we should probably let the adapter do the whole query, including starting the parent transaction?

blatyo:
@josevalim, I have the same hesitance with `try`/`rescue`. I think switching to `:ok` / `{:error, reason}` style is possible for all the `raise`s that the migrator produces. However, if an exception happens in the user code, we'll still need to rescue that and convert it to an error tuple. Also, since the error is in the user's code, we'll want to preserve the original error and stacktrace.

As long as `raise`s aren't used in the migrator itself, I think it would be fine to do the transaction and lock in the adapter.

josevalim:
Sounds good. Then we should use `catch kind:reason`, store it as `{kind, reason, System.stacktrace}` and then re-raise it with `:erlang.raise/3`.

blatyo:
I implemented the adapter and changed the error propagation. 

Is what I did for the adapter ok? It felt like maybe I was doing something dirty by using something at a higher level abstraction than what I was implementing. Maybe there's a reasonable way to call `execute/6` directly, but I didn't really understand what its arguments were.

For errors that were raised by the migrator, I instead had it return `{:error, exception}` and that will get raised outside the `lock_for_migration`. User error is caught as `{kind, reason, stacktrace}` and re-raised with `:erlang.raise/3`.

lib/ecto/adapters/sql.ex
```diff
@@ -596,6 +602,23 @@ defmodule Ecto.Adapters.SQL do
...
 
+  ## Migrations
+
+  @doc false
+  def lock_for_migrations(repo, opts, fun) do
+    alias Ecto.Migration.SchemaMigration
+    import Ecto.Query, only: [from: 2]
+
+    transaction(repo, opts, fn ->
+      from(p in {SchemaMigration.get_source(repo), SchemaMigration}, select: p.version)
```
josevalim: I don't like this dependency on schema migrations, it feels a bit backwards. Maybe we can pass the initial query as argument to `lock_for_migrations`? Also, I think we don't even need the `SchemaMigration` name, we can probably write it as:

    from(p in SchemaMigration.get_source(repo), select: type(p.version, :integer))

Other than that, the callback looks great! :+1:

lib/ecto/adapters/sql.ex
```diff
@@ -596,6 +602,23 @@ defmodule Ecto.Adapters.SQL do
...
+      |> Map.put(:prefix, opts[:prefix])
+      |> Map.put(:lock, "FOR UPDATE")
+      |> repo.all(timeout: :infinity, log: false)
+
+      fun.()
```
josevalim: I was thinking we could call this `fun.()` with the versions as argument. What do you think?

test/support/test_repo.exs
```diff
@@ -108,6 +108,11 @@ defmodule Ecto.TestAdapter do
 
   ## Migrations
 
+  def lock_for_migrations(repo, opts, fun) do
+    send self(), {:lock_for_migrations, fun}
+    {:ok, fun.()}
```
josevalim: If we call `fun.()` with the migration versions, we can probably get rid of [this hack](https://github.com/blatyo/ecto/blob/4808d1381a61e42c64e4cb5b798fb1f0e271bc16/test/support/test_repo.exs#L38-L41) and move it down here - which is a bit more appropriate.

michalmuskala:
How does it interact with adapters that don't support transactions?

josevalim:
@michalmuskala they will have to implement lock_for_migrations accordingly.

lib/ecto/adapters/sql.ex
blatyo: So, I could pass the query, but not every adapter would use the same query. If I instead passed the table name, is there a form of calling `execute/6`, that would work? The actually query for the adapters is static aside from the table name.

lib/ecto/adapters/sql.ex
blatyo: No problem with that. It'll reduce some of the duplication in the functions passed to this.

blatyo:
@josevalim I added table as an argument to the callback and moved the lookup of migration versions into the adapter. I also completely removed `SchemaMigration.migrated_versions/2`, since that work moved completely into the adapter. After I did that, I was able to remove the hack you mentioned.

Now that the calls to `SchemaMigration` are removed does the callback look good to you?

lib/ecto/adapter/migration.ex
```diff
@@ -50,4 +50,7 @@ defmodule Ecto.Adapter.Migration  do
     * `:log` - When false, does not log begin/commit/rollback queries
   """
   @callback execute_ddl(repo :: Ecto.Repo.t, command, options :: Keyword.t) :: :ok | no_return
+
+  @callback lock_for_migrations(repo :: Ecto.Repo.t, table :: String.t, options :: Keyword.t, function :: fun) ::
+    {:ok, any} | {:error, any}
```
josevalim: Here you say it can return `{:error, _}` but when you call it in migration you don't check for the error tuple. Should we always say it returns `{:ok, any}` then?

lib/ecto/migrator.ex
```diff
@@ -42,8 +42,9 @@ defmodule Ecto.Migrator do
   def migrated_versions(repo, opts \\ []) do
     verbose_schema_migration repo, "retrieve migrated versions", fn ->
       SchemaMigration.ensure_schema_migrations_table!(repo, opts[:prefix])
-      SchemaMigration.migrated_versions(repo, opts[:prefix])
```
josevalim: You have removed the `lock_for_migrations` outside of the `verbose_schema_migration` function but I think it should still be inside. Otherwise we lose the helpful error message that is part of `verbose_schema_migration` - which is meant to add folks with a bad schema_migration setup.

lib/ecto/adapters/sql.ex
```diff
@@ -596,6 +602,23 @@ defmodule Ecto.Adapters.SQL do
...
+  @doc false
+  def lock_for_migrations(repo, table, opts, fun) do
+    import Ecto.Query, only: [from: 2]
+
+    transaction(repo, opts, fn ->
+      versions =
+        from(p in table, select: type(p.version, :integer))
+        |> Map.put(:prefix, opts[:prefix])
+        |> Map.put(:lock, "FOR UPDATE")
+        |> repo.all()
```
josevalim: What happens if `repo.all` fails? How will this error cascade and show up to users? You can for example use "FOR WHATEVER" as the lock and see how it will behave.

josevalim: To clarify, an error can happen if the user already has a schema_migrations table but it doesn't have a version field or something of sorts.

lib/ecto/adapter/migration.ex
josevalim: We could also say it simply returns `any`. i.e. we don't expect the operation to fail. Or it returns the value or it raises.

josevalim:
Thank you @blatyo. I have added two last comments and we should be good to go!

josevalim:
Btw, the build is failing and also please take a look at these warnings: https://travis-ci.org/elixir-ecto/ecto/jobs/274429724#L813-L820

blatyo:
Both points are valid. I'll need to look at both a little longer to decide.

I've been watching the build, but didn't want to go far into changing the tests without knowing where the design would go. I've started to fix them. One thing I've noticed is that the tests assume that the migrations and transactions are running in the same process as the test. That's particularly problematic now that the migrations run within a `Task`. I'm not quite sure what to do for [this test](https://github.com/elixir-ecto/ecto/blob/master/test/ecto/migrator_test.exs#L298-L304).

Finally, in the test output I notice a log of error logs like the following. Would you accept a separate PR to capture logs in the tests so that these don't show up in the test output?

```
** (Mariaex.Error) (1049): Unknown database 'ecto_test'
    (db_connection) lib/db_connection/connection.ex:148: DBConnection.Connection.connect/2
    (connection) lib/connection.ex:622: Connection.enter_connect/5
    (stdlib) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
Last message: nil
```

josevalim:
> Finally, in the test output I notice a log of error logs like the following. Would you accept a separate PR to capture logs in the tests so that these don't show up in the test output?

Yes, definitely.

josevalim:
> I'm not quite sure what to do for this test.

I am not sure either. You can comment the problematic tests and I can take a look at them as well before merging. :D

josevalim:
Thank you @blatyo, I have merged this! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

I have done the following changes before merging:

1. I have removed part of the alter.exs tests the parent migration was causing the whole transaction to lock due to the migration locks

2. I have changed the lock_for_migrations contract to receive a query back again. The rationale is that, even for things such as the ORACLE query you posted, you can still generate the SQL and then use it as two separated commands.

If you have more questions or more improvements, please let us know or send more PRs. Thanks!

blatyo:
Thanks for taking it across the finish line! Also, thanks for all the feedback! 

I suppose for other adapters you can introspect the query passed to get the options to generate the locking query and that should be sufficient.

josevalim:
@blatyo yes, the Ecto.Query construct is part of the adapter contract, so they can do that. Getting the table name is a matter of calling query.from and then that becomes the exact case as you had before. :+1:

scouten:
@josevalim @blatyo I am having trouble updating SQLite adapter to this version of Ecto. SQLite has no locks and no `SELECT FOR UPDATE` syntax, so concurrent schema migration is probably not feasible.

Can you advise on a fallback position? Or construct a fallback approach for those of us maintaining adapters for less fancy database engines? :-)

josevalim:
@scouten the simplest implementation you can have is this:

```elixir
  def lock_for_migrations(repo, query, _opts, fun) do
    query
    |> repo.all()
    |> fun.()
end
```

Although we should wait for @blatyo's input in case the tools he consulted when implementing this do anything special for sqlite.

blatyo:
@scoutin Since SQLite lacks those features, it should define `lock_for_migrations` to fall back back to the previous behavior.

``` elixir
def lock_for_migrations(repo, query, _opts, fun) do
    query
    |> Map.put(:lock, nil)
    |> repo.all()
    |> fun.()
end
```

Basically what @josevalim wrote, but also setting `:lock` to `nil`, since your adapter may use that for something.

scouten:
OK, thank you both. Will try that soon.

scouten:
Yup. Works. Thank you!

fishcakez:
@josevalim the failing alter tests that got removed are very important tests for query type caching we need to bring back equivalent tests.

josevalim:
@fishcakez oh, I forgot about them. I have some WIP in a branch, I will push today.

josevalim:
@fishcakez 1f26fed3a0b4cd4ba2314fed18d38ae62b7a7ce4

fishcakez:
@josevalim 👍

marceldegraaf:
@josevalim it looks like this functionality isn't in the `2.2.6` release; do you know when it will be incorporated in a tagged release?

josevalim:
It will only be in the next major version.
-- 


marceldegraaf:
@josevalim thanks for your reply; you mean it will be available from Ecto `3.x.x` onwards? What would be my best option if I'd like to use this feature already?

josevalim:
You need to depend on master.
-- 


marceldegraaf:
OK, thanks!

PR #2218, opened 2017-09-15
Fix IN query related bug

Bharat311:
The query `params` are assigned wrong counters whenever we have any `WHERE` query followed by an `IN` query. This results in wrong params assigned to the final query.

I believe this effects only those databases which do not support prepared statements, or queries which cannot be cached.

The fixed test case should explain the issue much better.

I was able to replicate the issue using Ecto  and MongoDB. If needed, i can provide a sample from the same.

Thanks


josevalim:

PR #2224, opened 2017-09-18
add a type spec for source_repo_priv

BenMorganIO:


michalmuskala:
Just an FYI - the whole `Mix.Ecto` module is internal and is marked as private. This means there are no guarantees about the API and the module can change (or even go away) at any moment even in a patch version.

josevalim:
In addition to what @michalmuskala said, given that we use typespecs primarily for documentation and this module is private, we can probably skip this one. Thank you!

PR #2226, opened 2017-09-18
Remove hard-coded "Ecto.Association.NotLoaded"

aaronrenner:
I thought it was a little weird seeing the string `"Ecto.Association.NotLoaded"` in that module's Inspect implementation. This change calls `inspect @for` to get the module name dynamically.
It looks like the Inspect implementation for this module already has test coverage in test/ecto/schema_test.exs.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -16,7 +16,7 @@ defmodule Ecto.Association.NotLoaded do
   defimpl Inspect do
     def inspect(not_loaded, _opts) do
       msg = "association #{inspect not_loaded.__field__} is not loaded"
-      ~s(#Ecto.Association.NotLoaded<#{msg}>)
+      ~s(##{inspect @for}<#{msg}>)
     end
   end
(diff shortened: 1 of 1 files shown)


wojtekmach:
> I thought it was a little weird seeing the string "Ecto.Association.NotLoaded" in that module's Inspect implementation.

FWIW it's not that uncommon, here are some examples from stdlib:
- https://github.com/elixir-lang/elixir/blob/v1.5.1/lib/elixir/lib/calendar/date_range.ex#L83
- https://github.com/elixir-lang/elixir/blob/v1.5.1/lib/elixir/lib/version.ex#L666

given the module name is really unlikely to change (and if it does it seems the mistake will be caught quickly) I'd save one protocol dispatch on `inspect @for` :)

aaronrenner:
That makes sense. I didn't realize hard-coding the module name was a pattern in the stdlib. Thanks for the links and I'll close this PR.

PR #2227, opened 2017-09-19
Feature/dump nil

borodark:
Why change ecto this way?

It is important to remember that Database Schema is hard to have done right from the first time. It's even harder to change it especially when application is in production. This solution should facilitate greater adoption of Elixir/Ecto for folks dealing with legacy databases.

I had to work with a legacy database lately and among many idiosyncrasies of database design encountered the one was particularly difficult to dodge. The timestamp column stores time as value of seconds starting from 1970-01-01 00:00:00 in column declared as follows:
`
CREATE TABLE actions ( ... unix_time integer DEFAULT 0 ); ...
`
While `0` represents a valid date it's still the same as being `nil` from legacy applications point of view.

The Ecto does not treat nil differently for any type of its own or custom types. The [dump/3](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L281) always produces `{:ok, nil}` for `nil`.

The change allows custom type implementations to load, dump and cast any value of column into nil value in Ecto structure.

Legacy apps use `date +%s` to write values to this column. The `0` actually represents `1970-01-01 00:00:00` but this is a different issue.
This optional callback will allow custom types to take care of various legacy idiosyncrasies with values like
* "N/A"
* " "
* "NULL"
To be handled by custom types.
See the [sample project type_test](https://github.com/borodark/type_test) 
Not sure where to put [test/ecto/unix_date_time_test.exs](test/ecto/unix_date_time_test.exs) so I have duplicated it in the [sample project type_test](https://github.com/borodark/type_test).


josevalim:
Thanks @borodark although I am not sure we should merge this. This is the first time this use case appear and I am not convinced we should make the type API larger for a single case. I would rather push this concern to schema loading/dumping.

borodark:
Yes - I agree it's invasive. I'll be thinking on how to make this working in schema.

josevalim:
Hi @borodark! Since we won't merge this in this particular shape, I will go ahead and close the PR. Feel free to continue adding comments as we will be glad to keep this conversation going. :heart:

Issue #2229, opened 2017-09-19
Improve the differentiation between cast / dump in Ecto.Type documentation


LostKobrakai:
I've seen quite a few people on slack – as well as myself – initially not really understanding what the difference is for `cast` and `dump` callbacks in ecto types. It would be nice to have an example in the moduledoc, where the following 3 data types would be different to each other.

external data `--(cast)->` internal elixir data `--(dump)->` postgres data

People seem to think cast is for parsing data like this: external data `--(cast)->` postgres data

Sidenote: Having the function`cast/2` and the callback `cast/1` doesn't seem to make the finding the essential information better as the callback docs of `cast/1` do explain the above separation kinda ok. I still think most people will be guided by the big example implementation in the moduledoc.

josevalim:
Can you please send a PR that improves those? One possible example we could use is to have structured data as the app representation but it is converted to a raw map when written to the database.

PR #2230, opened 2017-09-19
Exchange the Ecto.Type example

LostKobrakai:
I'm not fully happy with it by now so I'm totally open on suggestions/improvements.

Regarding #2229

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -23,11 +23,11 @@ defmodule Ecto.Type do
         where: fragment("?->>? ILIKE ?", s.original_url, "port", "443")
 
   So the custom type does need to handle the conversion from 
-  external data to runtime data ([`cast/1`](#c:cast/1)) as well as 
+  external data to runtime data (`c:cast/1`) as well as 
   transforming that runtime data into the `:map` Ecto native type and 
-  back ([`dump/1`](#c:dump/1) and [`load/1`](#c:load/1)).
+  back (`c:dump/1` and `c:load/1`).
 
-      defmodule URI.Type do
+      defmodule EctoURI do
         @behaviour Ecto.Type
         def type, do: :map
 
@@ -68,7 +68,7 @@ defmodule Ecto.Type do
         use Ecto.Schema
 
         schema "posts" do
-          field :original_url, URI.Type
+          field :original_url, EctoURI
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/ecto/type.ex
```diff
@@ -10,55 +10,65 @@ defmodule Ecto.Type do
...
-  If `id` is an integer field, Ecto will fail in the query above
-  because it cannot cast the string to an integer. By using a
-  custom type, we can provide special casting behaviour while
-  still keeping the underlying Ecto type the same:
+      from s in ShortUrl,
+        where: fragment("?->>? ILIKE ?", s.original_url, "port", "443")
 
-      defmodule Permalink do
+  So the custom type does need to handle the conversion from 
+  external data to runtime data ([`cast/1`](#c:cast/1)) as well as
```
josevalim: You can write the callbacks as `c:cast/1` and so on and ExDoc will be able to link to it.

lib/ecto/type.ex
```diff
@@ -10,55 +10,65 @@ defmodule Ecto.Type do
...
+      from s in ShortUrl,
+        where: fragment("?->>? ILIKE ?", s.original_url, "port", "443")
 
-      defmodule Permalink do
+  So the custom type does need to handle the conversion from 
+  external data to runtime data ([`cast/1`](#c:cast/1)) as well as 
+  transforming that runtime data into the `:map` Ecto native type and 
+  back ([`dump/1`](#c:dump/1) and [`load/1`](#c:load/1)).
+
+      defmodule URI.Type do
```
josevalim: Let's call it `EctoURI`. We should avoid polluting namespaces of modules we don't own. :)

josevalim:
I definitely think this is an improvement over the previous example. I have added some comments, thank you!

josevalim:

LostKobrakai:
Before I forget. As my last pr to the docs wasn't backported into the 2.2 docs I though it would be a genius idea to issue this own against the v2.2 branch, but it's not on the master branch now 🤦‍♂️ . Do you guys have any guidelines on handling documentation improvements, so at best they don't sit in the master branch for a few month until the next mayor release? For this time should I add another pr against master?

josevalim:
Oh, I missed that. Please always send PRs against master and we backport when asked or if necessary.

LostKobrakai:
Ok, will do.

josevalim:
For now no worries, I have already ported it to master!

PR #2236, opened 2017-09-22
Add migration_lock repo configuration

### LINKED ISSUE #2233: Multiple concurrent migrators should be optional
With the recent commit from @blatyo to add support for multiple concurrent migrators, one side effect is that migrators are now locking the migration table.

The issue is I am using Ecto with CockroachDB, a Postgresql variant which does not support locking and therefore I can't run my migrations anymore on master.

Could we add the support of multiple concurrent migrators behind a flag so that if it's off, we dont lock the migration table ? Or if we want it to be the default, a way to turn it off ?

Thanks a lot in advance.

tlvenn:
Added migration_lock repo configuration in order to override the lock that is being used by migrators to handle multiple concurrent migrators.

By default "FOR UPDATE" is being used.

Fixes #2233

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -606,10 +606,11 @@ defmodule Ecto.Adapters.SQL do
 
   @doc false
   def lock_for_migrations(repo, query, opts, fun) do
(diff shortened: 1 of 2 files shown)


josevalim:

tlvenn:
Thanks @josevalim !

PR #2238, opened 2017-09-23
Fix code example in Schemaless changeset doc

AntonShevel:
I assume it should be with
`Ecto.Changeset.`
Basically, I had to add it to run the code.


josevalim:

PR #2241, opened 2017-09-26
Prefix support on unsafe_validate_unique

francocatena:
Hi,

I have added just that. Not sure about the correctness of the test though, guidance is greatly appreciated =)

Cheers,

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1416,6 +1416,7 @@ defmodule Ecto.Changeset do
       unsafe_validate_unique(changeset, [:email], repo)
       unsafe_validate_unique(changeset, [:city_name, :state_name], repo)
       unsafe_validate_unique(changeset, [:city_name, :state_name], repo, message: "city must be unique within state")
+      unsafe_validate_unique(changeset, [:city_name, :state_name], repo, prefix: "public")
 
   """
   def unsafe_validate_unique(changeset, fields, repo, opts \\ []) do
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1455,6 +1456,15 @@ defmodule Ecto.Changeset do
           end)
         end
 
+      query =
+        if is_list(opts) && opts[:prefix] do
```
josevalim: I think we can use `if prefix = opts[:prefix] do` and reuse the `prefix` variable later. Thoughts?

francocatena: Love that, thing is, the ```is_list``` check should go somewhere. Since you can use it like this:

```elixir
unsafe_validate_unique(changeset, [:city_name, :state_name], repo, "city must be unique within state")
```
Otherwise this test breaks: https://github.com/francocatena/ecto/blob/9f56ae90a1d19788d901d298c627d8314a04f5d4/test/ecto/changeset_test.exs#L1098

Personally I disagree with that, the message should always be set as an explicit ```message: "xx"``` option.

If we do ```if is_list(opts) && (prefix = opts[:prefix]) do``` (besides the ugliness) we get the warning 'the variable "prefix" is unsafe'.

Can we remove the "implicit" message conversion when a string is passed as an option? Is there another way to check it that I am not seeing?

josevalim: That test is wrong. :) It never triggers the error though because the validation doesn't fail. If the validation failed, it would error because elsewhere we expect opts to be a list. So you can assume that `opts` is a list and you can even add a guard to `unsafe_validate_unique`'s definition that guarantees so.

francocatena: Great, I've added the guard clause and the modified ```if``` statement. Now it looks better, tell me what you think =)

josevalim:

PR #2242, opened 2017-09-26
Get rid of extra 'to' in error message.

ConnorRigby:
(this)[https://github.com/elixir-ecto/ecto/blob/master/lib/mix/ecto.ex#L206] line would cause the message to be printed as:

`Expected SomeAdapter to implement Ecto.Adapter.Storage in order to to drop storage for SomeRepo`

This change fixes the sentence so it reads: 

`Expected SomeAdapter to implement Ecto.Adapter.Storage in order to drop storage for SomeRepo`


josevalim:

PR #2249, opened 2017-09-28
Add unsafe_fragment/1 query API

xadhoom:
This PR adds support for unsafe_fragment/1 query API.

This is meant to be used to insert fragments only from interpolated binaries, to handle cases that fragment/1 is not able to handle.

This API may open the app to injection attacks if not used wisely, but proper warnings are in the doc and errors.

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -272,6 +272,39 @@ defmodule Ecto.Query.API do
   def fragment(fragments), do: doc! [fragments]
 
   @doc """
+  Send an unsafe fragment directly to the database.
+
+  WARNING: sending an unsafe fragment may expose your application to sql
+  injection attacks. Think twice before using it and try to reach the same
+  result using the safe fragment/1 API.
+
+  Sometimes it is not possible to represent all possible databases queries
+  using Ecto's syntax or fragment/1, expecially when dealing with specific
+  database construct and dynamic fields.
+
+  To overcome this, it is possible to build the expression as a binary
+  outside the query expression and send it to the database by interpolating
+  the value in the query.
+
+      def window_sorting(field, dir) do
+        expr = "ROW_NUMBER() OVER(ORDER BY \#{field} \#{dir})"
+        from p in Post,
+          select: %{p | id: unsafe_fragment(^expr)}
+      end
+
+  In the example above, we are using the row_number() sql window function to
+  do a particular type of ordering.
+
+  It is very important to keep in mind that no input sanitization is done
+  by Ecto, so the application may be open to sql injection attacks if the
+  expression is not built properly.
+
+  Only an interpolated string (binary) is supported, any other value will
+  result into a compilation error or ArgumentError during runtime.
+  """
+  def unsafe_fragment(fragment), do: doc! [fragment]
+
+  @doc """
   Allows a field to be dynamically accessed.
 
       def at_least_four(doors_or_tires) do
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -271,7 +271,7 @@ defmodule Ecto.Query.API do
   """
   def fragment(fragments), do: doc! [fragments]
 
-  @doc """
+  @doc ~S"""
   Send an unsafe fragment directly to the database.
 
   WARNING: sending an unsafe fragment may expose your application to sql
@@ -292,11 +292,11 @@ defmodule Ecto.Query.API do
             :asc -> "ASC"
             :desc -> "DESC"
           end
-          "\#{key} \#{dir}"
+          "#{key} #{dir}"
         end
         |> Enum.join(",")
       
-        expr = "ROW_NUMBER() OVER(ORDER BY \#{sort})"
+        expr = "ROW_NUMBER() OVER(ORDER BY #{sort})"
         from p in Post,
           select: %{p | id: unsafe_fragment(^expr)}
       end
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -295,7 +295,7 @@ defmodule Ecto.Query.API do
           "#{key} #{dir}"
         end
         |> Enum.join(",")
-      
+
         expr = "ROW_NUMBER() OVER(ORDER BY #{sort})"
         from p in Post,
           select: %{p | id: unsafe_fragment(^expr)}
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -286,14 +286,24 @@ defmodule Ecto.Query.API do
   outside the query expression and send it to the database by interpolating
   the value in the query.
 
-      def window_sorting(field, dir) do
-        expr = "ROW_NUMBER() OVER(ORDER BY \#{field} \#{dir})"
+      def window_sorting(sorters) do
+        sort = sorters |> Enum.map(fn %{sort: key, direction: dir} ->
+          dir = case dir do
+            :asc -> "ASC"
+            :desc -> "DESC"
+          end
+          "\#{key} \#{dir}"
+        end
+        |> Enum.join(",")
+      
+        expr = "ROW_NUMBER() OVER(ORDER BY \#{sort})"
         from p in Post,
           select: %{p | id: unsafe_fragment(^expr)}
       end
 
   In the example above, we are using the row_number() sql window function to
-  do a particular type of ordering.
+  do a particular type of ordering, using a list of sorters which are not
+  known at compile time.
 
   It is very important to keep in mind that no input sanitization is done
   by Ecto, so the application may be open to sql injection attacks if the
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -553,8 +553,23 @@ defmodule Ecto.Query.Builder do
   """
   def keyword!(kw) do
     unless Keyword.keyword?(kw) do
-      raise ArgumentError, "to prevent sql injection, only a keyword list may be interpolated " <>
-                           "as the first argument to `fragment/1` with the `^` operator, got `#{inspect kw}`"
+      message = """
+      to prevent sql injection, only a keyword list may be interpolated
+      as the first argument to `fragment/1` with the `^` operator, got `#{inspect kw}`
+      """
+      message = if is_binary(kw) do
+        message <> """
+
+        For interpolated strings use `unsafe_fragment/1` to pass the argument
+        as-is to the database, paying attention to possible sql injection attack
+        vectors as no escaping is done by Ecto.
+        
+        Use `unsafe_fragment/1` only as last resort and wisely.
+        """
+      else
+        message
+      end
+      raise ArgumentError, message: message
     end
 
     kw
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -82,6 +82,17 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:fragment, [], merge_fragments(pieces, frags)]}, params_acc}
   end
 
+  def escape({:unsafe_fragment, _, [{:^, _, [var]} = _expr]}, _type, params_acc, _vars, _env) do
+    expr = quote do
+      Ecto.Query.Builder.binary!(unquote(var))
+    end
+    {{:{}, [], [:fragment, [], [{:raw, expr}]]}, params_acc}
+  end
+
+  def escape({:unsafe_fragment, _, _}, _type, _params_acc, _vars, _env) do
+    error! "unsafe_fragment(...) expects only an interpolated string"
+  end
+
   # interval
 
   def escape({:from_now, meta, [count, interval]}, type, params_acc, vars, env) do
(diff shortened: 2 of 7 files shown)


OvermindDL1:
If anything, I'd vote for a warning to be raised on every-single-compile to encourage users to get rid of it and/or get a proper API in.

However, on the other side, maybe put that warning behind an option that defaults to true or so, something like this would be very very useful in using SQL features that Ecto currently fails on, like multi-prefix joins or UNION's or etc...

OvermindDL1:
A few things, you need to add tests first of all, and are you able to, say, put a subquery into the unsafe_fragment for things like UNION's and so forth?

lib/ecto/query/api.ex
```diff
@@ -272,6 +272,39 @@ defmodule Ecto.Query.API do
...
+  Sometimes it is not possible to represent all possible databases queries
+  using Ecto's syntax or fragment/1, expecially when dealing with specific
+  database construct and dynamic fields.
+
+  To overcome this, it is possible to build the expression as a binary
+  outside the query expression and send it to the database by interpolating
+  the value in the query.
+
+      def window_sorting(field, dir) do
+        expr = "ROW_NUMBER() OVER(ORDER BY \#{field} \#{dir})"
```
OvermindDL1: This is not a good example as I already do the same via normal macro's.  Perhaps show how to do a multi-prefix join or a UNION or so?

xadhoom: mmmh I don't think is possible with normal macros when the fields are dynamic, or at least I was not able to do it.

For example, my use case:
```
  defp add_query_ordering_with_cursor(model, sorters) do
    sort = sorters
    |> Enum.map(fn %{sort: key, direction: dir} ->
      key = case key do
        k when is_atom(k) -> Atom.to_string(k)
        k when is_binary(k) -> k
      end

      dir = case dir do
        :asc -> "ASC"
        :desc -> "DESC"
      end

      "#{key} #{dir}"
    end)
    |> Enum.join(",")

    sort = case sort do
      "" -> "id"
      _ -> sort
    end

    fragment = "ROW_NUMBER() over(ORDER BY #{sort})"
    from m in model,
      select_merge: %{m |
        id: unsafe_fragment(^fragment)
      }

  end
```
and after that I add other selects, filters and limits.
Is really possible with standard macros?

xadhoom:
Well,

having a warning *always* can be annoying because is meant to be used when the current API is not sufficient and encourage switching makes no sense is simply because not possible.

I agree that adding a "switchable" flag could restrain newbies to use it while giving others the possibility to shut it off. If so, is worth the effort? What other thinks?

About tests, you mean integration tests?
Query building tests are already present.

xadhoom:
About using unsafe_fragments/1 for UNIONS and so on, I did not tested it and just followed how fragment/1 is tested.

Btw, unsafe_fragment is just the same AST as fragment without Keyword checks at runtime, while forcing only interpolated values at compile time and being a binary at runtime.
So the change is really minimal and where fragment/1 works also unsafe_fragment/1 will.

lib/ecto/query/api.ex
OvermindDL1: > Is really possible with standard macros?

Not macro's, it's a built-in function in Ecto.Query:  https://hexdocs.pm/ecto/Ecto.Query.API.html#field/2  :-)

xadhoom: yes I know it, but if you look how ROW_NUMBER works, it must be into the select clause and it has a special syntax with `OVER (order_by_clause)`  where the order_by_clause is *not* the order_by of the Ecto syntax.

an example query:
`SELECT *, ROW_NUMBER() OVER (ORDER by realm DESC) FROM "domains"`

how is rendered with standard Ecto DSL ? I've bashed by head for a couple of days and from what I've read around is just not possible.

xadhoom:
Another example where this may be useful:

https://elixirforum.com/t/querying-json-fields-with-dynamic-document-schema/7740/2

lib/ecto/query/api.ex
OvermindDL1: Here's an actual example from my source:
```elixir
              select: %{
                _rank: fragment("rank() OVER (PARTITION BY ?, ? ORDER BY ? DESC)",
                  crse.scbcrse_crse_numb, crse.scbcrse_subj_code, field(crse, sorted_field)),
...
```
Is that not what like you are wanting to do (you can of course use `field` on any/all of the fragment inputs too)?

xadhoom: my problem is that the number of fields is not fixed nor are the directions.
My order clause may be from one to N (where N are all schema fields), and each one being sorted in a different direction.

OvermindDL1:
> https://elixirforum.com/t/querying-json-fields-with-dynamic-document-schema/7740/2

In that specific example from that post why not just:
```elixir
  defmacrop ilike_jsonb_fragment(query, field, [_ | _] = keys, value) do
    frag =
      keys
      |> Enum.map(fn _ -> "?" end)
      |> Enum.join("->")
      |> case do f -> "#{f}->>?" end
    keys = Enum.map(keys, &quote(do: ^unquote(&1))
    quote do
      from q in query,
        where: fragment(unquote(frag), field(q, ^String.to_existing_atom(field)), unquote_splicing(keys), ^value)
    end
  end
```
Or something like that?  You could then use it like:
```elixir
query = ilike_jsonb_fragment(query, "some_field", [key1, key2, key3], 42)
```
Which does not seem to require an unsafe_fragment at all.  I'm really not for adding `unsafe_fragment` unless it provides a clear benefit for something that is otherwise impossible in Ecto as it is, and yes it does have some use-cases, such as UNIONs and cross-schema joins and other such things, and tests for those kind of things really should be tested for, not for things that are already possible.

xadhoom:
Well,

in the forum link, the problem was that the depth of the json document queried was not know at compile time.

I'm not a macro expert at all, but your example works if the keys are known at compile time, not runtime.
Or I'm mistaken?

Btw yes, I don't want to contradict on a more proper example you I'm just trying to learn a bit more from your answers :smile: 

Said that, what do you think for a more proper example to put into the docs ?

OvermindDL1:
> I'm not a macro expert at all, but your example works if the keys are known at compile time, not runtime.
Or I'm mistaken?

The 'count' of keys have to be known at compile-time (you can branch on this function easily in that case), but the keys themselves do not have to be known, they can be literals or bindings, whichever.

> Btw yes, I don't want to contradict on a more proper example, I'm just trying to learn a bit more from your answers :smile:

All good.  :-)

> Said that, what do you think for a more proper example to put into the docs ?

If you could get this to work, I'd be so happy:

```elixir
query = from q in unsafe_fragment("public.blah")
query = join(query, :inner, [q], s in unsafe_fragment("someschema.sometable"), s.id == q.id)
```

Or this:

```elixir
query = from q in Blah, group_by: q.name, select: max(q.value)
query = unsafe_query("? UNION SELECT 1", subquery(query))
```

xadhoom:
Well having the "count" known a compile time when you write a sorting function for all your models, even with branching is not really feasible, unless you put a max on how many fields you may sort at a time (which can be reasonable after all).

I'll try to try your examples and report back :)

OvermindDL1:
Also an idea, still disallow string interpolation in unsafe_fragment and go back to `?` inputs, but instead combine them *without* escaping them (which is what `fragment/?` currently does).  The API would be a lot more consistent then and if someone wants the original functionality as it stands then they could just do `unsafe_fragment("?", "blah blah blah")`.  but for those that like to use it more like functions then it would be a lot nicer.

_/me is still unsure if wants it added, need to see how well it handles the truly 'currently impossible' cases..._

xadhoom:
Well, in your example also the plain fragment/1 with interpolation (yes does not makes sense, but just to check) blames about "cannot use outside match clause" and so on.

So mine does, since it is just a fragment/1 underneath.

Before going on a different path, I would like some feedback also from top mantainers in order to have a more shared/general view on where to go.

After all I just did what I understood from @josevalim from a quick email exchange on https://groups.google.com/forum/#!topic/elixir-ecto/8aLKRvp5P3k

lib/ecto/query/builder.ex
```diff
@@ -82,6 +82,17 @@ defmodule Ecto.Query.Builder do
...
   end
 
+  def escape({:unsafe_fragment, _, [{:^, _, [var]} = _expr]}, _type, params_acc, _vars, _env) do
+    expr = quote do
+      Ecto.Query.Builder.binary!(unquote(var))
+    end
+    {{:{}, [], [:fragment, [], [{:raw, expr}]]}, params_acc}
+  end
+
+  def escape({:unsafe_fragment, _, _}, _type, _params_acc, _vars, _env) do
```
josevalim: We need to match on the third element of the tuple being a list with a single argument.

lib/ecto/query/builder.ex
```diff
@@ -82,6 +82,17 @@ defmodule Ecto.Query.Builder do
...
+    error! "unsafe_fragment(...) expects only an interpolated string"
```
josevalim: We can probably do better on the error message. In particular, interpolated string is amibiguous. Something like:

error! """
unsafe_fragment(...) expects a single argument to be interpolated as ^argument

The argument is a string that will be sent to the database as is. For this reason,
use this feature with extreme care and instead prefer to use fragment as much 
as possible.
"""
```

Can you also make sure `fragment/1` raises a nice error message if someone is trying to interpolate something into it?

File: lib/ecto/query/builder.ex [hunk 4]
```diff
@@ -544,6 +555,21 @@ defmodule Ecto.Query.Builder do
   end
 
   @doc """
+  Called by escaper at runtime to verify binaries.
+  """
+  def binary!(v) do
+    unless is_binary(v) do
+      raise ArgumentError, "`unsafe_fragment/1` expects only an interpolated string. " <>
```
josevalim: This error message is quote good but I think we can probably just say `unsafe_fragment/1 expects a string at compile time` and move the bulk of it to the error message I mentioned above!

lib/ecto/query/builder.ex
```diff
@@ -544,6 +555,21 @@ defmodule Ecto.Query.Builder do
   end
 
   @doc """
+  Called by escaper at runtime to verify binaries.
+  """
+  def binary!(v) do
```
josevalim: We should probably call it "unsafe_fragment!" since we are mentioning the fragment in the error message.

lib/ecto/query/builder.ex
```diff
@@ -544,6 +555,21 @@ defmodule Ecto.Query.Builder do
...
+    unless is_binary(v) do
+      raise ArgumentError, "`unsafe_fragment/1` expects only an interpolated string. " <>
+                           "You should build your binary and pass as an interpolated value " <>
+                           "with the `^` operator. " <>
+                           "WARNING: interpolating the fragment may open you to sql injection attacks! " <>
```
josevalim: This error message is quite good. We can probably use heredocs here too.

josevalim:
@xadhoom please check builder/join and builder/select, they explicitly handle fragments in some places, so we will need to do that there too and add some tests. Good news is that the implementation should be equally straight-forward.

lib/ecto/query/builder.ex
xadhoom: Ok for the error message.

About fragment/1 : there's already a check in place which makes sure that the interpolated is only a keyword list (via keyword!/1) . Do you mean to rephrase the message? Or something else?

josevalim: > About fragment/1 : there's already a check in place which makes sure that the interpolated is only a keyword list (via keyword!/1) . Do you mean to rephrase the message? Or something else?

Good point. I think we can improve that check so that, if a binary is given, we mention unsafe fragment and the same kind of pitfalls and so on.

lib/ecto/query/api.ex
```diff
@@ -272,6 +272,49 @@ defmodule Ecto.Query.API do
   def fragment(fragments), do: doc! [fragments]
 
   @doc """
```
josevalim: Please use `~S"""` this way you don't need to escape the interpolations down below. :)

xadhoom:
@josevalim PR updated to handle all your (great :) ) review cases

I'm learning a lot :)

lib/ecto/query/builder.ex
```diff
@@ -536,14 +553,47 @@ defmodule Ecto.Query.Builder do
   """
   def keyword!(kw) do
     unless Keyword.keyword?(kw) do
```
josevalim: This can be written with a cond:

```elixir
cond do
  Keyword.keyword?(kw) ->
    kw
  is_binary(kw) ->
    raise ArgumentError, "...binary error message..."
  true ->
    raise ArgumentError, "...general error message..."
end
```

josevalim: Actually, I ignore this, I don't think we need two different error messages. Let me write a single message that handles both cases, one sec.

lib/ecto/query/builder.ex
```diff
@@ -536,14 +553,47 @@ defmodule Ecto.Query.Builder do
...
-      raise ArgumentError, "to prevent sql injection, only a keyword list may be interpolated " <>
-                           "as the first argument to `fragment/1` with the `^` operator, got `#{inspect kw}`"
+      message = """
+      to prevent sql injection, only a keyword list may be interpolated
+      as the first argument to `fragment/1` with the `^` operator, got `#{inspect kw}`
+      """
+      message = if is_binary(kw) do
+        message <> """
+
+        For interpolated strings use `unsafe_fragment/1` to pass the argument
```
josevalim: I think "interpolated strings" is ambiguous because of string interpolation. I think we can have a single message like this:

```
"""
to prevent SQL injection attacks, fragment allows only keyword lists to be
interpolated as the first argument via the `^` operator.

If you are in a rare situation where the string that must be sent to the database
is defined dynamically, you can use `unsafe_fragment/1` to bypass Ecto checks
and send the string as is. Use `unsafe_fragment/1` only as last resort and wisely.

Got: `#{inspect kw}`
"""
```

josevalim:
Hi @xadhoom! I have some comments left that are exclusive to the error messages, so I will merge it and change those locally. The rest is great, thanks! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:

xadhoom:
thanks! I'm very happy for my first contribution to the Elixir ecosystem!

PR #2254, opened 2017-10-01
Add missing parameter names in Repo.load/2 typespec

jayjun:
Documentation refers to the second argument as `data`, but function signature shows `load(arg0, arg1)`.


josevalim:

PR #2262, opened 2017-10-04
Docs: What is Ecto

superherointj:
What is Ecto?


superherointj:
What is Ecto?
* Ecto is a domain specific language for writing queries and interacting with databases in Elixir.

PR #2270, opened 2017-10-12
Allow default to be a map when type is {:map, type} in migrations

evuez:
Allows to do this:

```elixir
add :field, {:map, :string}, default: %{}
```

Not sure if that makes sense since the `type` isn't actually used by any adapter (I think?) but since it's allowed in the migration it's nice to be able to add a default on these too.


josevalim:
Thanks but I would prefer if we don't do that for now exactly because the type is not enforced by the database in any way.

evuez:
Sure that makes sense.

PR #2271, opened 2017-10-12
Document using a fragment with order_by/3

nathanl:
The documentation didn't show that a fragment could be used with `order_by/3`, so I added my own use case.

Is it too weird? The idea is to have an order that's shuffled but stable for pagination (in the case where I'm using it, inserting and deleting records is fairly rare; such changes would change the ordering some because a new record might appear in the middle of the result set.)

The point is not to show this particular use case, it's just the example I have for why fragments might be needed.


josevalim:

PR #2272, opened 2017-10-12
Pass the executing repo module as the first argument to Multi.run

stavro:
Originating from forum post: https://groups.google.com/forum/#!topic/elixir-ecto/uqUkVCKhd7c

Passes the repo module through during execution to all run functions as the first argument.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -27,10 +27,11 @@ defmodule Ecto.Multi do
   ## Run
 
   Multi allows you to run arbitrary functions as part of your transaction
-  via `run/3` and `run/5`. This is very useful when an operation depends
-  on the value of a previous operation. For this reason, the function given
-  as callback to `run/3` and `run/5` will receive all changes performed by
-  the multi so far as a map in the first argument.
+  via `run/3` and `run/5`. This is especially useful when an operation
+  depends on the value of a previous operation. For this reason, the
+  function given as a callback to `run/3` and `run/5` will receive the repo
+  as the first argument, and all changes performed by the multi so far as a
+  map for the second argument.
 
   The function given to `run` must return `{:ok, value}` or `{:error, value}`
   as its result. Returning an error will abort any further operations
@@ -132,8 +133,8 @@ defmodule Ecto.Multi do
 
   ## Example
 
-      iex> lhs = Ecto.Multi.new |> Ecto.Multi.run(:left, &{:ok, &1})
-      iex> rhs = Ecto.Multi.new |> Ecto.Multi.run(:right, &{:error, &1})
+      iex> lhs = Ecto.Multi.new |> Ecto.Multi.run(:left, fn _, _ -> {:ok, :ok} end)
+      iex> rhs = Ecto.Multi.new |> Ecto.Multi.run(:right, fn _, _ -> {:error, :error} end)
       iex> Ecto.Multi.append(lhs, rhs) |> Ecto.Multi.to_list |> Keyword.keys
       [:left, :right]
 
@@ -150,8 +151,8 @@ defmodule Ecto.Multi do
 
   ## Example
 
-      iex> lhs = Ecto.Multi.new |> Ecto.Multi.run(:left, &{:ok, &1})
-      iex> rhs = Ecto.Multi.new |> Ecto.Multi.run(:right, &{:error, &1})
+      iex> lhs = Ecto.Multi.new |> Ecto.Multi.run(:left, fn _, _ -> {:ok, :ok} end)
+      iex> rhs = Ecto.Multi.new |> Ecto.Multi.run(:right, fn _, _ -> {:error, :error} end)
       iex> Ecto.Multi.prepend(lhs, rhs) |> Ecto.Multi.to_list |> Keyword.keys
       [:right, :left]
 
@@ -239,7 +240,7 @@ defmodule Ecto.Multi do
   def update(multi, name, %Changeset{} = changeset, opts \\ []) do
     add_changeset(multi, :update, name, changeset, opts)
   end
-  
+
   @doc """
   Inserts or updates a changeset depending on whether the changeset was persisted or not.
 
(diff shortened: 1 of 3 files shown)


stavro:
@josevalim Should we also pass `repo` to functions provided to `Multi.merge/2` and `Multi.merge/4`?

Any run blocks in merged multis will receive the repo just fine, but I'm not familiar with the intended use case of the function for `merge`.

Thanks!

lib/ecto/multi.ex
```diff
@@ -444,10 +446,13 @@ defmodule Ecto.Multi do
...
   defp apply_operation({:merge, merge}, acc, {wrap, return}, repo),
-    do: {:merge, __apply__(apply_fun(merge, acc), repo, wrap, return)}
+    do: {:merge, __apply__(apply_merge_fun(merge, acc), repo, wrap, return)}
+
+  defp apply_merge_fun({mod, fun, args}, acc), do: apply(mod, fun, [acc | args])
+  defp apply_merge_fun(fun, acc),              do: apply(fun, [acc])
 
-  defp apply_fun({mod, fun, args}, acc), do: apply(mod, fun, [acc | args])
-  defp apply_fun(fun, acc),              do: apply(fun, [acc])
+  defp apply_run_fun({mod, fun, args}, repo, acc), do: apply(mod, fun, [repo | [acc | args]])
```
stavro: Is there a cleaner way to prepend multiple variables to a list than `[repo | [acc | args]]`?

lib/ecto/multi.ex
```diff
@@ -132,8 +133,8 @@ defmodule Ecto.Multi do
 
   ## Example
 
-      iex> lhs = Ecto.Multi.new |> Ecto.Multi.run(:left, &{:ok, &1})
-      iex> rhs = Ecto.Multi.new |> Ecto.Multi.run(:right, &{:error, &1})
+      iex> lhs = Ecto.Multi.new |> Ecto.Multi.run(:left, fn _, _ -> {:ok, :ok} end)
```
stavro: Separate from this PR, I've had a few instances where using the capture operator would be useful, but the restriction that it _must_ name all functions (can't use `&2` if `&1` isn't defined) is limiting.

Is there a restriction I'm unaware of as to why we don't allow just `&2` to be used, and silently discard `&1`?

lib/ecto/multi.ex
```diff
@@ -303,10 +304,11 @@ defmodule Ecto.Multi do
   Adds a function to run as part of the multi.
 
   The function should return either `{:ok, value}` or `{:error, value}`,
-  and receives changes so far as an argument.
+  and receives the repo as the first argument, and the changes so far
+  as the second argument.
   """
   @spec run(t, name, (t -> {:ok | :error, any})) :: t
-  def run(multi, name, run) when is_function(run, 1) do
+  def run(multi, name, run) when is_function(run, 2) do
```
OvermindDL1: Can't this be backwards compatible with 2.x by accepting either a 1-arity or 2-arity function?

stavro: Anonymous functions and MFA functions should have the same signature.

While anonymous functions are _guaranteed_ to not have additional arguments, and thus we can test for the arity, doing the same for MFA can be ambiguous.

Example:

```
defmodule SomeModule do
  def some_function(changes, arg1 \\ nil) do
    {:ok, true}
  end 
end

multi1 = Multi.new |> Multi.run(:fun, SomeModule, :some_function, [arg1])
multi2 = Multi.new |> Multi.run(:fun, SomeModule, :some_function, [])
```

In this contrived example, SomeModule has a function, `some_function`, which has both arity 1 and 2.  Do we:

  * Check for arity 2 first and pass changes, repo? In this case we'd be wrong for multi2.
  * Check for arity 1 first and pass just changes? Can't do that.
  * Check for arity 2 + args length first? This is getting complicated.

While we could probably come up with something clever, I'm not sure if the cleverness would cause more confusing problems for some people than just documenting the upgrade path from Ecto 2.x to 3.x.

More than happy to hear any suggestions.

OvermindDL1: With MFA sure, but the feature of passing in `repo` now could be done via anonymous functions.  The MFA form can be added/changed in 3.0.  It is not much of a cost to wrap an MFA in an anonymous function anyway and it adds the feature in a backwards compatible way too.

OvermindDL1: Plus you could even use `:erlang.function_exported`(sp?) to test the mfa and args to see if it is exported (and if both are then throw an error saying to disambiguate via an anonymous function or so).

Or just add a new :run method with a new argument or name or so...

stavro: I'll wait for more direction from a maintainer as to how/if they'd like to proceed and how they want to handle backward compatibility.

Valid argument & suggestions though. Thanks!

lib/ecto/multi.ex
josevalim: This is not the place for this discussion. :) You can start something in the mailing list but short answer is yes.

lib/ecto/multi.ex
josevalim: `[repo, acc | args]`

lib/ecto/multi.ex
josevalim: We can emit  a warning for anonymous functions like this:

```elixir
def run(multi, name, run) when is_function(run, 1) do
  IO.warn "Giving an anonymous function with 1 argument to Ecto.Multi.run/3 is deprecated. " <>
            "Please pass an anonymous function that receives the repository and the changes so far"
  run(multi, name, fn _repo, changes -> run.(changes) end)
end
```

Given it is only 5 LOC and it will lead to a better upgrading experience, we should do it, yeah.

stavro:
This has been updated with the backwards-compatible anonymous function.

I also added a separate commit to fix the typespecs, which I believe were wrong before this PR (the anonymous run function should not receive the Multi as an argument).

Outstanding question: Should we also pass repo to functions provided to Multi.merge/2 and Multi.merge/4? Any run blocks in merged multis will receive the repo just fine, but I'm not familiar with the intended use case of the function for merge.

josevalim:

PR #2273, opened 2017-10-12
[docs] Document that Changeset.validate_required drops fields

novaugust:
This functionality is mentioned in the documentation for `update_change/3`, but since there's nothing about it in `validate_required` I wasn't sure what to believe =)  

>Note that the value of the change  can still be `nil` (unless the field was marked as required on `validate_required/3`).

Dug into the code and saw that it was the case. Thought this was an important functionality and that it should be documented.


josevalim:

PR #2276, opened 2017-10-13
Update Ecto.Changeset.optimistic_lock/3 typespec

mememori:
As of right now, the function typespec defines that the incrementor must receive an integer and return an integer but that's not true; particularly, on the documentation of the function we have an example that receives a binary and returns a binary:

```elixir
iex> Ecto.Changeset.optimistic_lock(post, :lock_uuid, fn _ -> Ecto.UUID.generate end)
```

and another possible example would be
```elixir
iex> Ecto.Changeset.optimistic_lock(user, :last_update, fn _ -> NaiveDateTime.utc_now() end)
```

So, this PR corrects the typespec to consider correct any 1-arity function


josevalim:

PR #2283, opened 2017-10-21
fix wrong arity in Ecto.Adapter.Transaction docs

n0isiv:


lib/ecto/adapter/transaction.ex
```diff
@@ -10,7 +10,7 @@ defmodule Ecto.Adapter.Transaction  do
   is the value return by the function or `{:error, value}` if the transaction
   was rolled back where `value` is the value given to `rollback/1`.
 
-  See `Ecto.Repo.transaction/1`.
+  See `Ecto.Repo.transaction/2`.
```
wojtekmach: while at it, since it's a callback it would be great if you could change to `c:Ecto.Repo.transaction/2` so that ex_doc autolinks it correctly. Same goes for `c:Ecto.Repo.rollback/1` elsewhere in this file

josevalim:
Fixed locally, thank you.

PR #2286, opened 2017-10-26
allow more options for migration_primary_key config

dallincrane:
Right now configuration for `:migration_primary_key` doesn't allow passing any additional options to the `add` function. This PR allows other options to be passed. Of the available options the only one that makes sense right now would be `default` but this change will keep things flexible for different options that might be rolled out in the future.

An example of how this might be utilized:
On a product my team is working on, sometimes it's convenient for our client-side developers to add a record directly to the dev database. We like to add a default to the `id` fields to make that manual entry easier.

`config :app, App.Repo, migration_primary_key: [id: :uuid, type: :binary_id, default: {:fragment, "gen_random_uuid()"}]`

used the `timestamps` function as a style reference

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -308,7 +308,12 @@ defmodule Ecto.Migration do
 
       if table.primary_key do
         opts = Runner.repo_config(:migration_primary_key, [])
-        add(opts[:name] || :id, opts[:type] || :bigserial, primary_key: true)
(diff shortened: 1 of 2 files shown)


josevalim:
Can you please add a test showing that custom options also work? Look at the existing tests for this feature for inspiration! Thanks!

josevalim:

PR #2291, opened 2017-11-02
add support for usec types

dallincrane:


lib/ecto/adapters/postgres/connection.ex
```diff
@@ -925,18 +940,21 @@ if Code.ensure_loaded?(Postgrex) do
...
+    defp ecto_to_db(:binary_id),           do: "uuid"
+    defp ecto_to_db(:string),              do: "varchar"
+    defp ecto_to_db(:binary),              do: "bytea"
+    defp ecto_to_db(:map),                 do: Application.fetch_env!(:ecto, :postgres_map_type)
+    defp ecto_to_db({:map, _}),            do: Application.fetch_env!(:ecto, :postgres_map_type)
+    defp ecto_to_db(:time_usec),           do: "time"
+    defp ecto_to_db(:utc_datetime),        do: "timestamp"
+    defp ecto_to_db(:utc_datetime_usec),   do: "timestamp"
+    defp ecto_to_db(:naive_datetime),      do: "timestamp"
+    defp ecto_to_db(:naive_datetime_usec), do: "timestamp"
```
josevalim: If we return `timestamp(6)` here, do we run into issues?

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -769,20 +779,23 @@ if Code.ensure_loaded?(Mariaex) do
...
+    defp ecto_to_db(:float, _query),               do: "double"
+    defp ecto_to_db(:binary, _query),              do: "blob"
+    defp ecto_to_db(:uuid, _query),                do: "binary(16)" # MySQL does not support uuid
+    defp ecto_to_db(:map, _query),                 do: "text"
+    defp ecto_to_db({:map, _}, _query),            do: "text"
+    defp ecto_to_db(:time_usec, _query),           do: "time"
+    defp ecto_to_db(:utc_datetime, _query),        do: "datetime"
+    defp ecto_to_db(:utc_datetime_usec, _query),   do: "datetime"
+    defp ecto_to_db(:naive_datetime, _query),      do: "datetime"
+    defp ecto_to_db(:naive_datetime_usec, _query), do: "datetime"
```
josevalim: If we return `datetime(6)` here, do we run into issues?

lib/ecto/type.ex
```diff
@@ -76,6 +76,14 @@ defmodule Ecto.Type do
 
   import Kernel, except: [match?: 2]
 
+  alias Ecto.Types.DateType
```
josevalim: Let's please keep the logic for this inlined in this single module instead of introducing a bunch of new modules.

josevalim:
Thanks @dallincrane, this looks great. I would just like to ask you to revert the splitting you did on Ecto.Type and keep a single module. They share a lot of common logic and it would make the PR easier to review too. We can always move it apart later.

lib/ecto/adapters/postgres/connection.ex
dallincrane: Having this just be `"timestamp"` here allows the precision option to be set. By default, Postgres has no explicit bound on precision. Because of this it will work for a `_usec` type without having to set the precision to `6` like we do for mysql.

lib/ecto/adapters/mysql/connection.ex
dallincrane: Having this just be `"timestamp"` here allows the precision option to be set. For mysql, a default precision of `6` is set for `_usec` types in `column_type/2`.

lib/ecto/adapters/postgres/connection.ex
josevalim: The "issue" is that we can also do something like `from "posts", select: type(p.published_at, :naive_datetime_utc)` and we want that to have precision of 6 and we want the `:naive_datetime` to have precision of 0. So maybe ecto_to_db should always return the precision and column_type should handle when the precision is duplicated, something like:

```elixir
defp column_type(type, opts) when type in ~w(time time_usec)a do
  if precision = Keyword.get(opts, :precision) do
    ["time", ?(, to_string(precision), ?)]
  else
    ecto_to_db(type)
  end
end

defp column_type(type, opts) when type in ~w(naive_datetime utc_datetime utc_datetime_usec naive_datetime_usec)a do
  if precision = Keyword.get(opts, :precision) do
    ["timestamp", ?(, to_string(precision), ?)]
  else
    ecto_to_db(type)
  end
end

defp ecto_to_db(:time), do: "time(0)"
defp ecto_to_db(:time_usec), do: "time(6)"
# ...
```

josevalim: Actually, after further thoughts, I believe your code is right since we remove the precision on the Elixir side.

josevalim:
This looks perfect now @dallincrane! :heart: There is still more work to do in other areas, would you like to do it or would you prefer if i merged it and finished it?

Here is what comes to mind:

  * We need docs in Ecto.Schema
  * Change test/ecto/repo/autogenerate_test.exs and test that generation works with both naive_datetime_usec and utc_datetime_usec (I would say they don't right now)
  * We need to add some integration tests, probably in integration_test/cases/interval.exs that uses usec types and make sure that everything works. See `test "datetime_add with utc_datetime"` at the end of the file for an example. Probably something exactly like that, except it uses the other types.

lib/ecto/adapters/mysql/connection.ex
josevalim: You are correct, thank you!

lib/ecto/adapters/postgres/connection.ex
josevalim: Actually, your current code is correct, apologies. :)

dallincrane: Hmm, I've been looking into your first response a bit deeper.

First of all, thanks for the explanation of the issue! That helped me look into something I missed — that `tagged_to_db/1` calls `ecto_to_db/1`. I didn't see that the first time through for some reason.

Wondering if you could help me understand the example you gave with this because I was actually persuaded by your first comment that this does need to be different. haha

I'm reading the following function in `Postgres.Connection` as building part of query that is actually sent to the DB. 
```
    defp expr(%Ecto.Query.Tagged{value: other, type: type}, sources, query) do
      [expr(other, sources, query), ?:, ?: | tagged_to_db(type)]
    end
```

So if we have a query  
`from "posts", select: type(p.published_at, :naive_datetime)`
with the changes in this PR it would build
`SELECT p0."published_at"::timestamp FROM "posts" AS p0`
but with a when using time, naive_datetime, or utc_datetime wouldn't the user be wanting the microseconds dropped if they exist?
`SELECT p0."published_at"::timestamp(0) FROM "posts" AS p0`

That's what I thought you were referring to anyway.   :)

dallincrane:
@josevalim I would love to keep working on this if that's alright!? I should be able to get some good work in on it this weekend

lib/ecto/adapters/postgres/connection.ex
dallincrane: Currently I'm thinking if I change the column_type function for non-usec types to be

```
    defp column_type(type, _opts) when type in ~w(time utc_datetime naive_datetime)a,
      do: ecto_to_db(type)
```

and adjust the `ecto_to_db` functions for not-usec types to include `(0)` that it would cover all of the scenarios.

```
    defp ecto_to_db(:time),                do: "time(0)"
    defp ecto_to_db(:time_usec),           do: "time"
    defp ecto_to_db(:utc_datetime),        do: "timestamp(0)"
    defp ecto_to_db(:utc_datetime_usec),   do: "timestamp"
    defp ecto_to_db(:naive_datetime),      do: "timestamp(0)"
    defp ecto_to_db(:naive_datetime_usec), do: "timestamp"
```

Thoughts?

Whatever way we go I'll definitely also add some tests for these `Query.type/2` scenarios!

josevalim: > That's what I thought you were referring to anyway. :)

Yes, that's what I was referring to but then I realized it doesn't matter much because if we get more precision, we will just go ahead and discard it on the Ecto side. So let's definitely leave it as is. :)

dallincrane: Ah, that's right! Gets pushed through the cast anyway. Thanks!

josevalim:
> @josevalim I would love to keep working on this if that's alright!? I should be able to get some good work in on it this weekend

Excellent!

dallincrane:
A quick list so its obvious where I'm at as I add commits

- [x] usec types for migrations
- [x] usec types for schemas
- [ ] usec types for timestamps autogen
- [ ] Ecto.Schema docs
- [ ] integration tests

dallincrane:
@josevalim This should be ready for another review. Let me know if I'm missing anything. I'm more than happy to keep contributing!

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -390,7 +390,7 @@ if Code.ensure_loaded?(Mariaex) do
 
     defp expr({:datetime_add, _, [datetime, count, interval]}, sources, query) do
       ["CAST(date_add(", expr(datetime, sources, query), ", ",
-       interval(count, interval, sources, query) | ") AS datetime)"]
+       interval(count, interval, sources, query) | ") AS datetime(6))"]
```
josevalim: I think this won't work in earlier MySQL versions. :O

lib/ecto/schema.ex
```diff
@@ -176,6 +173,33 @@ defmodule Ecto.Schema do
...
+  Having these different types allows developers to choose a type that will
+  be compatible with the database and your project's precision requirements.
+  For example, some older versions of MySQL do not support microseconds in
+  datetime fields.
+
+  When choosing what datetime type to work with, keep in mind that Elixir
+  functions like `NaiveDateTime.utc_now/0` have a default precision of 6.
+  Casting a value with a precision greater than 0 to a non-`usec` type will
+  truncate all microseconds and set the precision to 0, making the raw utc_now
+  value ≠ to the cast value.
```
josevalim: Let's spell out "different" instead of using the unicode char. :)

integration_test/cases/interval.exs
```diff
@@ -297,4 +298,39 @@ defmodule Ecto.Integration.IntervalTest do
     assert [^datetime] =
            TestRepo.all(from p in User, select: datetime_add(p.inserted_at, ^dec, "year"))
   end
+
+  test "datetime_add with naive_datetime_usec" do
+    TestRepo.insert!(%Article{published_at: ~N[2014-01-01 02:00:00.000001]})
+    dec = Decimal.new(1500)
+
+    assert [~N[2014-01-01 02:00:00.001501]] =
```
josevalim: Please add a clause where the value we are calling `datetime_add` is an interpolated value and not a database column, something like:

```elixir
  assert [~N[2014-01-01 02:00:00.001501]] =
    TestRepo.all(from p in Article, select: datetime_add(^~N[2014-01-01 02:00:00.000001], 1500, "microsecond"))
```

and the same to the test below. I think it won't work because we are casting external values to naive_datetime, which means loss of precision. If so, you will have to change `:naive_datetime` in the snippet below to `:naive_datetime_usec`:

https://github.com/elixir-ecto/ecto/blob/295482fe5323614ae71078e93929657c7538bf71/lib/ecto/query/builder.ex#L120-L121

josevalim:
@dallincrane this looks perfect, I have added only three comments, one is very minor, one is a potential bug and the other one is a compatibility issue with MySQL.

I am afraid we can't use `datetime(6)` for `datetime_add` for MySQL. At the same time I think you had to add that to make the new tests work. I am not sure how to proceed yet. Thoughts?

josevalim:
@dallincrane I have some time now so I will go ahead and merge it and do the 0.1% left. Thanks for the great contribution!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #2293, opened 2017-11-04
Custom Associations

mgwidmann:
See #2122 for reasoning. Updated gist to go with these changes https://gist.github.com/mgwidmann/ad2ef5138378dc3772a321567610520d

@josevalim I hope this is more in line with the direction you were thinking. I've tested this with a nested preload and it is working just fine for loading via `Repo.one` and `Repo.all` (unlike the previous PR).

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -1040,7 +1040,7 @@ defmodule Ecto.Association.ManyToMany do
   @doc false
   def preload(%{join_keys: [{_, owner_key}, {_, _}]} = refl, repo, query, structs, opts) do
     ids = structs |> Enum.map(&(Map.get(&1, owner_key)))
-    
+
     refl
     |> assoc_query(query, Enum.uniq(ids))
     |> Ecto.Repo.Preloader.normalize_query(refl, {-2, owner_key})
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -117,7 +117,7 @@ defmodule Ecto.Association do
   @doc """
   Preforms a preload operation (typically using `Ecto.Repo.Preloader.fetch_query/7`).
   """
-  @callback preload(t, module, Ecto.Query.t, [atom], String.t, [Ecto.Schema.t], Keyword.t) :: Ecto.Schema.t | [Ecto.Schema.t] | Ecto.Query.t
+  @callback preload(t, repo :: module, Ecto.Query.t, structs :: [Ecto.Schema.t], opts :: Keyword.t) :: [Ecto.Schema.t]
 
   @doc """
   Performs the repository change on the association.
@@ -529,14 +529,18 @@ defmodule Ecto.Association.Has do
   end
 
   @doc false
-  def preload(%{related_key: related_key, owner_key: owner_key} = refl, repo, query, fields, prefix, structs, opts) do
+  def preload(_refl, _repo, _query, [], _opts) do
+    []
+  end
+
+  @doc false
+  def preload(%{related_key: related_key, owner_key: owner_key} = refl, repo, query, structs, opts) do
     ids = structs |> Enum.map(&(Map.get(&1, owner_key)))
-    query = assoc_query(refl, query, Enum.uniq(ids))
-    Ecto.Repo.Preloader.fetch_query(structs, refl, query, prefix, {0, related_key}, fields)
-    |> case do
-      list when is_list(list) -> list
-      query -> repo.all(query, opts)
-    end
+
+    refl
+    |> assoc_query(query, Enum.uniq(ids))
+    |> Ecto.Repo.Preloader.normalize_query(refl, {0, related_key})
+    |> repo.all(opts)
   end
 
   @doc false
@@ -683,14 +687,18 @@ defmodule Ecto.Association.HasThrough do
   end
 
   @doc false
-  def preload(%{through: through, owner_key: owner_key} = refl, repo, query, fields, prefix, structs, opts) do
+  def preload(_refl, _repo, _query, [], _opts) do
+    []
+  end
+
+  @doc false
+  def preload(%{through: through, owner_key: owner_key} = refl, repo, query, structs, opts) do
     ids = structs |> Enum.map(&(Map.get(&1, owner_key)))
-    query = assoc_query(refl, query, Enum.uniq(ids))
-    Ecto.Repo.Preloader.fetch_query(structs, refl, query, prefix, through, fields)
-    |> case do
-      list when is_list(list) -> list
-      query -> repo.all(query, opts)
-    end
+
+    refl
+    |> assoc_query(query, Enum.uniq(ids))
+    |> Ecto.Repo.Preloader.normalize_query(refl, through)
+    |> repo.all(opts)
   end
 
   @doc false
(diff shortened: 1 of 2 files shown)


lib/ecto/association.ex
```diff
@@ -524,6 +529,17 @@ defmodule Ecto.Association.Has do
...
+    query = assoc_query(refl, query, Enum.uniq(ids))
+    Ecto.Repo.Preloader.fetch_query(structs, refl, query, prefix, {0, related_key}, fields)
+    |> case do
+      list when is_list(list) -> list
+      query -> repo.all(query, opts)
+    end
+  end
+
+  @doc false
   def preload_info(%{related_key: related_key} = refl) do
```
mgwidmann: I wasnt sure if you intended me to remove this callback but it does seem to be used by other parts of the code... should this be removed with this change or is it still valuable?

lib/ecto/association.ex
```diff
@@ -115,6 +115,11 @@ defmodule Ecto.Association do
               {:assoc, t, {integer, atom}} | {:through, t, [atom]}
 
   @doc """
+  Preforms a preload operation (typically using `Ecto.Repo.Preloader.fetch_query/7`).
+  """
+  @callback preload(t, module, Ecto.Query.t, [atom], String.t, [Ecto.Schema.t], Keyword.t) :: Ecto.Schema.t | [Ecto.Schema.t] | Ecto.Query.t
```
mgwidmann: I hope this isn't too many things being passed into this callback. I was trying to keep it to a minimum but if you have any suggestions on how to remove some parameters I'm all ears...

lib/ecto/repo/preloader.ex
```diff
@@ -168,18 +177,15 @@ defmodule Ecto.Repo.Preloader do
     query = update_in query.select.expr, &{:{}, [], [field, &1]}
```
mgwidmann: So, ecto doesn't have any problem with this but when using my gist if I happen to come across a polymorphic relationship where two records have the same `id` but were fetched in different tables, I believe one will overwrite the other and that may be a very subtle bug to discover. Any ideas on how that might work?

lib/ecto/repo/preloader.ex
```diff
@@ -113,52 +113,65 @@ defmodule Ecto.Repo.Preloader do
...
-  defp fetch_ids(structs, module, assoc, opts) do
+  defp fetch_structs(structs, module, assoc, opts) do
     %{field: field, owner_key: owner_key, cardinality: card} = assoc
     force? = Keyword.get(opts, :force, false)
 
     Enum.reduce structs, {[], [], []}, fn
       nil, acc ->
         acc
-      struct, {fetch_ids, loaded_ids, loaded_structs} ->
+      struct, {fetch_structs, loaded_ids, loaded_structs} ->
```
josevalim: Can you please explain why you have to return structs now?

lib/ecto/association.ex
```diff
@@ -524,6 +529,17 @@ defmodule Ecto.Association.Has do
   end
 
   @doc false
+  def preload(%{related_key: related_key, owner_key: owner_key} = refl, repo, query, fields, prefix, structs, opts) do
+    ids = structs |> Enum.map(&(Map.get(&1, owner_key)))
+    query = assoc_query(refl, query, Enum.uniq(ids))
+    Ecto.Repo.Preloader.fetch_query(structs, refl, query, prefix, {0, related_key}, fields)
+    |> case do
+      list when is_list(list) -> list
+      query -> repo.all(query, opts)
```
josevalim: When do we get a list and when do we get a query?

lib/ecto/association.ex
josevalim: Can you name the parameters? IT will help understand what they are supposed to do, for example: `opts :: Ketword.t`.

Also, regarding the result type, doesn't it always returns a list of schemas?

josevalim:
Thanks @mgwidmann this definitely helps move the conversation forward.

lib/ecto/association.ex
mgwidmann: Yeah it does always return schemas. At one point I think I had it returning a query and forgot to update it.

lib/ecto/repo/preloader.ex
mgwidmann: So from what I could understand this function divides up the ones that are already loaded from the ones we need to fetch. If I didn't return the structs that need fetching and left this function alone then I'd be passing in `structs` (which is all loaded and needing to be fetched) which would end up fetching an association that was already loaded.

lib/ecto/association.ex
mgwidmann: `fetch_query` can return either, in the case where theres an empty list or a function in the preload, this will actually return the data rather than the query. I forgot to add type specs for `fetch_query`, I'll do that too.

mgwidmann: Perhaps a better name too now that it's public if I can come up with one.

mgwidmann:
@josevalim I've refactored the preload callback to remove `prefix` and `fields` and merged them into the `query` parameter (which was actually a bit trickier than it sounds).

josevalim:
Hi @mgwidmann, I have reviewed your code.

It definitely moves in the right direction but looking at your gist and the updated code, it seems you are still dependent on internals. I am wondering if it would be best if we simply delegated everything to a custom code. 

Basically, you could  have the preload_info for your associations return something like `{:custom, &some_function/N}`, where N is the number of parameters (it may be one but it may have to be more). Then you change the `expand` function in the preloader to return `{assocs, throughs, customs}` and around this line you would invoke all customs, passing all structs and receiving all structs:

https://github.com/mgwidmann/ecto/blob/d4dc96dc1ac604c419dc9b6cdb3c49419ad711cf/lib/ecto/repo/preloader.ex#L76

This way you should have the flexibility to do whatever you want. You may need to duplicate some of the functionality in here but it will come with the guarantee there is nothing we could do on the preloader that would break your code.

What do you think?

mgwidmann:
Hmmm, ok well I definitely agree it still depends upon internals. I'll give this another look and see what I can change to help remove those internal dependencies. Just to be clear, you're suggesting [here](https://github.com/mgwidmann/ecto/blob/d4dc96dc1ac604c419dc9b6cdb3c49419ad711cf/lib/ecto/repo/preloader.ex#L80) to add something like the following:

```elixir
struct = Enum.reduce customs, struct, &load_custom/2
```

I'm assuming that for the implementation ecto will give there won't be anything done in this function? (just return the struct)

My only hesitation is that custom preloaders won't be able to benefit from any optimizations made within the ecto source since it will be just a black box to ecto. Do you see that as a potential drawback of this approach?

josevalim:
That's exactly what I mean.

> My only hesitation is that custom preloaders won't be able to benefit from any optimizations made within the ecto source since it will be just a black box to ecto. Do you see that as a potential drawback of this approach?

Yup, it means you will have to roll those optimizations yourself. It adds code duplication but it adds decoupling and you should be able to do optimizations that only apply to your cases that Ecto would have no idea about.

josevalim:
Hi 👋!

We haven't had much progress here for 3 months, so I will go ahead and close this. Feel free to reopen and continue the discussion at any time. Thank you!

imranismail:
@mgwidmann do you need help?

I've been meaning to test this out.

mgwidmann:
@imranismail I haven't had time to get back to it, if you think you can build on top of what I've done please give it a shot.

PR #2294, opened 2017-11-06
Add comment on table prefix usage with constraints

maartenvanvliet:
Make clear on how to use schema prefixes with adding constraints. Ran into this when I tried to add a constraint in a non-default postgres schema and could not find it immediately in the documentation.


josevalim:

PR #2295, opened 2017-11-06
remove unnecessary cast in datetime_add

dallincrane:
In the commit https://github.com/elixir-ecto/ecto/commit/8388d4b148ae7e6a2d87fb4eb9617b8eafdfc361 that fixes `datetime_add` microseconds for MySQL, the surrounding cast was removed as it isn't needed — a great solution to the adding microseconds compatibility problem! This PR removes that cast for Postgres as well. Although there isn't a compatibility problem, there isn't a scenario where a timestamp +/- an interval returns anything other than a timestamp.

Postgres docs reference  
https://www.postgresql.org/docs/current/static/functions-datetime.html


josevalim:

PR #2300, opened 2017-11-08
Implement Poison.Encoder for the metadata struct.

michalmuskala:
This allows us to provide a nice error message to the user instead of the
"encoder is not implemented for tuples" message they would get right now.


josevalim:

PR #2310, opened 2017-11-16
feat(Numeric Type Casting): Allow trailing whitespace when casting bi…

tomciopp:
…nary integers and floats

On occasion users of our app will accidentally type in a trailing space when dealing with numeric values which results in an "is invalid" error. I understand if you want to be extremely strict when parsing values but I believe this change is reasonable.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -523,6 +523,10 @@ defmodule Ecto.Type do
       {:ok, 1}
       iex> cast(:integer, "1")
       {:ok, 1}
+      iex> cast(:integer, "1 ")
+      {:ok, 1}
+      iex> cast(:integer, " 1")
+      :error
       iex> cast(:integer, "1.0")
       :error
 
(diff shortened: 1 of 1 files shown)


michalmuskala:
I'm not sure about this change. Thoughts @josevalim ?

wojtekmach:
In our phoenix app we have a custom plug that `String.trim`s input `params` and I believe web layer is the right layer to perform such cleanup work.

I could swear that [`Phoenix.Controller.scrub_params/2`](https://hexdocs.pm/phoenix/Phoenix.Controller.html#scrub_params/2) performed that cleanup in our app (again, it was a custom plug). @josevalim Maybe there could be `scrub_params "user", trim: true` which does that then?

josevalim:
Yup, we should probably move it up in the tack. If you want to discuss a trim option, please open up an issue on Phoenix.

tomciopp:
@wojtekmach @josevalim That's fair was implementing a custom plug for this behavior and just thought it might make sense at the casting level.

PR #2312, opened 2017-11-19
Move preprocessing out of adapters

fishcakez:
Simplify execute (and stream) by removing processing argument. This is a breaking change but it is possible to get backwards compatibility from Ecto 2.2 for an adapter by defining a default preprocessing argument as `nil` or `fn x -> x end` depending on the adapter logic.
```elixir
def execute(repo, query_meta, query, params :: list(), process \\ fn x -> x end, options)
```
I am not aware of a noSQL adapter that has supported https://github.com/elixir-ecto/ecto/pull/2155 and this PR builds on top of that in a way.

This will fix the log entry result for SQL drivers. It would be possible to do this by just changing the SQL adapter.

### DIFF
--- a/lib/ecto/repo/assoc.ex
+++ b/lib/ecto/repo/assoc.ex
@@ -7,20 +7,20 @@ defmodule Ecto.Repo.Assoc do
   Transforms a result set based on query assocs, loading
   the associations onto their parent schema.
   """
-  @spec query([Ecto.Schema.t], list, tuple) :: [Ecto.Schema.t]
-  def query(rows, assocs, sources)
+  @spec query([list], list, tuple, (list -> list)) :: [Ecto.Schema.t]
+  def query(rows, assocs, sources, fun)
 
-  def query([], _assocs, _sources), do: []
-  def query(rows, [], _sources), do: rows
+  def query([], _assocs, _sources, _fun), do: []
+  def query(rows, [], _sources, fun), do: Enum.map(rows, fun)
 
-  def query(rows, assocs, sources) do
+  def query(rows, assocs, sources, fun) do
     # Create rose tree of accumulator dicts in the same
     # structure as the fields tree
     accs = create_accs(0, assocs, sources, [])
 
     # Populate tree of dicts of associated entities from the result set
     {_keys, _cache, rows, sub_dicts} = Enum.reduce(rows, accs, fn row, acc ->
-      merge(row, acc, 0) |> elem(0)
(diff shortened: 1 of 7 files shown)


lib/ecto/repo/assoc.ex
```diff
@@ -7,20 +7,20 @@ defmodule Ecto.Repo.Assoc do
...
   """
-  @spec query([Ecto.Schema.t], list, tuple) :: [Ecto.Schema.t]
-  def query(rows, assocs, sources)
+  @spec query([list], list, tuple, (list -> list)) :: [Ecto.Schema.t]
+  def query(rows, assocs, sources, fun)
 
-  def query([], _assocs, _sources), do: []
-  def query(rows, [], _sources), do: rows
+  def query([], _assocs, _sources, _fun), do: []
+  def query(rows, [], _sources, fun), do: Enum.map(rows, fun)
```
fishcakez: We would be able to delay this `map` until next step (preloading) to minimize traversals. However if Ecto controls all processing perhaps we can do better.

AndrewDryga:
I guess it will be a ring thing to mention @elixir-ecto/adapters here

michalmuskala:
You're right @AndrewDryga. I also added @ankhers to the team since he took over the maintenance of the mongodb adapter.

scottmessinger:
Hi! I'm working on upgrading the Mongo adapter to be compatible with Ecto 3 and this is causing a problem I'm not sure how to fix and I was wondering if you all had any ideas. 

Prior to this change, if a field didn't exist in the document retrieved from the database, Ecto would use the `default` value. e.g. If the database has this:

```elixir
%{
  _id: 1,
  title: "Blog Post",
}
```

and the schema is:

```elixir
defmodule App.Post do 
  schema "posts" do
    field(:title, :string),
    field(:year, :number, default: 2020)
  end
end
```

And you were to fetch the doc: 
```elixir
doc = App.Repo.get(App.Post, 1)
```

In Ecto 2.2, Ecto fills in the non-existent field with the default. E.g

```elixir
doc == %{
  _id: 1,
  title: "Blog Post",
  year: 2020,
}
```

After this change, Ecto does not fill in the field with the default. In this example, the year is `nil`. E.g.
```elixir
doc == %{
  _id: 1,
  title: "Blog Post",
  year: nil,
}
```

In our app, we're relying on the default throughout our app and getting a ton of errors as fields we assume to be maps or lists are now nil.

Any idea how we can restore the old functionality with this new model?

PR #2315, opened 2017-11-21
Getting Started fixes: missing Repo.update and shell output arrows

jesperp:
The output of one of the examples seem to imply a Repo.update, but the function that produces the result is changeset.

Also some missing (from what I can tell) arrows that signify output


josevalim:

PR #2316, opened 2017-11-22
Be explicit that cast/4 does not validate the data.

stefanchrobot:
I would expect `cast/4` to validate the types in the passed `data`, but it does not. I think it makes sense to be explicit about this in the docs and in the tests.


josevalim:

PR #2320, opened 2017-11-24
Deprecate use Ecto.Schema in favour of import

michalmuskala:
Importing the schema and embedded_schema macros is completely enough to
acheive what we need. Import is clearer for the end user and should be
preffered over use.

But I know this can be a controversial change, so I wanted to open a discussion before merging.

I split the change in two commits - the first one does the change in code required to make it happen, the second one replaces `use` with `import` in docs & tests.

\cc @josevalim @ericmj @chrismccord @wojtekmach

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ defmodule Sample.Repo do
 end
 
 defmodule Sample.Weather do
-  use Ecto.Schema
+  import Ecto.Schema, only: [schema: 2]
 
   schema "weather" do
     field :city     # Defaults to type :string
--- a/examples/friends/lib/friends/person.ex
+++ b/examples/friends/lib/friends/person.ex
@@ -1,5 +1,5 @@
 defmodule Friends.Person do
-  use Ecto.Schema
+  import Ecto.Schema, only: [schema: 2]
 
   schema "people" do
     field :first_name, :string
--- a/guides/Associations.md
+++ b/guides/Associations.md
@@ -114,7 +114,7 @@ And the following schema:
 ```elixir
 # lib/ecto_assoc/user.ex
 defmodule EctoAssoc.User do
-  use Ecto.Schema
+  import Ecto.Schema, only: [schema: 2]
 
   schema "users" do
     field :name, :string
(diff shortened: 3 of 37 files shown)


wojtekmach:
I can add minor anecdotal evidence: when I was first learning Ecto I'd often type `import Ecto.Schema` instead of `use ...` it (because that's how one uses changesets), and I remember I was surprised it didn't work. 
So I think it would be a good addition to make it work by just importing schema macro.

It'd also mean we can change:

```elixir
defmodule Post do
  use Ecto.Schema
  import Ecto.Changeset
end
```

to

```elixir
defmodule Post do
  import Ecto.Schema, only: [schema: 2]
  import Ecto.Changeset
end
```

(https://github.com/phoenixframework/phoenix/blob/v1.3.0/priv/templates/phx.gen.schema/schema.ex#L2:L3)

which I think would be a good change too, seems more "standard".

(that being said, `import Ecto.Schema` _without_ `only:` is not great either because you'll get functions like `belongs_to` which are useless at that point. Also, yet another different story if you should have changesets in the same module as schema which was brought up in previous phoenix 1.3 discussions :))

I agree that dropping `use` might hinder future additions but obviously it's a crystal ball-y argument. One thing for sure, it'd be less than ideal to deprecate `use` now only to re-add it in Ecto 4 or something.

michalmuskala:
We could move other functions from `Ecto.Schema` to `Ecto.Schema.DSL` or something like that, so `import Ecto.Schema` would only bring `schema/2` and `embedded_schema/1`. This is even more disruptive, though, and makes documentation slightly less convenient.

josevalim:
I agree with @chrismccord. If we ever need to actually do something, such as a before_compile or after_compile, then we will have our hands tied.

I also really really want to minimize the amount of changes users will need to go through when migrating to Ecto 3.

michalmuskala:
I'm going to close this. Looking into ways to improve documentation I decided to put the callbacks we generate for `Ecto.Schema` behind a behaviour and describing them properly. This means a `@behaviour Ecto.Schema` call would be nice to have, which in turn means the `use` call is desirable.

Also having a behaviour, makes it a bit easier to explain the `use` instead of `import`.

PR #2324, opened 2017-11-27
Improve transaction handling

fishcakez:
This is the first step towards upgrading to take advantage of the latest DBConnection. This change ensures we rely fully on the database's status to get a transaction's state. There are 3 main benefits.

Transactions now mark SQL transactions as failed if attempt to begin, rollback or commit a transaction when SQL status does not allow it. For example we will reject connections that have an open transaction when we checkout from the pool and disconnect them. Similarly if we try to commit or rollback a transaction but there isn't a transaction we will return an error. This could apply if a query manually committed or rolled back the transaction.

Nested transaction now mark SQL transactions as failed if a nested transaction starts or ends outside a transaction. This works in a similar way to previous benefit, except we can detect at the start and end of a nested transaction whether we are inside a transaction. This ensures we have a strict flattened transaction.

Nested transactions now mark pg transactions as failed when a SQL error occurs that put transaction in error state without an uncaught exception or explicit rollback to notify Ecto. This means we can detect a query error causing PostgreSQL to enter the error state where all queries fail and the transaction would rollback on commit. On error for nested and outter transactions an error forces a rollback of the whole transaction. Previously we could log these as commits!

Unfortunately there is a bug in DBConnection (https://github.com/elixir-ecto/db_connection/issues/106) that is blocking this.

### DIFF
--- a/integration_test/mysql/test_helper.exs
+++ b/integration_test/mysql/test_helper.exs
@@ -4,7 +4,8 @@ Logger.configure(level: :info)
 # but that is not yet supported in travis.
 ExUnit.start exclude: [:array_type, :read_after_writes, :returning, :modify_column,
                        :strict_savepoint, :create_index_if_not_exists,
-                       :transaction_isolation, :rename_column, :with_conflict_target]
+                       :transaction_error_status, :transaction_isolation,
+                       :rename_column, :with_conflict_target]
 
 # Configure Ecto for support and tests
 Application.put_env(:ecto, :lock_for_update, "FOR UPDATE")
--- a/integration_test/sql/alter.exs
+++ b/integration_test/sql/alter.exs
@@ -35,6 +35,24 @@ defmodule Ecto.Integration.AlterTest do
     end
   end
 
+  defmodule AlterMigrationThree do
+    use Ecto.Migration
+
+    def up do
+      alter table(:alter_col_type) do
+        remove :value
+        add :value, :timestamp
+      end
+    end
+
+    def down do
+      alter table(:alter_col_type) do
+        remove :value
+        add :value, :numeric
+      end
+    end
+  end
+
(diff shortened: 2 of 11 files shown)


michalmuskala:
@fishcakez What do you think about generally changing how we handle transactions at the adapter level. 

Basically the only adapters I saw, using transactions in any way are SQL adapters and the mnesia adapter. For SQL doing separate begin/commit/rollback instead of the closure-based solution seems like a better approach, but it doesn't really work for mnesia which expects a closure.

I wonder if maybe we could move over the whole transactional support over to the separate begin/commit/rollback approach and instead of having the common case extracted into `Ecto.Adapters.SQL` have the mnesia-specific code in `EctoMnesia`.

\cc @AndrewDryga

fishcakez:
@michalmuskala do you mean for Repo to provide the transaction logic (in `Ecto.SQL.Adapters` in this PR) and adapters to provide begin/commit/rollback as adapter behavior callbacks? If an adapter doesn't support transaction it returns an error from begin? Then `EctoMnesia` would override transaction?

michalmuskala:
Yeah, you put it much better 😄

josevalim:
I would prefer to keep this SQL only because the version without functions
is harder to implement and easier to get wrong.
-- 


AndrewDryga:
Right now, having transaction implementation with function callback is much easier, just because Mnesia by itself already does that and adapter doesn't need to do anything about that. I'm not sure how to workaround that, need some time to think about it.

fishcakez:
> the version without functions
> is harder to implement and easier to get wrong.

Definitely this. I wouldn't trust myself to handle this in all but simple cases. Of course we have much more protection to prevent a mistake in SQL here but I don't think all adapters can provide this.

josevalim:
We have decided to implement this on the DBConnection side of things.

PR #2330, opened 2017-12-01
Put constraint name in error metadata for check & exclude constraints

zoldar:
In our project we have 2 check constraints on the same field. We want to be able to easily distinguish errors triggered by them without matching on error message strings.

We also propose to change other constraint functions to have this metadata shape, e.g.:

`[constraint: {type, constraint_name}`

Even though it's not common to e.g. have multiple unique constraints on the same field, the DB allows that.

This is obviously a breaking change.

An alternative for our use case is to be able to do e.g.:

```elixir
|> check_constraint(:price, name: :positive_price, metadata: [constraint: {:check, "positive_price"}])
```

(done pairing with @wojtekmach)

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2272,7 +2272,7 @@ defmodule Ecto.Changeset do
     constraint_string = to_string(constraint)
     message    = message(opts, "violates an exclusion constraint")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :exclude, constraint_string, match_type, field, {message, [constraint: :exclude, constraint_name: constraint_string]})
+    add_constraint(changeset, :exclude, constraint_string, match_type, field, {message, [constraint: :exclusion, constraint_name: constraint_string]})
   end
 
   defp no_assoc_message(:one), do: "is still associated with this entry"
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1990,9 +1990,10 @@ defmodule Ecto.Changeset do
   """
   def check_constraint(changeset, field, opts \\ []) do
     constraint = opts[:name] || raise ArgumentError, "must supply the name of the constraint"
+    constraint_string = to_string(constraint)
     message    = message(opts, "is invalid")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :check, to_string(constraint), match_type, field, {message, [constraint: :check]})
+    add_constraint(changeset, :check, constraint_string, match_type, field, {message, [constraint: :check, constraint_name: constraint_string]})
   end
 
   @doc """
@@ -2264,9 +2265,10 @@ defmodule Ecto.Changeset do
   """
   def exclusion_constraint(changeset, field, opts \\ []) do
     constraint = opts[:name] || "#{get_source(changeset)}_#{get_field_source(changeset, field)}_exclusion"
+    constraint_string = to_string(constraint)
     message    = message(opts, "violates an exclusion constraint")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :exclude, to_string(constraint), match_type, field, {message, [constraint: :exclusion]})
+    add_constraint(changeset, :exclude, constraint_string, match_type, field, {message, [constraint: :exclude, constraint_name: constraint_string]})
   end
 
   defp no_assoc_message(:one), do: "is still associated with this entry"
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1993,7 +1993,7 @@ defmodule Ecto.Changeset do
     constraint_string = to_string(constraint)
     message    = message(opts, "is invalid")
     match_type = Keyword.get(opts, :match, :exact)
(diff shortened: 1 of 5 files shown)


lib/ecto/changeset.ex
```diff
@@ -1990,9 +1990,10 @@ defmodule Ecto.Changeset do
   """
   def check_constraint(changeset, field, opts \\ []) do
     constraint = opts[:name] || raise ArgumentError, "must supply the name of the constraint"
+    constraint_string = to_string(constraint)
     message    = message(opts, "is invalid")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :check, to_string(constraint), match_type, field, {message, []})
+    add_constraint(changeset, :check, constraint_string, match_type, field, {message, [constraint: {:check, constraint_string}]})
```
wojtekmach: implementation-wise going with our 1st proposal would mean we could replace this with:

```elixir
add_constraint(changeset, :check, constraint_string, match_type, field, message)
```

and inside`add_constraint` we could change:

```elixir
%{changeset | constraints: [%{type: type, constraint: constraint, ..., error: error} | constraints])
```

to remove `error: error` because it's redundant: it can be computed from `type` and `constraint` in the place where we convert constraint errors to changeset errors.

wojtekmach:
proposal 3: convert to changeset errors in the following way, e.g.:

```elixir
changeset.errors == [from: {"violates an exclusion constraint", [constraint: :exclude, constraint_name: "cannot_overlap"}]}]

changeset.errors == [uuid: {"has already been taken", [constraint: :unique, constraint_name: "posts_uuid_index"]}]
```

imho not quite as elegant as proposal 1, but it's backwards-compatible.

josevalim:
We don't need to worry about backwards compatibility because the commit that added the constraint key to metadata is only on master. So, regardless of our decision here, I believe we should change all constraints to behave the same. We cannot add `constraint: :unique` in some cases and `constraint: {:exclusion, ...}` to others.

IMO, I would prefer to use two keys, `:constraint` and `:constraint_name`. One of the use cases for metadata is to use it in internationalization and having distinct keys means we can use the keys directly without matching on them. I also think the constraint name should come from the adapter, since we have features that allow matching on the prefix or the suffix of the constraint.

So I propose to:

1. We always add the `:constraint` key to metadata (today it is not done consistently)
2. Once the repo/adapter converts a constraint into an error, we prepend `:constraint_name` to the metadata

Thoughts?

michalmuskala:
I like that approach.

josevalim:
@zoldar I have gone ahead and added the :constraint key to check and exclusion constraints in master since someone was asking for it on IRC. All that is left is the adding of the constraint_name based on the adapter information. Can you please take a look at that? Thanks!

zoldar:
@josevalim Sure, will do!

zoldar:
I have rebased the PR and added `constraint_name` to returned meta in all constraint functions.

Additionally, I've made a slight change, renaming `constraint` meta for `exclusion` to `exclude`:

https://github.com/elixir-ecto/ecto/pull/2330/files#diff-da02bd57818ebeb0b1941e76bf4c3d88L68

I'm not sure whether the original name choice was deliberate but `exclude` seems to be more in line with what's passed to `add_constraint` for that case. I also see however, that for instance, `no_assoc` is an exception (for understandable reasons). If that change does not feel right, I'll happily revert it.

\cc @wojtekmach @josevalim

lib/ecto/changeset.ex
```diff
@@ -1990,9 +1990,10 @@ defmodule Ecto.Changeset do
   """
   def check_constraint(changeset, field, opts \\ []) do
     constraint = opts[:name] || raise ArgumentError, "must supply the name of the constraint"
+    constraint_string = to_string(constraint)
     message    = message(opts, "is invalid")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :check, to_string(constraint), match_type, field, {message, [constraint: :check]})
+    add_constraint(changeset, :check, constraint_string, match_type, field, {message, [constraint: :check, constraint_name: constraint_string]})
```
wojtekmach: given all constraint functions have a very similar shape now we could remove some duplication by calling `add_constraint` like this:

```elixir
add_constraint(changeset, :check, constraint_string, match_type, field, message)
```

in the case where constraint _type_ doesn't match what we want to put into metadata (e.g.: exclude&exclusion, no_assoc&foreign_key etc), we could do this:

```elixir
add_constraint(changeset, :exclude, constraint_string, match_type, field, message, :exclusion)
```

josevalim:
I am sorry because maybe I was not clear enough. I believe it would be better if the constraint name was added here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L555-L578

That's because the changeset functions may make partial matches on the constraint name and we only have the full name in the repository. So I think `add_constraint` should only store the error message and **all** metadata should come from the function linked above.

josevalim:
Actually, I will go ahead and do the remaining changes, since most of the work here is done. :) Thanks!

josevalim:

zoldar:
@josevalim 

> I am sorry because maybe I was not clear enough.

Sorry, I have rushed the changes instead of asking for clarifications.

Happy to be at least of some help though :)

josevalim:
No worries, I am glad you spotted this and everything now works consistently.

PR #2332, opened 2017-12-06
Fix query traversal for :update_all using MySQL adapter

### LINKED ISSUE #2303: Bad ordering of interpolated expression when running `update_all`
### Environment

* Elixir version (elixir -v): Elixir 1.5.2 (Erlang/OTP 20)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): MySQL 5.7.16
* Ecto version (mix deps): 2.2.6
* Operating system: macOS High Sierra version 10.13.1

### Current behavior

When I run an update_all, it doesn't update the specified row. Examining the query that is produced shows that it fails due to sticking the join interpolated value at the end instead of the beginning of the list:

```elixir
user = %User{id: 2, email: "foo@example.com", first_name: "needs", last_name: "changing"}
changeset = User.changeset(user, %{first_name: "José", last_name: "Valim"})
query = 
  from p in "people", join: u in "users",
  on: p.id == u.person_id and u.id == ^changeset.data.id,
  update: [set: ^Map.to_list(changeset.changes)]

Ecto.Adapters.SQL.to_sql(:update_all, Repo, query)
# => {"UPDATE `people` AS p0 INNER JOIN `users` AS u1 ON (p0.`id` = u1.`person_id`) AND (u1.`id` = ?) SET p0.`first_name` = ?, p0.`last_name` = ?",
#    ["José", "Valim", 2]}
```

### Expected behavior

I expect the list of values to use in the query to be in the correct order so that my update will work:
`[2, "José", "Valim"]`

zoldar:
Closes #2303 

This is my first stab at solving the issue. Up for scrutiny :)

The source of the problem is incorrect expression traversal order when processing the query. It affects the order in which parameters are extracted, which, in the specific case of MySQL `UPDATE`/`update_all` operation, differs from the actual query constructed by MySQL adapter. Namely, `JOIN` expressions are before `SET` ones. To illustrate the issue, here's the same query built using 2 different adapters:

```elixir
# query in Ecto DSL
id = Ecto.UUID.generate()
query = 
  from p in "people", join: u in "users",
  on: p.id == u.person_id and u.id == ^id,
  update: [set: [first_name: ^"Tom", last_name: ^"Jones"]]

# result of `to_sql` for Postgres adapter - correct parameters order
{"UPDATE \"people\" AS p0 SET \"first_name\" = $1, \"last_name\" = $2 FROM \"users\" AS u1 WHERE ((p0.\"id\" = u1.\"person_id\") AND (u1.\"id\" = $3))",
 ["Tom", "Jones", "72c5acf3-c465-4c8b-9c66-b48c11cd1386"]}

# result of `to_sql` for MySQL adapter - incorrect parameters order
{"UPDATE `people` AS p0 INNER JOIN `users` AS u1 ON (p0.`id` = u1.`person_id`) AND (u1.`id` = ?) SET p0.`first_name` = ?, p0.`last_name` = ?",
 ["Tom", "Jones", "72c5acf3-c465-4c8b-9c66-b48c11cd1386"]}
```

As mentioned above, the order of parameters is directly dependent on the order of traversal of different parts/expressions in the query (`:froms`, `:updates`, `:joins` etc.). Currently, the order is hardwired in `@exprs` module attribute in `Ecto.Query.Planner`.

This PR changes two things:

1. It moves the keyword list of expressions under `Ecto.Adapters.SQL` under `expressions_for_traversal/1` callback function (added to `Ecto.Adapter` behavior), accepting a single atom argument designating the action in context of which it's called (like `:update_all`). 

2. It overrides the new callback for MySQL adapter.

The proposed changes are not breaking backwards compatibility (at least they shouldn't).


josevalim:
Great catch @zoldar! Before we go on this route, do you think it is possible to rewrite the MySQL query using WHERE or a similar syntax to PostgreSQL?

josevalim:
And regarding the test, it would probably be best to move it to test/ecto/adapters/mysql_test.exs :)

zoldar:
> Great catch @zoldar! Before we go on this route, do you think it is possible to rewrite the MySQL query using WHERE or a similar syntax to PostgreSQL?

I looked into that first - however I've only checked MySQL's `UPDATE` docs. At least in this case, I don't see a way to write the query differently while supporting full range of `JOIN`s. There's a simplified syntax for referring to multiple tables and joining them in `WHERE` conditions but it basically equals `(INNER) JOIN`, so it's pretty limited.

I didn't check that possibility for Postgres though. I'll look into that sometime later today hopefully.

> And regarding the test, it would probably be best to move it to test/ecto/adapters/mysql_test.exs :)

Ouch, sure, will do :+1:

josevalim:
> I looked into that first - however I've only checked MySQL's UPDATE docs. At least in this case, I don't see a way to write the query differently while supporting full range of JOINs.

I see. To be honest, I would rather also limit MySQL to only support inner joins (same as Postgres) - that's preferable than pushing complexity to all adapters. That also aligns with the behaviour in SQL Server (which follows PostgreSQL).

zoldar:
@josevalim 

> I see. To be honest, I would rather also limit MySQL to only support inner joins (same as Postgres) - that's preferable than pushing complexity to all adapters. That also aligns with the behaviour in SQL Server (which follows PostgreSQL).

Oh shoot, somehow it escaped me that Postgres doesn't support other types of joins 🙈 

I'll look into changing that then.

josevalim:
@zoldar ping! :)

zoldar:
@josevalim Hey 😅 I have the query syntax change ready. I'll put it out in another PR (while closing this one) either today or tomorrow.

josevalim:
No rush, just checking if you were still tackling this or someone else could take it. :)

zoldar:
Closing in favor of #2345

PR #2335, opened 2017-12-07
Raise error when not association passed to Ecto.assoc_loaded?/1

Apelsinka223:
I noticed that `Ecto.assoc_loaded?/1` returns true on any input except `%Ecto.Association.NotLoaded{}`. I think it can be source of errors when somebody will try to put atom instead of entry `Ecto.assoc_loaded?(:assoc)` or something else.

### DIFF
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -531,8 +531,8 @@ defmodule Ecto do
 
   """
   def assoc_loaded?(%Ecto.Association.NotLoaded{}), do: false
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @Apelsinka223! I have just removed the catch-all clause in favor of a FunctionClauseError, which are quite reliable from Elixir v1.5 and OTP 20+.

josevalim:

PR #2342, opened 2017-12-12
change comment in schema has_many :through docs

michallepicki:
I find the current comment confusing. It does not describe what's happening in the code. I'd either change it to this one or remove it.

edit: I know it does describe what happens underneath, but I think the placement is unfortunate. Code comments describe the purpose of code and neither of these two lines' purpose is getting comments. It is a side effect of (or the means of getting the value of) the second line, and if we want to mention that side effect, I think it should be noted that it is a side effect.


josevalim:

PR #2345, opened 2017-12-13
Use stricter syntax for update_all with joins in MySQL adapter

### LINKED ISSUE #2303: Bad ordering of interpolated expression when running `update_all`

zoldar:
Closes #2303 

Refs #2332 

This PR addresses the issue #2303 by way of supporting only a strict subset of `JOIN` syntax for `update_all` queries, as suggested in https://github.com/elixir-ecto/ecto/pull/2332#issuecomment-349582570.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -102,9 +102,9 @@ if Code.ensure_loaded?(Mariaex) do
         update_fields(:update, query, sources)
       end
 
-      join   = join(query, sources)
+      {join, wheres} = using_join(query, :update_all, sources)
       prefix = prefix || ["UPDATE ", from, " AS ", name, join, " SET "]
-      where  = where(query, sources)
+      where  = where(%{query | wheres: wheres ++ query.wheres}, sources)
 
       [prefix, fields | where]
     end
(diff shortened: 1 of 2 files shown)


josevalim:

Issue #1432, opened 2016-05-14
Ecto.Changeset.put_assoc() fails when trying to change belongs_to association (sometimes)


carusso:
Elixir: 1.2.4
Postgres: 9.5.2
ecto 2.0.0-rc.5 (Hex package) (mix)
phoenix_ecto 3.0.0-rc.0 (Hex package) (mix)
postgrex 0.11.1 (Hex package) (mix)
OSX 10.11.3

### Following simple models were created:
```
defmodule Demo.Club do
  use Demo.Web, :model

  schema "clubs" do
    field :name, :string
    has_many :members, Demo.Member
    
    timestamps
  end
end

defmodule Demo.Member do
  use Demo.Web, :model

  schema "members" do
    field :name, :string
    belongs_to :club, Demo.Club

    timestamps
  end
end
```

### The following test to put_assoc succeeds
```
  test "can associate existing member with existing club" do
    member = Repo.insert!(%Member{name: "Dave"}) |> Repo.preload(:club)
    club = Repo.insert!(%Club{name: "Wentworth"})

    changeset = member
                |> Changeset.change
                |> Changeset.put_assoc(:club, club)
    assert changeset.valid?
    member = Repo.update! changeset
    assert member
  end

```
### The following test fails trying to change the association using actual db model references
```
  test "can change association for existing member with existing club" do
    member = Repo.insert!(%Member{name: "Dave"}) |> Repo.preload(:club)
    club = Repo.insert!(%Club{name: "Wentworth Club"})
    club2 = Repo.insert!(%Club{name: "Palm Club"})

    changeset = member
                |> Changeset.change
                |> Changeset.put_assoc(:club, club)
    assert changeset.valid?
    member = Repo.update! changeset
    assert member

    changeset = member
                |> Changeset.change
                |> Changeset.put_assoc(:club, club2)
    assert changeset.valid?
    member = Repo.update! changeset
    assert member
  end
```
  
### of the actual Member from the insert!().
```
  test "can change association for existing member with existing club using struct reference for member" do
    member = Repo.insert!(%Member{name: "Dave"}) |> Repo.preload(:club)
    club = Repo.insert!(%Club{name: "Wentworth Club"})
    club2 = Repo.insert!(%Club{name: "Palm Club"})

    changeset = member
                |> Changeset.change
                |> Changeset.put_assoc(:club, club)
    assert changeset.valid?
    member = Repo.update! changeset
    assert member

    changeset = %Member{id: member.id}
[...]

PR #2350, opened 2017-12-15
Update docs for Ecto.Changeset where :on_replace is :nilify

axelson:
Even though I've previously read issue #1432 the knowledge did not stick and I didn't correctly set `on_replace` to `:nilify`. Hopefully this doc update will help.

The intent of this PR is to make it more clear what to set `on_replace` to if you want to allow updating a `belongs_to` association.


axelson:
Awesome, thanks!

PR #2352, opened 2017-12-19
Add `migration_default_prefix` to repo config

coryodaniel:
When you are working with a database that *always* uses an alternate
schema prefix to `public`, allow setting a default for all migrations
to use. From the model standpoint its easy to set @schema_prefix in a
module then use that module in your models, from the migration
perspective you have to remember to run migrations with `--prefix` or
set the prefix on each DDL statement which is prone to forgetfulness :D

The default is overridable at the DDL level by specifying
`prefix: "foo"` on a create table, index, or drop statement.

It is also overridable by passing `--prefix NAME` to the runner.

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -174,6 +174,11 @@ defmodule Ecto.Migration do
 
           config :app, App.Repo, migration_lock: nil
 
+    * `:migration_default_prefix` - Ecto defaults to `nil` for the database prefix for
+      migrations but you can configure it via:
+
+          config :app, App.Repo, migration_default_prefix: "kewl"
+
   """
 
   defmodule Index do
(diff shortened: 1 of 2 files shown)


lib/ecto/migration.ex
```diff
@@ -921,11 +926,12 @@ defmodule Ecto.Migration do
 
   @doc false
   def __prefix__(%{prefix: prefix} = index_or_table) do
+    default_prefix = Runner.repo_config(:migration_default_prefix, nil)
     runner_prefix = Runner.prefix()
```
michalmuskala: I wonder if the `Runner.prefix()` call should handle this logic internally.

josevalim: Sounds like a good candidate, yeah!

coryodaniel: I had the runner handle it internally, but then the exception condition below raised. To make it handle overriding and not raising an exception on the match I ended up checking to see if the `migration_default_prefix` was set or not, and at that point I figured since I already had to pull in the value, that this approach had the same effect with less code.

josevalim: @coryodaniel why did the exception raised though? Why were both being given?

coryodaniel: It raised when it was internal to `Runner.prefix()` it fails the following scenario, which I think we should allow to be overridden.

```elixir
@tag repo_config: [migration_default_prefix: "baz"]
  test "forward: creates a table with prefix from migration overriding the default prefix configuration" do
    create(table(:posts, prefix: "foo"))
    flush()

    {_, table, _} = last_command()
    assert table.prefix == "foo"
  end
```

josevalim:
Thanks for clarifying, I agree with your conclusions! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

coryodaniel:
Awesome! Thanks for all your hard work and making my job easier! @josevalim @michalmuskala

PR #2357, opened 2017-12-22
Allow string without size on pg

chulkilee:
Although there is no performance difference between `text` and `varchar` without
limit, users may want to use `:string`, not `:text`, in migrations.

The existing default size 255 for `:string` is preserved. However when
`nil` is given for `:size` on `:string`, do not set the length limit.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -837,6 +837,16 @@ if Code.ensure_loaded?(Postgrex) do
       end
     end
 
+    defp column_type(:string = type, opts) do
+      size      = Keyword.get(opts, :size, 255)
+      type_name = ecto_to_db(type)
+
+      case size do
+        nil -> type_name
+        _   -> [type_name, ?(, to_string(size), ?)]
+      end
+    end
+
     defp column_type(type, opts) do
       size      = Keyword.get(opts, :size)
       precision = Keyword.get(opts, :precision)
@@ -844,10 +854,9 @@ if Code.ensure_loaded?(Postgrex) do
       type_name = ecto_to_db(type)
 
       cond do
-        size            -> [type_name, ?(, to_string(size), ?)]
-        precision       -> [type_name, ?(, to_string(precision), ?,, to_string(scale || 0), ?)]
-        type == :string -> [type_name, "(255)"]
-        true            -> type_name
+        size      -> [type_name, ?(, to_string(size), ?)]
+        precision -> [type_name, ?(, to_string(precision), ?,, to_string(scale || 0), ?)]
+        true      -> type_name
       end
     end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Apologies but I don't understand the argument. :text is the correct type to use here, so why not use it?

chulkilee:
- In PostgreSQL `varchar` without length and `text` are still different type (although they use the same internal structure)
- For a column which contains small string but unknown length (e.g. post title), using `text` seems to give wrong impresson that the column is intended to store very long text.
  - Forcing the use of `:text` for any unlimited string makes the purpose of the column less clear
Let's see this example:



```elixir
create table(:post1) do
  add :title, :text
  add :body, :text
end

create table(:post2) do
  add :title, :string, size: nil
  add :body, :text
end
```

They behaves same - except actual type of `:title` columns are different between two tables. However, I think using `:string` with `size: nil` shows the exact intention of person who wrote the migration.


BTW I think for PostgreSQL `:string` default size should not be set (so no limit), like Rails. If db adopters requires size, then those adopters should require them...

josevalim:
> Forcing the use of :text for any unlimited string makes the purpose of the column less clear

So, I don't agree with the conclusion. :) Text doesn't have to mean long, it rather means unlimited, which is how we should go here.

chulkilee:
> So, I don't agree with the conclusion. :) Text doesn't have to mean long, it rather means unlimited, which is how we should go here.

@josevalim I don't think *everyone* have to agree with that to include that feature. What I said is that someone *may* want to  use `varchar` and `text` for different purpose, *although* they works same.

Even if you thin that's meaningless (or even wrong), since PostgreSQL *does* keep the information (not using `text` when `varchar` without length given), so I believe Ecto should allow passing it, unless Ecto is more opinionated than just just a "wrapper".

However, I found I could just use `:varchar` - so this PR is not needed :)

OvermindDL1:
> BTW I think for PostgreSQL :string default size should not be set (so no limit), like Rails. If db adopters requires size, then those adopters should require them...

I do very much agree with this.  I've read on multiple occasions that unbounded varchar's in PostgreSQL execute faster than bounded varchar's due to a variety of optimizations unless they are very very short (mostly interning and so forth).  Having 255 as a default was indeed quite surprising for me when I came to Ecto long ago.

michalmuskala:
We set the size to provide the same behaviour across different databases.

I really don't think a lot of "if you're using this db" things in migration docs. The more consistency we can provide, the better.

OvermindDL1:
> I really don't think a lot of "if you're using this db" things in migration docs is a great idea. The more consistency we can provide, the better.

Even regardless of the database, still surprising that varchar is default sized, since it is not default sized in SQL, and I'd expect it to match SQL semantics for less friction.

PR #2360, opened 2017-12-25
correctly log validate_confirmation in Changeset

voltone:
All other validations record `{field, {:validation_type, opts}}` in the Changeset's `validations` field, but `validate_confirmation` logs `{:validation, opts}` instead. I believe this is a regression introduced in #1739.


josevalim:

Issue #2365, opened 2018-01-09
Joining two queries together results in query planning error


zachdaniel:
### Environment

* Elixir version (elixir -v): 1.5.1
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgresql 9.6
* Ecto version (mix deps): 2.2.7
* Database adapter and version (mix deps): postgrex 0.13.3
* Operating system: OSX

### Current behavior

```elixir
alias MyApp.Repo
import Ecto.Query

q1 = from row in MyApp.Post, where: row.published == true
q2 = from row in MyApp.User, where: row.super == false

qx = 
  from post in q1,
    join: user in ^q2,
    on: user.id == post.user_id

Repo.all(qx)
```

raises

```
** (Ecto.QueryError) deps/ecto/lib/ecto/query/planner.ex:18: field `super` in `join` does not exist in schema MyApp.Post in query:

from a0 in MyApp.Post,
  join: a1 in MyApp.User,
  on: a1.super == false and a1.id == a0.user_id,
  where: a0.published == true,
  select: a0

    (elixir) lib/enum.ex:1357: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir) lib/enum.ex:1357: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir) lib/enum.ex:1357: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir) lib/enum.ex:1811: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto) lib/ecto/repo/queryable.ex:124: Ecto.Repo.Queryable.execute/5
    (ecto) lib/ecto/repo/queryable.ex:37: Ecto.Repo.Queryable.all/4
iex(6)>
```

### Expected behavior

It should correctly build the where clauses as `on` statements and not apply them to the main query that I'm joining to.

EDIT: Fixed an alias name in the example.

josevalim:
Thanks @zachdaniel!

Is the error really reporting this: `join: a1 in MyApp.,`? Notice the alias is incomplete.

zachdaniel:
No, my apologies! I went through and replaced our model names, and got that wrong. Fixing now.

josevalim:
[...]

PR #2368, opened 2018-01-10
adding join test case

zachdaniel:
A test case for #2365


zachdaniel:
Sorry, this is wrong. Fixing now.

PR #2371, opened 2018-01-11
Add {:replace, columns}, closes #2370

### LINKED ISSUE #2370: Update specific fields from EXCLUDED table on conflict
Is it possible to do something like this on Postgres?

```
Repo.insert_all(Struct, entries, on_conflict: [set: [:foo, :bar])
```

I found solution via fragment (#1853) but this way I can't provide ```on_conflict``` value as an argument (I can't use ```fragment``` outside of ```from``` scope). Eventually I'd like to make the next code work:

```
def save(struct, entries, opts \\ []) do
  Repo.insert_all(struct, entries, opts)
end

save(MyStruct, [%MyStruct{foo: "foo", bar: "bar", bazz: "bazz"}], on_conflict: [set: [:foo, :bar])
```

I've checked *Ecto.Repo.Schema.on_conflict/6* and concluded it's just not supported at all. Are there any hacks here? :)

Many thanks!

IcyEagle:


josevalim:

PR #2372, opened 2018-01-11
Correct Default Message for validate_confirmation

mfcollins3:
The documentation for validate_confirmation contained the wrong message
for the default error message if no error message was provided. The
documentation indicated that the default error message was "does not
match," but the actual default error message in the code is "does not
match confirmation." I corrected the documentation to match the source
code.


josevalim:

PR #2373, opened 2018-01-11
add an exclusion for all the other join types

BenMorganIO:
This should ideally add support for excluding `:inner_join`, `:cross_join`, `:left_join`, `:right_join`, :full_join`, `, :inner_lateral_join`, and finally `:left_lateral_join`.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -99,14 +99,14 @@ defmodule Ecto.Query do
 
   `nil` comparison in filters, such as where and having, is forbidden
   and it will raise an error:
-  
+
       # Raises if age is nil
       from u in User, where: u.age == ^age
-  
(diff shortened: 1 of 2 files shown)


josevalim:
Beautiful. Can we have some tests and updated docs as well? Thanks!

BenMorganIO:
@josevalim just finished that before you asked. I've moved that case statement to a pattern matched private function. I was trying to figure out how to keep it all DRY with the tests but I gave in to hard coding it all. Not that big of a problem until SQL decides a new join type; which probably won't happen for a really long time.

josevalim:

BenMorganIO:
And the docs: https://github.com/elixir-ecto/ecto/pull/2374

PR #2374, opened 2018-01-11
update docs for Ecto.Query.exclude/2 with more examples and new joins

BenMorganIO:


lib/ecto/query.ex
```diff
@@ -475,11 +475,34 @@ defmodule Ecto.Query do
   @doc """
   Resets a previously set field on a query.
 
-  It can reset any query field except the query source (`from`).
+  It can reset many fields except the query source (`from`).
```
BenMorganIO: Considering I discovered it couldn't update any query field such as inner_join, probably write "many" instead in case any else finds others that are missing.

lib/ecto/query.ex
```diff
@@ -475,11 +475,34 @@ defmodule Ecto.Query do
...
 
   ## Example
 
-      query |> Ecto.Query.exclude(:select)
+      Ecto.Query.exclude(query, :join)
```
BenMorganIO: I should write a note that says that this will remove _all_ joins and not just the default inner_join.

josevalim: Yes, please!

lib/ecto/query.ex
```diff
@@ -475,11 +475,34 @@ defmodule Ecto.Query do
...
+      Ecto.Query.exclude(query, :select)
+      Ecto.Query.exclude(query, :limit)
+      Ecto.Query.exclude(query, :offset)
+      Ecto.Query.exclude(query, :lock)
+      Ecto.Query.exclude(query, :preload)
+
+  You can also remove specific joins as well such as `left_join` and
+  `inner_join`.
+
+  ## Example
```
josevalim: You don't need this other exemple header in here.

lib/ecto/query.ex
```diff
@@ -475,11 +475,34 @@ defmodule Ecto.Query do
   @doc """
   Resets a previously set field on a query.
 
-  It can reset any query field except the query source (`from`).
+  It can reset many fields except the query source (`from`).
 
   ## Example
```
josevalim: Can you use the opportunity to rename this to "## Examples"? Tks!

lib/ecto/query.ex
BenMorganIO: added

lib/ecto/query.ex
BenMorganIO: updated

lib/ecto/query.ex
BenMorganIO: removed

josevalim:

Issue #2325, opened 2017-11-27
Make Changeset.cast accept only atoms


michalmuskala:
Today, is't quite confusing that some changeset functions accept strings for field names, while others accept atoms. On couple occasions there were people confused about this on slack channel, who didn't know why their `cast` call was working OK, while `validate_required` was failing.

I suggest changing `cast` to accept only a list of atoms.

Unfortunately, this might be a disruptive change and I'm not sure what would be the best way to handle it, or even if it's really worth it. Maybe a warning instead of an outright error would be better?

josevalim:
Definitely a warning.

ybod:
Can I take a look at this?

josevalim:
Yes pls.

PR #2375, opened 2018-01-12
Output warning when using non-atom list in cast

alexgaribay:
Adds a warning whenever the `permitted` param contains non-atom values in `cast`.

Addresses #2325.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -437,6 +437,7 @@ defmodule Ecto.Changeset do
 
   defp cast(%{} = data, %{} = types, %{} = changes, :invalid, permitted, opts) when is_list(permitted) do
     {empty_values, _opts} = Keyword.pop(opts, :empty_values, @empty_values)
+    validate_permitted_list(permitted)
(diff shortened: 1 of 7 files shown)


lib/ecto/changeset.ex
```diff
@@ -465,6 +467,12 @@ defmodule Ecto.Changeset do
                           message: "expected params to be a :map, got: `#{inspect params}`"
   end
 
+  defp validate_permitted_list(permitted) do
+    if Enum.any?(permitted, &not is_atom(&1)) do
```
josevalim: I used `not` instead of `!` since is_atom is guaranteed to return a  boolean.

lib/ecto/changeset.ex
```diff
@@ -465,6 +467,12 @@ defmodule Ecto.Changeset do
...
+      IO.warn("expected permitted parameters given to cast/3 to be a list of atoms, got: `#{inspect permitted}`")
```
josevalim: I tried to make it a bit clearer. :)

josevalim:

Issue #2317, opened 2017-11-22
Parent updated_at not changing when adding children 


hisapy:
### Precheck

This is considered a possible bug, based on conversation with @michalmuskala on #ecto Slack channel.

### Environment

* Elixir version (elixir -v): **1.4.2**
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): **PostgreSQL 9.6.3**
* Ecto version (mix deps): **2.2.6**
* Database adapter and version (mix deps): **postgrex 0.13.3**
* Operating system: **macOS Sierra 10.12.6, Ubuntu 10.04 LTS** 

### Current behavior

Given an existing record associated with zero or more children, adding children to the association via the parent's changeset, doesn't change `updated_at` on parent. 

For example, given the following schema

```elixir
# parent
defmodule Webapp.Company do 
  schema "companies" do 
    field :name, :string
    has_many :members, Webapp.CompanyMember, on_replace: :delete
    # other fields ...
    timestamps()
  end
  
  def changeset(struct, params \\ %{}) do
    struct
    |> cast(params, [:name])
    |> validate_required(:name)
    |> cast_assoc(:members)
  end
end

# child
defmodule Webapp.CompanyMember do
  use Webapp.Web, :model

  schema "company_members" do
    belongs_to :company, Webapp.Company
    field :is_executive, :boolean
    # other fields ...
    timestamps()
  end

  def changeset(model, params \\ %{}) do
    model
    |> cast(params, [:company_id, :is_executive])
    |> assoc_constraint(:company)
  end
end 
```

... and given an existing `Company`, adding new members to it, doesn't update its `updated_at`.

```elixir
# args is coming from the outside (client)
args = %{"members" => [%{"isExecutive" => false}]} 
company
 |> Repo.preload(:members)
 |> Company.changeset(args)
 |> Repo.update 
```

### Expected behavior

Given an existing `Company` adding members to it should update its `updated_at`. A behavior similar to passing `args` like above but with data for `:name` - or any other `Company` _specific_ field, or like the following:

```elixir
company
 |> Repo.preload(:members)
 |> Company.changeset(args)
 |> Repo.update(force: true)
```

but without _bypassing the logic to skip query if there are no changes in the changeset_

josevalim:
Thank you for the report @hisapy. I would like to hear why changing the child should update the parent. Is it because we are using cast_assoc? Should that also apply to put_assoc? Wouldn't it be better to do it under a specific flag?

[...]

PR #2376, opened 2018-01-13
Add force update option for casting relations

alexgaribay:
Adds an option on `cast_assoc/3` and `cast_embeds/3` to add `[force: true]` in the `repo_opts` on a changeset.

Addresses #2317.

Note: I wasn't quite sure how to validate that record's `updated_at` is updated in the test suite.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -669,6 +669,7 @@ defmodule Ecto.Changeset do
     * `:required` - if the association is a required field
     * `:required_message` - the message on failure, defaults to "can't be blank"
     * `:invalid_message` - the message on failure, defaults to "is invalid"
+    * `:force_update_on_change` - if the update should mark the parent struct as updated
   """
   def cast_assoc(changeset, name, opts \\ []) when is_atom(name) do
     cast_relation(:assoc, changeset, name, opts)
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset.ex
```diff
@@ -669,6 +669,7 @@ defmodule Ecto.Changeset do
     * `:required` - if the association is a required field
     * `:required_message` - the message on failure, defaults to "can't be blank"
     * `:invalid_message` - the message on failure, defaults to "is invalid"
+    * `:force_update_on_change` - if the update should mark the parent struct as updated
```
josevalim: We should change the description to something like: "force the "parent" schema to be updated in the repository if there is a change in the association, defaults to true".

lib/ecto/changeset.ex
```diff
@@ -719,6 +721,13 @@ defmodule Ecto.Changeset do
         {changeset, false}
       end
 
+    changeset =
+      if opts[:force_update_on_change] do
```
josevalim: `Keyword.get(opts, :force_update_on_change, true)` since we want to default to true.

alexgaribay:
It should be good for a second look now.

lib/ecto/changeset.ex
```diff
@@ -719,6 +723,7 @@ defmodule Ecto.Changeset do
         {changeset, false}
       end
 
+    forced_update? = Keyword.get(opts, :force_update_on_change, true)
```
josevalim: You can pass the opts to `force_update` and read it there. There is no need to read it before if we are not going to use it. :)

josevalim:
One last comment and we should be good to go!

josevalim:

josevalim:

hisapy:
Thank you guys 
😃

PR #2385, opened 2018-01-19
Add .formatter.exs for from/2

taiansu:
As elixir 1.6 introduce format feature, the Ecto query with `from/2` will be formatted like
```elixir
from(
   p in Post,
   select: p.title
)
```

If we like to encourage the use of documented syntax, by default (which is my favorite part of Ecto, aesthetically)
```
from p in Post,
  select: p.title
```

I propose to add this `.formatter.exs`, for the library dependent on Ecto, like Phoenix, can then import it.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,0 +1,6 @@
+[
+  locals_without_parens: [from: 2],
+  export: [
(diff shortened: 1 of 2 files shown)


josevalim:

PR #2386, opened 2018-01-21
Fix ConstraintError message for exclude constraint

arcz:
Before the change, the error message was advising to call `exclude_constraint/3` but the function is named `exclusion_constraint/3`. I've tested messages for other constraints to be sure they match actual function name. I have also a suggestion about the message copy itself, what do you think about this:

If you would like to convert this **exception into a changeset error**,
please call `#{func}_constraint/3` **on** your changeset with the constraint
`:name` as an option.


josevalim:

josevalim:
improvements to the error message are also welcome.

PR #2390, opened 2018-01-22
Move errors to the top of the error message

blatyo:
Really long error messages can get truncated when written to things like
logs. This moves the most useful information about a changeset error to
the top, so that it won't get truncated in most cases.


josevalim:

PR #2391, opened 2018-01-23
add Changeset.validation_specs/1

pragdave:
Joeé:

This is just the first part of the stuff we talked about. I wanted to check it was inline with your expectations.

I called it `Changeset.validation_specs` because I felt that plain `validations` implied the running of them.

Let me know if its OK and I'll finish it off.

Dave

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1213,6 +1213,7 @@ defmodule Ecto.Changeset do
     update_in changeset.changes, &Map.delete(&1, key)
   end
 
+  
   @doc """
   Applies the changeset changes to the changeset data.
 
@@ -1274,6 +1275,62 @@ defmodule Ecto.Changeset do
   end
 
   ## Validations
+
+    @doc """
+  Returns a keyword list of the validations for this changeset.
+
+  The keys in the list are the names of fields, and the values are a
+  validation associated with the field. A field may occur multiple
+  times in the list.
+
+  A quick example:
+
+      iex> %Post{}    |>
+      ...> change()   |>
+      ...> validate_format(:title, ~r/^\w+:\s/,
+      ...>      message: "must start with a topic")  |>
+      ...> validate_length(:title, max: 100)
+      ...> validation_specs()
+      [
+        title: { :length, [ max: 100 ]},
+        title: { :format, ~r/^\w+:\s/ }
+      ]
+
+  The following validations may be included in the result. 
+  
+  This first group contains validations that take a keyword list of validators,
+  where the validators are show immediately following the validation type. 
+  This list may also include a `message:` key.
+
+    * `{ :length, [ «validators»… ] }`
+
+      * `min: n`
+      * `max: n`
+      * `is: n`
+      * `count: :graphemes | :codepoints`
+
+  * `{ :number,  [ «validators»… ] }`
+    
+     * `equal_to: n`
+     * `greater_than: n`
+     * `greater_than_or_equal_to: n`
+     * `less_than: n`
+     * `less_than_or_equal_to: n`
+
+  The other validators simply take a value:
+
+  * `{ :exclusion, «Enum.t» }`
+  * `{ :format, ~r/pattern/ }`
+  * `{ :inclusion, «Enum.t» }`
+  * `{ :subset, «Enum.t» }`
+
+  (since: 1.7.0)
+  """
+
+  @spec validation_specs(t) :: [ atom: { atom, ([ atom: any ] | any) } ]
+  def validation_specs(%Changeset{validations: validations}) do
+    validations
+  end
 
   @doc """
   Adds an error to the changeset.
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1289,17 +1289,19 @@ defmodule Ecto.Changeset do
       ...> change()   |>
       ...> validate_format(:title, ~r/^\w+:\s/,
       ...>      message: "must start with a topic")  |>
-      ...> validate_length(:title, max: 100)
-      ...> validation_specs()
+      ...> validate_length(:title, max: 100)         |>
+      ...> validation_list()
       [
         title: { :length, [ max: 100 ]},
         title: { :format, ~r/^\w+:\s/ }
       ]
 
-  The following validations may be included in the result. 
+  The following validations may be included in the result. The list is
+  not necessarily exhaustive. For example, custom validations written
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset.ex
```diff
@@ -1275,6 +1276,62 @@ defmodule Ecto.Changeset do
...
+     * `greater_than_or_equal_to: n`
+     * `less_than: n`
+     * `less_than_or_equal_to: n`
+
+  The other validators simply take a value:
+
+  * `{ :exclusion, «Enum.t» }`
+  * `{ :format, ~r/pattern/ }`
+  * `{ :inclusion, «Enum.t» }`
+  * `{ :subset, «Enum.t» }`
```
michalmuskala: I'm not sure about documenting all the possible values this way, since it's generally open-ended. The values provided here are generally supplied by the `validate_change/4` function. If you have a custom validation function (or things from some library) it can put just about any value in there - documenting only the things provided by the `Ecto.Changeset` module itself might be surprising for the user when they get something different.

josevalim: I think it is good to document them but we should make it clear that any value can be added and that it is open ended.

josevalim:
@pragdave thanks! I am fine with changing the name away from `validations` but I don't like the name `specs` because it is already an overloaded term in Elixir. What do you think about `get_validations` or something that means "executed validations"?

lib/ecto/changeset.ex
```diff
@@ -1275,6 +1276,62 @@ defmodule Ecto.Changeset do
...
+  times in the list.
+
+  A quick example:
+
+      iex> %Post{}    |>
+      ...> change()   |>
+      ...> validate_format(:title, ~r/^\w+:\s/,
+      ...>      message: "must start with a topic")  |>
+      ...> validate_length(:title, max: 100)
+      ...> validation_specs()
```
mememori: missing a pipe

lib/ecto/changeset.ex
```diff
@@ -1275,6 +1276,62 @@ defmodule Ecto.Changeset do
...
+
+  * `{ :exclusion, «Enum.t» }`
+  * `{ :format, ~r/pattern/ }`
+  * `{ :inclusion, «Enum.t» }`
+  * `{ :subset, «Enum.t» }`
+
+  (since: 1.7.0)
+  """
+
+  @spec validation_specs(t) :: [ atom: { atom, ([ atom: any ] | any) } ]
```
mememori:
```elixir
@spec validation_specs(t) :: Keyword.t({atom, Keyword.t | term})
```
or
```elixir
@spec validation_specs(t) :: [{atom, {atom, [{atom, term}] | term}]
```

lib/ecto/changeset.ex
```diff
@@ -1275,6 +1276,62 @@ defmodule Ecto.Changeset do
...
+      ...> validation_specs()
+      [
+        title: { :length, [ max: 100 ]},
+        title: { :format, ~r/^\w+:\s/ }
+      ]
+
+  The following validations may be included in the result. 
+  
+  This first group contains validations that take a keyword list of validators,
+  where the validators are show immediately following the validation type.
```
mememori: i think the correct here is _shown_ but not sure :P

test/ecto/changeset_test.exs
```diff
@@ -614,6 +614,29 @@ defmodule Ecto.ChangesetTest do
...
+      result = 
+        changeset(%{})
+        |> validation_specs()
+      
+      assert result == []
+    end
+
+    test "returns the appropriate validations" do
+      result =
+        changeset(%{})
```
mememori: i think that, according to the styleguidelines, the first param of the first function of a pipe should be on it's own line, so

```elixir
result =
  %{}
  |> changeset()
  |> ...
```

lib/ecto/changeset.ex
pragdave: José:

I'm a little wary of looking like a getter (even though the function is), simply because it sends the wrong message. How about `all_validations` or `validation_list`?

test/ecto/changeset_test.exs
pragdave: I've changed my code, and the other 20 or so occurrences of this already there. 

However, in future I'd suggest the style guidelines be amended so that this only applies if the parameter is the actual starting value of the pipeline. In this case, the parameter was a default passed to the function.

michalmuskala: Yeah, I agree with you @pragdave here. This is one of the checks that credo does, but we have it disabled for ecto, because in many cases the starting value isn't really part of the pipeline (e.g. the data type of the first element is different from what you pipe through the other calls).

wojtekmach:
Out of curiosity: since `changeset.validations` is a public field on the struct, wouldnt it be enough to keep the documentation on the field, of there are reasons to have s function for this?

Just throwing it out there: Have you considered calling it `validators`? It would be awkward to use both validations and validators in the codebase  but to me it would be obvious that such function wouldnt do any transformations but just return unchanged data.

pragdave:
This is the end result of a very long email thread between José and myself.

The upshot of it all was that we agreed that there are two kinds of field
in structures: those that are part of the publicly accessible state (think
DateTime.year) and those that are intended for internal use (think
Regex.re_version).

In order to reduce coupling, we shouldn't expect users of a module to
directly access those internal fields, because if we change the internal
implementation we could break their code. This is bad, as the
implementation is supposed to be independent of the API.

So the idea is to wrap any access to these "private" fields in a function
call. If we need to change the implementation, we can keep the same API by
updating the function to deal with the changes.

This all came about because I objected to Phoenix.Ecto poking around in the
%Changeset{} struct.


Dave

josevalim:
> I'm a little wary of looking like a getter (even though the function is), simply because it sends the wrong message.

Good point.

Let's go with `validations` then. Ecto does not have a separate stage for defining the validations and another for running them, so the confusion here should exist. We can also be clear about it in the docs.

wojtekmach:
right, thanks @pragdave! Just saw another example: https://hexdocs.pm/elixir/master/Date.Range.html#t:t/0, where we show the fields in the docs, but explicitly tell not to use some of them.

josevalim:
@wojtekmach we should put those fields under private types so their format do not appear in the doc either. Similar to how we do here: https://hexdocs.pm/elixir/master/Macro.Env.html#t:t/0

josevalim:
Thanks @pragdave. I have merged it locally as `validations` and also added `constraints`.

Issue #2394, opened 2018-01-25
Error while working with DEFERRED foreign key constraint (Postgres)


jsangilve:
### Environment

* Elixir version (elixir -v):  1.5-1.6
* Database and version (PostgreSQL 9.4-9.6):
* Ecto version (mix deps): >= 2.1.x
* Database adapter and version (mix deps): postgrex 0.13.3
* Operating system: Debian 8-9

### Current behavior

I'm using Ecto (2.2.8) to work with an existent `PostgreSQL` database (wasn't generated through Ecto's migrations). I defined two schemas to represent `owner` and `house`. The `house` schema has a FK (`belongs_to`) `owner`. The `owner`'s schema and change were defined as follows:

```elixir
     @primary_key {:id, :id, autogenerate: true}
      schema "house" do
        belongs_to :owner, Owner, foreign_key: :owner_id
        field :name, :string
      end
    
      def changeset(house, params \\ %{}) do
        house
        |> cast(params, [:name, :owner_id])
        |> validate_required([:owner_id])
        |> foreign_key_constraint(:owner_id)
      end
```

I expected the following code to return a tuple `{:error, changeset}`, when there isn't an `owner` record with `id` `10` in database:

```elixir
    House.changeset(%House{}, %{name: "Whatever",  owner_id: 10}) |> Repo.insert
```

This raises a **Postgrex.Error**: 

`
** (Postgrex.Error) ERROR 23503 (foreign_key_violation): insert or update on table "house" violates foreign key constraint "house_owner_id_fkey".`

If I check the content of `changeset.constraints` before calling `Repo.insert`, this is what I get:

```elixir
    House.changeset(%House{}, %{name: "Whatever",  owner_id: 10}) 
    |> Map.get(:constraints)

    [
      %{
        constraint: "house_owner_id_fkey",
        error: {"does not exist", []},
        field: :owner_id,
        match: :exact,
        type: :foreign_key
      }
    ]
```

The `Postgrex.Error` is raised when the database `FOREIGN KEY` constraint is defined as `DEFERRED`:

`"house_owner_id_fkey" FOREIGN KEY (owner_id) REFERENCES owner(id) DEFERRABLE INITIALLY DEFERRED`

I re-created the database using Ecto.Migration and realized that Foreign Keys are created as `NOT DEFERRED` by default

`"house_owner_id_fkey" FOREIGN KEY (datacenter_id) REFERENCES datacenter(id)`

[...]

PR #2396, opened 2018-01-28
Add docs about limitations with deferred foreign key constraints

jsangilve:
This PR add info to the docs as discussed in #2394


michalmuskala:
The text looks good to me. But since this is not really specific to foreign keys, I wonder if maybe we should place this somewhere else. I'm not really sure where - maybe in the module docs where we talk about constrains?

josevalim:

PR #2397, opened 2018-01-28
Fix a few places that don't use `__schema__(:query)`

zachdaniel:
I figured it was best to call into the protocol, but I could also do them just as `schema.__schema__(:query)` if necessary.

These changes basically seem like innocent changes that should be expected behavior, but also let me potentially implement workarounds for #2395. And if that actually gets implemented, then this is important too.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -507,7 +507,7 @@ defmodule Ecto.Association.Has do
   def joins_query(%{queryable: queryable, related_key: related_key,
                     owner: owner, owner_key: owner_key}) do
     from o in owner,
-      join: q in ^queryable,
+      join: q in ^Ecto.Queryable.to_query(queryable),
       on: field(q, ^related_key) == field(o, ^owner_key)
   end
 
@@ -768,7 +768,7 @@ defmodule Ecto.Association.BelongsTo do
   def joins_query(%{queryable: queryable, related_key: related_key,
                     owner: owner, owner_key: owner_key}) do
     from o in owner,
-      join: q in ^queryable,
+      join: q in ^Ecto.Queryable.to_query(queryable),
       on: field(q, ^related_key) == field(o, ^owner_key)
   end
 
@@ -953,7 +953,7 @@ defmodule Ecto.Association.ManyToMany do
     [{join_owner_key, owner_key}, {join_related_key, related_key}] = join_keys
     from o in owner,
       join: j in ^join_through, on: field(j, ^join_owner_key) == field(o, ^owner_key),
-      join: q in ^queryable, on: field(j, ^join_related_key) == field(q, ^related_key)
+      join: q in ^Ecto.Queryable.to_query(queryable), on: field(j, ^join_related_key) == field(q, ^related_key)
   end
 
   @doc false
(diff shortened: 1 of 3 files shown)


zachdaniel:
I didn't add tests yet because I know this might be a questionable change (and I'm not sure what exactly to test, given that its not recommended to alter the query underlying these schemas at the current time anyway).

Feel free to tear it apart!

lib/ecto/association.ex
```diff
@@ -507,7 +507,7 @@ defmodule Ecto.Association.Has do
   def joins_query(%{queryable: queryable, related_key: related_key,
                     owner: owner, owner_key: owner_key}) do
     from o in owner,
-      join: q in ^queryable,
```
josevalim: I am afraid this is not the correct place to fix this. Somewhere internally we handle `join: q in Foo` and we need to make sure the conversion happens there.

lib/ecto/query/builder/from.ex
```diff
@@ -58,18 +58,19 @@ defmodule Ecto.Query.Builder.From do
         schema when is_atom(schema) ->
           # Get the source at runtime so no unnecessary compile time
           # dependencies between modules are added
-          source = quote do: unquote(schema).__schema__(:source)
-          prefix = quote do: unquote(schema).__schema__(:prefix)
-          {1, query(prefix, source, schema)}
+          source = quote do: Ecto.Queryable.to_query(unquote(schema))
```
josevalim: The reason we had this code is because the query builder tries to generate as much as possible from the query at compilation time. By making this dynamic, we lose this flexibility.

josevalim:
Thanks @zachdaniel!

I don't think we should go in this direction before we have a plan in hand because I don't want developers to override `__schema__(:queryable)`. For all purposes, it should be a private API and we do important optimizations because of it.

lib/ecto/query/builder/join.ex
```diff
@@ -62,15 +62,15 @@ defmodule Ecto.Query.Builder.Join do
   end
 
   def escape({:__aliases__, _, _} = module, _vars, _env) do
-    {:_, {nil, module}, nil, %{}}
+    {:_, quote(do: Ecto.Queryable.to_query(unquote(module))), nil, %{}}
```
josevalim: Same thing here. It is probably best to do it on the planner and not when the query is being built.

lib/ecto/query/builder/from.ex
michalmuskala: It should be fine to call it at compile-time, though. Should be basically the same as we do manually today.

josevalim: No because the optimizer only works if the query API we are invoking receives an `Ecto.Query` at *compilation time*. This is effectively hiding the query behind a runtime call, so it will all be lost.

michalmuskala: I'm saying we could call the protocol _at compile-time_. It shouldn't be different from doing what the implementation does manually.

josevalim: No, we cannot do that, because otherwise you can easily generate deadlocks. For example, imagine some query in MyApp.User using a join on Post and a query on MyApp.Post doing a join on User.

josevalim: Also the query returned by `__schema__(:query)` returns a value which we would need to escape and I am not sure we can guarantee all parameters in the query would actually be "escapable".

Plus it would add compile time dependencies between files, which we have generally avoided.

zachdaniel: @josevalim I definitely can get on board w/ not introducing compile time dependencies :D.

PR #2400, opened 2018-01-31
Implementing infix mathematical operators

### LINKED ISSUE #1757: Support mathematical infix operators

costaraphael:
Closes #1757 

I tried to think about Ecto as a really thin layer over the database while implementing this. As a result, any kind of mathematical operation that the database supports (like dates and intervals on PostgreSQL) should work just fine.

I also changed the `type/2` macro to work with mathematical operations, so you can enforce return types and change how parameters will be casted to the database:

```elixir
# casts the parameter as a float and converts the result to float
from p in Post, select: type(p.visits / ^"2", :float)
```

The drawbacks I found in this implementation are:

- as it simply delegates the operations to the database, we don't have compile time errors (unless you try to do some crazy stuff, like `1 == 1 + (1 > 1)`);
- the queries are not fully portable between databases, as some operations can be fundamentally different in PostgreSQL, MySQL or whichever database you might use (e.g. `1 / 2` returns `0` in PostgreSQL and `#Decimal<0.5>` in MySQL).

There isn't much we can do for the second one without losing the "thin layer" status. As for the first one, maybe we could add some more checks (like allowing only some certain types in these operations) at the expense of increasing the distance to the database.

What do you guys think?

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -860,7 +860,7 @@ defmodule Ecto.Type do
   # Checks if a value is of the given primitive type.
   defp of_base_type?(:any, _),         do: true
   defp of_base_type?(:id, term),       do: is_integer(term)
-  defp of_base_type?(:float, term),    do: is_float(term)
+  defp of_base_type?(:float, term),    do: is_float(term) or is_integer(term)
   defp of_base_type?(:integer, term),  do: is_integer(term)
   defp of_base_type?(:boolean, term),  do: is_boolean(term)
   defp of_base_type?(:binary, term),   do: is_binary(term)
--- a/test/ecto/query/builder_test.exs
+++ b/test/ecto/query/builder_test.exs
@@ -20,6 +20,10 @@ defmodule Ecto.Query.BuilderTest do
     assert {Macro.escape(quote do &0.y > &1.z end), %{}} ==
            escape(quote do x.y > y.z end, [x: 0, y: 1], __ENV__)
 
+    import Kernel, except: [+: 2]
+    assert {Macro.escape(quote do &0.y + &1.z end), %{}} ==
+           escape(quote do x.y + y.z end, [x: 0, y: 1], __ENV__)
+
     assert {Macro.escape(quote do avg(0) end), %{}} ==
            escape(quote do avg(0) end, [], __ENV__)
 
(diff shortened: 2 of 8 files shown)


michalmuskala:
This looks good to me on first sight.

There's a missing test for `-` - I don't think it's checked in any of the tests, we should cover it for completeness.

costaraphael:
> There's a missing test for - - I don't think it's checked in any of the tests, we should cover it for completeness.

@michalmuskala Good catch!

I'll add a test for it and remove that trailing space that I missed :sweat_smile:

test/ecto/query/builder_test.exs
```diff
@@ -20,6 +20,10 @@ defmodule Ecto.Query.BuilderTest do
     assert {Macro.escape(quote do &0.y > &1.z end), %{}} ==
            escape(quote do x.y > y.z end, [x: 0, y: 1], __ENV__)
 
+    import Kernel, except: [+: 2]
+    assert {Macro.escape(quote do &0.y + &1.z end), %{}} ==
```
josevalim: Let's also add a unit test that calls `type(x.y + x.z, x.y)`.

lib/ecto/type.ex
```diff
@@ -860,7 +860,7 @@ defmodule Ecto.Type do
   # Checks if a value is of the given primitive type.
   defp of_base_type?(:any, _),         do: true
   defp of_base_type?(:id, term),       do: is_integer(term)
-  defp of_base_type?(:float, term),    do: is_float(term)
+  defp of_base_type?(:float, term),    do: is_float(term) or is_integer(term)
```
josevalim: I don't like this change because it means we would be able to write `%{measurement | intensity: some_integer}` even when intensity is tagged as `float`. What happens if we remove this?

One option is to try to mark all of them as decimal because decimal accept integers, floats and even binaries, mirroring closer to the database behaviour.

josevalim:
Thanks @costaraphael! I really like this. Just two comments and we should be good to go!

josevalim:
Tackled locally and merged manually, thank you! :heart:

costaraphael:
@josevalim Awesome! I'm going to close that other PR then ^^

lib/ecto/type.ex
costaraphael: Sorry, I didn't see this comment. Without this, I can't use integers in a operation that requires a float, like `type(p.views / 2, :float)` (`p.views / 2` works just fine IIRC).

My reasoning behind this change is that although you can't use a float where an integer is expected, the only operation I can think off where the opposite is true is the `is_float/1` function.

So I didn't see as much of an issue allowing integers where floats are expected. Thoughts?

test/ecto/query/builder_test.exs
costaraphael: I can add this one on the documentation PR ^^

josevalim:
I believe I have addressed both comments already in a later commit, so no worries!

PR #2401, opened 2018-02-01
Migrate from Poison to Jason

josevalim:


CHANGELOG.md
```diff
@@ -27,7 +33,15 @@ And should be rewritten as:
...
-This means Ecto no longer requires Ecto-specific extensions on databases. This leads to better integration with 3rd party libraries and faster compilation times. Ecto also now requires the JSON library to implement an `encode_to_iodata!/1` function instead of `encode!` in line with the interface required by adapters such as Postgrex.
+If you want to rollback to Poison, you need to configure your adapter accordingly:
+
+    # For Postgres
+    config :postgrex, :json_library, Poison
+
+    # For MySQL
+    config :mariaex, :json_library, Poison
+
+We recommend everyone to migrate to Jason. Built-in support for Poison will be removed in future Ecto 3.x releases.
```
michalmuskala: Maybe we should say something less definite, like:
```
We recommend everyone to migrate to Jason. Built-in support for Poison will be removed in future Ecto 3.x releases if breaking changes in Poison occur.
```

josevalim: I still want to reserve the right to remove it. Since this is 3.0, we could have removed it now, but it wouldn't be the best upgrade experience. :)

josevalim:

uri:
What makes Jason a better choice over Poison? I couldn't find any discussion in the elixir-ecto mailing list.

josevalim:
The main difference is that it is more performant (while still being
implemented in pure Elixir).
-- 

Issue #2404, opened 2018-02-03
Add common DSLs to .formatter.exs


darkbaby123:
Hello guys, we recently upgraded our app (with Phoenix & Ecto) to Elixir 1.6. And we currently add many DSL functions to `.formatter.exs`. Below are some for Ecto:

```elixir
[
  locals_without_parens: [
    # Ecto

    ## schema
    field: :*,
    belongs_to: :*,
    has_one: :*,
    has_many: :*,
    many_to_many: :*,
    embeds_one: :*,
    embeds_many: :*,

    ## migration
    create: :*,
    create_if_not_exists: :*,
    alter: :*,
    drop: :*,
    drop_if_exists: :*,
    rename: :*,
    add: :*,
    remove: :*,
    modify: :*,
    execute: :*
  ]
]
```

 I think they're very common for most of the users. If they're in the Ecto's `.formatter.exs` then developers can simply `import_deps: [:ecto]`. If you agree I'd like to make a PR for it.

josevalim:
That's the plan! Although we think we could handle migrations in a slightly different way. We are not quite sure yet.

josevalim:
We can start with the schema one though. Could you please send a PR? Thank you!

darkbaby123:
The schema DSLs are added. But I found that adding things to both `locals_with_parens` and `export` is duplicated. I can extract a variable in `.formatter.exs` like below:

```elixir
locals_without_parens = [
  dsl_1: :arity,
  dsl_2: :arity
]

[
  locals_without_parens: locals_without_parens,
  export: [
    locals_without_parens: locals_without_parens
  ]
]
```

It works for `mix format`, but I'm not sure if it's a preferred way?

darkbaby123:
After checking the source of `mix format`, I found that the `.formatter.exs` is evaluated with `Code.eval_file/2`. Since we only use evaluation result but not bindings, it's safe to use variables in `.formatter.exs`. The PR is also updated.

josevalim:
Yup, it is totally safe. Closing this in favor of the PR. :)

hauleth:
It seem that `.formatter.exs` has `export` section, but it seem that this file isn't present in release which mean that `import_deps: [:ecto]` does nothing.

josevalim:
Pushed to v2.2. It iwll be fixed in the next release.

rbino:
[...]

PR #2405, opened 2018-02-04
Add schema DSLs to .formatter.exs

darkbaby123:
ref #2404

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,6 +1,32 @@
+locals_without_parens = [
+  # Query
+  from: 2,
+
+  # Schema
+  field: 1,
+  field: 2,
+  field: 3,
+  timestamps: 0,
+  timestamps: 1,
+  belongs_to: 2,
+  belongs_to: 3,
+  has_one: 2,
+  has_one: 3,
+  has_many: 2,
+  has_many: 3,
+  many_to_many: 2,
+  many_to_many: 3,
+  embeds_one: 2,
+  embeds_one: 3,
+  embeds_one: 4,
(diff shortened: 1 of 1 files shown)


.formatter.exs
```diff
@@ -1,6 +1,20 @@
+locals_without_parens = [
+  # Query
+  from: 2,
+
+  # Schema
+  field: :*,
```
josevalim: Can we please be explicit about the arities? `:field` should be 1, 2, 3. The others should be 2 and 3. :)

darkbaby123: No problem.

.formatter.exs
```diff
@@ -1,6 +1,30 @@
+locals_without_parens = [
+  # Query
+  from: 2,
+
+  # Schema
+  field: 1,
```
fertapric: Should `schema`, `embedded_schema` and `timestamps` be also be included? The doc examples have all of them without parens https://hexdocs.pm/ecto/Ecto.Schema.html#content

josevalim: We don't need schema and embedded_schema because they are used in do-blocks and those don't have parens by default. We should add timestamps/1 though. Good catch!

darkbaby123: After Elixir 1.5, functions without arity is required to have parens. For `timestamps` I think if we need parens "sometimes", it's not consistent and I prefer always adding parens. That's the reason I didn't added it originally.

One example:

```elixir
# must have parens
timestamps()

# no parens
timestamps inserted_at: :created_at
```

What do you think?

josevalim:

Issue #2389, opened 2018-01-22
Support named joins


josevalim:
## Tasks

  * [x] Allow multiple options to be given to join/5
  * [x] Support the `:as` construct in queries
  * [x] Include `:as` on inspected queries
  * [x] Update the documentation
  * [x] Add support for `:as` in `from` too

## The problem

A commonly asked feature about Ecto is to support named joins. The idea is that, if you need to compose a query by parts, you may want to refer to previous joins. The only way to do so in Ecto today is via positional bindings:

      query = Post

      # Filter by the join
      query = from p in query,
                join: c in Comment, where: c.post_id == p.id

      # Extend the query
      query = from [p, c] in query,
                select: {p.title, c.body}

There is also the `...` identifier that helps select the last bindings:

      # Extend the query
      query = from [p, ..., c] in query,
                select: {p.title, c.body}

However, those approaches are still restrictive, as they require you to track when a given join was added. 

For those reasons, having a mechanism that refers to a previously established join would be very useful.

## Solution

There were previous attempts to address this issue but all of them attempted to conflate the positional binding with the named join. This approach can be very error prone because in Ecto the name of the binding has no effect on the generated query. So if bindings suddenly had to be unique, many queries would fail.

Another approach would be to identify the joins by their schema but that is also limited as you would be unable to join the same schema/table more than once, which may be desired from time to time.

So it is clear that the only approach to solve this is by introducing an explicit naming mechanism. We propose to do so via the `:as` option in queries:

      query = Post

      # Filter by the join
      query = from p in query,
                join: c in Comment, as: :comments, where: c.post_id == p.id

      # Extend the query
      query = from [p, comments: c] in query,
                select: {p.title, c.body}

If the `:as` key is present, Ecto will do the following:

  1. Has another join be specified with the `:as` key? If so, we raise
  2. Otherwise we store it for further lookups

We could even leverage this solution to include the `:as` names in the resulting SQL query, as it may be handy in certain situations.

## Feedback

Your turn. :)

OvermindDL1:
First some glee.  Oh My Word This Would Simplify *SOOO* Much Of My Code And Shorten My Compile Time Whoooo!!!

Now that that is out of the way.  ^.^

I'm not sure why the `:as` needs to be in the join query, instead of this:

```elixir
query = from p in query,
            join: c in Comment, as: :comments, where: c.post_id == p.id
```

Why not do this:

```elixir
query = from p in query,
[...]

Issue #2392, opened 2018-01-24
Allow passing table hints to the db via Ecto


JamesFerguson:
### Environment

* Elixir version (elixir -v): 1.5.1
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): MySQL 5.6.38
* Ecto version (mix deps): 2.1.6
* Database adapter and version (mix deps): mariaex 0.8.1
* Operating system: OSX 10.11.6

### Current behavior

Previously discussed on SO here: https://stackoverflow.com/questions/48394148/how-to-give-an-index-hint-to-mysql-through-ecto

Please bear in mind I'm an Elixir/Ecto noob.

I need to pass an index hint to MySQL in an Ecto query and it seems this is not currently supported.

The syntax for MySQL is defined [here](https://dev.mysql.com/doc/refman/5.7/en/index-hints.html). It seems from a brief googling that it's not standard SQL (Postgres doesn't support it at all) and each DB adapter would need to handle it differently.

For example, if the following query needs manual optimization:

```
SELECT *
FROM posts AS p
  JOIN comments AS c ON c.post_id = p.id
```

I might want the following query produced:

```
SELECT *
FROM posts AS p
  JOIN comments AS c USE INDEX(comments_index) ON c.post_id = p.id
```

I have tried a number of hacks, like:

```
from p in Post
  join: c in fragment("comments AS c USE INDEX(comments_index) ON c.post_id = p.id")

# produces the SQL

SELECT *
FROM posts AS p
  JOIN (comments AS c USE INDEX(comments_index) ON c.post_id = p.id) AS f1

# which is obviously invalid SQL syntax
```

and also

```
from p in Post
  join: c in "comments AS c USE INDEX(comments_index) ON c.post_id = p.id"

# produces the SQL

SELECT *
FROM posts AS p
  JOIN `comments AS c USE INDEX(comments_index) ON c.post_id = p.id`

# which is results in a 'table name too long' error
```

### Expected behavior

Ideally there'd be a feature added so I could do:

```
from p in Post
  join: c in Comment
  use: "comments_index"

# produces the SQL

SELECT *
FROM posts AS p
  JOIN comments AS c USE INDEX(comments_index) ON c.post_id = p.id
```

if nothing else it would be nice if I could do something like:

```
from p in Post
  join: c in fragment("comments AS c USE INDEX(comments_index) ON c.post_id = p.id", raw: true)

# produces the SQL

SELECT *
FROM posts AS p
[...]

PR #2411, opened 2018-02-09
[WIP] Add support for named joins

zoldar:
Closes https://github.com/elixir-ecto/ecto/issues/2389

I took a stab at implementing support for named joins as discussed in #2389, using the ideas from initial post and https://github.com/elixir-ecto/ecto/issues/2389#issuecomment-359481044.

Example usage:

```elixir
query =
  Post
  |> join(:inner, [p], {:comment, c} in assoc(p, :comments))
  |> order_by([comment: c], c.id)
  |> select([p, comment: c], {p, c})
```

Marking it as work in progress because I'm definitely not sure whether 
  1. the API is acceptable 
  2. it's implemented optimally. 

Other than that, I'll want to refactor `Query.Builder.escape_binding` for sure, because there's a lot of redundancy and excessive nesting at the moment. Also, docs will need update eventually as well when/if the proposed change gets into acceptable shape.

Feedback appreciated.

### DIFF
--- a/test/ecto/query/builder/dynamic_test.exs
+++ b/test/ecto/query/builder/dynamic_test.exs
@@ -11,7 +11,9 @@ defmodule Ecto.Query.Builder.DynamicTest do
   end
 
   defp named_query do
-    from p in "posts", join: {:comments, c} in "comments", on: p.id == c.post_id
+    from p in "posts",
+      join: {:authors, a} in "authors", on: a.id == p.author_id,
+      join: {:comments, c} in "comments", on: p.id == c.post_id
   end
 
   describe "fully_expand/2" do
@@ -83,12 +85,12 @@ defmodule Ecto.Query.Builder.DynamicTest do
       dynamic = dynamic([comments: c], c.bar == ^"bar")
       assert {expr, _, params, _, _} = fully_expand(named_query(), dynamic)
       assert Macro.to_string(expr) ==
-        "&1.bar() == ^0"
-      assert params == [{"bar", {1, :bar}}]
+        "&2.bar() == ^0"
+      assert params == [{"bar", {2, :bar}}]
     end
 
     test "with ... and named join" do
-      dynamic = dynamic([p, ..., comments: c], p.foo == c.bar and c.baz == ^"baz")
+      dynamic = dynamic([p, ..., authors: a], p.foo == a.bar and a.baz == ^"baz")
       assert {expr, _, params, _, _} = fully_expand(named_query(), dynamic)
       assert Macro.to_string(expr) ==
         "&0.foo() == &1.bar() and &1.baz() == ^0"
--- a/test/ecto/query/builder/dynamic_test.exs
+++ b/test/ecto/query/builder/dynamic_test.exs
@@ -10,6 +10,10 @@ defmodule Ecto.Query.Builder.DynamicTest do
     from p in "posts", join: c in "comments", on: p.id == c.post_id
   end
 
+  defp named_query do
+    from p in "posts", join: {:comments, c} in "comments", on: p.id == c.post_id
+  end
+
   describe "fully_expand/2" do
     test "without params" do
       dynamic = dynamic([p], p.foo == true)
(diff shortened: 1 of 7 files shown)


josevalim:
Thanks for the PR @zoldar! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

This is great I have a couple suggestions though.

1. I still prefer `join c in "comments", as: :comments` because it allows us to also use the `:as` option when generating SQL queries. It is more verbose but it is also clearer. For example, imagine reading `join: {:comments, c} in {"comments", Comment}` by using tuples it is really hard to figure out what is what.

2. The way you compute the names right now is very expensive because it is linear based on the number of joins. My suggestion is to add a new field to Ecto.Query called aliases and store it there. Seeing if a field exists or not should be a simple lookup.

Because this PR is potentially quite long, I would recommend breaking this into smaller tasks and we work step by step towards the end result:

  1. The first step is to make `Ecto.Query.join/5` support keyword list as last argument. Today `join/5` expects the last argument to a list of expressions:

        join(Post, :left, [p], c in assoc(p, :comments), c.public)

    We need to change it to support a keyword list as last argument and emit a warning if a keyword list is not given. So we want this:

        join(Post, :left, [p], c in assoc(p, :comments), on: c.public)

    This will also be necessary for us to support index hinting (#2392) in the future. But for now we will support only `:on` anything else will raise.

  2. Now that we made joins extensible, we can add support to `:as` in both `join` and `from`. The aliases should be stored in the query, so lookups are fast. Trying to register an already given name should be an error. Trying to use a name that does not exist should be an error too.

  3. Used `:as` should also be included in the generated SQL statements too.

Thoughts?

test/ecto/query_test.exs
```diff
@@ -268,6 +268,42 @@ defmodule Ecto.QueryTest do
     end
   end
 
+  describe "named joins" do
+    test "assigns a name to a join" do
```
josevalim: Those tests are very fragile because they are asserting on the AST. The reason why we test them above is because we want to check if the query is optimizable but we are not worried about optimization here. My recommendation is to build actual queries and then traverse `query.joins` (or `query.aliases` if we follow my proposal) checking for the generated aliases.

test/ecto/query/builder_test.exs
```diff
@@ -124,6 +124,13 @@ defmodule Ecto.Query.BuilderTest do
            escape(quote(do: ^([] ++ [])), [], __ENV__)
   end
 
+  test "escape_binding with named binding in the input" do
```
josevalim: This is a fragile test too, we don't need to check the generated AST on expansion.

test/ecto/query/builder/dynamic_test.exs
```diff
@@ -66,5 +72,29 @@ defmodule Ecto.Query.Builder.DynamicTest do
              "&0.foo() == ^0 and &1.bar() == ^1 and &0.baz() == ^2"
       assert params == [{"foo", {0, :foo}}, {"bar", {1, :bar}}, {"baz", {0, :baz}}]
     end
+
+    test "with ... binding in the middle" do
```
josevalim: Those tests are beautiful! I would recommend moving them to `query_test.exs` as there is nothing specific to fragments about them. Have them work directly on the query and then check the result of `query.joins` (or `query.aliases` according to my proposal).

test/ecto/query_test.exs
josevalim: It is also very important to test what happens when we give an alias that already exists. We also need to check what happens when you try to use an alias that does not exist.

zoldar:
@josevalim Thank you for the feedback. All the points you made are valid to me and make sense. If nobody else has any objections, I'd start with stage one, as described by you, some time next week.

zoldar:
I need a word of advice. When I've started working on introducing keyword opts in join/5, I've hit an issue with the fact that join actually allows to accept keyword list for `on` expression here:

https://github.com/zoldar/ecto/blob/cea93504d0dd609ad92733692e5bdd6830b685a9/test/ecto/query/builder/join_test.exs#L23

In order for this to work, I guess it would be necessary to tell apart the case where it's meant to be treated as `on` expression from the one where it's to be treated as a keyword list of options. I'm not sure whether there's a sensible solution for that. The `on` expression could very well to refer to fields such as `on` or `as`. Any ideas?

josevalim:
@zoldar yeah, you will have to check that it is either an empty list or it has `as` or `on` in it. There is a very small chance of a conflict but this is a major release so we can afford it.

Theoretically we could just go and break everything but that would be annoying for users. :)

zoldar:
Closing in favor of plan laid out in https://github.com/elixir-ecto/ecto/pull/2411#issuecomment-364642910 and #2428. Leaving for reference though.

PR #2416, opened 2018-02-11
Feature/respect schema prefix in joins

### LINKED ISSUE #2369: Problem to join with different postgresql schemas
### Environment

* Elixir version: Elixir 1.4.4  ( Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] )
* Database and version: PostgreSQL 9.5.10
* Ecto version: ecto 2.1.4
* Database adapter and version: postgrex 0.13.2
* Operating system: Linux 3e30dded55e2 4.9.0-5-amd64 #1 SMP Debian 4.9.65-3+deb9u2

### Current behavior

We are getting problem to make join with tables in different postgresql schemas

With a simple join with two tables 

revenue_metrics = (from r in Ppa.RevenueMetric,
      join: university in Ppa.University, on: university.id == r.university_id,

in the revenue_metrics.ex model definition there is a 

  @schema_prefix "ppa"

and in the university.ex model definition there is 

  @schema_prefix "qb"

>>> the sql generated looks like
SELECT ... FROM "ppa"."revenue_metrics" AS r0 INNER JOIN "ppa"."universities"

### Expected behavior

>>> the sql generated looks like
SELECT ... FROM "ppa"."revenue_metrics" AS r0 INNER JOIN "qb"."universities"

isaacsanders:
Currently a WIP

Looking for help writing tests for multi-prefix queries.

When finished, fixes #2369

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -71,7 +71,7 @@ if Code.ensure_loaded?(Mariaex) do
     ## Query
 
     alias Ecto.Query
-    alias Ecto.Query.{BooleanExpr, JoinExpr, QueryExpr}
+    alias Ecto.Query.{BooleanExpr, FromExpr, JoinExpr, QueryExpr}
 
     def all(query) do
       sources = create_names(query)
@@ -504,21 +504,28 @@ if Code.ensure_loaded?(Mariaex) do
 
     defp create_names(prefix, sources, pos, limit) when pos < limit do
       current =
-        case elem(sources, pos) do
-          {table, schema} ->
-            name = [create_alias(table) | Integer.to_string(pos)]
-            {quote_table(prefix, table), name, schema}
-          {:fragment, _, _} ->
-            {nil, [?f | Integer.to_string(pos)], nil}
-          %Ecto.SubQuery{} ->
-            {nil, [?s | Integer.to_string(pos)], nil}
-        end
+        sources
+        |> elem(pos)
+        |> create_name(prefix, pos)
       [current | create_names(prefix, sources, pos + 1, limit)]
     end
 
     defp create_names(_prefix, _sources, pos, pos) do
       []
     end
+
+    defp create_name({table, schema}, prefix, pos),
+      do: {quote_table(prefix, table),
+           [create_alias(table) | Integer.to_string(pos)],
+           schema}
+    defp create_name({:fragment, _, _}, _prefix, pos),
+      do: {nil, [?f | Integer.to_string(pos)], nil}
+    defp create_name(%Ecto.SubQuery{}, _prefix, pos),
+      do: {nil, [?s | Integer.to_string(pos)], nil}
+    defp create_name(%FromExpr{} = from, prefix, pos),
+      do: create_name(from.source, from.prefix || prefix, pos)
+    defp create_name(%JoinExpr{} = join, prefix, pos),
+      do: create_name(join.source, join.prefix || prefix, pos)
 
     defp create_alias(<<first, _rest::binary>>) when first in ?a..?z when first in ?A..?Z do
       <<first>>
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -109,7 +109,7 @@ if Code.ensure_loaded?(Postgrex) do
     end
 
     alias Ecto.Query
-    alias Ecto.Query.{BooleanExpr, JoinExpr, QueryExpr}
+    alias Ecto.Query.{BooleanExpr, FromExpr, JoinExpr, QueryExpr}
 
     def all(query) do
       sources = create_names(query)
@@ -568,21 +568,28 @@ if Code.ensure_loaded?(Postgrex) do
 
     defp create_names(prefix, sources, pos, limit) when pos < limit do
       current =
-        case elem(sources, pos) do
-          {table, schema} ->
-            name = [create_alias(table) | Integer.to_string(pos)]
-            {quote_table(prefix, table), name, schema}
-          {:fragment, _, _} ->
-            {nil, [?f | Integer.to_string(pos)], nil}
-          %Ecto.SubQuery{} ->
-            {nil, [?s | Integer.to_string(pos)], nil}
-        end
+        sources
+        |> elem(pos)
+        |> create_name(prefix, pos)
       [current | create_names(prefix, sources, pos + 1, limit)]
     end
 
     defp create_names(_prefix, _sources, pos, pos) do
       []
     end
+
+    defp create_name({table, schema}, prefix, pos),
+      do: {quote_table(prefix, table),
+           [create_alias(table) | Integer.to_string(pos)],
+           schema}
+    defp create_name({:fragment, _, _}, _prefix, pos),
+      do: {nil, [?f | Integer.to_string(pos)], nil}
+    defp create_name(%Ecto.SubQuery{}, _prefix, pos),
+      do: {nil, [?s | Integer.to_string(pos)], nil}
+    defp create_name(%FromExpr{} = from, prefix, pos),
+      do: create_name(from.source, from.prefix || prefix, pos)
+    defp create_name(%JoinExpr{} = join, prefix, pos),
+      do: create_name(join.source, join.prefix || prefix, pos)
 
     defp create_alias(<<first, _rest::binary>>) when first in ?a..?z when first in ?A..?Z do
       <<first>>
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -33,7 +33,7 @@ defmodule Ecto.Association do
                field: atom,
                unique: boolean}
 
-  alias Ecto.Query.{BooleanExpr, QueryExpr}
+  alias Ecto.Query.{BooleanExpr, FromExpr, QueryExpr}
 
   @doc """
   Builds the association struct.
@@ -169,7 +169,7 @@ defmodule Ecto.Association do
   end
 
   def assoc_query(refl, t, query, values) do
-    query = query || %Ecto.Query{from: {"join expression", nil}, prefix: refl.queryable.__schema__(:prefix)}
+    query = query || %Ecto.Query{from: %FromExpr{source: {"join expression", nil}, prefix: refl.queryable.__schema__(:prefix)}}
 
     # Find the position for upcoming joins
     position = length(query.joins) + 1
@@ -210,7 +210,8 @@ defmodule Ecto.Association do
     |> Map.put(:op, :and)
   end
 
-  defp merge_from({"join expression", _}, assoc_source), do: assoc_source
+  defp merge_from(%FromExpr{source: {"join expression", _}} = from, assoc_source),
+    do: %{ from | source: assoc_source }
   defp merge_from(from, _assoc_source), do: from
 
   # Rewrite all later joins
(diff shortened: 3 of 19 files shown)


lib/ecto/adapters/mysql/connection.ex
```diff
@@ -504,22 +504,29 @@ if Code.ensure_loaded?(Mariaex) do
...
+      do: {quote_table(prefix, table),
+           [create_alias(table) | Integer.to_string(pos)],
+           schema}
+    defp create_name({:fragment, _, _}, _prefix, pos),
+      do: {nil, [?f | Integer.to_string(pos)], nil}
+    defp create_name(%Ecto.SubQuery{}, _prefix, pos),
+      do: {nil, [?s | Integer.to_string(pos)], nil}
+    defp create_name(%FromExpr{} = from, prefix, pos),
+      do: create_name(from.source, from.prefix || prefix, pos)
+    defp create_name(%JoinExpr{} = join, prefix, pos),
```
josevalim: Minor: this and the line above should pattern match on `source: source, prefix: prefix`. :)

lib/ecto/query/inspect.ex
```diff
@@ -82,6 +82,13 @@ defimpl Inspect, for: Ecto.Query do
   defp unbound_from(%Ecto.Query{} = query) do
     "^" <> inspect(query)
   end
+  defp unbound_from(%Ecto.Query.FromExpr{} = from) do
+    if is_nil(from.prefix) do
+      unbound_from(from.source)
+    else
+      "FromExpr(\"#{from.prefix}\", #{unbound_from(from.source)})"
```
josevalim: TODO: we should emit valid queries as much as possible. So we need to figure out how we are going to support from prefixes in the keyword query.

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -504,22 +504,29 @@ if Code.ensure_loaded?(Mariaex) do
...
+      do: create_name(join.source, join.prefix || prefix, pos)
```
josevalim: I don't like that we are calling `create_name` recursively. This means if that if we accidentally put a `FromExpr` inside another FromExpr or if we forget to wrap a `from` inside a `FromExpr` at some point, it will go unnoticed.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -568,22 +568,29 @@ if Code.ensure_loaded?(Postgrex) do
...
+           [create_alias(table) | Integer.to_string(pos)],
+           schema}
+    defp create_name({:fragment, _, _}, _prefix, pos),
+      do: {nil, [?f | Integer.to_string(pos)], nil}
+    defp create_name(%Ecto.SubQuery{}, _prefix, pos),
+      do: {nil, [?s | Integer.to_string(pos)], nil}
+    defp create_name(%FromExpr{} = from, prefix, pos),
+      do: create_name(from.source, from.prefix || prefix, pos)
+    defp create_name(%JoinExpr{} = join, prefix, pos),
+      do: create_name(join.source, join.prefix || prefix, pos)
```
josevalim: Same comments as above.

lib/ecto/association.ex
```diff
@@ -169,7 +169,7 @@ defmodule Ecto.Association do
   end
 
   def assoc_query(refl, t, query, values) do
-    query = query || %Ecto.Query{from: {"join expression", nil}, prefix: refl.queryable.__schema__(:prefix)}
+    query = query || %Ecto.Query{from: %FromExpr{source: {"join expression", nil}, prefix: refl.queryable.__schema__(:prefix)}}
```
josevalim: We need to discuss this. Should we set the prefix in the whole query or only in from? Maybe we should only have the query prefix and the prefix for every join but not a prefix for the `from`?

lib/ecto/query/inspect.ex
```diff
@@ -239,6 +246,7 @@ defimpl Inspect, for: Ecto.Query do
   end
 
   defp from_sources(%Ecto.SubQuery{query: query}), do: from_sources(query.from)
+  defp from_sources(%Ecto.Query.FromExpr{source: source}), do: from_sources(source)
```
josevalim: I don't like that we are calling `from_sources` recursively. This means if that if we accidentally put a `FromExpr` inside another FromExpr or if we forget to wrap a `from` inside a `FromExpr` at some point, it will go unnoticed.

lib/ecto/query/planner.ex
```diff
@@ -236,13 +236,29 @@ defmodule Ecto.Query.Planner do
...
+  defp prepare_prefix(query, {_source, schema}) when is_atom(schema) and schema != nil,
+    do: schema.__schema__(:prefix) || query.prefix
+  defp prepare_prefix(query, {:fragment, _, _}),
+    do: query.prefix
+  defp prepare_prefix(query, %FromExpr{} = from),
+    do: from.prefix || query.prefix
+  defp prepare_prefix(query, %JoinExpr{} = join),
+    do: join.prefix || query.prefix
+  defp prepare_prefix(query, %Ecto.SubQuery{} = subquery),
+    do: subquery.query.prefix || query.prefix
```
josevalim: I don't like that we are calling `prepare_prefix` recursively. This means if that if we accidentally put a `FromExpr` inside another FromExpr or if we forget to wrap a `from` inside a `FromExpr` at some point, it will go unnoticed.

Also, we don't need to fallback to `|| query.prefix` here as we ask the adapters to do that later on. Unless we want to consistently do so here... in this case, we should remove the `|| query.prefix` from the adapters. I assume normalizing it here is better as it requires less logic in the adapters?

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -504,22 +504,29 @@ if Code.ensure_loaded?(Mariaex) do
...
-            name = [create_alias(table) | Integer.to_string(pos)]
-            {quote_table(prefix, table), name, schema}
-          {:fragment, _, _} ->
-            {nil, [?f | Integer.to_string(pos)], nil}
-          %Ecto.SubQuery{} ->
-            {nil, [?s | Integer.to_string(pos)], nil}
-        end
+        sources
+        |> elem(pos)
+        |> create_name(prefix, pos)
```
josevalim: It seems that, when preparing the queries, we already do `|| query.prefix`. This means we don't need to pass the `query.prefix` here. Shall we remove it from here then?

lib/ecto/query/planner.ex
josevalim: Same for `prepare_source` above.

lib/ecto/query/planner.ex
```diff
@@ -569,6 +590,8 @@ defmodule Ecto.Query.Planner do
   defp prepend_if(cache, true, prepend), do: prepend ++ cache
   defp prepend_if(cache, false, _prepend), do: cache
 
+  defp source_cache(%FromExpr{source: source}, params),
+    do: source_cache(source, params)
```
josevalim: We are calling it recursively too.

lib/ecto/query/planner.ex
```diff
@@ -1114,7 +1141,14 @@ defmodule Ecto.Query.Planner do
   end
 
   defp get_source!(where, %{sources: sources} = query, ix) do
-    elem(sources, ix)
+    case elem(sources, ix) do
+      %FromExpr{source: source} ->
+        source
+      %JoinExpr{source: source} ->
+        source
+      source ->
```
josevalim: Could it ever be this line? When?

lib/ecto/query/planner.ex
```diff
@@ -1162,6 +1196,8 @@ defmodule Ecto.Query.Planner do
     case get_source!(kind, query, ix) do
       {_, schema} ->
```
josevalim: Could it ever be this line now?

lib/ecto/association.ex
josevalim: Actually, I think we should completely ignore the prefix here. Instead, we should change `merge_from` to called as `merge_from(from, assoc)` and then copy both `prefix` and `source` from assoc into from.

josevalim:
Thank you @isaacsanders! This looks beautiful.

My only concern is that the code is less assertive now. It seems in a bunch of places we are receiving both `FromExpr` and `JoinExpr` but also the "old" tuple/subquery/fragment formats. It feels like they should either be `FromExpr` and `JoinExpr` or the "old" ones. The fact we call some of those functions recursively add to the confusion.

Regarding the tests, we can write them at test/ecto/adapters/postgres_test.exs and test/ecto/adapters/mysql_test.exs. We basically need to test what happens when:

  1. from has a prefix and join doesn't
  2. from doesn't have a prefix and join does
  3. from has a prefix and join has another
  4. all of the cases above again but with the query having a prefix too

lib/ecto/adapters/postgres/connection.ex
isaacsanders: With the `JoinExpr`s, what is the desired behavior with respect to `FromExpr`s in this case? They aren't like direct sources, nor are they like `SubQuery`s...

Unless I did something incorrect.

josevalim: The root level will always be a FromExpr or a JoinExpr. Their source will then always be a fragment, a tuple or a subquery.

isaacsanders:
The lack of assertion is because I don't really always feel comfortable making an assertion. I haven't read all of your comments yet, but if you have an idea were the assertions are too weak, I can try to work on them more.

lib/ecto/adapters/mysql/connection.ex
isaacsanders: I'm okay with that.

lib/ecto/adapters/postgres/connection.ex
isaacsanders: But here, there can be a `%JoinExpr{source: %FromExpr{}}`. How deep should this go?

isaacsanders: I would rather get the source from the from before it is put into the join and just put that source in the join at that point, but I don't understand the codebase well enough to know where that change should occur.

lib/ecto/query/planner.ex
isaacsanders: `prepare_prefix` isn't being called recursively.

lib/ecto/adapters/postgres/connection.ex
josevalim: @isaacsanders oh, definitely. A FromExpr should *never* be inside a JoinExpr. My suggestion: change at least one of those functions so that they are no longer recursive and it will reveal you which kind of query this is happening, then you can go and fix the code. :)

With the assertive code the code will fail, then we "just" need to fix it. :)

josevalim:
Someone referenced this PR by mistake. Reopening. :)

lib/ecto/query/inspect.ex
isaacsanders: `"prefix"."table"` is how Postgres has this feature specified. Do you know how other systems make this syntax?

josevalim: This is inspecting in the Elixir syntax. So we need to reproduce the same query syntax as the developer would write when building the query.

lib/ecto/query/planner.ex
josevalim: You are correct, only prepare_source above.

josevalim:
Hi @isaacsanders! Since we hvaen't had progress here in almost 2 months, I will go ahead and close this. Also note we pushed FromExpr to master and this should make your life easier. Feel free to reopen or submit a new PR at any time, thank you.

PR #2417, opened 2018-02-12
Associations Guide proofread

radar:
Read through the 👌 excellent 👌 associations guide today to update my knowledge on how to work with the different association types. I noticed a couple of minor inconsistencies in the guide, and so I've fixed them up in this PR.

I've also added the `examples/ecto_assoc` directory. It contains the code from the guide, which I think might be helpful. I did this because the "Getting Started" does the same thing with its `examples/friends` directory.

Thanks :heart:


josevalim:

PR #2425, opened 2018-02-15
updated Readme to include the Tds MSSQL adapter again

ewitchin:
We have added Ecto2 support into Tds and Tds.Ecto months ago. Adding it back on the list as available adapters since it is now up to date and being maintained


michalmuskala:
Thank you! :heart:

That's a great news. Congratulations! 🎉

PR #2426, opened 2018-02-15
Update docs for mix ecto.migrate task and custom priv path.

davepersing:
Hello!  Thank you so much for the work on Ecto.  It's a wonderful library!

This is a result of debugging a custom `priv` path with my repo.  I didn't realize `migrations` was getting appended to the specified path.  I'm happy to put a PR up to fix the `migrations_path` function if you'd prefer, but in the meantime, I wanted to update the documentation so no one else gets confused.  

Thanks again!


josevalim:

PR #2428, opened 2018-02-16
Support passing `on` to join/5 as option and soft-deprecate raw argument 

zoldar:
Refs #2389 
Refs #2411 

This PR is attempt at stage one of plan to implement named joins in Ecto, as laid out by @josevalim in https://github.com/elixir-ecto/ecto/pull/2411#issuecomment-364642910.

The actual changes to the logic are implemented in the [first commit](https://github.com/elixir-ecto/ecto/commit/96d4c64b2fc544769c6395d88cb6082db076e41a). The [second commit](https://github.com/elixir-ecto/ecto/commit/ac84d0a2f9f99db71045c31679e07bd98627970f) contains exclusively updates to the new syntax across tests and docs.

Feedback appreciated.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -776,17 +776,17 @@ defmodule Ecto.Query do
   @join_opts [:on]
 
   defp parse_join_opts([]), do: {:opts, []}
-  defp parse_join_opts(lst) when is_list(lst) do
-    opts = Keyword.take(lst, @join_opts)
+  defp parse_join_opts(list) when is_list(list) do
+    opts = Keyword.take(list, @join_opts)
 
-    case {lst, opts} do
-      {lst, []} ->
-        {:expr, lst}
+    case {list, opts} do
+      {list, []} ->
+        {:expr, list}
       {opts, opts} ->
         {:opts, opts}
-      {lst, _} ->
+      {list, _} ->
         raise ArgumentError, "invalid option(s) passed to Ecto.Query.join/5: " <>
-          "#{inspect(Keyword.keys(lst) -- @join_opts)}. Valid options: #{inspect @join_opts}."
+          "#{inspect(Keyword.keys(list) -- @join_opts)}. Valid options: #{inspect @join_opts}."
     end
   end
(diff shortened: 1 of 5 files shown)


lib/ecto/query.ex
```diff
@@ -757,12 +757,40 @@ defmodule Ecto.Query do
...
     query
-    |> Join.build(qual, binding, expr, on, nil, __CALLER__)
+    |> Join.build(qual, binding, expr, opts[:on], nil, __CALLER__)
     |> elem(0)
   end
 
+  @join_opts [:on]
+
+  defp parse_join_opts([]), do: {:opts, []}
+  defp parse_join_opts(lst) when is_list(lst) do
```
josevalim: This is beautiful. My only comment is to not abbreviate variable names. So `list` instead of `lst`. :)

lib/ecto/query.ex
```diff
@@ -757,12 +757,40 @@ defmodule Ecto.Query do
...
+          opts
+        {:expr, on_expr} ->
+          IO.warn "Passing raw `on` expression as the last argument to Ecto.Query.join/5 is deprecated. " <>
+            "Please use :on keyword option instead."
+          [on: on_expr]
+      end
+
     query
-    |> Join.build(qual, binding, expr, on, nil, __CALLER__)
+    |> Join.build(qual, binding, expr, opts[:on], nil, __CALLER__)
```
josevalim: If you want, you can already start passing a list of options to Join.build. But you can also do it later too. Completely up to you. :)

josevalim:
Just two comments and we should be good to go.

josevalim:

lib/ecto/query.ex
zoldar: I was about to push the second change :D Oh well, it'll just become part of the next PR.

fertapric:
Nice PR @zoldar! One minor thing, should this be included in the `CHANGELOG.md`?

josevalim:
PRs should never touch the CHANGELOG because they are very conflict prone. :)

PR #2430, opened 2018-02-18
Use "expr1 not in expr2" instead of "not expr1 in expr2"

ppraisethesun:


josevalim:
This requires Elixir v1.5 and we still support v1.4, so we need to wait a bit more, thanks. :)

PR #2431, opened 2018-02-19
[WIP] Add support for named joins

zoldar:
Refs #2389 

This PR implements support for named joins following the feedback from #2411 and building on top of what was implemented in #2428.

The commits are organized in a logical order with atomic changes, so that the work can be reviewed one-by-one (which may come in handy due to size of the PR).

1. https://github.com/elixir-ecto/ecto/commit/e3796abda9bc1f36d06837faa007f9e4fe44d06b alters `Ecto.Query.Builder.Join.build` to accept opts as the last argument. This change landed here because it didn't make it to #2428 
2. https://github.com/elixir-ecto/ecto/commit/443f3b81cded190d3c67bae85760cf3837c5e67c adds `:as` option support and assures that it's passed from both `Ecto.Query.join` and `Ecto.Query.from`. There's a small inconsistency in how the extra parameters are treated. For `Ecto.Query.from`, `as` has to be always present immediately after `join` DSL construct, while in case of `Ecto.Query.join`, there are no constraints with regards to the position of `as` relative to `on`. While at it, `Ecto.Query.join` does not support multiple `on` occurrences the way `Ecto.Query.from` does. Is it desired to strive for 100% consistency in behavior between those 2?
3. https://github.com/elixir-ecto/ecto/commit/bdff7e78a7c327d4d6e65ec07eec55beac426724 handles storing join name to position mapping on the basis of `as` alias passed when building. It follows the advice of storing those mappings explicitly in top-level query struct for quick access later. I'm not 100% sure whether the approach taken is correct one, however following the structure of the code, I've decided to build/update aliases map in the `apply` step, because that seemed the only place where query is modified in any way (a join expression is added)
4. https://github.com/elixir-ecto/ecto/commit/0c075f74b1fdad17b19d8f0c26c9afd6532cf85a handles named binds in binding lists. It's mostly iterates on the initial attempt made in #2428 with a couple refactorings along the way to make that area of the code a bit clearer (I hope).
5. https://github.com/elixir-ecto/ecto/commit/11c2f8ff5926f6a8f83793f5ad5d47d0e89ffb96 adds an extra integration test so that at least basic use case is tested end to end.

As always, feedback appreciated.

## TODO (there may be more, based on the feedback)

- [ ] add documentation
- [ ] add extra tests around passing opts to `Ecto.Query.from`, especially for failing cases

### DIFF
--- a/integration_test/cases/joins.exs
+++ b/integration_test/cases/joins.exs
@@ -77,6 +77,18 @@ defmodule Ecto.Integration.JoinsTest do
     assert [{^p1, ^c1}, {^p2, ^c1}] = TestRepo.all(query)
   end
 
+  test "named joins" do
+    _p = TestRepo.insert!(%Post{title: "1"})
+    p2 = TestRepo.insert!(%Post{title: "2"})
+    c1 = TestRepo.insert!(%Permalink{url: "1", post_id: p2.id})
+
+    query =
+      from(p in Post, join: c in assoc(p, :permalink), as: :permalink, order_by: p.id)
+      |> select([p, permalink: c], {p, c})
+
+    assert [{^p2, ^c1}] = TestRepo.all(query) 
+  end
+
   @tag :left_join
   test "left joins with missing entries" do
     p1 = TestRepo.insert!(%Post{title: "1"})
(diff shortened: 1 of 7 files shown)


josevalim:
This is beautiful. I will go ahead and merge this and push some commits on top with small changes. Documentation and more tests can come as we improve this. :)

josevalim:

josevalim:
@zoldar I have pushed two commits that tidies up the implementation a bit. I have also updated #2389 to include pending tasks.

PR #2432, opened 2018-02-19
provide more insight into how on_replace affects action

atomkirk:
This is what I've deduced for myself. Is this correct?

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -134,12 +134,14 @@ defmodule Ecto.Changeset do
       changeset, and it will be marked as invalid
     * `:nilify` - sets owner reference column to `nil` (available only for
       associations). Use this on a `belongs_to` column to allow the association
-      to be cleared out so that it can be set to a new value.
+      to be cleared out so that it can be set to a new value. Will set `action`
+      on associated changesets to `:replace`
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -134,12 +134,14 @@ defmodule Ecto.Changeset do
...
-      to be cleared out so that it can be set to a new value.
+      to be cleared out so that it can be set to a new value. Will set `action`
+      on associated changesets to `:replace`
     * `:update` - updates the association, available only for `has_one` and `belongs_to`.
       This option will update all the fields given to the changeset including the id
       for the association
     * `:delete` - removes the association or related data from the database.
-      This option has to be used carefully
+      This option has to be used carefully. Will set `action` on associated changesets
+      to `:delete`
```
atomkirk: @josevalim I think I may have been mistaken. After some more testing I found changesets are still set to `:replace` even though `on_replace: :delete` is on the schema. Does that sound right? Seems like it *should* be set to `:delete`, that would make more sense to me. So is that a bug in ecto or should I create a new PR to fix my doc change?

josevalim: Good call. Actually we always mark them as a `:replace`, regardless of the behaviour, and we apply the nillify/update/delete later on. Can you please send a new PR to update the docs? :heart:

PR #2433, opened 2018-02-19
fix bad assumption

atomkirk:
@josevalim


josevalim:

PR #2434, opened 2018-02-20
Minor doc fixup: Update field name to reflect doc example code.

asummers:
The rest of the code in these docs refer to an `Item`'s `:title` but the field name was `:name` in the `schema`. Alternative is to change uses of `:title` to `:name` but this seems less invasive.


josevalim:

PR #2435, opened 2018-02-20
Do not create double newline when updating config/config.exs

lexmag:


josevalim:

Issue #2436, opened 2018-02-21
BadMapError in select_merge/2


AndrewDryga:
### Environment

* Elixir version (elixir -v): Elixir 1.7.0-dev (252974e) (compiled with OTP 20)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6
* Ecto version (mix deps): 2.2.8
* Database adapter and version (mix deps): 0.13.5
* Operating system: macOS

### Current behavior

After updating from 2.2.7 we observe a regression bug for a query that uses `select_merge/2`:

```
** (BadMapError) expected a map, got: nil
 code: assert [%Draft{id: ^draft1_id}, %Draft{id: ^draft2_id}] = Loader.all_for_profile_id(owner.profile.id)
 stacktrace:
   (ecto) lib/ecto/repo/queryable.ex:276: anonymous fn/2 in Ecto.Repo.Queryable.process_update/6
   (elixir) lib/enum.ex:1905: Enum."-reduce/3-lists^foldl/2-0-"/3
   (ecto) lib/ecto/repo/queryable.ex:274: Ecto.Repo.Queryable.process_update/6
   (ecto) lib/ecto/repo/queryable.ex:194: Ecto.Repo.Queryable.process/5
   (ecto) lib/ecto/repo/queryable.ex:174: Ecto.Repo.Queryable.preprocess/5
   (postgrex) lib/postgrex/query.ex:77: DBConnection.Query.Postgrex.Query.decode_map/3
   (postgrex) lib/postgrex/query.ex:64: DBConnection.Query.Postgrex.Query.decode/3
   (db_connection) lib/db_connection.ex:1019: DBConnection.decode/6
   (ecto) lib/ecto/adapters/postgres/connection.ex:73: Ecto.Adapters.Postgres.Connection.prepare_execute/5
   (ecto) lib/ecto/adapters/sql.ex:256: Ecto.Adapters.SQL.sql_call/6
   (ecto) lib/ecto/adapters/sql.ex:426: Ecto.Adapters.SQL.execute_and_cache/7
   (ecto) lib/ecto/repo/queryable.ex:133: Ecto.Repo.Queryable.execute/5
   (ecto) lib/ecto/repo/queryable.ex:37: Ecto.Repo.Queryable.all/4
   (ecto) lib/ecto/repo/queryable.ex:78: Ecto.Repo.Queryable.one!/4
[...]

PR #2437, opened 2018-02-21
Fix select_merge/2 regression

AndrewDryga:
Related to #2436

### DIFF
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -235,6 +235,9 @@ defmodule Ecto.Repo.Queryable do
         {map, row}
     end
   end
+  defp process(row, {:map, {:source, :from}, args}, nil, _prefix, _adapter) do
+    {Map.new(args), row}
+  end
   defp process(row, {:map, data, args}, from, prefix, adapter) do
     {data, row} = process(row, data, from, prefix, adapter)
     process_update(data, args, row, from, prefix, adapter)
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1120,6 +1120,13 @@ defmodule Ecto.Integration.RepoTest do
       assert [%{title: "2"}] =
              Post |> select([p], %{title: p.title}) |> select_merge([p], %{title: "2"}) |> TestRepo.all()
     end
+
+    test "merge with update" do
+      %Post{} = TestRepo.insert!(%Post{title: "1", counter: 1})
+
+      assert [%Post{title: "1", counter: 2}] =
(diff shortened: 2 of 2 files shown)


AndrewDryga:
After merging, I can submit the same PR to the branch `v2.2` so we can release an update.

lib/ecto/repo/queryable.ex
```diff
@@ -235,6 +235,9 @@ defmodule Ecto.Repo.Queryable do
         {map, row}
     end
   end
+  defp process(row, {:map, {:source, :from}, args}, nil, _prefix, _adapter) do
+    {Map.new(args), row}
```
josevalim: Should the recursive [call here](https://github.com/AndrewDryga/ecto/blob/3eb6b5d4e7f4b4a72e29b35bd88cf1ef08620674/lib/ecto/repo/queryable.ex#L242) handle this case?

AndrewDryga: It did not work, because from is nil. And that call is basically calling [this code](https://github.com/AndrewDryga/ecto/blob/3eb6b5d4e7f4b4a72e29b35bd88cf1ef08620674/lib/ecto/repo/queryable.ex#L256-L258):
```
  defp process(row, {:source, :from}, from, _prefix, _adapter) do
    {from, row}
  end
```

But here args knowledge is lost, so within this clause we can't really do anything.

josevalim: Ok, I think we need to generalize this a bit. I will merge and play with it. :) Thank you! :heart:

AndrewDryga: You're welcome. And in case you feel like guiding me won't take more time than fixing it by yourself, I can take care of it :). Just need to know a place to start.

josevalim: I would love to guide but I still have no idea what needs to be done or if something needs to be done. I want to understand why we are passing `nil` in the first place. :)

josevalim:
I have pushed 9374766059bd884bd5c8fb46eb77b85dd0e042f4 which generalizes `{:source, :from}`. It seems to be the proper commit here.

I would like to say that your merges could be clearer, there is no reason to merge on yourself. Do you have your actual production query? I believe it can be simplified a lot by passing pure maps on select_merge and not map updates. Basically:

    select_merge: %{foo: bar}

instead of

    select_merge: %{c | foo: bar}

AndrewDryga:
@josevalim I already saw the code 👍.

Regarding merge we got into a funny situation, we initially tried to write that query properly but it did not work on v2.2.7. But then you fixed the bug (elixir-ecto/ecto@d189cb8) that stopped us from using simple version AND broken our query that is ugly but works.

So it's like we weren't able to write proper code because of bug, and then we have a bug because ugly code stopped to work :).

PR #2438, opened 2018-02-21
Fix Ecto.Query.API.map for single nil column with join

### LINKED ISSUE #2380: Ecto.Query.API.map broken on single nil column with join
### Environment

* Elixir version: Elixir 1.7.0-dev (4175ddebd) (compiled with OTP 20)
* Database and version: psql (PostgreSQL) 9.6.5
* Ecto version (mix deps): master
* Database adapter and version (mix deps): postgrex 0.14.0-dev
* Operating system: macOS Sierra

### Current behavior

```elixir
# works on simple query
iex> TestRepo.one(from p in Post, where: p.title == "1", select: map(p, [:counter]))
%{counter: nil}

# doesn't work with join
iex> TestRepo.one(from p in Post, join: c in assoc(p, :comments), where: p.title == "1", select: map(p, [:counter]))
nil

# when selecting more columns it works again
iex> TestRepo.one(from p in Post, join: c in assoc(p, :comments), where: p.title == "1", select: map(p, [:title, :counter]))
%{counter: nil, title: "1"}
```

### Expected behavior

```
iex> TestRepo.one(from(p in Post, join: c in assoc(p, :comments), where: p.title() == "1", select: [:counter]))
%{counter: nil}
```

here's a test case!

```
diff --git a/integration_test/cases/repo.exs b/integration_test/cases/repo.exs
index aac73cf0..18d62030 100644
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1091,6 +1091,13 @@ defmodule Ecto.Integration.RepoTest do
              TestRepo.one(from p in Post, where: p.title == "1", select: [:counter])
     end
 
+    test "take with join assocs and single nil column" do
+      %{id: pid} = TestRepo.insert!(%Post{title: "1"}, counter: nil)
+      TestRepo.insert!(%Comment{post_id: pid, text: "comment"})
+      assert %{counter: nil} =
+             TestRepo.one(from p in Post, join: c in assoc(p, :comments), where: p.title == "1", select: [:counter])
+    end
+
     test "field source" do
       TestRepo.insert!(%Permalink{url: "url"})
       assert ["url"] = Permalink |> select([p], p.url) |> TestRepo.all()

```

```
  1) test query select take with join assocs and single nil column (Ecto.Integration.RepoTest)
     integration_test/cases/repo.exs:1094
     match (=) failed
     code:  assert %{counter: nil} = TestRepo.one(from(p in Post, join: c in assoc(p, :comments), where: p.title() == "1", select: [:counter]))
     right: nil
     stacktrace:
       integration_test/cases/repo.exs:1097: (test)
```

uesteibar:
Fixes #2380

Not sure if this is the best approach, but as far as I could test it solves the issue.

Used the test case provided by @wojtekmach in the issue as a regression test (thanks!)

### DIFF
--- a/integration_test/cases/joins.exs
+++ b/integration_test/cases/joins.exs
@@ -112,7 +112,7 @@ defmodule Ecto.Integration.JoinsTest do
 
     query = from(p in Post, right_join: c in assoc(p, :comments),
                  preload: :permalink, order_by: c.id)
-    assert [p1, p2, nil] = TestRepo.all(query)
+    assert [p1, p2, _] = TestRepo.all(query)
     assert p1.id == pid1
     assert p2.id == pid2
 
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1093,6 +1093,13 @@ defmodule Ecto.Integration.RepoTest do
              TestRepo.one(from p in Post, where: p.title == "1", select: [:counter])
     end
 
+    test "take with join assocs and single nil column" do
+      %{id: post_id} = TestRepo.insert!(%Post{title: "1"}, counter: nil)
+      TestRepo.insert!(%Comment{post_id: post_id, text: "comment"})
+      assert %{counter: nil} ==
+              TestRepo.one(from p in Post, join: c in assoc(p, :comments), where: p.title == "1", select:  map(p, [:counter]))
+    end
+
     test "field source" do
       TestRepo.insert!(%Permalink{url: "url"})
       assert ["url"] = Permalink |> select([p], p.url) |> TestRepo.all()
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -184,11 +184,9 @@ defmodule Ecto.Repo.Queryable do
     end
   end
 
-  defp preprocessor({_, {:source, source_schema, fields}}, preprocess, sources, prefix, adapter) do
-    all_nil? = tuple_size(sources) != 1
-
+  defp preprocessor({_, {:source, source_schema, fields}}, preprocess, _sources, prefix, adapter) do
     fn row ->
-      {entry, rest} = process_source(source_schema, fields, row, all_nil?, prefix, adapter)
+      {entry, rest} = process_source(source_schema, fields, row, false, prefix, adapter)
       preprocess(rest, preprocess, entry, prefix, adapter)
     end
   end


josevalim:
Sorry @uesteibar. I have just changed this part of the code.

I think a simpler fix would be to completely ignore the `all_nil?` value here and always pass it as false:

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/queryable.ex#L188

Could you please try that out?

josevalim:
Btw, I am aware that my solution will break a right join test but I think we should fix that test instead. The issue is that it is impossible for us to know, in the case of right joins, if all of the fields are nil or if the record entry is missing. But I would rather have the faulty behaviour on the right join rather than on `from`.

uesteibar:
@josevalim that's the first thing I tried, but found out that the test for the right join so tried to make it work without braking it. I'll give that a try :)

uesteibar:
With ignoring `all_nil?` and just passing false, it will still use the struct as result, which I understand is not the behaviour we want (I changed this behaviour with [this](https://github.com/elixir-ecto/ecto/pull/2438/files#diff-ed753f133e94caf75aab64df9433dbb8R283)).

```
code:  assert %{counter: nil} == TestRepo.one(from(p in Post, join: c in assoc(p, :comments), where: p.title() == "1", select: [:counter]))
     left:  %{counter: nil}
     right: %Ecto.Integration.Post{
              __meta__: #Ecto.Schema.Metadata<:loaded, "posts">,
              author: #Ecto.Association.NotLoaded<association :author is not loaded>,
              author_id: nil,
              bid: nil,
              comments: #Ecto.Association.NotLoaded<association :comments is not loaded>,
              comments_authors: #Ecto.Association.NotLoaded<association :comments_authors is not loaded>,
              comments_authors_permalinks: #Ecto.Association.NotLoaded<association :comments_authors_permalinks is not loaded>,
              constraint_users: #Ecto.Association.NotLoaded<association :constraint_users is not loaded>,
              cost: nil,
              counter: nil,
              id: nil,
              inserted_at: nil,
              intensity: nil,
              links: nil,
              meta: nil,
              permalink: #Ecto.Association.NotLoaded<association :permalink is not loaded>,
              post_user_composite_pk: #Ecto.Association.NotLoaded<association :post_user_composite_pk is not loaded>,
              posted: nil,
              public: true,
              temp: "temp",
              text: nil,
              title: nil,
              unique_users: #Ecto.Association.NotLoaded<association :unique_users is not loaded>,
              update_permalink: #Ecto.Association.NotLoaded<association :update_permalink is not loaded>,
              updated_at: nil,
              users: #Ecto.Association.NotLoaded<association :users is not loaded>,
              users_comments: #Ecto.Association.NotLoaded<association :users_comments is not loaded>,
              uuid: nil,
              visits: nil
            }
```

I don't see a way of knowing whether the user passed all the columns in the `from` or none without passing `take` all the way down here, but I might be missing something.

Any thoughts?

josevalim:
> With ignoring all_nil? and just passing false, it will still use the struct as result

The test is wrong. When we do `select: [:counter]`, it will still keep the struct format. You can change it to `select: map(p, [:counter])` though.

uesteibar:
@josevalim redone following your proposal

josevalim:

PR #2439, opened 2018-02-22
Fix crash ecto.create/drop on log: false for mysql adapter

utah-KT:
Fixed crash on ecto.create/drop with mysql.

I referred to https://github.com/elixir-ecto/ecto/commit/4d4f1cc5e586bbeb80e0c1b0eb2a3e1a8a1fa7f4 for this pull request.


josevalim:

PR #2440, opened 2018-02-22
Simple integer to string

devvit:
Issue [here](https://stackoverflow.com/questions/40613577/why-does-ectos-cast-not-convert-an-integer-to-a-string)

The same case for me too, I use [paper_trail](https://github.com/izelnakri/paper_trail) for tracing records, the type of ```item_id``` is one of ```string```,

```uuid```, ```integer```, and fetch them through ecto reflections, save as ```text``` type.

```uuid``` to ```string``` is automatic well done, but ```integer``` to ```string``` blocked me.

The PR  helps  simplify similar works.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -302,6 +302,10 @@ defmodule Ecto.Type do
     {:ok, value}
   end
 
+  def dump(:string, value, _dumper) when is_integer(value) do
+    {:ok, "#{value}"}
+  end
+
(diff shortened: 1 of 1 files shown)


fertapric:
This PR seems to introduce an unexpected coercion between types. Could a [custom `Ecto.Type`](https://hexdocs.pm/ecto/Ecto.Type.html#content) work in your case?

devvit:
closed, I will patch paper_trail.

PR #2443, opened 2018-02-23
Add Dynamic Repo Configuration

InAbsentia:
This is a dirty, barely-working implementation of dynamic repos (#1964). Now I could use some advice on whether I'm on the right track and where to go from here.

I've added a dynamic supervisor that will take a repo module, name, OTP app name (which could come from the repo, instead), and options as arguments to `start_child/4` to start a dynamic repo.

The dynamic repo then starts a named supervisor for the connection pool, which allows the registry and query cache to work without changes by calling the query functions with the repo name as the first argument. Only `DynamicRepo.all/3` is currently implemented.

Example:

```elixir
defmodule MyDynamicRepo do
  use Ecto.DynamicRepo, otp_app: :my_app, adapter: Ecto.Adapters.Postgres
end

Ecto.DynamicRepo.Supervisor.start_link() # or in MyApp.Application ;)

opts = [username: "user", hostname: "localhost", database: "my_database"]
Ecto.DynamicRepo.Supervisor.start_child(MyDynamicRepo, :my_repo_name, :my_app, opts)

opts = [username: "user", hostname: "localhost", database: "my_other_database"]
Ecto.DynamicRepo.Supervisor.start_child(MyDynamicRepo, :my_other_repo_name, :my_app, opts)

require Ecto.Query
query = Ecto.Query.from("my_table", select: [:id, :name])

MyDynamicRepo.all(:my_repo_name, query)
#> [%{id: 1, name: "one"}, %{id: 2, name: "two"}]

MyDynamicRepo.all(:my_other_repo_name, query)
#> [%{id: 3, name: "three"}, %{id: 4, name: "four"}]
```


michalmuskala:
I think this is going in the right direction. One thing, that I think is important and the thing that blocked us the most toactually implement this in the past is that we actually want to avoid naming the dynamic repo process - at least not using atoms and possibly through Registry - the best way would be to allow the user to specify the name similar to what you can do in a regular GenServer.

InAbsentia:
Thanks for the feedback!

Changing it to allow not providing a name seems relatively straightforward. The dynamic repo can use its module name, instead of a user-supplied name when none is given.

As far as not needing to use an atom for the name, the dynamic repo module is using the user-supplied name as the process name, so unless it causes a problem in the registry, I think users can use any name they can with a GenServer. I’ll test other naming methods to be sure, though. 🙂

Can you provide an example of a situation you expect it to handle that you think would be a problem with the current approach? Or is there somewhere other than the referenced issue where I can see some of your previous discussions on the topic?

josevalim:
Thank you @InAbsentia, this looks great!

As @michalmuskala said, we don't want to generate a module at all, but rather have a Ecto.DynamicRepo that is used directly and receives the name as argument.

That will complicate things in regards to the query cache, which also need to be started based on the `:name` and no longer based on the module name.

InAbsentia:
Ohhh, ok. Thanks for clarifying, @josevalim!

To be sure I understand what you want: `Ecto.DynamicRepo` should be a static module that behaves like a generated `Ecto.Repo`, but with the added ability to start new named repos and query them. Should `Ecto.DynamicRepo` also be the supervisor or would you prefer to have a separate dynamic supervisor module like I have now?

InAbsentia:
Also, I don't think the query cache will be a problem, unless there's some context I'm missing with more complicated situations. It works with the repo name with my current changes and `DynamicRepo.all` queries. The supervisor creates the ETS table when starting the repo and the query cache works normally, as far as I can tell.

josevalim:
> Also, I don't think the query cache will be a problem, unless there's some context I'm missing with more complicated situations. It works with the repo name with my current changes and DynamicRepo.all queries. 

Yes, the issue is that we will no longer have a repo name when we do dynamic repos, only process names. So we need to change the query cache to start based on the process name  and no longer on the repo. :)

InAbsentia:
👍🏼 By "repo name", I was referring to the user-provided name that I use to set the process name, so I've already done that and was planning to carry it over into the new version. I'll use "process name" from now on for clarity. 🙂

InAbsentia:
@josevalim @michalmuskala Are my latest changes more in line with what you're thinking?

If so, the main problem moving forward is that `Ecto.Adapters.SQL` relies on the generated functions `repo.__log__/1`, `repo.__adapter__/0`, and `repo.__sql__/0` when executing queries. My initial thought is to move those into the config that gets cached in the registry, since it's fetched for every query, anyway. Do you have a preference for how to handle that?

josevalim:
@InAbsentia in my mind we wouldn't have a separate module. We would start multiple repositories by calling `Ecto.Repo.start_link(name: Foo.Bar)` and so on and we would have a static module, not one that is generated, for the dispatch.

In any case, you are correct this is the main problem:

> If so, the main problem moving forward is that Ecto.Adapters.SQL relies on the generated functions 

I believe we always call the repository module for `__log__` and `__sql__`. The adapter is a trickier one. There are some operations that likely can never work on the named process, such as `mix ecto.create` and `mix ecto.drop`, as we don't even have a database running in those cases.

josevalim:
@InAbsentia maybe a good next step here is to change Ecto.Adapter callbacks and say some of them will receive a name (instead of a `Ecto.Repo.t`).

InAbsentia:
It's become clear to me that I don't actually understand how you want this feature to work. I assumed that I did, but my assumptions have proven to be faulty. 🙂 I've been interpreting your comments through those assumptions and making minor adjustments to my own mental model, instead of asking for further clarification.

When you have time, would you mind describing what this looks like to you at the user API level, so I can hopefully get back on track?

To start, a user calls `Ecto.Repo.start_link(name: Foo.Bar)`. What does the name `Foo.Bar` refer to in this context and what happens next?

josevalim:
Apologies for not specifying the problem properly @InAbsentia. Good call on breaking the miscommunication pattern.

Here is what I have in mind. Initially, the user needs to define a repository via the usual means:

```elixir
defmodule MyApp.Repo do
  use Ecto.Repo, otp_app: ...
end
```

Except that, if the user wants to start multiple instances, they can do so by calling `start_link`, passing different options and a different name to each instance:

```elixir
MyApp.Repo.start_link(name: Client1.Repo, database: ...)
```

That's already possible today although you need to pass the query_cache_owner option as false. Otherwise it will try to create a query cache entry with the same name. This is the first issue that needs to be solved. Before we solve this, there isn't much purpose in discussing the API.

How we are going to solve this varies a lot. Today, we need to know the repository name and the adapter. By adding the name and allowing the query cache to change per name, it means we also need to know the name of the process on every repo operation. It just happens that for the regular case, the repo name and the process name are the same.

There are other parts that would need changing. For example, today the adapter always receives the repository but now we know it is not enough, because the same repository may have multiple instances with different names and options. So all needs to be changed to work based on PIDs too.

The challenge is how to do those changes without causing major breaking changes to the adapter API.

InAbsentia:
Thanks for the detailed explanation @josevalim! This all makes a lot more sense to me now.

I'll start looking into updating the query cache and adapter and get back to you with any potential solutions or further questions.

InAbsentia:
Sorry I haven't had a chance to get back to this yet. What I have here currently fulfills my needs for the moment, so I copied it over and adapted it to work directly in my application. I still plan on working on the general solution, but I can't devote much of my work time to it right now thanks to the deadlines of having a startup as a client and I haven't had any more personal time to work on it, either. Hopefully I'll have some more time to dig into it in the next couple weeks.

Meanwhile, if anyone else wants to take this on, please feel free! I can come back and help out once things quiet down for me.

josevalim:
Hi @InAbsentia! I will go ahead and close this. If you pick it back up or someone else does it, please re-open it or send a new pull request. Thank you!

PR #2445, opened 2018-02-23
Include `:as` in inspected queries

zoldar:
Refs #2389 

I have (re)introduced `:as` attribute in `JoinExpr` struct for convenience of converting expressions on inspect. I'm not too happy with the way the `as` expression is put in the list in `Inspect.join` but couldn't think of a better way ATM that wouldn't hinder the readability even further.


josevalim:

Issue #2441, opened 2018-02-22
Ecto.LogEntryTest "converts entry to iodata" fails


Farijo:
### Environment

* Elixir version (elixir -v): Elixir 1.5.0 Erlang/OTP 19
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):
* Ecto version (mix deps): 3.0.0-dev
* Database adapter and version (mix deps):
* Operating system: Windows

### Current behavior

The test "converts entry to iodata" in Ecto.LogEntryTest fails whenever a time of the entry is set to a value other than 0. For exemple l.28: 

```elixir
    diff = :erlang.convert_time_unit(1, :micro_seconds, :native)
    entry = %{entry | params: [1, 2, 3], query_time: 2100 * diff, queue_time: 100 * diff}
    assert to_binary(entry) == "QUERY OK db=2.1ms queue=0.1ms\ndone [1, 2, 3]"
```

As the result of `:erlang.convert_time_unit` is rounded, the value of `diff` is incorrect and makes the test fails. I think the solution might be to pass to `:erlang.convert_time_unit` the value we want as time. The code would be:

```elixir
    query_time = :erlang.convert_time_unit(2100, :micro_seconds, :native)
    queue_time = :erlang.convert_time_unit(100, :micro_seconds, :native)
    entry = %{entry | params: [1, 2, 3], query_time: query_time , queue_time: queue_time}
    assert to_binary(entry) == "QUERY OK db=2.1ms queue=0.1ms\ndone [1, 2, 3]"
```

The values must be high enouth to avoid precision loss.
i.e.
```elixir
y == :erlang.convert_time_unit(x, :micro_seconds, :native)
x == :erlang.convert_time_unit(y, :native, :micro_seconds)
```

also, for consistency I think we should use `System.convert_time_unit` instead of `:erlang.convert_time_unit` in this test, since it is the one used in Ecto.LogEntry.time/3

josevalim:
Can you please send a pull request? Passing the values as you specified instead of getting a multiplying factor will indeed be better.

PR #2446, opened 2018-02-23
PR for issue 2441

Farijo:
A proposition for the issue [#2441](https://github.com/elixir-ecto/ecto/issues/2441)
I stored the different times in variables, and increased the times to avoid precision loss. But it only repels the problem. I think we should compute the expected values depending on the input value.
```elixir
query_time = :erlang.convert_time_unit(2100, :micro_seconds, :native)
expected_query_ms = query_time |> :erlang.convert_time_unit(:native, :micro_seconds) |> div(1000)
expected_query_ms_dec = query_time |> :erlang.convert_time_unit(:native, :micro_seconds) |> div(100) |> rem(10)
"... db=#{expected_query_ms }.#{expected_query_ms_dec }ms ..."
```
But I am not sure about doing these operations inside a test.


josevalim:

Issue #2277, opened 2017-10-13
Changeset.filters (Ecto.Changeset.optimistic_lock/3) doesn't work with nil fields


umamaistempo:
## Environment

* Elixir version (elixir -v): Elixir 1.4.2
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6
* Ecto version (mix deps): 2.2.0
* Database adapter and version (mix deps): postgrex 0.13
* Operating system: Arch Linux

### Current behavior

Right now, if using `Changeset.optimistic_lock` on a nil field, it will always return an `Ecto.StaleEntryError`

Example code:
```elixir
user = %User{updated_at: nil} |> Repo.insert!()

  User
  |> Repo.get!(user.id)
  |> Ecto.Changeset.change(%{})
  |> Ecto.Changeset.optimistic_lock(:updated_at, fn _ -> NaiveDateTime.utc_now() end)
  |> Repo.update!()
```

This causes an `Ecto.StaleEntryError`.
Inspecting the changeset filters key it returns `%{updated_at: nil}`
The update query is `UPDATE "users" SET "updated_at" = $1 WHERE "updated_at" = $2 AND "id" = $3 [{{2017, 10, 13}, {19, 15, 57, 888374}}, nil, 10]`

Based on the query, i'd assume the problem is with translating the `nil` value. In this case, for this to work, i'd assume that instead of binding and using the nil value, the `IS NULL` comparator would have to be used.

### Expected behavior
When using optimistic_lock with a nil field, the update query would pass as long as the row in database is still null

josevalim:
Good catch!

qcam:
I can work on this if it has not been taken :).

josevalim:
Yes please! Now that we support dynamic in Ecto.Query, I am thinking we can
allow any dynamic to be set as a filter.
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

qcam:
@josevalim can you please elaborate on that? What do you think would be the desired fix for this?

josevalim:
@qcam apologies for the quick answer.

[...]

PR #2447, opened 2018-02-24
Support nil field in Ecto.Changeset.optimistic lock

qcam:
I would like to propose a fix for #2277.

As described in the issue, if `Changeset.filters` has a nil `updated_at` field, the SQL query generated is `"updated_at" = $1` while the desired output would be `"updated_at" IS NULL`.

I suppose this change **WILL break other external adapters**, since it changes the input of [Ecto.Adapters.SQL.update/5](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapters/sql/connection.ex#L82), so I'm not sure this is the desired way to fix. I'm open for any suggestion.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -242,15 +242,54 @@ defmodule Ecto.Integration.RepoTest do
     import Ecto.Changeset, only: [cast: 3, optimistic_lock: 2]
     base_post = TestRepo.insert!(%Comment{})
 
-    cs_ok =
+    changeset_ok =
       base_post
       |> cast(%{"text" => "foo.bar"}, ~w(text)a)
       |> optimistic_lock(:lock_version)
-    TestRepo.update!(cs_ok)
+    TestRepo.update!(changeset_ok)
 
-    cs_stale = optimistic_lock(base_post, :lock_version)
-    assert_raise Ecto.StaleEntryError, fn -> TestRepo.update!(cs_stale) end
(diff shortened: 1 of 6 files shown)


integration_test/support/schemas.exs
```diff
@@ -76,6 +76,7 @@ defmodule Ecto.Integration.Comment do
   schema "comments" do
     field :text, :string
     field :lock_version, :integer, default: 1
+    field :nullable_lock_version, :integer, default: nil
```
josevalim: Instead of introducing a new field, can we simply use the previous field but we make sure to set it to nil when we first create the record?

integration_test/cases/repo.exs
```diff
@@ -253,6 +253,27 @@ defmodule Ecto.Integration.RepoTest do
...
+    import Ecto.Changeset, only: [cast: 3, optimistic_lock: 3]
+    base_post = TestRepo.insert!(%Comment{lock_version: nil})
+
+    incrementer =
+      fn
+        nil -> 1
+        old_value -> old_value + 1
+      end
+
+    cs_ok =
```
josevalim: Let's not abbreviate variables please. :)

josevalim:
The fix is perfect @qcam! I have just added some questions regarding the tests.

qcam:
Great thanks for the comments, I will make the same change to `DELETE` too.

josevalim:

PR #2449, opened 2018-02-27
Add Ecto.Association.NotLoaded.t/0 typespec

yordis:


lib/ecto/association.ex
```diff
@@ -9,8 +9,19 @@ defmodule Ecto.Association.NotLoaded do
     * `__field__` - the association field in `owner`
     * `__owner__` - the schema that owns the association
     * `__cardinality__` - the cardinality of the association
+  """
 
+  @typedoc """
```
josevalim: I don't think we need those docs since they are a copy of the struct docs.

lib/ecto/association.ex
```diff
@@ -9,8 +9,19 @@ defmodule Ecto.Association.NotLoaded do
...
+  * `__field__` - the association field in `owner`
+  * `__owner__` - the schema that owns the association
+  * `__cardinality__` - the cardinality of the association
   """
+  @type t :: %__MODULE__{
+    __field__: atom(),
```
wojtekmach: because of underscores it seems to me it's private-ish, maybe the the `t()` type should be opaque or these fields private?

josevalim: They are not private. The reason why they are underscored is to avoid accidental matches in code like this `comment.post.owner` for the case the post is not loaded. :)

wojtekmach: nevermind, the fields are documented so they're public. the reason it's `__field` was probably to minimize the chance of having the same fields in this struct as well as loaded schema struct. Sorry for noise, I'm gonna remove both comments.

yordis:
@josevalim I thought on that but so have a feedback. 

I find myself clicking on types and trying to read the typedocs before anything, because ex_doc put it closer to the type documentation.

Maybe we should move the struct field doc there or just be happy with duplicated documentation? 

I would prefer to keep the documentation as close as possible to the struct. 

What do you want me to do ?

josevalim:
ExDoc supports defining `@doc` before defstruct too. I am fine with any approach, honestly.

yordis:
The doc code is small enough. It would leave as it is

josevalim:
To clarify, I am fine with any approach that does not involve duplication. :)

yordis:
@josevalim done!

josevalim:

PR #2451, opened 2018-02-27
Correct spec on Repo.transaction: Multi.name, not atom

novaugust:
A multi's name can is type `any()`, so `atom` is misleading here.


josevalim:

PR #2456, opened 2018-03-04
Adding documentation for arithmetic operators

### LINKED ISSUE #1757: Support mathematical infix operators

costaraphael:
Resolves #1757

I thought it was interesting to add some notes on how the operators could be used in any way supported by the adapter, but I didn't want to repeat the same paragraph for each operator, so I created a top level section explaining this and left the individual operators documentation pretty much bare.

What do you guys think?


josevalim:

Issue #2455, opened 2018-03-03
Some binaries fail required validation


haljin:
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 20 [erts-9.2.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.1 (compiled with OTP 20)
```
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): `postgres (PostgreSQL) 10.2`
* Ecto version (mix deps): `ecto 2.2.8 (Hex package) (mix)`
* Database adapter and version (mix deps): `postgrex 0.13.5 (Hex package) (mix)`
* Operating system: `OS X High Sierra`

### Current behavior

I have the following schema defined:
```elixir
defmodule PaintSchemer.Paints.Paint do
  use Ecto.Schema
  import Ecto.Changeset
  alias PaintSchemer.Paints.Paint


  schema "paints" do
    field :color, :binary
    field :name, :string
    belongs_to :manufacturer, PaintSchemer.Paints.Manufacturer
    belongs_to :type, PaintSchemer.Paints.PaintType

    timestamps()
  end

  @doc false
  def changeset(%Paint{} = paint, attrs) do
    paint
    |> cast(attrs, [:name, :color, :type_id, :manufacturer_id])
    |> validate_required([:name, :color, :type_id, :manufacturer_id])
  end
end
```

With the table created like this:
```elixir
defmodule PaintSchemer.Repo.Migrations.CreatePaints do
  use Ecto.Migration

  def change do
    create table(:paints) do
      add :name, :string
      add :color, :binary
      add :manufacturer_id, references(:paint_manufacturers, on_delete: :nothing)
      add :type_id, references(:paint_types, on_delete: :nothing)

      timestamps()
    end

    create index(:paints, [:manufacturer_id])
    create index(:paints, [:type_id])
  end
end
```

I also have some unit tests, that want to create this schema:

```elixir
@create_attrs %{color: <<12, 12, 12>>, name: "some name"}
  @update_attrs %{color: <<13, 12, 12>>, name: "some updated name"}
  @invalid_attrs %{color: nil, name: nil}

  def fixture(:paint) do
    {:ok, type} = Paints.create_paint_type(%{name: "Some paint type"})
    {:ok, manufacturer} = Paints.create_manufacturer(%{name: "Some manufacturer"})
[...]

PR #2457, opened 2018-03-06
Skip trim value for binary data type for required validation

hlhr:
Fix for issue #2455 

When checking validate_required, get the data type of the field and skip trim value when the datatype is binary.

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -41,6 +41,7 @@ defmodule Ecto.ChangesetTest do
       field :title, :string, default: ""
       field :body
       field :uuid, :binary_id
+      field :color, :binary
       field :decimal, :decimal
       field :upvotes, :integer, default: 0
       field :topics, {:array, :string}
@@ -55,7 +56,7 @@ defmodule Ecto.ChangesetTest do
   end
 
   defp changeset(schema \\ %Post{}, params) do
-    cast(schema, params, ~w(id token title body upvotes decimal topics virtual)a)
+    cast(schema, params, ~w(id token title body upvotes decimal color topics virtual)a)
   end
 
   ## cast/4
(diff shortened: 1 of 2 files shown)


michalmuskala:
I'm not a huge fan of special-casing  the `:binary` type since it means any custom type using `:binary` underneath won't work correctly. But I'm also not sure what else we could do.

test/ecto/changeset_test.exs
```diff
@@ -723,6 +724,13 @@ defmodule Ecto.ChangesetTest do
     assert changeset.changes == %{}
     assert changeset.errors == [title: {"is blank", [validation: :required]}, body: {"is blank", [validation: :required]}]
 
+    # When type is binary skip value triming (byte 12 == emtpy character in strings)
+    changeset =
+        changeset(%{color: <<12, 12, 12>>})
```
fertapric: Alignment here is a bit weird. 

Maybe a oneliner could work here `validate_required(changeset(%{color: <<12, 12, 12>>}), :color)`

fertapric:
@hlhr @michalmuskala could be an option to provide an `:only_nil` (or `:allow_blank`) option?

```
validate_required(changeset, :color, only_nil: true)
validate_required(changeset, :color, allow_blank: true)
```

As an alternative, other option is to pass an `:empty_values` option like in `cast/4` (which defaults to `[nil, ""]`).

test/ecto/changeset_test.exs
hlhr: Thanks for the feedback, I changed the format to a oneliner but used the same format as in the tests above (see test when missing).

michalmuskala:
Maybe a solution would be to have a `trim: false` option? I think this would be the simplest way to solve this. I'm not a huge fan of adding more options, but I don't really see another clean way to do this.

hlhr:
@michalmuskala That's the option with the best backwards compatibility and it's pretty simple, so it's makes sense to do this. 
I have some free time and would close this PR, develop the new option, add it as new PR and would link this PR to #2455 and this discussion #2457. Is this the right way to handle this or should I create a additional new Issue?

josevalim:
@hlhr I like the trim approach too!

You don't need to open up an new issue, you can just close this and send a new PR.

Thank you! :heart:

PR #2464, opened 2018-03-14
Add trim option for validate_required

hlhr:
Fix for issue #2455 
Discussed in #2457

A new option for `validate_required` to skip the `String.trim_leading` before the check if the value is empty. The default is `true` to ensure backwards compatibility.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1464,6 +1464,7 @@ defmodule Ecto.Changeset do
   ## Options
 
     * `:message` - the message on failure, defaults to "can't be blank"
+    * `:trim` - trims value before required check to remove whitespace, defaults to true
 
   ## Examples
 
@@ -1475,11 +1476,12 @@ defmodule Ecto.Changeset do
   def validate_required(%Changeset{} = changeset, fields, opts \\ []) do
     %{required: required, errors: errors, changes: changes} = changeset
     message = message(opts, "can't be blank")
+    trim = Keyword.get(opts, :trim, true)
     fields = List.wrap(fields)
 
     fields_with_errors =
       for field <- fields,
-          missing?(changeset, field),
+          missing?(changeset, field, trim),
           ensure_field_exists!(changeset, field),
           is_nil(errors[field]),
           do: field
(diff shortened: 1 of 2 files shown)


test/ecto/changeset_test.exs
```diff
@@ -723,6 +724,15 @@ defmodule Ecto.ChangesetTest do
     assert changeset.changes == %{}
     assert changeset.errors == [title: {"is blank", [validation: :required]}, body: {"is blank", [validation: :required]}]
 
+    # When :trim option is false
```
fertapric: Sorry for the off-topic, but do you guys think it could be better to put these commens in different tests so in case of errors it's easier to know what's failing?

```elixir
# Option 1
describe "validate_required/2" do
  test "when valid" do
    # ...
  end

  test "when missing" do
    # ...
  end

  test "when :trim option is false" do
    # ...
  end
end

# Option 2
test "validate_required/2 when valid" do
  # ...
end

test "validate_required/2 when missing" do
  # ...
end

test "validate_required/2 when :trim option is false" do
  # ...
end
```

If you like any of the options I could apply those changes in a different PR :)

lib/ecto/changeset.ex
```diff
@@ -1464,6 +1464,7 @@ defmodule Ecto.Changeset do
   ## Options
 
     * `:message` - the message on failure, defaults to "can't be blank"
+    * `:trim` - trims value before required check to remove whitespace, defaults to true
```
fertapric: What do you think about?

* `:trim` - boolean, sets whether leading whitespaces are removed before running the validation. Defaults to true

hlhr: Sounds definitely better and more precise.

test/ecto/changeset_test.exs
hlhr: I like option 1 more but this should be decided by someone which is more involved in the project like @michalmuskala or @josevalim

lib/ecto/changeset.ex
```diff
@@ -1464,6 +1464,8 @@ defmodule Ecto.Changeset do
   ## Options
 
     * `:message` - the message on failure, defaults to "can't be blank"
+    * `:trim` - boolean, sets whether leading whitespaces are removed
+                before running the validation, defaults to true
```
fertapric: Looking at the codebase, it seems "Options" are indented differently:

```
  * `:trim` - boolean, sets whether leading whitespaces are removed
    before running the validation, defaults to true
```

Check https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L390-L391

test/ecto/changeset_test.exs
josevalim: @fertapric a PR that moves to this style would be appreciated. This project started way before Elixir v1.0, so we didn't have such constructs at the time. :)

josevalim:

Issue #2465, opened 2018-03-17
Test Environment Setup Documentation


philippneugebauer:
I wanted to extend my existing Elixir application with Ecto. While the setup of the normal application was straight-forward through the documentation, I struggled a lot to set up the test environment. Finally, I was able to set it up to work following this post:
https://elixirforum.com/t/how-to-set-up-the-testing-environment-for-a-library-that-uses-ecto/4770/3

I was searching quite a time to find possible test setups but they all didn't work till the last one right now. I think, we should extend the documentation about the information of the post since it is relevant information.

josevalim:
Thanks @philippneugebauer! Could you please send a pull request with this documentation?

I would add a new file [here](https://github.com/elixir-ecto/ecto/tree/master/guides) saying "Adding Ecto to a project.md" that covers regular and the test environment setup.

Let me know if you have any questions!

philippneugebauer:
@josevalim what happens then to the at least partially redundant information in getting started?
Should it be added to the Readme as well?

josevalim:
@philippneugebauer oh, I see. Let's add them a "Testing with Ecto" section and make sure to link to it from the Getting Started. :)

PR #2466, opened 2018-03-17
Chapter for Test Environment Setup

philippneugebauer:
https://github.com/elixir-ecto/ecto/issues/2465

### DIFF
--- a/guides/Getting Started.md
+++ b/guides/Getting Started.md
@@ -119,6 +119,9 @@ This tells our application about the repo, which will allow us to run commands s
 
 We've now configured our application so that it's able to make queries to our database. Let's now create our database, add a table to it, and then perform some queries.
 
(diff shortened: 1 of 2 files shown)


philippneugebauer:
@josevalim I've added the chapter

philippneugebauer:
Should the `aliases` function for the `test` command also be included? I am not sure about that.

josevalim:
Thanks @philippneugebauer! I have done some changes that expand a bit more the point of using the sandbox. Can you please review my changes?

I think we should also mention the aliases in mix.exs because it makes running tests more convenient, could you please add that too?

Thanks!

philippneugebauer:
I have also added the `aliases`. Should be good to merge now

josevalim:

PR #2471, opened 2018-03-21
Add missing formatter file

bokay:
By doing `{"ecto", "~> 2.2.9", override: true}` in my mix.exs and  `import_deps: [:ecto]` in my own formatter, I was expecting to have the ecto rules set in my project.

However, it seems that the file `.formatter.exs` is missing from the package on hex. 

I then make the assumption that it should be present in the branch `v2.2`.
This commit aims to complete this one https://github.com/elixir-ecto/ecto/commit/9463e1c9564cd971754e3324033cdc5546ac428f

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,0 +1,32 @@
+locals_without_parens = [
+  # Query
+  from: 2,
+
+  # Schema
+  field: 1,
+  field: 2,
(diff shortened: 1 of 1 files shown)


josevalim:
:heart: :green_heart: :blue_heart: :y...

PR #2472, opened 2018-03-22
Queries in Associations

zachdaniel:
There wasn't much to do to make this work as expected, because everything aside from the validations was expecting associations to be expressed as queryables anyway (as far as I can tell). I have a few test failures that I'm not 100% sure on, but they appear to be due to changes I made to test schemas. I didn't want to just change them off hand though, because for all I know I broke something real. So I left those tests failing, and wrote new tests to express both the capability of building the query with queries for assocs, and the planning of it. I was going to write integration tests for it, but I figured that if the planner is breaking it out into the correct joins, that it would be unnecessary to have an integration test.

This is in service of providing a lot of the value #2395 was looking for, without having to introduce the larger change of default where clauses.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -1048,8 +1048,18 @@ defmodule Ecto.Association.ManyToMany do
   defp validate_join_through(name, nil) do
     raise ArgumentError, "many_to_many #{inspect name} associations require the :join_through option to be given"
   end
-  defp validate_join_through(_, %Ecto.Query{}) do
-    :ok
+  defp validate_join_through(name, query = %Ecto.Query{}) do
+    allowed_keys = [:wheres, :sources, :from, :aliases]
+    default = %Ecto.Query{}
+    query
+    |> Map.from_struct()
+    |> Enum.each(fn {key, value} ->
+      unless (key in allowed_keys) or Map.get(default, key) == value do
+        raise ArgumentError,
+          "many_to_many #{inspect name} was given a query with a `#{inspect key}` statement. Associations " <>
+          "cannot accept queries with select `#{inspect key }` statements"
+      end
+    end)
   end
   defp validate_join_through(_, join_through) when is_atom(join_through) or is_binary(join_through) do
     :ok
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -261,6 +261,8 @@ defmodule Ecto.Association do
   """
   def related_from_query(atom) when is_atom(atom), do: atom
   def related_from_query({source, schema}) when is_binary(source) and is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: schema}) when is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: {source, schema}}) when is_binary(source) and is_atom(schema), do: schema
   def related_from_query(queryable) do
     raise ArgumentError, "association queryable must be a schema " <>
       "or {source, schema}, got: #{inspect queryable}"
@@ -884,13 +886,7 @@ defmodule Ecto.Association.ManyToMany do
   def struct(module, name, opts) do
     join_through = opts[:join_through]
 
-    if join_through && (is_atom(join_through) or is_binary(join_through)) do
-      :ok
-    else
-      raise ArgumentError,
-        "many_to_many #{inspect name} associations require the :join_through option to be " <>
-        "given and it must be an atom (representing a schema) or a string (representing a table)"
-    end
+    validate_join_through(name, join_through)
 
     join_keys = opts[:join_keys]
     queryable = Keyword.fetch!(opts, :queryable)
@@ -1047,6 +1043,21 @@ defmodule Ecto.Association.ManyToMany do
       {:error, changeset} ->
         {:error, changeset}
     end
+  end
+
+  defp validate_join_through(name, nil) do
+    raise ArgumentError, "many_to_many #{inspect name} associations require the :join_through option to be given"
+  end
+  defp validate_join_through(_, %Ecto.Query{}) do
+    :ok
+  end
+  defp validate_join_through(_, join_through) when is_atom(join_through) or is_binary(join_through) do
+    :ok
+  end
+  defp validate_join_through(name, _join_through) do
+    raise ArgumentError,
+      "many_to_many #{inspect name} associations require the :join_through option to be " <>
+      "an atom (representing a schema) or a string (representing a table)"
   end
 
   defp insert_join?(%{action: :insert}, _, _field, _related_key), do: true
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -210,6 +210,19 @@ defmodule Ecto.Association do
     |> distinct([x], true)
   end
 
+  @doc """
+  Add the default assoc query where clauses a provided query
+  """
+  def combine_assoc_query(assoc_queryable, nil), do: assoc_queryable
+  def combine_assoc_query(%Ecto.Query{wheres: assoc_wheres}, provided_queryable) do
+    provided_queryable
+    |> Ecto.Queryable.to_query()
+    |> Map.update!(:wheres, &Kernel.++(&1, assoc_wheres))
+  end
+  def combine_assoc_query(_, provided_queryable) do
+    provided_queryable
+  end
+
   defp assoc_to_where(%{on: %QueryExpr{} = on}) do
     on
     |> Map.put(:__struct__, BooleanExpr)
@@ -520,13 +533,14 @@ defmodule Ecto.Association.Has do
 
   @doc false
   def assoc_query(%{queryable: queryable, related_key: related_key}, query, [value]) do
-    from x in (query || queryable),
+
+    from x in Ecto.Association.combine_assoc_query(queryable, query),
       where: field(x, ^related_key) == ^value
   end
 
   @doc false
   def assoc_query(%{queryable: queryable, related_key: related_key}, query, values) do
-    from x in (query || queryable),
+    from x in Ecto.Association.combine_assoc_query(queryable, query),
       where: field(x, ^related_key) in ^values
   end
 
@@ -781,13 +795,13 @@ defmodule Ecto.Association.BelongsTo do
 
   @doc false
   def assoc_query(%{queryable: queryable, related_key: related_key}, query, [value]) do
-    from x in (query || queryable),
+    from x in Ecto.Association.combine_assoc_query(queryable, query),
       where: field(x, ^related_key) == ^value
   end
 
   @doc false
   def assoc_query(%{queryable: queryable, related_key: related_key}, query, values) do
-    from x in (query || queryable),
+    from x in Ecto.Association.combine_assoc_query(queryable, query),
       where: field(x, ^related_key) in ^values
   end
 
@@ -970,7 +984,7 @@ defmodule Ecto.Association.ManyToMany do
     # We need to go all the way using owner and query so
     # Ecto has all the information necessary to cast fields.
     # This also helps validate the associated schema exists all the way.
-    from q in (query || queryable),
+    from q in Ecto.Association.combine_assoc_query(queryable, query),
       join: o in ^owner, on: field(o, ^owner_key) in ^values,
       join: j in ^join_through, on: field(j, ^join_owner_key) == field(o, ^owner_key),
       where: field(j, ^join_related_key) == field(q, ^related_key)
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -262,21 +262,30 @@ defmodule Ecto.Association do
 
   ## Examples
 
-      iex> Ecto.Association.related_from_query({"custom_source", Schema})
+      iex> Ecto.Association.related_from_query({"custom_source", Schema}, :comments_v1)
       Schema
 
-      iex> Ecto.Association.related_from_query(Schema)
+      iex> Ecto.Association.related_from_query(Schema, :comments_v1)
       Schema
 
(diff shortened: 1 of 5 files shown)


zachdaniel:
Additionally, I added a validation for only allowing certain fields of the provided query to be modified, but I'm not really sure if thats a good way to do it or not. I also wasn't sure if perhaps we should just *ignore* any additional values in the query, in which case that validation would not be necessary. Due to that, I have not yet written tests for it.

test/ecto/query/planner_test.exs
```diff
@@ -16,9 +16,49 @@ defmodule Ecto.Query.PlannerTest do
       field :temp, :string, virtual: true
       field :posted, :naive_datetime
       field :uuid, :binary_id
+      field :special, :boolean
       belongs_to :post, Ecto.Query.PlannerTest.Post
       has_many :post_comments, through: [:post, :comments]
     end
+
+    def special() do
```
honkycat: Special isn't the best name, a name that describes WHAT is special would be better.

zachdaniel: Its defined on the `Comment` schema, so `Comment.special()` returns a query for special comments.

test/ecto/schema_test.exs
```diff
@@ -516,6 +633,42 @@ defmodule Ecto.SchemaTest do
...
+      join_through: AssocEditors.active()
+    }
+
+    assert AssocSchema.__schema__(:association, :awesome_editors) == struct
+    assert AssocSchema.__changeset__.awesome_editors == {:assoc, struct}
+
+    awesome_editors = (%AssocSchema{}).awesome_editors
+    assert %Ecto.Association.NotLoaded{__cardinality__: :many} = awesome_editors
+    assert inspect(awesome_editors) == "#Ecto.Association.NotLoaded<association :awesome_editors is not loaded>"
+  end
```
josevalim: Those tests are testing implementation details and not any behaviour and I avoid those kind of tests generally. For example, we may rename a field, everything will continue to work, but we would have failing tests. I suggest to remove the tests added to this file.

test/ecto/query/planner_test.exs
```diff
@@ -16,9 +16,49 @@ defmodule Ecto.Query.PlannerTest do
...
+      from comment in __MODULE__,
+        where: comment.special
+    end
+  end
+
+  defmodule Author do
```
josevalim: We prefer to using the existing modules instead of adding new modules, otherwise it becomes impossible to manage all of the schemas in this file. Would that be possible?

josevalim:
Thanks @zachdaniel!

The code looks good but I think we missed the proper coverage with tests. Please take a look at the existing tests in "test/ecto/associations_test.exs". Notice how they are checking for a specific behaviour in assoc_query/join_query. Those are the tests we need to write. For each association type, we need to add a test that covers `assoc_query` and `join_query` using the new association definition. Note there we use `inspect` to compare different queries without the line limits.

At the end of the same file, there are preload tests, that we should add too.

In the planner test, we should add an integration test, that would verify the new mechanism works via a complex join. Imagine something like:

    from a in A, join: b in assoc(a, :b), join: c in assoc(b, :c), join: d in assoc(c, :d)

Where each join uses a queryable association. This is the most important test to have because we need to make sure all of the query bindings are properly rewritten.

In all cases, please avoid adding new schemas to the test files.

lib/ecto/association.ex
```diff
@@ -261,6 +261,8 @@ defmodule Ecto.Association do
   """
   def related_from_query(atom) when is_atom(atom), do: atom
   def related_from_query({source, schema}) when is_binary(source) and is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: schema}) when is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: {source, schema}}) when is_binary(source) and is_atom(schema), do: schema
```
josevalim: We don't need two clauses here because queries are always in the form `from: {source, schema}`. However, we do need to check that it only has a `where` field and nothing more. Something like this:

```elixir
case query do
      %Ecto.Query{order_bys: [], limit: nil, offset: nil, group_bys: [], joins: [],
                  havings: [], preloads: [], assocs: [], distinct: nil, lock: nil} ->
        query
      _ ->
        raise ...
    end
```

lib/ecto/association.ex
```diff
@@ -1049,6 +1045,31 @@ defmodule Ecto.Association.ManyToMany do
     end
   end
 
+  defp validate_join_through(name, nil) do
+    raise ArgumentError, "many_to_many #{inspect name} associations require the :join_through option to be given"
+  end
+  defp validate_join_through(name, %Ecto.Query{} = query) do
```
josevalim: Let's do this validation with pattern matching. I have just posted an example above. :)

lib/ecto/association.ex
```diff
@@ -1049,6 +1045,31 @@ defmodule Ecto.Association.ManyToMany do
...
+      end
+    end)
+  end
+  defp validate_join_through(_, join_through) when is_atom(join_through) or is_binary(join_through) do
+    :ok
+  end
+  defp validate_join_through(name, _join_through) do
+    raise ArgumentError,
+      "many_to_many #{inspect name} associations require the :join_through option to be " <>
+      "an atom (representing a schema) or a string (representing a table)"
```
josevalim: We also support a query now. :)

lib/ecto/association.ex
```diff
@@ -261,6 +261,8 @@ defmodule Ecto.Association do
   """
   def related_from_query(atom) when is_atom(atom), do: atom
   def related_from_query({source, schema}) when is_binary(source) and is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: schema}) when is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: {source, schema}}) when is_binary(source) and is_atom(schema), do: schema
   def related_from_query(queryable) do
     raise ArgumentError, "association queryable must be a schema " <>
       "or {source, schema}, got: #{inspect queryable}"
```
josevalim: We also support a query now. So we need to update the error message.

test/ecto/query/planner_test.exs
zachdaniel: I definitely can, but it will be a bit strange modeling the join association that way. If I could add one schema for the join then that would be enough to do it cleanly.

josevalim: Sure! Just add the minimum necessary. :)

lib/ecto/association.ex
zachdaniel: Done!

zachdaniel:
@josevalim I'm curious, it looks like when I provide a query for `assoc_query/3`, like `Ecto.Association.Has.assoc_query(assoc, query, [])`, that the conditions in the association are not honored (which I can see why that would be happening). It also looks like the `Ecto.assoc` helper and friends all pass in explicitly `nil`. With that in mind, I'm not sure if it is correct or incorrect for a query passed into that function to override `where` clauses on the configured query. Thoughts?

josevalim:
@zachdaniel yes, we need to "port" the conditions in `where` over to the query. :)

josevalim:
We probably need to do something similar to join queries as well, porting the where conditions to an `on` clause.

zachdaniel:
Awesome, will do :)

zachdaniel:
Is there perhaps some utility for porting where's from one query to another that I should use instead? Along that same line of questioning, I will need to rewrite the where clauses from the join_through to `on` clauses, which I know is something that happens somewhere in ecto ( like when you join on a query), so I'll need to find a way to tap into that.

josevalim:
@zachdaniel it is no bother at all!

> I'm trying to determine the context in which additional queries can be passed into assoc_query/3 it looks like that is only done during preload.

Yes, I do think this is the case!

> Seeing as it would make no sense to have the source of a query that isn't the same as the destination of the preload, I think that (in the case of everything but many_to_many, haven't looked at that yet) its as simple as appending the where clauses from the default query onto the provided query.

Beautiful! Agreed.

There is no utility, just concatenate the wheres. :)

zachdaniel:
I'm realizing I can use this same logic for the join_through and the existing stuff will adapt, so I think I'll be able to take care of the rest of it.

zachdaniel:
@josevalim so I *think* I've addressed everything you mentioned, and I added a (perhaps pathological) test case in the planner test, and implemented the necessary code to combine where clauses when an override is provided. I still have two failing tests in planner_test and I'm really not sure why. It looks like one of them might have something to do with a cache key that might be based on the fields/associations on the struct, so should I just hardcode the new value? The other one also seems to be testing a complete list of selected fields, so I think I might just need to add the new fields I added to the schema. Any thoughts?

lib/ecto/association.ex
```diff
@@ -256,14 +269,15 @@ defmodule Ecto.Association do
       Schema
 
       iex> Ecto.Association.related_from_query("wrong")
-      ** (ArgumentError) association queryable must be a schema or {source, schema}, got: "wrong"
+      ** (ArgumentError) association queryable must be a schema,  or {source, schema}, or a query. got: "wrong"
 
   """
   def related_from_query(atom) when is_atom(atom), do: atom
   def related_from_query({source, schema}) when is_binary(source) and is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: {source, schema}}) when is_binary(source) and is_atom(schema), do: schema
```
josevalim: We also need to validate here that the query only has `wheres`. :)

test/ecto/query/planner_test.exs
```diff
@@ -229,6 +256,44 @@ defmodule Ecto.Query.PlannerTest do
...
+
+    assert Macro.to_string(join1.on.expr) == "&1.special() and &1.post_id() == &0.id()"
+    assert Macro.to_string(join2.on.expr) == "&2.id() == &1.post_id()"
+    assert Macro.to_string(join3.on.expr) == "not(&6.deleted()) and &6.post_id() == &0.id()"
+    assert Macro.to_string(join4.on.expr) == "&3.special() and &6.comment_id() == &3.id()"
+    assert Macro.to_string(join5.on.expr) == "&4.comment_id() == &3.id()"
+    assert Macro.to_string(join6.on.expr) == "&5.special() and &5.id() == &4.special_comment_id()"
+  end
+
+  test "prepare: joins associations with queries" do
```
josevalim: Nitpick: let's reverse the order of those two tests so the less complex ones comes first. Let's also change the test name to "prepare: joins associations with custom queries".

josevalim:
@zachdaniel beautiful. I have left two minor comments. Then we just need to write documentation and we are good to go! :heart:

josevalim:
@zachdaniel I have noticed that you are relying on the join queries for subqueries. This is actually quite interesting because it means it should be possible to lift the where restriction.

But before we lift any restriction, which I don't think we should now anyway, do you think we will have a difference in performance? Imagine those two queries:

    from s in Summary, join: p in ^(from p in Post, where: p.upvotes >= 1000), on: p.summary_id == s.id

    from s in Summary, join: p in Post, on: p.summary_id == s.id and p.upvotes >= 1000

Which one would be more efficient? Or is it not relevant enough to matter? Could you please investigate by analyzing the generated SQL?

lib/ecto/association.ex
```diff
@@ -256,14 +269,15 @@ defmodule Ecto.Association do
...
   def related_from_query(queryable) do
-    raise ArgumentError, "association queryable must be a schema " <>
-      "or {source, schema}, got: #{inspect queryable}"
+    raise ArgumentError, "association queryable must be a schema, " <>
+      " or {source, schema}, or a query. got: #{inspect queryable}"
```
AndrewDryga: There is already a space on the previous line, you don't need to add it here.

zachdaniel:
I'll analyze the SQL tonight, and look at the plan difference. From my experience, I would not expect the performance difference to be significant, because we aren't really using subqueries, we are just using where vs on. If there is a performance difference, I don't think it would be predictably for one way or the other, but more dependent on the specific usage.

zachdaniel:
@josevalim So since its not using subqueries, but just using the queries as the subject of the join, it actually generates the same sql! The only difference is that the order of the expressions in the `on` are different. So one says `ON (post.id = summary.post_id) and (p.upvotes > 1000)` and the other one says `ON (p.upvotes > 1000) and (post.id = summary.post_id)`, I doubt that would affect performance in any way.

josevalim:
@zachdaniel perfect! This means we can generalize this later to include things like order_by and what not without much hassle. @zachdaniel, we are missing just the docs and we are good to go! A one paragraph description to each association function should be enough. :)

zachdaniel:
@josevalim I added some documentation. I'm sure the wording could be improved, so I'll do another pass through later this evening, but wanted to make sure I hit the key points.

lib/ecto/schema.ex
```diff
@@ -650,6 +650,26 @@ defmodule Ecto.Schema do
         end
       end
 
+  A query with *only* where clauses can be provided, instead of a related schema, and the where clauses
```
josevalim: I think the current description focuses too much on the implementation. My suggestion:

> A query can also be given instead of a schema. Querying, joining or preloading the association will use the given query. Currently only `where` clauses can be provided in queries. Let's see an example:
> ...
> Note building the association does not consider the query filters. For example, if the given query requires the `active` field of the associated records to be true, building such association won't automatically set the `active` field to true.

zachdaniel: Sounds great :) I'll make those changes this evening

zachdaniel:
@josevalim pretty much just replaced mine with what you provided :) Should be good to go!

josevalim:

michalmuskala:
Thanks @zachdaniel for working on this feature and advocating for it earlier. I really like the final outcome.

PR #2475, opened 2018-03-25
Support CTEs

zachdaniel:
I'm primarily PR-ing this because I think the proposal only really makes sense to come along with an MVP (okay, its not *viable*, its just an MP). There are a lot of hacky workarounds for the things that I haven't taken the time to understand, especially the cache. But the point was to get a very minimal implementation of CTEs working for just one example.

Once I realized that there was minimal support for named joins, I realized that it was perhaps an entrypoint to using CTEs. If there are already efforts underway for this, or if this feature is just not desired, it is no big deal, it was just a few hours and was very educational anyway!

If we decide that we like it and should actually implement it, then I figured getting my earliest attempts at an implementation out there might be a good idea, to catch any misconceptions I might have and to avoid me spending too much time going down a bad path.

Testing locally, I was able to use this like so:
```elixir
iex(20)> cte = from row in "users", where: row.super, select: %{id: row.id}
#Ecto.Query<from u in "users", where: u.super, select: %{id: u.id}>
iex(6)> joined_to_cte = from row in "posts", with_cte: cte, as: :super_users, join: super_user in cte(:super_users), on: super_user.id == row.user_id, select: row.id
#Ecto.Query<with: :super_users as: #Ecto.Query<from u in "users", where: u.super, select: %{id: u.id}>,
 from p in "posts", join: s in cte(:super_users), on: s.id == p.user_id,
 select: p.id>
```

Which would result in the following SQL (formatted)
```sql
WITH super_users AS (
  SELECT u0."id" 
  FROM "users" AS u0 
  WHERE (u0."super")
) SELECT p0."id" FROM "posts" AS p0 
INNER JOIN super_users AS super_users 
ON super_users."id" = p0."user_id"
```

Personally I think having `WITH` in our toolbelt would be *great* especially for composability. For instance, lets say that if someone added a different `with` clause w/ the same alias, I think we would just override it, which allows for the quick swapping of certain contexts of the query very easily.

```
super_users_query = from row in "users", where: row.super
from row in query,
  with_cte: super_users_query,
  as: :users

# ...some time later...

from row in query,
  with_cte: (from row in "users"),
  as: :users
```

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -112,21 +112,32 @@ if Code.ensure_loaded?(Postgrex) do
     alias Ecto.Query.{BooleanExpr, JoinExpr, QueryExpr}
 
     def all(query) do
+      IO.inspect(query)
       sources = create_names(query)
       {select_distinct, order_by_distinct} = distinct(query.distinct, sources, query)
 
-      from     = from(query, sources)
-      select   = select(query, select_distinct, sources)
-      join     = join(query, sources)
-      where    = where(query, sources)
-      group_by = group_by(query, sources)
-      having   = having(query, sources)
-      order_by = order_by(query, order_by_distinct, sources)
-      limit    = limit(query, sources)
-      offset   = offset(query, sources)
-      lock     = lock(query.lock)
+      with_ctes = with_ctes(query.with_ctes)
+      from      = from(query, sources)
+      select    = select(query, select_distinct, sources)
+      join      = join(query, sources)
+      where     = where(query, sources)
+      group_by  = group_by(query, sources)
+      having    = having(query, sources)
+      order_by  = order_by(query, order_by_distinct, sources)
+      limit     = limit(query, sources)
+      offset    = offset(query, sources)
+      lock      = lock(query.lock)
 
-      [select, from, join, where, group_by, having, order_by, limit, offset | lock]
+      [with_ctes, select, from, join, where, group_by, having, order_by, limit, offset | lock]
+    end
+
+    def with_ctes(query, current_with_statement \\ [])
+    def with_ctes([], statement), do: statement
+    def with_ctes([with_cte|rest], []) do
+      with_ctes(rest, ["WITH ", to_string(with_cte.as), " AS (", all(with_cte.query), ") "])
+    end
+    def with_ctes([with_cte|rest], current_with_statement) do
+      with_ctes(rest, current_with_statement ++ [", ", to_string(with_cte.as), " AS (", all(with_cte.query), ") "])
     end
 
     def update_all(%{from: from} = query, prefix \\ nil) do
@@ -892,14 +903,20 @@ if Code.ensure_loaded?(Postgrex) do
 
     ## Helpers
 
+    defp get_source(_, _, _, {:cte, cte}) do
+       string_name = to_string(cte)
+       {string_name, string_name}
+    end
     defp get_source(query, sources, ix, source) do
       {expr, name, _schema} = elem(sources, ix)
       {expr || paren_expr(source, sources, query), name}
     end
 
     defp quote_qualified_name(name, sources, ix) do
-      {_, source, _} = elem(sources, ix)
-      [source, ?. | quote_name(name)]
+      case elem(sources, ix) do
+        {[_, "cte", _], _, cte_alias} -> [to_string(cte_alias), ?. | quote_name(name)]
+        {_, source, _} -> [source, ?. | quote_name(name)]
+      end
     end
 
     defp quote_name(name) when is_atom(name) do
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -112,7 +112,6 @@ if Code.ensure_loaded?(Postgrex) do
     alias Ecto.Query.{BooleanExpr, JoinExpr, QueryExpr}
 
     def all(query) do
-      IO.inspect(query)
       sources = create_names(query)
       {select_distinct, order_by_distinct} = distinct(query.distinct, sources, query)
 
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -307,7 +307,7 @@ defmodule Ecto.Query do
 
   defstruct [prefix: nil, sources: nil, from: nil, joins: [], aliases: %{}, wheres: [], select: nil,
              order_bys: [], limit: nil, offset: nil, group_bys: [], updates: [],
-             havings: [], preloads: [], assocs: [], distinct: nil, lock: nil]
+             havings: [], preloads: [], assocs: [], with_ctes: [], distinct: nil, lock: nil]
   @type t :: %__MODULE__{}
 
   defmodule DynamicExpr do
@@ -341,6 +341,19 @@ defmodule Ecto.Query do
     # * tag is the directly tagged value, like Ecto.UUID
     # * type is the underlying tag type, like :string
     defstruct [:value, :tag, :type]
+  end
+
+  defmodule WithCte do
+    @moduledoc false
+    defstruct [:query, :as, :file, :line]
+
+    def apply(%Ecto.Query{aliases: aliases, with_ctes: with_ctes} = query, with_cte, as) do
+      %{query | aliases: Map.put(aliases, as, :with_cte), with_ctes: [with_cte | with_ctes]}
+    end
+    def apply(query, with_cte, as) do
+      __MODULE__.apply(Ecto.Queryable.to_query(query), with_cte, as)
+    end
+
   end
 
   alias Ecto.Query.Builder
@@ -629,8 +642,22 @@ defmodule Ecto.Query do
     from(t, env, count_bind, quoted, to_query_binds(binds))
   end
 
+  defp from([{:with_cte, cte}|t], env, count_bind, quoted, binds) do
+    {t, _on, as} = collect_on_and_as(t, nil, nil)
+    unless as do
+      Builder.error! "`as` keyword is required for with_cte"
+    end
+    with_cte =
+      quote do
+        %WithCte{query: unquote(cte), as: unquote(as)}
+      end
+
+    quoted = Builder.apply_query(quoted, WithCte, [with_cte, as], env)
+    from(t, env, count_bind, quoted, binds)
+  end
+
   defp from([{:as, _value}|_], _env, _count_bind, _quoted, _binds) do
-    Builder.error! "`as` keyword must immediately follow a join"
+    Builder.error! "`as` keyword must immediately follow a join or a with"
   end
 
   defp from([{:on, _value}|_], _env, _count_bind, _quoted, _binds) do
@@ -798,6 +825,11 @@ defmodule Ecto.Query do
     end
   end
   defp parse_join_opts(expr), do: {:expr, expr}
+
+
+  # defmacro with_cte(query, qual, binding \\ [], expr, opts \\ []) do
+
+  # end
 
   @doc """
   A select query expression.
--- a/lib/ecto/query/builder/join.ex
+++ b/lib/ecto/query/builder/join.ex
@@ -89,6 +89,10 @@ defmodule Ecto.Query.Builder.Join do
     {:_, quote(do: Ecto.Query.Builder.Join.join!(unquote(expr))), nil, %{}}
   end
 
+  def escape({:cte, _, [cte]} = expr, _vars, _env) do
+    {:_, {:cte, cte}, nil, %{}}
+  end
+
   def escape(join, vars, env) do
     case Macro.expand(join, env) do
       ^join ->
--- a/lib/ecto/query/inspect.ex
+++ b/lib/ecto/query/inspect.ex
@@ -47,6 +47,8 @@ defimpl Inspect, for: Ecto.Query do
       |> generate_names
       |> List.to_tuple
 
+    with_ctes = with_ctes(query.with_ctes)
+
     from      = bound_from(query.from, elem(names, 0))
     joins     = joins(query.joins, names)
     preloads  = preloads(query.preloads)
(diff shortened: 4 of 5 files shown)


zachdaniel:
I will say this about CTEs though: If you don't use them right, they will hurt your performance significantly. The example I posted above would, in fact, could even be a very bad idea if your users table is large, or if you actually really only wanted to target a single user. I personally have a few good use cases for CTEs, but its possible that others might instinctively use it incorrectly.

josevalim:
My concern with CTEs is how they are going to compose. I do like the idea of having an equivalent to `subquery` called `with_cte` and absorb the complexity in there. This way we don't even need to track table bindings because we can generate all of those automatically in the adapter. We can also reuse the whole SubQuery structure we have for planning. My only concern is that it imposes some limitations, like the `select` must be a map or a source selector such as `p` but I think those limitations are fine for both subqueries and CTEs.

If we are going in this direction, I would actually prefer to not introduce a new struct (WithCTE) and simply introduce a new field in `Ecto.SubQuery` which says which kind of subquery it is. It may either be `:subquery` itself or `:with_cte`. Thoughts?

michalmuskala:
One limitation with this approach, is that I don't think it would be possible for one CTE to refer to another, or would it?

josevalim:
@michalmuskala I think it would with named bindings.

zachdaniel:
@josevalim I agree on reusing the subquery struct.

I actually think having the ability for one CTE to reference past CTEs also plays into a feature that I'd really like to see, which is embedding a query and allowing it referencing the outer query for things like `WHERE EXISTS` and `WHERE IN`. Since each query is built in isolation, I don't see a way that we could reference one from the other, without having one of two things:
Some kind of "trust this binding will be available" option when building the query, or a way to tell a query that it is being evaluated in the context of another query and give it the list of bindings. I think the second one would be limiting in that it would require the query definitions to be actually nested. I'd bet the planner could work with something like this:

```
cte1 = from user in "users", where: user.super, select: %{id: user.id}
cte2 = from post in "posts", where: user.awesome, join: super_user in outer(:super_users), on: super_user.id == user.user_id

from photo in "photos",
  with_cte: cte1,
  as: :super_users,
  with_cte: cte2,
  as: :super_user_posts,
  join: post in cte(:super_user_posts),
  on: post.id == photo.post_id
```

The operative thing here is `outer(:super_user_posts)`. We could support inner queries referencing outer query values with `outer` as well, with something like

```
from user in "users",
  where_exists: (
    from photo in "photos",
      where: photo.id == outer(user.id) 
  )
```

Note before I forget: We also need to have the ability to use a cte reference in the `from` of a subsequent CTE or in the query itself.

zachdaniel:
Honestly, I think having to use `outer` to sort of breakout of the current query context significantly adds to the readability of nested queries as it is, and makes accessing outer bindings much more obvious which is a net win when embedding queries. I might like to see that feature supported before this one maybe.

zachdaniel:
From a technical standpoint, having to support a binding refence in the form of `outer(<binding>)` everywhere may be non-ideal. Perhaps having an `outer` keyword, where you list the bindings you're going to use from the outer query.

```
cte1 = from user in "users", where: user.super, select: %{id: user.id}
cte2 = 
  from post in "posts",
    outer: [:super_users],
    where: user.awesome, 
    join: super_user in :super_users, 
    on: super_user.id == user.user_id

from photo in "photos",
  with_cte: cte1,
  as: :super_users,
  with_cte: cte2,
  as: :super_user_posts,
  join: post in cte(:super_user_posts),
  on: post.id == photo.post_id
```

josevalim:
> plays into a feature that I'd really like to see, which is embedding a query and allowing it to reference the outer query

This is a much more complex problem and it deserves a separate discussion. If this is a requirement for CTEs, then we should stop this and start this discussion instead. I don't think it is a problem at the planner level, because the planner would have the full query already. It is rather at the DSL+composition level.

The query syntax is already complex, adding `outer`, `where_exists`, `with_cte`, `cte` does not make me happy. It also means that queries can have "holes" in them and then the query composition is no longer linear. Sure, it is more flexible, but much harder to understand in the long term. So if we can't come up with something simple, I am fine with saying you will have to write those by hand. :)

All of the queries above can be written linearly, without outer/holes. The only case you would need holes (besides where exists which is a separate discussion, are inefficient and can be written as a join) is in co-recursive CTE which I am not even sure they are possible (and also a case I would be glad to say to write a SQL by hand).

zachdaniel:
@josevalim I agree that it leads to composition issues. I'd almost rather see a separate, higher level abstraction over subqueries that relies on ecto's superior composability when compared to SQL.  That last example I posted, where two CTEs are used and the second one references the first, the query itself only references the second one, meaning that the first two CTEs could have been combined. I wonder if we could do something automatically. This might just confuse things, or cause ecto to produce poorly performant SQL (in which case its definitely not worth it), but we may be able to work it out.

Something along the lines of determining if a subquery is referenced
```
super_users = from user in "users", where: user.super

from user in "users",
  join: super_user in subquery(super_users),
  on: super_user.id == user.id
```
Would generate:
```
SELECT * FROM users
JOIN (
  SELECT * FROM users
  WHERE users.super
) AS super_users
```
But what if (and this is a contrived example)
```
from user in "users",
  join: super_user in subquery(super_users),
  on: super_user.id == user.id,
  join: super_user2 in subquery(super_users),
  on: super_user2.id == user.id
```
Generated:
```
WITH super_users AS (
  SELECT * FROM users
  WHERE users.super
)
SELECT * FROM users
JOIN super_users AS su1
ON super_users.id = users.id
JOIN super_users AS su2
ON super_users.id = users.id
```

Just kind of a random idea. The point being that perhaps we could support using things like `WHERE EXISTS` and CTEs like this where each thing that wants to use brings in the subquery itself, effectively meaning that each portion of the query expresses in what way it depends on a specific subquery, and ecto decides how to shape the queries. This means that each inner query/query expresses itself completely, and we no longer have to allow holes in our queries.

zachdaniel:
Hmm... thinking about it a little more, this doesn't give us what we need for inner queries to reference the parent query. I would say our best bet at this point to be productive would be to evaluate the value of CTEs that *can't* reference the prior CTEs. We should determine if that is valuable/desired on its own, and then move a discussion on embedding queries/referencing values from an outer query to a google discussion.

josevalim:
@zachdaniel can you provide queries from your application that you would like to write?

A regular CTE, assuming it is a select without recursion, is equivalent to subqueries. CTE help you organize your code in SQL but we have other ways of organizing your queries in Ecto, so this use case of CTE is not strictly needed.

If you need CTE only to avoid multiple SELECTs, then we can provide one solution. If you need CTE to embed a UPDATE/DELETE in a query, that's another solution. If you need CTE for recursive cases, it may be another case.

Seeing some examples will help us move forward.

zachdaniel:
@josevalim I can't give you any examples of a CTE that we are using currently, primarily because we've avoided using them since it would currently require the entire query to be raw sql, but I can definitely formulate the use case that made me think of this :D.  Our API is very graphql-ish(we follow the JSON API spec) in that you fetch a list of records, sortable/filterable on attributes, and request additional metadata given some context (which are further filterable and sortable). An example of this might be the request to `fetch me users, sorted by count of comments, with content, permissions, group, and parental controls metadata`. In order to support the arbitrary sorting/filtering based on any piece of metadata (most importantly we support this *with* pagination), we embed metadata in the query. This pattern relies on there always being a filter applied that significantly reduces the scope of the query (which has just been true in practice w/o us having to enforce it), like requests are made for "users in this group" not "for all users". So with that in mind, we apply any non-metadata filters, then pass the query to a metadata generator which then adds metadata based on the provided metadata (a combo of requested metadata and parameterization for that requested metadata). Then that query adds to the select statement of the primary query using `select_merge` and joining to a subquery. We keep this pattern performant by virtue of passing in the base query (the one scoped to some small non-metadata based filter scope) and using that as the source for each subquery.

```
def add_metadata(query, requested_context) do
  query
  |> add_group_metadata(query, requested_context)
  |> add_content_metadata(query, requested_context)
end

defp add_group_metadata(query, base, %{"group" => group_id}) do
  from user in query,
    left_join: group_stat in subquery(group_stats(base, group_id)),
    on: user.id == group_stat.id,
    select_merge: %{
      count_of_posts_in_group: group_stat.count_of_posts_in_group,
      sum_of_upvotes_on_posts_in_group: group_stat.sum_of_upvotes_on_posts_in_group
    }
end
defp add_group_metadata(query, _, _), do: query

defp add_content_metadata(query, base, %{"content" => true}) do
  from user in query,
    left_join: content_stat in subquery(content_stats(base)),
    on: user.id == content_stat.id,
    select_merge: %{
      count_of_private_photos: content_stat.count_of_private_photos,
      count_of_photos: content_stat.count_of_photos,
      sum_of_storage_used: content_stat.sum_of_storage_used
    }
end
defp add_group_metadata(query, _, _), do: query

defp group_stats(query, group_id) do
  from user in query,
    left_join: posts in Post,
    on: posts.user_id == user.id,
    on: posts.group_id == ^group_id,
    group_by: user.id,
    select: %{
      id: user.id,
      count_of_posts_in_group: count(post.id),
      sum_of_upvotes_on_posts_in_group: sum(post.upvotes)
    }
end

defp content_stats(query) do
  from user in query,
    left_join: photos in Photo,
    on: photos.user_id == user.id,
    group_by: user.id,
    select: %{
      id: user.id,
      count_of_private_photos: fragment("COUNT ? FILTER (WHERE ?)", photos.id, photos.private),
      count_of_photos: count(photos.id),
      sum_of_storage_used: sum(photos.bytes)
    }
end
```

I'm sure this doesn't make a lot of sense in isolation, but I have code that will, given the results of this query, and requested filters/sorts, find the subquery that selected the field requested and adds the necessary expressions to further filter and sort the query based on that (I know this is bad, but we heavily test all of this, and I'm actively looking to use new features to avoid it). This, however, doesn't perform as well as it would if the `base` that we pass into each subquery was a CTE instead. Right now it performs well, and queries all have very predictable performance, since this pattern just scales linearly with the size of the base. If we had a CTE, it would perform even better because we could use the base query results everywhere and only run it once.

zachdaniel:
I will also say that this design went through *a lot* of iterations. Directly joining to the query caused small changes in the method behind adding one kind of metadata to drastically effect the performance of the query, and it was also very hard to predict how a change would affect the performance of that metadata generation, because if you have 10-15 joins all sibling to each other, we found it to be very fragile.

josevalim:
Right, it feels like named bindings will make a bunch of this less convoluted too, including allowing you to refer to something you have built earlier. For now, I would say that unless you want to use a UPDATE/DELETE, a recursive CTE, or filter a subquery in two different ways, the subquery and upcoming named bindings should be enough.

So maybe we should put this on the shelf until we have more concrete use cases?

zachdaniel:
The example I posed above isn't reusing anything, but is really just doing it's best to emulate the performance benefits of using a CTE (bounding complex queries by some smaller scoped query result). Named bindings will help clean up a lot, but wont really allow us to change the pattern there. I've just realized that a CTE result is referenced the same as a table though, and that there could be a lot simpler way to reference them. I think just using a string as if the value was a raw sql table would work, in the same way that referencing a view result as a table would work.  Which would mean that all we would have to add support for is running CTEs before. A given query.  What if we supported that like `Repo.all(query, with: [super_users: super_users_query])`? Then we leave it up the the consumers to use it in their query as a table, and we keep them disconnected.

zachdaniel:
Honestly, implementing the thing I just mentioned would be extremely easy.  I can see how maybe providing an option like that to report would be a bad pattern, since then the query wouldn't be a standalone object.  So I think I would amend that above proposal to make it so that queries have a `with_ctes: [super_users: cte1, super_user_posts: cte2]`.  Then it is self contained, and they are referenced simply with strings still.

josevalim:
> The example I posed above isn't reusing anything, but is really just doing it's best to emulate the performance benefits of using a CTE (bounding complex queries by some smaller scoped query result). 

So please provide an example of how you would like to rewrite that code and how that would look like with CTE?  You can even show me two SQL queries, one without CTE and another with CTE and describe how one is more performant than the other.

We need to implement a general solution that works for multiple people and the first step here is to understand the use cases before jumping to solutions. If we don't have use cases now, then let's start this discussion later.

zachdaniel:
I'll get an example going soon :)

zachdaniel:
@josevalim fun fact! I spent a long time writing up the SQL by hand to illustrate the performance improvements of using a CTE, instead of embedding the filters in every subquery over that same table. It appears that due to the implementation of CTEs in postgres, they are Optimization Fences, meaning that the planner does not bring optimizations in/out of a CTE, causing our use case to *dramatically* slow down when implemented with a CTE. So this means that the postgres planner is adapting appropriately to the fact that we have duplicate selects/filters in each subquery. So it appears we've found the optimal way to achieve the modular, arbitrarily sortable/filterable metadata that we need, and that I was incorrect on the value of CTEs for our use case. I appreciate the request for concrete evidence to back up the idea that CTEs have better performance. Additionally, now that I know that CTEs are fenced in that fashion to allow inserts/updates/side effects, I think they definitely have no place in ecto, or at the very least no reason that *I* would need them in ecto.

More about CTEs and Optimization Fences https://blog.2ndquadrant.com/postgresql-ctes-are-optimization-fences/

zachdaniel:
I still plan on advocating for a few new features, but that can be done on the google group. Thanks for looking at this with me and helping me learn more about the issue :)

josevalim:
Thanks @zachdaniel!

For documentation purposes, the use cases for CTE, as far as I see, are:

1. To make writing large SQLs more manageable
2. To avoid computing a given subquery twice
3. To compute recursive data
4. To embed a UPDATE/DELETE in a regular query

The first one does not matter for Ecto because we provide composition. The remaining 2, 3 and 4 could be beneficial for Ecto. However, 3 requires `UNION ALL` queries, but those are not supported by Ecto and they would require you to write your queries as strings/fragments. While 4 should be fine within Ecto but it means we can't use the subquery structure (they are all selects).

That's why I asked for use cases, to see which one of those problems you wanted to solve. :) However, the final solution should probably support all cases above and your final `with_ctes` proposal does so. So if the topic ever comes back, I think we do have a good place to get started.

PR #2478, opened 2018-03-26
Update Ecto.Query inspect to include `on` with `assoc`

wojtekmach:


test/ecto/query/inspect_test.exs
```diff
@@ -65,6 +65,9 @@ defmodule Ecto.Query.InspectTest do
   end
 
   test "join" do
+    assert i(from(x in Post, join: y in Comment)) ==
+           ~s{from p in Inspect.Post, join: c in Inspect.Comment, on: true}
```
wojtekmach: I'm leaving `on: true` here because generated SQL has it too.

josevalim:

PR #2483, opened 2018-03-29
Paramaterizing ChangeSet Type

zachdaniel:
I often find myself wanting to be more specific in my functions when I say that a function takes a changeset. Our schemas pretty much all express a `.t()` that is usually just `@type t :: %__MODULE__{}`.

I think its nice and very informative to be able to say something like

```
@spec validate_author_permissions(Ecto.Changeset.t(Post.t())) :: Ecto.changeset.t(Post.t())
```

The only real question I have here is whether or not we should do the `| nil` in the paramaterized or non paramaterized version.


josevalim:

PR #2487, opened 2018-03-30
Support for using FILTER with aggregates

zachdaniel:
I haven't added any tests or documentation yet, since I figured there might still be some discussion about this. I also kinda winged it on the path from select expression -> builder -> SQL. I'm not sure how far away from the right answer I am there. This works for me in my manual testing, and if it doesn't look too far off, I can get started writing actual tests for it.

This was discussed here: https://groups.google.com/forum/#!topic/elixir-ecto/9APrq9XfDp0

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -248,6 +248,13 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:filter, _, [agg, filters]}, type, params_acc, vars, env) do
+    {agg, params_acc} = escape(agg, type, params_acc, vars, env)
+    {filters, params_acc} = escape(filters, type, params_acc, vars, env)
+    expr = {:{}, [], [:filter, [], [agg, filters]]}
+    {expr, params_acc}
+  end
+
   def escape({:=, _, _} = expr, _type, _params_acc, _vars, _env) do
     error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
             "The match operator is not supported: `=`. " <>
@@ -790,6 +797,7 @@ defmodule Ecto.Query.Builder do
   def quoted_type({agg, _, [expr]}, vars) when agg in [:max, :min, :sum] do
     quoted_type(expr, vars)
   end
+  def quoted_type({:filter, _, [aggregate_type, _]}, vars), do: quoted_type(aggregate_type, vars)
 
   # Literals
   def quoted_type(literal, _vars) when is_float(literal),   do: :float
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -248,13 +248,6 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
-  def escape({:filter, _, [agg, filters]}, type, params_acc, vars, env) do
-    {agg, params_acc} = escape(agg, type, params_acc, vars, env)
-    {filters, params_acc} = escape(filters, type, params_acc, vars, env)
-    expr = {:{}, [], [:filter, [], [agg, filters]]}
-    {expr, params_acc}
-  end
-
   def escape({:=, _, _} = expr, _type, _params_acc, _vars, _env) do
     error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
             "The match operator is not supported: `=`. " <>
@@ -797,7 +790,6 @@ defmodule Ecto.Query.Builder do
   def quoted_type({agg, _, [expr]}, vars) when agg in [:max, :min, :sum] do
     quoted_type(expr, vars)
   end
-  def quoted_type({:filter, _, [aggregate_type, _]}, vars), do: quoted_type(aggregate_type, vars)
 
   # Literals
   def quoted_type(literal, _vars) when is_float(literal),   do: :float
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -248,6 +248,15 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:filter, _, [aggregate]}, type, params_acc, vars, env) do
+    escape(aggregate, type, params_acc, vars, env)
+  end
+  def escape({:filter, _, [aggregate | filter_expr]}, type, params_acc, vars, env) do
+    {aggregate, params_acc} = escape(aggregate, type, params_acc, vars, env)
+    {filter_expr, params_acc} = escape(filter_expr, type, params_acc, vars, env)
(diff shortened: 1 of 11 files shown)


josevalim:
I think this is almost all you need to do. There is a type-based concern I will comment on soon. For the tests, adding simple cases to test/ecto/adapters/postgres and checking that mysql raises for similar construct is enough.

lib/ecto/query/builder/select.ex
```diff
@@ -58,6 +58,16 @@ defmodule Ecto.Query.Builder.Select do
     {{:{}, [], [:%{}, [], [{:{}, [], [:|, [], [data, pairs]]}]]}, params_take}
   end
 
+  # filter
```
josevalim: Let's move this to the builder. The reason is that this construct is also allowed on UPDATEs according to Postgres documentation. So let's allow everything on the Ecto side. If the user writes an invalid query, they can raise.

josevalim:
You will also need to add a clause around here that handles filters:

https://github.com/elixir-ecto/ecto/blob/8d5fd3857ab16bb0521d172044ec1f343900a347/lib/ecto/query/planner.ex#L954

You can probably find a bug that will show the need for such clause by adding some tests to integration_test/cases/types.exs that uses the new filter. The test will show that Ecto is not properly casting the result.

That's because we need to teach Ecto that only the first argument of filter matters for the type computation.

lib/ecto/query/builder/select.ex
zachdaniel: :+1:

zachdaniel:
@josevalim I will add that to the planner :) I'll see if I can write some failing tests w/o it first, so I can be confident that I covered that behavior.

zachdaniel:
Test failed, like you said :)

zachdaniel:
@josevalim I feel like just the documentation in `Query.Api` isn't enough, but I'm not sure where else to put it. Any thoughts?

zachdaniel:
@josevalim interesting. I didn't realize it, but support for this feature was only added in postgres 9.4. Is there some way to annotate this in the tests, or does that rule out this feature?

zachdaniel:
Found the spot where I can configure it in test_helper.ex

zachdaniel:
Alright, ready to go!

lib/ecto/query/builder.ex
```diff
@@ -248,6 +248,15 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:filter, _, [aggregate]}, type, params_acc, vars, env) do
+    escape(aggregate, type, params_acc, vars, env)
+  end
+  def escape({:filter, _, [aggregate | filter_expr]}, type, params_acc, vars, env) do
```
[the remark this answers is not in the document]
josevalim: Agreed!

lib/ecto/query/planner.ex
```diff
@@ -938,17 +938,14 @@ defmodule Ecto.Query.Planner do
...
-      case agg do
-        :count -> :integer
-        :avg -> :any
-        :sum -> :any
-        _ -> source_type!(:select, query, select, ix, field)
-      end
+    {{:value, aggregate_type(agg, field, query)}, [expr | fields], from}
+  end
+
+  defp collect_fields({:filter, _, [{agg, _, [field | _]}, _]} = expr, fields, from, query, _take)
```
josevalim: I would prefer to make it recursive so it always works regardless of changes elsewhere. Something like:

    defp collect_fields({:filter, _, [call, _]} = expr, fields, from, query, take) do
      {type, _, _} = collect_fields(call, fields, from, query, take)
      {type, [expr | fields], from}
    end

test/ecto/adapters/mysql_test.exs
```diff
@@ -166,6 +166,13 @@ defmodule Ecto.Adapters.MySQLTest do
     assert all(query) == ~s{SELECT TRUE FROM `schema` AS s0 LIMIT 3 OFFSET 5}
   end
 
+  test "aggregate filters" do
+    query = Schema |> select([r], count(r.x) |> filter(r.x > 10)) |> normalize
+    assert_raise Ecto.QueryError, fn ->
```
josevalim: Let's do a quick assertion on the message too.

test/ecto/adapters/postgres_test.exs
```diff
@@ -116,6 +116,15 @@ defmodule Ecto.Adapters.PostgresTest do
     assert all(query) == ~s{SELECT count(DISTINCT s0."x") FROM "schema" AS s0}
   end
 
+  @tag :aggregate_filters
```
josevalim: We don't need to tag here because this is a unit test. It will always work regardless of version as the database is not involved.

josevalim: And we need a similar test for MySQL, except it will raise. :)

lib/ecto/query/builder.ex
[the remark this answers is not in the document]
zachdaniel: Yeah, that makes sense :)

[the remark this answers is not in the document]
zachdaniel: Should I do anything other than only having a pattern for `{:filter, _, [aggregate, expr]}`? Like should I do any custom error messaging if extra arguments are passed in, or is that already handled?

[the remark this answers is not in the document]
zachdaniel: Looks like the error message for having too many arguments is fine already (not a valid query expression), so I will just only match on a two element list for args.

lib/ecto/query/planner.ex
zachdaniel: Yeah, much simpler. I'll replace the original `collect_fields` for aggregates as well, so there will be less change.

test/ecto/adapters/mysql_test.exs
zachdaniel: Hmm... after making the change to make it not use `varargs` anymore, the mysql adapter no longer throws any errors during planning. I assume that it would end up error-ing in the db w/ something about the filter function not existing. Should I add some manual intervention there in the mysql adapter to prevent filter from being used?

josevalim: Yes, you need to check it in the MySQL adapter and raise accordingly. You should find examples there.

zachdaniel:
Alright, added custom error messaging in the mysql adapter for the error message, simplified the implementation, and removed the varargs aspect of `filter`, so the examples just use `and` for multiple conditions. I totally agree the varargs didn't make much sense, I think I was just learning how to implement it, and thought to myself "what if there are more things in the list of arguments", and then the shape of that just kinda snuck into my work.

zachdaniel:
@josevalim @michalmuskala Responded to your feedback.

zachdaniel:
Not sure whats going on with ebert. Could that be related to my changes at all?

josevalim:

PR #2489, opened 2018-04-02
Support coalesce

zachdaniel:
This PR adds support for `coalesce`. Per a discussion on a different PR, I opted not to make it variadic, even though it is in sql. It banks on the fact that `COALESCE(COALESCE(x, y), z)` is the same as `COALESCE(x, y, z)` With that in mind, it looks like:

```elixir
p.foo |> coalesce(p.bar) |> coalesce(5)
# or
coalesce(coalesce(p.foo, p.bar), 5)
```
I didn't really check if there is any kind of optimization to be done in the adapter or planner to unwrap these into a single `coalesce`. I wouldn't expect there to be any kind of performance difference in the two forms.

The only other thing I wasn't entirely sure about was determining the type for a coalesce call. It would seem to me that the best thing to do would be to take "the most specific type" of the two arguments. I wasn't really sure of a good way to do that aside from choosing whichever one is not `:any` if there is one. One thing I noticed is that the type of coalesce *could* be informed by a literal value, like
`from t in "table", select: coalece(t.field, 10)`. We could potentially return `{:value, :integer}` from `collect_fields` for that. I couldn't find any mapping of values -> type like that anywhere, so I figure it was best to leave it alone for now.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -257,6 +257,11 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:filter, [], [aggregate, filter_expr]]}, params_acc}
   end
 
+  def escape({:coalesce, _, values}, type, params_acc, vars, env) do
+    {arg, params_acc} = escape(values, type, params_acc, vars, env)
+    {{:{}, [], [:coalesce, [], arg]}, params_acc}
+  end
+
   def escape({:=, _, _} = expr, _type, _params_acc, _vars, _env) do
     error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
             "The match operator is not supported: `=`. " <>
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -257,10 +257,6 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:filter, [], [aggregate, filter_expr]]}, params_acc}
   end
 
-  def escape({:coalesce, _, values}, type, params_acc, vars, env) do
-    {arg, params_acc} = escape(values, type, params_acc, vars, env)
-    {{:{}, [], [:coalesce, [], arg]}, params_acc}
-  end
   def escape({:coalesce, _, [left, right]}, type, params_acc, vars, env) do
     {left, params_acc} = escape(left, type, params_acc, vars, env)
     {right, params_acc} = escape(right, type, params_acc, vars, env)
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -261,6 +261,11 @@ defmodule Ecto.Query.Builder do
     {arg, params_acc} = escape(values, type, params_acc, vars, env)
     {{:{}, [], [:coalesce, [], arg]}, params_acc}
   end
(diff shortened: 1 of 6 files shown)


lib/ecto/query/planner.ex
```diff
@@ -952,6 +952,20 @@ defmodule Ecto.Query.Planner do
     {{:value, type}, [expr | fields], from}
   end
 
+  defp collect_fields({:coalesce, _, [left, right]} = expr, fields, from, query, take) do
+    {left_type, _, _} = collect_fields(left, fields, from, query, take)
+    {right_type, _, _} = collect_fields(right, fields, from, query, take)
+
+    type =
+      case {left_type, right_type} do
+        {_, {:value, type}} when type != :any -> {:value, type}
```
josevalim: Can we coalesce two different types?

In any case, I think this type could be written in two clauses?

    {{:value, :any}, type} -> type
    {type, _} -> type

josevalim:
@zachdaniel I think we got some conflicts from the other PR too. :)

lib/ecto/query/planner.ex
zachdaniel: @josevalim I think the reason that we can't do that is because one of the types could be a value literal, like just `5` (from `coalesce(p.foo, 5)`), and I noticed that if I return that from this function that it will just use that value in all cases which would definitely be very wrong.. So its important here to only use the type returned by one of them *only if* `{:value, type}`. Of course correct me if I'm wrong in that, but that is what I noticed.

zachdaniel:
Conflicts should be easy enough to resolve!

lib/ecto/query/planner.ex
josevalim: I see! So let's add an assertion on this scenario on coalesce type as well.

zachdaniel: so asserting that in the case that the value is non-null and we are coalescing w/ a provided value `coalesce(p.some_not_null_field, 5)` that the value returned is not 5, but the value of the field?

zachdaniel: I also did not answer your question. In postgres, at least, you cannot coalesce values of different types. I have not quite done enough research to confirm this, but it appears to be dynamically defined for all types (or something along those lines) such that calling it with different types fails.

josevalim: > so asserting that in the case that the field is non-null and we are coalescing w/ a provided value coalesce(p.some_not_null_field, 5) that the value returned is not 5, but the value of the field?

:+1:

zachdaniel: Added the corresponding test.

lib/ecto/query/builder.ex
```diff
@@ -257,6 +257,16 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:filter, [], [aggregate, filter_expr]]}, params_acc}
   end
 
+  def escape({:coalesce, _, values}, type, params_acc, vars, env) do
```
josevalim: Why do we have this clause?

josevalim: If it is for varargs, let's not add it now as not all of us agree with it.

zachdaniel: It was accidentally left in :/

josevalim:

PR #2490, opened 2018-04-04
Decouple source and prefix in metadata

josevalim:


lib/ecto/schema.ex
```diff
@@ -346,12 +346,12 @@ defmodule Ecto.Schema do
...
-    collection) where the struct is or should be persisted. It is represented
-    as a tuple consisting of two fields:
+    The `:source` tracks the (table or collection) where the struct is or should
+    be persisted to.
 
-      * source - tracking the table or collection name.
-      * prefix - tracking the ecto prefix of the table or collection.
+    ## Prefix
+
+    Tracks the struct prefix in the data storage.
```
michalmuskala: Would "Tracks the source's prefix in the data storage" be more appropriate?

josevalim: :+1:

josevalim:
@michalmuskala just simplifying. :)

PR #2494, opened 2018-04-05
Add ClickHouse adapter to README

santaux:
Hi! ✋We created Ecto adapter for ClickHouse database (https://clickhouse.yandex/) some days ago. It works and we use it in our projects. It would be nice if it were in the adapters list in REAME.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,14 +64,15 @@ Also checkout the ["What's new in Ecto 2.1"](http://pages.plataformatec.com.br/e
 
 You need to add both Ecto and the database adapter as a dependency to your `mix.exs` file. The supported databases and their adapters are:
 
-Database   | Ecto Adapter           | Dependency                   | Ecto 2.0 compatible?
-:----------| :--------------------- | :----------------------------| :-------------------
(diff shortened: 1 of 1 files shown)


michalmuskala:
Before adding the adapter to the list, we generally require it to pass at least some tests from the integration suite that is distributed with ecto itself https://github.com/elixir-ecto/ecto/tree/master/integration_test. This allows us to make sure that the features that are supported by the particular adapter work the same across all adapters.

If there are some features that the database does not support and the tests are not tagged properly, we're open to adding necessary tags to the integration suite.

santaux:
Ok, thank you for the answer. Yes, this database doesn't support a lot of things (it sounds strange, but it doesn't support data updating or deleting). But we'll try to do our best 👍

AndrewDryga:
@santaux there is a lot of tags in integration tests, just start with simple ones and create a PR if you need an additional tag.

josevalim:
Hi @santaux! I will go ahead and close this but let us know when you have the test suites integrated or if you have any questions!

PR #2497, opened 2018-04-06
Add `FromExpr` and use it instead of a tuple in `Ecto.Query.from`

zoldar:
Refs #2389 

This PR introduces `Ecto.Query.FromExpr` which replaces `{source, schema}` tuple for holding query source data in the `from` field of `Ecto.Query`. The user facing API still accepts the from expression as `{source, schema}`, although it's internally converted to the new dedicated struct. That is at least my intent - I could have missed something, though as far as test coverage is concerned, it seems to be working (and again, I could have incidentally mangled tests while refactoring).

Notes about particular changes will follow in separate comments.

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -527,7 +527,7 @@ if Code.ensure_loaded?(Mariaex) do
     defp create_names(prefix, sources, pos, limit) when pos < limit do
       current =
         case elem(sources, pos) do
-          {table, schema} ->
+          %Ecto.Query.FromExpr{source: table, schema: schema} ->
             name = [create_alias(table) | Integer.to_string(pos)]
             {quote_table(prefix, table), name, schema}
           {:fragment, _, _} ->
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -585,7 +585,7 @@ if Code.ensure_loaded?(Postgrex) do
     defp create_names(prefix, sources, pos, limit) when pos < limit do
       current =
         case elem(sources, pos) do
-          {table, schema} ->
+          %Ecto.Query.FromExpr{source: table, schema: schema} ->
             name = [create_alias(table) | Integer.to_string(pos)]
             {quote_table(prefix, table), name, schema}
           {:fragment, _, _} ->
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -260,7 +260,7 @@ defmodule Ecto.Adapters.SQL do
   end
 
   defp put_source(opts, %{sources: sources}) when tuple_size(elem(sources, 0)) == 2 do
-    {source, _} = elem(sources, 0)
+    %Ecto.Query.FromExpr{source: source} = elem(sources, 0)
     Keyword.put(opts, :source, source)
   end
   defp put_source(opts, _) do
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -39,7 +39,7 @@ defmodule Ecto.Association do
                field: atom,
                unique: boolean}
 
-  alias Ecto.Query.{BooleanExpr, QueryExpr}
+  alias Ecto.Query.{BooleanExpr, QueryExpr, FromExpr}
 
   @doc """
   Builds the association struct.
@@ -175,7 +175,7 @@ defmodule Ecto.Association do
   end
 
   def assoc_query(refl, t, query, values) do
-    query = query || %Ecto.Query{from: {"join expression", nil}, prefix: refl.queryable.__schema__(:prefix)}
+    query = query || %Ecto.Query{from: %FromExpr{source: "join expression"}, prefix: refl.queryable.__schema__(:prefix)}
 
     # Find the position for upcoming joins
     position = length(query.joins) + 1
@@ -229,7 +229,7 @@ defmodule Ecto.Association do
     |> Map.put(:op, :and)
   end
 
-  defp merge_from({"join expression", _}, assoc_source), do: assoc_source
+  defp merge_from(%FromExpr{source: "join expression"}, assoc_source), do: assoc_source
   defp merge_from(from, _assoc_source), do: from
 
   # Rewrite all later joins
@@ -262,7 +262,7 @@ defmodule Ecto.Association do
 
   ## Examples
 
-      iex> Ecto.Association.related_from_query({"custom_source", Schema}, :comments_v1)
+      iex> Ecto.Association.related_from_query(%Ecto.Query.FromExpr{source: "custom_source", schema: Schema}, :comments_v1)
       Schema
 
       iex> Ecto.Association.related_from_query(Schema, :comments_v1)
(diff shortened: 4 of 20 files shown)


lib/ecto/association.ex
```diff
@@ -272,8 +272,8 @@ defmodule Ecto.Association do
       ** (ArgumentError) association :comments_v1 queryable must be a schema, a {source, schema}, or a query. got: "wrong"
   """
   def related_from_query(atom, _name) when is_atom(atom), do: atom
-  def related_from_query({source, schema}, _name) when is_binary(source) and is_atom(schema), do: schema
-  def related_from_query(%Ecto.Query{from: {source, schema}} = query, name) when is_binary(source) and is_atom(schema) do
+  def related_from_query(%FromExpr{source: source, schema: schema}, _name) when is_binary(source) and is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{from: %FromExpr{source: source, schema: schema}} = query, name) when is_binary(source) and is_atom(schema) do
```
zoldar: Regarding error message at https://github.com/elixir-ecto/ecto/pull/2497/files#diff-f547e032c73d51297b726338f16756c5R284 - I have left it unchanged b/c I assume it's a hint for the end user who will input and see external representation of `from`, which still is `{source, schema}`.

lib/ecto/query/inspect.ex
```diff
@@ -70,11 +70,11 @@ defimpl Inspect, for: Ecto.Query do
...
-  defp unbound_from(nil),           do: "query"
-  defp unbound_from({source, nil}), do: inspect source
-  defp unbound_from({nil, schema}),  do: inspect schema
-  defp unbound_from(from = {source, schema}) do
-    inspect if source == schema.__schema__(:source), do: schema, else: from
+  defp unbound_from(nil), do: "query"
+  defp unbound_from(%FromExpr{source: source, schema: nil}), do: inspect source
+  defp unbound_from(%FromExpr{source: nil, schema: schema}),  do: inspect schema
+  defp unbound_from(%FromExpr{source: source, schema: schema}) do
+    inspect if source == schema.__schema__(:source), do: schema, else: {source, schema}
```
zoldar: Outputting `{source, schema}` here, at it is still external representation of `from`

lib/ecto/query/planner.ex
```diff
@@ -576,9 +577,9 @@ defmodule Ecto.Query.Planner do
   defp prepend_if(cache, true, prepend), do: prepend ++ cache
   defp prepend_if(cache, false, _prepend), do: cache
 
-  defp source_cache({_, nil} = source, params),
-    do: {source, params}
-  defp source_cache({bin, schema}, params),
+  defp source_cache(%FromExpr{source: source, schema: nil}, params),
+    do: {{source, nil}, params}
```
zoldar: I'm unsure if I got that part right.

lib/ecto/queryable.ex
```diff
@@ -42,8 +42,19 @@ end
 
 defimpl Ecto.Queryable, for: Tuple do
```
zoldar: Still accepting tuple here, which is then converted to `FromExpr`.

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -527,7 +527,7 @@ if Code.ensure_loaded?(Mariaex) do
     defp create_names(prefix, sources, pos, limit) when pos < limit do
       current =
         case elem(sources, pos) do
-          {table, schema} ->
+          %Ecto.Query.FromExpr{source: table, schema: schema} ->
```
josevalim: I think this is wrong. When we first create the sources, which I believe is in the planner, we should go ahead and store only the `{table, schema}` there. Making it consistent with Joins and others.

josevalim: Think as the sources tuple as containing already unstructured source expressions.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -585,7 +585,7 @@ if Code.ensure_loaded?(Postgrex) do
     defp create_names(prefix, sources, pos, limit) when pos < limit do
       current =
         case elem(sources, pos) do
-          {table, schema} ->
+          %Ecto.Query.FromExpr{source: table, schema: schema} ->
```
josevalim: Same as above!

lib/ecto/adapters/sql.ex
```diff
@@ -260,7 +260,7 @@ defmodule Ecto.Adapters.SQL do
   end
 
   defp put_source(opts, %{sources: sources}) when tuple_size(elem(sources, 0)) == 2 do
-    {source, _} = elem(sources, 0)
+    %Ecto.Query.FromExpr{source: source} = elem(sources, 0)
```
josevalim: Same here.

lib/ecto/query/planner.ex
```diff
@@ -243,10 +243,10 @@ defmodule Ecto.Query.Planner do
     end
   end
 
-  defp prepare_source(_query, {nil, schema}, _adapter) when is_atom(schema) and schema != nil,
-    do: {schema.__schema__(:source), schema}
-  defp prepare_source(_query, {source, schema}, _adapter) when is_binary(source) and is_atom(schema),
-    do: {source, schema}
+  defp prepare_source(_query, %FromExpr{source: nil, schema: schema}, _adapter) when is_atom(schema) and schema != nil,
```
josevalim: I wonder when this scenario would happen. Wouldn't it be better to always guarantee that the source is already there when we give a schema? I think to do this we would need to change the builder/from to always include the source too.

lib/ecto/queryable.ex
```diff
@@ -42,8 +42,19 @@ end
...
   def to_query({source, %Ecto.Query{from: {_, schema}} = query}) when is_binary(source),
```
josevalim: I actually did not even rmember this syntax being allowed. I will investigate. :D

lib/ecto/queryable.ex
```diff
@@ -42,8 +42,19 @@ end
...
 
-  def to_query({source, schema} = from) when is_binary(source) and is_atom(schema) and not is_nil(schema),
+  def to_query({source, %Ecto.Query{from: %Ecto.Query.FromExpr{schema: schema}} = query}) when is_binary(source),
+    do: %{query | from: %Ecto.Query.FromExpr{source: source, schema: schema}}
+
+  def to_query({source, schema}) when is_binary(source) and is_atom(schema) and not is_nil(schema),
+    do: %Ecto.Query{from: %Ecto.Query.FromExpr{source: source, schema: schema}, prefix: schema.__schema__(:prefix)}
+end
+
+defimpl Ecto.Queryable, for: Ecto.Query.FromExpr do
```
josevalim: This one shouldn't be necessary. It means we are leaking FromExpr in a place where we should not.

josevalim:
Hi @zoldar!

This looks great. However, I do think you have converted too many of the tuples to FromExpr :D

At some places we likely still want to work with the tuple internally but the query representation should be the most complete one. I will take a look at those points myself as they are not super clear to me either. :heart:

josevalim:
Ok, after further review, my original assessment is correct. We already only to convert `from` into FromExpr. All from expressions were previously `{source, schema}` but not all `{source, schema}` is a FromExpr. For example, we have `{source, schema}` inside `JoinExpr` and that should not be converted to a `FromExpr`. :) I am pushing some changes soon. :)

josevalim:
@zoldar I have pushed your commit to master, thank you!

The change I did was to have only a `:source` in `from`, this is the same `:source` as in JoinExpr which may either be `{source, schema}` or a subquery. This keeps it inline with my comments above.

Everything else was perfect, great job! :)

zoldar:
@josevalim Yeah, sorry for winging that. I had been so fed up wrestling with it for the past month on and off that the moment I've made it "work", with a bit of a cleanup, I just wanted put it out to lift that weight off my shoulders.

josevalim:
No problem at all! This touches many parts of Ecto and some of those parts
have probably only been touched by me in the last 2 years or so. So I am
really glad you are also getting familiar with it. Once your PR was done,
it was easy to amend the remaining spots.

And good call on pushing it unpolished. It is indeed better to get things
moving and this way we can help each other. :)
-- 

Issue #2496, opened 2018-04-06
Ecto.Query.API.type/2 does not compiles in order_by macro


Eiji7:
### Environment

```bash
$ asdf current
elixir         1.6.4-otp-20
erlang         20.3.2
postgres       10.3

$ mix deps
# …
* postgrex 0.13.5 (Hex package) (mix)
  locked at 0.13.5 (postgrex) 3d931aba
  ok
* ecto 2.2.8 (Hex package) (mix)
  locked at 2.2.8 (ecto) a4463c09
  ok
# …
```
Funtoo Linux current x86-64bit on intel64-ivybridge

### My situation

I already have created my own query builder (from parameters send to API).

Everything is working, but we found issue that our API does not sort well fields in embedded schema.
I found that it's because we do not store some types in json column (which is normal in this format) and sorting its values causes sorting based on stored string and not on real values (like numbers).

To solve that I decided to dynamically find embedded field type and cast it. All variables are already prepared, but it fails when compiling project.

I tried to create a simple query in console and found that `order_by` does not allow calling: `&Ecto.Query.API.type/2`.

### Current behavior

Same code in `select` works, but it does not work in `order_by`.

### Example

```
import Ecto.Query

# this works:
Repo.all from m in Model, select: type(fragment("(?->>?)", field(m, :field_name), "json_key_name"), :type)

# this one also works
Repo.all from m in Model, order_by: fragment("(?->>?)", field(m, :field_name), "json_key_name")

# but this not
Repo.all from m in Model, order_by: type(fragment("(?->>?)", field(m, :field_name), "json_key_name"), :type)
```

### Error

```
== Compilation error in file path/to/file.extension ==
** (Ecto.Query.CompileError) `type(fragment("(?->>?)", field(query, ^field_name), ^json_key_name), ^cast_type)` is not a valid query expression.

* If you intended to call a database function, please check the documentation
  for Ecto.Query to see the supported database expressions

* If you intended to call an Elixir function or introduce a value,
  you need to explicitly interpolate it with ^

    expanding macro: Ecto.Query.order_by/3
    path/to/file.extension:line: Called.Project.function/arity
    (elixir) lib/kernel/parallel_compiler.ex:198: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6
```

josevalim:
The reason why `type/2` works with fragments in select is because we are using type to customize the output value. In the order by case, there is no output value, so type/2 doesn't make much sense there.

[...]

PR #2498, opened 2018-04-07
Port select-only type validation logic to builder

Eiji7:
solves #2496

@josevalim I was not sure how to write tests well, so I created scenario for my case as described in issue, so I'm casting type of json value in `order_by` macro for both adapters.

Hope I did not missed anything.

### DIFF
--- a/test/ecto/adapters/mysql_test.exs
+++ b/test/ecto/adapters/mysql_test.exs
@@ -231,11 +231,9 @@ defmodule Ecto.Adapters.MySQLTest do
         order_by: type(fragment("(?->>?)", field(e, ^:map), ^"value"), ^:decimal),
         select: true)
 
-    result =
-      "SELECT TRUE FROM `schema3` AS s0 " <>
-      "ORDER BY (s0.`map`->>?) + 0"
+    result = "SELECT TRUE FROM `schema3` AS s0 ORDER BY (s0.`map`->>?) + 0"
 
-    assert all(query) == String.trim(result)
+    assert all(query) == result
   end
 
   test "fragments" do
--- a/test/ecto/adapters/mysql_test.exs
+++ b/test/ecto/adapters/mysql_test.exs
@@ -44,6 +44,16 @@ defmodule Ecto.Adapters.MySQLTest do
     end
   end
 
+  defmodule Schema4 do
+    use Ecto.Schema
+
+    schema "schema4" do
+      embeds_one :map do
+        field :value, :decimal
+      end
+    end
+  end
+
   defp normalize(query, operation \\ :all, counter \\ 0) do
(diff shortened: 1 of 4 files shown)


test/ecto/adapters/mysql_test.exs
```diff
@@ -44,6 +44,16 @@ defmodule Ecto.Adapters.MySQLTest do
     end
   end
 
+  defmodule Schema4 do
```
josevalim: Please don't add a new schema. Just use one of the existing ones.

test/ecto/adapters/postgres_test.exs
```diff
@@ -46,6 +46,16 @@ defmodule Ecto.Adapters.PostgresTest do
     end
   end
 
+  defmodule Schema4 do
```
josevalim: Please don't add a new schema, just use one of the existing ones.

test/ecto/adapters/postgres_test.exs
```diff
@@ -395,6 +405,19 @@ defmodule Ecto.Adapters.PostgresTest do
...
+    query =
+      normalize from(e in "schema4",
+        order_by: type(fragment("(?->>?)", field(e, ^:map), ^"value"), ^:decimal),
+        select: true)
+
+    result =
+      "SELECT TRUE FROM \"schema4\" AS s0 " <>
+      "ORDER BY (s0.\"map\"->>$1)::decimal"
+
+    assert all(query) == String.trim(result)
```
josevalim: You don't need trim here. YOu can keep everything in a single line:

    assert all(query) ==  "SELECT TRUE FROM \"schema4\" AS s0 ORDER BY (s0.\"map\"->>$1)::decimal"

test/ecto/adapters/mysql_test.exs
```diff
@@ -222,6 +232,19 @@ defmodule Ecto.Adapters.MySQLTest do
...
+    query =
+      normalize from(e in "schema4",
+        order_by: type(fragment("(?->>?)", field(e, ^:map), ^"value"), ^:decimal),
+        select: true)
+
+    result =
+      "SELECT TRUE FROM `schema4` AS s0 " <>
+      "ORDER BY (s0.`map`->>?) + 0"
+
+    assert all(query) == String.trim(result)
```
josevalim: You don't need trim here. YOu can keep everything in a single line:

    assert all(query) ==  "SELECT TRUE FROM \"schema4\" AS s0 ORDER BY (s0.\"map\"->>$1)::decimal"

josevalim:
@Eiji7 i have dropped some comments and this should be good to go!

Eiji7:
@josevalim Is it ok now?

josevalim:

PR #2499, opened 2018-04-07
Support naming query `from` with `as`

zoldar:
Refs #2389 

This PR adds support for using `as` for naming query source.

Example:

```elixir
from(p in Post, as: :post)
|> where([post: p], p.title == "hello")
```

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -47,6 +47,16 @@ defmodule Ecto.Integration.RepoTest do
     end
   end
 
+  test "fetch using named from" do
+    TestRepo.insert!(%Post{title: "hello"})
+
+    query =
+      from(p in Post, as: :post)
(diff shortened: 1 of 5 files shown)


josevalim:

josevalim:
Now we just need docs and we are done! :D

zoldar:
@josevalim I've just realized that there are cases where `from` binding expression will be silently ignored. Namely, when the referenced source is not a source and/or schema.

Example:

```elixir
query = from p in "posts", as: :post
query = from p in query, as: :post2
IO.inspect query.aliases
# %{post: 0} 
```

Shouldn't we be explicit about it and crash when the source is something different than source/schema and `as` is defined?

josevalim:
@zoldar hrm, this is a very good question. I raising when the source is already an `Ecto.Query` is a good limitation.

PR #2501, opened 2018-04-07
Fix type casting coalesce

zachdaniel:
I wasn't sure if we wanted to add a pattern there for *any* "function" that we are trying to escape, but it seems to me like it was designed to guard against escaping non-ecto-syntax functions, so I figured adding just an explicit check for coalesce would be desired. I also added a pattern to fix `filter` since that would be broken in the same way.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -80,6 +80,14 @@ defmodule Ecto.Query.Builder do
     escape_with_type(expr, type, params_acc, vars, env)
   end
 
+  def escape({:type, _, [{:coalesce, _, _} = expr, type]}, _type, params_acc, vars, env) do
+    escape_with_type(expr, type, params_acc, vars, env)
+  end
+
+  def escape({:type, _, [{:filter, _, _} = expr, type]}, _type, params_acc, vars, env) do
+    escape_with_type(expr, type, params_acc, vars, env)
+  end
+
   # fragments
   def escape({:fragment, _, [query]}, _type, params_acc, vars, env) when is_list(query) do
     {escaped, params_acc} =


.gitignore
```diff
@@ -3,3 +3,5 @@
 /doc
 /tmp
 erl_crash.dump
+
+.elixir_ls/
```
zachdaniel: this seems like it is going to end up as a pretty commonly gitignored folder, as elixir_ls gets more popular.

josevalim: Anything that is specific to an editor should be in your global gitignore.

josevalim:
@zachdaniel as I said in the issue, we take a conservative approach to supporting `type` as it exists mostly for casting parameters. Unless we have a very solid reason for adding those extensions, I don't think we should. :)

.gitignore
zachdaniel: I've been doing this for way too long not to have known about the global gitignore :D 🤦‍♂️

josevalim:
Especially because for coalesce and filter you can use type inside withthe
arguments and that is goingto propagate out.
-- 


zachdaniel:
Trying to reproduce, but I'm getting `invalid query expression` when trying to use the `type` to cast the `sum()`. I'll get back to you with a solid reproduction case.

zachdaniel:
So here is what I'm seeing.  I'm getting back `Decimal`s from `sum` (even though I'm summing an integer field), and that is getting JSON serialized as a string. NBD, so I'm selecting:
```
          filter(sum(guess.time_elapsed), question.difficulty == 3),
```

zachdaniel:
So it seemed to me like I was being forced to add it to the outer call in order to make it possible.

josevalim:
Two questions: can you confirm that without the filter it returns an
integer? And what happens if you add a type/2 around the sum/1 call inside
the filter?
-- 


zachdaniel:
This is a snippet from iex
```
iex(7)> from post in "posts", select: sum(post.points)
#Ecto.Query<from p in "posts", select: sum(p.points)>
iex(8)> from post in "posts", select: type(sum(post.points), :integer)
#Ecto.Query<from p in "posts", select: type(sum(p.points), :integer)>
iex(9)> from post in "posts", select: filter(type(sum(post.points), :integer), post.special == true)
** (Ecto.Query.CompileError) `type(sum(post.points()), :integer)` is not a valid query expression.

  for Ecto.Query to see the supported database expressions

    (ecto) expanding macro: Ecto.Query.select/3
    iex:9: (file)
    (ecto) expanding macro: Ecto.Query.from/2
    iex:9: (file)
```
So that highlights the primary issue. I'm still trying to nail down exactly why the `sum`s that I'm making are returning strings (its like a huge complicated query, joining to a bunch of subqueries and stuff).

zachdaniel:
Side note: Having a clause for `filter` or `coalesce` seems in line w/ having a clause for all of the aggregate functions.

zachdaniel:
*oops*. Sorry for wasting your time! I was on master as of *very* recently and I just didn't think anything had changed in that regard. Sorry for the waste of time! On latest master, both examples I showed for coalesce *also* work, so we are good to go in all regards. Thanks!

josevalim:
I actually just realized you needed master as of your last comments, so no worries!

PR #2512, opened 2018-04-15
Improve Multi module documentation.

soteras:
Improve documentation.

https://hexdocs.pm/ecto/Ecto.Multi.html#content

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -219,6 +219,17 @@ defmodule Ecto.Multi do
   Adds an insert operation to the multi.
 
   Accepts the same arguments and options as `c:Ecto.Repo.insert/2` does.
+
+  ## Example
+
+      iex> post = %Post{title: "first"}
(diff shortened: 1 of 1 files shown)


lib/ecto/multi.ex
```diff
@@ -219,6 +219,17 @@ defmodule Ecto.Multi do
   Adds an insert operation to the multi.
 
   Accepts the same arguments and options as `c:Ecto.Repo.insert/2` does.
+
+  ## Example
+
+      iex> post = %Post{title: "first"}
```
josevalim: This module is doctested, which means we cannot actually leverage `%Post` because there is no guarantee a Post module was defined before.

My suggestion is to remove the doctests (i.e.`iex>`). I would also remove to_list in favor of transaction. Something like:

```elixir
post = %Post{title: "first"}
multi = Ecto.Multi.new |> Ecto.Multi.insert(:insert, post) 
MyApp.Repo.transaction(multi)
```

josevalim:
Thank you @soteras! Please note that, even though we removed `iex>`, we still need to nest the examples by 4 columns to keep the code formatting in markdown. Could you please nest them?

soteras:
Sure

Issue #2507, opened 2018-04-14
Add error function to Multi


josevalim:
For example, `Multi.error(multi, :subscription, "subscription failed")`. This is a shortcut to return a failed multi to an API that expected a multi.

If an error is present, we shouldn't even start the transaction (which is similar to how the changeset functions work).

/cc @ericmj @michalmuskala

hlhr:
I would give it a try.

josevalim:
Thanks @hlhr! Someone pointed out that this function is already part of the API, so it is just a matter of guaranteeing it fails early and updating the docs. :)

hlhr:
@josevalim Yes I already saw that. Just to make sure, what do you mean with "just a matter of guaranteein"? Do you mean change the ordering of the calls or put the `:error` as the first call?

PR #2513, opened 2018-04-15
Multi.error skips transaction and directly returns error

hlhr:
Implementation of issue #2507 

If a error operation is part of the multi operations no transaction will get started and the error gets returned. The tests got changed to reflect the new behaviour.
I only changed the docs for `Multi.error`, the documentation for `Repo.transaction` could get changed too but I think that's not necessary.

Just give me a feedback if additional changes have to be made e.g. in the documentation.


josevalim:

PR #2514, opened 2018-04-16
Update and prepare for Decimal.new floats deprecation

wojtekmach:


wojtekmach:
@josevalim ok, I got a bit carrier away, fixed! 😅

PR #2517, opened 2018-04-18
Fix link to associations code samples

zachahn:
Hello!

I found a broken link while reading about Ecto associations.

I'm guessing it happened cause of a word-wrap setting and the placement of the `-`. I'm not sure if this would prevent it from happening in the future, but I can reformat this to something like the following if you'd like!

```
If you want to see the code from this guide, you can view it [at
ecto/examples/ecto_assoc on 
GitHub](https://github.com/elixir-lang/ecto/tree/master/examples/ecto_assoc).
```

Hope this tiny change helps! :p


josevalim:

PR #2518, opened 2018-04-18
Mention removal of Ecto.DataType in CHANGELOG

wojtekmach:


wojtekmach:
I believe there was no deprecation period but perhaps in this particular case that's not a big deal? We had an implementation in hexpm but it wasn't used: https://github.com/hexpm/hexpm/commit/a9004ead49cecd96042e29de18885551e6a62708.

josevalim:

Issue #2480, opened 2018-03-27
Queries in associations must be known at compile time


zachdaniel:
So I went about using this more in depth in my own codebase, and I'm realizing that there is probably a flaw in the initial implementation (sorry D:). Lets say we have two separate files: post.ex, and comment.ex

```
# post.ex
defmodule Post do
  use Ecto.Schema
  schema "posts" do
    field :active, :boolean
    has_many :active_comments, Comment.active()
  end

  def active() do
    from row in __MODULE__,
      where: row.active
  end
end

#comment.ex
defmodule Comment do
  use Ecto.Schema
  schema "comments" do
    field :active, :boolean
    has_many :posts, Post.active()
  end

  def active() do
    from row in __MODULE__,
      where: row.active
  end
end
```

For obvious reasons, this can't work. Does anyone have any ideas for this one?

zachdaniel:
This *also* doesn't work if the functions are in separate files I think, becuase `from` calls `.__schema__()`, so these would still have a transitive compile time dependency on eachother in that case.

josevalim:
I was aware of the first part of the flaw but I forgot it would also hit for regular queries. I am not sure of the proper fix right now.

zachdaniel:
@josevalim we could make the destination query a runtime concern as opposed to a compile time concern? `has_many, :posts, Post, queryable: &Post.active/0` That honestly seems pretty heavy handed, and I feel like there are tools already in use in ecto to resolve the unresolvable things at runtime. We just need to find a way to make a query used in the association do the same thing.

zachdaniel:
I think this might also make self association w/ a query impossible.

josevalim:
Your option is one way to go but we would need to use an atom and/or MFArgs tuple:

    has_many :posts, Post, queryable: :active
    has_many :posts, Post, queryable: {Post, :active, []}

Another option is to allow the user to pass the `where` clauses directly:

    has_many :posts, Post, where: [active: true]

This would allow us to automatically set active to true when using build_assoc but it would limit our filters only to comparisons unless we allow `dynamic`:

    has_many :posts, Post, where: dynamic([p], ...p.active...)

Both cases hurt composition though as you need to define the query inline instead of reusing a previou definition.

zachdaniel:
I wonder if we could do something here:
```
  defmacro has_many(name, queryable, opts \\ []) do
    related = parse_related_out_of_ast(queryable) #I'm sure this helper exists already
[...]

PR #2520, opened 2018-04-19
Remove queries as an association source, add `where` and `join_through_where` keywords

zachdaniel:
Stems from discussion in issue #2480

I encountered some troubles trying to implement this. Even just locking it down to remote function calls, `dynamic`, and keyword lists, I just can't figure out how to build the `dynamic` at run time. They can't be quoted, so it can't be done at runtime, but they can't be built dynamically (can't `apply/3` macros), wasn't able to construct it manually.

Examples
```
  has_many :posts, Post, where: [special: true] # works
  has_many :posts, Post, where: Post.special() # works
  has_many :posts, Post, where: SomeoneElse.special() # works
  belongs_to :post, Post, where: dynamic([p], p.special()) # does not work
```

The ones that work above are both tested at different spots in this PR, but documentation hasn't been changed and comprehensive tests have not been added, just because I'm not sure if this will even work. Feedback much appreciated, thanks!

Edit: I also think I'm doing something wrong by having to `Macro.escape` the mfa and then `Macro.expand` it later. It seems weird to do `Macro.expand(expr, __ENV__)` when that expression should act as if it was evaluated in the caller's context (not use any aliases that are in the current context, that kind of thing)

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -37,7 +37,8 @@ defmodule Ecto.Association do
                owner: atom,
                owner_key: atom,
                field: atom,
-               unique: boolean}
+               unique: boolean,
+               where: mfa | Macro.t() | Keyword.t()}
 
   alias Ecto.Query.{BooleanExpr, QueryExpr, FromExpr}
 
@@ -217,14 +218,14 @@ defmodule Ecto.Association do
   @doc """
   Add the default assoc query where clauses a provided query
   """
-  def combine_assoc_query(assoc_queryable, nil), do: assoc_queryable
-  def combine_assoc_query(%Ecto.Query{wheres: assoc_wheres}, provided_queryable) do
-    provided_queryable
-    |> Ecto.Queryable.to_query()
-    |> Map.update!(:wheres, &Kernel.++(&1, assoc_wheres))
+  def combine_assoc_query(%{queryable: queryable} = assoc, nil), do: combine_assoc_query(assoc, queryable)
+  def combine_assoc_query(%{where: nil}, queryable), do: queryable
+  def combine_assoc_query(%{where: []}, queryable), do: queryable
+  def combine_assoc_query(%{where: {module, function, args}} = assoc, queryable) do
+    combine_assoc_query(%{assoc | where: apply(Macro.expand(module, __ENV__), function, args)}, queryable)
   end
-  def combine_assoc_query(_, provided_queryable) do
-    provided_queryable
+  def combine_assoc_query(%{where: where}, queryable) do
+    Ecto.Query.where(queryable, _, ^where)
   end
 
   defp assoc_to_where(%{on: %QueryExpr{} = on}) do
@@ -275,24 +276,13 @@ defmodule Ecto.Association do
       Schema
 
       iex> Ecto.Association.related_from_query("wrong", :comments_v1)
-      ** (ArgumentError) association :comments_v1 queryable must be a schema, a {source, schema}, or a query. got: "wrong"
+      ** (ArgumentError) association :comments_v1 queryable must be a schema or a {source, schema}. got: "wrong"
   """
   def related_from_query(atom, _name) when is_atom(atom), do: atom
   def related_from_query({source, schema}, _name) when is_binary(source) and is_atom(schema), do: schema
-  def related_from_query(%Ecto.Query{from: %FromExpr{source: {_, schema}}} = query, name) when is_atom(schema) do
-    case query do
-      %Ecto.Query{order_bys: [], limit: nil, offset: nil, group_bys: [], joins: [],
-                  havings: [], preloads: [], assocs: [], distinct: nil, lock: nil} ->
-        schema
-      _ ->
-        raise ArgumentError,
-          "A query was provided for association #{inspect name}, but that query included a statement other" <>
-          "than a `where` clause. Queries in associations only support `where` clauses, nothing else."
-    end
-  end
   def related_from_query(queryable, name) do
-    raise ArgumentError, "association #{inspect name} queryable must be a schema, " <>
-      "a {source, schema}, or a query. got: #{inspect queryable}"
+    raise ArgumentError, "association #{inspect name} queryable must be a schema or " <>
+      "a {source, schema}. got: #{inspect queryable}"
   end
 
   @doc """
@@ -457,7 +447,7 @@ defmodule Ecto.Association.Has do
   @on_replace_opts [:raise, :mark_as_invalid, :delete, :nilify]
   @has_one_on_replace_opts @on_replace_opts ++ [:update]
   defstruct [:cardinality, :field, :owner, :related, :owner_key, :related_key, :on_cast,
-             :queryable, :on_delete, :on_replace, unique: true, defaults: [], relationship: :child]
+             :queryable, :on_delete, :on_replace, :where, unique: true, defaults: [], relationship: :child]
 
   @doc false
   def after_compile_validation(%{queryable: queryable, related_key: related_key}, env) do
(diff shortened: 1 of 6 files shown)


zachdaniel:
An alternative to the issues I was having w/ `dynamic` is *only* allowing remote function calls.

lib/ecto/association.ex
```diff
@@ -1074,25 +1080,13 @@ defmodule Ecto.Association.ManyToMany do
...
-  end
   defp validate_join_through(_, join_through) when is_atom(join_through) or is_binary(join_through) do
     :ok
   end
   defp validate_join_through(name, _join_through) do
     raise ArgumentError,
       "many_to_many #{inspect name} associations require the :join_through option to be " <>
-      "an atom (representing a schema), a string (representing a table) or a query" <>
-      "with only `where` clauses."
+      "an atom (representing a schema) or a string (representing a table)."
```
josevalim: Let's remove the trailing dot (error messages do not have trailing dot).

lib/ecto/association.ex
```diff
@@ -978,12 +971,24 @@ defmodule Ecto.Association.ManyToMany do
...
-      join: j in ^join_through, on: field(j, ^join_owner_key) == field(o, ^owner_key),
-      join: q in ^queryable, on: field(j, ^join_related_key) == field(q, ^related_key)
+      join: j in ^join_through_query, on: field(j, ^join_owner_key) == field(o, ^owner_key),
+      join: q in ^Ecto.Association.combine_assoc_query(assoc, nil), on: field(j, ^join_related_key) == field(q, ^related_key)
+  end
+
+  defp join_through_query(queryable, []), do: queryable
+  defp join_through_query(queryable, nil), do: queryable
+  defp join_through_query(queryable, {module, function, args}) do
+    join_through_query(queryable, apply(Macro.expand(module, __ENV__), function, args))
```
josevalim: We shouldn't need Macro.expand here. We need to evalute the tuple contents at compile time. See the comment on schema.

lib/ecto/association.ex
```diff
@@ -217,14 +218,14 @@ defmodule Ecto.Association do
...
-  def combine_assoc_query(assoc_queryable, nil), do: assoc_queryable
-  def combine_assoc_query(%Ecto.Query{wheres: assoc_wheres}, provided_queryable) do
-    provided_queryable
-    |> Ecto.Queryable.to_query()
-    |> Map.update!(:wheres, &Kernel.++(&1, assoc_wheres))
+  def combine_assoc_query(%{queryable: queryable} = assoc, nil), do: combine_assoc_query(assoc, queryable)
+  def combine_assoc_query(%{where: nil}, queryable), do: queryable
+  def combine_assoc_query(%{where: []}, queryable), do: queryable
+  def combine_assoc_query(%{where: {module, function, args}} = assoc, queryable) do
+    combine_assoc_query(%{assoc | where: apply(Macro.expand(module, __ENV__), function, args)}, queryable)
```
josevalim: No need for macro expand here. See the comment on schema.

lib/ecto/schema.ex
```diff
@@ -1621,6 +1629,23 @@ defmodule Ecto.Schema do
     struct
   end
 
+  @spec escape_association_where(any, Macro.env) :: any
```
josevalim: We don't need to escape_association_where. Let's just invoke the code at compile time and store the where. Basically, we can remove this code and the keyword update. If somebody wants to lazily invoke a function, then they can use the MFA-tuple approach. This should simplify the code.

For the dynamic case, I will remove the anonymous functions it has inside.

josevalim:
@zachdaniel I took a look and supporting dynamic is actually not that easy. :( So what if we support only keyword lists and the MFA tuple? The MFA tuple can then return keywords or a dynamic though.

In any case, we won't escape `where` at compile time. We will just let it evaluate. This means we can also do stuff like this:

    where = [public: true]
    has_many :comments, Comment, where: where

zachdaniel:
I'm on board. For the MFA version, should I just require that they pass in `{Module, function, args}` then? That stop sme from having to do anything funky.

zachdaniel:
Nevermind, I know what your answer will be :) we will just evaluate at compile time, and accept only keywords or MFAs. Got it :) will address sometime today.

lib/ecto/association.ex
```diff
@@ -217,14 +218,14 @@ defmodule Ecto.Association do
...
-  def combine_assoc_query(assoc_queryable, nil), do: assoc_queryable
-  def combine_assoc_query(%Ecto.Query{wheres: assoc_wheres}, provided_queryable) do
-    provided_queryable
-    |> Ecto.Queryable.to_query()
-    |> Map.update!(:wheres, &Kernel.++(&1, assoc_wheres))
+  def combine_assoc_query(%{queryable: queryable} = assoc, nil), do: combine_assoc_query(assoc, queryable)
+  def combine_assoc_query(%{where: nil}, queryable), do: queryable
+  def combine_assoc_query(%{where: []}, queryable), do: queryable
+  def combine_assoc_query(%{where: {module, function, args}} = assoc, queryable) do
+    combine_assoc_query(%{assoc | where: apply(module, function, args)}, queryable)
```
josevalim: I think we should call Ecto.Query.where with the result of `apply`. Otherwise we will accidentally support returning `nil` or another MFA from the callback.

lib/ecto/association.ex
```diff
@@ -978,12 +971,24 @@ defmodule Ecto.Association.ManyToMany do
...
-      join: j in ^join_through, on: field(j, ^join_owner_key) == field(o, ^owner_key),
-      join: q in ^queryable, on: field(j, ^join_related_key) == field(q, ^related_key)
+      join: j in ^join_through_query, on: field(j, ^join_owner_key) == field(o, ^owner_key),
+      join: q in ^Ecto.Association.combine_assoc_query(assoc, nil), on: field(j, ^join_related_key) == field(q, ^related_key)
+  end
+
+  defp join_through_query(queryable, []), do: queryable
+  defp join_through_query(queryable, nil), do: queryable
+  defp join_through_query(queryable, {module, function, args}) do
+    join_through_query(queryable, apply(module, function, args))
```
josevalim: Same here. We should avoid the recursion and just call `Ecto.Query.where`.

lib/ecto/association.ex
```diff
@@ -978,12 +971,24 @@ defmodule Ecto.Association.ManyToMany do
...
+  end
+  defp join_through_query(queryable, where) do
+    from row in queryable,
+      where: ^where
```
josevalim: Let's call Ecto.Query.where here too for consistency.

josevalim:
@zachdaniel beautiful! Some minor comments and we should be good to go!

lib/ecto/association.ex
zachdaniel: Isn't that what this eventually does? This just updates the `where` and recurses the next clause that does call `Ecto.Query.where/3`

zachdaniel: Ah, I see your other feedback to directly call where. I don't mind that at all :) Will add that hange.

josevalim: Exactly. We don't want to recurse. :)

zachdaniel:
Alright, this is updated with your feedback. I also added an `@type t` for `DynamicExpr` so that I could annotate the `where` key of associations.

lib/ecto/query.ex
```diff
@@ -337,6 +337,9 @@ defmodule Ecto.Query do
 
   defmodule DynamicExpr do
     @moduledoc false
+
+    @type t :: %__MODULE__{}
```
josevalim: Since this is private, I don't think we should expose a type. What if we add `Ecto.Query.dynamic` instead? It should also be opaque, so something like:

    @opaque dynamic :: %DynamicExpr{}

josevalim:
@zachdaniel one last comment on the type and we should be ready!

lib/ecto/query.ex
zachdaniel: Sounds good :)

zachdaniel:
@josevalim change the type as suggested. I don't like where I put it in the file, but I wasn't sure of a better place.

josevalim:

PR #2522, opened 2018-04-20
Assign alias to source at runtime when `from` can't be expanded

zoldar:
Refs #2502 

~That's another attempt at ensuring that source alias won't get silently ignored when it doesn't apply.~

~I'm not entirely sure whether this is the right way to go - it's possible I overdid it with the escape/unescape dance. Also, https://github.com/elixir-ecto/ecto/compare/master...zoldar:zoldar-crash-on-from-as-query-source-redux?expand=1#diff-2bc73ed242e9a034ae5755e404196341R126 feels a bit fishy to me (it leaks over to `From.apply` even though escape logic is applied) - I don't see a better way to deal with that ATM.~

This PR handles aliasing source in cases where `from` expression can't be expanded at compile time.

### DIFF
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -115,8 +115,15 @@ defmodule Ecto.Query.Builder.From do
   end
 
   defp maybe_apply_as(query, nil), do: query
+  defp maybe_apply_as(%{from: %{as: from_as}}, as) when not is_nil(from_as) do
+    Builder.error! "can't apply alias `#{inspect as}` - source binding has `#{inspect from_as}` alias already"
+  end
   defp maybe_apply_as(%{from: from, aliases: aliases} = query, as) do
-    %{query | aliases: Map.put(aliases, as, 0), from: %{from | as: as}}
+    if Map.has_key?(aliases, as) do
+      Builder.error! "alias `#{inspect as}` already exists"
+    else
+      %{query | aliases: Map.put(aliases, as, 0), from: %{from | as: as}}
+    end
   end
 
   defp check_binds(query, count) do
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -118,7 +118,7 @@ defmodule Ecto.Query.Builder.From do
       %{from: from, aliases: aliases} = query
       %{query | aliases: Map.put(aliases, as, 0), from: %{from | as: as}}
     else
-      Builder.error! "can't alias query expression in `from` with `#{inspect as}`"
+      query
     end
   end
 
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -118,7 +118,7 @@ defmodule Ecto.Query.Builder.From do
       %{from: from, aliases: aliases} = query
       %{query | aliases: Map.put(aliases, as, 0), from: %{from | as: as}}
     else
-      query
+      Builder.error! "can't alias query expression in `from` with `#{inspect as}`"
     end
   end
 
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -57,29 +57,29 @@ defmodule Ecto.Query.Builder.From do
 
     {query, binds} = escape(query, env)
 
-    {count_bind, quoted} =
-      case expand_from(query, env) do
-        schema when is_atom(schema) ->
-          # Get the source at runtime so no unnecessary compile time
-          # dependencies between modules are added
-          source = quote do: unquote(schema).__schema__(:source)
-          prefix = quote do: unquote(schema).__schema__(:prefix)
-          {1, query(prefix, source, schema, as)}
+    case expand_from(query, env) do
+      schema when is_atom(schema) ->
+        # Get the source at runtime so no unnecessary compile time
+        # dependencies between modules are added
+        source = quote do: unquote(schema).__schema__(:source)
+        prefix = quote do: unquote(schema).__schema__(:prefix)
+        {query(prefix, source, schema, as), binds, 1}
 
-        source when is_binary(source) ->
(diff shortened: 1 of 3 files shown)


test/ecto/query_test.exs
```diff
@@ -227,6 +227,20 @@ defmodule Ecto.QueryTest do
...
+      posts_source = "posts"
+      query = from p in posts_source, as: :post
+
+      assert %{post: 0} == query.aliases
+    end
+
+    test "assign to source fails when query provided as source" do
+      message = ~r"can't alias query expression in `from` with `:post`"
+      assert_raise Ecto.Query.CompileError, message, fn ->
+        from p in subquery("posts"), as: :post
```
josevalim: @zoldar it feels like this should be allowed, no?

I guess the big question is: what do we want to allow and what we don't want to?

zoldar: Yeah, indeed - haven't thought that one through properly, hah. Even without the check though, moving the aliasing logic to `apply` enables applying the alias where it was previously ignored, right?

josevalim:
@zoldar after some thoughts. What do you think about this:

```diff
diff --git a/lib/ecto/query/builder/from.ex b/lib/ecto/query/builder/from.ex
index da69631e..68eacdc6 100644
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -57,29 +57,30 @@ defmodule Ecto.Query.Builder.From do

     {query, binds} = escape(query, env)

-    {count_bind, quoted} =
-      case expand_from(query, env) do
-        schema when is_atom(schema) ->
-          # Get the source at runtime so no unnecessary compile time
-          # dependencies between modules are added
-          source = quote do: unquote(schema).__schema__(:source)
-          prefix = quote do: unquote(schema).__schema__(:prefix)
-          {1, query(prefix, source, schema, as)}
-
-        source when is_binary(source) ->
-          # When a binary is used, there is no schema
-          {1, query(nil, source, nil, as)}
-
-        {source, schema} when is_binary(source) and is_atom(schema) ->
-          prefix = quote do: unquote(schema).__schema__(:prefix)
-          {1, query(prefix, source, schema, as)}
-
-        other ->
-          {nil, other}
-      end
-
-    quoted = Builder.apply_query(quoted, __MODULE__, [length(binds)], env)
-    {quoted, binds, count_bind}
+    case expand_from(query, env) do
+      schema when is_atom(schema) ->
+        # Get the source at runtime so no unnecessary compile time
+        # dependencies between modules are added
+        source = quote do: unquote(schema).__schema__(:source)
+        prefix = quote do: unquote(schema).__schema__(:prefix)
+        {query(prefix, source, schema, as), binds, 1}
+
+      source when is_binary(source) ->
+        # When a binary is used, there is no schema
+        {query(nil, source, nil, as), binds, 1}
+
+      {source, schema} when is_binary(source) and is_atom(schema) ->
+        prefix = quote do: unquote(schema).__schema__(:prefix)
+        {query(prefix, source, schema, as), binds, 1}
+
+      other ->
+        quoted = quote do
+          query = unquote(query) # Unquote the query for any binding variable
+          Ecto.Query.From.apply(query, unquote(length(binds)), unquote(as))
+        end
+
+        {quoted, binds, nil}
+    end
   end

   defp query(prefix, source, schema, as) do
@@ -103,8 +104,9 @@ defmodule Ecto.Query.Builder.From do
   @doc """
   The callback applied by `build/2` to build the query.
   """
-  @spec apply(Ecto.Queryable.t, non_neg_integer) :: Ecto.Query.t
-  def apply(query, binds) do
+  @spec apply(Ecto.Queryable.t, non_neg_integer, atom) :: Ecto.Query.t
+  def apply(query, binds, as) do
+    # TODO: validate `as`
     query = Ecto.Queryable.to_query(query)
     check_binds(query, binds)
     query
```

Basically, the first three clauses we know the query exists at compile time and that the alias is valid, because we just build it, so we don't need to check it. For the remaining ones, we just always do it at runtime.

josevalim:
I have updated the gist. But basically, the thing to implement is in the TODO and it should work fully at runtime (no macro magic).

Note I haven't tested this in any way at all :P

zoldar:
@josevalim Yeah, that definitely looks better and is less invasive. There's no use in escaping/unescaping `from` when applying to other query parts than `From`.

zoldar:
@josevalim Thanks for pointers. I'll make the adjustments and see how it works out a bit later. As to whether alias indiscriminately or not - I'm not sure now. I'll try to think it over, do some poking around and perhaps dig through Postgres/MySQL docs.

zoldar:
@josevalim I have applied your version with a tiny correction or two. I have also added logic for aliasing at runtime. The alias, when defined, is assigned in virtually all cases. Referring to source was and is possible by other means, so there's not much sense limiting it for named bindings anyway.

lib/ecto/query/builder/from.ex
```diff
@@ -103,13 +103,22 @@ defmodule Ecto.Query.Builder.From do
...
+      |> Ecto.Queryable.to_query()
+      |> maybe_apply_as(as)
+
     check_binds(query, binds)
     query
   end
 
+  defp maybe_apply_as(query, nil), do: query
+  defp maybe_apply_as(%{from: from, aliases: aliases} = query, as) do
+    %{query | aliases: Map.put(aliases, as, 0), from: %{from | as: as}}
```
josevalim: We have two issues here:

  1. What happens if the alias already exists?
  2. What happens if the `from` already has an `as`?

We need to check for those two cases and raise accordingly. :)

zoldar: @josevalim Fixed in https://github.com/elixir-ecto/ecto/pull/2522/commits/77f499588a618f40a5fb212880765727a974c5db. Don't know what I was thinking.

josevalim:

PR #2523, opened 2018-04-22
Use MixProject instead of Mixfile in mix.exs files

fertapric:
This change was introduced in Elixir in https://github.com/elixir-lang/elixir/commit/c6a0edcd32359eca284ff0166d19a6dc979ee954.

The new templates in Elixir also include `MixProject` https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L291


michalmuskala:
Thank you! :heart:

PR #2529, opened 2018-04-30
test case and a fix for tagged types in join queries

zachdaniel:
Tagged types in the `on` of queries that were joined weren't getting rewritten with the correct index. The test I added was failing before my change and passing afterwards. Totally fine if this isn't the preferred way to fix it, but the behavior is at least confirmed in this PR.


zachdaniel:
I needed the field names to not line up, so I added random fields like `crazy_post` and `crazy_comment` to the schemas.

test/ecto/query/planner_test.exs
```diff
@@ -54,12 +60,18 @@ defmodule Ecto.Query.PlannerTest do
...
       has_many :comments, Ecto.Query.PlannerTest.Comment
       has_many :extra_comments, Ecto.Query.PlannerTest.Comment
       has_many :special_comments, Ecto.Query.PlannerTest.Comment, where: {Ecto.Query.PlannerTest.Comment, :special, []}
+      many_to_many :crazy_comments, Comment, join_through: CommentPost, where: {Comment, :crazy, []}
 
       many_to_many :shared_special_comments, Comment, join_through: CommentPost, where: {Comment, :special, []}, join_through_where: {CommentPost, :inactive, []}
     end
+
+    def crazy() do
+      Ecto.Query.dynamic([row], row.crazy_post == "crazy")
```
michalmuskala: Have you tried a test that does not use a literal? Something like:

```elixir
many_to_many ..., where: {Comment, :crazy, ["crazy"]}

def crazy(value) do
 Ecto.Query.dynamic([row], row.crazy_post == ^value)
end
```

I want to make sure we handle params correctly.

zachdaniel: I'll add a case for that in the morning :)

lib/ecto/query/planner.ex
```diff
@@ -446,6 +446,8 @@ defmodule Ecto.Query.Planner do
       Macro.prewalk expr, fn
         {:&, meta, [join_ix]} ->
           {:&, meta, [rewrite_ix(join_ix, ix, last_ix, source_ix, inc_ix)]}
+        expr = %Ecto.Query.Tagged{type: {type_ix, type}} ->
```
josevalim: Should we guard that `type_ix` is an integer? I am not sure that it is always an integer, especially because of types such as `{:array, :string}`.

zachdaniel: Yeah, we definitely should.

test/ecto/query/planner_test.exs
zachdaniel: So that actually doesn't handle params correctly, but also requires a bit more thought. I can't just protected against `{:^, _, ix}` and rewrite that, because the prewalk could find that for other reasons, right? It would seem to me like I'd need to protect against

zachdaniel: Actually, I think I'm looking at the wrong pattern there, so nevermind.

zachdaniel: Okay, so I was originally hesitant to guard against `{:^, _, ix}` because I was worried of what would happen if someone literally interpolated `^0`, but I'm realizing now that isn't how that would work. It would instead be set as a parameter in this same way.

zachdaniel:
@josevalim addressed your comment by protecting against non-integers

test/ecto/query/planner_test.exs
zachdaniel: @michalmuskala got it. I had to additionally rewrite parameters. I imagine there may be more rewrites that need to happen here, but I'm not sure what they are. Those fixes *could* be done in separate PRs as well, if there are any. I'm happy to do them in this one as well though.

zachdaniel:
@josevalim @michalmuskala I think this should be ready to go :)

lib/ecto/query/planner.ex
```diff
@@ -442,16 +442,23 @@ defmodule Ecto.Query.Planner do
...
+            {:&, meta, [rewrite_ix(join_ix, ix, last_ix, source_ix, inc_ix)]}
+          expr = %Ecto.Query.Tagged{type: {type_ix, type}} when is_integer(type_ix) ->
+            %{expr | type: {rewrite_ix(type_ix, ix, last_ix, source_ix, inc_ix), type}}
+          other ->
+            other
+        end
+      end)
+      |> Map.update!(:params, fn params ->
+        Enum.map(params, &rewrite_param_ix(&1, ix, last_ix, source_ix, inc_ix))
+      end)
```
michalmuskala: I'd prefer to have this written without `Map.update!`
```elixir
%{expr: expr, params: params} = on
expr = 
  ...
params =
  ...

%{join | on: %{on | expr: expr, params: params}, qual: qual ...}
```

I think it is slightly clearer what's happening.

Issue #2537, opened 2018-05-08
Fragment generates parentheses for jsonb_each in join clause


sashman:
Hello, I've posted a question on SO here: https://stackoverflow.com/questions/50177729/ecto-fragment-without-parentheses

I'm not 100% if this is intended behaviour or not?

### Environment

* Elixir version (elixir -v): `Elixir 1.6.3 (compiled with OTP 19)`
* Database and version (in docker):

```
PostgreSQL 10.1 on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18) 6.3.0 20170516, 64-bit
```
* Ecto version (mix deps):
```
* ecto 2.2.10 (Hex package) (mix)
  locked at 2.2.10 (ecto) e7366dc8
* phoenix_ecto 3.3.0 (Hex package) (mix)
  locked at 3.3.0 (phoenix_ecto) 702f6e16
```
* Database adapter and version (mix deps):
```
* postgrex 0.13.5 (Hex package) (mix)
  locked at 0.13.5 (postgrex) 3d931aba
  ok
```
* Operating system: `Ubuntu 16.04.4 LTS`

### Current behavior

As per the SO post, I'm trying to query by a field in JSON, using a wild card key. My ecto query:

```elixir
(from c in Stuff,
  join: fragment("jsonb_each(?)", c.stuff),
  where: fragment("value->>'ref' = ?", ^ref)
)
|> Repo.all
```

Creates:

```
[debug] QUERY ERROR source="stuff" db=0.3ms
SELECT ... FROM "stuff" AS c0 CROSS JOIN (jsonb_each(c0."stuff")) AS f1 WHERE (value->>'ref' = $1) ["ref_1"]
```

Results in the following run time error:

```
** (Postgrex.Error) ERROR 42601 (syntax_error): syntax error at or near ")"
```

A working version of the query:

```
SELECT * FROM "stuff" AS c0 CROSS JOIN jsonb_each(c0."stuff") AS f1 WHERE value->>'ref' = 'ref_1';
```

### Expected behavior

Exclude the surrounding parentheses in `CROSS JOIN (jsonb_each(c0."stuff")) AS f1`

PR #2538, opened 2018-05-08
Dont wrap json functons in parentheses

hubertpompecki:
Addresses https://github.com/elixir-ecto/ecto/issues/2537 by making sure fragments containing PostgreSQL json functions are not wrapped in parentheses.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -415,7 +415,7 @@ if Code.ensure_loaded?(Postgrex) do
     defp operator_to_boolean(:or), do: " OR "
 
     defp parens_for_select([first_expr | _] = expr) do
-      if String.match?(first_expr, ~r/SELECT/i) do
+      if is_binary(first_expr) and String.starts_with?(first_expr, ["SELECT", "select"]) do
         [?(, expr, ?)]
       else
         expr
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -413,6 +413,14 @@ if Code.ensure_loaded?(Postgrex) do
 
     defp operator_to_boolean(:and), do: " AND "
     defp operator_to_boolean(:or), do: " OR "
+    
+    # Don't wrap json functions in parentheses
+    defp paren_expr(expr = {:fragment, _, [{:raw, raw_expr} | _] }, sources, query) do
+      cond do
+        String.match?(raw_expr, ~r/^\s*jsonb?/i) -> [expr(expr, sources, query)]
+        true -> [?(, expr(expr, sources, query), ?)]
+      end
+    end
 
     defp paren_expr(expr, sources, query) do
       [?(, expr(expr, sources, query), ?)]
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -413,12 +413,13 @@ if Code.ensure_loaded?(Postgrex) do
 
     defp operator_to_boolean(:and), do: " AND "
     defp operator_to_boolean(:or), do: " OR "
-    
+
     # Don't wrap json functions in parentheses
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks @hubertpompecki! We shouldn't special case to json though as it may happen in other cases. My suggestion is to change this `paren_expr` [here](https://github.com/hubertpompecki/ecto/blob/93c3285169775c6508da81afded66b1b8e4f3698/lib/ecto/adapters/postgres/connection.ex#L913) to simply `expr`. Some tests may fail but we can update them to explicitly add parens.

hubertpompecki:
Cheers @josevalim - makes sense, I'll try your approach and will make changes as required.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -414,6 +414,15 @@ if Code.ensure_loaded?(Postgrex) do
     defp operator_to_boolean(:and), do: " AND "
     defp operator_to_boolean(:or), do: " OR "
 
+    # Don't wrap json functions in parentheses
+    defp paren_expr({:fragment, _, [{:raw, raw_expr} | _]} = expr, sources, query) do
+      if String.match?(raw_expr, ~r/^\s*jsonb?/i) do
```
josevalim: We are still checking for json. We should never ever check for something as specific as json. We should just change the call to `paren_expr` to `expr` in `get_source`.

hubertpompecki: This is because I have not pushed anything new since yesterday yet. I am hoping to rework the PR later this week. I am fairly new to Ecto and still getting my head around how it is structured, apologies!

josevalim: Oh, sorry. I saw a commit and thought you updated it but it was just about keeping up ebert. :)

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -414,12 +414,11 @@ if Code.ensure_loaded?(Postgrex) do
     defp operator_to_boolean(:and), do: " AND "
     defp operator_to_boolean(:or), do: " OR "
 
-    # Don't wrap json functions in parentheses
-    defp paren_expr({:fragment, _, [{:raw, raw_expr} | _]} = expr, sources, query) do
-      if String.match?(raw_expr, ~r/^\s*jsonb?/i) do
-        [expr(expr, sources, query)]
+    defp parens_for_select([first_expr | _] = expr) do
+      if String.match?(first_expr, ~r/SELECT/i) do
```
michalmuskala: Could `String.starts_with?(first_expr, "SELECT")` work here?

hubertpompecki: I am not sure, it would then become case sensitive. Also, I think it would break the resulting SQL if somebody accidentally left whitespace at the beginning of the fragment string, e.g. `fragment(" SELECT ...")`.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -414,6 +414,14 @@ if Code.ensure_loaded?(Postgrex) do
     defp operator_to_boolean(:and), do: " AND "
     defp operator_to_boolean(:or), do: " OR "
 
+    defp parens_for_select([first_expr | _] = expr) do
+      if String.match?(first_expr, ~r/SELECT/i) do
```
josevalim: Let's go with: `is_binary(first_expr) and String.starts_with?(first_expr, ["SELECT", "select"])`

That's because the first expression is not always a binary and binary matches are much faster than regexes.

lib/ecto/adapters/postgres/connection.ex
josevalim: Let's check for select and SELECT. The issue with a regex is also that the behaviour changes if you accidentally have a select in it. For example, if you have a select inside a string.

lib/ecto/adapters/postgres/connection.ex
hubertpompecki: Sounds good. I take it we are not worried about whitespace accidentally left at the beginning of the string then?

josevalim:

PR #2541, opened 2018-05-11
Prevent noop migrations directory creation on gen.migration task

simonprev:
The "migrations" directory is always printed as "just been created" and it can be confusing.
This simple validation before calling the "create_directory" function prevents the logger from printing useless information 😄 

**Before**
```shell
Generated test_ecto_migration app
* creating priv/repo/migrations
* creating priv/repo/migrations/20180511174539_create_foo.exs
```
**After**
```shell
Generated test_ecto_migration app
* creating priv/repo/migrations/20180511174539_create_foo.exs
```


josevalim:

PR #2542, opened 2018-05-11
Support named and ... bindings when using dynamic in join's :on

### LINKED ISSUE #2244: Dynamic with `...` binding in `on:` takes the last binding from previous binding rather than the join itself
### Environment

* Elixir version: Erlang/OTP 20 [erts-9.0.5] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:10] [hipe] [kernel-poll:false] Elixir 1.5.1
* Database and version: PostgreSQL 9.6.5
* Ecto version (mix deps): 2.2.4
* Database adapter and version (mix deps): postgrex 0.13.3
* Operating system: Linux (Docker)

### Issue

Demo app: https://github.com/Driftrock/ecto-dynamic-binding-in-join

```elixir
query = from c in Customer,
  where: c.name == "Test Customer"

# inline two joins - the use case is that these can be added dynamically from list of params

# the dynamic is used to allow construction of conditions from list of params (here only static for simplicity)
join_on = dynamic([c, ..., ev], ev.name == "Event 1" and ev.location == "Location 1")

# append the inner join to the query, with extra ON condition from dynamic above
query = from c in query,
  inner_join: ev in assoc(c, :events),
  on: ^join_on # the `ev` binding from dynamic will be here expanded as `c` instead of `e0`

join_on = dynamic([c, ..., ev], ev.name == "Event 2" and ev.location == "Location 2")

query = from c in query,
  inner_join: ev in assoc(c, :events),
  on: ^join_on # and here the `ev` binding will be here expanded as `e0` - ie `ev` from the previous inner_join

result = App.Repo.all(query)
```

#### Current behavior
```elixir
     ** (Ecto.QueryError) deps/ecto/lib/ecto/association.ex:509: field `location` in `join` does not exist in schema App.Customer in query:

     from c in App.Customer,
       join: e0 in App.Event,
       on: e0.customer_id == c.id and (c.name == ^"Event 1" and c.location == ^"Location 1"),
       join: e1 in App.Event,
       on: e1.customer_id == c.id and (e0.name == ^"Event 2" and e0.location == ^"Location 2"),
       where: c.name == "Test Customer",
       select: c
```

#### Expected behavior

```elixir
     from c in App.Customer,
       join: e0 in App.Event,
       on: e0.customer_id == c.id and (e0.name == ^"Event 1" and e0.location == ^"Location 1"),
       join: e1 in App.Event,
       on: e1.customer_id == c.id and (e1.name == ^"Event 2" and e1.location == ^"Location 2"),
       where: c.name == "Test Customer",
       select: c
```

### Notes

I presume that the issue here is that the `dynamic` when used inside `on:` doesn't know about the new binding from join yet, because the join is not yet part of the query.

zoldar:
Closes #2244 

This PR tries to address the issue of expanding dynamic expression as `:on` expression of join query when either named binding or `...` is used for referring to the joined relation. This approach may seem a bit hackish but I couldn't sort out any better approach to that sort of chicken-and-egg problem (also, I've spent more time trying to wrap my head around the inner workings of dynamic than I'd be willing to admit 🙈 ).

### DIFF
--- a/integration_test/cases/joins.exs
+++ b/integration_test/cases/joins.exs
@@ -89,6 +89,27 @@ defmodule Ecto.Integration.JoinsTest do
     assert [{^p2, ^c1}] = TestRepo.all(query)
   end
 
+  test "joins with dynamic in :on" do
+    p = TestRepo.insert!(%Post{title: "1"})
+    c = TestRepo.insert!(%Permalink{url: "1", post_id: p.id})
+
+    join_on = dynamic([p, ..., c], c.id == ^c.id)
+
+    query =
+      from(p in Post, join: c in Permalink, on: ^join_on)
+      |> select([p, c], {p, c})
+
+    assert [{^p, ^c}] = TestRepo.all(query)
+
+    join_on = dynamic([p, permalink: c], c.id == ^c.id)
+
+    query =
+      from(p in Post, join: c in Permalink, as: :permalink, on: ^join_on)
+      |> select([p, c], {p, c})
+
+    assert [{^p, ^c}] = TestRepo.all(query)
+  end
+
   @tag :cross_join
   test "cross joins with missing entries" do
     p1 = TestRepo.insert!(%Post{title: "1"})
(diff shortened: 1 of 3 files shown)


josevalim:
Great work! This is perfect.  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:

PR #2552, opened 2018-05-17
Document non-atom Multi name example

aspett:
The current example usages of `Ecto.Multi` do not make it obvious that the `name` you use in each operation can actually be anything. The purpose of this pull request is to add an example which illustrates how you might reduce over a collection, using a non-atom as the name of each operation.


josevalim:

PR #2554, opened 2018-05-18
Extend Ecto.Migration docs

febeling:
* precedences of `up/0` and `down/0` over `change/0`
* reversibility of plain SQL steps with `execute/2`


febeling:
Well, that was lightening fast - thanks! ⚡️💛😁

PR #2555, opened 2018-05-18
Update Ecto.Assocation.NotLoaded documentation

pwightman:
This struct is returned by all associations, not just one to one.

I may be off-base here, not sure.


michalmuskala:
Thank you! :heart:

PR #2559, opened 2018-05-23
mix ecto.load should fail on SQL errors

jeremyjh:
Submitting as requested in the mailing [list](https://groups.google.com/forum/#!topic/elixir-ecto/yiptot9z8OY). 

This changes arguments to the Postgres & MySQL CLIs such that they fail with a non-zero error when SQL errors are raised. For example, if you run `mix ecto.load` with a database that is already populated with your schema, that operation will fail after applying this change. Today, both providers silently succeed since that is the default in both command programs.

Compatibility risk is low and the change is worth it. I thought I'd mention the one possible issue I know this could cause in case someone comes looking for this later. At work we have this issue:  we have a long history of migrations, and some of it was "compressed" by removing all the migration files and doing a `mix ecto.load` in the first new migration file. e.g. we literally have:

```elixir
  def up do
    Mix.Task.run "ecto.load", ["--force"]
  end
```

In the first migration file. That sort of stuff will need to go away after this change, because `schema_migrations` will be created before loading the script, and so the script will fail trying to create `schema_migrations`. Arguably this should have been done another way to begin with, which for us will probably be simply adding `ecto.load` to our setup alias: e.g.

```elixir 
defp aliases do
    [ "ecto.setup": ["ecto.create", "ecto.load", "ecto.migrate"],
...
```

### DIFF
--- a/integration_test/pg/storage_test.exs
+++ b/integration_test/pg/storage_test.exs
@@ -89,10 +89,10 @@ defmodule Ecto.Integration.StorageTest do
   end
 
   test "structure load will fail on SQL errors" do
+    File.mkdir_p!(tmp_path())
     error_path = Path.join(tmp_path(), "error.sql")
     File.write!(error_path, "DO $$ BEGIN RAISE EXCEPTION 'failing SQL'; END $$;")
-    {:error, message} = Postgres.structure_load(tmp_path(), [dump_path: error_path] ++
-                                                            TestRepo.config())
+    {:error, message} = Postgres.structure_load(tmp_path(), [dump_path: error_path] ++ TestRepo.config())
     assert message =~ ~r/ERROR.*failing SQL/
   end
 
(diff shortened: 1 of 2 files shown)


michalmuskala:
Overall this looks good to me.

The MySQL tests are failing with `unknown option '--abort-source-on-error'` - is it some recent addition and the database used on travis is older?

jeremyjh:
I should have researched that a bit more I guess as I'm not that familiar with the MySQL ecosystem. I tested with homebrew's MariaDB distribution. It looks like this option is only available in Maria's version of `mysql` CLI, since 6.0. It is not in the manual for MySQL 8.

I can imagine doing a match for it in the output from `mysql --help` to see if its supported before using it.

josevalim:
@jeremyjh so let's add this behaviour only for Postgres for now. :)

jeremyjh:
Updated to just include Postgres. One of build versions failed with a file path error on the tmp_path not found. Is that a periodic issue?

josevalim:
@jeremyjh i pushed a fix. you need to make sure the tmp directory exists otherwise you become order dependent.

josevalim:

PR #2560, opened 2018-05-23
Updates Ecto.Multi examples.

soteras:
Refactored examples and added new examples for the new implementation.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -250,6 +250,16 @@ defmodule Ecto.Multi do
   ## Example
 
       Ecto.Multi.new()
+      |> Ecto.Multi.insert(:insert, %Post{title: "first"})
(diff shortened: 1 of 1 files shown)


michalmuskala:
In cases where we use a function I think it would be good to showcase why that is - i.e. leverage data in `changes`. Without it, I think it might be confusing why use a functions instead of a regular changeset. It might be even a good idea to add some comment explaining it.

soteras:
I updated i followed the example:

```
 Ecto.Multi.new()
      |> Ecto.Multi.insert(:post, %Post{title: "first"})
      |> Ecto.Multi.insert(:comment, fn %{post: post} ->
           Ecto.build_assoc(post, :comments)
         end)
      |> MyApp.Repo.transaction()
```

michalmuskala:
Thank you! :heart:

PR #2562, opened 2018-05-23
update LogEntry query_time typespec to include nil

mitchellhenke:
I have seen some issues with assuming that a `query_time` is an integer, and it seems as though Ecto  also understands it may be nil. I think the typespec for `query_time` should be expanded to include `nil`, like `queue_time` and `decode_time` do.

https://github.com/elixir-ecto/ecto/blob/64648910450dc1762857fd7b4bca20d358cf1c27/lib/ecto/log_entry.ex#L90-L91 covers the case where a `query_time` may be nil.

https://github.com/elixir-ecto/ecto/blob/64648910450dc1762857fd7b4bca20d358cf1c27/test/ecto/log_entry_test.exs#L8-L9 tests logging an entry with a nil `query_time`.


josevalim:

Issue #2561, opened 2018-05-23
Problem setting default value in a migration for :binary type


maxim:
### Environment

* Elixir version (elixir -v): 1.6.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 10.3
* Ecto version (mix deps): 2.2.10
* Database adapter and version (mix deps): postgrex 0.13.5
* Operating system: MacOS High Sierra 10.13.4

### Current behavior

Hi folks, thanks for amazing work. Not 100% sure if this is Ecto adapter or Postgrex issue.

Looks like migration `:default` parameter for a `:binary` doesn't allow using the `<<…>>` syntax. Given the following migration:

```elixir
defmodule MyApp.Repo.Migrations.CreateThings do
  use Ecto.Migration

  def change do
    create table(:things) do
      add(:data, :binary, default: <<0, 0>>)
    end
  end
end
```

When I try to run `mix ecto.migrate` I get the following error:

```
[info] == Running MyApp.Repo.Migrations.CreateThings.change/0 forward
[info] create table things
** (Postgrex.Error) ERROR 08P01 (protocol_violation): invalid message format
    (ecto) lib/ecto/adapters/sql.ex:200: Ecto.Adapters.SQL.query!/5
    (ecto) lib/ecto/adapters/postgres.ex:96: anonymous fn/4 in Ecto.Adapters.Postgres.execute_ddl/3
    (elixir) lib/enum.ex:1899: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto) lib/ecto/adapters/postgres.ex:96: Ecto.Adapters.Postgres.execute_ddl/3
    (ecto) lib/ecto/migration/runner.ex:104: anonymous fn/2 in Ecto.Migration.Runner.flush/0
    (elixir) lib/enum.ex:1899: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto) lib/ecto/migration/runner.ex:102: Ecto.Migration.Runner.flush/0
    (stdlib) timer.erl:181: :timer.tc/2
    (ecto) lib/ecto/migration/runner.ex:26: Ecto.Migration.Runner.run/6
    (ecto) lib/ecto/migrator.ex:128: Ecto.Migrator.attempt/6
    (ecto) lib/ecto/migrator.ex:73: anonymous fn/4 in Ecto.Migrator.do_up/4
[...]

PR #2563, opened 2018-05-23
Fix specifying binary column defaults for Postgres

maxim:
This addresses issue #2561. I think this is just an accidental omission.


maxim:
Huh, looks like my integration test fails because it's run in MySQL, but only Postgres allows defaults for binaries. Any advice? I think a test like that is important to make sure database accepts the encoded value, but can we exclude it in MySQL?

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -971,6 +972,10 @@ if Code.ensure_loaded?(Postgrex) do
       :binary.replace(value, "'", "''", [:global])
     end
 
+    defp format_binary(value) when is_binary(value) do
```
michalmuskala: We already have a guard on all uses of this function, let's skip it here.

maxim:
@michalmuskala One more thing. This might be a breaking change. If people specified default with `"\\x1234"`, or `"'\\x1234'"`, or `"'\\x1234'::bytea"` I believe it will stop working. We can have `format_binary` match `"\\x"` and format differently to support both ways, e.g.

maxim:
@michalmuskala If we were to support the \\x style, it would be inconsistent with where clause usage, where it wouldn't work. So my opinion: we either support both everywhere, or only `<<0>>` everywhere, and embrace the breaking change.

maxim:
@michalmuskala I've thought more about it, and I'd say we should just drop `\\x` style since it was a bug in the first place. However, if this is possibly a breaking change, how is it typically handled? Do you want me to add it to changelog and make it a part of major release? (I wanted to backport to 2.2 frankly but now I'm not sure).

integration_test/pg/binary_default_test.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Ecto.Integration.BinaryDefaultTest do
```
josevalim: Let's please not add a whole new module only for this test. Instead, let's add it to `integration_test/sql/migration.exs` and make sure it also passes for MySQL.

maxim: Any advice on how to make it pass for MySQL? It doesn’t support defaults for binaries.

josevalim: Oh, if it doesn't support default for binaries let's use a tag. Something like this:

    @tag :binary_default_in_migration

And then you can skip this tag in the MySQL test suite. :)

maxim:
@josevalim before people could only set a default like `"\\x1234"` but now it will be formatted into ["'\\x", encoded("\\x1234"), "'::bytea"], which would produce a different default than they intended.

josevalim:
Hrm, I see. I wonder if that would, after all, be the better solution to this problem then? We just need to point people towards the proper direction.

The issue with checking on `:binary` is that we may still have issues in other scenarios, such as custom postgres types built on top of binaries. So a general solution that doesn't require us to check `:binary` would be best.

Can we figure out which bytes postgres considers to be valid in a string? Or is this issue exclusive to null bytes? Would <<1, 1>> work without this patch? What about <<255, 255>>?

Thanks for looking into this! :heart:

maxim:
@josevalim 

> The issue with checking on :binary is that we may still have issues in other scenarios, such as custom postgres types built on top of binaries.

I just made a custom binary type, which is how I came across the bug. You would always add `def type, do: :binary` in your custom type, and cast your input values into binaries, so things will just work. It works as of today, and we aren't breaking it with this change.

>  is this issue exclusive to null bytes?

It's a very simple problem, nothing really to do with data. Ecto accepts `<<1,2,3>>` literals In `where` clauses, but not in `default` clause in a migration. I think elixir binaries are the only sane thing you should be passing into postgres binary type, so just trying to bring `default` in line with `where`.

Documentation says that a literal for `:binary` type is `<<int, int, …>>` but that's not true for setting a default.

maxim:
@josevalim Original issue #2561 has a good, brief problem description.

Also, I've updated the PR to not use a separate test file.

maxim:
I'd like to clarify 2 things:

1. This is a breaking change, but only for people who want to re-run their old migrations, and they set defaults on binary columns (rare probably).
2. If anyone actually did set a binary column default, they used an undocumented way to do that. You only come up with `"\\x123abc"` after reading postgres docs. It's a hack from Elixir's perspective, since Elixir's docs tell you to use `<<int, int, …>>`.

josevalim:
I understand the problem. :)

The table you linked talks about the query syntax, because we can prepare
queries and get their type info from the database, but not with migrations.
Not all constructs of the query syntax are allowed in migrations, actually
very few.

The issue with this PR are not Ecto custom types, but other Postgres types
that may expect binaries. For them, the issue won’t be solved and he would
have the same poor experience as you. :)
-- 


maxim:
@josevalim Ah I think I get the concern. In create table migrations we have types, but in alter table where a column default might be added/altered, we wouldn't always know that the column is :binary. So we wouldn't know whether to cast the default value into a string or bytea blob. Instead we want users to always use Postgres literals in migrations, not Elixir literals, because we don't know what those columns are in postgres.

That's kind of a bummer, it's not intuitive to go from `<<byte, byte>>` to `"\\x[HEX][HEX][HEX][HEX]"` (which is how postgres wants binary data to be passed in), and we should probably document this clearly.

Is this what you had in mind?

Thanks for bearing with me! :)

maxim:
@josevalim Got it, thanks for explaining. I honestly assumed that bytea column can ONLY be formatted as `"\\x[HEX…]"` and plain strings or binary data won't be accepted but that's not true. It accepts "foo" in the default. However, you're right that we need to dig in and figure out what we can pass.

✅  This works:

```elixir
add(:foo, :binary, default: "foo")
```

❌ This doesn't:

```elixir
add(:foo, :binary, default: <<0>>)
# ERROR 08P01 (protocol_violation): insufficient data left in message
```

✅ This works:

```elixir
add(:foo, :binary, default: <<1>>)
```

❌ This doesn't:

```elixir
add(:foo, :binary, default: <<2,213>>)
# ERROR 22021 (character_not_in_repertoire): invalid byte sequence for encoding "UTF8": 0xd5 0x27
```

josevalim:
It seems we need to check that String.valid? and that it doesn’t contain
any null byte. Also, we don’t need an integration test, I think
test/Ecto/adapters/postgres_test is enough. Maybe a separate PR so we can
compare them side by side? Thank you!
-- 


maxim:
How's this for error phrasing?

```
** (ArgumentError) unknown default `<<213>>` for type `:binary`. Invalid strings should be encoded into bytea hex format as follows: "\\xd5"
```

maxim:
Added an alternative PR #2563.

maxim:
Closing this in favor of #2567.

PR #2567, opened 2018-05-26
Raise a helpful error on invalid SQL default

maxim:
This is an alternative approach to #2563 (original issue #2561).

In this one, we don't try to encode default values, we only raise a helpful error when the value is an invalid UTF-8 or contains null bytes.

Here's what the error currently looks like:

> ** (ArgumentError) default values are interpolated as UTF-8 strings, and cannot contain null bytes. `<<102, 111, 111, 0>>` is invalid. Refer to PostgreSQL documentation for instructions on how to escape this SQL type.

@fishcakez and I have been back-and-forth in IRC about the wording of this error (and I still tweaked this version a little), but open to further suggestions.

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -826,7 +826,7 @@ if Code.ensure_loaded?(Postgrex) do
       else
         encoded = "\\x" <> Base.encode16(literal, case: :lower)
         raise(ArgumentError, "default values are interpolated as UTF-8 strings and cannot contain null bytes. " <>
-                             "`#{inspect literal}` is invalid. If you want to write it as a binary, use #{encoded}, " <>
+                             "`#{inspect literal}` is invalid. If you want to write it as a binary, use \"#{encoded}\", " <>
                              "otherwise refer to PostgreSQL documentation for instructions on how to escape this SQL type")
       end
     end
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -821,13 +821,13 @@ if Code.ensure_loaded?(Postgrex) do
       ["ARRAY[",  Enum.map(list, &default_type(&1, inner)) |> Enum.intersperse(?,), "]::", ecto_to_db(type)]
     end
     defp default_type(literal, _type) when is_binary(literal) do
-      if :binary.match(literal, <<0>>) == :nomatch && String.valid?(literal) do
+      if :binary.match(literal, <<0>>) == :nomatch and String.valid?(literal) do
         [?', escape_string(literal), ?']
       else
         encoded = "\\x" <> Base.encode16(literal, case: :lower)
-        raise(ArgumentError, "default values are interpolated as UTF-8 strings and cannot contain null bytes. " <>
+        raise ArgumentError, "default values are interpolated as UTF-8 strings and cannot contain null bytes. " <>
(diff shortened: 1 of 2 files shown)


maxim:
The string validation could be slightly optimized if we did null-byte and valid-UTF8 check in one pass, but I bet it's not as trivial to do as it might seem.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -820,7 +820,14 @@ if Code.ensure_loaded?(Postgrex) do
     defp default_type(list, {:array, inner} = type) when is_list(list) do
       ["ARRAY[",  Enum.map(list, &default_type(&1, inner)) |> Enum.intersperse(?,), "]::", ecto_to_db(type)]
     end
-    defp default_type(literal, _type) when is_binary(literal),  do: [?', escape_string(literal), ?']
+    defp default_type(literal, _type) when is_binary(literal) do
+      if :binary.match(literal, <<0>>) == :nomatch && String.valid?(literal) do
+        [?', escape_string(literal), ?']
+      else
+        raise(ArgumentError, "default values are interpolated as UTF-8 strings, and cannot contain null bytes. `#{inspect literal}` is invalid. " <>
```
josevalim: I like this a lot and I have one suggestion: what if we include the encoded string in the error message? So something like this:

```elixir
"default values are interpolated as UTF-8 strings and cannot contain null bytes. " <>
  "`#{inspect literal}` is invalid. If you want to write it as a binary, use #{inspect(encoded)}, " <>
  "otherwise refer to PostgreSQL documentation for instructions on how to escape this SQL type"
```

test/ecto/adapters/postgres_test.exs
```diff
@@ -883,6 +883,51 @@ defmodule Ecto.Adapters.PostgresTest do
...
+    assert_raise ArgumentError, ~r"UTF-8", fn ->
+      execute_ddl(create)
+    end
+  end
+
+  test "create table with binary column and null-byte-containing default" do
+    create = {:create, table(:blobs),
+              [{:add, :blob, :binary, [default: "foo" <> <<0>>]}]}
+
+    assert_raise ArgumentError, ~r"UTF-8", fn ->
```
josevalim: If we do my suggestion above, let's check the suggestion in the error message is correct. :)

josevalim:
@maxim I like this a lot. What do you think, compared to the previous PR? I dropped two comments and we should be good to go!

maxim:
@josevalim Definitely agree. I don't think Ecto should know how to produce a string version of every possible type. In my previous PR I missed the fact that we normally delegate var binding to db, I thought elixir code did it, and we forgot to do it in default interpolation. Now it's clear, string representations aren't Ecto's responsibility, so we should just point people to docs to find the right SQL literal for their default.

lib/ecto/adapters/postgres/connection.ex
josevalim: That's a good point. Given that we clearly say that "if you want to have it as a binary", we should be covered. Plus, if you send a poorly utf-8 encoded binary to a text column, I am hoping postgres will provide a better error message (error than a protocol error).

maxim:
@josevalim changes pushed, plz don't hesitate to ask for more fixes, no matter how nit-picky, I care about the little things too.

josevalim:
Tks @maxim! I removed the parens from `raise` and used `and` instead of `&&` since the operands are returning booleans. That's all!

It was a pleasure to work together on this one! If you would like to further contribute to Ecto, feel free to pick any other open issue and/or ask questions!

josevalim:

PR #2568, opened 2018-05-28
Avoid catch-all clause in Ecto.Registry

lexmag:


fishcakez:
Should we remove this clause and crash instead as we don't expect other messages?

josevalim:
We will remove `super` from Elixir v1.8 so I would keep it as is.

@fishcakez It depends if you interpret stray info messages as a bug or not. I personally don't, so I am fine with keeping it as is. :)

fishcakez:
We should then instead of being silent

josevalim:

PR #2572, opened 2018-05-30
Allow prefix option in from/join

josevalim:
Currently there are two backwards incompatible changes in this PR:

  1. Adapter backwards incompatible: Each source in the tuple of sources now have three elements instead of two since we now include the prefix. This means adapters will break when building queries but the fix is trivial and performing the fix means this feature will be supported. See the `create_names` change.

  2. User backwards incompatible: The `@schema_prefix` attribute in schemas now behaves differently but we have not decided this is the way to go yet.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -176,10 +176,14 @@ defmodule Ecto.Association do
   end
 
   def assoc_query(refl, t, query, values) do
-    query = query || %Ecto.Query{
-      from: %FromExpr{source: {"join expression", nil}},
-      prefix: refl.queryable.__schema__(:prefix)
-    }
+    query =
+      query ||
+      %Ecto.Query{
+        from: %FromExpr{
+          source: {"join expression", nil},
+          prefix: refl.queryable.__schema__(:prefix)
+        }
+      }
 
     # Find the position for upcoming joins
     position = length(query.joins) + 1
@@ -210,9 +214,15 @@ defmodule Ecto.Association do
     rewrite_ix = assoc.ix
     [assoc | joins] = Enum.map([assoc | joins], &rewrite_join(&1, rewrite_ix))
 
-    %{query | wheres: [assoc_to_where(assoc) | query.wheres], joins: joins,
-              from: merge_from(query.from, assoc.source), sources: nil}
-    |> distinct([x], true)
+    query = %{
+      query
+      | wheres: [assoc_to_where(assoc) | query.wheres],
+        joins: joins,
+        from: merge_from(query.from, assoc.source),
+        sources: nil
+    }
+
+    distinct(query, [x], true)
   end
 
   @doc """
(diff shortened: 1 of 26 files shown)


lib/ecto/association.ex
```diff
@@ -176,10 +176,14 @@ defmodule Ecto.Association do
...
-      prefix: refl.queryable.__schema__(:prefix)
-    }
+    query =
+      query ||
+      %Ecto.Query{
+        from: %FromExpr{
+          source: {"join expression", nil},
+          prefix: refl.queryable.__schema__(:prefix)
+        }
+      }
```
josevalim: User backwards incompatible!!!

lib/ecto/query/builder/from.ex
```diff
@@ -84,26 +89,28 @@ defmodule Ecto.Query.Builder.From do
...
+      from: {:%, [], [Ecto.Query.FromExpr, {:%{}, [], from_fields}]},
+      aliases: {:%{}, [], aliases}
+    ]
 
-    {:%, [], [Ecto.Query,
-              {:%{}, [],
-               [from: {:%, [], [Ecto.Query.FromExpr,
-                                {:%{}, [], [source: {source, schema}, as: as]}]},
-                prefix: prefix,
-                aliases: {:%{}, [], aliases}]}]}
```
josevalim: User backwards incompatible!!! We may need to keep the source prefix and from prefix separated.

lib/ecto/query/builder/join.ex
```diff
@@ -117,18 +117,30 @@ defmodule Ecto.Query.Builder.Join do
...
     join_params = Builder.escape_params(join_params)
 
-    qual = validate_qual(qual)
+    prefix =
+      case join_source do
+        {_, schema} when prefix == nil and schema != nil ->
+          quote(do: unquote(schema).__schema__(:prefix))
+        _ ->
+          prefix
+      end
```
josevalim: User backwards incompatible!!!

lib/ecto/queryable.ex
```diff
@@ -44,10 +44,8 @@ defimpl Ecto.Queryable, for: Atom do
 end
 
 defimpl Ecto.Queryable, for: Tuple do
-  def to_query({source, schema} = from) when is_binary(source) and is_atom(schema) and not is_nil(schema) do
-    %Ecto.Query{
-      from: %Ecto.Query.FromExpr{source: from},
-      prefix: schema.__schema__(:prefix)
-    }
```
josevalim: User backwards incompatible!!!

lib/ecto/schema.ex
```diff
@@ -515,8 +514,10 @@ defmodule Ecto.Schema do
 
         def __schema__(:query) do
           %Ecto.Query{
-            from: %Ecto.Query.FromExpr{source: {unquote(source), __MODULE__}},
-            prefix: unquote(prefix)
```
josevalim: User backwards incompatible!!!

lib/ecto/schema.ex
```diff
@@ -50,11 +50,10 @@ defmodule Ecto.Schema do
...
       which generates structs and queries without prefix. When set, the
-      prefix will be used by every built struct and on queries where the
-      current schema is used in `from` (and only `from` exclusively). If
-      a schema is used as a join or part of an assoc, `@schema_prefix` won't
-      be obeyed. In PostgreSQL, the prefix is called "SCHEMA" (typically
-      set via Postgres' `search_path`). In MySQL the prefix points to databases.
+      prefix will be used by every built struct and on queries whenever
+      the schema is used in a `from` or a `join`. In PostgreSQL, the prefix
+      is called "SCHEMA" (typically set via Postgres' `search_path`).
+      In MySQL the prefix points to databases.
```
josevalim: User backwards incompatible!!!

josevalim:

lib/ecto/association.ex
OvermindDL1: Maybe `static_prefix` for a prefix that should always apply for a given schema (unless overridden on-site)?

PR #2575, opened 2018-06-04
New function `count`

soteras:
i know there is a aggregate function, but i think it is not explicit. I found a lot of examples like this:

```
Repo.one(from p in "people", select: count(p.id))
```

It will be better if has a count function.

It is just a idea. Thanks


josevalim:
`aggregate` is fine. :) The good think about it is that it also points people towards other functionality.

PR #2577, opened 2018-06-05
Repo.exists?

### LINKED ISSUE #2454: Existence query in `Repo`
In #234 there were proposal for adding `Repo.exists?/1` function that will check if given query will result with any data.

I think that now it is good idea to reopen that question as Ecto stabilised a little, we have adapters for some NoSQLs and there is Ecto 3.0 around the corner. However mine "SQL" solution is a little bit different from the original

```elixir
  def exists?(query) do
    import Ecto.Query

    result =
      query
      |> exclude(:select)
      |> limit(1)
      |> select(1)
      |> one()

    not is_nil(result)
  end
```

mitchellhenke:
Attempts to close #2454

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -713,8 +713,8 @@ defmodule Ecto.Integration.RepoTest do
   end
 
   test "exists?" do
-    TestRepo.insert!(%Post{title: "1", text: "hai"})
-    TestRepo.insert!(%Post{title: "2", text: "hai"})
+    TestRepo.insert!(%Post{title: "1", text: "hai", visits: 2})
+    TestRepo.insert!(%Post{title: "2", text: "hai", visits: 1})
 
     query = from p in Post, where: not is_nil(p.title), limit: 2
     assert query |> TestRepo.exists? == true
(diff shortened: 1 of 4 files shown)


josevalim:
Some questions: what happens if a select is already set? What about a
limit? Also, can we have unit tests for this? Please look into
test/ecto/repo_test.exs but it may be we don’t unit test this functionality
due to the complexity. I can take a look at it later too!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
Also let’s move the integration test to a separate test. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

mitchellhenke:
oops, good catch on already having select set. needed to `exclude(:select)`. The limit should be overridden if it's already set.

I've moved the integration test to a separate test and added cases around having a `select` and `limit` already set.

I looked at existing unit tests, and `exists?` seemed most comparable to `one`, which doesn't have any. Is there something that could be checked in the unit test?

josevalim:
@mitchellhenke I will play with the integration tests a bit, I am hoping we can at least store the query it is ultimately sent to adapters and check that. We will see. :)

josevalim:
@mitchellhenke sorry for the delay but I've added some unit tests to aggregate:

https://github.com/elixir-ecto/ecto/commit/a21e56f90a73da054f000bdbfa6ebba72cb4bde6

We can use the same strategy to test exists. :)

mitchellhenke:
@josevalim no worries 🙂 

I've added some tests and updated the functionality of `exists?` to exclude other query attributes to reflect some potential issues I found in adding those tests.

josevalim:

Issue #2576, opened 2018-06-05
Custom Type:  load/1 not called for null values


namxam:
### Environment

* Elixir version (elixir -v): Elixir 1.6.4 (compiled with OTP 20)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 10.2
* Ecto version (mix deps): 2.2.8
* Database adapter and version (mix deps): postgrex, 0.13.4
* Operating system: MacOS

### Current behavior

I have defined a schema with a custom type like this:
```
schema "users" do
  field(:gender, Types.Gender)
end
```

The database column allows null values for historic reasons. For our new API we would like to convert all given values to predefined ones from an enum. A basic form of a custom type would be:

```
defmodule Auth.Types.Gender do
  @behaviour Ecto.Type
  def type, do: :string

  def cast(:not_provided), do: {:ok, nil}
  def cast(value) when is_atom(value), do: {:ok, Atom.to_string(value)}
  def cast(_), do: :error

  def load(nil), do: {:ok, :not_provided}
  def load(value), do: {:ok, String.to_atom(value)}

  def dump(value), do: {:ok, value}
end
```

When running in iex, everything seems to be working as expected:
```
iex(1)> Auth.Types.Gender.load(nil)
{:ok, :not_provided}
```

But when fetching data via `Repo.all`, all i get is `gender: nil`

### Expected behavior

I would expect that `load` would be called no matter what value the database column would have.

josevalim:
@namxam Yes, `nil` are always bypassed and never given to the type, since most types do not care about `nil` and this seems to be a corner case.

Unfortunately I am afraid we can't do this change, as it would basically break all of the existing types, as they will suddenly start receiving nils.

namxam:
Too bad. Although it is a bit counterintuitive. I guess I have to find another way to get it working for me. To help others, I created a small PR with a note in load/1 documentation. Just in case you want it.

PR #2578, opened 2018-06-05
Add small note that nil values are bypassed in custom types.

namxam:
Fixes https://github.com/elixir-ecto/ecto/issues/2576


josevalim:
Thanks for the PR @namxam! Since this applies to all operations and not only dump, maybe we should move the note [to the moduledoc](https://github.com/namxam/ecto/blob/3b1658487d65f3572f504acc4d0181c08da97292/lib/ecto/type.ex#L10) instead?

josevalim:

PR #2583, opened 2018-06-08
Run prepare_changes on embeds

### LINKED ISSUE #2545: Ecto.Changeset.prepare_changes not being triggered for embeds
### Environment

* Elixir version (elixir -v): 1.6.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): 10
* Operating system: Mac OS X

### Current behavior

https://github.com/LostKobrakai/ecto/commit/70b5ba66e8ca2ca67060535cbe486378056df7df

### Expected behavior

I'd expect that all `Ecto.Changeset.prepare_changes` functions are called no matter if it's on the "root" changeset or any nested assoc/embed changeset.

wojtekmach:
Closes #2545 

@LostKobrakai thanks for test case! I've performed minor adjustments to it.

### DIFF
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -171,9 +171,7 @@ defmodule Ecto.Embedded do
   defp action_to_auto(:insert), do: :autogenerate
   defp action_to_auto(:update), do: :autoupdate
 
-  @doc """
-  Callback invoked to build relations.
-  """
+  @impl true
   def build(%Embedded{related: related}) do
     related.__struct__
   end
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -48,33 +48,35 @@ defmodule Ecto.Embedded do
   loaded into the schema struct afterwards.
   """
   def prepare(changeset, adapter, repo_action) do
-    %{changes: changes, data: %{__struct__: schema}, types: types} = changeset
-    prepare(Map.take(changes, schema.__schema__(:embeds)), types, adapter, repo_action)
+    %{changes: changes, data: %{__struct__: schema}, types: types, repo: repo} = changeset
+    prepare(Map.take(changes, schema.__schema__(:embeds)), types, adapter, repo, repo_action)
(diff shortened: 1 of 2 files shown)


josevalim:
Beautiful! Shall we move the test to test/ecto/repo/embedded_test.exs?

LostKobrakai:
I feel like it's where it belongs, as the issue is not only releated to embeds. I mean afaik it did not run prapare_changes() callbacks of nested assocs as well before, but just the one callback of the root changeset.

josevalim:
@LostKobrakai then it requires a test in both test/ecto/repo/embedded_test and in test/ecto/repo/*assocs_test. `test/ecto/repo_test` is about the repo entry point API.

lib/ecto/embedded.ex
```diff
@@ -111,6 +113,19 @@ defmodule Ecto.Embedded do
     |> apply_embeds(struct)
   end
 
+  defp run_prepare(changeset, repo) do
```
wojtekmach: this is copy-paste from: https://github.com/wojtekmach/ecto/blob/e8fbe9ad269c0cb5d7c7f544249e537ba930f9a5/lib/ecto/repo/schema.ex#L481

In Repo.Schema.run_prepare we get `prepare` list as separate argument because we use that argument in `wrap_in_transaction` call as well. Also, in Repo.Schema changeset got repo filled in using `put_repo_and_action`.

I wasn't sure if I should make that code more similar, so I tried making the minimal change only.

josevalim: @wojtekmach we could move the `Ecto.Changeset.Relation.run_prepare(changeset, prepare)` but I don't think the gain is clear, I can see both sides of the argument. Therefore I will gladly leave this up to you. :) Your call!

wojtekmach: Let's leave this as is for now. I'm gonna read whole embedded.ex and relation.ex one of these days :)

josevalim:

Issue #2547, opened 2018-05-15
Crash when validating NaN number


lukaszsamson:
### Environment
* Ecto version (mix deps): 2.10

### Current behavior
When validating
```
MyStruct{}
|> cast(%{my_field: "NaN"})
|> validate_required([:my_field])
|> validate_number(: my_field, greater_than: 0, less_than: 100)
```
ecto crashes with
```
** (CaseClauseError) no case clause matching: #Decimal<NaN>
(decimal) lib/decimal.ex:349: Decimal.cmp/2
(ecto) lib/ecto/changeset.ex:1727: Ecto.Changeset.validate_number/6
(elixir) lib/enum.ex:2914: Enum.find_value_list/3
(ecto) lib/ecto/changeset.ex:1312: Ecto.Changeset.validate_change/3
```
### Expected behavior
'validate_required' should check if the number is not NaN (just like it checks if string is not empty)
'validate_number' should not try to compare with NaN (or use Decimal.compare/2 instead of Decimal.cmp/2 and ignore NaN results)

josevalim:
I don't think we should accept NaN. i.e. casting should fail.

I know some users may actually want NaN, but then they can leverage a custom type, especially once we add compare to the type API.

josevalim:
Btw, thanks for the report! :)

lukaszsamson:
I'm not sure if casting should fail. Some databases do support NaN values (e.g. Postgres, Oracle).
I worked around this issue with simple substitution - in my use case I can treat NaN as a missing value.
```
defp validate_not_nan(cs, field) do
    case get_field(cs, field) do
      %Decimal{coef: coef} when coef in [:sNaN, :qNaN] ->
        cs |> put_change(field, nil)
      _ ->
        cs
    end
  end

MyStruct{}
|> cast(%{my_field: "NaN"})
|> validate_not_nan(: my_field)
|> validate_required([:my_field])
|> validate_number(:my_field, greater_than: 0, less_than: 100)
```

josevalim:
It should fail out of the box because I would say most users don't actually expect NaN in there. If you want NaN, then you can opt into it with a custom type.

zachdaniel:
As long as the way NaN is handled is documented somewhere, along w/ the advice to use a custom type, I agree.

lukaszsamson:
@josevalim @wojtekmach the patch does not fix the original issue.
```
MyStruct{}
[...]

PR #2585, opened 2018-06-11
Return :error when casting invalid decimals instead of crashing

wojtekmach:
Ref #2547 

```elixir
Ecto.Changeset.cast({%{}, %{x: :decimal, y: :decimal}}, %{x: "1", y: "nan"}, ~w(x y)a)
#Ecto.Changeset<
  action: nil,
  changes: %{x: #Decimal<1>},
  errors: [y: {"is invalid", [type: :decimal, validation: :cast]}],
  data: %{},
  valid?: false

```

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -620,7 +620,7 @@ defmodule Ecto.Type do
   def cast(:decimal, term) when is_binary(term) do
     term
     |> Decimal.parse()
-    |> validate_decimal(term)
+    |> validate_decimal()
   end
   def cast(:decimal, term) when is_integer(term) do
     {:ok, Decimal.new(term)}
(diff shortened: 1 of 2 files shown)


lib/ecto/type.ex
```diff
@@ -928,10 +928,8 @@ defmodule Ecto.Type do
...
   end
 
-  defp validate_decimal({:ok, %Decimal{coef: coef}}, value) when coef in [:inf, :qNaN, :sNaN] do
-    message = "cannot cast #{inspect value} to :decimal. Define custom type to handle NaN or infinite decimals"
-    raise Ecto.CastError, type: :decimal, value: value, message: message
-  end
-
-  defp validate_decimal(other, _), do: other
+  defp validate_decimal({:ok, %Decimal{coef: coef}}) when coef in [:inf, :qNaN, :sNaN],
+    do: :error
```
michalmuskala: I probably should have asked that on the previous PR, but is the internal structure of the `Decimal` struct and the values we match on in here documented?

wojtekmach: Yes, they are public API.

PR #2592, opened 2018-06-17
[WIP] Build UPDATE SQL with "Ecto.Query"

### LINKED ISSUE #2277: Changeset.filters (Ecto.Changeset.optimistic_lock/3) doesn't work with nil fields

qcam:
An attempt to fix #2277.


test/ecto/adapters/postgres_test.exs
```diff
@@ -769,6 +770,12 @@ defmodule Ecto.Adapters.PostgresTest do
 
     query = update("prefix", "schema", [:x, :y], [id: 1, updated_at: nil], [])
     assert query == ~s{UPDATE "prefix"."schema" SET "x" = $1, "y" = $2 WHERE "id" = $3 AND "updated_at" IS NULL}
+
+    query = from(m in Schema, update: [set: [x: 0, y: 1]], where: [id: 0]) |> plan(:update_all) |> update()
+    assert query == ~s{UPDATE "schema" AS s0 SET "x" = 0, "y" = 1 WHERE (s0."id" = 0)}
+
+    query = from(m in Schema, update: [set: [x: 0, y: 1]], where: [id: 0], select: [:id]) |> plan(:update_all) |> update()
+    assert query == ~s{UPDATE "schema" AS s0 SET "x" = 0, "y" = 1 WHERE (s0."id" = 0) RETURNING s0."id"}
```
qcam: @josevalim it would be greatly appreciated if you could help to take a look at this. I wonder if this return is the desired outcome of the function, or should we keep it the same as the previous one?

josevalim:
Thanks @qcam! I think this is a step in the right direction. However, I would propose that instead of introducing a new operation, could we change [the filters parameter](https://github.com/qcam/ecto/blob/9bc7e01a5eae542e89dacdf481b17fe39bc79ff9/lib/ecto/adapter.ex#L179) in the existing update operation to be the query or the dynamic parameters of the query? This way we don't add new functions to the sql connection. What do you think?

qcam:
@josevalim thanks. Yeah I think that's the better approach too. I will continue working on the PR toward that then.

PR #2593, opened 2018-06-18
Automatically define schema typespecs

gVolop:
i'm not the owner of this code.. 
but y not merge it?
i think that's can be really helpfully!!

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -525,6 +525,13 @@ defmodule Ecto.Schema do
             {args, body} <- clauses do
           def __schema__(unquote_splicing(args)), do: unquote(body)
         end
+
+        field_types =
+          for {name, type} <- @changeset_fields do
+            {name, Ecto.Type.__typespec__(type)}
+          end
+
+        @type t() :: %__MODULE__{unquote_splicing(field_types)}
       end
 
     quote do
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -224,6 +224,67 @@ defmodule Ecto.Type do
     end
   end
 
+  @doc false
+  def __typespec__(:id) do
+    quote do: integer()
+  end
+  def __typespec__(:binary_id) do
+    quote do: binary()
+  end
+  def __typespec__(:integer) do
+    quote do: integer()
+  end
+  def __typespec__(:float) do
+    quote do: float()
+  end
+  def __typespec__(:boolean) do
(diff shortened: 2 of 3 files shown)


josevalim:
@wojtekmach is a frequent contributor to Ecto, so we will leave it up to him to contribute it or send a pull request when he feels it is ready. Thanks!

wojtekmach:
@gVolop thanks :) I haven't opened the PR because there are 2 unresolved things still:

- how to generate types for custom modules. On the branch we assume the custom type defined `CustomType.t()`, which may be incorrect assumption and we don't have a good way to enforce it. Some other options are adding callbacks to `Ecto.Type` which doesn't seem ideal either
- whether we should allow to overwrite the generated schema's `t()` type

@josevalim I'm happy to open the PR if it moves this forward. (I'm not sure if it does, I still don't know how to handle custom types!)

gVolop:
@wojtekmach 
in cases the custom struct have no defined type - y can't you generate the type like you did for the ecto schema?
you can take the default values of the struct `%CustomType{}` - and mapping them into type.. or - put `any()` for all fields
no?!

wojtekmach:
> you can take the default values of the struct %CustomType{} - and mapping them into type.. or - put any() for all fields
no?!

@gVolop a custom type doesn't have to be a struct, it can be just a string, for example.

gVolop:
i see there is `type` callback in the `Ecto.Type` behaviour, 
so, conceptually, we can say that if we can raise an error in case of not implemented callback (instead of warning) - you can use this callback to build the type ?

wojtekmach:
that callback has different purpose, it specifies the underlying DB type. For example, if you define custom `money` type in Postgres, type would return `:money`

gVolop:
humm... i thought that the underlying type have to be one of the existing ecto types...

@wojtekmach
sorry for bothering you.. but i think that feature can really be very useful!
 
what dyou think about add option for the `Ecto.Schema.field`, something like:
`field :name, :string, struct_type: String.t`
and while generating the type - use the `struct_type` OR map the ecto type and if custom - use the `t()` if defined or - `any()`

wojtekmach:
We shouldnt define the typespec when the type is used on a schema, rather when the type is defined (either build in or custom type) if that makes sense. Also for typespecs we need quote expressions.

I have opened a PR and the discussion has been moved there: https://github.com/elixir-ecto/ecto/pull/2594

Issue #425, opened 2015-02-04
Ecto.Model typespec


whatyouhide:
I have a `MyModel` model where I `use Ecto.Model`. In the `MyModel` module, I also keep a bunch of functions that operate on instances of `MyModel`.
When I write the `@spec`s for those functions, I'd like to directly use the type `t` when dealing with an instance; it would be cool to have the `MyModel.t` type automatically defined from the schema data.

I'm not sure this is something that can be done (dynamic typespecs generation) and that's why I'm opening this issue. I'd like to give this a try if it's possible :smiley:

josevalim:
It would be nice to automatically generate the spec with the types from the schema. However that would require extending Ecto.Type to add yet another function. Thoughts? /cc @drewolson @ericmj

drewolson:
My experience with type specs is very limited so I'll bow out of this
discussion.

ericmj:
I'm -1 because the callback added to Ecto.Type would have to return a quote, so people implementin it need to learn about quoting and typespecs. We could make the callback optional but I'm not sure if that's better.

@whatyouhide It's easy to create a type for your model to use in typespecs. Just do `@type t :: %__MODULE__{}`.

whatyouhide:
@ericmj yes, that would work too but we wouldn't have any type of specs about the types of the fields in the struct; this is a bummer because we have those things in the schema definition. However, when I opened the issue I didn't think of custom `Ecto.Type`s at all actually.

That said, maybe making the callback optional but allowing people to specify the type via a quoted callback is not such a bad idea. Let's see what others think about this maybe?

josevalim:
Right, that's exactly the advantage of moving it to `Ecto.Type`. We already give the type information when mounting the schema.

@ericmj your suggestion won't have any type information though, just lay out the struct structure.

whatyouhide:
@josevalim I think even just laying out the struct info is a good starting point. I also think that most people would implement the callback in `Ecto.Type` if we give good documentation on how to do it; it doesn't sound like a very complicated task if you have a "template" to look at.

josevalim:
I am closing this. It would be nice to have but currently is low priority.

hickscorp:
I know this is a bit outdated, but I was wondering whether or not there was progress on this?

We have an app that uses Ecto and OTP, and we would really like to be able to spec our GenServer related inputs as `User.id` type etc. How would one do that right now?

michalmuskala:
[...]

PR #2594, opened 2018-06-18
Automatically define schema typespecs

wojtekmach:
Ref #425

### DIFF
--- a/examples/friends/mix.lock
+++ b/examples/friends/mix.lock
@@ -1,7 +1,7 @@
 %{
   "connection": {:hex, :connection, "1.0.4", "a1cae72211f0eef17705aaededacac3eb30e6625b04a6117c1b2db6ace7d5976", [], [], "hexpm"},
   "db_connection": {:hex, :db_connection, "1.1.3", "89b30ca1ef0a3b469b1c779579590688561d586694a3ce8792985d4d7e575a61", [:mix], [{:connection, "~> 1.0.2", [hex: :connection, repo: "hexpm", optional: false]}, {:poolboy, "~> 1.5", [hex: :poolboy, repo: "hexpm", optional: true]}, {:sbroker, "~> 1.0", [hex: :sbroker, repo: "hexpm", optional: true]}], "hexpm"},
-  "decimal": {:hex, :decimal, "1.4.1", "ad9e501edf7322f122f7fc151cce7c2a0c9ada96f2b0155b8a09a795c2029770", [:mix], [], "hexpm"},
+  "decimal": {:hex, :decimal, "1.5.0", "b0433a36d0e2430e3d50291b1c65f53c37d56f83665b43d79963684865beab68", [:mix], [], "hexpm"},
   "ecto": {:hex, :ecto, "2.0.1", "cf97a4d353e14af3d3cc3b4452cfbd18b3aeee1fb4075475efeccec3853444a9", [:mix], [{:poison, "~> 1.5 or ~> 2.0", [hex: :poison, optional: true]}, {:sbroker, "~> 1.0-beta", [hex: :sbroker, optional: true]}, {:mariaex, "~> 0.7.7", [hex: :mariaex, optional: true]}, {:postgrex, "~> 0.11.2", [hex: :postgrex, optional: true]}, {:db_connection, "~> 1.0-rc.2", [hex: :db_connection, optional: true]}, {:decimal, "~> 1.0", [hex: :decimal, optional: false]}, {:poolboy, "~> 1.5", [hex: :poolboy, optional: false]}]},
   "poolboy": {:hex, :poolboy, "1.5.1", "6b46163901cfd0a1b43d692657ed9d7e599853b3b21b95ae5ae0a777cf9b6ca8", [], [], "hexpm"},
   "postgrex": {:hex, :postgrex, "0.13.5", "3d931aba29363e1443da167a4b12f06dcd171103c424de15e5f3fc2ba3e6d9c5", [:mix], [{:connection, "~> 1.0", [hex: :connection, repo: "hexpm", optional: false]}, {:db_connection, "~> 1.1", [hex: :db_connection, repo: "hexpm", optional: false]}, {:decimal, "~> 1.0", [hex: :decimal, repo: "hexpm", optional: false]}], "hexpm"},
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -526,12 +526,15 @@ defmodule Ecto.Schema do
           def __schema__(unquote_splicing(args)), do: unquote(body)
         end
 
-        field_types =
+        types =
           for {name, type} <- @changeset_fields do
             {name, Ecto.Type.__typespec__(type)}
           end
 
-        @type t() :: %__MODULE__{unquote_splicing(field_types)}
+        types = [{:__struct__, __MODULE__} | types]
+        types = if meta?, do: [{:__meta__, quote do: Ecto.Schema.Metadata.t()} | types], else: types
+
+        @type t() :: %{unquote_splicing(types)}
       end
 
     quote do


lib/ecto/type.ex
```diff
@@ -224,6 +224,67 @@ defmodule Ecto.Type do
...
+    quote do: binary()
+  end
+  def __typespec__({:array, inner}) do
+    quote do: [Macro.escape(Ecto.Type.__typespec__(unquote(inner)))]
+  end
+  def __typespec__(:map) do
+    quote do: map()
+  end
+  def __typespec__({:map, inner}) do
+    quote do: [Macro.escape(Ecto.Type.__typespec__(unquote(inner)))]
```
michalmuskala: This should be 
```
quote do: %{String.t => Macro.escape(Ecto.Type.__typespec__(unquote(inner)))}
```

lib/ecto/type.ex
```diff
@@ -224,6 +224,67 @@ defmodule Ecto.Type do
...
+    quote do: String.t()
+  end
+  def __typespec__(:binary) do
+    quote do: binary()
+  end
+  def __typespec__({:array, inner}) do
+    quote do: [Macro.escape(Ecto.Type.__typespec__(unquote(inner)))]
+  end
+  def __typespec__(:map) do
+    quote do: map()
```
michalmuskala: This should probably be `%{String.t => term}`

lib/ecto/type.ex
```diff
@@ -224,6 +224,67 @@ defmodule Ecto.Type do
...
+    quote do: [Macro.escape(Ecto.Type.__typespec__(unquote(inner)))]
+  end
+  def __typespec__(:decimal) do
+    quote do: Decimal.t()
+  end
+  def __typespec__(:date) do
+    quote do: Decimal.t()
+  end
+  def __typespec__(:time) do
+    quote do: Decimal.t()
```
michalmuskala: This (and `:date`) are wrong as well

michalmuskala:
The main problem here is that we cannot reliably implement this because of custom types lacking the `t` type.

What if we first introduced a `@typecallback` mechanism in elixir? It seems like something that could be a natural extension of the current `@callback` and `@macrocallback` mechanism.

josevalim:
I don’t think we need a @typecallback. IMO we should rather have a callback
in Ecto.Type that returns the AST of the type.

lib/ecto/type.ex
wojtekmach: Good call. Macro.escape cannot be used here like that because:

```elixir
iex(1)> t Friends.Person
@type t() :: %Friends.Person{
        __meta__: term(),
        age: integer(),
        first_name: String.t(),
        id: integer(),
        last_name: String.t(),
        map: [Macro.escape(Ecto.Type.__typespec__(:integer))]
      }
```

I'll investigate and add unit test. Looks like it wasn't quite ready after all, sorry for noise.

michalmuskala:
Returning AST is going to be problematic. Defining an ecto type is already more ceremony than it, IMHO, should be. This will only make it harder.

josevalim:
@michalmuskala we make it optional, defaults to `any`. But there is no way to extract more information without requiring new callbacks. And I doubt `@typecallback` would pass an Elixir core proposal, it is not a common issue and there is a solution in this specific case. Even if it did, we are looking at 2 years to see this problem solved.

josevalim:
Btw, there is another discussion for making the other callbacks optional too, so we will clear up the ceremony a bit anyway.

wojtekmach:
Let's say that we have the following:

```elixir
defmodule Post do
  schema "posts" do
    field :permalink, Permalink
  end
end

defmodule Permalink do
  @behaviour Ecto.Type

  @impl true
  def typespec() do
    quote do: String.t()
  end
end
```

having a new callback in Ecto.Type would allow us to "infer" the type in schema, so we end up with:

```elixir
iex> t Post
@type t() :: %Friends.Person{
  permalink: String.t(),
  # ...
}
```

That was the main idea behind the callback, right?

However, I think it's actually preferred not to "infer" and instead of have something like:

```elixir
iex> t Post
@type t() :: %Friends.Person{
  permalink: Permalink.t(),
  # ...
}
```

Looking at `Person.t()` we have more information, and if someone generates docs it's nice to link to custom type module's page for more information.

(I tried making this point in the previous discussion, sorry for repetition!)

Thus the callback can reutrn just boolean:

```elixir
defmodule Permalink do
  @type t() :: String.t()

  def typespec?(), do: true

  # or:

  def typespec(), do: :t # I haven't considered parametrized custom types
end
```

And ideally, someday, we don't need the callback but look just whether the type is defined.

P.S. FWIW in the previous approach we can still do:

```elixir
defmodule Permalink do
  @type t() :: String.t()

  def typespec() do
    quote do: Permalink.t()
  end
end
```

josevalim:
Hrm. I see.

The @type t is indeed very simple. I would go with that and let dialyzer
warm on missing types.
-- 


wojtekmach:
I had a pending test for handling overwriting the autogenerated type, I'm not sure if this is really needed so I removed it in 9f420a5. I'm using private `Code.Typespec.fetch_specs` to test that types are generated; is there a nicer way to test it, or leave this untested? Apart from this, I think this is ready.

wojtekmach:
Actually, Code.Typespec is not even in 1.6 yet, I've removed the tests for now.

michalmuskala:
One problem with having dialyzer warn on missing types is that library authors will probably never get that warning. That's why I proposed a `typecallback`.

wojtekmach:
Just saw @ejpcmac's [Typed Elixir structs without boilerplate](https://medium.com/@ejpcmac/typed-elixir-structs-without-boilerplate-c565520691a3); I wonder if generated typespecs should actually be e.g.:

```elixir
@type t() :: %Post{
        title: String.t() | nil
      }
```

However, often `nil` isn't actually possible because we have non-null constraint, so maybe we should be able to do: `field :title, :string, required: true` (matches null: false in migrations), which means we don't have to call `validate_required` for that field.

fertapric:
I think `field :title, :string, required: true` could be misleading and reduce the flexibility that the current schema provide. Different `def <operation>_changeset` could require some fields for that operation, but not for the other ones :)

wojtekmach:
@fertapric I don't think we lose flexibility because it's a deliberate choice: if the field is not required at all times then we don't mark it as such and we mark the column as nullable.

michalmuskala:
I personally don't think it's a good idea to modify our current API for the sake of generating dialyzer types that are probably (and possibly unfortunately) not that frequently used.

michalmuskala:
What if we always generated all fields as nullable? Especially if we're not given a default value - that's what the newly constructed `%Struct{}` would have.

wojtekmach:
Yes, making everything nullable is a solution and you're right it matches newly constructed struct; let me repeat the previous point though, if we autogenerated `title: String.t | nil` but we also have not null constraint then the type is unnecessarily wide. (and ugly)

fertapric:
Could module attributes be an option? In the same fashion as `@primary_key`, `@foreign_key_type`, or `@timestamp_opts` :)

wojtekmach:
@fertalric what module variable name you have in mind and what would it do?

fertapric:
@wojtekmach I was thinking on having a list of required fields, similar to what `@enforce_keys` does with structs. 

Following that train of thought, I considered using `@enforce_keys` as a way to communicate non-nullable fields, with the benefit of making the generated structs more robust. However, there were some use cases in my personal codebase in which, at some point, all the fields of any schema were `nil` even if the columns for those fields were marked as `null: false` in the migration (database). The documentation of `Ecto.Changeset` contains some examples of what I'm talking about:

```elixir
defmodule User do
  use Ecto.Schema
  import Ecto.Changeset

  schema "users" do
    field :name
    field :email
    field :age, :integer
  end

  def changeset(user, params \\ %{}) do
    user
    |> cast(params, [:name, :email, :age])
    |> validate_required([:name, :email])
    |> validate_format(:email, ~r/@/)
    |> validate_inclusion(:age, 18..100)
    |> unique_constraint(:email)
  end
end

changeset = User.changeset(%User{}, %{age: 0, email: "mary@example.com"})
```

In that example, `%User{}` would create the struct `%User{name: nil, email: nil, age: nil}`, even if the columns `name` and `email` of the `users` table were defined as `null: false`. Having `field :name, :string, required: true` would not be accurate then.

wojtekmach:
@fertapric thanks for explanation,

> In that example, %User{} would create the struct %User{name: nil, email: nil, age: nil}, even if the columns name and email of the users table were defined as null: false. Having field :name, :string, required: true would not be accurate then.

Are you proposing to do `@enforce_keys [:name]`, or just pointing out even if we have stuff like `required: true` (or `@required_fields` for that matter) it's possible to bypass it? In any case, users will always be able to do stuff like explicitly set `%User{name: nil}` and that's fine, it'll be caught later when casting.

fertapric:
I'm suggesting that the generated `@type` might not reflect all the possible values. 

In fact, I made a small PoC and Dialyzer would warn in such cases:

```elixir
defmodule MyApp.User do
  use Ecto.Schema
  import Ecto.Changeset

  @type t :: %__MODULE__{
    name: String.t(),
    email: String.t(),
    age: integer | nil
  }

  schema "users" do
    field :name
    field :email
    field :age, :integer
  end

  @spec changeset(t, map) :: Ecto.Changeset.t()
  def changeset(user, params \\ %{}) do
    user
    |> cast(params, [:name, :email, :age])
    |> validate_required([:name, :email])
    |> validate_format(:email, ~r/@/)
    |> validate_inclusion(:age, 18..100)
    |> unique_constraint(:email)
  end
end

changeset = MyApp.User.changeset(%MyApp.User{}, %{age: 0, email: "mary@example.com"})
```

Dyalyzer emits:

```elixir
The call 'Elixir.MyApp.User':changeset(#{'__meta__':=#{'__struct__':='Elixir.Ecto.Schema.Metadata', 'context':='nil', 'source':={'nil',<<_:40>>}, 'state':='built'}, '__struct__':='Elixir.MyApp.User', 'age':='nil', 'email':='nil', 'id':='nil', 'name':='nil'},#{'age':=0, 'email':=<<_:128>>}) breaks the contract (t(),map()) -> 'Elixir.Ecto.Changeset':t()
```

The warning disappears if I define the `@type` as:

```elixir
@type t :: %__MODULE__{
    name: String.t(),
    email: String.t(),
    age: integer | nil
  }
```

wojtekmach:
Good catch 👍

michalmuskala:
Here's a crazy idea.

What if for each field we generated a type with the same name, so:

```
schema "foo" do
  field :abc, :integer
  field :def, :string
end
```

would produce:

```
@type abc :: integer | nil
@type def :: String.t() | nil
@type t :: %__MODULE__{abc: abc, def: def}
```

At the same time, if you wanted to override a type, you'd just define `@type abc :: pos_integer` and we'd use that - we wouldn't override it.

wojtekmach:
@michalmuskala neat, I will investigate

fertapric:
Could be a good compromise to generate all the fields as nullable in the typespec `my_field: MyEctoType.t() | nil` and leave further customizations off the table? As I shown above, Dialyzer will warn with empty structs `%MySchema{}` if the typespec contains non-nullable fields; and I would dare to say that's the most common use case.

If developers need to define a different typespec, they can always do it before the schema `@type t` and it won't be overriden. If, in the future, there are use cases for having non-nullable fields or custom typespecs per field, we can always go back to the ideas of this PR :)

Thoughts?

gVolop:
@wojtekmach @michalmuskala 

y not use the first way of this PR? 
as i understand there is 2 points: 
1) how to handle custom types 
2) how to allow overriding of type-spec by the developer

my thought - we can add option for the `Ecto.Schema.field`, (like the `default`, and `virtual`)
can be called - `type_spec`, 
and while we generating the schema type - if specified type_spec - use it, else - we can map the ecto type to Elixir type (`:integer --> integer()`) , and if it's a custom type - use `any()`
so - if the developer want to specify the typespec for his custom type - he can, but we don't need to guess it, and don't need to rely on the `t()` implementation of his type - everything under his responsibility.

e.g. 
```
schema "person" do
   field :name, :string, struct_type: String.t
   field :age, :integer
   field :custom_details, CustomDetails, struct_type: CustomDetails.t
   field :spouse_custom_details, CustomDetails
end
```
```
defmodule CustomDetails do
  @behaviour Ecto.Type
  @type t() :: ____
 ......
```

PR #2596, opened 2018-06-18
Specify the type expected on schema source option

stevedomin:
The docs should indicate that it expects this value to be an atom.


josevalim:

PR #2599, opened 2018-06-19
Many to many documentation: :on_replace flag

Errol-Hassall:
Added docs to better explain many to many relationships in regards to deleting and updating associations, also explains the flags that you can use for on_replace for many to many and how they differ to one to one and one to many associations

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -202,7 +202,7 @@ defmodule Ecto.Changeset do
     This works fine, if you keep adding more `tags` to a post, but if you want to then remove some `tags` ecto will throw a `:raise` error.
     By default `:on_replace` is set to `:raise`, it is expecting the same amount of `tags` if you are to do another upsert.
 
-    Setting `:on_replace` to `:delete` will tell ecto that you want to delete the row in the linkning table called `posts_tags`.
+    Setting `:on_replace` to `:delete` will tell ecto that you want to delete the row in the linking table called `posts_tags`.
     Now you can add and remove `tags` to your `posts` at will.
 
     `many_to_many many_to_many(:tags, Tag, join_through: "posts_tags", :on_replace: :delete)`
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -171,6 +171,42 @@ defmodule Ecto.Changeset do
...
+
+          many_to_many many_to_many(:tags, Tag, join_through: "posts_tags")
+        end
+      end
+
+    `tags` has many `posts`, `posts` can have many `tags` thus they have a `many_to_many` realtionship.
+    This works fine, if you keep adding more `tags` to a post, but if you want to then remove some `tags` ecto will throw a `:raise` error.
+    By default `:on_replace` is set to `:raise`, it is expecting the same amount of `tags` if you are to do another upsert.
+
+    Setting `:on_replace` to `:delete` will tell ecto that you want to delete the row in the linkning table called `posts_tags`.
```
jackmarchant: linking* typo

michalmuskala:
To be honest, I'm not convinced about this change. There's nothing specific to `many_to_many` in the `:delete` option - it works the same with other association types as well. Or is there something I'm missing?

josevalim:
Thanks for clarifying, @Errol-Hassall .

> It might also be worth adding some more explanation on how the :delete will only remove the entry from the linking table not the table that your joining too. This is something that I would have found very helpful myself, and I'm sure it would help future people.

I think this belongs in the `many_to_many` section in `Ecto.Schema`. It may even have already been there? However, how to make sure developers will read those sections? :S

Errol-Hassall:
@josevalim I had a quick look at the `many_to_many` section, you're right it is already there. I didn't see that before I think it's because I'm still new to Ecto and Elixir as a whole and didn't know where to look. This can probably be closed as it's already in the schema section. Unless theres something I could add?

josevalim:
Thanks @Errol-Hassall, I do believe the current docs are good as is then. We will keep listening to feedback though and see if there is anything else to improve!

Issue #2589, opened 2018-06-13
Using on_conflict option of Repo.insert with a conditional update raises Ecto.StaleEntryError when the condition is not met


skateinmars:
### Environment

* Elixir version (elixir -v): Elixir 1.6.5 (compiled with OTP 19)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6.9 on x86_64-pc-linux-gnu (Debian 9.6.9-2.pgdg90+1), compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit
* Ecto version (mix deps): 2.2.10
* Database adapter and version (mix deps): postgrex 0.13.5
* Operating system: Ubuntu 17.10

### Current behavior

I am trying to use `Repo.insert` with its `on_conflict` option to conditionally update my record, based on its exiting attributes. Here is a small example to reproduce the issue:

```
defmodule User do
  use Ecto.Schema

  schema "users" do
    field(:email, :string)
    field(:comment, :string)

    timestamps()
  end
end
```

```
defmodule UserTest do
  use ExUnit.Case
  import Ecto.Query, only: [from: 2, update: 2]

  setup do
    :ok = Ecto.Adapters.SQL.Sandbox.checkout(Repo)
  end

  test "Raises an error" do
    email = "email@example.com"
    first_updated_at = DateTime.from_naive!(~N[2017-06-12 11:11:11.004], "Etc/UTC")
    _first_record = Repo.insert!(%User{email: email, updated_at: first_updated_at})

    second_updated_at = DateTime.from_naive!(~N[2010-01-01 11:11:11.004], "Etc/UTC")

    on_conflict_changes = [comment: "Duplicate found!"]

    conditional_update_query =
      from(p in User, where: p.updated_at < ^second_updated_at)
      |> update(set: ^on_conflict_changes)

    try do
      Repo.insert(
        %User{email: email, updated_at: second_updated_at},
        returning: true,
        on_conflict: conditional_update_query,
        conflict_target: [:email]
      )
    rescue
      Ecto.StaleEntryError ->
        IO.inspect "Should not have raised?"
    end
  end
end
```

### Expected behavior

I am not exactly certain what should be returned in this case. Returning the existing record may be confusing, so maybe a changeset error?

The current behaviour may actually be the one expected, but if so it could be documented.

josevalim:
[...]

PR #2603, opened 2018-06-22
Repo.delete exception not documented

fmcgeough:
If a row is already deleted from db Repo.delete raises Ecto.StaleEntryError.
This change docs this for the API.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -928,7 +928,8 @@ defmodule Ecto.Repo do
   Deletes a struct using its primary key.
 
   If the struct has no primary key, `Ecto.NoPrimaryKeyFieldError`
-  will be raised.
+  will be raised. If the struct has been removed from db prior to
+  call, `Ecto.StaleEntryError` will be raised.
 
   It returns `{:ok, struct}` if the struct has been successfully
   deleted or `{:error, changeset}` if there was a validation


josevalim:

mguimas:
@josevalim I completely disagree with this solution. It is not the documentation that needs to be fixed, it is the `Repo.delete` API that must be fixed to return `{:error, ...}` in case of an error.

If one wants to get an exception in case of a failure during deletion, e.g., because the item was not found, then one should call `Repo.delete!(...)` which is there exactly to throw exceptions.

The same issue happens with `Repo.update`.

josevalim:
Please see #2589. We are going to allow `{:error, ...}` to be returned with an option. We need an option because we need to know on which field to put the error on, as both update/delete operations return a changeset on errors and have it return something else would be breaking.

mguimas:
@josevalim I have a suggestion: why not set the action to `:stale` on the changeset, instead of throwing the exception? This would make sense because the `Repo` functions already set this action according to the function, and in this case `:stale` seems clean and appropriate.

josevalim:
There are two issues with this approach:

  1. action is about the repository action that was trigger, such as Repo.insert. I think setting it to stale would cause lost of information, as you no longer know what you called, and it would be a misuse of the config

  2. most projects are not used to check the value of action, because doing so is not required at all. So imagine someone is building an API. They may call the repository, they will call the repo and get `{:error, changeset}` and render `changeset.errors` back to the client. Except the errors are empty because the information of what went wrong is hidden inside action, which you don't care about 99% of the time. That's why I believe the best way to handle it is by explicitly adding it to a field of your choice as an error

mguimas:
@josevalim Ok, I tested this

```
iex(156)> x = %Avatar{nick_name: "fellowguy", user_id: 1} |> change |> foreign_key_constraint(:user_id) |>  Repo.insert

18:24:04.979 [debug] QUERY OK db=0.7ms
begin []

18:24:04.982 [debug] QUERY ERROR db=2.5ms
INSERT INTO "avatars" ("nick_name","user_id") VALUES ($1,$2) RETURNING "id" ["fellowguy", 1]

18:24:04.983 [debug] QUERY OK db=1.7ms
rollback []
{:error,
 #Ecto.Changeset<
   action: :insert,
   changes: %{nick_name: "fellowguy", user_id: 1},
   errors: [user_id: {"does not exist", []}],
   data: #EctoAssoc.Avatar<>,
   valid?: false
 >}
```

and adding an entry in `errors` seems more appropriate then.

But, why not put the error automatically on the primary key, like in this example? I am not seeing other `Ecto.StaleEntryError` situations when passing a struct or changset to a `Repo.update` or `Repo.delete` functions.

josevalim:
Because:

1. Doing this change would be backwards compatible without an option
2. Often in the UI you don't remember to check or show errors that happen in the primary key field - which would trigger the same issue I mentioned before, where errors silently pass through

mguimas:
@josevalim 

Point 1, I did not understand.

Point 2, if developers no longer handle the stale exception, I see them changing their code to look for errors of the primary key field. When the API changes, that is the expectable change  as well.

mguimas:
@josevalim I think we are probably talking about different things. I don't mind adding an option to generate an error, instead of throwing an exception, for backwards compatibility. What I was asking is, in case the option to generate the error is passed, why not generate the error on the primary key field?

josevalim:
Oh, I see, sorry! You can choose that if you want to. The argument to the option is exactly the field to add the error to. I can think of cases people would rather have it elsewhere (for example, a permalink, title or another unique field).

mguimas:
@josevalim Ok, so this option will come for the `Repo.update` and `Repo.delete` functions in what version of Ecto? the sooner the better, as always :-)

josevalim:
It is already in master and we are actively working towards ecto 3.0, so it should be out soon.

PR #2604, opened 2018-06-25
Note that Repo.in_transaction?/0 doesn't reflect per-test transactions

nathanl:
Clarifying the docs based on conversation at https://github.com/elixir-ecto/postgrex/issues/392

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1055,6 +1055,15 @@ defmodule Ecto.Repo do
   @doc """
   Returns true if the current process is inside a transaction.
 
+  If you are using the `Ecto.Adapters.SQL.Sandbox` in tests, note that even
+  though each test is inside a transaction, `in_transaction?/0` will only
+  return true inside transactions explicitly created with `transaction/2`. This
+  is done so the test environment mimics dev and prod.
+
+  If you are trying to debug transaction-related code while using
+  `Ecto.Adapters.SQL.Sandbox`, it may be more helpful to configure the database
+  to log all statements and consult those logs.
+
   ## Examples
 
       MyRepo.in_transaction?


lib/ecto/repo.ex
```diff
@@ -1055,6 +1055,12 @@ defmodule Ecto.Repo do
   @doc """
   Returns true if the current process is inside a transaction.
 
+  Note that this may not reflect transactions created for tests; since tests
+  almost always run in a transaction, it would almost always report `true`,
+  even when the same code running in production would not be inside a
+  transaction. See the database logs for reliable information on when
+  transactions begin and end.
```
josevalim: I think we can phrase this differently. Because it is not specific to tests but only to the sql sandbox. Something like this:

> If you are using the `Ecto.Adapters.SQL.Sandbox` in tests, note that even though
> each test is inside a transaction, `in_transaction?/1` will only return true if you also
> call `transaction/2`. This is done so the test environment mimics dev and prod.
> However, this may be surprising if you run XYZ...

Thoughts?

fertapric: I was about to comment something similar. Sounds good to me.

Minor nitpick, maybe "will only return true inside transactions explicitly created with `transaction/2`" instead of "will only return true if you also call `transactions/2`?

fertapric: Ah! And should a similar paragraph be added to `Ecto.Adapters.SQL.Sandbox`?

josevalim:
I don’t think we need to change it there as we describe this in the detail
there. Thoughts?
-- 


nathanl:
@josevalim @fertapric I updated the commit to incorporate your suggestions.

lib/ecto/repo.ex
```diff
@@ -1055,6 +1055,15 @@ defmodule Ecto.Repo do
   @doc """
   Returns true if the current process is inside a transaction.
 
+  If you are using the Ecto.Adapters.SQL.Sandbox in tests, note that even
```
fertapric: I would include backticks to enable auto-linking:

- `Ecto.Adapters.SQL.Sandbox`
- `in_transaction?/1`
- `transaction/2`

nathanl: Good idea. Done.

fertapric:
> I don’t think we need to change it there as we describe this in the detail
there. Thoughts?

Yeah. This is probably enough :)

josevalim:

PR #2606, opened 2018-06-26
Allows support for '...' in bindings with select_merge macro

### LINKED ISSUE #2605: select_merge throws an exception when '...' is used in the bindings list
* Elixir version (elixir -v): 1.6.5
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 10.4
* Ecto version (mix deps): 2.2-3.0.0-dev
* Database adapter and version (mix deps): 0.13.5-Postgrex from ecto master mix.lock
* Operating system: macOS High Sierra

### Current behavior

select_merge throws an exception when '...' is used in the bindings list.

```
  query =
        "posts"
        |> select_merge([..., p], %{title: p.title})

  ** (CompileError) test/ecto/query/builder/select_test.exs:121: undefined function escape_count/0
      (stdlib) lists.erl:1338: :lists.foreach/2
      (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
      (elixir) lib/code.ex:677: Code.require_file/2
      (elixir) lib/kernel/parallel_compiler.ex:201: anonymous fn/4 in   Kernel.ParallelCompiler.spawn_workers/6
```

### Expected behavior

Should not throw exception

ideanl:
This PR fixes a bug that causes an exception when '...' is used in the bindings for select_merge. This is done by unquoting the query prior to the `Builder.Select.merge/2` call. Currently, '...' only appears to work when a select_merge is already used previously in the query.

This bug is also in Ecto version 2.2 and maybe earlier.

Resolves #2605

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -200,7 +200,8 @@ defmodule Ecto.Query.Builder.Select do
       Builder.apply_query(query, __MODULE__, [select], env)
     else
       quote do
-        Builder.Select.merge(unquote(query), unquote(select))
+        query = unquote(query)
+        Builder.Select.merge(query, unquote(select))
       end
     end
   end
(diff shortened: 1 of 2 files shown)


josevalim:

PR #2607, opened 2018-06-26
Ecto benchmarks

tallysmartins:
The initial benchmark suite for Ecto. Tests rely on the [Benchee](https://github.com/PragTob/benchee)  library.


bench/README.md
```diff
@@ -7,7 +7,13 @@ need `Postgres` and `MySQL` up and running.
...
 
 ```
-$ export BENCHMARKS_OUTPUT_PATH=bench/results && mix run bench/bench_helper.exs
+# POSIX-compatible shells
+$ BENCHMARKS_OUTPUT_PATH=bench/results && mix run bench/bench_helper.exs
+```
+
+```
+# other shells
+$ env BENCHMARKS_OUTPUT_PATH=bench/results && mix run bench/bench_helper.exs
```
hauleth: In both cases without the `&&` part.

tallysmartins:
Here are some of the results that I've gotten from running in my laptop.

[Repo.insert!/1](https://gist.github.com/tallysmartins/e0b31af78b6210f4a7fb7e7c1c993f4b)
[Repo.all/2](https://gist.github.com/tallysmartins/ef36d56cf31d4f8309af60c295507703)
[Repo.to_sql/2](https://gist.github.com/tallysmartins/48644d770a0d6d47ff40ca5b87ab86f4)
[Repo.load/2](https://gist.github.com/tallysmartins/6ceb5b62369c80e604287b263abcc0d8)

PragTob:
Any way we can help move this PR along? :grin:

bench/benchmarks/macro/all_bench.exs
```diff
@@ -0,0 +1,48 @@
...
+Benchee.run(
+  jobs,
+  formatters: [Benchee.Formatters.JSON],
+  formatter_options: [json: [file: file]],
+  time: 10
+)
+
+# Clean inserted data
+Ecto.Bench.PgRepo.delete_all(User)
+Ecto.Bench.MySQLRepo.delete_all(User)
```
michalmuskala: Could we move the cleanup code to the beginning of the benchmark instead of the end? That way if something goes wrong during the benchmark the next run should be fine.

tallysmartins: Yeah, good point.. but now that you said, I wonder if we need to do both, a pre and a post cleanup... subsequent benchmarks might be affected by this data...

michalmuskala: You're right

bench/benchmarks/macro/all_bench.exs
```diff
@@ -0,0 +1,48 @@
...
+# We need to insert data to fetch
+Ecto.Bench.PgRepo.insert_all(User, users)
+Ecto.Bench.MySQLRepo.insert_all(User, users)
+
+jobs = %{
+  "Pg Repo.all/2" => fn -> Ecto.Bench.PgRepo.all(User, limit: limit) end,
+  "MySQL Repo.all/2" => fn -> Ecto.Bench.MySQLRepo.all(User, limit: limit) end
+}
+
+path = System.get_env("BENCHMARKS_OUTPUT_PATH") || raise "I DON'T KNOW WHERE TO WRITE!!!"
```
fertapric: What do you think about using `bench/results` as default value? 

And maybe move it to Application env?

```
Application.put_env(
  :ecto, 
  :bench_results_path, 
  System.get_env("BENCHMARKS_OUTPUT_PATH") || "bench/results"
)
```

bench/README.md
```diff
@@ -0,0 +1,29 @@
...
+# POSIX-compatible shells
+$ BENCHMARKS_OUTPUT_PATH=bench/results MIX_ENV=bench mix run bench/bench_helper.exs
+```
+
+```
+# other shells
+$ env BENCHMARKS_OUTPUT_PATH=bench/results MIX_ENV=bench mix run bench/bench_helper.exs
+```
+
+Benchmarks are inside the benchmarks/ directory and are divided into two
```
fertapric: Minor nitpicks: 

1. I wouldn't use backticks for Potgres and MySQL:

```
and need PostgreSQL and MySQL up and running.
```

Also, I would use "PostgreSQL" instead of "Postgres

2. I would use backticks for the dir:

```
Benchmarks are inside the `benchmarks/` directory
```

3. And maybe use bullet points for the list of categories?

```
* `micro benchmarks`: Operations that don't actually interface with the database,
but might need it up and running to start the Ecto agents and processes.
* `macro benchmarks`: Operations that are actually run in the database. This are
more likely to integration tests.
```

bench/benchmarks/macro/all_bench.exs
tallysmartins: Well, it is concerned to ElixirBench to set this variable in order to collect and clean the results, so raising here means that something went wrong in our runners. In the near future  we want to wrap this kind of settings, so I don't think it will make big difference moving this now.

bench/README.md
tallysmartins: Thanks for your considerations! :heart:

josevalim:

Issue #2595, opened 2018-06-18
`count(<field>)` can sometimes be much slower than `count(*)`


hauleth:
### Environment

* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL (possibly other SQL DBs, unsure about NoSQLs)
* Ecto version (mix deps): 2.0+

### Current behavior

Currently there is no way (other than using fragment) to run query like `count(*)`. This can be problematic as `count(<field>)` on `<field>` that isn't indexed can be terribly slow (found by @mkaszubowski). The reason why this is happening is that by SQL standard `COUNT(<field>)` should return count of non-null values in DB so that mean it need to check wether any value in given field is empty or not, while `COUNT(*)` do not have this limitation and always return count of all matched rows.

Currently there is available workaround in form of `count(true)` in Ecto, but this isn't obvious solution and for sure isn't idiomatic.

### Expected behavior

There should be `count/0` function that would expand to counting all entries matching the query which I believe that most DBs have optimised enough.

josevalim:
Sounds good to me! PR please?

mkaszubowski:
Just to clarify: the problem exists even if the fields used in `count` is indexed properly. The difference between two approaches is that when using `count(<field>)`, PostgreSQL uses an `index scan` and still has to fetch the rows. With `count(*)` it can use `index only scan` without fetching the rows, which is way faster.

soteras:
I have a closed pr about *count* function #2575 if you want i can modify for that situation.

josevalim:
@soteras a PR is definitely welcome. Although note that we are not adding new callbacks to the repo module. The new count function will belong to the `Ecto.Query.API` instead. Please let me know if you have any questions!

hauleth:
@josevalim any sources where I could learn how are aggregate functions in `Ecto.Query.API` handled? Because it seems a little convoluted (as most macro systems).

josevalim:
There is a step in the middle, which is the Ecto.Query.Planner. My suggestion is to search the codebase for a less commonly used aggregator, such as `avg`, and see where in the codebase we touch it.

soteras:
@hauleth are you doing that or wants to?

hauleth:
I want to, but if you are willing then you can try as I do not have time right now to dig into it properly.

hauleth:
It seem it was merged. @soteras thanks for your work.

PR #2610, opened 2018-06-29
New function `count/0` for `Ecto.Query.API`

soteras:
### Objective
Create new function `count/0` for `Ecto.Query.API`

```
from(p in Post, select: count()) |> Repo.one

#Ecto.Query<from p in Post, select: count(:*)>                             

00:20:18.358 [debug] QUERY OK source="posts" db=0.4ms                      
SELECT count(*) FROM "posts" AS p0 []                                      
27                                   
iex(27)>   
```

### Information
There is a problem with `count/1` sometimes it can be slower.
See more #2595 

### Acceptance Criteria

- [x] New function in Ecto.Query.API
- [x] New functions Ecto.Query.Builder
- [x] Add new expression in postgres connection
- [x] Update docs
- [x] Add tests

### DIFF
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -513,6 +513,8 @@ if Code.ensure_loaded?(Postgrex) do
       [?(, intersperse_map(elems, ?,, &expr(&1, sources, query)), ?)]
     end
 
+    defp expr({:count, _, []}, _sources, _query), do: "count(*)"
+
     defp expr({fun, _, args}, sources, query) when is_atom(fun) and is_list(args) do
       {modifier, args} =
         case args do
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -8,7 +8,7 @@ defmodule Ecto.Query.API do
     * Inclusion operator: `in/2`
     * Search functions: `like/2` and `ilike/2`
     * Null check functions: `is_nil/1`
-    * Aggregates: `count/1`, `avg/1`, `sum/1`, `min/1`, `max/1`
+    * Aggregates: `count/0`, `count/1`, `avg/1`, `sum/1`, `min/1`, `max/1`
     * Date/time intervals: `datetime_add/3`, `date_add/3`, `from_now/2`, `ago/2`
     * Inside select: `struct/2`, `map/2`, `merge/2` and literals (map, tuples, lists, etc)
     * General: `fragment/1`, `field/2` and `type/2`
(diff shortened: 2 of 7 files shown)


soteras:
I was a little confused in *Ecto.Query.Builder* about the return expressions. Could someone explain to me?

lib/ecto/query/api.ex
```diff
@@ -158,6 +158,13 @@ defmodule Ecto.Query.API do
   """
   def count(value, :distinct), do: doc! [value, :distinct]
 
+  @doc """
+  Counts the entries in the table.
+
+      from p in Post, select: count()
+  """
+  def count, do: doc! ["*"]
```
josevalim: We should use `doc! []` since there are no arguments to `count`.

lib/ecto/query/builder.ex
```diff
@@ -261,9 +261,15 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:count, _, []}, type, params_acc, vars, env) do
+    expr = {:{}, [], [:count, [], [:*]]}
```
josevalim: This should expand to `{:{}, [], [:count, [], []]}`.

josevalim: The idea here is that `count(:*)` is a behaviour that is specific to SQL databases. We should only work with `count()` and the database can convert that to `count(:*)` if they find necessary.

josevalim:
> I was a little confused in Ecto.Query.Builder about the return expressions. Could someone explain to me?

Which return expressions? From which function? :)

josevalim:
This looks great! I have added some comments and we should be good to go!

soteras:
Like this *{:{}, [], [:count, [], [:*]]}* i read some example but i didn't understand much. What are the meaning of each part of tuple? 

```
{
  :{} => ?
  [] => What can i put here?
  [:count, [], [:*]] => The second param i didn't understand
}
```

josevalim:
Good question!

Since we are doing this at compile time, we need to return a data structured that, when compiled and executed will return Elixir AST. Therefore, what we need, is the **Escaped AST**.

The AST for `count()` is:

    iex> quote do: count()
    {:count, [], []}

The escaped AST is the one after we call `count()`:

    iex> Macro.escape quote do: count()
    {:{}, [], [:count, [], []]}

If we evaluate the escaped AST, we get the AST back:

    iex> Code.eval_quoted(Macro.escape quote do: count()) |> elem(0)
    {:count, [], []}

soteras:
Now i understood. Thanks @josevalim

soteras:
Done!!!

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -513,6 +513,8 @@ if Code.ensure_loaded?(Postgrex) do
       [?(, intersperse_map(elems, ?,, &expr(&1, sources, query)), ?)]
     end
 
+    defp expr({:count, _, []}, _sources, _query), do: ["count(*)"]
```
michalmuskala: This doesn't need to be wrapped in a list.

soteras:
@michalmuskala i did the changes and added this change in the mysql.

josevalim:

PR #2612, opened 2018-06-29
Make sure that FunctionClauseError is raised for nil fields

### LINKED ISSUE #2609: Ecto.Changeset.validate_required/2 should not silently pass nil as list of fields
Currently, this code would silently pass the validation and leave changeset valid:
```
Ecto.Changeset.validate_required(changeset, nil)
```

I believe it's bad because if you use module attributes to keep list of fields it's easy to get `nil` value in the attribute after a minor renaming, eg:
```
@settings_types %{application_id: :string, page_id: :string, access_token: :string}
@settings_fields Map.keys(@settings_types)
@settings_required_fields @fields
```

`@settings_required_fields` would contain `nil`, because it refers to `@fields` which is now called `@settings_fields`.

AndrewDryga:
This PR closes #2609.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -429,7 +429,7 @@ defmodule Ecto.Changeset do
 
   def cast(_data, %{__struct__: _} = params, _permitted, _opts) do
     raise Ecto.CastError, type: :map, value: params,
-                          message: "expected params to be a :map, got: `#{inspect params}`"
+                          message: "expected params to be a :map, got: `#{inspect(params)}`"
   end
 
   def cast({data, types}, params, permitted, opts) when is_map(data) do
@@ -573,7 +573,7 @@ defmodule Ecto.Changeset do
   This function should be used when working with the entire association at
   once (and not a single element of a many-style association) and using data
   external to the application.
-  
+
   `cast_assoc/3` works matching the records extracted form the database (preload)
   and compares it with the parameters provided form an external source.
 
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -573,7 +573,7 @@ defmodule Ecto.Changeset do
   This function should be used when working with the entire association at
   once (and not a single element of a many-style association) and using data
   external to the application.
-  
+
```
AndrewDryga: Trailing spaces are removed by the editor, please tell if you don't want this change.

lib/ecto/changeset.ex
```diff
@@ -429,7 +429,7 @@ defmodule Ecto.Changeset do
 
   def cast(_data, %{__struct__: _} = params, _permitted, _opts) do
     raise Ecto.CastError, type: :map, value: params,
-                          message: "expected params to be a :map, got: `#{inspect params}`"
+                          message: "expected params to be a :map, got: `#{inspect(params)}`"
```
AndrewDryga: I guess Elixir formatter would wrap argument in parentheses, I noticed it and fixed it manually. Please tell if it's not desired.

test/ecto/changeset_test.exs
```diff
@@ -249,6 +249,10 @@ defmodule Ecto.ChangesetTest do
     assert_raise FunctionClauseError, fn ->
       cast(%Post{}, %{}, %{})
     end
+
+    assert_raise FunctionClauseError, fn ->
+      cast(%Post{}, %{"title" => "foo"}, nil)
+    end
```
AndrewDryga: There already was no way to pass `nil` to `cast/3` so I only added the test to prevent regressions.

PR #2618, opened 2018-07-13
OVER, PARTITION BY and WINDOW

Anber:
Hi

PR implements window functions for PostgreSQL.
It can be used instead of fragments for the more clear and flexible solution of problems like https://github.com/coderplanets/coderplanets_server/issues/16 or https://github.com/absinthe-graphql/absinthe_relay/issues/100#issuecomment-357557000
```elixir
# SELECT row_number(s0."x") OVER () FROM "schema" AS s0
Schema |> select([r], row_number(r.x) |> over)

# SELECT count(s0."x") OVER w FROM "schema" AS s0 WINDOW w AS (PARTITION BY s0."x" ORDER BY s0."a", s0."b")
from s in Schema,
  window: (w as partition_by s.x, order_by: [s.a, s.b]),
  select: (count(s.x) |> over(w))

# SELECT count(s0."x") OVER (PARTITION BY s0."x" ORDER BY s0."y") FROM "schema" AS s0
partition = partition_by([r], r.x) |> order_by([r], r.y)
Schema |> select([r], count(r.x) |> over(^partition))
```

MySQL also supports window functions since 8.0, but we don't know version at compile time. So I can add the same implementation, but it will throw runtime errors if a user uses MySQL <8, or I can add compile-time errors like "… is not supported by MySQL".

If you are ok with this feature, I will add documentation and some more unit tests.

- [x] traverse of `:windows`
- [x] inspect
- [x] MySQL
- [x] tests
- [x] documentation

### DIFF
--- a/lib/ecto/adapters/mysql/connection.ex
+++ b/lib/ecto/adapters/mysql/connection.ex
@@ -82,12 +82,13 @@ if Code.ensure_loaded?(Mariaex) do
       where    = where(query, sources)
       group_by = group_by(query, sources)
       having   = having(query, sources)
+      window   = window(query, sources)
       order_by = order_by(query, sources)
       limit    = limit(query, sources)
       offset   = offset(query, sources)
       lock     = lock(query.lock)
 
-      [select, from, join, where, group_by, having, order_by, limit, offset | lock]
+      [select, from, join, where, group_by, having, window, order_by, limit, offset | lock]
     end
 
     def update_all(query, prefix \\ nil) do
@@ -320,13 +321,34 @@ if Code.ensure_loaded?(Mariaex) do
        end)]
     end
 
+    defp window(%Query{windows: []}, _sources), do: []
+    defp window(%Query{windows: windows} = query, sources) do
+      [" WINDOW " |
+       intersperse_map(windows, ", ", fn
+         {name, definition} ->
+           [quote_name(name), " AS ", partition_by(definition, sources, query)] end)]
+    end
+
+    defp partition_by(%QueryExpr{expr: opts}, sources, %Query{} = query) do
+      fields = Keyword.get(opts, :fields)
+      order_bys = Keyword.get_values(opts, :order_by) |> Enum.concat
+      fields = fields |> intersperse_map(", ", &expr(&1, sources, query))
+      ["(PARTITION BY ",
+        fields,
+        order_by(order_bys, query, sources),
+        ?)]
+    end
+
     defp order_by(%Query{order_bys: []}, _sources), do: []
     defp order_by(%Query{order_bys: order_bys} = query, sources) do
+      order_bys = Enum.flat_map(order_bys, & &1.expr)
+      order_by(order_bys, query, sources)
+    end
+
+    defp order_by([], _query, _sources), do: []
+    defp order_by(order_bys, query, sources) do
       [" ORDER BY " |
-       intersperse_map(order_bys, ", ", fn
-         %QueryExpr{expr: expr} ->
-           intersperse_map(expr, ", ", &order_by_expr(&1, sources, query))
-       end)]
+       intersperse_map(order_bys, ", ", &order_by_expr(&1, sources, query))]
     end
 
     defp order_by_expr({dir, expr}, sources, query) do
@@ -455,6 +477,21 @@ if Code.ensure_loaded?(Mariaex) do
 
     defp expr({:ilike, _, [_, _]}, _sources, query) do
       error!(query, "ilike is not supported by MySQL")
+    end
+
+    defp expr({:over, _, [agg, %QueryExpr{} = window]}, sources, query) do
+      aggregate = expr(agg, sources, query)
+      [aggregate, " OVER ", partition_by(window, sources, query)]
+    end
+
+    defp expr({:over, _, [agg, nil]}, sources, query) do
+      aggregate = expr(agg, sources, query)
+      [aggregate, " OVER ()"]
+    end
+
+    defp expr({:over, _, [agg, name]}, sources, query) when is_atom(name) do
+      aggregate = expr(agg, sources, query)
+      [aggregate, " OVER ", quote_name(name)]
     end
 
     defp expr({:{}, _, elems}, sources, query) do
--- a/lib/ecto/adapters/postgres/connection.ex
+++ b/lib/ecto/adapters/postgres/connection.ex
@@ -121,12 +121,13 @@ if Code.ensure_loaded?(Postgrex) do
       where    = where(query, sources)
       group_by = group_by(query, sources)
       having   = having(query, sources)
+      window   = window(query, sources)
       order_by = order_by(query, order_by_distinct, sources)
       limit    = limit(query, sources)
       offset   = offset(query, sources)
       lock     = lock(query.lock)
 
-      [select, from, join, where, group_by, having, order_by, limit, offset | lock]
+      [select, from, join, where, group_by, having, window, order_by, limit, offset | lock]
     end
 
     def update_all(%{from: %{source: source}} = query, prefix \\ nil) do
@@ -382,9 +383,32 @@ if Code.ensure_loaded?(Postgrex) do
        end)]
     end
 
+    defp window(%Query{windows: []}, _sources), do: []
+    defp window(%Query{windows: windows} = query, sources) do
+      [" WINDOW " |
+       intersperse_map(windows, ", ", fn
+         {name, definition} ->
+           [quote_name(name), " AS ", partition_by(definition, sources, query)] end)]
+    end
+
+    defp partition_by(%QueryExpr{expr: opts}, sources, %Query{} = query) do
+      fields = Keyword.get(opts, :fields)
+      order_bys = Keyword.get_values(opts, :order_by) |> Enum.concat
+      fields = fields |> intersperse_map(", ", &expr(&1, sources, query))
+      ["(PARTITION BY ",
+        fields,
+        order_by(order_bys, query, [], sources),
+        ?)]
+    end
+
     defp order_by(%Query{order_bys: []}, _distinct, _sources), do: []
     defp order_by(%Query{order_bys: order_bys} = query, distinct, sources) do
       order_bys = Enum.flat_map(order_bys, & &1.expr)
+      order_by(order_bys, query, distinct, sources)
+    end
+
+    defp order_by([], _query, _distinct, _sources), do: []
+    defp order_by(order_bys, query, distinct, sources) do
       [" ORDER BY " |
        intersperse_map(distinct ++ order_bys, ", ", &order_by_expr(&1, sources, query))]
     end
@@ -508,6 +532,21 @@ if Code.ensure_loaded?(Postgrex) do
     defp expr({:filter, _, [agg, filter]}, sources, query) do
       aggregate = expr(agg, sources, query)
       [aggregate, " FILTER (WHERE ", expr(filter, sources, query), ?)]
+    end
+
+    defp expr({:over, _, [agg, %QueryExpr{} = window]}, sources, query) do
+      aggregate = expr(agg, sources, query)
+      [aggregate, " OVER ", partition_by(window, sources, query)]
+    end
+
+    defp expr({:over, _, [agg, nil]}, sources, query) do
+      aggregate = expr(agg, sources, query)
+      [aggregate, " OVER ()"]
+    end
+
+    defp expr({:over, _, [agg, name]}, sources, query) when is_atom(name) do
+      aggregate = expr(agg, sources, query)
+      [aggregate, " OVER ", quote_name(name)]
     end
 
     defp expr({:{}, _, elems}, sources, query) do
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -177,6 +177,168 @@ defmodule Ecto.Query.API do
   def coalesce(value, expr), do: doc! [value, expr]
 
   @doc """
+  Invokes specified aggregation function in context of window.
+
+  Accepts any built-in or user-defined aggregate function as well as window functions.
+
+  `window` can be either existing window name or `partition_by/2` expression.
+  """
+  def over(agg_fun, window), do: doc! [agg_fun, window]
+
+  @doc """
+  Defines a partition for `over` or `window` clause.
+
+  `fields` can be either atom or a list of atoms and refers
+  fields in the source which will be used for grouping rows.
+
+  You can also specify an `:order_by` option which syntax and behaviour
+  are similar to `Ecto.Query.order_by/2` except the fact that it affects
+  rows only inside the window.
+
+      from e in Employee,
+           select: {
+             e.depname,
+             e.empno,
+             row_number() |> over(partition_by(e.depname, order_by: e.name))
+           }
+  """
+  def partition_by(fields, opts), do: doc! [fields, opts]
+
+  @doc """
+  Returns number of the current row within its partition, counting from 1.
+
+      from p in Post,
+           select: row_number() |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def row_number(), do: doc! []
+
+  @doc """
+  Returns rank of the current row with gaps; same as `row_number/0` of its first peer.
+
+      from p in Post,
+           select: rank() |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def rank(), do: doc! []
+
+  @doc """
+  Returns rank of the current row without gaps; this function counts peer groups.
+
+      from p in Post,
+           select: dense_rank() |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def dense_rank(), do: doc! []
+
+  @doc """
+  Returns relative rank of the current row: (rank - 1) / (total rows - 1).
+
+      from p in Post,
+           select: percent_rank() |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def percent_rank(), do: doc! []
+
+  @doc """
+  Returns relative rank of the current row:
+  (number of rows preceding or peer with current row) / (total rows).
+
+      from p in Post,
+           select: cume_dist() |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def cume_dist(), do: doc! []
+
+  @doc """
+  Returns integer ranging from 1 to the argument value, dividing the partition as equally as possible.
+
+      from p in Post,
+           select: ntile(10) |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def ntile(num_buckets), do: doc! [num_buckets]
+
+  @doc """
+  Returns value evaluated at the row that is the first row of the window frame.
+
+      from p in Post,
+           select: first_value(p.id) |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def first_value(value), do: doc! [value]
+
+  @doc """
+  Returns value evaluated at the row that is the last row of the window frame.
+
+      from p in Post,
+           select: last_value(p.id) |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def last_value(value), do: doc! [value]
+
+  @doc """
+  Returns value evaluated at the row that is the nth row of the window
+  frame (counting from 1); null if no such row.
+
+      from p in Post,
+           select: nth_value(p.id, 4) |> over(partition_by(p.category_id, order_by: p.date))
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def nth_value(value, nth), do: doc! [value, nth]
+
+  @doc """
+  Returns value evaluated at the row that is offset rows before
+  the current row within the partition; if there is no such row,
+  instead return default (which must be of the same type as value).
+  Both offset and default are evaluated with respect to the current row.
+  If omitted, offset defaults to 1 and default to null.
+
+      from e in Events,
+           windows: [w: partition_by(e.name, order_by: e.tick)],
+           select: {
+             e.tick,
+             e.action,
+             e.name,
+             lag(e.action) |> over(:w), # previous_action
+             lead(e.action) |> over(:w) # next_action
+           }
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def lag(value, offset, default), do: doc! [value, offset, default]
+
+  @doc """
+  Returns value evaluated at the row that is offset rows after
+  the current row within the partition; if there is no such row,
+  instead return default (which must be of the same type as value).
+  Both offset and default are evaluated with respect to the current row.
+  If omitted, offset defaults to 1 and default to null.
+
+      from e in Events,
+           windows: [w: partition_by(e.name, order_by: e.tick)],
+           select: {
+             e.tick,
+             e.action,
+             e.name,
+             lag(e.action) |> over(:w), # previous_action
+             lead(e.action) |> over(:w) # next_action
+           }
+
+  Note that this function must be invoked using window function syntax.
+  """
+  def lead(value, offset, default), do: doc! [value, offset, default]
+
+  @doc """
   Applies the given expression as a FILTER clause against an
   aggregate. This is currently only supported by Postgres.
 
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -266,6 +266,35 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  @over_aggs [
+    row_number: {[], :integer},
+    rank: {[], :integer},
+    dense_rank: {[], :integer},
+    percent_rank: {[], :float},
+    cume_dist: {[], :float},
+    ntile: {[:integer], :integer},
+    first_value: {[:any], :any},
+    last_value: {[:any], :any},
+    nth_value: {[:any, :integer], :any},
+    lag: {[:any, :integer, :any], :any},
+    lead: {[:any, :integer, :any], :any},
+    lag: {[:any, :integer], :any},
+    lead: {[:any, :integer], :any},
+    lag: {[:any], :any},
+    lead: {[:any], :any},
+  ]
+  @over_agg_names Enum.uniq(Keyword.keys(@over_aggs))
+  @over_agg_names_with_arity Enum.map(@over_aggs, fn {name, {in_types, _}} -> {name, length(in_types)} end)
+
+  def escape({agg, _, nil}, _type, _params_acc, _vars, _env) when {agg, 0} in @over_agg_names_with_arity do
+    error! "#{agg}/#{0} must be invoked using window function syntax"
+  end
+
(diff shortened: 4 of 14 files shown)


josevalim:
This looks amazing! We are currently focused on getting Elixir v1.7 out but we plan to review this in a week or two. Thanks for this great work!

> MySQL also supports window functions since 8.0, but we don't know version at compile time. So I can add the same implementation, but it will throw runtime errors if a user uses MySQL <8, or I can add compile-time errors like "… is not supported by MySQL".

For MySQL, we should emit the correct SQL and then it will fail if they are using an old version.

zachdaniel:
I want to go over this a little more in depth, but this all looks pretty great! Very excited about this feature. (Also hi everyone! I still read all the issues, I'm just constantly busy)

lib/ecto/query/builder.ex
```diff
@@ -266,6 +266,12 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  def escape({:row_number, _, [arg]}, type, params_acc, vars, env) do
```
josevalim: If row_number is a regular "function", you don't need this here. :)

josevalim: I.e. you can just remove this function and the default transformations will apply.

josevalim:
Hi @Anber! I did some review of the codebase and of the feature and it looks great! The `over/2` feature is excellent.

I have only two comments for now. You added this example:

```
partition = partition_by([r], r.x) |> order_by([r], r.y)
Schema |> select([r], count(r.x) |> over(^partition))
```

This feature (interpolation) is actually really hard to implement. That's because if `order_by` has any interpolation inside itself, we need to surface those parameters to the outer select in a separate pass. This will require changes in the planner, inspect, and other places around Ecto.

However, we do have another feature that is very similar: the `window` function. Both the interpolation syntax that you propose and the Window function have the same purpose, which is to avoid duplication and allow some sort of composition:

> The purpose of a WINDOW clause is to specify the behavior of window functions appearing in the query's SELECT List or ORDER BY Clause. These functions can reference the WINDOW clause entries by name in their OVER clauses. A WINDOW clause entry does not have to be referenced anywhere, however; if it is not used in the query it is simply ignored. It is possible to use window functions without any WINDOW clause at all, since a window function call can specify its window definition directly in its OVER clause. However, the WINDOW clause saves typing when the same window definition is needed for more than one window function.

My suggestion is the following: remove `partition_by` as a function in Ecto.Query because making it work as an interpolation under all scenarios will be extremely complex and ultimately undesired.

test/ecto/adapters/postgres_test.exs
```diff
@@ -545,6 +545,109 @@ defmodule Ecto.Adapters.PostgresTest do
...
+    query = from s in Schema,
+              window: (w as partition_by s.x, order_by: [s.a, s.b]),
+              select: (count(s.x) |> over(w))
+    query = query |> plan()
+    assert all(query) == ~s{SELECT count(s0."x") OVER w FROM "schema" AS s0 WINDOW w AS (PARTITION BY s0."x" ORDER BY s0."a", s0."b")}
+  end
+
+  test "count over unknown window" do
+    query = Schema
+            |> window([r], w as partition_by r.x)
```
josevalim: The use `as` as operator here is really smart but a bit foreign here since it is not an actual operator. My suggestion is the following:

```elixir
windows(r, w: partition_by(r.x))
```

i.e. rename the macro to `windows` and allow it to be a keyword list of windows names and their definitions. Similar to how `order_by` itself works. Thoughts?

lib/ecto/query/builder.ex
Anber: If I just remove this `escape` function, we will get "is not a valid query expression."
`row_number` is a kind of aggregation function, but it can be invoked only over a window, so we can't just mix it with regular aggregators. And there are a bunch of functions like this (`rank`, `dense_rank`, `first_value`, etc) which also have to be implemented. Should I add a `call_type` instead of `escape` for them?

btw, do you have any plans to extend the list of aggregate functions?

test/ecto/adapters/postgres_test.exs
Anber: I tried to find a clear syntax for `WINDOW w AS …`, but I missed this obvious variant. Thank you!

lib/ecto/query/builder.ex
josevalim: > Should I add a call_type instead of escape for them?

You are right. `call_type` is the way to go here.

> btw, do you have any plans to extend the list of aggregate functions?

@Anber we could add more functions but it would be nice to know how they are handled across databases. Which ones are supported by all databases (PG, MySQL and MSSQL)? Which ones are PG specific?

Anber: > Which ones are supported by all databases

Maybe it can be made extendable? DB driver (or even user's code) can provide supported functions to the builder.

josevalim: The thing is that we need the type information to be able to use in the query syntax correctly. So I think for now we should support what is common. Folks can use fragments for everything else.

Anber:
@josevalim,
I hope I've fixed all the mentioned issues, but maybe I didn't catch your idea about `partition_by`. I've removed it from `Ecto.Query`, but this syntax is still available:
```elixir
fields = [:y, :z]
Schema |> select([r], count(r.x) |> over(partition_by(r.x, order_by: ^fields)))
```

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -380,6 +381,22 @@ if Code.ensure_loaded?(Postgrex) do
        end)]
     end
 
+    defp window(%Query{windows: []}, _sources), do: []
+    defp window(%Query{windows: windows} = query, sources) do
+      [" WINDOW " |
+       intersperse_map(windows, ", ", fn
+         {name, definition} ->
+           [Atom.to_string(name), " AS ", partition_by(definition, sources, query)] end)]
```
josevalim: We need to use `quoted_name` or similar instead of `Atom.to_string(name)` to avoid SQL injection attacks.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -380,6 +381,22 @@ if Code.ensure_loaded?(Postgrex) do
...
+    end
+
+    defp partition_by(%PartitionByExpr{fields: fields, order_bys: order_bys}, sources, %Query{} = query) do
+      partition_query = %Query{ query | order_bys: order_bys }
```
josevalim: No spaces after `{` and before `}`.

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -509,6 +526,25 @@ if Code.ensure_loaded?(Postgrex) do
...
+      [aggregate, " OVER ", partition_by(window, sources, query)]
+    end
+
+    defp expr({:over, _, [agg, nil]}, sources, query) do
+      aggregate = expr(agg, sources, query)
+      [aggregate, " OVER ()"]
+    end
+
+    defp expr({:over, _, [agg, name]}, sources, %Ecto.Query{ windows: windows } = query) when is_atom(name) do
+      if Keyword.has_key?(windows, name) do
```
josevalim: This validation should happen in Ecto.Query.Planner. The adapter should always assume that query is correct (unless it doesn't support a particular construct).

lib/ecto/adapters/postgres/connection.ex
```diff
@@ -509,6 +526,25 @@ if Code.ensure_loaded?(Postgrex) do
...
+        aggregate = expr(agg, sources, query)
+        [aggregate, " OVER ", Atom.to_string(name)]
```
josevalim: We need to quote the `name` here too.

lib/ecto/query/api.ex
```diff
@@ -176,6 +176,11 @@ defmodule Ecto.Query.API do
   """
   def coalesce(value, expr), do: doc! [value, expr]
 
+  ## FIXME doc
+  @doc """
+  """
+  def over(value, window), do: doc! [value, window]
```
josevalim: We will also need to document `partition_by` here too. :)

lib/ecto/query/builder.ex
```diff
@@ -282,6 +282,32 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:coalesce, [], [left, right]]}, params_acc}
   end
 
+  def escape({:over, _, [aggregate]}, type, params_acc, vars, env) do
+    {aggregate, params_acc} = escape(aggregate, type, params_acc, vars, env)
+    {{:{}, [], [:over, [], [aggregate, nil]]}, params_acc}
+  end
+
+  def escape({:over, _, [aggregate, {window_name, _, nil}]}, type, params_acc, vars, env) when is_atom(window_name) do
```
josevalim: I am wondering... now that we changed to `windows`, should we make the window name an atom here instead of a binding? So it would be `over(count(p.foo), :window_name)`?

josevalim: I see that you support the atom clause below, so maybe it is a matter of getting rid of this one.

lib/ecto/query/builder.ex
```diff
@@ -463,6 +489,8 @@ defmodule Ecto.Query.Builder do
   defp merge_fragments([h1], []),
     do: [{:raw, h1}]
 
+  @over_agg_1 ~w(row_number rank dense_rank percent_rank cume_dist)a
```
josevalim: Some of those return integers, don't they? for those, I think the correct return would be `{:any, :integer}`. We will also need to document those as aggregates in `Ecto.Query.API`.

lib/ecto/query/builder/partition_by.ex
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  @doc """
+  Builds a partition by expression.
+  """
+  @spec build([Macro.t], Tuple.t, Macro.Env.t) :: Macro.t
+  def build(binding, expr, env) do
+    {fields, opts} = escape(expr)
+    {fields, {_, :acc}} = Builder.escape(fields, :any, {%{}, :acc}, binding, env)
+    order_bys = order_by(binding, opts, env)
```
josevalim: I think this could be `Enum.map(opts, &order_by(&1, binding, env)` and we would be able to remove the loop from `order_by`?

lib/ecto/query/builder/windows.ex
```diff
@@ -0,0 +1,46 @@
...
+  @spec build(Macro.t, [Macro.t], Keyword.t, Macro.Env.t) :: Macro.t
+  def build(query, binding, windows, env) do
+    {_, binding} = Builder.escape_binding(query, binding, env)
+    windows = windows
+              |> Enum.map(fn {name, expr} -> {name, PartitionBy.build(binding, expr, env)} end)
+
+    Builder.apply_query(query, __MODULE__, [windows], env)
+  end
+
+  @spec apply(Ecto.Queryable.t, Tuple.t) :: Ecto.Query.t
```
josevalim: `@spec` in the wrong place. :)

lib/ecto/query/builder/windows.ex
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  @doc """
+  The callback applied by `build/4` to build the query.
+  """
+  @spec apply(Ecto.Queryable.t, Keyword.t) :: Ecto.Query.t
+  def apply(%Ecto.Query{} = query, definitions) do
+    definitions
+    |> Enum.reduce(query, &apply_window/2)
```
josevalim: I would prefer something like:

```elixir
def apply(%{windows: windows} = query, definitions) do
  %{query | windows: windows ++ Enum.map(definitions, &validate_window!(&1, windows))}
end
```

Otherwise we are accessing the query and prepending the list of windows multiple times.

josevalim:
Hi @Anber! I have done a more careful review now. They are mostly minor details but there is one big issue: you have not changed `Ecto.Query.Planner` to traverse `:windows`. The main consequence of not doing this is that any binding inside `:windows` won't be considered. So I think you should write some queries that has parameters inside the window, maybe something silly in the order_by field, such as `order_by: [desc: ^param]`.

In order to make this work, there are a couple things you need to have in mind:

1. You need to make sure that you are consistent in the order things are processed. So if you change the planner to traverse windows immediately after `from`, then the drivers need to do it exactly after `from` too. See `traverse_exprs` in the planner.

2. I am afraid the fact you are using order_bys inside PartitionByExpr is going to be more trouble than help. That's because order_bys will have their own bindings and you need to make sure they are traversed and changed accordingly. It is probably better if PartitionByExpr simply calls `OrderBy.escape` to do the escape work but you keep them as regular ASTs inside PartitionByExpr and not as full blown Ecto.Query.QueryExpr. I would possibly give this a try before doing any other changes.

I am really excited about this! If you have any questions, please let me know!

josevalim:
I have updated the list above. :)

Anber:
It turned out to be a bit more complicated than I thought :)
I have added the checklist and I will try to finish it in several days.

lib/ecto/query/builder.ex
```diff
@@ -266,6 +266,33 @@ defmodule Ecto.Query.Builder do
     {expr, params_acc}
   end
 
+  @over_agg %{
+    row_number: [:integer],
+    rank: [:integer],
+    dense_rank: [:integer],
+    percent_rank: [:float],
+    cume_dist: [:float],
+    ntile: [:integer, :integer], # TODO: what should we do with functions which require exactly integer type?
```
Anber: Hi @josevalim! I have some questions here. 
The first question is about `integer` and `bigint`. In some cases, they are interchangeable but if function requires `integer` we can't just cast a value to `bigint`.

lib/ecto/query/builder.ex
```diff
@@ -282,6 +309,17 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:coalesce, [], [left, right]]}, params_acc}
   end
 
+  def escape({:over, _, [aggregate | args]}, type, {params, acc}, vars, env) do
+    {aggregate, {params, acc}} = escape(aggregate, type, {params, Map.put(acc, :over, true) }, vars, env)
```
Anber: The second question is about `acc` argument of `escape/5`. I want to know whether window function is called inside `over` or not. So, I use `acc` here for storing a flag which shows that we are in `over` scope.

lib/ecto/query/builder.ex
```diff
@@ -419,12 +457,34 @@ defmodule Ecto.Query.Builder do
...
+    do: {window_name, params_acc}
+  defp escape_window([partition_by], type, params_acc, vars, env),
+    do: escape(partition_by, type, params_acc, vars, env)
+
+  defp escape_call({_, _, args} = call, type, params, vars, env) when is_atom(type) do
+    arity = length(args)
+    escape_call(call, List.duplicate(type, arity), params, vars, env)
+  end
+
+  defp escape_call({name, _, args}, types, params, vars, env) when is_list(types) and length(types) >= length(args) do
```
Anber: I have also added support of function with different arity. If you approve it, we can replace tuples [here](https://github.com/Anber/ecto/blob/7c9096c4a3622977b96c78fabf964e590fb85435/lib/ecto/query/builder.ex#L526)  to lists like `[out_type | in_types]`.

lib/ecto/query/builder.ex
josevalim: Before we get to do that: why do you need to know if the window function is called inside `over` or not? :)

lib/ecto/query/builder.ex
josevalim: @Anber we don't make a distinction from this side of things, so just `:integer` should be enough.

lib/ecto/query/builder.ex
josevalim: @Anber I don't think we should generalize the `length(types) >= length(args)` as we are getting a behaviour specific to two functions and generalizing it to the whole codebase. This is bound to cause issues further down the road. I think you should move the [zip logic to before you call escape_call](https://github.com/elixir-ecto/ecto/pull/2618/files#diff-c04711315c01e29757522ea2764fd871R287).

josevalim: In fact, I would prefer if for now we only added `row_number` and the `*_rank` functions, treat them as aggregators and keep this simple for now. Otherwise there is a chance for this PR to only continue growing in size, making it harder to review and merge. :)

lib/ecto/query/builder.ex
Anber: We don't, but queries like `Schema |> select([r], nth_value(r.x, "42") |> over)` throw a error `function nth_value(integer, bigint) does not exist`.

lib/ecto/query/builder.ex
Anber: Because window function without `over` can't be called? :)

lib/ecto/query/builder.ex
Anber: Actually, it doesn't matter how many window functions we will implement, because even `row_number` has zero arity and it can't be implemented by using current `call_type`.

lib/ecto/query/builder.ex
josevalim: We don't need to check this. Let the SQL engine fail later on. :)

lib/ecto/query/builder.ex
josevalim: Right. You need to cast or pass a proper value. There is not much we can do here because Postgres is thinking that it is expecting a bigint.

lib/ecto/query/builder.ex
josevalim: @Anber I see. So maybe we should do this:

1. Change call_type to return `{[in], out}`

2. Have a specific annotation that tells all functions with variadic arity. For example `@varargs [:lag, :lead, :row_number]`

What do you think?

josevalim: To be more precise, I am just worried about always enabling this behaviour. But if we put it behind a `@varargs` flag, my concerns are addressed and I will be happy. :)

lib/ecto/query/builder.ex
Anber: Why not? We can prevent runtime error just by adding a few lines of code.

lib/ecto/query/builder.ex
Anber: I agree that we can't do anything here right now, but maybe there is a possibility to add more strict type system in the future?

lib/ecto/query/builder.ex
Anber: Modification of `call_type` definitely looks better than my solution, but I didn't want to touch it without your blessing :)
Maybe I can just modify my `@over_agg` and use it in `call_type`? 
```elixir
  defp call_type(agg, 1) when agg in @over_agg_names, do: Map.get(@over_agg, agg)
```

lib/ecto/query/builder.ex
josevalim: If this is the goal, then it is better to not mixture those functions with `escape_call` at all and explicitly check for them in the `over` clause. We cannot use the accumulator because the accumulator can be of any type.

lib/ecto/query/builder.ex
josevalim: @Anber that's a linear cost. We should implement the proper `call_type/2` clauses and not rely on the map.

However, given the previous discussion about over+window functions, maybe we shouldn't use `call_type` at all and keep them completely distinct, as they seem to have a bunch of behaviour that is specific to them.

lib/ecto/query/builder.ex
Anber: It can't be checked in `over` clause, because there is no `over` in case of error  :)
I don't see any usages of `acc`. How and when it can be used? Maybe I can just override `acc` for `over` branch of AST and restore it to the original value before exit from `escape(:over, …)`?

josevalim: > It can't be checked in over clause, because there is no over in case of error :)

What I meant is that we will only traverse those functions if we are inside over. So over should not call `escape` on the first argument, it should rather call `escape_window_function` or something similar.

> I don't see any usages of acc. How and when it can be used?

@Anber the accumulator is used by `select` to collect take/2 calls. So please don't change the semantics of the accumulator. It is used in some places to collect information about the AST being traversed. Even if it is not an issue now, it may be an issue in the future.

lib/ecto/query/builder.ex
Anber: > that's a linear cost

And what if we unfold call_type on compile-time?
```elixir
  for {agg, types} <- @over_agg do
    defp call_type(unquote(agg), 1), do: types
  end
```

lib/ecto/query/builder.ex
Anber: So, it's not a context. I've got it. Also, your solution with `escape_window_function` sounds better. Thank you :)

lib/ecto/query/builder.ex
josevalim: That's fine but given the solution above, I think we won't be overriding call_type anyway, will we?

Anber: Looks like we won't.
Thank you for your patience!

Anber:
Hi @josevalim!

> They are mostly minor details but there is one big issue

I hope I have fixed it, but I don't like how I escape windows. Typically macros are injected to the AST on escape stage for expanding to `%QueryExpr{}` on the later stages, but I don't have a `partition_by` macro, so I just return `%QueryExpr{}` from `escape_window`. It works, but I can't test it clearly in `builder_test.exs`.

josevalim:
@Anber looks great to me! Now we need to go ahead with MySQL support and docs. :)

Anber:
Done!

lib/ecto/adapters/mysql/connection.ex
```diff
@@ -320,13 +321,34 @@ if Code.ensure_loaded?(Mariaex) do
        end)]
     end
 
+    defp window(%Query{windows: []}, _sources), do: []
+    defp window(%Query{windows: windows} = query, sources) do
+      [" WINDOW " |
+       intersperse_map(windows, ", ", fn
+         {name, definition} ->
+           [quote_name(name), " AS ", partition_by(definition, sources, query)] end)]
```
michalmuskala: This is formatter a bit weirdly. Can we do:
```
```

michalmuskala: This is formatter a bit weirdly. Can we do:
```elixir
[" WINDOW " |
  intersperse_map(windows, ", ", fn {name, definition} ->
```

Anber:
Hi @josevalim 
Now we have an ability to fix preloader behavior which mentioned in the note [here](https://hexdocs.pm/ecto/Ecto.Query.html#preload/3-preload-queries), but it will be a breaking change. May I fix it?

josevalim:
@Anber shall we discuss it in another issue? I also think this behaviour may been fixed on master.

Btw, we should merge this PR next week, thanks! :)

Anber:
@josevalim It definitely will be another issue and another PR :)

Anber:
Hi @josevalim
When you said that we should merge this PR, did you mean that I forgot to do something? :)

hauleth:
@Anber damn you. This is awesome PR and obsoletes all my work that I have done on supporting window functions via fragments. Now I need to focus on adding grouping sets to Ecto.

Anber:
@hauleth, we also tried to use fragments, but then we faced this [issue](https://github.com/elixir-ecto/ecto/issues/2281#issuecomment-357195145).

hauleth:
@Anber [my solution worked for me](https://github.com/hauleth/ecto_olap/blob/622042722394e0e26bf259a12fa56e0abb9fd0a4/lib/window/over.ex)

Anber:
@hauleth btw, I haven't implemented `OVER (ORDER BY …)` and `OVER (RANGE …)` yet. This PR became too big, so I decided to do it in another one.

Anber:
Hi @josevalim!
Do you have any news? :)

josevalim:
I have started working on this on a branch and it should be merged later today. :)

lib/ecto/query/planner.ex
```diff
@@ -577,6 +577,20 @@ defmodule Ecto.Query.Planner do
     end
   end
 
+  defp merge_cache(:windows, query, exprs, {cache, params}, adapter) do
+    {expr_cache, {params, cacheable?}} =
+      Enum.map_reduce exprs, {params, true}, fn {_, expr}, {params, cacheable?} ->
+        {params, current_cacheable?} = cast_and_merge_params(:windows, query, expr, params, adapter)
+        {expr_to_cache(expr), {params, cacheable? and current_cacheable?}}
```
josevalim: Key must be part of the cache.

lib/ecto/query/planner.ex
```diff
@@ -1023,6 +1049,29 @@ defmodule Ecto.Query.Planner do
     {type, [expr | fields], from}
   end
 
+  # OVER ()
+  defp collect_fields({:over, _, [call, nil]} = expr, fields, from, query, take) do
```
josevalim: We should probably make this arity 1.

test/ecto/query/builder_test.exs
```diff
@@ -115,6 +126,14 @@ defmodule Ecto.Query.BuilderTest do
                  fn ->
       escape(quote(do: Foo.bar(x)), [x: 0], __ENV__) |> elem(0) |> Code.eval_quoted([], __ENV__)
     end
+
+    assert_raise Ecto.Query.CompileError, ~r"lag/2 must be invoked using window function syntax", fn ->
+      escape(quote(do: lag(:a, 1)), [], __ENV__)
+    end
+
+    assert_raise Ecto.Query.CompileError, ~r"window function lag/0 is undefined.", fn ->
```
josevalim: Remove trailing dot.

test/ecto/adapters/postgres_test.exs
```diff
@@ -544,6 +544,112 @@ defmodule Ecto.Adapters.PostgresTest do
...
+    query = from s in Schema,
+                 windows: [w1: partition_by(s.a)],
+                 windows: [w2: partition_by(s.x, order_by: s.a, order_by: [desc: ^param])],
+                 select: (count(s.x) |> over(:w2))
+
+    query = query |> plan()
+    assert all(query) == ~s{SELECT count(s0."x") OVER "w2" FROM "schema" AS s0 WINDOW "w1" AS (PARTITION BY s0."a"), "w2" AS (PARTITION BY s0."x" ORDER BY s0."a", s0."b" DESC)}
+  end
+
+  test "count over unknown window" do
```
josevalim: This should be mover to the planner suite.

test/ecto/adapters/mysql_test.exs
```diff
@@ -438,6 +438,112 @@ defmodule Ecto.Adapters.MySQLTest do
...
+    query = from s in Schema,
+                 windows: [w1: partition_by(s.a)],
+                 windows: [w2: partition_by(s.x, order_by: s.a, order_by: [desc: ^param])],
+                 select: (count(s.x) |> over(:w2))
+
+    query = query |> plan()
+    assert all(query) == ~s{SELECT count(s0.`x`) OVER `w2` FROM `schema` AS s0 WINDOW `w1` AS (PARTITION BY s0.`a`), `w2` AS (PARTITION BY s0.`x` ORDER BY s0.`a`, s0.`b` DESC)}
+  end
+
+  test "count over unknown window" do
```
josevalim: This should be mover to the planner suite.

josevalim:
Thank you @Anber! I have added some notes but I can tackle them locally as they are minor. Amazing work overall!

josevalim:

josevalim:
Hi @Anber! I have noticed that SQL databases allow you to specify `order_by` without a `partition_by`, so for consistency, i have converted all of them into keyword lists. So now you specify:

    windows: [w: [partition_by: table.x, order_by: table.y]]

This looks very nice with over:

    over(agv(table.salary), partition_by: table.x, order_by: table.y)

So if you want to go ahead and add support for ranges, the infrastructure is there and we can start this discussion. The only thing left from this PR is to allow dynamic fields in `partition_by` and improve the coverage in the integration tests.

PR #2622, opened 2018-07-17
Fix Mix.Ecto.ensure_started/2 specs

barthez:
Function can return pid or nil as the second item in the 3-elements tuple.

I have used this function in my project and copied `pid && repo.stop(pid)` pattern from migration task. Dialyzer was complaining that `pid` cannot be false.


josevalim:

wojtekmach:
Worth mentioning this module is private API, expect breaking changes any time in your app and definitely dont use it in libraries.

PR #2625, opened 2018-07-23
Parsing "infinity" in Repo url config

fbergero:
With these changes an user can specify `"infinity"` for `pool_timeout` and `timeout` in Repo's URL config.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -64,7 +64,7 @@ defmodule Ecto.Repo do
   options `ssl`, `timeout`, `pool_timeout`, `pool_size`:
 
     config :my_app, Repo,
-        url: "ecto://postgres:postgres@localhost/ecto_simple?ssl=true&pool_size=10"
+        url: "ecto://postgres:postgres@localhost/ecto_simple?ssl=true&pool_size=10&timeout=infinity&pool_timeout=infinity"
 
   In case the URL needs to be dynamically configured, for example by
   reading a system environment variable, such can be done via the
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @fbergero. Thanks for the PR. What is the use case here?

Setting the timeout to infinity is generally speaking not a good idea.

fbergero:
Hi @josevalim I'm aware that setting the timeout to infinity is *definetely* not a good idea. Ecto already allows this. I'm only enabling using that option from the URL config.

josevalim:
I am OK with letting you be explicit this in your configuration rather than doing it via the URL. :)

josevalim:
That's always the case though, as you can't pass everything in the URL. We can wait for practical use cases on this feature. For now, it seems it exists rather for completeness and I don't think we need to be consistent here. Thanks!

fbergero:
I see an inconsistency here:
* `Ecto.Repo` allows to set the timeouts from the URL as stated in the docs
> URL can include query parameters to override shared and adapter-specific options ssl, timeout, pool_timeout, pool_size
* The timeouts are an integer or the `:infinity` atom
> :timeout - The time in milliseconds to wait for the query call to finish, :infinity will wait indefinitely (default: 15000);
:pool_timeout - The time in milliseconds to wait for calls to the pool to finish, :infinity will wait indefinitely (default: 5000);
* But the `:infinity` option is not accepted from the URL.

josevalim:
Exactly. But as said:

> For now, it seems it exists rather for completeness and I don't think we need to be consistent here

Arguing for consistency here is like saying you have to shoot your right foot only because you shot your left foot earlier. :)

PR #2626, opened 2018-07-25
Allows `--protocol` to be configurable for the `mysql` cmd

fredwu:
Hi,

Today I ran into an issue that took me hours to figure out. I was trying to use `mix ecto.load` to load a schema file, after seeing the file's successfully loaded message, I checked the database and saw nothing was imported.

I dug into the code to see what goes behind the scenes of `ecto.load`, and noticed that it uses the `mysql` command. So I ran the `mysql` command manually and it imported correctly.

Then I started debugging what exactly was the command generated by the ecto's mysql adapter. It turns out, the `--protocol=tcp` part was the culprit. As for my local mysql installation, it uses socket to connect.

I'm not sure why `--protocol=tcp` was hard-coded, so I'm going to go out on a limb and guess that it was an oversight?

This PR retains the `--protocol=tcp` behaviour, but allows users to configure it to other values (`tcp`, `memory`, `pipe`, etc).

I couldn't figure out a way to add tests (there are no tests for testing adapter's load/dump behaviour, other than a few for a mocked version) so I'm going to simply make these straightforward changes. Please let me know if you have any suggestions and I'm more than happy to look into it.

Thanks!

### DIFF
--- a/lib/ecto/adapters/mysql.ex
+++ b/lib/ecto/adapters/mysql.ex
@@ -30,6 +30,7 @@ defmodule Ecto.Adapters.MySQL do
     * `:port` - Server port (default: 3306)
     * `:username` - Username
     * `:password` - User password
+    * `:protocol` - Protocol used for the mysql client connection (default: tcp)
     * `:ssl` - Set to true if ssl should be used (default: false)
     * `:ssl_opts` - A list of ssl options, see Erlang's `ssl` docs
     * `:parameters` - Keyword list of connection parameters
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/mysql.ex
```diff
@@ -30,6 +30,7 @@ defmodule Ecto.Adapters.MySQL do
     * `:port` - Server port (default: 3306)
     * `:username` - Username
     * `:password` - User password
+    * `:protocol` - Protocol used for the mysql client connection (default: tcp)
```
josevalim: We should clarify that this is only used for the ecto.create and ecto.drop commands. :)

fredwu: Correct me if I'm wrong, but it's only used for `ecto.load` and `ecto.dump` AFAICS?

josevalim: Yes, sorry. :)

fredwu: Cool, just pushed up the clarification. :)

michalmuskala:
Would it be maybe better if we didn't have a repo-wide configuration for this, but instead allowed passing an option to the `ecto.load` and `ecto.dump` tasks? 

It seems to me a `protocol` is a fairly "wide" option we might want to use for something - if anything, actually connection to the database daemon via the unix sockets.

It should be also pretty easy to make it a default with mix aliases for tasks.

fredwu:
@michalmuskala The only thing is that this option is `mysql` specific, and the current options for `ecto.load` are all non-adapter specific.

michalmuskala:
Ah you're right. It doesn't make sense to add it there. Could we maybe rename this option to `cli_protocol` or something similar? What do you think?

fredwu:
Sure thing, updated!

josevalim:

PR #2635, opened 2018-08-02
fix brocken link in "GettingStarted.md"

apre:
Github markdown dosn't support well hyperlink with space in the filename.
Replaced the space with CamelCase style.


apre:
Sure, 
I thought that would also fix the broken link in
https://hexdocs.pm/ecto/getting-started.html#test-environment-setup
after regenerating the doc. 
However I am not sure it will fix it.

josevalim:
Thanks @apre but as @fertapric said, the links shuold be cased. The version live will fixed in the next release!

PR #2650, opened 2018-08-09
Update Changeset Documentation

mmacheerpuppy:
Clarified changeset documentation with valid code...

Changes include...

+ The documentation conflated 'Elixir structs' and 'structs generated by an Ecto schema'. Whilst in the underlying implementation they are equivalent, I've provided the improvement of making the documentation more beginner friendly by making it more explicit that you really can just feed it a valid struct (you aren't forced to use an Ecto schema, that then generates a struct)! This will be helpful to newcomers who aren't familiar with Ecto schemas and have just read the basic documentation and want to get started with changesets. I assume _if_ the newcomers read the Ecto documentation on Ecto schemas, then they will understand the Ecto schema will generate a regular Elixir struct (with the value of added metadata), just not necessarily the other way around (which is certainly a challenge I faced when I wanted to jump straight into changesets). 

+ Within the previous examples, Ecto.Changeset would not accept the user input as expecting a map (params). I've provided the improvement of providing valid examples (which I tested in iex to make sure 🤓). Now our awesome Ecto athletes can copy and paste from the examples to see what working code looks like and go from there!

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -174,20 +174,54 @@ defmodule Ecto.Changeset do
   ## Schemaless changesets
 
   In the changeset examples so far, we have always used changesets to
-  validate and cast data contained in a struct, such as the `%User{}`
-  struct defined by the `User` module.
-
-  However, changesets can also be used with data in a plain map, by
+  validate and cast data contained in a struct defined by an Ecto Schema, such as the `%User{}`
+  struct defined by the `User` module. 
+  
+  However, changesets can also be used without defining our structs using an Ecto schema:
+  
+      user = %User{}
+      types = %{first_name: :string, last_name: :string, email: :string}
+      changeset =
+        {user, types}
+        |> Ecto.Changeset.cast(params["name"], Map.keys(types))      
+      #   |> Ecto.Changeset.validate_required(...)
+      #   |> Ecto.Changeset.validate_length(...)
+                
+  Where the user struct refers to the definition in the following module. 
+  
+    defmodule User do
+      defstruct [:name, :age]
+    end 
+  
+  Returning the Changeset as a response:
+  
+    #Ecto.Changeset<
+    action: nil,
+    changes: %{name: "Callum"},
+    errors: [],
+    data: #User<>,
+    valid?: true>
+        
+  Elixir structs are just bare maps underneath. Changesets can also be used with data in a plain map, by
   passing a tuple containing both the data and the supported types:
 
-      data  = %{}
-      types = %{first_name: :string, last_name: :string, email: :string}
-
-      changeset =
-        {data, types}
-        |> Ecto.Changeset.cast(params["sign_up"], Map.keys(types))
-        |> Ecto.Changeset.validate_required(...)
-        |> Ecto.Changeset.validate_length(...)
+    data  = %{}
+    types = %{name: :string}
+    params = %{name: "Callum"}
+    changeset =
+      {data, types}
+        |> Ecto.Changeset.cast(params, Map.keys(types))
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -174,20 +174,54 @@ defmodule Ecto.Changeset do
...
-  struct defined by the `User` module.
-
-  However, changesets can also be used with data in a plain map, by
-  passing a tuple containing both the data and the supported types:
-
-      data  = %{}
+  validate and cast data contained in a struct defined by an Ecto Schema, such as the `%User{}`
+  struct defined by the `User` module. 
+  
+  However, changesets can also be used without defining our structs using an Ecto schema:
```
fertapric: What do you think about?

```
However, changesets can also be used with data in a plain map, by
passing a tuple containing both the data and the supported types:

    data  = %{}
    types = %{name: :string}
    params = %{name: "Callum"}

    changeset =
      {data, types}
      |> Ecto.Changeset.cast(params, Map.keys(types))
      |> Ecto.Changeset.validate_required(:name)

Elixir structs are just bare maps underneath. For this reason, changesets
can also be used with any struct:

    defmodule User do
      defstruct [:name, :email, :age]
    end 

    user = %User{}
    types = %{name: :string, email: :string, age: :integer}
    params = %{name: "Callum", email: "callum@example.com"}

    changeset =
      {user, types}
      |> Ecto.Changeset.cast(params, Map.keys(types))
      |> Ecto.Changeset.validate_required(:email)

Such functionality makes Ecto extremely useful to cast, validate and prune
data even if it is not meant to be persisted to the database.
```

Note also that I change the `types` variable to match the ones in the `User` struct. I've also indented the examples with 4 spaces, so they are rendered as code :)

mmacheerpuppy: The change in structure offers a much more readable improvement. I would further add doctest `iex` statements to the examples, just as a written proof to readers "This code is guaranteed to work" but I'm wasn't too sure how well that would play with your repository.

mmacheerpuppy: I think you would have to give the structs literal definitions in order for that to work, because I don't think doctest allows you to define Elixir structs directly in your code.

mmacheerpuppy:
Update I think before pushing this it might be also worth mentioning the Schemaless changeset doesn't mean 'schemaless' in the sense that you can't know what your types are going to be in advance.

We find in the implementation...

```
defp cast_key(key) when is_binary(key) do
    try do
      {String.to_existing_atom(key), key}
    rescue
      ArgumentError ->
        raise ArgumentError, "could not convert the parameter `#{key}` into an atom, `#{key}` is not a schema field"
    end
end
```

Where if we try to cast types that aren't already existing atoms, then we get an error. But if our casting was _truly schemaless_ then we would be able to dynamically construct our types based on user input. For example...
```
data = # some source of user input data
changeset = {data, types}
      |> Ecto.Changeset.cast(params, Map.keys(types))
well_formed_types = # Some map of well defined types, each of these are atoms.
user_input_types = # Some source of user input, each of these are strings
types = Map.merge(required_types, Map.new(variables_list, fn variable -> {String.to_atom(variable), :string} end))
```

Imagining this use case people _could_ get confused by what we mean as 'schemaless'.  We don't want to mean "allow users to define types without a schema". We want to mean "allow users to define schemas using maps and not Ecto schemas, in particular". 

This might already be apparent in the documentation revisions, but worth keeping in mind.

michalmuskala:
You need to pass the list of allowed fields to `cast/3` so the atom has to exist by the time we reach the code calling `String.to_existing_atom` if the field is allowed in the cast.

josevalim:

PR #2651, opened 2018-08-09
Semantic comparisons in changesets

### LINKED ISSUE #2473: Decimal type always changes
* Elixir version (elixir -v): 1.6.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Mysql 5.7.18
* Ecto version (mix deps): 2.2.9
* Operating system: Mac OS Sierra

### Current behavior

When I inspect changesets for schemas with `:decimal` type, `:amount` is always in `changes` when its value did not change.

```
[
  #Ecto.Changeset<
    action: :update,
    changes: %{amount: #Decimal<599>},
    errors: [],
    data: #ZB.JournalEntryLine<>,
    valid?: true
  >,
  #Ecto.Changeset<
    action: :update,
    changes: %{amount: #Decimal<599>, chart_account_id: 7377718},
    errors: [],
    data: #ZB.JournalEntryLine<>,
    valid?: true
  >
]
```

### Expected behavior

In that case above, the only thing that should be in the changes is `chart_account_id: 7377718` cause its the only thing that is different

wojtekmach:
Fixes #2473 

* Add `Ecto.Type.equal?/3` callback
* Add `Ecto.Type.equal?/2` function

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -918,9 +918,8 @@ defmodule Ecto.Type do
     do: requires_semantic_comparison?(type)
   defp requires_semantic_comparison?({:map, type}),
     do: requires_semantic_comparison?(type)
-  defp requires_semantic_comparison?(module) when is_atom(module) do
-    function_exported?(module, :equal?, 2)
-  end
+  defp requires_semantic_comparison?(module) when is_atom(module),
+    do: loaded_and_exported?(module, :equal?, 2)
 
   defp do_equal?(:decimal, %Decimal{} = term1, %Decimal{} = term2) do
     Decimal.equal?(term1, term2)
@@ -1023,6 +1022,11 @@ defmodule Ecto.Type do
       {int, ""} -> int
       _ -> nil
     end
+  end
+
+  defp loaded_and_exported?(module, fun, arity) do
+    # TODO: Rely only on Code.ensure_loaded? when targetting Erlang/OTP 21+
+    (:erlang.module_loaded(module) or Code.ensure_loaded?(module)) and function_exported?(module, fun, arity)
   end
 
   defp validate_decimal({:ok, %Decimal{coef: coef}}) when coef in [:inf, :qNaN, :sNaN],
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -94,12 +94,14 @@ defmodule Ecto.Type do
                       :utc_datetime_usec | :naive_datetime_usec | :time_usec
   @typep composite :: {:array, t} | {:map, t} | {:embed, Ecto.Embedded.t} | {:in, t}
 
-  @base ~w(
-    integer float boolean string map
-    binary decimal id binary_id
-    utc_datetime naive_datetime date time any
+  @calendar ~w(
+    utc_datetime naive_datetime date time
     utc_datetime_usec naive_datetime_usec time_usec
   )a
+  @base ~w(
+    integer float boolean string map
+    binary decimal id binary_id any
+  )a ++ @calendar
   @composite ~w(array map in embed)a
 
   @doc """
@@ -144,6 +146,13 @@ defmodule Ecto.Type do
   and it needs to validate them and convert it to an Ecto native type.
   """
   @callback dump(term) :: {:ok, term} | :error
+
+  @doc """
+  Checks if two terms are semantically equal.
+  """
+  @callback equal?(term, term) :: boolean
+
+  @optional_callbacks [equal?: 2]
 
   ## Functions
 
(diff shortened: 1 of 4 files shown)


lib/ecto/type.ex
```diff
@@ -145,6 +147,13 @@ defmodule Ecto.Type do
   """
   @callback dump(term) :: {:ok, term} | :error
 
+  @doc """
+  Checks if two terms are semantically equal.
+  """
+  @callback equal?(term, term) :: boolean
```
wojtekmach: other callbacks have `{:ok, term} | :error` return, not sure if we need it here too

lib/ecto/type.ex
```diff
@@ -875,6 +884,94 @@ defmodule Ecto.Type do
...
+  end
+  defp do_equal?(type, %DateTime{} = term1, %DateTime{} = term2)
+       when type in [:utc_datetime, :utc_datetime_usec] do
+    DateTime.compare(term1, term2) == :eq
+  end
+  defp do_equal?(type, term1, term2) when type in @base do
+    term1 == term2
+  end
+  defp do_equal?(module, term1, term2) when is_atom(module) do
+    module.equal?(term1, term2)
```
fertapric: Could `requires_semantic_comparison?/1` be removed if this part is changed to?

```elixir
defp do_equal?(module, term1, term2) when is_atom(module) do
  if function_exported?(module, :equal?, 2) do
    module.equal?(term1, term2)
  else
    term1 == term2
  end
end
```

In that case, `do_equal?` could be renamed to `equal?`

wojtekmach: No because this wont handle typed arrays and maps.

wojtekmach: Ah, this probably would work as you described; we are introducing the extra check to avoid semantical comparison unless we know it’s required as an optimization - vast majority of cases its not needed.

fertapric: Could we get similar performance benefits just be re-arranging the clauses?

```elixir
def equal?(type, term1, term2) when type in @base do
  term1 == term2
end
def equal?({:array, type}, xs, ys) do
  equal_list?(type, xs, ys)
end
def equal?({:map, _}, map1, map2) when map_size(map1) != map_size(map2) do
  false
end
def equal?({:map, type}, map1, map2) do
  equal_map?(type, Map.to_list(map1), Map.to_list(map2))
end
def equal?(type, %Time{} = term1, %Time{} = term2) when type in [:time, :time_usec] do
  Time.compare(term1, term2) == :eq
end
def equal?(type, %NaiveDateTime{} = term1, %NaiveDateTime{} = term2)
     when type in [:naive_datetime, :naive_datetime_usec] do
  NaiveDateTime.compare(term1, term2) == :eq
end
def equal?(type, %DateTime{} = term1, %DateTime{} = term2)
     when type in [:utc_datetime, :utc_datetime_usec] do
  DateTime.compare(term1, term2) == :eq
end
def equal?(:decimal, %Decimal{} = term1, %Decimal{} = term2) do
  Decimal.equal?(term1, term2)
end
def equal?(module, term1, term2) when is_atom(module) do
  if function_exported?(module, :equal?, 2) do
    module.equal?(term1, term2)
  else
    term1 == term2
  end
end
```

wojtekmach: The code above does unnecessary comparisons for eg `{:array, :integer}` but I see where youre going with this, I will investigate.

wojtekmach: @fertapric yes, that close (fyi: decimal and calendar types are in `@base` but that's a minor detail that's easy to fix)

We would still be inefficient for nested composites like `{:array, {:array, integer()}}` but to be fair that should be rather uncommon.

@josevalim do you have a preference?

lib/ecto/type.ex
```diff
@@ -875,6 +884,94 @@ defmodule Ecto.Type do
...
+  defp requires_semantic_comparison?(type) when type in @calendar,
+    do: true
+  defp requires_semantic_comparison?(type) when type in @base,
+    do: false
+  defp requires_semantic_comparison?({:array, type}),
+    do: requires_semantic_comparison?(type)
+  defp requires_semantic_comparison?({:map, type}),
+    do: requires_semantic_comparison?(type)
+  defp requires_semantic_comparison?(module) when is_atom(module) do
+    function_exported?(module, :equal?, 2)
```
josevalim: This will return false if the module is not loaded and I think there is no guarantee the module will be loaded by then.

lib/ecto/type.ex
```diff
@@ -875,6 +884,93 @@ defmodule Ecto.Type do
...
+  defp requires_semantic_comparison?(type) when type in @calendar,
+    do: true
+  defp requires_semantic_comparison?(type) when type in @base,
+    do: false
+  defp requires_semantic_comparison?({:array, type}),
+    do: requires_semantic_comparison?(type)
+  defp requires_semantic_comparison?({:map, type}),
+    do: requires_semantic_comparison?(type)
+  defp requires_semantic_comparison?(module) when is_atom(module),
+    do: loaded_and_exported?(module, :equal?, 2)
```
josevalim: I pushed a fix for it.

josevalim:

Issue #2648, opened 2018-08-09
Ecto.Schema.load!/5 cannot load postgres "timestamp with time zone" fields


nimf:
### Environment

* Elixir version (elixir -v):
Erlang/OTP 20 [erts-9.3.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.6.5 (compiled with OTP 20)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):
PostgreSQL 9.6.8

* Ecto version (mix deps):
```
[
  {:ecto, git: "https://github.com/elixir-ecto/ecto.git", ref: "master"},
  {:postgrex, "~> 0.14.0-dev", git: "https://github.com/elixir-ecto/postgrex.git"},
  {:db_connection, "~> 1.1",
   override: true, git: "https://github.com/elixir-ecto/db_connection.git", ref: "4947966"}
]
```

* Database adapter and version (mix deps):
  {:postgrex, "~> 0.14.0-dev", git: "https://github.com/elixir-ecto/postgrex.git"},

* Operating system:
macOS 10.12.6 (Sierra)

### Current behavior

When the table created externally (not by ecto migrations) and it contains "timestamp with time zone" fields they can't be loaded into Ecto.Schema.

```elixir
defmodule MyApp.NotOk do
  use Ecto.Schema

  schema "not_oks" do
    field :created_at, :utc_datetime
  end
end
```

```sql
PostgreSQL

create table not_oks (id integer, created_at timestamp with time zone);
insert into not_oks(id, created_at) values(1, current_timestamp);
```

```
iex(6)> MyApp.NotOk |> MyApp.Repo.all

10:54:50.069 [debug] QUERY OK source="not_oks" db=3.3ms
SELECT n0."id", n0."created_at" FROM "not_oks" AS n0 []
** (ArgumentError) cannot load `#DateTime<2018-08-09 04:44:57.364987Z>` as type :utc_datetime for field `created_at` in schema MyApp.NotOk
    (ecto) lib/ecto/schema.ex:1794: Ecto.Schema.load!/5
    (ecto) lib/ecto/schema.ex:1742: Ecto.Schema.safe_load_zip/4
    (ecto) lib/ecto/schema.ex:1743: Ecto.Schema.safe_load_zip/4
[...]

PR #2653, opened 2018-08-13
Loading PG timestamp with time zone into utc_datetime

nimf:
This is my take on #2648 

It allows loading 'timestamp with time zone' fields (for example if a table was created manually) with Ecto.

The problem is that when we have 'timestamp with time zone' field, the adapter returns it as DateTime, not as NaiveDateTime and `lib/ecto/type.ex` has no matching function to load DateTime.

From my observation, when we put a non-UTC value into timestamp with time zone, it gets read as UTC by the adapter. However, maybe we should add an additional check that it is UTC and convert to UTC if not? Just how it is done here: https://github.com/elixir-ecto/ecto/blob/022f244aa4839d61dc4cf413813e5d02fd4fbd64/lib/ecto/type.ex#L860-L866

This PR does not fully return the ability to work with 'timestamp with time zone' we had in 2.2 because such fields cannot be used in where conditions and inserted/updated.

I.e. If we try to insert into the field, we got:
```
iex(2)> MyApp.Repo.insert!(MyApp.NotOk.changeset(%MyApp.NotOk{}, %{created_at: DateTime.utc_now}))

10:32:12.492 [debug] QUERY ERROR db=1.8ms
INSERT INTO "not_oks" ("created_at") VALUES ($1) RETURNING "id" [~N[2018-08-13 04:32:12]]
** (ArgumentError) Postgrex expected %DateTime{}, got ~N[2018-08-13 04:32:12]. Please make sure the value you are passing matches the definition in your table or in your query or convert the value accordingly.
    (postgrex) lib/postgrex/type_module.ex:714: Postgrex.DefaultTypes.encode_params/3
    (postgrex) lib/postgrex/query.ex:45: DBConnection.Query.Postgrex.Query.encode/3
    (db_connection) lib/db_connection.ex:1658: DBConnection.encode/5
    (db_connection) lib/db_connection.ex:1719: DBConnection.run_prepare_execute/6
    (db_connection) lib/db_connection.ex:1834: DBConnection.run/6
    (db_connection) lib/db_connection.ex:673: DBConnection.prepare_execute/4
    (ecto) lib/ecto/adapters/postgres/connection.ex:81: Ecto.Adapters.Postgres.Connection.execute/4
    (ecto) lib/ecto/adapters/sql.ex:288: Ecto.Adapters.SQL.sql_call/5
```

Looks like again the adapter expects a DateTime for timestamp with timezone field, but Ecto gives NaiveDateTime for all utc_datetime fields. I have no idea how to fix it yet; haven't dig into that.

### DIFF
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -3,16 +3,18 @@ Code.require_file "../support/types.exs", __DIR__
 defmodule Ecto.Integration.TypeTest do
   use Ecto.Integration.Case, async: Application.get_env(:ecto, :async_integration_tests, true)
 
-  alias Ecto.Integration.{Custom, Item, Order, Post, User, Tag}
+  alias Ecto.Integration.{Custom, ExternalEntity, Item, Order, Post, User, Tag}
   alias Ecto.Integration.TestRepo
   import Ecto.Query
 
   test "primitive types" do
-    integer  = 1
-    float    = 0.1
-    text     = <<0, 1>>
-    uuid     = "00010203-0405-4607-8809-0a0b0c0d0e0f"
-    datetime = ~N[2014-01-16 20:26:51]
+    integer           = 1
+    float             = 0.1
+    text              = <<0, 1>>
+    uuid              = "00010203-0405-4607-8809-0a0b0c0d0e0f"
+    datetime          = ~N[2014-01-16 20:26:51]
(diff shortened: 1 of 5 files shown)


josevalim:
Thanks @nimf! We actually don't need the integration test changes, just the remaining ones. Can you please remove them?

josevalim:
Actually, no worries, merged manually, thank you!

PR #2655, opened 2018-08-14
Stricter handling of usec and non-usec calendar types

wojtekmach:
Ref https://github.com/elixir-ecto/ecto/issues/2018#issuecomment-411356454

* `cast` was already truncating microseconds for non-usec, it's now padding them for usec types too:

      iex> Ecto.Type.cast(:time, ~T[09:00:00.123])
      {:ok, ~T[09:00:00]}

      iex> Ecto.Type.cast(:time_usec, ~T[09:00:00])
      {:ok, ~T[09:00:00.000000]}

* similarly, `dump` enforces precision:

      iex> Ecto.Type.dump(:time_usec, ~T[09:00:00.123456])
      {:ok, ~T[09:00:00.123456]}
      iex> Ecto.Type.dump(:time, ~T[09:00:00])
      {:ok, ~T[09:00:00]}

      iex> Ecto.Type.dump(:time_usec, ~T[09:00:00.123])
      :error
      iex> Ecto.Type.dump(:time, ~T[09:00:00.123456])
      :error

* `load` was already truncating, it's now padding too.
  Note, it's less strict than dump as e.g. Postgrex always has microsecond precision
  and Mariaex has second precision if microseconds are 0.

      iex> Ecto.Type.load(:time, ~T[09:00:00.123])
      {:ok, ~T[09:00:00]}
      iex> Ecto.Type.load(:time_usec, ~T[09:00:00])
      {:ok, ~T[09:00:00.000000]}
      iex> Ecto.Type.load(:time_usec, ~T[09:00:00.123])
      {:ok, ~T[09:00:00.123000]}

**Note**: this effectively prohibits `~T[09:00:00.123]` as _internal_ data,
the millisecond precision might be in user input and in "external" DB
but this would be handled by Ecto internally as either second or microsecond precision.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -469,9 +469,7 @@ defmodule Ecto.Type do
   end
 
   def load(:utc_datetime_usec, %NaiveDateTime{} = naive_datetime, _loader) do
-    with {:ok, %DateTime{} = datetime} <- DateTime.from_naive(naive_datetime, "Etc/UTC") do
-      {:ok, pad_usec(datetime)}
-    end
+    naive_datetime |> pad_usec() |> DateTime.from_naive("Etc/UTC")
   end
 
   def load(type, value, _loader) do
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -336,23 +336,23 @@ defmodule Ecto.Type do
     {:ok, Decimal.from_float(term)}
   end
 
-  def dump(:time, %Time{} = time, _dumper) do
+  def dump(:time, %Time{microsecond: {0, 0}} = time, _dumper) do
     {:ok, zerofy_microsecond(time)}
   end
 
-  def dump(:time_usec, %Time{} = time, _dumper) do
+  def dump(:time_usec, %Time{microsecond: {_, 6}} = time, _dumper) do
     {:ok, time}
   end
 
-  def dump(:naive_datetime, %NaiveDateTime{} = naive_datetime, _dumper) do
+  def dump(:naive_datetime, %NaiveDateTime{microsecond: {0, 0}} = naive_datetime, _dumper) do
     {:ok, zerofy_microsecond(naive_datetime)}
   end
 
-  def dump(:naive_datetime_usec, %NaiveDateTime{} = naive_datetime, _dumper) do
+  def dump(:naive_datetime_usec, %NaiveDateTime{microsecond: {_, 6}} = naive_datetime, _dumper) do
     {:ok, naive_datetime}
   end
 
-  def dump(:utc_datetime, %DateTime{time_zone: time_zone} = datetime, _dumper) do
+  def dump(:utc_datetime, %DateTime{time_zone: time_zone, microsecond: {0, 0}} = datetime, _dumper) do
     if time_zone != "Etc/UTC" do
       message = ":utc_datetime expects the time zone to be \"Etc/UTC\", got `#{inspect(datetime)}`"
       raise ArgumentError, message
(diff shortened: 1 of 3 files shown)


lib/ecto/type.ex
```diff
@@ -637,25 +639,25 @@ defmodule Ecto.Type do
   end
 
   def cast(:time, term) do
```
fertapric: What do you think about keeping the private `cast_time` as before, and move the `pad_usec` logic to the `cast/2` function like in the previous clauses?

```elixir
def cast(:time, term) do
  case cast_time(term) do
    {:ok, time} -> {:ok, zerofy_microsecond(time)}
    :error -> :error
  end
end

def cast(:time_usec, term) do
  case cast_time(term) do
    {:ok, time} -> {:ok, pad_usec(time)}
    :error -> :error
  end
end

defp pad_usec(nil), do: nil
defp pad_usec(%{calendar: _, microsecond: {microsecond, _precision}} = struct) do
  %{struct | microsecond: {microsecond, 6}}
end
```

It would behave similarly to `zerofy_microsecond/1` :)

wojtekmach: @fertapric you're of course right, done!

lib/ecto/type.ex
```diff
@@ -449,27 +449,29 @@ defmodule Ecto.Type do
...
   end
 
   def load(:utc_datetime, %NaiveDateTime{} = naive_datetime, _loader) do
-    naive_datetime |> zerofy_microsecond |> DateTime.from_naive("Etc/UTC")
+    naive_datetime |> truncate_usec() |> DateTime.from_naive("Etc/UTC")
   end
 
   def load(:utc_datetime_usec, %NaiveDateTime{} = naive_datetime, _loader) do
-     DateTime.from_naive(naive_datetime, "Etc/UTC")
+    with {:ok, %DateTime{} = datetime} <- DateTime.from_naive(naive_datetime, "Etc/UTC") do
```
fertapric: This `with` might suggest that an error response might be forwarded from `DateTime.from_naive/2`. However, that function always returns `{:ok, datetime}`. 

In addition, I think the pattern match on the struct `%DateTime{}` is not needed in this case :)

What do you think then about?

```elixir
{:ok, naive_datetime |> DateTime.from_naive!("Etc/UTC") |> pad_usec()}

# or:

{:ok, pad_usec(DateTime.from_naive!(naive_datetime, "Etc/UTC"))}

# or:

datetime = DateTime.from_naive!(naive_datetime, "Etc/UTC")
{:ok, pad_usec(datetime)}

# or:

{:ok, datetime} = DateTime.from_naive(naive_datetime, "Etc/UTC")
{:ok, pad_usec(datetime)}
```

wojtekmach: good catch! I went with:

    naive_datetime |> pad_usec() |> DateTime.from_naive("Etc/UTC")

to get a nice symmetry with a clause above:

    naive_datetime |> truncate_usec() |> DateTime.from_naive("Etc/UTC")

lib/ecto/type.ex
```diff
@@ -1041,4 +1047,10 @@ defmodule Ecto.Type do
     do: :error
   defp validate_decimal(other),
     do: other
+
+  defp truncate_usec(nil), do: nil
+  defp truncate_usec(struct), do: %{struct | microsecond: {0, 0}}
```
josevalim: We should probably do this only if `microsecond` is not already `{0, 0}`, this can save us some memory allocation.

lib/ecto/type.ex
```diff
@@ -1041,4 +1047,10 @@ defmodule Ecto.Type do
...
+
+  defp pad_usec(nil), do: nil
+  defp pad_usec(%{microsecond: {microsecond, _}} = struct), do: %{struct | microsecond: {microsecond, 6}}
```
josevalim: We should probably do this only if `microsecond` is not already `{_, 6}`, this can save us some memory allocation.

wojtekmach:
@josevalim tests fixed in https://github.com/elixir-ecto/ecto/pull/2655/commits/4b84b61cf76f057bc559f195bbe5fd2ca4846b51 and optimisation done in 367d4f009cf08eed33b5ddc8991bb060abea1280 - thanks for feedback!

josevalim:

PR #2663, opened 2018-08-20
Rework Ecto.Type

wojtekmach:
- [x] Optimize `Ecto.Type.equal?/3`
- [x] Make `Ecto.Type.cast/1` callback optional
- [x] Make `Ecto.Type.dump/1` callback optional
- [x] Make `Ecto.Type.load/1` callback optional

---

Before, for `{:array, Custom}` we were hitting `function_exported?(Custom, :equal, 2)` for each element in the list.
Now, we compute `equal_fun` once and then just call it for each element.

cc @michalmuskala

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -296,7 +296,6 @@ defmodule Ecto.Type do
       iex> dump({:array, :binary}, ["1", "2", "3"])
       {:ok, ["1", "2", "3"]}
 
-  A `dumper` function may be given for handling recursive types, see: `dump/3`.
   """
   @spec dump(t, term) :: {:ok, term} | :error
   def dump(_type, nil) do
@@ -308,9 +307,10 @@ defmodule Ecto.Type do
   end
 
   @doc """
-  Dumps a value to the given type using a custom function.
+  Dumps a value to the given type.
 
-  See `dump/2` for more information.
+  This function behaves the same as `dump/2`, except for composite types
+  the given `dumper` function is used.
   """
   @spec dump(t, term, (t, term -> {:ok, term} | :error)) :: {:ok, term} | :error
   def dump(_type, nil, _dumper) do
@@ -461,7 +461,6 @@ defmodule Ecto.Type do
       iex> load(:integer, "10")
       :error
 
-  A `loader` function may be given for handling recursive types, see: `load/3`.
   """
   @spec load(t, term) :: {:ok, term} | :error
   def load({:embed, embed}, value) do
@@ -477,9 +476,10 @@ defmodule Ecto.Type do
   end
 
   @doc """
-  Dumps a value to the given type using a custom function.
+  Loads a value with the given type.
 
-  See `dump/2` for more information.
+  This function behaves the same as `load/2`, except for composite types
+  the given `loader` function is used.
   """
   @spec load(t, term, (t, term -> {:ok, term} | :error)) :: {:ok, term} | :error
   def load({:embed, embed}, value, loader) do
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -301,90 +301,130 @@ defmodule Ecto.Type do
 
   A `dumper` function may be given for handling recursive types.
   """
-  @spec dump(t, term, (t, term -> {:ok, term} | :error)) :: {:ok, term} | :error
-  def dump(type, value, dumper \\ &dump/2)
+  # @spec dump(t, term, (t, term -> {:ok, term} | :error)) :: {:ok, term} | :error
 
   def dump(_type, nil, _dumper) do
     {:ok, nil}
-  end
-
-  def dump(:binary_id, value, _dumper) when is_binary(value) do
-    {:ok, value}
-  end
-
-  def dump(:any, value, _dumper) do
-    {:ok, value}
   end
 
   def dump({:embed, embed}, value, dumper) do
     dump_embed(embed, value, dumper)
   end
 
-  def dump({:array, type}, value, dumper) when is_list(value) do
-    array(value, type, dumper, [])
+  def dump({:in, type}, value, dumper) do
+    case dump({:array, type}, value, dumper) do
+      {:ok, value} -> {:ok, {:in, value}}
+      :error -> :error
+    end
   end
 
   def dump({:map, type}, value, dumper) when is_map(value) do
     map(Map.to_list(value), type, dumper, %{})
   end
 
-  def dump({:in, type}, value, dumper) do
-    case dump({:array, type}, value, dumper) do
-      {:ok, v} -> {:ok, {:in, v}}
-      :error -> :error
-    end
+  # TODO: this head is required to pass `dumper`, see
+  # the head below ignroes dumper. Maybe return 2-arity
+  # function from `dump_fun` and always pass dumper instead?
+  def dump({:array, type}, value, dumper) do
+    array(value, type, dumper, [])
   end
 
-  def dump(:decimal, term, _dumper) when is_integer(term) do
-    {:ok, Decimal.new(term)}
+  def dump(type, value, _) do
+    dump_fun(type).(value)
   end
 
-  def dump(:decimal, term, _dumper) when is_float(term) do
-    {:ok, Decimal.from_float(term)}
+  def dump(_type, nil) do
+    {:ok, nil}
   end
 
-  def dump(:time, %Time{microsecond: {0, 0}} = time, _dumper) do
-    {:ok, time}
+  def dump(type, value) do
+    dump_fun(type).(value)
   end
 
-  def dump(:time_usec, %Time{microsecond: {_, 6}} = time, _dumper) do
-    {:ok, time}
+  defp dump_fun(:integer), do: &dump_integer/1
+  defp dump_fun(:float), do: &dump_float/1
+  defp dump_fun(:boolean), do: &dump_boolean/1
+  defp dump_fun(:map), do: &dump_map/1
+  defp dump_fun(:string), do: &dump_binary/1
+  defp dump_fun(:binary), do: &dump_binary/1
+  defp dump_fun(:id), do: &dump_integer/1
+  defp dump_fun(:binary_id), do: &dump_binary/1
+  defp dump_fun(:any), do: &{:ok, &1}
+  defp dump_fun(:decimal), do: &dump_decimal/1
+  defp dump_fun(:date), do: &dump_date/1
+  defp dump_fun(:time), do: &dump_time/1
+  defp dump_fun(:time_usec), do: &dump_time_usec/1
+  defp dump_fun(:naive_datetime), do: &dump_naive_datetime/1
+  defp dump_fun(:naive_datetime_usec), do: &dump_naive_datetime_usec/1
+  defp dump_fun(:utc_datetime), do: &dump_utc_datetime/1
+  defp dump_fun(:utc_datetime_usec), do: &dump_utc_datetime_usec/1
+
+  # TODO: this is similar to `def dump({:array, type}` above, see if can be removed
+  defp dump_fun({:array, type}) do
+    &array(&1, dump_fun(type), [])
   end
 
-  def dump(:naive_datetime, %NaiveDateTime{microsecond: {0, 0}} = naive_datetime, _dumper) do
-    {:ok, naive_datetime}
+  defp dump_fun({:map, type}) do
+    &map(&1, dump_fun(type), %{})
   end
 
-  def dump(:naive_datetime_usec, %NaiveDateTime{microsecond: {_, 6}} = naive_datetime, _dumper) do
-    {:ok, naive_datetime}
+  defp dump_fun(mod) when is_atom(mod) do
+    &mod.dump(&1)
   end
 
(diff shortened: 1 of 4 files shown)


lib/ecto/type.ex
```diff
@@ -916,75 +914,78 @@ defmodule Ecto.Type do
...
-    NaiveDateTime.compare(term1, term2) == :eq
-  end
-  defp do_equal?(type, %DateTime{} = term1, %DateTime{} = term2)
-       when type in [:utc_datetime, :utc_datetime_usec] do
-    DateTime.compare(term1, term2) == :eq
-  end
-  defp do_equal?(type, term1, term2) when type in @base do
-    term1 == term2
+  defp equal_fun({:array, type}) do
+    if fun = equal_fun(type) do
```
wojtekmach: @josevalim the biggest gain is here, we only call this function "once" (its recursive) instead of per element in the list which is important for custom types that require function_exported? check.

We were discussing that we should perform similar optimisation for load/dump/cast but I don't think it's necessary; in the `cast/2` call here https://github.com/elixir-ecto/ecto/blob/9f8ddb3da57305d526e5fba845e3700297a78f25/lib/ecto/type.ex#L613 we could already know that it's say :decimal so we could have some `cast_decimal` but letting beam find matching clause of cast/2 should be fast enough right? Maybe I misunderstood.

josevalim: @wojtekmach we will do the exact same changes for load/dump/cast. Since we will make the callback optional, we will know in many situations that we can just return `{:ok, value}` without traversing at all. For example, for an array of integers, we don't need to traverse it either, we just need to return `{:ok, value}`.

wojtekmach:
@josevalim I've pushed the change to `cast/1` as we've discussed. I've also changed PR title & description to reflect what's been done and what's left.

lib/ecto/type.ex
```diff
@@ -600,117 +598,95 @@ defmodule Ecto.Type do
...
   def cast({:in, _type}, nil), do: :error
   def cast(_type, nil), do: {:ok, nil}
 
-  def cast(:binary_id, value) when is_binary(value) do
-    {:ok, value}
+  def cast(type, value) do
+    caster(type).(value)
+  end
+
+  defp caster(type) when type in @basic, do: &cast_basic(type, &1)
```
fertapric: What do you think about having different cast functions instead of `cast_basic`? It would pattern match only once then; and it wouldn't need the new `@basic` module attribute:

```elixir
defp caster(:float), do: &cast_float/1
defp caster(:boolean), do: &cast_boolean/1
defp caster(:binary_id), do: &cast_binary_id/1
defp caster(:id), do: &cast_integer/1
defp caster(:integer), do: &cast_integer/1
defp caster(:decimal), do: &cast_decimal/1
defp caster(:date), do: &cast_date/1
defp caster(:time), do: &maybe_truncate_usec(cast_time(&1))
defp caster(:time_usec), do: &maybe_pad_usec(cast_time(&1))
defp caster(:naive_datetime), do: &maybe_truncate_usec(cast_naive_datetime(&1))
defp caster(:naive_datetime_usec), do: &maybe_pad_usec(cast_naive_datetime(&1))
defp caster(:utc_datetime), do: &maybe_truncate_usec(cast_utc_datetime(&1))
defp caster(:utc_datetime_usec), do: &maybe_pad_usec(cast_utc_datetime(&1))
defp caster(type) when type in [:string, :binary, :map, :any], do: &cast_base(type, &1)

# rest of caster/1 functions

defp cast_float(term) when is_binary(term) do
  case Float.parse(term) do
    {float, ""} -> {:ok, float}
    _ -> :error
  end
end

defp cast_float(term) when is_integer(term), do: {:ok, :erlang.float(term)}
defp cast_float(term) when is_float(term), do: {:ok, term}
defp cast_float(_term), do: :error

defp cast_boolean(term) when term in ~w(true 1),  do: {:ok, true}
defp cast_boolean(term) when term in ~w(false 0), do: {:ok, false}
defp cast_boolean(_term), do: :error

defp cast_base(type, term) do
  if of_base_type?(type, term) do
    {:ok, term}
  else
    :error
  end
end
```

lib/ecto/type.ex
```diff
@@ -600,117 +598,95 @@ defmodule Ecto.Type do
...
+  defp caster(:decimal), do: &cast_decimal/1
+  defp caster(:date), do: &cast_date/1
+  defp caster(:time), do: &maybe_truncate_usec(cast_time(&1))
+  defp caster(:time_usec), do: &maybe_pad_usec(cast_time(&1))
+  defp caster(:naive_datetime), do: &maybe_truncate_usec(cast_naive_datetime(&1))
```
fertapric: I would also rename this to `cast_fun`, to mirror `equal_fun` naming :)

lib/ecto/type.ex
fertapric: In fact, this idea is not new in the codebase. It would be similar to how `equal_fun` is calculated:

```elixir
  defp equal_fun(nil), do: nil
  defp equal_fun(:decimal), do: &equal_decimal?/2
  defp equal_fun(t) when t in [:time, :time_usec], do: &equal_time?/2
  defp equal_fun(t) when t in [:utc_datetime, :utc_datetime_usec], do: &equal_utc_datetime?/2
  defp equal_fun(t) when t in [:naive_datetime, :naive_datetime_usec], do: &equal_naive_datetime?/2
  defp equal_fun(t) when t in @base, do: nil
```

The "catch-rest-base-type" clause could be changed then from 

```elixir
defp caster(type) when type in [:string, :binary, :map, :any], do: &cast_base(type, &1)
```

to

```elixir
defp caster(type) when type in @base, do: &cast_base(type, &1)
```

josevalim: The goal here is to avoid the type lookups, so we should do our best to avoid any type based dispatching.

michalmuskala:
Here's an idea - what if we made all the types proper modules with all the callbacks (going a bit against the recent changes)?

I think this would actually offer a more efficient solution, because it would allow skipping the type dispatch in `Ecto.Type` functions (since you'd call the type function directly). It also naturally works for arrays and maps and there's no need for some special optimisation code. We could map the simple atom types into internal modules during compilation.

I think in this set-up we could also come up with a way to have optional callbacks. What if we "resolved" the callbacks at compile-time? I believe today schemas already have a compile-time dependency on the type module since it's a literal in the AST.

lib/ecto/type.ex
fertapric: I think right now the type-based dispatching is split in two, first in the `caster(type) when type in @base` and then in the `cast_basic(:float, term)` or `cast_basic(:boolean, term)`. My idea was to merge the dispatching logic into the `caster` function. Would that make sense? :)

fertapric:
@michalmuskala Would that work with ["schemaless changeset"](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-schemaless-changesets)?

josevalim:
@michalmuskala actually, I don't think we should go with optional callbacks for cast/dump/load. This PR makes it clear that even the simplest of types requires at least a guard check. An Ecto.Type that accepts any Elixir value without checks is very rare to the point it is not worth specializing.

We could also move to individual `Ecto.Type` for each built-in type. I don't see a problem with that but let's treat it as a separate discussion. If anything, this PR is a stepping stone exactly because we are already isolating the functionality into individual functions. The other thing, already mentioned by @fertapric, there are many runtime places where types are used, so such proposal should be taken more carefully, as it has a larger scope than the optimization scope of this PR.

The other thing is that, if we want to overhaul types, then I would also wait for the overhaul before doing larger changes.

lib/ecto/type.ex
```diff
@@ -152,7 +150,7 @@ defmodule Ecto.Type do
   """
   @callback equal?(term, term) :: boolean
 
-  @optional_callbacks [equal?: 2]
+  @optional_callbacks [cast: 1, dump: 1, load: 1, equal?: 2]
```
wojtekmach: I've spoken with @josevalim, we're reverting this change because the default behaviour of passing through the value is too relaxed. E.g. if people wrap a string they should use `is_binary` guards for all calback implementations.

@michalmuskala we talked a while ago about limiting type boilerplate, do you have some ideas?

yordis:
@michalmuskala I am not sure if you remember the conversation about the primitive types but since this module is getting refactor that much I would like to bring it up.

Since this module is implementing a lot of code related to the `primitive` types, the idea is to actually move those codes into its own modules as treat the primitive types are regular types.

At least for me, it is simple to refactor or map the implementation in your head that way.

Sure the number of files increase but this module become leaner and focused on its responsibility which it shouldn't be implementing types.

lib/ecto/type.ex
```diff
@@ -916,92 +932,141 @@ defmodule Ecto.Type do
...
+  defp of_base_type?(:integer, term), do: is_integer(term)
+  defp of_base_type?(:boolean, term), do: is_boolean(term)
+  defp of_base_type?(:binary, term), do: is_binary(term)
+  defp of_base_type?(:string, term), do: is_binary(term)
+  defp of_base_type?(:map, term), do: is_map(term) and not Map.has_key?(term, :__struct__)
+  defp of_base_type?(:decimal, value), do: Kernel.match?(%Decimal{}, value)
+  defp of_base_type?(:date, value), do: Kernel.match?(%Date{}, value)
+  defp of_base_type?(_, _), do: false
+
+  # TODO: pasted from array/4, remove after changing dump/1 and load/1
```
josevalim: Is this TODO still relevant? :)

wojtekmach: oops, nope; we need both array/3 and array/4

lib/ecto/type.ex
```diff
@@ -301,92 +296,124 @@ defmodule Ecto.Type do
       iex> dump({:array, :binary}, ["1", "2", "3"])
       {:ok, ["1", "2", "3"]}
 
-  A `dumper` function may be given for handling recursive types.
+  A `dumper` function may be given for handling recursive types, see: `dump/3`.
```
fertapric: `dump/2` does not include the `dumper` argument, so I would replace the ``` `dumper` ``` reference in this paragraph with something like “An additional function” or similar :)

lib/ecto/type.ex
```diff
@@ -301,92 +296,124 @@ defmodule Ecto.Type do
...
-  def dump(:binary_id, value, _dumper) when is_binary(value) do
-    {:ok, value}
+  def dump(type, value) do
+    dump_fun(type).(value)
   end
 
-  def dump(:any, value, _dumper) do
-    {:ok, value}
+  @doc """
+  Dumps a value to the given type using a custom function.
```
fertapric: I think this paragraph is not accurate, otherwise I would expect the catch-rest function to be like:

```elixir
def dump(_type, value, dumper) do
  dumper.(value)
end
```

(which makes no sense because you could just use the `dumper` function directly)

yordis:
@wojtekmach after watching that Gist I would definitely love to read the source code like that. Maybe its own folder with the files so we do not have the massive file, but either way, this is cleaner.

Folks, any opinions since the issue is fixed? Maybe do the refactoring before moving forward?

yordis:
btw, using modules and just following the same specification for the primitive as well could potentially remove code like this one https://github.com/elixir-ecto/ecto/blob/5914556cdf98828da2ad3afb6b6e9029188cfb7d/lib/ecto/type.ex#L224

Where we need to be aware of situation so we treat the types differently.

lib/ecto/type.ex
```diff
@@ -434,71 +461,89 @@ defmodule Ecto.Type do
       iex> load(:integer, "10")
       :error
 
-  A `loader` function may be given for handling recursive types.
+  A `loader` function may be given for handling recursive types, see: `load/3`.
```
fertapric: The same applies here, `loader` is not an argument of `load/2` :)

fertapric: The same applies here, `loader` is not an argument of `load/2` :)

lib/ecto/type.ex
```diff
@@ -434,71 +461,89 @@ defmodule Ecto.Type do
...
-    naive_datetime |> truncate_usec() |> DateTime.from_naive("Etc/UTC")
+  def load(type, value) do
+    load_fun(type).(value)
   end
 
-  def load(:utc_datetime, %DateTime{} = datetime, _loader) do
-    {:ok, truncate_usec(datetime)}
-  end
+  @doc """
+  Dumps a value to the given type using a custom function.
```
fertapric: I think this is a copy & paste error. This should be something like “Loads a value...” and reference `load/2` instead of `dump/2` :)

josevalim:

yordis:
@josevalim @michalmuskala so what do you think about the feedback I gave?

I was working locally on that before I saw this PR so

josevalim:
I have commented on this earlier:

> We could also move to individual Ecto.Type for each built-in type. I don't see a problem with that but let's treat it as a separate discussion. If anything, this PR is a stepping stone exactly because we are already isolating the functionality into individual functions. The other thing, already mentioned by @fertapric, there are many runtime places where types are used, so such proposal should be taken more carefully, as it has a larger scope than the optimization scope of this PR.

> The other thing is that, if we want to overhaul types, then I would also wait for the overhaul before doing larger changes.

yordis:
@josevalim I am so sorry I missed that comment completely, sorry.

PR #2667, opened 2018-08-23
Ecto type for postgres daterange

dgvncsz0f:
## summary

This patch creates an ecto type for `daterange` using `postgrex` range
type. It also contains a module with primitives to create another
discrete range types, like `intrange`.

The idea is to define types for all other discrete range types as well
but I'd like to test the reception of a patch of this nature and
collect feedback before doing more deep work as currently we only use
`daterange`.

As far as I know this type is Postgres only the code probably makes
this hypothesis very evident as it is completely dependent of
`postgrex`.

## code

there are two modules. `discrete_range.ex` contains functions to create range types. In short it requires a `successor` function and a `comparison` function to work. The other module is defined in `postgres_date_range.ex` and contains the actual ecto type.

### DIFF
--- a/lib/ecto/discrete_range.ex
+++ b/lib/ecto/discrete_range.ex
@@ -1,0 +1,311 @@
+defmodule Ecto.DiscreteRange do
+  @moduledoc """
+  Define useful functions to create an ecto.type for the postgres
+  range type for discrete types. It represents the interval as a
+  `closed-open` interval.
+
+  To use this module your type must have a well define successor
+  function and a comparison function.
+  """
+
+  @typedoc """
+  The type of a point in the range.
+  """
+  @type t :: :infinity | :_infinity | term
+
+  @typedoc """
+  The successor function. For example, for integer type, returns,
+  `succ_fn(n)` must return `n+1`.
+  """
+  @type succ_fn :: (t -> t)
+
+  @typedoc """
+  The comparison function.
+  """
+  @type comp_fn :: (t, t -> :lt | :eq | :gt)
+
+  @typedoc """
+  Allen's relantionships between intervals. Refer to
+  https://en.wikipedia.org/w/index.php?title=Allen%27s_interval_algebra&oldid=846947944 for more information.
+  """
+  @type allens_relation ::
+          :equals | :finishes | :starts | :meets | :before | :after | :during | :overlaps
+
+  @doc """
+  Creates a new interval.
+
+    * `:succ_fn` computes the next value;
+    * `:comp_fn` comparator for the given type;
+    * `:lower` the lower term of the interval;
+    * `:upper` the upper term of the interval;
+    * `:mode` defines how to interpret interval boundaries;
+  """
+  @spec new(succ_fn, comp_fn, Ecto.Type, t, t, {:open | :closed, :open | :closed}) ::
+          {:ok, {t, t}} | :error
+  def new(succ_fn, comp_fn, ecto_ty, lower, upper, mode) do
+    interval_fn =
+      case mode do
+        {:closed, :open} ->
+          fn a, b -> {:ok, {a, b}} end
+
+        {:closed, :closed} ->
+          fn a, b -> {:ok, {a, succ_term(succ_fn, b)}} end
+
+        {:open, :open} ->
+          fn a, b -> {:ok, {succ_term(succ_fn, a), b}} end
+
+        {:open, :closed} ->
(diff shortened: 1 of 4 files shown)


lib/ecto/discrete_range.ex
```diff
@@ -0,0 +1,311 @@
+defmodule Ecto.DiscreteRange do
+  @moduledoc """
+  Define useful functions to create an ecto.type for the postgres
```
umamaistempo: In this case it would be nice to quote to the ecto type module as ex_doc would autolink it's docs

Ie: 

```
Black black `Ecto.Type` food desu
```

lib/ecto/discrete_range.ex
```diff
@@ -0,0 +1,311 @@
...
+  def dump(ecto_ty, lower, upper) do
+    with {:ok, lower} <- dump_term(ecto_ty, lower),
+         {:ok, upper} <- dump_term(ecto_ty, upper) do
+      {:ok,
+       %Postgrex.Range{
+         lower: lower,
+         upper: upper,
+         lower_inclusive: lower != nil,
+         upper_inclusive: false
+       }}
```
umamaistempo: This will make the code not compile if the application uses ecto with any driver besides postgrex. To be viable, there should be a transparent way to create ranges in Ecto that is then passed to the driver

lib/ecto/discrete_range.ex
```diff
@@ -0,0 +1,311 @@
...
+          :u -> {:ok, :infinity}
+          :l -> {:ok, :_infinity}
+        end
+
+      _value ->
+        Ecto.Type.load(ecto_ty, value)
+    end
+  end
+
+  @spec cast_term(:r | :u, Ecto.Type.t(), term) :: {:ok, t} | :error
```
umamaistempo: In this spec `mode :: :r | :u` but on the code you check if it's `:l`

lib/ecto/discrete_range.ex
```diff
@@ -0,0 +1,311 @@
...
+  @doc """
+  Creates a new interval.
+
+    * `:succ_fn` computes the next value;
+    * `:comp_fn` comparator for the given type;
+    * `:lower` the lower term of the interval;
+    * `:upper` the upper term of the interval;
+    * `:mode` defines how to interpret interval boundaries;
+  """
+  @spec new(succ_fn, comp_fn, Ecto.Type, t, t, {:open | :closed, :open | :closed}) ::
```
umamaistempo: Here `Ecto.Type` means the module, I think you means `Ecto.Type.t`

dgvncsz0f:
there is no difference between `null` and `infinity` using the `&&` operator. if i recall correctly, the difference seems to be whether or not the interval is inclusive or not and when using the `=` operator. 

in the code, an unbounded interval is always represented as an open interval and it uses `null`. as long as the database is consistently using `null` to represent an unbounded interval, the issue would never appear. that until postgrex is fixed.

lib/ecto/discrete_range.ex
dgvncsz0f: i've avoided using `if Code.ensure_loaded?(Postgrex)` like ecto does in the adapters modules at this point because (a) i don't know if that is acceptable here (b) i'd like to probe if something like this is wanted in the ecto library before doing any more work.

is that `if` the way to go here?

lib/ecto/discrete_range.ex
```diff
@@ -0,0 +1,311 @@
...
+      {:infinity, _} ->
+        :gt
+
+      {_, :infinity} ->
+        :lt
+
+      {_, :_infinity} ->
+        :gt
+
+      _otherwise ->
```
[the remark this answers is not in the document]
dgvncsz0f: i've written this way the first time but changed it to this form to force using `comp_fn` when it is a bounded interval. ok.

josevalim:
Thanks @dgvncsz0f for the PR. We don't plan to support new Ecto Types in core per-se but custom types can always be added as a separate package (like the postgis types are in a separate package). In case there is no package for handling pg ranges in Ecto, this is a great opportunity to start one. :)

lib/ecto/discrete_range.ex
```diff
@@ -0,0 +1,311 @@
...
+         lower: lower,
+         upper: upper,
+         lower_inclusive: lower != nil,
+         upper_inclusive: false
+       }}
+    end
+  end
+
+  @doc """
+  Compute the allen relationship between two proper intervals:
```
wojtekmach: just wanted to say I'm really happy to see Allen's Interval Algebra getting some love :)

Issue #2614, opened 2018-07-06
Mixed migration rollback order


phunehehe:
### Environment

* Elixir version (elixir -v): 1.6.6
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6
* Ecto version (mix deps): 2.2.10
* Database adapter and version (mix deps): postgrex 0.13.5
* Operating system: Alpine

### Current behaviour

1. Let's say I create a migration `20180701_first.exs`, but shelve the branch to work on something else.
2. Then I create another migration `20180702_second.exs`, finish and deploy it. Everything is good.
3. Then I go back to `20180701_first.exs`, finish and also deploy it. Everything is still good.
4. Now I find a problem with `20180701_first.exs` and want to roll back.

- `mix ecto.rollback` reverts `20180702_second.exs`.
- `mix ecto.rollback --to 20180701` reverts `20180702_second.exs`, then it also reverts `20180701_first.exs`.

### Expected behaviour

It's hard to say what should be expected here. Basically the problem happens because Ecto assumes the file name order is the same as the applied order.

One way is to honour the applied order:
- `mix ecto.rollback` should revert the latest migration that was applied, that is `20180701_first.exs`
- `mix ecto.rollback --to XXX` should revert migrations based on the applied order, i.e. first revert `20180701_first.exs`, then revert `20180702_second.exs`.

Another way is to honour the timestamp:
- At step (3), `ecto.migrate` should complain that the file name indicates an incompatible version. Then I can change the file name and things will be good again.

Thoughts?

josevalim:
It seems that using the insertion time and define the order based on that is a good way to go.

fertapric:
I have some doubts about using the insertion time to define the order. I think it might be confusing to order the list of `mix ecto.migrations` in a different order than the one defined in the migrations directory; or if the migrations are ordered using the timestamp (I guess this is less likely), then `--step`, `--to` and similar flags would be quite hard to use.

In addition, it could make migrations order inconsistent and different per environments. For example, if the both migrations given in the example  are rolled back and applied again, the order of the migrations would probably change because, in this new execution, migrations would be applied based on the timestamp.

Besides, for what I’ve experienced, the step (3) usually requires fixing conflicts in `db/structure.sql`. In such cases, I tend to generate a new timestamp and dump the database structure again. I’m aware this is my personal flow, but I wanted to mention it as a “workaround” for this issue :)

[...]

PR #2669, opened 2018-08-24
Implement --strict-version-order for migrations

phunehehe:
I have not actually been able to use the option from the command line to see if it works, since I am using v2.2.10 in my project (which is the latest release, but has drifted quite far away from master).

The tests seem to work fine though. Is this something useful?

https://github.com/elixir-ecto/ecto/issues/2614

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -66,6 +66,7 @@ defmodule Ecto.Migrator do
     * `:log` - the level to use for logging. Defaults to `:info`.
       Can be any of `Logger.level/0` values or `false`.
     * `:prefix` - the prefix to run the migrations on
+    * `:strict_version_order` - abort when applying a migration with old timestamp
   """
   @spec up(Ecto.Repo.t, integer, module, Keyword.t) :: :ok | :already_up
   def up(repo, version, module, opts \\ []) do
@@ -82,7 +83,7 @@ defmodule Ecto.Migrator do
         if version != Enum.max([version | versions]) do
           latest = Enum.max(versions)
 
-          Logger.warn """
+          message = """
           You are running migration #{version} but an older \
           migration with version #{latest} has already run.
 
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/ecto.migrate.ex
```diff
@@ -65,7 +65,8 @@ defmodule Mix.Tasks.Ecto.Migrate do
 
     {opts, _, _} = OptionParser.parse args,
       switches: [all: :boolean, step: :integer, to: :integer, quiet: :boolean,
-                 prefix: :string, pool_size: :integer, log_sql: :boolean],
+                 prefix: :string, pool_size: :integer, log_sql: :boolean,
+                 strict_version_order: :boolean],
```
josevalim: Can you please update the docs for this module to mention the --struct-version-order switch?

lib/ecto/migrator.ex
```diff
@@ -93,6 +93,10 @@ defmodule Ecto.Migrator do
           If this can be an issue, we recommend to rollback #{version} and change \
           it to a version later than #{latest}.
           """
+
+          if opts[:strict_version_order] do
+            raise Ecto.MigrationError
+          end
```
josevalim: Can you please update the docs for this function to mention the `:strict_version_order` option?

josevalim: Also, what do you think of including the message in the exception? Something like this:

```elixir
message = """
...
"""

if opts[:strict_version_order] do
  raise Ecto.MigrationError, message
else
  Logger.warn(message)
end
```

phunehehe: Nice idea. Updated.

josevalim:

wojtekmach:
> I have not actually been able to use the option from the command line to see if it works, since I am using v2.2.10 in my project

A quick way to test changes is using the [`examples/friends`](https://github.com/elixir-ecto/ecto/tree/master/examples/friends) app, that's what I usually do :)

PR #2671, opened 2018-08-26
[WIP] Use module attribute for prefix lookup

yordis:
I was about to create an issue about this but I thought it would work with this PR but I got his error

```
(ArgumentError) could not call get_attribute with argument Ecto.AssociationTest.Author because the module is already compiled
```
So I need your help but I think is worth to show you what I am trying to do.

### Why

As did some refactoring in my code I used to have this module name `StrawHat.IAM.PolicyStatement` and then I changed it to `StrawHat.IAM.Policy.Statement`.

I assumed that nothing would go wrong but turns out that the module name matters in the particular case that you create the foreign keys.

Since the ID in the relation is `policy_statement_id` and not `statement_id` I am kind of confused what should I do.

I would prefer to at least have some simple way to configure this. Maybe add some other metadata that it is simple to set up.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -155,8 +155,20 @@ defmodule Ecto.Association do
 
   """
   def association_key(module, suffix) do
-    prefix = module |> Module.split |> List.last |> Macro.underscore
+    prefix =
+      module
+      |> Module.get_attribute(:find_a_better_name)
+      |> association_prefix(module)
+
     :"#{prefix}_#{suffix}"
+  end
+
+  defp association_prefix(nil, module) do
+    module |> Module.split |> List.last |> Macro.underscore
+  end
+
+  defp association_prefix(prefix, _module) do
+    prefix
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


wojtekmach:
> Since the ID in the relation is policy_statement_id and not statement_id I am kind of confused what should I do.

You can use `:foreign_key` option on has_* associations to override the key name that's inferred from the module, have you tried that? Or is the issue that you'd need to do this all over the place?

yordis:
@wojtekmach well everyone that has some relationship with that module will need to do that.

To be honest, right now is just in one place but I would like to know what do you think about my approach and if it is worth to do it.

josevalim:
@yordis as @wojtekmach said, we already have a mechanism to do so via the option. I don't think this scenario is common enough to warrant introducing another mechanism for configuration.

The reason the defaults now work is because they are simple: it is based on the module name or an option. Introducing another lookup means that everybody else's applications become more complex, because now if I look at `belongs_to Foo.Bar`, I never know if it will be `:bar_id` or if `Bar` has some sort of special configuration behind it.

PR #2673, opened 2018-08-28
add function support for get_field/3 default value

btkostner:
What it says on the tin. Let me know if you want me to do this for the other changeset functions as well.


josevalim:
Unfortunately this makes it impossible to retrieve a function as a default value, which may be the case if you are using a virtual field of type `:any`. If you want to execute a function depending on if there is a value or not, you can use `fetch_field` and pattern match on the results. Thanks!

PR #2676, opened 2018-08-31
[WIP] Support multiple migration paths

anthonator:
ref https://groups.google.com/forum/#!topic/elixir-ecto/y41IAhOKZZk

Add support for using migrations from multiple paths when migrating.

### DIFF
--- a/lib/ecto/migrator.ex
+++ b/lib/ecto/migrator.ex
@@ -30,10 +30,19 @@ defmodule Ecto.Migrator do
   """
   @spec migrations_path(Ecto.Repo.t) :: String.t
   def migrations_path(repo) do
+    hd(List.first(migrations_paths(repo)))
+  end
+
+  @doc """
+  Gets the migrations paths from a repository.
+  """
+  @spec migrations_paths(Ecto.Repo.t) :: [String.t]
+  def migrations_paths(repo) do
     config = repo.config()
     priv = config[:priv] || "priv/#{repo |> Module.split |> List.last |> Macro.underscore}"
     app = Keyword.fetch!(config, :otp_app)
-    Application.app_dir(app, Path.join(priv, "migrations"))
+    paths = config[:migrations_paths] || [Application.app_dir(app)]
+    Enum.map(paths, &(Path.join(&1, priv)))
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


anthonator:
This is a WIP but wanted to start getting initial thoughts on `Ecto.Migrator.migrations_paths/1`. `Ecto.Migrator.migrations_paths/1` will return a list of paths specified by the new `:migrations_paths` config option. The `:priv` config option as been removed.

I'll keep working on adding support for providing a list of migration paths down the call chain.

lib/ecto/migrator.ex
```diff
@@ -30,10 +30,24 @@ defmodule Ecto.Migrator do
   """
   @spec migrations_path(Ecto.Repo.t) :: String.t
   def migrations_path(repo) do
+    hd(List.first(migrations_paths(repo)))
```
josevalim: hd is enough, we don't need both hd and List.first :)

lib/ecto/migrator.ex
```diff
@@ -30,10 +30,24 @@ defmodule Ecto.Migrator do
...
+  end
+
+  @doc """
+  Gets the migrations paths from a repository.
+  """
+  @spec migrations_paths(Ecto.Repo.t) :: [String.t]
+  def migrations_paths(repo) do
+    config = repo.config()
+    config[:migrations_paths] || [default_migration_path(repo)]
```
josevalim: Let's pass the config as argument to `default_migration_path` to avoid reading it twice.

anthonator: That's what I was originally doing but

https://github.com/elixir-ecto/ecto/blob/8cb17abc4d9914297b7ca8e3b5760c0ffc4b9fc7/lib/ecto/migrator.ex#L48

requires the repo to build the priv path.

josevalim:
I wrote a comment about this but I think I forgot to send it. :) We have two options:

1. Add an extra_migrations_paths option that is always appended to the default one

2. Or have migrations_paths override the default one if one is given. So if you want to keep the priv one, which is likely, you will have to list it explicitly

anthonator:
Does `:extra_migrations_paths` replace `:migrations_paths`? Might help if I had some examples of the two options.

josevalim:
Ping!

josevalim:
I will close this one as we are moving the migrations code to the ecto_sql repository and this no longer applies to this repo. If there is still an interest on this feature, we can bring it to the new repo. Thanks!

PR #2678, opened 2018-09-04
Add UNION and UNION ALL support

feymartynov:
This PR adds support of UNION and UNION ALL:

```elixir
customer_city_query = from c in Customer, select: c.city
from s in Supplier, select: s.city, union: customer_city_query
```

```elixir
customer_city_query = Customer |> select([c], c.city)
Supplier |> select([s], s.city) |> union_all(customer_query)
```

Both can be used as keyword or expression versions. Multiple unions are also supported.

I made this because I had to optimize a complex query that had some OR conditions which made the optimizer unable to use indexes. This could be optimized by rewriting it with UNION ALL but Ecto didn't have such an ability. Rewriting the query as plain SQL was impossible because it was constructed of different subqueries spreaded all accross of my app's domain logic.

This is my first contribution here. Hope I got it correctly how Ecto.Query works under the hood, especially the Planner module. If anything is wrong please let me know and I'll update the PR.


josevalim:
Thank you! I really like the approach here. We do however need tests where
both queries contain parameters interpolated to make sure we are taking
them into account properly. In particular, in the query planner, you need
to make sure you process the unions in the same order as you process the
union in MySQL / Postgres. So instead of adding your own normalize_unions,
you need to hook into the existing union.

I also think we can get rid of Ecto.Query.Builder.Union. We can implement
everything at runtime, similar to Ecto.Query.exclude.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

lib/ecto/query.ex
```diff
@@ -1188,8 +1188,12 @@ defmodule Ecto.Query do
     supplier_query = Supplier |> select([s], s.city)
     Customer |> select([c], c.city) |> union(supplier_query)
   """
-  defmacro union(query, other_query) do
-    Union.build(:union, query, other_query, __CALLER__)
+  def union(%Ecto.Query{unions: unions} = query, %Ecto.Query{} = other_query) do
```
josevalim: What happens if `other_query` has LIMIT / OFFSET / ORDER BY. Maybe we should check that those values have not been set and raise if they are set?

lib/ecto/query.ex
```diff
@@ -353,7 +353,7 @@ defmodule Ecto.Query do
   """
 
   defstruct [prefix: nil, sources: nil, from: nil, joins: [], aliases: %{}, wheres: [], select: nil,
-             order_bys: [], limit: nil, offset: nil, group_bys: [], updates: [],
+             order_bys: [], limit: nil, offset: nil, group_bys: [], unions: [], updates: [],
```
josevalim: Proposal: let's call this field "combines" or "combinations"? This way we can also add [intersect/except](https://www.postgresql.org/docs/8.3/static/queries-union.html) in the feature without adding more new fields.

lib/ecto/query/builder/union.ex
```diff
@@ -0,0 +1,30 @@
+import Kernel, except: [apply: 3]
+
+defmodule Ecto.Query.Builder.Union do
```
josevalim: I believe we can remove this file now?

test/ecto/adapters/mysql_test.exs
```diff
@@ -161,6 +161,16 @@ defmodule Ecto.Adapters.MySQLTest do
...
   end
 
+  test "union" do
+    other_query = Schema |> select([r], r.y)
+
+    query = Schema |> select([r], r.x) |> union(other_query) |> plan()
+    assert all(query) == ~s{SELECT s0.`x` FROM `schema` AS s0 UNION SELECT s0.`y` FROM `schema` AS s0}
+
+    query = Schema |> select([r], r.x) |> union_all(other_query) |> plan()
+    assert all(query) == ~s{SELECT s0.`x` FROM `schema` AS s0 UNION ALL SELECT s0.`y` FROM `schema` AS s0}
```
josevalim: We need to add a test for the precedence. If I do `query1 |> union_all(query2 |> union_all(query3))`, we need parens around the second pair. An easier solution is also to always add parens around the operators whenever UNION is used.

test/ecto/adapters/postgres_test.exs
```diff
@@ -212,6 +212,16 @@ defmodule Ecto.Adapters.PostgresTest do
...
   end
 
+  test "union" do
+    other_query = Schema |> select([r], r.y)
+
+    query = Schema |> select([r], r.x) |> union(other_query) |> plan()
+    assert all(query) == ~s{SELECT s0."x" FROM "schema" AS s0 UNION SELECT s0."y" FROM "schema" AS s0}
+
+    query = Schema |> select([r], r.x) |> union_all(other_query) |> plan()
+    assert all(query) == ~s{SELECT s0."x" FROM "schema" AS s0 UNION ALL SELECT s0."y" FROM "schema" AS s0}
```
josevalim: The precedence test is needed here.

test/ecto/adapters/postgres_test.exs
```diff
@@ -405,17 +415,21 @@ defmodule Ecto.Adapters.PostgresTest do
...
             |> plan()
 
     result =
       "SELECT s0.\"id\", $1 FROM \"schema\" AS s0 INNER JOIN \"schema2\" AS s1 ON $2 " <>
       "INNER JOIN \"schema2\" AS s2 ON $3 WHERE ($4) AND ($5) " <>
       "GROUP BY $6, $7 HAVING ($8) AND ($9) " <>
-      "ORDER BY $10, s0.\"x\" LIMIT $11 OFFSET $12"
+      "UNION SELECT s0.\"id\", $10 FROM \"schema1\" AS s0 WHERE ($11) " <>
+      "UNION ALL SELECT s0.\"id\", $12 FROM \"schema2\" AS s0 WHERE ($13) " <>
+      "ORDER BY $14, s0.\"x\" LIMIT $15 OFFSET $16"
```
josevalim: The MySQL adapter has tests similar to this, we should update it too!

josevalim:
Thanks @feymartynov ! I really like the improvements and new tests you have added. I have added a last batch of comments and we should be good to go.

test/ecto/adapters/mysql_test.exs
feymartynov: That's true. We also can have ORDER BY, OFFSET and LIMIT in union queries and without parens it would be invalid SQL so I'm going to test this case too.

lib/ecto/query.ex
feymartynov: Good idea.

feymartynov:
@josevalim I've added a couple of commits regarding your comments. Please check it out.

josevalim:
Merged manually, thank you!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #2679, opened 2018-09-05
Use String.pad_trailing to format ecto.migrations

aquarhead:
Decided to play around with the id a bit in a recent project, and I got this output via `mix ecto.migrations`:

```
  Status    Migration ID    Migration Name
--------------------------------------------------
  up        0  base_db
```

So made some changes to this task to better format both the "Status" and "Migration ID". 

Also applied `mix format` to this file

### DIFF
--- a/lib/mix/tasks/ecto.migrations.ex
+++ b/lib/mix/tasks/ecto.migrations.ex
@@ -34,33 +34,30 @@ defmodule Mix.Tasks.Ecto.Migrations do
   def run(args, migrations \\ &Ecto.Migrator.migrations/1, puts \\ &IO.puts/1) do
     repos = parse_repo(args)
 
-    result = Enum.map(repos, fn repo ->
-      ensure_repo(repo, args)
-      ensure_migrations_path(repo)
-      {:ok, pid, _} = ensure_started(repo, all: true)
+    format = fn x, n -> x |> to_string() |> String.pad_trailing(n) end
 
-      repo_status = migrations.(repo)
+    result =
+      Enum.map(repos, fn repo ->
+        ensure_repo(repo, args)
+        ensure_migrations_path(repo)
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/ecto.migrations.ex
```diff
@@ -34,33 +34,30 @@ defmodule Mix.Tasks.Ecto.Migrations do
...
+        Repo: #{inspect(repo)}
 
-        "  #{status}      #{number}  #{description}"
-      end) <> "\n"
-    end)
+          Status    Migration ID    Migration Name
+        --------------------------------------------------
+        """ <>
+          Enum.map_join(repo_status, "\n", fn {status, number, description} ->
+            "  #{format.(status, 4)}      #{format.(number, 14)}  #{description}"
```
aquarhead: Actually I can just increase the numbers here and remove all manual whitespaces, not sure which is better.

lib/mix/tasks/ecto.migrations.ex
```diff
@@ -34,33 +34,30 @@ defmodule Mix.Tasks.Ecto.Migrations do
   def run(args, migrations \\ &Ecto.Migrator.migrations/1, puts \\ &IO.puts/1) do
     repos = parse_repo(args)
 
-    result = Enum.map(repos, fn repo ->
-      ensure_repo(repo, args)
-      ensure_migrations_path(repo)
-      {:ok, pid, _} = ensure_started(repo, all: true)
+    format = fn x, n -> x |> to_string() |> String.pad_trailing(n) end
```
josevalim: Let's move this to a private function instead of using an anonymous function.

josevalim:
Thanks @aquarhead! One tiny comment and we are good to go!

josevalim:

PR #2684, opened 2018-09-15
reverse modify in migration

### LINKED ISSUE #2682: make modify work with ecto.rollback
* Elixir version (elixir -v): 1.7.0
* Database and version: psql (PostgreSQL) 10.4
* Ecto version (mix deps): ecto 2.2.10
* Database adapter and version (mix deps): postgrex 0.13.5
* Operating system: macOS 10.13.3

see also https://github.com/elixir-ecto/ecto/issues/858,
is it possible to make some modify to `modify` function (like add some option called `:from`) in order to support rollback/down?

```elixir
def change do
  alter table(:user) do
    # when down the change, just read it as "modify :username, :integer"
    # if not supplied with the :from option, just refuse rollback like the behavior now
    modify :username, :string, from: :integer
  end
end
```

otherwise I have to use `execute` for every single modify

```elixir
# lost the generalization and may not work in some database
execute """ALTER TABLE "user" ALTER COLUMN "username" TYPE VARCHAR""",
        """ALTER TABLE "user" ALTER COLUMN "username" TYPE INT"""
```

(and thus function `modify` would be never useful for me)

not to mention it is hard to hand write `execute` command like

```elixir
modify :user_id, references(:users, type: :string), from: :string
```

I really need `ecto.rollback`, but I won't write `down` unless `modify` is the only action in `change` (otherwise, I'd have to revert every action in `down` even if it could be automatically `down`)

clouds56:
This pull request fixes #2682 via adding ability for `modify` to revert.

If use `modify` with `:from` option, e.g.
```elixir
alter table(:some_table) do
  # the option :from is optional, if not supplied, it works just as before (could not revert at all)
  modify :some_field, new_type, from: old_type
end
```
it could revert the change when `ecto.rollback`.

Note that since we get more information in of the `old` type, it is possible to do more thing when modify the type, such as to remove the foreign key constraint for the modified column if exists when the `old_type` is some reference.
(This feature only available for adapters of postgresql now, I think it could be port to mysql as well.)
Be careful when use with rename, you'd like to rename/remove the constraint as well.
```elixir
alter table(:some_table) do
  # remove the constraints some_table_old_name_fkey
  modify :old_name, :old_type, from: reference(:table1, type: :old_type)
end
alter table(:table1) do
  modify :old_name, :new_type, from: :old_type
end
rename table(:some_table), :old_name, to: :new_name
alter table(:some_table) do
  # add the constraints some_table_new_name_fkey
  modify :new_name, reference(:table1, type: :new_type), from: old_type
end
```

All the change would remain compatibility with old code (that is if no `:from` supplied, no constraints would be removed and no `ecto.rollback` could be performed.

### DIFF
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -79,27 +79,21 @@ defmodule Ecto.Integration.MigrationTest do
         add :id, :integer, primary_key: true
       end
       create table(:modify_from_posts) do
-        add :author, references(:modify_from_authors, type: :integer)
+        add :author_id, references(:modify_from_authors, type: :integer)
       end
 
       if direction() == :up do
-        execute "INSERT INTO modify_from_authors (id) VALUES (1)"
-        execute "INSERT INTO modify_from_posts (author) VALUES (1)"
+        flush()
+        PoolRepo.insert_all "modify_from_authors", [[id: 1]]
+        PoolRepo.insert_all "modify_from_posts", [[author_id: 1]]
       end
 
       alter table(:modify_from_posts) do
-        # remove the constraints modify_from_posts_author_fkey
-        modify :author, :integer, from: references(:modify_from_authors, type: :integer)
+        # remove the constraints modify_from_posts_author_id_fkey
+        modify :author_id, :integer, from: references(:modify_from_authors, type: :integer)
       end
       alter table(:modify_from_authors) do
         modify :id, :bigint, from: :integer
-      end
-      case PoolRepo.__adapter__ do
-        Ecto.Adapters.MySQL ->
-          execute ~s{ALTER TABLE modify_from_posts CHANGE author author_id BIGINT UNSIGNED},
-                  ~s{ALTER TABLE modify_from_posts CHANGE author_id author BIGINT UNSIGNED}
-        _ ->
-          rename table(:modify_from_posts), :author, to: :author_id
       end
       alter table(:modify_from_posts) do
         # add the constraints modify_from_posts_author_id_fkey
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -94,7 +94,13 @@ defmodule Ecto.Integration.MigrationTest do
       alter table(:modify_from_authors) do
         modify :id, :bigint, from: :integer
       end
-      rename table(:modify_from_posts), :author, to: :author_id
+      case PoolRepo.__adapter__ do
+        Ecto.Adapters.MySQL ->
+          execute ~s{ALTER TABLE modify_from_posts CHANGE author author_id BIGINT UNSIGNED},
+                  ~s{ALTER TABLE modify_from_posts CHANGE author_id author BIGINT UNSIGNED}
+        _ ->
+          rename table(:modify_from_posts), :author, to: :author_id
+      end
       alter table(:modify_from_posts) do
         # add the constraints modify_from_posts_author_id_fkey
         modify :author_id, references(:modify_from_authors, type: :bigint), from: :integer
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -71,6 +71,37 @@ defmodule Ecto.Integration.MigrationTest do
     end
   end
 
+  defmodule AlterColumnFromMigration do
+    use Ecto.Migration
+
+    def change do
+      create table(:modify_from_authors, primary_key: false) do
+        add :id, :integer, primary_key: true
+      end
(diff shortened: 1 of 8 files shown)


josevalim:
Thank you @clouds56, this looks great! Can you please update MySQL too? You can run all tests with `mix test.all`. I believe we also need an [integration test here](https://github.com/elixir-ecto/ecto/blob/master/integration_test/sql/migration.exs) and to update the docs. Let me know if you have any questions!

clouds56:
`DROP FOREIGN KEY IF EXISTS` for mysql/mariadb wouldn't work until 10.1.26. https://stackoverflow.com/questions/17161496/drop-foreign-key-only-if-it-exists
We could not achieve the exact same thing in mysql in earlier version.

There're 3 options:
1. use `DROP FOREIGN KEY IF EXISTS` and this works only for mariadb 10.1.26
2. use `DROP FOREIGN KEY` directly and it would fail if some rename of the field or drop of the constraint happens before the `modify`, (anyway, they could write `modify some_field, :new_type, from: :bigint` to avoid the `DROP FOREIGN KEY`)
3. do not drop the constraints at all for mysql/mariadb (user could drop it themselves before modify types)

I prefer the option 2, people could either take one of the following action before modify.
a) rename the field to the old name and `modify :some_field, :new_type, from: :references(type: :actual_type)`
b) drop the constraints manually and `modify :some_field, :new_type, from: :actual_type`

clouds56:
Sorry, I don't have a mysql configured my side. (and did not know that mysql adapter doesn't support rename column)
@josevalim, is there any documentation on how to configure a mysql server for test? I suffered from `randomuser` not exists.
and I wonder how RenameColumnMigration could work?

josevalim:
Thanks @clouds56! I would go with this solution for MySQL:

> use DROP FOREIGN KEY IF EXISTS and this works only for mariadb 10.1.26

It is the simplest and if people want to use this feature, then they can update their MySQL version. We can add a tag to the test and disable the tag by default for MySQL.

Also note that Ecto 3.1 will require MySQL 5.7, so it is likely that we will be close to the required MySQL version anyway.

integration_test/sql/migration.exs
```diff
@@ -71,6 +71,43 @@ defmodule Ecto.Integration.MigrationTest do
...
+    def change do
+      create table(:modify_from_authors, primary_key: false) do
+        add :id, :integer, primary_key: true
+      end
+      create table(:modify_from_posts) do
+        add :author, references(:modify_from_authors, type: :integer)
+      end
+
+      if direction() == :up do
+        execute "INSERT INTO modify_from_authors (id) VALUES (1)"
```
josevalim: Let's use `Repo.insert_all "modify_from_authors", [[id: 1]]` and `Repo.insert_all "modify_from_posts", [[author: 1]]`. I think you will have to call `flush()` before though, to ensure the migration is executed before.

integration_test/sql/migration.exs
```diff
@@ -71,6 +71,43 @@ defmodule Ecto.Integration.MigrationTest do
...
+      end
+
+      alter table(:modify_from_posts) do
+        # remove the constraints modify_from_posts_author_fkey
+        modify :author, :integer, from: references(:modify_from_authors, type: :integer)
+      end
+      alter table(:modify_from_authors) do
+        modify :id, :bigint, from: :integer
+      end
+      case PoolRepo.__adapter__ do
```
josevalim: We avoid doing this in tests. Since you have added a tag to the test as `@tag :modify_column_with_from`, we can go to `integration_test/mysql/test_helper.exs` and exclude that tag. But please make sure to leave a TODO on top of the tag saying we can enable the test once we require 5.7+.

integration_test/sql/migration.exs
```diff
@@ -71,6 +71,43 @@ defmodule Ecto.Integration.MigrationTest do
...
+        Ecto.Adapters.MySQL ->
+          execute ~s{ALTER TABLE modify_from_posts CHANGE author author_id BIGINT UNSIGNED},
+                  ~s{ALTER TABLE modify_from_posts CHANGE author_id author BIGINT UNSIGNED}
+        _ ->
+          rename table(:modify_from_posts), :author, to: :author_id
```
josevalim: Is the rename column a necessary part of this test? I would remove it from here, as it means we either cannot run this test for MySQL or we need to special case the MySQL logic, none which are desired.

clouds56:
I thought someone would rename the field or drop the constraint before the modify, but you are right, inconsistency may happen here.

josevalim:

PR #2687, opened 2018-09-18
Introduce stale_error_field option

qcam:
I would like to propose a small behavior change to `Ecto.Repo` upserting with an `on_conflict` query.

Given the following code trying to upsert an entry with a condition in `on_conflict` query.

```elixir
post = %Post{title: "first", uuid: Ecto.UUID.generate()}
MyRepo.insert!(post)

on_conflict = from Post, where: [deleted: false], update: [set: [title: "first"]]
MyRepo.insert(post, on_conflict: on_conflict, conflict_target: [:uuid])
```

The current behavior for this case raises `Ecto.StaleEntryError`, which is a bit confusing IMO, because there isn't any optimistic lock set. I think the ideal behavior in this case could be exactly as `on_conflict: :nothing`.

Suggestions are very appreciated. I could update the needed documentation if this PR is accepted. Thank you.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -591,7 +591,16 @@ defmodule Ecto.Repo.Schema do
       {:invalid, _} = constraints ->
         constraints
       {:error, :stale} ->
-        raise Ecto.StaleEntryError, struct: changeset.data, action: action
+        opts = List.last(args)
+
+        case Keyword.fetch(opts, :stale_error_field) do
+          {:ok, stale_error_field} ->
+            changeset = Ecto.Changeset.add_error(changeset, stale_error_field, "stale")
+            {:error, changeset}
+
+          :error ->
+            raise Ecto.StaleEntryError, struct: changeset.data, action: action
+        end
     end
   end
 
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1447,6 +1447,25 @@ defmodule Ecto.Integration.RepoTest do
       assert TestRepo.get!(Post, inserted.id).title == "second"
     end
 
+    @tag :with_conflict_target
+    test "on conflict query having condition" do
+      post = %Post{title: "first", counter: 1, uuid: Ecto.UUID.generate()}
+      {:ok, inserted} = TestRepo.insert(post)
+
+      on_conflict = from Post, where: [counter: 2], update: [set: [title: "second"]]
+
+      insert_options = [
+        on_conflict: on_conflict,
+        conflict_target: [:uuid],
+        stale_error_field: :counter
+      ]
+
+      assert {:error, changeset} = TestRepo.insert(post, insert_options)
+      assert changeset.errors == [counter: {"stale", []}]
+
+      assert TestRepo.get!(Post, inserted.id).title == "first"
+    end
+
     @tag :without_conflict_target
     test "on conflict replace_all" do
       post = %Post{title: "first", text: "text", uuid: Ecto.UUID.generate()}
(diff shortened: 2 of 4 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -612,7 +612,11 @@ defmodule Ecto.Adapters.SQL do
         {:ok, Enum.zip(returning, values)}
 
       {:ok, %{num_rows: 0}} ->
-        if on_conflict == :nothing, do: {:ok, []}, else: {:error, :stale}
+        if on_conflict == :nothing do
+          {:ok, []}
+        else
+          {:error, :no_change}
```
josevalim: I don't think we should change this code. The change in ecto/repo/schema should be enough. :)

lib/ecto/repo/schema.ex
```diff
@@ -590,8 +590,12 @@ defmodule Ecto.Repo.Schema do
...
       {:invalid, _} = constraints ->
         constraints
-      {:error, :stale} ->
-        raise Ecto.StaleEntryError, struct: changeset.data, action: action
+      {:error, :no_change} ->
+        if changeset.filters != %{} do
+          raise Ecto.StaleEntryError, struct: changeset.data, action: action
+        else
+          {:ok, []}
+        end
```
josevalim: Instead of changing this here, which is used by insert/update/delete, I would probably change the `insert` operation to no longer call this apply function, as we want to handle `{:error, :stale}` differently.

josevalim:
Hi @qcam!

I believe this is a duplicate of #2687. We do want to change this behaviour, but we want to do it behind an option. You can see the discussion in the original issue.

I do think StaleEntryError makes sense here, as the record we are using as reference does not exist (or it was changed).

So my suggestion is to add an option, that will be supported on insert/update/delete. Maybe something called: `add_stale_errors_to: :field_name`. This will convert the stale error into a changeset error on the `field_name`.

Also, can you please add some unit tests? I have just pushed some tests to test/ecto/repo_test that check for staleentryerrors and you can use them as template. Make sure to rebase before! Thanks!

qcam:
@josevalim I changed the behavior as what you suggested, please let me know if I misunderstood anything.

I think `:add_stale_errors_to` is rather long so I went with `stale_error_field`, but I don't have a strong opinion on that so do advise if there's a better name.

I also made an assumption that we still want to keep raising `Ecto.StaleEntryError` when there's no `stale_error_field` configured.

I suppose some documentations need updates. I could do it later if the implementation looks fine.

josevalim:
This is beautiful! We just need the docs and we are ready to go!

josevalim:

lib/ecto/repo/schema.ex
```diff
@@ -591,7 +591,16 @@ defmodule Ecto.Repo.Schema do
       {:invalid, _} = constraints ->
         constraints
       {:error, :stale} ->
-        raise Ecto.StaleEntryError, struct: changeset.data, action: action
+        opts = List.last(args)
+
+        case Keyword.fetch(opts, :stale_error_field) do
+          {:ok, stale_error_field} ->
+            changeset = Ecto.Changeset.add_error(changeset, stale_error_field, "stale")
```
wojtekmach: great work @qcam. I'd like to propose a couple minor changes:

1. we should probably change clause to be a bit stricter:

   ```elixir
   {:ok, stale_error_field} when is_atom(stale_error_field) ->
   ```

2. add error metadata, e.g.: `add_error(changeset, stale_error_field, message, [stale: true])`. This way, it's similar to `:validation` and `:constraint` error metadata keys, so we can programatically handle this error without trying to look for the message string.

3. change `"stale"` to `"is stale"`, or maybe even: `Keyword.get(opts, :stale_error_message, "is stale")`

wdyt?

qcam: @wojtekmach your proposal sounds great, I'm up for changing but the PR is merged. Shall I open a new one with these changes?

josevalim: Yes, please open a new one! :D

PR #2688, opened 2018-09-18
Make `dump` stricter for Decimal

wojtekmach:
Previously we only made cast stricter.


josevalim:

PR #2689, opened 2018-09-18
Add notes about :decimal and calendar types

wojtekmach:
Since we enforce the type, this fails because the default timestamp is `:naive_datetime`, so without microseconds.

```elixir
iex(1)> Friends.Repo.insert(%Friends.Person{inserted_at: NaiveDateTime.utc_now()})
** (Ecto.ChangeError) value `~N[2018-09-18 10:10:51.596624]` for `Friends.Person.inserted_at` in `insert` does not match type :naive_datetime
    (ecto) lib/ecto/repo/schema.ex:831: Ecto.Repo.Schema.dump_field!/6
    (ecto) lib/ecto/repo/schema.ex:840: anonymous fn/6 in Ecto.Repo.Schema.dump_fields!/5
    (stdlib) maps.erl:257: :maps.fold_1/3
    (ecto) lib/ecto/repo/schema.ex:838: Ecto.Repo.Schema.dump_fields!/5
    (ecto) lib/ecto/repo/schema.ex:780: Ecto.Repo.Schema.dump_changes!/6
    (ecto) lib/ecto/repo/schema.ex:219: anonymous fn/15 in Ecto.Repo.Schema.do_insert/3
```

I think ideally the error would provide more information. Should I investigate? 

Mentioned to me by @AndrewDryga. WDYT about these docs update, Andrew?

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -235,17 +235,21 @@ defmodule Ecto.Schema do
 
   **Notes:**
 
-  - For the `{:array, inner_type}` and `{:map, inner_type}` type,
-    replace `inner_type` with one of the valid types, such as `:string`.
+    * For the `{:array, inner_type}` and `{:map, inner_type}` type,
+      replace `inner_type` with one of the valid types, such as `:string`.
 
-  - For the `:decimal` type, `+Infinity`, `-Infinity`, and `NaN` values are not supported,
-    even though the `Decimal` library handles them. To support them, create a custom type.
+    * For the `:decimal` type, `+Infinity`, `-Infinity`, and `NaN` values
+      are not supported, even though the `Decimal` library handles them.
+      To support them, you can create a custom type.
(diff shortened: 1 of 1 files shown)


josevalim:
@wojtekmach i did a pass for formatting and clarity, let me know what you think!

wojtekmach:
@josevalim :shipit:

AndrewDryga:
@wojtekmach I think we should additionally mention that when you insert a struct or update with changes that are not casted, you need to take care of precision manually.

Your code example is a good one, and here is the one we discussed:
```
channel
|> Ecto.Changeset.change(%{
  verification_token_hash: nil,
  verification_token: nil,
  verification_token_sent_at: nil,
  verified_at: DateTime.utc_now()
})
|> Repo.update!()
```

Having better error would ease migration for noobs like me :).

josevalim:

PR #2696, opened 2018-09-20
Allow integration tests to run on PostgreSQL 10.4

AndrewDryga:
I just tried to run integration tests locally with official PostgreSQL 10.4 container, got the following error:

```
MIX_ENV=pg mix test
** (Version.InvalidVersionError) invalid version: "10.4 (Debian 10.4-2.pgdg90+1)"
    (elixir) lib/version.ex:337: Version.to_matchable/2
    (elixir) lib/version.ex:179: Version.match?/3
    integration_test/pg/test_helper.exs:75: (file)
    (elixir) lib/code.ex:767: Code.require_file/2
```

I think using new regex would help us to not get back to PG version match issue over and over again.

### DIFF
--- a/integration_test/pg/test_helper.exs
+++ b/integration_test/pg/test_helper.exs
@@ -67,7 +67,7 @@ _   = Ecto.Adapters.Postgres.storage_down(TestRepo.config)
 %{rows: [[version]]} = TestRepo.query!("SHOW server_version", [])
 
 version =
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks! We should probably change [this line](https://github.com/elixir-ecto/ecto/blob/b00db1ce96514209790bad11a5a35948668cbc72/integration_test/pg/test_helper.exs#L76) to be `>= 9.5` as well.

josevalim:

Issue #2694, opened 2018-09-20
It's not possible to know was changeset inserted or not on Upsert


AndrewDryga:
### Environment

* Elixir version (elixir -v): 1.7.2
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 10
* Ecto version (mix deps): master, commit bba34c5ba218537bf196630179e4934946ab94e0
* Database adapter and version (mix deps): 6cabc3b018cabc6e036e7d162414889a20785b96
* Operating system: MacOS

### Current behavior

Quote from Ecto.Repo docs:
> Now we can insert with the same title but do nothing on conflicts:
> ```elixir
> {:ok, ignored} = MyRepo.insert(%Post{title: "this is unique"}, on_conflict: :nothing)
> 
> assert ignored.id == nil
> ```

This way of determining was something inserted or not only works when you use structs, but it fails with changesets, here is pseudo-code:

```elixir
post = Post.changeset(%{"title" => "this is unique"})
{:ok, ignored} = MyRepo.insert(post, on_conflict: :nothing, returning: true)
assert ignored.id == # it's not nil any more, but a randomly generated UUID
```

I believe it’s because ID is generated when we turn changeset into schema right before inserting it, not sure why Ecto does not read the ID back though.

(My specific use case is inserting changeset via `Ecto.Multi.insert`, but it should not be relevant. Primary keys are set to `:binary_id, autogenerate: true`.)

### Expected behavior

A way to determine was changeset inserted or not.

cc @wojtekmach

josevalim:
In cases like that we simply can't know because the database does not return us any information. So you need to have at least some field with read_after_writes set to true or let the database generate the UUID for you. You can also perform a separate read  but there is nothing "built-in".

AndrewDryga:
Thank you, got it. As a workaround for people came from Google I'm removing `on_conflict` from `insert/2` and adding a unique constraint to the changeset. So I can read back inserted struct only when conflict occurred and proceed with loaded data.

AndrewDryga:
@josevalim just to confirm, `returning: true` also doesn't help there? I thought it should read all the fields back? Similar to having `read_after_writes` set on all the fields.

AndrewDryga:
@josevalim should not just override all fields in the returned schema? I'm not sure how to define the diff there because I do not get two copies of the data.

josevalim:
>  should not just override all fields in the returned schema?

It should.

[...]

PR #2697, opened 2018-09-20
Returning should return actual record ID on insert conflict

AndrewDryga:
Related #2694. 

This PR is based on another one that fixes the version issue (#2696), I'll rebase it's merged.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1418,6 +1418,17 @@ defmodule Ecto.Integration.RepoTest do
       assert not_inserted.id == nil
     end
 
+    @tag :returning
+    @tag :with_conflict_target
+    test "on conflict ignore and returning (changeset)" do
+      custom_changeset = Custom.changeset(%Custom{}, %{uuid: Ecto.UUID.generate()})
+      {:ok, inserted} = TestRepo.insert(custom_changeset, on_conflict: :nothing, conflict_target: [:uuid])
+      assert inserted.bid
+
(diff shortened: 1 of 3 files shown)


integration_test/cases/repo.exs
```diff
@@ -1418,6 +1418,17 @@ defmodule Ecto.Integration.RepoTest do
...
 
+    @tag :returning
+    @tag :with_conflict_target
+    test "on conflict ignore and returning (changeset)" do
+      custom_changeset = Custom.changeset(%Custom{}, %{uuid: Ecto.UUID.generate()})
+      {:ok, inserted} = TestRepo.insert(custom_changeset, on_conflict: :nothing, conflict_target: [:uuid])
+      assert inserted.bid
+
+      {:ok, not_inserted} = TestRepo.insert(custom_changeset, on_conflict: :nothing, conflict_target: [:uuid], returning: true)
+      assert not_inserted.bid == inserted.bid
```
AndrewDryga: I believe this should be true.

josevalim: on_conflcit nothing the database is not going to return anything information. There is nothing we can do about it. You have to do a separate query.

josevalim: The returning true only applies if the operation succeeds.

josevalim: To be super clear, now you can precisely use the fact that the `bid` is not there to know that the operation failed. This means that now you can do a `Repo.get` and get the latest version. :) If bid was there, then the operation suceeded, so you don't need a separate query.

I know this is annoying but we are doing all we can with what the database returns.

PR #2700, opened 2018-09-21
 index/3 raises if given multiple `where:` clauses

### LINKED ISSUE #2690: `Ecto.Migration.index/3` silently mishandles multiple `:where` clauses
### Environment

* Elixir version (elixir -v): Elixir 1.7.1 (compiled with Erlang/OTP 21)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 10.2
* Database adapter and version (mix deps): Postgrex 0.13.5
* Operating system: MacOS High Sierra (10.13.6)

## Problem

`Ecto.Query.from` allows multiple `where:` clauses. So it seems natural to use multiple `where:` clauses when creating an index with `Ecto.Migration`. However, doing so fails in a surprising way.

### Current behavior

This statement:

```elixir
    create(
      index(
        "seasons",
        :team_id,
        where: "started_at is not null",
        where: "ended_at is null",
        unique: true,
        name: "only_one_active_season_per_team"
      )
    )
```

...creates an index where only the second `where:` clause is used.

### Expected behavior

Either `index/3` should combine all `where:` clauses using `AND`, or it should raise an error when multiple `where:` clauses are given, suggesting that the user combine them.

nathanl:
Because multiple `where:` keywords are supported by Ecto.Query, users
may provide multiple `where:` keywords when declaring a partial index.
However, this is not supported.

Previously, this would fail in a surprising way: the final `where:`
would be used and the other discarded. Now we raise an error, explaining
that these clauses should be combined into a single string.

Fixes #2690

### DIFF
--- a/lib/ecto/migration.ex
+++ b/lib/ecto/migration.ex
@@ -582,6 +582,7 @@ defmodule Ecto.Migration do
   end
 
   def index(table, columns, opts) when is_binary(table) and is_list(columns) and is_list(opts) do
+    validate_index_opts!(opts)
     index = struct(%Index{table: table, columns: columns}, opts)
     %{index | name: index.name || default_index_name(index)}
   end
@@ -955,6 +956,17 @@ defmodule Ecto.Migration do
   defp validate_type!(%Reference{} = reference) do
     reference
   end
+
+  defp validate_index_opts!(opts) when is_list(opts) do
+    case Keyword.get_values(opts, :where) do
+      [_, _ | _] ->
+        raise ArgumentError, message:
+       "only one 'where' keyword is supported when declaring a partial index. To specify multiple conditions, write a single WHERE clause using AND between them."
+      _ -> :ok
+    end
+  end
+
+  defp validate_index_opts!(opts), do: opts
 
   @doc false
   def __prefix__(%{prefix: prefix} = index_or_table) do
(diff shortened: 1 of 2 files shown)


lib/ecto/migration.ex
```diff
@@ -956,6 +960,21 @@ defmodule Ecto.Migration do
...
 
+  defp validate_index_opts!(opts) when is_list(opts) do
+    where_count = Enum.count(opts, fn entry ->
+      case entry do
+        {:where, _} -> true
+        _ -> false
+      end
+    end)
+    if where_count > 1 do
+      raise Index.Error, message:
```
josevalim: We don't need a new exception for this, ArgumentError is fine

lib/ecto/migration.ex
```diff
@@ -956,6 +960,21 @@ defmodule Ecto.Migration do
     reference
   end
 
+  defp validate_index_opts!(opts) when is_list(opts) do
+    where_count = Enum.count(opts, fn entry ->
```
josevalim: Maybe:

```elixir
case Keyword.get_values(opts, :where) do
  [_, _ | _] -> raise ...
  _ -> :ok
end
```

nathanl:
@josevalim Thanks for the feedback! Updated.

josevalim:

PR #2702, opened 2018-09-22
Filter out unknown fields when loading changes

### LINKED ISSUE #2699: Returned entity contains nonexistent field
### Environment

Ecto 2.2.10
Elixir 1.7.3
macOS High Sierra 10.13.6

### Current behavior

Trying to change a nonexistent field in a post seems fine:

```
iex(56)> %Post{} |> change(%{nonexistent_field: "thing"}) |> apply_changes
%EctoAssoc.Post{
  __meta__: #Ecto.Schema.Metadata<:built, "posts">,
  body: nil,
  header: nil,
  id: nil,
  lock_version: 1,
  tags: #Ecto.Association.NotLoaded<association :tags is not loaded>,
  user: #Ecto.Association.NotLoaded<association :user is not loaded>,
  user_id: nil
}
```

Now doing similarly but this time inserting the post, the insert succeeds but 
the returned struct contains the `nonexistent_field` field:

```
iex(57)> %Post{} |> change(%{nonexistent_field: "thing"}) |> Repo.insert!

14:44:41.979 [debug] QUERY OK db=1.7ms
begin []

14:44:41.983 [debug] QUERY OK db=3.1ms
INSERT INTO "posts" ("lock_version") VALUES ($1) RETURNING "id" [1]

14:44:41.985 [debug] QUERY OK db=2.6ms
commit []
%{
  __meta__: #Ecto.Schema.Metadata<:loaded, "posts">,
  __struct__: EctoAssoc.Post,
  body: nil,
  header: nil,
  id: 6,
  lock_version: 1,
  nonexistent_field: "thing",
  tags: #Ecto.Association.NotLoaded<association :tags is not loaded>,
  user: #Ecto.Association.NotLoaded<association :user is not loaded>,
  user_id: nil
}
iex(58)>
```

### Expected behavior

Either one of these:

1. Passing a nonexisting field to `change` or `cast` should raise an `Ecto.UnknownFieldError`

1. The `Repo` operations must not add nonexisting fields to their results.

1. some other solution

I prefer the first because a program using an unkown field clearly has a bug.

qcam:
This PR is an attempt to fix #2699.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -677,12 +677,21 @@ defmodule Ecto.Repo.Schema do
     # already filtered the results to be only about fields.
     data =
       changeset.data
-      |> Map.merge(changes)
+      |> merge_changes(changes)
       |> Map.merge(embeds)
       |> merge_autogen(autogen)
       |> apply_metadata(state, schema_meta)
       |> load_each(values, types, adapter)
     Map.put(changeset, :data, data)
+  end
+
+  defp merge_changes(data, changes) do
+    changes =
+      Enum.reduce(changes, changes, fn {key, _value}, changes ->
+        if Map.has_key?(data, key), do: changes, else: Map.delete(changes, key)
+      end)
+
+    Map.merge(data, changes)
   end
 
   defp merge_autogen(data, autogen) do
--- a/test/ecto/repo_test.exs
+++ b/test/ecto/repo_test.exs
@@ -358,6 +358,22 @@ defmodule Ecto.RepoTest do
       assert {:ok, %MySchema{}} = TestRepo.delete(valid)
     end
 
+    test "insert, update, insert_or_update and delete filters out unknown field" do
+      valid = Ecto.Changeset.change(%MySchema{id: 1}, %{unknown: "foo"})
+
+      assert {:ok, %MySchema{} = inserted} = TestRepo.insert(valid)
+      refute Map.has_key?(inserted, :unknown)
+
+      assert {:ok, %MySchema{} = updated} = TestRepo.update(valid)
+      refute Map.has_key?(updated, :unknown)
+
+      assert {:ok, %MySchema{} = upserted} = TestRepo.insert_or_update(valid)
+      refute Map.has_key?(upserted, :unknown)
+
+      assert {:ok, %MySchema{} = deleted} = TestRepo.delete(valid)
+      refute Map.has_key?(deleted, :unknown)
+    end
+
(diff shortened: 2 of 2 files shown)


integration_test/cases/repo.exs
```diff
@@ -106,6 +106,18 @@ defmodule Ecto.Integration.RepoTest do
            TestRepo.delete!(updated)
   end
 
+  @tag :focus
+  test "insert, update and delete with invalid field" do
+    post = %Post{title: "title"}
+    changeset =
+      post
+      |> Ecto.Changeset.change()
+      |> Ecto.Changeset.put_change(:nonexistent_field, "foo")
```
qcam: @josevalim Could you please advise what would be the desired behavior in this case?

wojtekmach: I can add some historical context: https://github.com/elixir-ecto/ecto/pull/1883#issuecomment-269858878. Back then, the consensus was that it's OK to store some additional information on the changeset.

At the same time though, I agree that non-existing fields shouldn't end up on the final struct so maybe we should filter them out before returning from repo operations.

qcam: thanks, the PR is updated.

lib/ecto/repo/schema.ex
```diff
@@ -677,14 +677,22 @@ defmodule Ecto.Repo.Schema do
...
+      |> merge_changes(changes)
       |> Map.merge(embeds)
       |> merge_autogen(autogen)
       |> apply_metadata(state, schema_meta)
       |> load_each(values, types, adapter)
     Map.put(changeset, :data, data)
   end
 
+  defp merge_changes(data, changes) do
+    Enum.reduce(changes, data, fn {key, value}, data ->
```
josevalim: The most efficient approach is actually to remove any unknown change and then still call `Map.merge`  at the end, as it requires less modifications. :)

josevalim: so something like:
```
changes = Enum.reduce(changes, changes, fn {key, value}, acc -> if not in key remove from acc end)
Map.merge(data, changes)
```

test/ecto/repo_test.exs
```diff
@@ -358,6 +358,22 @@ defmodule Ecto.RepoTest do
       assert {:ok, %MySchema{}} = TestRepo.delete(valid)
     end
 
+    test "insert, update, insert_or_update and delete filters nonexisitent field" do
```
AndrewDryga: You have a typo in `nonexisitent`.

josevalim:

PR #2707, opened 2018-09-24
Add statement cache to insert/update/delete

josevalim:
This makes introducing the Ecto.Integration.Post schema a thousand times in a row almost twice faster since we avoid a round trip to the database.

### DIFF
--- a/lib/ecto/adapters/sql.ex
+++ b/lib/ecto/adapters/sql.ex
@@ -62,6 +62,12 @@ defmodule Ecto.Adapters.SQL do
   `DBConnection` library.
   """
 
+  @type result_set :: %{
+    :rows => nil | [[term] | binary],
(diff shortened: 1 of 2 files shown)


lib/ecto/adapters/sql.ex
```diff
@@ -320,18 +330,41 @@ defmodule Ecto.Adapters.SQL do
...
+
+    case sql_call(adapter_meta, op, args, params, opts) do
+      {:ok, query, result} ->
+        :ets.insert(cache, {name, query})
+        {:ok, result}
+
+      {:ok, _} = ok ->
+        ok
+
+      {:error, _} = error ->
```
fishcakez: never reset cache?

josevalim:
Closing this for now as the one in Postgres will have less contention (at the cost of memory).

PR #2711, opened 2018-09-25
Make updated_at have the same value as inserted_at

### LINKED ISSUE #2633: Timestamps: updated_at greater that inserted_at on creation
### Environment

* Elixir version (elixir -v): Elixir 1.5.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6
* Database adapter and version (mix deps): postgrex 0.13.5
* Operating system: Mac OS 

### Current behavior

I found that `updated_at` is always greater then `inserted_at` timestamp. Is there any reason why autogen works like this?

### Expected behavior

Once record created `updated_at` equals to `inserted_at`

qcam:
An attempt to fix #2633. Kudos to @fertapric for the idea.

### DIFF
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -159,13 +159,17 @@ defmodule Ecto.Embedded do
   end
 
   defp autogenerate(changes, action, schema) do
-    Enum.reduce schema.__schema__(action_to_auto(action)), changes, fn
-      {k, {mod, fun, args}}, acc ->
-        case Map.fetch(acc, k) do
-          {:ok, _} -> acc
-          :error   -> Map.put(acc, k, apply(mod, fun, args))
-        end
-    end
+    autogen_fields = schema.__schema__(action_to_auto(action))
+
+    Enum.reduce(autogen_fields, changes, fn {fields, {method, fun, args}}, acc ->
+      generated = apply(method, fun, args)
+
+      Enum.reduce(fields, acc, fn field, acc2 ->
+        if Map.has_key?(acc2, field),
+           do: acc2,
+           else: Map.put(acc2, field, generated)
+      end)
+    end)
   end
 
   defp action_to_auto(:insert), do: :autogenerate
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -159,16 +159,12 @@ defmodule Ecto.Embedded do
   end
 
   defp autogenerate(changes, action, schema) do
-    autogen_fields = schema.__schema__(action_to_auto(action))
+    autogen_fields = action |> action_to_auto() |> schema.__schema__()
 
-    Enum.reduce(autogen_fields, changes, fn {fields, {method, fun, args}}, acc ->
-      generated = apply(method, fun, args)
-
-      Enum.reduce(fields, acc, fn field, acc2 ->
-        if Map.has_key?(acc2, field),
-           do: acc2,
-           else: Map.put(acc2, field, generated)
-      end)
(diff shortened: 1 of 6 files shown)


fertapric:
I miss some tests asserting that both values are the same 🙂

lib/ecto/embedded.ex
```diff
@@ -159,13 +159,17 @@ defmodule Ecto.Embedded do
...
-      {k, {mod, fun, args}}, acc ->
-        case Map.fetch(acc, k) do
-          {:ok, _} -> acc
-          :error   -> Map.put(acc, k, apply(mod, fun, args))
-        end
-    end
+    autogen_fields = schema.__schema__(action_to_auto(action))
+
+    Enum.reduce(autogen_fields, changes, fn {fields, {method, fun, args}}, acc ->
+      generated = apply(method, fun, args)
```
josevalim: We should only `apply` if at least one field is not set. So I think we need to change things a bit. Something like:
```elixir
Enum.reduce(autogen_fields, changes, fn {fields, {method, fun, args}}, acc ->
  case Enum.reject(fields, Map.has_key?(acc, &1)) do
    [] ->
      acc

    fields ->
      generated = apply(method, fun, args)
      Enum.reduce(fields, acc, &Map.put(&2, &1, generated))
  end
end)
```

lib/ecto/repo/schema.ex
```diff
@@ -800,16 +800,12 @@ defmodule Ecto.Repo.Schema do
...
-      generated = apply(mod, fun, args)
-
-      Enum.flat_map(fields, fn field ->
-        if Map.has_key?(changes, field),
-           do: [],
-           else: [{field, generated}]
-      end)
+      fields
+      |> Enum.reject(&Map.has_key?(changes, &1))
+      |> Enum.map(&{&1, apply(mod, fun, args)})
```
fertapric: I think this is equivalent to what we have before. 

For each field, it calls the auto-generate function passed as the second argument. The idea is to generate the value once, and then assign that same value to all the fields :)

test/ecto/repo/autogenerate_test.exs
```diff
@@ -85,6 +85,7 @@ defmodule Ecto.Repo.AutogenerateTest do
     default = TestRepo.insert!(%Company{})
     assert %NaiveDateTime{microsecond: {0, 0}} = default.inserted_at
     assert %NaiveDateTime{microsecond: {0, 0}} = default.updated_at
+    assert default.inserted_at == default.updated_at
```
fertapric: I would either include this assertion in all the tests, or [here](https://github.com/elixir-ecto/ecto/blob/f676c115210894287a89a8f43a7dd9b4ef15ba7a/test/ecto/repo/autogenerate_test.exs#L170).

I think this is a flaky test with the current implementation.

lib/ecto/repo/schema.ex
fertapric: I would directly copy José's code :)

qcam: arrrrrh that's right 🤦‍♂️.

josevalim:

josevalim:
Thank you @qcam for all the help in all of those issues and thank you @fertapric for the amazing reviews! \o/

josevalim:
Hi @qcam and @fertapric! I have invited you to the contributors team in the Ecto organization! The invitation is just to formalize your contributions. You don't have to worry about it or change anything in the way you are contributing to the project. Thank you for all of your work!

PR #2718, opened 2018-09-27
Revert mix application change

wojtekmach:
Ref: https://github.com/elixir-ecto/ecto/commit/cb1fa78300d02200c082a7f29fd57ea65275e921#diff-6023be6004fce4718dad3dafb576d258R35

The previous code was correct or I'm missing something!

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -32,7 +32,7 @@ defmodule Ecto.MixProject do
 
   def application do
     [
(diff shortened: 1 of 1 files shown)


josevalim:
Yup, I have reverted it already. There was a mixture of issues since build_per_environment was disabled but i believe they are all fixed now.

PR #2720, opened 2018-09-28
Log Postgrex notice when running migration

### LINKED ISSUE #2463: Show Postgres notices when running migrations
#### Elixir version

Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.6.3 (compiled with OTP 19

#### Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):

postgres (PostgreSQL) 10.1

#### Ecto version (mix deps):

ecto 2.1.6

#### Database adapter and version (mix deps):

postgrex 0.13.5 (Hex package) (mix)

#### Operating system:

MacOs 10.12.6

### Current behavior

Postgres just silently truncate very long index name to 63 bytes:

https://til.hashrocket.com/posts/8f87c65a0a-postgresqls-max-identifier-length-is-63-bytes

Also, because of this the `unique_constraint()` validation suggest the full index name, while in DB the index name is truncated.

### Expected behavior

`Ecto.Migration.Runner` should show the notice from postgres, or just exit with error, like it is done in ActiveRecord (but it may break old ecto migrations)

I didn't check how it is working with table names, column names, etc... But should have the same silent truncating behavior

qcam:
Fix #2463.

### DIFF
--- a/integration_test/pg/constraints_test.exs
+++ b/integration_test/pg/constraints_test.exs
@@ -32,7 +32,6 @@ defmodule Ecto.Integration.ConstraintsTest do
   end
 
   setup_all do
-    PoolRepo.delete_all(Ecto.Migration.SchemaMigration)
     up(PoolRepo, 20040906120000, ConstraintMigration, log: false)
   end
 
--- a/integration_test/pg/constraints_test.exs
+++ b/integration_test/pg/constraints_test.exs
@@ -1,7 +1,6 @@
 defmodule Ecto.Integration.ConstraintsTest do
-  # We can keep this test async as long as it
-  # is the only one access the constraints table
-  use ExUnit.Case, async: true
+  # Cannot be async as other tests may migrate
+  use ExUnit.Case
 
   alias Ecto.Integration.PoolRepo
   import Ecto.Migrator, only: [up: 4]
@@ -33,6 +32,7 @@ defmodule Ecto.Integration.ConstraintsTest do
   end
 
   setup_all do
+    PoolRepo.delete_all(Ecto.Migration.SchemaMigration)
     up(PoolRepo, 20040906120000, ConstraintMigration, log: false)
   end
 
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -339,6 +339,8 @@ defmodule Ecto.Integration.MigrationTest do
     :ok
   end
 
+  @moduletag capture_log: true
+
   test "create and drop table and indexes" do
     assert :ok == up(PoolRepo, 20050906120000, CreateMigration, log: false)
     assert :ok == down(PoolRepo, 20050906120000, CreateMigration, log: false)
--- a/integration_test/sql/migration.exs
+++ b/integration_test/sql/migration.exs
@@ -339,8 +339,6 @@ defmodule Ecto.Integration.MigrationTest do
     :ok
   end
 
-  @moduletag capture_log: true
-
   test "create and drop table and indexes" do
     assert :ok == up(PoolRepo, 20050906120000, CreateMigration, log: false)
     assert :ok == down(PoolRepo, 20050906120000, CreateMigration, log: false)
@@ -377,10 +375,12 @@ defmodule Ecto.Integration.MigrationTest do
     assert :ok == down(PoolRepo, 20050906120000, ReferencesRollbackMigration, log: false)
   end
 
+  @tag capture_log: true
   test "create table if not exists and drop table if exists does not raise on failure" do
     assert :ok == up(PoolRepo, 20050906120000, NoErrorTableMigration, log: false)
   end
 
+  @tag capture_log: true
   @tag :create_index_if_not_exists
   test "create index if not exists and drop index if exists does not raise on failure" do
     assert :ok == up(PoolRepo, 20050906120000, NoErrorIndexMigration, log: false)
--- a/lib/ecto/adapter/migration.ex
+++ b/lib/ecto/adapter/migration.ex
@@ -54,7 +54,7 @@ defmodule Ecto.Adapter.Migration do
       to finish, `:infinity` will wait indefinitely (default: 5000);
     * `:log` - When false, does not log begin/commit/rollback queries
   """
-  @callback execute_ddl(adapter_meta, command, options :: Keyword.t()) :: {:ok, list()}
+  @callback execute_ddl(adapter_meta, command, options :: Keyword.t()) :: [{Logger.level, Logger.message, Logger.metadata}]
 
   @doc """
   Locks the migrations table and emits the locked versions for callback execution.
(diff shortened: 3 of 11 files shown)


integration_test/sql/migration.exs
```diff
@@ -349,6 +358,12 @@ defmodule Ecto.Integration.MigrationTest do
     assert :ok == down(PoolRepo, 20050906120000, InferredDropIndexMigration, log: false)
   end
 
+  @tag :very_long_index
+  test "create very long index name" do
```
qcam: Unfortunately I haven't been able to make Postgrex return a hint yet 🤔. Anyone has a hint?

josevalim: Here is a migration that emits a warning:

```
    def change do
      create_if_not_exists table(:duplicate_table)
      create_if_not_exists table(:duplicate_table)
    end
```

but this does not work right now. So I will push a fix for it and then you can try again on master.

Also, this is very likely a PG specific warning, so we will have to move this test to inside `integration_test/pg/migrations_test.exs`. See the existing `constraints_test.exs` for an example.

josevalim: Ok, [I pushed to master](29adbb95). The migration above should work and emit warnings.

lib/ecto/migration/runner.ex
```diff
@@ -263,10 +263,27 @@ defmodule Ecto.Migration.Runner do
...
+  defp log_notices(level, [message | notices]) do
+    # TODO: what would we want to log??
+    %{
+      code: _code,
+      file: _file,
+      line: _line,
+      message: message,
+      routine: _routine,
+      severity: _severity
+    } = message
```
josevalim: I don't think we should access this, as this internal representation will change per adapter. My suggestion is for us to add a new callback to `Ecto.Adapters.SQL.Connection` called `ddl_logs` that receives each `Postgrex.Result` returned in `execute_ddl` and it must return a list in this shape: `{level, msg, metadata}`. Then on this side, we should call `Logger.log(level, msg, metadata)`.

josevalim: If we do in this direction, we should probably change [this](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapter/migration.ex#L57) so the return result is no longer `list()` but rather something like `[{Logger.level, Logger.message, Logger.metadata}]`.

integration_test/sql/migration.exs
qcam: Thanks! I'll fix this.

integration_test/sql/migration.exs
```diff
@@ -339,6 +339,8 @@ defmodule Ecto.Integration.MigrationTest do
     :ok
   end
 
+  @moduletag capture_log: true
```
josevalim: Let's set `@tag capture_log: true` so we don't silence the whole test case by accident.

lib/ecto/adapter/migration.ex
```diff
@@ -54,7 +54,7 @@ defmodule Ecto.Adapter.Migration do
       to finish, `:infinity` will wait indefinitely (default: 5000);
     * `:log` - When false, does not log begin/commit/rollback queries
   """
-  @callback execute_ddl(adapter_meta, command, options :: Keyword.t()) :: {:ok, list()}
+  @callback execute_ddl(adapter_meta, command, options :: Keyword.t()) :: [{Logger.level, Logger.message, Logger.metadata}]
```
josevalim: Sorry, I don't think it was precise though, it should still be `{:ok, [{...}]}`. It is important to return `{:ok, something}` in case we want to return `:error` in the future.

lib/ecto/adapters/sql.ex
```diff
@@ -187,7 +187,9 @@ defmodule Ecto.Adapters.SQL do
 
       @doc false
       def execute_ddl(meta, definition, opts) do
-        Ecto.Adapters.SQL.execute_ddl(meta, @conn.execute_ddl(definition), opts)
+        meta
+        |> Ecto.Adapters.SQL.execute_ddl(@conn.execute_ddl(definition), opts)
+        |> Enum.flat_map(&@conn.ddl_logs/1)
```
josevalim: It is best to avoid doing the work here and instead do it inside `Ecto.Adapters.SQL`. So I propose to call something like `Ecto.Adapters.SQL.execute_ddl(meta, @conn, definition, opts)` and do all of the work there.

lib/ecto/adapters/sql.ex
```diff
@@ -663,8 +665,6 @@ defmodule Ecto.Adapters.SQL do
     for sql <- List.wrap(sqls) do
       query!(meta, sql, [], opts)
     end
-
-    {:ok, []}
```
josevalim: As mentioned, let's continue returning `{:ok, ...}`. :) This is important in case we want to return `:error` or something similar in the future.

lib/ecto/adapters/sql/connection.ex
```diff
@@ -104,4 +104,9 @@ defmodule Ecto.Adapters.SQL.Connection do
   Receives a DDL command and returns a query that executes it.
   """
   @callback execute_ddl(command :: Ecto.Adapter.Migration.command) :: String.t | [iodata]
+
+  @doc """
+  Receives a DDL command and returns a query that executes it.
+  """
+  @callback ddl_logs(result :: term) :: [{Logger.level, Logger.message, Logger.metadata}]
```
josevalim: Beautiful!

lib/ecto/adapters/sql/connection.ex
```diff
@@ -104,4 +104,9 @@ defmodule Ecto.Adapters.SQL.Connection do
   Receives a DDL command and returns a query that executes it.
   """
   @callback execute_ddl(command :: Ecto.Adapter.Migration.command) :: String.t | [iodata]
+
+  @doc """
+  Receives a DDL command and returns a query that executes it.
```
qcam: @qcam fix it.

integration_test/pg/constraints_test.exs
```diff
@@ -33,6 +32,7 @@ defmodule Ecto.Integration.ConstraintsTest do
   end
 
   setup_all do
+    PoolRepo.delete_all(Ecto.Migration.SchemaMigration)
```
josevalim: Why is this necessary? Nothing else should have ran the migrations in this file, right?

qcam: Somehow I got a warning when running `ECTO_ADAPTER=pg mix test` saying that a migration with a higher timestamp has been run (in `integration_tests/sql/migration.exs`) 🤔. Did I miss out anything?

qcam:
@josevalim thank you so much for the great reviews ❤️!

integration_test/pg/constraints_test.exs
qcam: I deleted the line and fortunately [this Travis job](https://travis-ci.org/elixir-ecto/ecto/jobs/435019781) has the warning I mention.

josevalim: Oh, I see. Leave the warning on and async: true, we will find a way to fix it. :)

josevalim:

PR #2721, opened 2018-09-30
Update migration docs

gj:


lib/ecto/migration/schema_migration.ex
```diff
@@ -1,6 +1,8 @@
 defmodule Ecto.Migration.SchemaMigration do
-  # Define a schema that works with the a table, which is schema_migrations by default
```
gj: Started off just meaning to fix the small typo here (`with the a table`), but ended up reading through the `Ecto.Migration` docs as well and fixing up the first section there.

lib/ecto/migration.ex
```diff
@@ -67,14 +67,14 @@ defmodule Ecto.Migration do
         end
       end
 
-  Notice not all commands are reversible though. Trying to rollback
+  However, note that not all commands are reversible. Trying to rollback
```
fertapric: I think it should be “roll back” instead of “rollback” when it’s a verb 🙂

lib/ecto/migration.ex
```diff
@@ -457,7 +459,7 @@ defmodule Ecto.Migration do
   end
 
   @doc """
-  Returns a table struct that can be given on create, alter, etc.
+  Returns a table struct that can be given to create, alter, etc.
```
fertapric: What do you think about?

> Returns a table struct that can be given to `create/2`, `alter/2`, `drop/1`, etc.

lib/ecto/migration.ex
```diff
@@ -496,20 +499,20 @@ defmodule Ecto.Migration do
...
-    * `:concurrently` - if the index should be created/dropped concurrently
-    * `:using` - configures the index type
-    * `:prefix` - prefix for the index
-    * `:where` - the conditions for a partial index
+    * `:name` - the name of the index. Defaults to "#{table}_#{column}_index".
+    * `:unique` - indicates whether the index should be unique. Defaults to
+      `false`.
+    * `:concurrently` - indicates whether the index should be created/dropped
+      concurrently.
+    * `:using` - configures the index type (for PostgreSQL).
```
fertapric: Is `(for PostgreSQL)` accurate? [There is a test on the MySQL side that test `using: :hash`](https://github.com/elixir-ecto/ecto/blob/master/test/ecto/adapters/mysql_test.exs#L1092-L1096), and [documentation of the MySQL seem to accept `USING`.](https://dev.mysql.com/doc/refman/5.7/en/create-index.html)

lib/ecto/migration.ex
```diff
@@ -524,32 +527,33 @@ defmodule Ecto.Migration do
 
   ## Index types
 
-  PostgreSQL supports several index types like B-tree, Hash or GiST. When
+  PostgreSQL supports several index types like B-tree, Hash, and GiST. When
```
fertapric: Should we also mention `GIN`? What about also mentioning MySQL?

lib/ecto/migration/schema_migration.ex
```diff
@@ -1,6 +1,8 @@
...
-  @moduledoc false
```
fertapric: I'm not sure if we should publicly document this module.

lib/ecto/migration.ex
```diff
@@ -796,12 +805,14 @@ defmodule Ecto.Migration do
...
-    * `:from` - provide the current type
-    * `:size` - the size of the type (for example the numbers of characters). Default is no size.
-    * `:precision` - the precision for a numeric type. Required when `scale` is specified.
-    * `:scale` - the scale of a numeric type. Default is 0 scale.
+    * `:null` - determines whether the column accepts null values.
+    * `:default` - changes the default value of the column.
+    * `:from` - specifies the current type of the column.
+    * `:size` - specifies the size of the type (for example, the number of characters).
+      The default is no size.
+    * `:precision` - the precision for a numeric type. Required when `scale` is
```
fertapric: Minor nitpick: `:scale` instead of `scale`

lib/ecto/migration.ex
fertapric: Or, what do you think about?

> ## Index types
>
> When creating an index, the index type can be specified with the `:using` option. The `:using` option can be an atom or a string; and its value is passed to the `USING` clause as is.
>
> For example, PostgreSQL supports several index types like B-tree (the default), Hash, GIN, or GiST. More information on index types can be found in the [PostgreSQL docs](http://www.postgresql.org/docs/9.4/static/indexes-types.html).

I think in that case we don't need to mention MySQL 🙂

lib/ecto/migration.ex
gj: screen shot 2018-09-30 at 2 26 53 pm

gj: I was going off of that, but I don't have strong feelings either way ¯\_(ツ)_/¯

lib/ecto/migration.ex
gj: Yep, I think linking the functions is a good idea.

lib/ecto/migration/schema_migration.ex
gj: Sure; I'll update the original comment instead of turning it into a moduledoc.

lib/ecto/migration.ex
gj: Yeah, that seems like a good compromise. Used your phrasing 🙂

lib/ecto/migration.ex
gj: Made this change; thanks for the catch!

gj:
Forgot to say it earlier, but thanks for the thorough review @fertapric ❤️

josevalim:

PR #2723, opened 2018-10-01
Correct typo in upsert example

brucepom:
It looks like in this example the result of the initial insert is bound to `updated` and then referenced on the next line of code as `inserted`. I think the intention was probably that the initial result be bound to `inserted`.


brucepom:
Sorry, please ignore. I misread the example.

PR #2724, opened 2018-10-01
Pattern match to better track which structs can be passed to which functions

gj:


lib/ecto/migration.ex
```diff
@@ -593,6 +608,16 @@ defmodule Ecto.Migration do
     %{index | name: index.name || default_index_name(index)}
   end
 
+  defp default_index_name(index) do
```
gj: Just moved this helper function up closer to the only place it's used

lib/ecto/migration.ex
```diff
@@ -828,6 +843,7 @@ defmodule Ecto.Migration do
       raise ArgumentError, "column #{Atom.to_string(column)} is missing precision option"
     end
 
+    validate_type!(type)
```
gj: The same type validations relevant to column additions would also seem to be relevant to column modifications.

lib/ecto/migration.ex
```diff
@@ -984,17 +1000,49 @@ defmodule Ecto.Migration do
   defp validate_index_opts!(opts), do: opts
 
   @doc false
-  def __prefix__(%{prefix: prefix} = index_or_table) do
```
gj: This `index_or_table` argument is what prompted the thought that we might want to pattern match to be a bit more stringent about what type of structs these functions are operating on. The test suite passes `%Constraint{}` structs into this `__prefix__/1` function

josevalim:
I believe this unfortunately removes one of the features in the original design of the code, which is to leverage structs with a shared subset of fields, so I am a :-1: on this change. Maybe it would be fine for public APIs, so it raises better error, but the private APIs should stay the same, IMO. :) Thanks!

gj:
Sounds good @josevalim — I will put a PR back up that updates `index_or_table` to be `index_or_table_or_constraint`. How do you feel about the changes in 4c27aa8 and 99b1481?

PR #2725, opened 2018-10-01
Handle relation changeset error in one place

### LINKED ISSUE #2722: Missing error metadata on invalid association
Ref: https://elixirforum.com/t/problem-with-phoenix-and-errorhelpers/16979/12

### Environment

* Ecto version (mix deps): master

### Current behavior

```elixir
  # test/ecto/repo/belongs_to_test.ex
  test "bad assoc" do
    changeset =
      %MySchema{assoc: :bad}
      |> Ecto.Changeset.change()

    {:error, changeset} = TestRepo.insert(changeset)
    assert [assoc: {"is invalid", _}] = changeset.errors
  end
```

this test fails because `changeset.errors` are `[assoc: "is invalid"]`
Worth mentioning that if we put the assoc through `put_assoc` then it works as expected:

```elixir
    changeset =
      %MySchema{}
      |> Ecto.Changeset.change()
      |> Ecto.Changeset.put_assoc(:assoc, :bad)

    {:error, changeset} = TestRepo.insert(changeset)
    assert [assoc: {"is invalid", [type: :map]}] = changeset.errors
```

### Expected behavior

errors should be: `[assoc: {"is invalid", [type: :map]}]`, same as when using `put_assoc`.

qcam:
This PR intended to fix #2722, but ended up together with some slight refactoring of `Ecto.Changeset.Relation` module.

`cast/4` and `change/3` have been changed to return `{:error, Ecto.Changeset.error()}` instead of just `:error`, making error handling easier for the caller. Error messages and metadata are also encapsulated at where the data get casted/changed, making future improvements of these information, e.g based on relation type, relatively easier.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -750,9 +750,9 @@ defmodule Ecto.Changeset do
               changeset = %{force_update(changeset, opts) | changes: changes, valid?: valid?}
               missing_relation(changeset, key, current, required?, relation, opts)
 
-            :error ->
-              meta = [validation: type, type: expected_relation_type(relation)]
-              error = {key, {message(opts, :invalid_message, "is invalid"), meta}}
+            {:error, {message, meta}} ->
+              meta = [validation: type] ++ meta
+              error = {key, {message(opts, :invalid_message, message), meta}}
               %{changeset | errors: [error | changeset.errors], valid?: false}
 
             # ignore or ok with change == original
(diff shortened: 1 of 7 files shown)


lib/ecto/changeset.ex
```diff
@@ -1105,8 +1102,8 @@ defmodule Ecto.Changeset do
         {Map.put(changes, key, change), errors, valid? and relation_valid?}
       :ignore ->
         {changes, errors, valid?}
-      :error ->
-        error = {key, {"is invalid", [type: expected_relation_type(relation)]}}
+      {:error, reason} ->
```
wojtekmach: nitpick: reason is usually an atom/string/etc, in this case it's `{message, metadata_tuple}`. It's not ideal because we're using word error a lot here, but maybe the variable should be called `error` still (which matches `t:Ecto.Changeset.error/0` type)?

btw, thanks for this PR <3

qcam: Thanks for pointing it out! It's fixed.

josevalim:

Issue #2734, opened 2018-10-04
Ecto Schema assumes that a changeset/2 function will be provided with no warning when one isn't provided.


coby-spotim:
### Environment

* Elixir version (elixir -v): 1.6.6
* Database and version: PostgreSQL 9.4
* Ecto version (mix deps): latest version
* Database adapter and version (mix deps): latest version of postgrex
* Operating system: Mac OS High Sierra

### Current behavior

When creating an Ecto Schema, there is no warning during compilation about not including a `changeset/2` function. It is assumed that every schema will have one, but none is provided by default. This is problematic when using `cast_assoc` as it expects a `changeset/2` function in your schema module.

### Expected behavior

During compilation, provide a warning for the user to see that they do not have a `changeset/2` function in their module. Optionally, inject a basic `changeset/2` function.

This behaviour is suggested as it is similar to the behaviour of GenServer with regards to the `init` function.

josevalim:
The changeset/2 function is definitely not required. We shouldn't always define it only in the off chance someone may use `cast_assoc/2`. I would rather make sure that we raise a good error message on cast_assoc/2 if changeset/2 is not there rather then define it upfront. Thanks!

josevalim:
It seems we already raise a proper error message in those cases: https://github.com/elixir-ecto/ecto/blob/9785114f545525cc57a89f3726b979619cbf3dc5/lib/ecto/changeset.ex#L772-L795

coby-spotim:
@josevalim Thanks for the quick answer!

I'd like to discuss this a bit further if you don't mind, as I'm not entirely sure if I agree with this.

While raising a good error is also important, I don't think it's explicit enough to the user. I believe there should be a warning for this in compilation, since then the user at least has some sort of knowledge of this before runtime, and it's a more explicit way of letting them know, otherwise they wouldn't know until runtime when it happens.

Defining it automatically might not be the best idea (I just got that idea from the `gen_server.ex` implementation), but at least it should be something that is apparent to the user before deploying and running code (disregarding tests as they don't always catch everything).

coby-spotim:
I apologize if I come off as rude and ignorant, I'm new here, just getting my feet wet for the first time!

josevalim:
[...]

PR #2735, opened 2018-10-04
Added a warning for when an Ecto Schema does not contain a changeset/2 function

coby-spotim:
This addresses issue #2734

I apologize if I'm missing any standards and welcome edits and comments.


josevalim:
Let's continue the discussion on #2734 but I don't think this is the way to go. Thanks! :)

PR #2736, opened 2018-10-08
remove unwanted end in docs example

fatboypunk:
Seemed like it was missed once when a fun was removed


josevalim:

PR #2737, opened 2018-10-08
extract base types to their own modules

dallincrane:
Gives each base type its own `Ecto.Type` module.

I believe the most recent discussion on doing this was in PR #2663 "Rework Ecto.Type". This largely follows the [gist](https://gist.github.com/wojtekmach/9c59e3b552109e1cfaca0b00efcdef66) suggestion given there by @wojtekmach.

Although no significant changes were made to test code, test code changes are separated into the second commit so old the tests can easily be run against all of the new lib code.

Looking forward to the discussion, changes, nitpicks, anything — hopefully this is helpful!


josevalim:
Thanks @dallincrane! The changes in PR are all excellent. However, I am still a bit on the fence if we should merge this. I think the whole `Ecto.Type` is pending a redesign and I would avoid changing things too much before we are sure of the direction we want to take.

Keeping everything in the same file also has some advantages, such as faster compilation, less remote dispatches and lower footprint, which makes me even more torn on this effort.

dallincrane:
Thanks, @josevalim. And that makes sense — there is for sure some risk in changing code like this.

What are some of the directions you see an `Ecto.Type` redesign going?
Is it possible that these changes could make a redesign easier or more accessible to contributions?

As far as compilation times and remote dispatches go, if they are a deciding factor, I'd be happy to explore profiling the impact. Are those impacts well known?

PR #2738, opened 2018-10-09
Add Ecto.Query.has_named_binding?/2

AndrewDryga:
Related to #2389.

The main use case is preloading an association if it's not already preloaded:
```
  def with_preloaded_account(queryable) do
    if has_named_binding?(queryable, :account) do
      queryable
    else
      queryable
      |> join(:left, [assignment], account in assoc(assignment, :account), as: :account)
      |> preload([assignment, account: account], account: account)
    end
  end
```

We may also add a doctest with example usage:
```

  ## Examples

        iex> import Ecto.Query
        iex> query = from(p in "posts", as: :posts, \
                join: b in "blogs", \
                join: l in "links", as: :link)
        iex> has_named_binding?(query, :posts)
        true
        iex> has_named_binding?(query, :link)
        true
        iex> has_named_binding?(query, :blogs)
        false
```

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1832,7 +1832,7 @@ defmodule Ecto.Query do
   end
 
   @doc """
-  Returns `true` if query has binding with a given name, otherwise - `false`.
+  Returns `true` if query has binding with a given name, otherwise `false`.
 
   For more information on named bindings see "Named bindings" in this module doc.
   """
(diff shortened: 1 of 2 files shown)


josevalim:

OvermindDL1:
Woooo!  This was the last bit I needed to be able to get rid of all my state passing around and go back to base Ecto Query's and base Repo calls!  Thanks much!

PR #2740, opened 2018-10-09
Fix ecto validate_length messages

allmonty:
While using Ecto `validate_length` i saw that the error message seemed wrong.
The string returned had what looked like a mistyped interpolation and the `:count` element of the error list was not counting the inputted value as expected.
ex:
ecto_error
In this example the `writable_line` was inputed with only 40 characters.

This PR fixes it.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1891,22 +1891,22 @@ defmodule Ecto.Changeset do
   end
 
   defp wrong_length(_type, value, value, _opts), do: nil
-  defp wrong_length(:string, _length, value, opts), do:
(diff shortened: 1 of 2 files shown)


josevalim:
@allmonty the current strings are correct. The interpolation look as such on purpose, to support tools like gettext. See the `Ecto.Changeset.traverse_errors` function on how to convert those error messages in pure binaries without the interpolation. Thanks!

allmonty:
@josevalim, thanks for the quick answer,
But the `:count` shouldn't be the size of the casted attribute?

josevalim:
@allmonty no, count in this case is the value of the parameter you passed to the validation. It is the expected, not the actual.

PR #2742, opened 2018-10-10
Fix type comparisons when both data and params are nil

dallincrane:
Currently all types that have a semantic `equal?` implementation don't consider `nil == nil` as equal.

Because of this, when the following conditions are met the the changeset field will register a change when it should not.

1. has a type with a semantic `equal?` comparison -or- has a composite type holding the same
2. has a nil value to begin with
3. is passed a nil value in params

Affected types
- decimal
- naive_datetime
- naive_datetime_usec
- time
- utc_datetime
- utc_datetime_usec
- {:array, [with base type above]}
- {:map, [with base type above]}


lib/ecto/type.ex
```diff
@@ -988,6 +991,14 @@ defmodule Ecto.Type do
     equal_map?(fun, Map.to_list(map1), map2)
   end
 
+  defp equal_map?(fun, [{key, nil} | tail], other_map) do
+    case Map.fetch(other_map, key) do
+      {:ok, nil} -> equal_map?(fun, tail, other_map)
```
josevalim: Pattern matching here is a bit more efficient:

```elixir
case other_map do
  %{^key => nil} -> equal_map?(fun, tail, other_map)
  _ -> false
end
```

You can also change the clause below if you want to. :)

josevalim:

PR #2747, opened 2018-10-15
Fix broken links in README

GregMefford:
I'm actually not sure whether these should have been links to `ecto_sql` like the others, but figured it's worth opening the PR to either directly fix it or point out the problem. 🚀


josevalim:

PR #2749, opened 2018-10-15
Provide beter errors for `Ecto.Type.cast`

### LINKED ISSUE #2746: Allow for beter errors for `Ecto.Type`
At the moment, you can only pass `:error` when a `cast` fails with `Ecto.Type`. 

It would be really useful to be able to pass custom error messages. For example `{:error, "some custom error message}`, to help finding out why something failed.

tcoopman:
This fixes #2746 

Checking if this is the right approach. If it is, then I'll clean it up, update the documentation,...

Feedback is very welcome, because I've got no idea if this is the way you would solve it.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -528,6 +528,11 @@ defmodule Ecto.Changeset do
         {changes, errors, valid?}
       :invalid ->
         {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
+      {:invalid, custom_errors} ->
+        {msg, custom_errors} = Keyword.pop(custom_errors, :message)
+        msg = msg || "is invalid"
+        new_errors = Keyword.merge([type: type, validation: :cast], custom_errors)
+        {changes, [{key, {msg, new_errors}} | errors], false}
     end
   end
 
@@ -569,6 +574,9 @@ defmodule Ecto.Changeset do
 
           :error ->
             :invalid
+
+          {:error, custom_errors} ->
+            {:invalid, custom_errors}
         end
 
       _ ->
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -526,13 +526,10 @@ defmodule Ecto.Changeset do
         {Map.put(changes, key, value), errors, valid?}
       :missing ->
         {changes, errors, valid?}
-      :invalid ->
-        {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
       {:invalid, custom_errors} ->
-        {msg, custom_errors} = Keyword.pop(custom_errors, :message)
-        msg = msg || "is invalid"
-        new_errors = Keyword.merge([type: type, validation: :cast], custom_errors)
-        {changes, [{key, {msg, new_errors}} | errors], false}
+        {message, custom_errors} = Keyword.pop(custom_errors, :message, "is invalid")
+        new_errors = [type: type] ++ Keyword.put_new(custom_errors, :validation, :cast)
+        {changes, [{key, {message, new_errors}} | errors], false}
     end
   end
 
@@ -573,7 +570,7 @@ defmodule Ecto.Changeset do
             end
 
           :error ->
-            :invalid
+            {:invalid, message: "is invalid"}
 
           {:error, custom_errors} ->
             {:invalid, custom_errors}
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -570,9 +570,9 @@ defmodule Ecto.Changeset do
             end
 
           :error ->
-            {:invalid, message: "is invalid"}
+            {:invalid, []}
 
-          {:error, custom_errors} ->
+          {:error, custom_errors} when is_list(custom_errors) ->
             {:invalid, custom_errors}
         end
 
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -123,6 +123,15 @@ defmodule Ecto.Type do
     1. When casting values by `Ecto.Changeset`
     2. When passing arguments to `Ecto.Query`
 
+
+  When returning `{:error, keyword()}`, the returned keyword list
+  will be preserved in the changeset errors, similar to
+  `Changeset.add_error/4`. Passing a `:message` key, will override
+  the default message.
+  It is not possible to override the `:type` key.
+
+  For `{:array, CustomType}` or `{:map, CustomType}` the returned
+  keyword list will be erased and the default error will be shown.
   """
   @callback cast(term) :: {:ok, term} | {:error, keyword()} | :error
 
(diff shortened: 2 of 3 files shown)


lib/ecto/changeset.ex
```diff
@@ -528,6 +528,11 @@ defmodule Ecto.Changeset do
         {changes, errors, valid?}
       :invalid ->
         {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
+      {:invalid, custom_errors} ->
+        {msg, custom_errors} = Keyword.pop(custom_errors, :message)
+        msg = msg || "is invalid"
+        new_errors = Keyword.merge([type: type, validation: :cast], custom_errors)
```
fertapric: Should Ecto raise an error if the custom error contains reserved keys like `:type` and `:validation`?

lib/ecto/changeset.ex
```diff
@@ -569,6 +574,9 @@ defmodule Ecto.Changeset do
 
           :error ->
             :invalid
```
josevalim: Let's change this to return `{:invalid, message: "is invalid"}`.

lib/ecto/changeset.ex
```diff
@@ -528,6 +528,11 @@ defmodule Ecto.Changeset do
         {changes, errors, valid?}
       :invalid ->
         {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
+      {:invalid, custom_errors} ->
+        {msg, custom_errors} = Keyword.pop(custom_errors, :message)
+        msg = msg || "is invalid"
```
josevalim: Should we raise if the message is not present? Thoughts @wojtekmach?

lib/ecto/changeset.ex
josevalim: I think we want to allow the user to change `validation` but I think is no reason to allow the type to be changed. I would go with something like:

```elixir
[type: type] ++ Keyword.put_new(custom_errors, :validation, :cast)
```

josevalim:
Thanks @tcoopman! Note that we also need a test for what happens when you declare the type to be `{:array, CustomError}` and `{:map, CustomError}`. I think you will have to do some changes in `Ecto.Type` for this to work. I think we will have to drop the information that you return today and make the whole array/map as invalid (without information of its inner contents).

lib/ecto/changeset.ex
fertapric: I would make it optional to reduce boilerplate. 

For example, `{:error, reason: :invalid_country}` would need to rewritten to `{:error, reason: :invalid_country, message: "is invalid"}` even if the user does not want to provide a custom error message.

wojtekmach:
@josevalim do we need to make similar changes to `load` and `dump` callbacks? I don't think we do as we don't use them for errors, but just double-checking.

josevalim:
@wojtekmach i don't think so, yeah.

zachdaniel:
We should probably change the behavior as well to indicate this new return format: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L648

lib/ecto/changeset.ex
tcoopman: I've changed it to the remarks of @josevalim. If we want to inform the user that we are overriding the `type`, that's something for the docs?

tcoopman:
> Thanks @tcoopman! Note that we also need a test for what happens when you declare the type to be `{:array, CustomError}` and `{:map, CustomError}` and you pass a list/map with an invalid entry. I think you will have to do some changes in `Ecto.Type` for this to work. I think we will have to drop the information that you return today and make the whole array/map as invalid (without information of its inner contents).

So you mean a test like this:

```
 defmodule Ecto.ChangesetTest do
     schema "custom_error" do
       field :custom_error, CustomError
       field :custom_error_2, CustomError2
+      field :array_custom_error, {:array, CustomError}
     end
   end
 
@@ -269,6 +270,15 @@ defmodule Ecto.ChangesetTest do
     refute changeset.valid?
   end
 
+  test "cast/4: field has a custom invalid error message on an array field" do
+    params = %{"array_custom_error" => [:error]}
+    struct = %CustomErrorTest{}
+
+    changeset = cast(struct, params, ~w(array_custom_error)a)
+    assert changeset.errors == [array_custom_error: {"is invalid", [type: Ecto.ChangesetTest.CustomError, validation: :cast, reason: :foobar]}]
+    refute changeset.valid?
+  end
```

Should the error be changed to note that it is an error in an array?

josevalim:
@tcoopman I don't think we should propagate the internal error, because it can get very confusing. So I would suggest this:

```
assert changeset.errors == [array_custom_error: {"is invalid", [type: {:array, Ecto.ChangesetTest.CustomError}, validation: :cast]}]
```

lib/ecto/changeset.ex
```diff
@@ -568,7 +570,10 @@ defmodule Ecto.Changeset do
             end
 
           :error ->
-            :invalid
+            {:invalid, message: "is invalid"}
```
josevalim: Since we do have a default message, I think we can go with:

```suggestion
            {:invalid, []}
```

lib/ecto/changeset.ex
```diff
@@ -568,7 +570,10 @@ defmodule Ecto.Changeset do
...
+
+          {:error, custom_errors} ->
```
josevalim: I would add a guard, just in case:

```suggestion
          {:error, custom_errors} when is_list(custom_errors) ->
```

lib/ecto/changeset.ex
```diff
@@ -526,8 +526,10 @@ defmodule Ecto.Changeset do
         {Map.put(changes, key, value), errors, valid?}
       :missing ->
         {changes, errors, valid?}
-      :invalid ->
-        {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
+      {:invalid, custom_errors} ->
+        {message, custom_errors} = Keyword.pop(custom_errors, :message, "is invalid")
+        new_errors = [type: type] ++ Keyword.put_new(custom_errors, :validation, :cast)
+        {changes, [{key, {message, new_errors}} | errors], false}
```
josevalim: This looks beautiful now, thank you!

tcoopman:
> ```
> assert changeset.errors == [array_custom_error: {"is invalid", [type: {:array, Ecto.ChangesetTest.CustomError}, validation: :cast]}]
> ```

I've changed this and erased the custom error in `Ecto.Type`.
I've also started to add some documentation. Not sure if the style is correct though.

josevalim:

lib/ecto/changeset.ex
```diff
@@ -526,8 +526,10 @@ defmodule Ecto.Changeset do
         {Map.put(changes, key, value), errors, valid?}
       :missing ->
         {changes, errors, valid?}
-      :invalid ->
-        {changes, [{key, {"is invalid", [type: type, validation: :cast]}} | errors], false}
+      {:invalid, custom_errors} ->
+        {message, custom_errors} = Keyword.pop(custom_errors, :message, "is invalid")
+        new_errors = [type: type] ++ Keyword.put_new(custom_errors, :validation, :cast)
```
fertapric: If the user provides the `:type` key, the resulting error will contain two `:type` keys, which might lead to some side effects similar to overriding the `:type` key. For example:

```elixir
iex(1)> errors = [type: MyType] ++ [type: MyErrorType, validation: :my_validation]
[type: MyType, type: MyErrorType, validation: :my_validation]
iex(2)> Enum.into(errors, %{})
%{type: MyErrorType, validation: :my_validation}
```

PR #2752, opened 2018-10-17
Frame clause for windows

Anber:
The last part of WINDOW clause. (and ecto_sql [PR](https://github.com/elixir-ecto/ecto_sql/pull/20))

### DIFF
--- a/integration_test/cases/windows.exs
+++ b/integration_test/cases/windows.exs
@@ -4,8 +4,7 @@ defmodule Ecto.Integration.WindowsTest do
   alias Ecto.Integration.TestRepo
   import Ecto.Query
 
-  alias Ecto.Integration.Comment
-  alias Ecto.Integration.User
+  alias Ecto.Integration.{Comment, User, Post}
 
   test "over" do
     u1 = TestRepo.insert!(%User{name: "Tester"})
@@ -26,5 +25,29 @@ defmodule Ecto.Integration.WindowsTest do
     # Over window
     query = from(c in Comment, windows: [w: [partition_by: c.author_id]], select: [c, count(c.id) |> over(:w)])
     assert [[^c1, 3], [^c2, 3], [^c3, 3], [^c4, 1]] = TestRepo.all(query)
+  end
+
+  test "frame" do
+    posts = Enum.map(0..6, &%{counter: &1, visits: round(:math.pow(2, &1))})
+    TestRepo.insert_all(Post, posts)
+
+    n = 1
+    query = from(p in Post,
+      windows: [w: [order_by: p.counter, frame: fragment("ROWS BETWEEN ? PRECEDING AND ? FOLLOWING", ^n, ^n)]],
+      select: [p.counter, sum(p.visits) |> over(:w)]
+    )
+    assert [[0, 3], [1, 7], [2, 14], [3, 28], [4, 56], [5, 112], [6, 96]] = TestRepo.all(query)
+
+    query = from(p in Post,
+      windows: [w: [order_by: p.counter, frame: fragment("ROWS BETWEEN 1 FOLLOWING AND UNBOUNDED FOLLOWING")]],
+      select: [p.counter, sum(p.visits) |> over(:w)]
+    )
+    assert [[0, 126], [1, 124], [2, 120], [3, 112], [4, 96], [5, 64], [6, nil]] = TestRepo.all(query)
+
+    query = from(p in Post,
+      windows: [w: [order_by: p.counter, frame: fragment("ROWS CURRENT ROW")]],
+      select: [p.counter, sum(p.visits) |> over(:w)]
+    )
+    assert [[0, 1], [1, 2], [2, 4], [3, 8], [4, 16], [5, 32], [6, 64]] = TestRepo.all(query)
   end
 end
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -504,6 +504,18 @@ defmodule Ecto.Query do
       windows: [ordered_names: [order_by: e.name]]
 
   It works exactly as the keyword query version of `order_by/3`.
+
+  ### :frame
+
+  A fragment which defines the frame for window functions.
+
+  ## Examples
+
+      # compare each employee's salary for each month with his average salary for previous 3 months
+      from p in Payroll,
+        select: {p.empno, p.date, p.salary, over(avg(p.salary), :prev_months)},
+        windows: [prev_months: [partition_by: p.empno, order_by: p.date, frame: fragment("ROWS 3 PRECEDING EXCLUDE CURRENT ROW")]]
+
   """
   defmacro windows(query, binding \\ [], expr) do
     Windows.build(query, binding, expr, __CALLER__)
--- a/lib/ecto/query/builder/group_by.ex
+++ b/lib/ecto/query/builder/group_by.ex
@@ -15,7 +15,7 @@ defmodule Ecto.Query.Builder.GroupBy do
         13],
        {%{}, :acc}}
   """
-  @spec escape(:group_by | :partition_by, {map, term}, Macro.t, Keyword.t, Macro.Env.t) ::
+  @spec escape(:group_by | :partition_by, Macro.t, {map, term}, Keyword.t, Macro.Env.t) ::
           {Macro.t, {map, term}}
   def escape(kind, {:^, _, [expr]}, params_acc, _vars, _env) do
     {quote(do: Ecto.Query.Builder.GroupBy.group_by!(unquote(kind), unquote(expr))), params_acc}
--- a/lib/ecto/query/builder/windows.ex
+++ b/lib/ecto/query/builder/windows.ex
@@ -34,14 +34,26 @@ defmodule Ecto.Query.Builder.Windows do
     {{:order_by, fields}, params_acc}
   end
 
+  defp do_escape({:frame, frame_clause}, params_acc, vars, env) do
+    {frame_clause, params_acc} = escape_frame(frame_clause, params_acc, vars, env)
+    {{:frame, frame_clause}, params_acc}
+  end
+
   defp do_escape(other, _params_acc, _vars, _env) do
     error!(other)
+  end
+
+  defp escape_frame({:fragment, _, _} = fragment, params_acc, vars, env) do
+    Builder.escape(fragment, :any, params_acc, vars, env)
+  end
+  defp escape_frame(other, _, _, _) do
+    Builder.error!("expected a fragment in `:frame`, got: `#{inspect other}`")
   end
 
   defp error!(other) do
     Builder.error!(
       "expected window definition to be a keyword list " <>
-        "with partition_by and order_by as keys, got: `#{inspect other}`"
+        "with partition_by, order_by or frame as keys, got: `#{inspect other}`"
     )
   end
 
--- a/lib/ecto/query/window_api.ex
+++ b/lib/ecto/query/window_api.ex
@@ -73,7 +73,7 @@ defmodule Ecto.Query.WindowAPI do
   Returns number of the current row within its partition, counting from 1.
 
       from p in Post,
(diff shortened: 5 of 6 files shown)


Anber:
Frame clause (rows or range) can be specified as tuple (`{start, end}` where `start` and `end` are integers or `nil`), integer (which is equivalent to `{start, nil}`) or range (`start..end` where `start` and `end` are integers).

```elixir
from(p in Post,
  select: [p.counter, sum(p.visits) |> over(order_by: p.counter, rows: {-1, 1})]
)
```

josevalim:
Hi @Anber! After reading the specification, I have two different approaches:

1. Let's support just tuples. The specification would be `{non_neg_integer | :unbound | :current, non_neg_integer | :unbound | :current}`. I think `:unbound` is clearer than nil and `:current` is better than 0 to indicate current. All DBs say the integer are non negative (they include 0, so we don't need to handle `-`).

2. We support only a string. So the user wold do `rows: "BETWEEN UNBOUND PRECEDING AND CURRENT ROW"`. This is the simplest approach and it may also be clearer to the user, who won't have to translate between Ecto tuple syntax and the SQL query.

Thoughts? Which one do you prefer?

zachdaniel:
They don't necessarily preclude each other. We could implement the string variation of it, and then add more specific syntax if desired.

Anber:
Hi @josevalim and @zachdaniel!

I don’t know why SQL has such a terrible syntax for such a simple thing but they definitely have some historical reasons :)

Strings look simplier but we need to parse them if we want to use compile-time validation and I think we have to use validation here because it’s a static expression and we don’t have excuses to not do that :). Also, this syntax is a bit monstrous.

Why ranges? I think they are more clear: if I want to describe frame which should include two previous rows, current row, and two next rows, I use existing language construction (`-2..2`) instead of raw SQL expression (`"BETWEEN 2 PRECEDING AND 2 FOLLOWING"`) which even cannot be highlighted in IDE.

In tuples, we need negative numbers for showing the difference between PRECEDING and FOLLOWING. And here 0 becomes logical but we can add `:current` as a synonym.

I also don’t like nil here and initially, it was `:unbound`. nil has more sense in terms of limits of rows interval, whereas `:unbound` looks better with `:range` and `:current`. So we can either add a synonym here.

josevalim:
> Strings look simplier but we need to parse them if we want to use compile-time validation

We don't have to. In cases like this, such as locks, we just send the query as is to the database. Plus, at compile time we don't know the adapter, and some DBs may introduce custom behaviour here, which we won't be able to validate.

> In tuples, we need negative numbers for showing the difference between PRECEDING and FOLLOWING. And here 0 becomes logical but we can add :current as a synonym.

We don't really need negative numbers for tuples. The preceding can only be on the left side and the following on the right one. The issue with ranges is that it only supports integers, which ends up imposing a variety of forms, so we need to support both ranges and tuples which to me is a :-1:.

To me, the most important approach for now is to define a single format that supports all cases. We can always add "sugar" in the future, but starting small is more important, especially because this feature won't be used frequently.

If we really want a range, then the only idea I have is to use `_` for unbound. So `_` means unbound, `0` means current. Negative is preceding. Positive is following. Thoughts?

Anber:
> The preceding can only be on the left side and the following on the right one.
Actually, no. You can take frame from 10 preceding to 1 preceding which can be written as -10..-1.

Anber:
> some DBs may introduce custom behaviour here

And that is another point against raw sql. In different databases it can be implemented in different ways.

josevalim:
> Actually, no. You can take frame from 10 preceding to 1 preceding which can be written as -10..-1.

Ah, good call, only the unbound ones are restrict to the left or right.

> And that is another point against raw sql. In different databases it can be implemented in different ways.

Not really. We will likely never be able to support all use cases across all DBs in Ecto's query syntax, that's why we provide `fragment` and so on. Plus, since this particular area the DBs actually follow the standard, going with a string is actually very convenient.

josevalim:
So I think we have three proposals:

1. Tuples - `{left, right}`, where `left` and `right` are either numbers or the atom `:unbound`
2. Ranges - `left..right`, where `left` and `right` are either numbers or `_` (meaning unbound)
3. Strings

We can only pick one. :)

Anber:
I can’t check it right now but ‘_..10’ doesn’t look as a valid expression. Does it?

For me, tuples look like the most flexible solution. On the other hand, ranges better describe what we really do.

josevalim:
> I can’t check it right now but ‘_..10’ doesn’t look as a valid expression. Does it?

It is a valid AST, we will just have to handle it. :)

Anber:
`_..10` looks weird for me…
So I vote for the tuples.

josevalim:
Thanks @Anber! @wojtekmach @zachdaniel what are your thoughts?

Anber:
I've dropped support of ranges and integers.

PostgreSQL 11 was released today and it has two new features which may affect this PR: `GROUPS` mode and `frame_exclusion` option. What will we do with that?

josevalim:
@Anber link to the new features please? :)

Anber:
@josevalim oops… I thought I’ve attached the [link](https://www.postgresql.org/docs/11/static/sql-select.html). We can just add the third frame type as a new keyword `:groups` in addition to `:rows` and `:range` but I’m starting thinking that maybe it would be better to use something like `[partition_by: …, frame: [:rows, {-10, 10}, exclude: :current]]`

josevalim:
The exclusion makes me think that just supporting a string, as I mentioned earlier, is the easiest way to go about this. The implementation is small, the translation to SQL is straight-forward and it will literally support all cases. Folks are not using this feature frequently, so being verbose here is not a major problem.

zachdaniel:
I also think that there is no reason not to start with a string :) Might as well get the minimum version of the feature out and see how others use it.

Anber:
But we have to validate strings at least in order to prevent SQL injections. Without validation it would work like unsafe_fragment.

Anber:
@josevalim I have just realized that it is required to be constant in SQL query but it should be dynamic in Ecto. Otherwise, we cannot change frame settings in runtime and it makes whole this feature a bit useless in real-world apps.

josevalim:
We can still change it if we support fragments as well as strings.

-- 


josevalim:
Or to better move the discussion forward, can you provide concrete examples
of what you want to achieve and why strings do not make it possible? Thanks
:)
-- 


Anber:
Hm… I hope @hauleth can give us some real-life examples.

@josevalim how we can implement something like this?
```elixir
  def avg_for(months) do
    from p in Payroll,
         select: {p.empno, p.date, p.salary, over(avg(p.salary), :prev_months)},
         windows: [prev_months: [partition_by: p.empno, order_by: p.date, rows: {-^months, -1}]]
  end
```

hauleth:
@Anber what kind of real life examples would you like? But yeah, the best example of situation where dynamic bound would be desired is when we want to have rolling average or other statistic. However I think that for now string would be ok as @josevalim said, as we can add support for tuples later when we establish better interface we can add support for tuples/whatever.

However if we would implement it as @Anber proposes we should mark that `rows` will always mean physical rows, `range` mean logical rows and `groups` mean groups tied by the grouping window. This is quite hard to understand without proper example and as it is quite late here I will just post [jOOQ blog post](https://blog.jooq.org/2018/07/05/postgresql-11s-support-for-sql-standard-groups-and-exclude-window-function-clauses/) about the feature.

As jOOQ and Ecto have similar scope and even similar API (thin abstraction over query language rather than ActiveRecord or similar ORM) it would be highly encouraged for Ecto devs to follow their blog as they post a lot of valuable informations about SQL.

So in summary:

1. Go with the string for now and in some version Ecto 3.x (where x > 0) we can add tuple syntax.
2. Go with the tuples but provide 3 different options that are exclusive: `rows`, `range`, and `groups` where the last one would be supported only by PostgreSQL.

In general `range` and `groups` are almost identical with exception to gaps, `range` will ignore missing windows and `groups` will skip them and fetch next existing group.

josevalim:
> @josevalim how we can implement something like this?

We could totally use fragments:

    rows: fragment("BETWEEN ? PRECEDING AND 1 PRECEDING", ^months)

Since Postgres supports a parameter there (see example blow), it should just work™. So my suggestion is to support only strings and fragments (a fragment would be expected anyway).

```psql
demo_dev=# PREPARE foo(integer) AS SELECT
  payment_date,
  amount,
  avg(amount) OVER (
    ORDER BY payment_date, payment_id
    ROWS BETWEEN $1 PRECEDING AND 2 FOLLOWING
  )::DECIMAL(10, 2),
  array_agg(amount) OVER (
    ORDER BY payment_date, payment_id
    ROWS BETWEEN 2 PRECEDING AND 2 FOLLOWING
  )
FROM payment;
PREPARE
demo_dev=# EXECUTE foo(1);
 payment_date | amount | avg | array_agg
--------------+--------+-----+-----------
(0 rows)
```

@hauleth that's a very good reference to have in mind. Thank you!

hauleth:
As I said, we can add tuple/range support later without problem. However we still would need to support at least `rows` and `ranges` keywords.

Anber:
@josevalim but with fragments it’s impossible to pass UNBOUNDED instead of a number or change PRECEDING to FOLLOWING. Otherwise it will be possible to inject any SQL.

josevalim:
@Anber you can still do that conditionally:

```elixir
if argument == :unbound do
  from p in Query, windows: [foo: [..., rows: "... UNBOUNDED ..."]
else
  from p in Query, windows: [foo: [..., rows: "... CURRENT ROW ..."]
end
```

If there is a lot of duplication it can be solved with macros too. It is not perfect, sure, but it is definitely enough to get started and cover a wide range of use cases.

Anber:
@josevalim I don’t understand why we can’t cover all cases by tuples? :)
Instead of a simple solution, we allow people inserting whatever they want in the middle of the query.

We've already decided that tuples are ok for simple frames but you don’t want to use it with EXCLUDE. Why we cannot just add :exclude as a 4th keyword argument to the window clause?

josevalim:
Because it is simpler implementation wise, it will always support all use cases DBs come up in the future, and users do not need to learn how to translate between SQL and tuple for a task that is so specific.

However, given adding a string now may mean supporting both strings and tuples in the future, I would like to propose for us to start only with fragments. We need to support fragments anyway, regardless of tuples and strings, because almost everywhere in Ecto supports fragments, so let's go ahead with it. This will allow us to at least have this feature and we can decide on strings/tuples/ranges later.

Can you please change the PR to allow only fragments? Thanks :heart:!

Anber:
@josevalim only fragments as a start sounds ok.
The last question:
`[…, frame: fragment(“ROWS BETWEEN …”)]`
or
`[…, rows: fragment(“BETWEEN …”)]`

josevalim:
@Anber I like the `:frame` approach, especially because it can allow us to do: `[rows: {-1, 1}, exclude: ...]` in the future. So let's go with frame+fragment right now, thanks! :+1:

Anber:
Done!

lib/ecto/query/builder/frame_clause.ex
```diff
@@ -0,0 +1,24 @@
+import Kernel, except: [apply: 2]
+
+defmodule Ecto.Query.Builder.FrameClause do
```
josevalim: I don't think we need a separate module for this, we can check it directly in the `Ecto.Query.Builder.Windows`. We can also merge the test with the windows test.

josevalim: We should also add some tests that make sure that it works both on over and on window, probably inside `windows_test`. Before we would add those tests to the PG/MySQL unit tests, but now that they live in another repository, it means we need better coverage here. :)

josevalim:
Thanks @Anber! I have added a comment, which is mostly about improving tests and moving the code around. Then I think we are ready to go!

Anber:
@josevalim, done.

I've also fixed documentation for WindowAPI.

josevalim:

PR #2756, opened 2018-10-19
Add specific error when interval is not a string. Fix docs.

TheFirstAvenger:
The error message when passing an atom into intervals is somewhat confusing if you don't immediately realize the valid intervals it is listing in the error message are strings, not atoms. For example, we spent too much time trying to figure out what was going on with this: `(Ecto.Query.CompileError) invalid interval: `:month` (expected one of year, month, week, day, hour, minute, second, millisecond, microsecond)`
Given that the most common mistake here is probably passing a "valid" interval in atom form instead of a string, this PR updates the error to catch a non-string and explicitly call out the fact that it needs to be a string. Still notes the valid strings when an invalid string is passed.

Docs also seemed to be out of date, updated them.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -811,13 +811,16 @@ defmodule Ecto.Query.Builder do
   end
 
   @doc """
-  Called by escaper at runtime to verify that value is an atom.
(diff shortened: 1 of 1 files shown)


josevalim:

PR #2757, opened 2018-10-20
Add Common Table Expressions (a.k.a. CTE or WITH)

feymartynov:
I know it's too late for 3.0-rc but anyway :)

This enables Common Table Expressions. There are two common use cases for them:
1. Recursive queries for retrieving tree structures (e.g. category tree or virtual file tree).
2. Reusable subqueries. Useful for scoping.

API may be disputable because it's more evident to use `with` as in plain SQL instead of `cte` but it may be a mess with `Kernel.SpecialForms.with/1`. I'm open for proposals.

Complementary PR in ecto_sql: https://github.com/elixir-ecto/ecto_sql/pull/23

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -356,7 +356,8 @@ defmodule Ecto.Query do
 
   defstruct [prefix: nil, sources: nil, from: nil, joins: [], aliases: %{}, wheres: [], select: nil,
              order_bys: [], limit: nil, offset: nil, group_bys: [], combinations: [], updates: [],
-             havings: [], preloads: [], assocs: [], distinct: nil, lock: nil, windows: []]
+             havings: [], preloads: [], assocs: [], distinct: nil, lock: nil, windows: [],
+             with_ctes: nil]
 
   defmodule FromExpr do
     @moduledoc false
@@ -388,6 +389,11 @@ defmodule Ecto.Query do
     defstruct [:qual, :source, :on, :file, :line, :assoc, :as, :ix, :prefix, params: [], hints: []]
   end
 
+  defmodule WithExpr do
+    @moduledoc false
+    defstruct [recursive: false, queries: []]
+  end
+
   defmodule Tagged do
     @moduledoc false
     # * value is the tagged value
@@ -401,7 +407,7 @@ defmodule Ecto.Query do
 
   alias Ecto.Query.Builder
   alias Ecto.Query.Builder.{Distinct, Dynamic, Filter, From, GroupBy, Join, Windows,
-                            LimitOffset, Lock, OrderBy, Preload, Select, Update}
+                            LimitOffset, Lock, OrderBy, Preload, Select, Update, CTE}
 
   @doc """
   Builds a dynamic query expression.
@@ -606,6 +612,7 @@ defmodule Ecto.Query do
       Ecto.Query.exclude(query, :distinct)
       Ecto.Query.exclude(query, :select)
       Ecto.Query.exclude(query, :combinations)
+      Ecto.Query.exclude(query, :with_ctes)
       Ecto.Query.exclude(query, :limit)
       Ecto.Query.exclude(query, :offset)
       Ecto.Query.exclude(query, :lock)
@@ -638,6 +645,7 @@ defmodule Ecto.Query do
   defp do_exclude(%Ecto.Query{} = query, :order_by), do: %{query | order_bys: []}
   defp do_exclude(%Ecto.Query{} = query, :group_by), do: %{query | group_bys: []}
   defp do_exclude(%Ecto.Query{} = query, :combinations), do: %{query | combinations: []}
+  defp do_exclude(%Ecto.Query{} = query, :with_ctes), do: %{query | with_ctes: nil}
   defp do_exclude(%Ecto.Query{} = query, :having), do: %{query | havings: []}
   defp do_exclude(%Ecto.Query{} = query, :distinct), do: %{query | distinct: nil}
   defp do_exclude(%Ecto.Query{} = query, :select), do: %{query | select: nil}
@@ -714,7 +722,7 @@ defmodule Ecto.Query do
   @from_join_opts [:as, :prefix, :hints]
   @no_binds [:lock, :union, :union_all, :except, :except_all, :intersect, :intersect_all]
   @binds [:where, :or_where, :select, :distinct, :order_by, :group_by, :windows] ++
-           [:having, :or_having, :limit, :offset, :preload, :update, :select_merge]
+           [:having, :or_having, :limit, :offset, :preload, :update, :select_merge, :with_ctes]
 
   defp from([{type, expr}|t], env, count_bind, quoted, binds) when type in @binds do
     # If all bindings are integer indexes keep AST Macro expandable to %Query{},
@@ -957,6 +965,98 @@ defmodule Ecto.Query do
     query
     |> Join.build(qual, binding, expr, nil, on, as, prefix, hints, __CALLER__)
     |> elem(0)
+  end
+
+  @doc """
+  A common table expression (CTE) also known as WITH expression.
+
+  `name` must be a compile-time literal string that is being used as the table name to join
+  the CTE in the main query or in the recursive CTE.
+
+  ## Options
+
+    * `:as` – the CTE query itself or a fragment
+
+  ## Recursive CTEs
+
+  Use `recursive_ctes/2` to enable recursive mode for CTEs.
+
+  In the CTE query itself use the same table name to leverage recursion that has been passed
+  to `name` argument. Make sure to write a stop condition to avoid infinite recursion loop.
+
+  ## Non-recursive CTEs
+
+  By default recursive mode for CTEs is off.
+
+  Keep in mind that from the performance perspective usually queries with non-recursive CTEs
+  can be rewritten to more optimal ones that leverage joins or subqueries instead.
+
+  ## Expression examples
+
+  Products and their category names for breadcrumbs:
+
+      category_tree_initial_query =
+        Category
+        |> where([c], is_nil(c.parent_id))
+
+      category_tree_recursion_query =
+        Category
+        |> join(:inner, [c], ct in "category_tree", on: c.parent_id == ct.id)
+
+      category_tree_query =
+        category_tree_initial_query
+        |> union_all(^category_tree_recursion_query)
+
+      Product
+      |> recursive_ctes(true)
+      |> with_cte("category_tree", as: ^category_tree_query)
+      |> join(:left, [p], c in "category_tree", on: c.id == p.category_id)
+      |> group_by([p], p.id)
+      |> select([p, c], %{p | category_names: fragment("ARRAY_AGG(?)", c.name)})
+
+  It's possible to cast CTE result rows as Ecto structs:
+
+      {"category_tree", Category}
+      |> recursive_ctes(true)
+      |> with_cte("category_tree", as: ^category_tree_query)
+      |> join(:left, [c], p in assoc(c, :products))
+      |> group_by([c], c.id)
+      |> select:([c, p], %{c | products_count: count(p.id)})
+
+  It's also possible to pass a raw SQL fragment:
+
+      @raw_sql_category_tree \"""
+      SELECT * FROM categories WHERE c.parent_id IS NULL
+      UNION ALL
+      SELECT * FROM categories AS c, category_tree AS ct WHERE ct.id = c.parent_id
+      \"""
+
+      Product
+      |> recursive_ctes(true)
+      |> with_cte("category_tree", as: fragment(@raw_sql_category_tree))
+      |> join(:inner, [p], c in "category_tree", on: c.id == p.category_id)
+
+  Keyword syntax is not supported for this feature.
+  """
+  defmacro with_cte(query, name, as: with_query) do
+    CTE.build(query, name, with_query, __CALLER__)
+  end
+
+  @doc """
+  Enables or disables recursive mode for CTEs.
+
+  According to the SQL standard it affects all CTEs in the query, not individual ones.
+
+  See `with_cte/3` on example of how to build a query with a recursive CTE.
+  """
+  def recursive_ctes(%__MODULE__{with_ctes: with_expr} = query, value) when is_boolean(value) do
+    with_expr = with_expr || %WithExpr{}
+    with_expr = %{with_expr | recursive: value}
+    %{query | with_ctes: with_expr}
+  end
+
+  def recursive_ctes(queryable, value) do
+    recursive_ctes(Ecto.Queryable.to_query(queryable), value)
   end
 
   @doc """
--- a/lib/ecto/query/inspect.ex
+++ b/lib/ecto/query/inspect.ex
@@ -1,9 +1,10 @@
 import Inspect.Algebra
 import Kernel, except: [to_string: 1]
 
-alias Ecto.Query.{DynamicExpr, JoinExpr, QueryExpr}
+alias Ecto.Query.{DynamicExpr, JoinExpr, QueryExpr, WithExpr}
 
-container_doc = if Version.match?(System.version(), ">= 1.6.0"), do: :container_doc, else: :surround_many
+container_doc =
+  if Version.match?(System.version(), ">= 1.6.0"), do: :container_doc, else: :surround_many
 
 defimpl Inspect, for: Ecto.Query.DynamicExpr do
   def inspect(%DynamicExpr{binding: binding} = dynamic, opts) do
@@ -12,26 +13,45 @@ defimpl Inspect, for: Ecto.Query.DynamicExpr do
 
     names = for {name, _, _} <- binding, do: Atom.to_string(name)
 
-    inspected =
-      Inspect.Ecto.Query.expr(expr, List.to_tuple(names), %{expr: expr, params: params})
+    inspected = Inspect.Ecto.Query.expr(expr, List.to_tuple(names), %{expr: expr, params: params})
 
     # TODO: Replace unquote with just `container_doc` when Elixir < 1.6.0 is no longer supported
-    unquote(container_doc)("dynamic(", [Macro.to_string(binding), inspected], ")", opts, fn str, _ -> str end)
+    unquote(container_doc)("dynamic(", [Macro.to_string(binding), inspected], ")", opts, fn str,
+                                                                                            _ ->
+      str
+    end)
   end
 end
 
 defimpl Inspect, for: Ecto.Query do
   @doc false
   def inspect(query, opts) do
-    list = Enum.map(to_list(query), fn
-      {key, string} ->
-        concat(Atom.to_string(key) <> ": ", string)
-      string ->
-        string
-    end)
+    list =
+      Enum.map(to_list(query), fn
+        {key, string} ->
+          concat(Atom.to_string(key) <> ": ", string)
+
+        string ->
+          string
+      end)
 
     # TODO: Replace unquote with just `container_doc` when Elixir < 1.6.0 is no longer supported
-    unquote(container_doc)("#Ecto.Query<", list, ">", opts, fn str, _ -> str end)
+    result = unquote(container_doc)("#Ecto.Query<", list, ">", opts, fn str, _ -> str end)
+
+    case query.with_ctes do
+      %WithExpr{recursive: recursive, queries: [_ | _] = queries} ->
+        with_ctes =
+          Enum.map(queries, fn {name, query} ->
+            cte = __MODULE__.inspect(query, opts)
+            concat(["|> with_cte(\"" <> name <> "\", as: ", cte, ")"])
+          end)
+
+        recursive_ctes = if recursive, do: "|> recursive_ctes(true)", else: ""
+        ([result, recursive_ctes] ++ with_ctes) |> Enum.intersperse(break("\n")) |> concat()
+
+      _ ->
+        result
+    end
   end
 
   @doc false
@@ -39,6 +59,7 @@ defimpl Inspect, for: Ecto.Query do
     Enum.map_join(to_list(query), ",\n  ", fn
       {key, string} ->
         Atom.to_string(key) <> ": " <> string
+
       string ->
         string
     end)
@@ -50,7 +71,7 @@ defimpl Inspect, for: Ecto.Query do
       |> collect_sources
       |> generate_letters
       |> generate_names
-      |> List.to_tuple
+      |> List.to_tuple()
 
     from = bound_from(query.from, binding(names, 0))
     joins = joins(query.joins, names)
@@ -58,73 +79,96 @@ defimpl Inspect, for: Ecto.Query do
     assocs = assocs(query.assocs, names)
     windows = windows(query.windows, names)
     combinations = combinations(query.combinations)
- 
+
     wheres = bool_exprs(%{and: :where, or: :or_where}, query.wheres, names)
     group_bys = kw_exprs(:group_by, query.group_bys, names)
     havings = bool_exprs(%{and: :having, or: :or_having}, query.havings, names)
     order_bys = kw_exprs(:order_by, query.order_bys, names)
     updates = kw_exprs(:update, query.updates, names)
- 
+
     lock = kw_inspect(:lock, query.lock)
     limit = kw_expr(:limit, query.limit, names)
     offset = kw_expr(:offset, query.offset, names)
     select = kw_expr(:select, query.select, names)
     distinct = kw_expr(:distinct, query.distinct, names)
 
-    Enum.concat [from, joins, wheres, group_bys, havings, windows, combinations, order_bys,
-                 limit, offset, lock, distinct, updates, select, preloads, assocs]
+    Enum.concat([
+      from,
+      joins,
+      wheres,
+      group_bys,
+      havings,
+      windows,
+      combinations,
+      order_bys,
+      limit,
+      offset,
+      lock,
+      distinct,
+      updates,
+      select,
+      preloads,
+      assocs
+    ])
   end
 
   defp bound_from(nil, name), do: ["from #{name} in query"]
+
   defp bound_from(%{source: source} = from, name) do
-    ["from #{name} in #{inspect_source source}"] ++ kw_as_and_prefix(from)
+    ["from #{name} in #{inspect_source(source)}"] ++ kw_as_and_prefix(from)
   end
 
   defp inspect_source(%Ecto.Query{} = query), do: "^" <> inspect(query)
-  defp inspect_source(%Ecto.SubQuery{query: query}), do: "subquery(#{to_string query})"
-  defp inspect_source({source, nil}), do: inspect source
-  defp inspect_source({nil, schema}), do: inspect schema
+  defp inspect_source(%Ecto.SubQuery{query: query}), do: "subquery(#{to_string(query)})"
+  defp inspect_source({source, nil}), do: inspect(source)
+  defp inspect_source({nil, schema}), do: inspect(schema)
+
   defp inspect_source({source, schema} = from) do
-    inspect if source == schema.__schema__(:source), do: schema, else: from
+    inspect(if source == schema.__schema__(:source), do: schema, else: from)
   end
 
   defp joins(joins, names) do
     joins
-    |> Enum.with_index
+    |> Enum.with_index()
     |> Enum.flat_map(fn {expr, ix} -> join(expr, binding(names, expr.ix || ix + 1), names) end)
   end
 
   defp join(%JoinExpr{qual: qual, assoc: {ix, right}, on: on} = join, name, names) do
-    string = "#{name} in assoc(#{binding(names, ix)}, #{inspect right})"
+    string = "#{name} in assoc(#{binding(names, ix)}, #{inspect(right)})"
     [{join_qual(qual), string}] ++ kw_as_and_prefix(join) ++ maybe_on(on, names)
   end
 
-  defp join(%JoinExpr{qual: qual, source: {:fragment, _, _} = source, on: on} = join = part, name, names) do
+  defp join(
+         %JoinExpr{qual: qual, source: {:fragment, _, _} = source, on: on} = join = part,
+         name,
+         names
+       ) do
     string = "#{name} in #{expr(source, names, part)}"
     [{join_qual(qual), string}] ++ kw_as_and_prefix(join) ++ [on: expr(on, names)]
   end
 
   defp join(%JoinExpr{qual: qual, source: source, on: on} = join, name, names) do
-    string = "#{name} in #{inspect_source source}"
+    string = "#{name} in #{inspect_source(source)}"
     [{join_qual(qual), string}] ++ kw_as_and_prefix(join) ++ [on: expr(on, names)]
   end
 
   defp maybe_on(%QueryExpr{expr: true}, _names), do: []
   defp maybe_on(%QueryExpr{} = on, names), do: [on: expr(on, names)]
 
-  defp preloads([]),       do: []
+  defp preloads([]), do: []
   defp preloads(preloads), do: [preload: inspect(preloads)]
 
-  defp assocs([], _names),    do: []
+  defp assocs([], _names), do: []
   defp assocs(assocs, names), do: [preload: expr(assocs(assocs), names, %{})]
 
   defp assocs(assocs) do
-    Enum.map assocs, fn
+    Enum.map(assocs, fn
       {field, {idx, []}} ->
         {field, {:&, [], [idx]}}
+
       {field, {idx, children}} ->
         {field, {{:&, [], [idx]}, assocs(children)}}
-    end
+    end)
   end
 
   defp windows(windows, names) do
@@ -140,20 +184,20 @@ defimpl Inspect, for: Ecto.Query do
   end
 
   defp bool_exprs(keys, exprs, names) do
-    Enum.map exprs, fn %{expr: expr, op: op} = part ->
+    Enum.map(exprs, fn %{expr: expr, op: op} = part ->
       {Map.fetch!(keys, op), expr(expr, names, part)}
-    end
+    end)
   end
 
   defp kw_exprs(key, exprs, names) do
-    Enum.map exprs, &{key, expr(&1, names)}
+    Enum.map(exprs, &{key, expr(&1, names)})
   end
 
   defp kw_expr(_key, nil, _names), do: []
-  defp kw_expr(key, expr, names),  do: [{key, expr(expr, names)}]
+  defp kw_expr(key, expr, names), do: [{key, expr(expr, names)}]
 
   defp kw_inspect(_key, nil), do: []
-  defp kw_inspect(key, val),  do: [{key, inspect(val)}]
+  defp kw_inspect(key, val), do: [{key, inspect(val)}]
 
   defp kw_as_and_prefix(%{as: as, prefix: prefix}) do
     kw_inspect(:as, as) ++ kw_inspect(:prefix, prefix)
@@ -169,7 +213,7 @@ defimpl Inspect, for: Ecto.Query do
   end
 
   # For keyword and interpolated fragments use normal escaping
-  defp expr_to_string({:fragment, _, [{_, _}|_] = parts}, _, names, part) do
+  defp expr_to_string({:fragment, _, [{_, _} | _] = parts}, _, names, part) do
     "fragment(" <> unmerge_fragments(parts, "", [], names, part) <> ")"
   end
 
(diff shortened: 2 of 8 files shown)


lib/ecto/query.ex
```diff
@@ -980,6 +992,69 @@ defmodule Ecto.Query do
     [on: expr]
   end
 
+  @doc """
+  A recursive common table expression (CTE) also known as WITH expression.
+
+  For keyword queries use `:with` or `:with_recursive` keywords.
```
josevalim: I would rather have `cte` and `recursive_cte`. With two functions, in order to keep everything consistent. :)

lib/ecto/query.ex
```diff
@@ -980,6 +992,69 @@ defmodule Ecto.Query do
...
+
+  ## Expression examples
+
+  Products and their category names for breadcrumbs:
+
+    Product
+    |> cte(category_tree_query, as: "category_tree", recursive: true)
+    |> join(:left, [p], c in "category_tree", on: c.id == p.category_id)
+    |> group_by([p], p.id)
+    |> select([p, c], %{p | category_names: fragment("ARRAY_AGG(?)", c.name)})
```
josevalim: Please indent with four spaces!

lib/ecto/query.ex
```diff
@@ -980,6 +992,69 @@ defmodule Ecto.Query do
...
+  """
+  def cte(%__MODULE__{with: with_expr} = query, with_query, opts) do
+    unless opts[:as], do: raise "`:as` option is required in `cte/3`"
+
+    with_expr = with_expr || %WithExpr{}
+
+    with_expr = %{
+      with_expr
+      | queries: with_expr.queries ++ [%WithQueryExpr{query: with_query, as: opts[:as]}],
```
josevalim: I would like to store this as `[{key, query}]`. This way we don't need the `WithQueryExpr` struct. Mirroring the structure we see in windows.

test/ecto/query_test.exs
```diff
@@ -97,6 +97,20 @@ defmodule Ecto.QueryTest do
...
+    test "adds CTEs" do
+      cte1 = from p in "posts1"
+      cte2 = from p in "posts2"
+      query = "posts" |> cte(cte1, as: :cte1, recursive: true) |> cte(cte2, as: :cte2)
+
+      assert query.with.recursive
+      assert query.with.queries |> Enum.at(0) |> Map.fetch!(:as) == :cte1
+      assert query.with.queries |> Enum.at(0) |> Map.fetch!(:query) == cte1
+      assert query.with.queries |> Enum.at(1) |> Map.fetch!(:as) == :cte2
+      assert query.with.queries |> Enum.at(1) |> Map.fetch!(:query) == cte2
```
josevalim: We would need tests for interpolation, to make sure we account for ctes correctly.

josevalim:
Thanks @feymartynov for another great pull request!

I have added some comments more specific to the code but, before addressing them, I think we should discuss the feature in general. Non-recursive CTEs do not provide anything meaningful to Ecto, because we are capable of expressing something similar using Ecto's query syntax. So I think the focus here should be on recursive ones. I would even go to the point of supporting only recursive (and maybe call the feature `recursive`).

The other thing is that, for recursive CTEs, we can actually pass "arguments" to the name, which is something I did not see explored here.

What are your thoughts?

zachdaniel:
I think non recursive CTEs are still important for performance reasons at least. If you have reason to join to a single CTE more than once, having that is better than joining to a duplicated subquery multiple times. I could be wrong though, Postgres might know what to do with repeating subqueries

feymartynov:
I agree with @zachdaniel. In my project for example I use them for scoping soft delete and data available for current user and there are cases when I join those scopes more than once.

feymartynov:
> The other thing is that, for recursive CTEs, we can actually pass "arguments" to the name, which is something I did not see explored here.

If I understand the standard correctly those "arguments" are just aliases for selected fields. We can specify them inside the query itself like I did in the test case so they don't look necessary. But if we want to stick with the standard I can add them but I need some insight about how to pass them with keyword syntax.

josevalim:
> If I understand the standard correctly those "arguments" are just aliases
for selected field

The question is: is this a syntactic behavior or does it change the
semantics too?
-- 


josevalim:
If so I believe we can use the keyword syntax. Assuming we need both recursive CTE and non-recursive ones, then there are two thing left:

1. Figure out the name: I guess `cte` and `recursive_cte` are good enough?
2. Figure out how to do the recursive bit. It seems that, once you set one single entry to recursive, it affects all of the other `cte`s, so this is not quite composable, so I am not quite sure how to handle this

Thoughts?

zachdaniel:
I agree with #1 but need to some more research on #2

zachdaniel:
yeah, after looking into it I think the only option is to have `cte: ...` and then a separate `recursive: ...`  for composability. We can either take the latest declaration for recursive, or allow only one to be set and raise an error if it’s overridden.

josevalim:
I have been thinking about this more and, the issue with how it is implemented today, is that we lose the Ecto type information. Before I make a different proposal, are the CTE requires to appear as the source of a FROM or JOIN in SQL?

zachdaniel:
Yeah, I think we could support argument-less for now, and just do `with_recursive: true`. And support arguments via some other value applied to that config.  As for your question, you can use a CTE in a from/join further on in the query.

josevalim:
> As for your question, you can use a CTE in a from/join further on in the query.

Right, but only then? Or can I use them elsewhere?

Because I am thinking that maybe we should implement this like we implement subqueries:

    from p in cte(query), ...

This would allow us to leverage the type checking/binding mechanism that exists for subqueries. The only difference is that we move those out into a WITH when we build the query.

feymartynov:
> Regarding the argument, queries like this does not work without them

In MySQL it really doesn't, in PostgreSQL it's ok.

> Yeah, I think we could support argument-less for now, and just do `with_recursive: true`.

I agree.

> Right, but only then? Or can I use them elsewhere?

> you can use a CTE in a from/join further on in the query.

Also in other CTEs and in all parts of UNION.

> ```from p in cte(query), ...```

This doesn't allow CTE to be used in the second part of UNION. However current implementation doesn't allow it too. I think we need a test case for that.

josevalim:
> This doesn't allow CTE to be used in the second part of UNION. However current implementation doesn't allow it too. I think we need a test case for that.

Good point. We will need a way to figure out how bindings will work then so we can keep the type information.

Something like `from f in "foo", cte: [foo: Post]` should return Post structs correctly.

feymartynov:
So what about the API? Is this ok?

```elixir
Product
|> with_recursive(true)
|> cte(category_tree_query, as: "category_tree")
|> join(:inner, [p], c in "category_tree", c.id == p.category_id)
```

```elixir
from p in Product,
  with_recursive: true,
  cte: category_tree_query,
  as: "category_tree",
  join: c in "category_tree",
  on: c.id == p.category_id
```

josevalim:
I still think we need to find a way to make the Ecto.Types work, as that is very important to how Ecto works. We could do it based on strings, but it would be too brittle.

The approach I proposed (i.e. `join: c in cte(foo)`) has some flaws, but maybe we can find something that mixes both together? Unsure.

zachdaniel:
I think in that context that we really just ought to make adding a cte introduce a binding the same way anything else would>

```
Product
|> with_recursive(true)
|> cte(category_tree in category_tree_query, as: :category_tree)
|> join(:inner, [p], c in category_tree, c.id == p.category_id)

from p in Product,
  with_recursive: true,
  cte: category_tree in category_tree_query,
  as: :category_tree,
  join: c in category_tree,
  on: c.id == p.category_id
```

Now, in the case that using a binding in the query *also* as a source in another part of the query causes insurmountable problems (something that seems likely), I think we can use the `cte(foo)` as a way to like...treat that as its own source. Basically the binding can be used as if it was something previously joined on, which also ought to work with how CTEs work in general, and then if used as a source it is wrapped in `cte(foo)`. So it would work like this:

```
Product
|> with_recursive(true)
|> cte(category_tree in category_tree_query, as: :category_tree)
|> join(:inner, [p], c in cte(category_tree), c.id == p.category_id)

from p in Product,
  with_recursive: true,
  cte: category_tree in category_tree_query,
  as: :category_tree,
  join: c in cte(category_tree),
  on: c.id == p.category_id
```

zachdaniel:
So basically we just store each CTE in the query, and when its referenced as a source we use the type information from that stored query.

josevalim:
Thanks Zach! I have a hunch that your approach is more flexible than the
one I proposed but I can’t come up with a concrete example. Can you provide
some? We still cannot use the CTE in an union in both approaches though.
-- 


zachdaniel:
I really just pulled this off of the top of my head, but something like this where you might have two sources from the same CTE and one source joining on that CTE might be a good example

```
WITH active_users AS (
  SELECT * FROM users WHERE deleted_at IS TRUE
)
SELECT 
  active_users.id,
  COUNT(employees.id) AS employee_count,
  COUNT(children.id) AS children_count,
FROM active_users
JOIN employees
ON employees.manager_id = active_users.id
JOIN active_users AS active_user_children
ON active_users.child IS TRUE
AND active_users_children.parent_id = active_users.id
```

I'm in the weird spot where I don't have a good dataset/database to actually build a useful query from lol

zachdaniel:
NOTE: I wrote all of this out, but I realized afterwards that it doesn't actually have an example in it of using the CTE just as a binding and not as a source. I'm leaving it here in case its worth talking about though. Skip to the bottom for the important part:

Fun fact, that actually highlights a small issue which is that you might want to use a CTE as the source of the query itself... Regardless, lets ignore that for now and do it this way instead:

```sql
WITH active_users AS (
  SELECT * FROM users WHERE deleted_at IS TRUE
)
SELECT 
  active_users.id,
  COUNT(employees.id) AS employee_count,
  COUNT(children.id) AS children_count,
FROM users
JOIN employees
ON employees.manager_id = active_users.id
JOIN active_users
ON active_users.id = 
JOIN active_users AS active_user_children
ON active_user_children.child IS TRUE
AND active_users_children.parent_id = active_users.id
```

The ecto version of that would look like:

```elixir
active_users = from user in User, where: is_nil(user.deleted_at)

from user in User,
  cte: au in active_users,
  join: employee in cte(au),
  on: employee.manager_id == employee.id
  join: active_user in cte(au),
  on: active_user.id == user.id,
  join: children in cte(au),
  on: children.child == true,
  on: children.parent_id == user.id
```


Important part:

The above doesn't have an actual example of using it in a binding. In reality, the way I see it being used in a binding is with the `in` operator, since you can say something like `WHERE thing IN cte` in sql and that would be very useful.  *aside* from that, I think we might not need it available except for as a source to `from` as you had originally stated.

josevalim:
Thanks @zachdaniel!

Reading your reply made me realize that @feymartynov was correct all along. If we provide his syntax and then we want it to also propagate the types upstream, we can just bind it again:

```
    from c in {"category_tree", CategoryTree}, # <---- this syntax! it should work just™
      with_recursive: category_tree_query,
      as: "category_tree",
      left_join: p in assoc(c, :products),
      group_by: c.id,
      select: %{c | products_count: count(p.id)}
```

So assuming we are happy with this, I would go with the original proposal of moving to a keyword list and the later suggestion of handling the recursion as a separate operation. So:

```
    from c in {"category_tree", CategoryTree}, # <---- this syntax! it should work just™
      ctes: [category_tree: category_tree_query],
      left_join: p in assoc(c, :products),
      group_by: c.id,
      select: %{c | products_count: count(p.id)}
```

The only big remaining question is how to handle arguments in the CTE.

hauleth:
I do not think that including this is good idea as IMHO. The reason is that CTE in Postgres have slightly different semantics form any other DB:

- PostgreSQL CTEs are recursive by default (what is already stated there)
- PostgreSQL CTEs are optimisation fences, so `WITH a AS (SELECT * FROM foos) SELECT * FROM a WHERE a.id = 1` will fetch **all** rows from `foo` and then do filtering
- PostgreSQL CTEs allow DML in them (that is the reason why these are optimisation fence)

So in general in 99.9% of the cases it is desirable to use subqueries instead which Ecto supports quite well. In that singular cases when CTEs would be needed it is IMHO better to use view or just raw SQL.

feymartynov:
@hauleth Raw SQL is fine until you have to compose queries. However I agree about performance. For example @zachdaniel's query above in real app most likely would be slow because at first PostgreSQL would do seq scan over the whole users table and then would be unable to do hash join. But there are many other ways of writing slow queries with Ecto, it's not the problem of the tool :)

zachdaniel:
@josevalim The only problem with doing a single argument of `ctes: [foo: bar]` is that later CTEs can reference earlier ones. So as long as we can specify that key multiple times, and/or we break it out into multiple `cte: ` flags, you'd lose that capability.

zachdaniel:
@hauleth I agree that it is often not the most performant, but I've also had many experiences where having CTEs available would have been my savior, it was just too long ago to recall the exact use case.

feymartynov:
Here one more synthetic example with UNION:

```sql
WITH a AS (
  SELECT * FROM addresses WHERE deleted_at IS NULL
)

SELECT c.name, a.*
FROM customers AS c
LEFT JOIN a
ON a.id = c.address_id

UNION ALL

SELECT p.name, a.*
FROM providers AS p
LEFT JOIN a
ON a.id = p.address_id
```

With providing type information through tuples like @josevalim mentioned above it colud be like this:

```elixir
from c in Customer,
  cte: addresses_query,
  as: :adresses,
  left_join: a in :addresses,
  on: a.id == c.address_id,
  select: [c.name, {a, Address}],
  union_all: (
    from p in Provider,
      cte: a in cte(addresses_query)
      left_join: a in :addresses,
      on: a.id == p.address_id,
      select: [p.name, {a, Address}]
  )
```

Coming back to column names list we could optionally provide them like this:

```elixir
as: {:addresses, [:city, :street, :house]}
```

hauleth:
For me it still "feels unnatural". It would be better to have something like:

```elixir
with foos: (from foo in Foo), bazs: (from baz in Baz) do
  from foo in foos, inner_join: baz from bazs, on: foo.id == baz.foo_id
end
```

Or similar as it more directly match SQL syntax and syntax proposed there can be sometimes confusing, as often (at least for me) the data returned from CTE query has (almost) nothing to do with the selected tables. Especially as I do not quite understand how it works, the top level `from` is the first entry in the CTE or it is the CTE query itself? If first then it can get enormously confusing, if last then how do you make query from tables introduced in CTE? For example the CTE from my application (live and running):

```sql
WITH RECURSIVE
  t AS (
    SELECT l.*
    FROM lab_tests l
    WHERE l.id = $1
    UNION ALL
    SELECT l.*
    FROM lab_tests l
    LEFT JOIN packages_to_tests pt ON pt.test_id = l.id
    INNER JOIN t ON t.id = pt.package_id OR t.id = l.parent_id
  )
SELECT * FROM t
```

How it would work in Ecto? Could you give me query example of such SQL? AFAIK there is no way to reference outer query in subquery.

josevalim:
Here are my two cents:

1. Let's keep this it out of the keyword query syntax. Since CTEs happen before the `from` (i.e. select, update, delete), it doesn't have a good match with the current DSL.

2. I would implement it as a simple macro, called `with_cte`:

```elixir
query
|> with_cte("name_of_cte", as: query)
|> with_cte("this_needs_to_be_a_string_at_compile_time(foo, bar)", as: query)
```

We can also add a `recursive_ctes/1` function (or `recursive_with/1`) that simply receives a query and marks it as recursive. 

Thoughts?

hauleth:
@josevalim what would be the `query` here? If the actual CTE query, then the order should be reversed IMHO.

josevalim:
Sorry, that was a horrible example as I used `query` everywhere. I have updated it. Please check again.

hauleth:
That still leave ambiguity. Should I understand that with your syntax my query would look like:

```elixir
query = from row in "t"

query
|> with_recursive_cte(
  "t",
  from(test in LabTests,
    where: test.id == ^id,
    union: (
      from test in LabTest,
      left_join: package in "packages_to_tests", on: package.test_id == test.id,
      inner_join: sub_test in "t", on: sub_test.id == test.parent_id or sub_test.id == package.package_id))
```

It kind of defeats the main selling point of CTEs which is the fact that you can read them from top to bottom. Ecto does quite nice job of being more like QUEL rather than SQL (which is good thing) and this syntax makes it a little bit weird again.

josevalim:
I am not sure it matters this much because it is unlikely you will build
the whole query at once anyway.

Plus you can always do this:

    foo_cte = from ...
    bar_cte = from ...
    query = from ...
    # wire it all up

I think trying to tie with and from somehow will just make it look weird on
the elixir side of things.
-- 


hauleth:
@josevalim even in that case it will make it weird to read

```elixir
sub_tests =
    from test in LabTest, # 4
    left_join: package in "packages_to_tests",
    on: package.test_id == test.id,
    inner_join: sub_test in "t", # 5
    on: sub_test.id == test.parent_id or sub_test.id == package.package_id
test =
    from test in LabTest, # 2
    where: test.id == ^id,
    union: sub_tests # 3
query = from row in {"t", LabTest} # 6

query
|> with_recursive_cte("t", test) # 1
```

Where with numbers in comments I have marked order of reading to understand all query. Don't you think it looks weird? Like hopping back and forth constantly until anything of it make any sense? In such situation I think that writing raw SQL query is clearer solution, even when I need to parse results like:

```query
results.rows
|> Enum.map(&Repo.load(LabTest, {results.columns, &1}))
|> Repo.preload([:parent])
```

josevalim:
If you think SQL is clearer in those cases then it is definitely your
choice to make. :)

I think the currently proposed API is the simplest one possible. Which
other alternatives are there?
-- 


hauleth:
@josevalim

```elixir
with_cte t <- (
    from test in LabTest, where: test.id == ^id, union: (
      from test in LabTest, # 4
      left_join: package in "packages_to_tests",
      on: package.test_id == test.id,
      inner_join: sub_test in "t", # 5
      on: sub_test.id == test.parent_id or sub_test.id == package.package_id
  ),
  from: row in {t, LabTest}
```

josevalim:
Or look at it like this: we need to support the pipe based syntax anyway.
So let’s start with that, since it is simpler, and we can develop a query
based one later on.
-- 


hauleth:
But no matter what, even with my proposal, it is hard to provide support for all functionalities, because (as I said) PostgreSQL allows DML in CTEs (following SQL standard IIRC), so this is perfectly valid CTE:

```
WITH
foo AS (
  INSERT INTO foos (id, val) VALUES (1, 10)
)
SELECT * FROM foos
```

And if I understand this proposal correctly there is no way to support such query.

josevalim:
The way to support those would be as:

    with_cte query, “foo”, as: fragment(“...”)
-- 


hauleth:
And I do not think that we should support *all* syntax via pipe form as not all SQL is meant to be written in such form. `WITH` is perfect (only?) example of that. IMHO good test is to rewrite this query using such syntax and check if this is still (quite) readable:

```sql
WITH RECURSIVE q(r, i, rx, ix, g) AS (
  SELECT r::DOUBLE PRECISION * 0.02, i::DOUBLE PRECISION * 0.02, 
        .0::DOUBLE PRECISION      , .0::DOUBLE PRECISION, 0
  FROM generate_series(-60, 20) r, generate_series(-50, 50) i
  UNION ALL
  SELECT r, i, CASE WHEN abs(rx * rx + ix * ix) <= 2 THEN rx * rx - ix * ix END + r, 
               CASE WHEN abs(rx * rx + ix * ix) <= 2 THEN 2 * rx * ix END + i, g + 1
  FROM q
  WHERE rx IS NOT NULL AND g < 99
)
SELECT array_to_string(array_agg(s ORDER BY r), '')
FROM (
  SELECT i, r, substring(' .:-=+*#%@', max(g) / 10 + 1, 1) s
  FROM q
  GROUP BY i, r
) q
GROUP BY i
ORDER BY i
```

This is [Mandelbrot set written as PostgreSQL CTE](https://explainextended.com/2013/12/31/happy-new-year-5/) and IMHO it is quite readable example, but quite complex.

So in my opinion it is better to not provide functionality at all instead of providing it in "not fairly usable way, because we want to fix it later".

This isn't simple task, and there will be more to come (like pivoting/unpivoting, temporal tables, row pattern matching, table samples, grouping sets, and stored procedures support). The question is how much we want to squeeze into Ecto and how much we can support from outside, because with every feature we will implement there will be more to come (especially if we want to also suport NoSQL DBs and/or Column-oriented DBs with SQL-like syntax). While CTEs are nice thing to support, I think it is exactly the feature we shouldn't support via pipe syntax. While pipes can be used for a lot of SQL, the CTEs aren't the thing as it will bring more confusion rather than a solutions (as this will be only thing in Ecto queries which needs to be read "upside down" without `fragments`).

josevalim:
> So in my opinion it is better to not provide functionality at all instead
of providing it in "not fairly usable way, because we want to fix it later".

I don’t see how supporting a syntax we will have to support anyway fits
this description.

Also, you were advocating to support only strings as a starting point for
range/rows over windows. How is this any different?
-- 


josevalim:
Anyway, we don’t have to support this. Ecto 3.0 is meant to be stable API
though. So if we don’t get it now, don’t expect it later either.
-- 


zachdaniel:
@hauleth I agree with @jose. It is a simple trade off: composability and compartmentalizations of concerns, in exchange for not being able to read the query from top to bottom as if it were sql.

josevalim:
> Just a question, if the pipe API si so simple then why Elixir itself decided to use with instead of "new kind" of pipe like for example ~>? I know the answer: flexibility. For me this is even more obvious case as there we can alias items created in WITH statement, so whole query could end very confusing to read, at least for me.

That's the point, isn't it? The `with` is an abstraction of an existing pattern. You can express what happens in Elixir's `with` today, just using function calls, and the regular pipe syntax. It is your choice to make between `with` and regular pipe plus pattern matching.

But for some reason you are saying that the only way to write this code is via a complex macro, which is what I disagree with. That would be the equivalent to say Elixir should remove pattern matching in function clauses now that we introduced `with` because `with` is more flexible.

Btw, the reason why `with` only arrived in Elixir v1.3 is because it took a while for us to figure out how it should look like. And since you have made this comparison, I would go ahead and say that we shouldn't rush it here either. In fact, shipping with a simple API will allow people to experiment with different syntaxes and approaches, which is exactly what happened before we added `with` to the language.

It is fine for us to disagree here. :) But I have made my mind that we will start with the simple API for now.

> Also in DBs without WITH should we emulate them by using subqueries or throw an error about unsupported features? IIRC this is how MSSQL treats WITH, I do not know how about other DBs though.

In Ecto we generally don't try to "rewrite queries". So if WITH can be emulated with something else that is not a WITH, we typically don't do it, and just ask the adapters to raise. Some complex combinations, like using WINDOW + LOCK raising an error in Postgres, we don't check for those and we let the DB do the work.

josevalim:
@feymartynov I will open up a new issue with an API proposal for CTE later today. Then you can either submit a new pull request or change this one. Thanks everyone for the discussion.

hauleth:
@josevalim what is so complex in my macro? For me it is quite simple to create with given PR as a different syntax. Both could be easily implemented alongside each other, however the problem is that I consider "pipe" syntax plainly harmful as very easy to abuse such syntax. It is even more troublesome than named joins as you cannot check if the queried table exists until you fail at DB boundary. IMHO it is very prone to the typos. As I said, the main concern for me is that you can query table that do not exist yet, and it will be created **later** in code. So this can be quite confusing for the users.

josevalim:
Your syntax is complex for the user, because it is something new to learn. Maybe it is the best macro-based approach but I don't want to decide it on a whim. On the other hand, everyone knows how function calls work. Again, your thoughts on the pipe syntax have been clear, and we will have to agree to disagree. We will go ahead with the call syntax for now, which you may pipe to it or not.

hauleth:
But that is my point. Everybody knows how function calls work, so this:

```elixir
foo(bar)
bar = 5
```

Is expected to fail, so why this:

```elixir
from foo in :foo,
  cte: bar,
  as: :foo
```

Should work?

For me it seems counterintuitive.

josevalim:
Thats late binding and it is exactly what a higher level construct allow us
to do. For example, Ecto.Query.dynamic works like that too.

I agree with you that writing the straight-forward way would be nice too,
but there are still things to figure out there, so the late binding
approach, which we will have anyway, will suffice for now.
-- 


OvermindDL1:
> Should work?

Just as you referenced `let`/`let*` above, think of this like a Haskell `where` clause:
```haskell
f x = y
  where
    y = z...
```

hauleth:
@OvermindDL1 however Haskell `where` do not bleed to the caller while `cte` does. So there is possibility that change in one function can cause other function to query wrong table. In Haskell it is not a case.

hauleth:
For example:

```elixir
from post in query,
  cte: cte_query, as: :users
```

And somewhere else down (or up) the pipeline I would do:

```elixir
from post in query,
  inner_join: user in assoc(post, :author)
```

Where schemas looks like:

```elixir
defmodule Post do
  use Ecto.Schema

  schema "posts" do
    belongs_to :author, User
  end
end

defmodule User do
  use Ecto.Schema

  schema "users" do
    has_many :posts, Posts, foreign_key: :author_id
  end
end
```

And now we can get quite confusing error about missing columns in table `users` when `cte_query` do not return all fields that `User` schema needs.

feymartynov:
> I will open up a new issue with an API proposal for CTE later today. Then you can either submit a new pull request or change this one. Thanks everyone for the discussion.

Thank you, that would be nice because I'm kind of lost in this variety of opinions :)

feymartynov:
#2769

feymartynov:
I have updated this PR and opened a new one to ecto_sql: https://github.com/elixir-ecto/ecto_sql/pull/25

Please, check 'em out.

lib/ecto/query.ex
```diff
@@ -635,6 +641,7 @@ defmodule Ecto.Query do
   defp do_exclude(%Ecto.Query{} = query, :order_by), do: %{query | order_bys: []}
   defp do_exclude(%Ecto.Query{} = query, :group_by), do: %{query | group_bys: []}
   defp do_exclude(%Ecto.Query{} = query, :combinations), do: %{query | combinations: []}
+  defp do_exclude(%Ecto.Query{} = query, :with), do: %{query | with: nil}
```
josevalim: We should probably call this `with_cte`?

lib/ecto/query.ex
```diff
@@ -993,6 +1000,98 @@ defmodule Ecto.Query do
...
 
+  @doc """
+  A common table expression (CTE) also known as WITH expression.
+
+  `name` must be a compile-time literal string that is being used as the table name to join
+  the CTE in the main query or in the recursive CTE.
+
+  ## Options
+
+  * `:as` – the CTE query itself or a fragment
```
josevalim:
```suggestion
    * `:as` – the CTE query itself or a fragment
```

lib/ecto/query/planner.ex
```diff
@@ -721,6 +739,23 @@ defmodule Ecto.Query.Planner do
...
+  defp plan_ctes(%Ecto.Query{with: %{queries: queries}} = query, operation, adapter, counter) do
+    queries =
+      Enum.map queries, fn
+        {name, %Ecto.Query{} = query} ->
+          {planned_query, _params, _key} = plan(query, operation, adapter, counter)
+          planned_query = planned_query |> ensure_select(true)
+          {name, planned_query}
+
+        {name, other} ->
+          {name, other}
```
josevalim: I think we need to call `plan` recursively here too so we handle any argument in the fragment.

lib/ecto/query/planner.ex
```diff
@@ -814,6 +849,30 @@ defmodule Ecto.Query.Planner do
...
+        {name, %Ecto.Query{} = query}, {queries, counter} ->
+          {query, counter} = traverse_exprs(query, operation, counter, fun)
+          {query, _} = normalize_select(query)
+          {_, select} = query |> subquery_select(adapter)
+          keys = %{select: select} |> subquery_types() |> Keyword.keys()
+          query = query.select.fields |> update_in(&Enum.zip(keys, &1))
+          {[{name, query} | queries], counter}
+
+        other, {queries, counter} ->
+          {[other | queries], counter}
```
josevalim: I think we need to call `fun` recursively here too so we handle any argument in the fragment.

test/ecto/query/planner_test.exs
```diff
@@ -414,6 +414,13 @@ defmodule Ecto.Query.PlannerTest do
     assert %Ecto.Query.SelectExpr{expr: {:&, [], [0]}} = query.select
   end
 
+  test "prepare: prepare CTEs" do
```
josevalim: We need to add a test that guarantees that we can handle parameter interpolation in both query case and in the fragment case.

lib/ecto/query.ex
```diff
@@ -354,7 +354,7 @@ defmodule Ecto.Query do
 
   defstruct [prefix: nil, sources: nil, from: nil, joins: [], aliases: %{}, wheres: [], select: nil,
              order_bys: [], limit: nil, offset: nil, group_bys: [], combinations: [], updates: [],
-             havings: [], preloads: [], assocs: [], distinct: nil, lock: nil, windows: []]
+             havings: [], preloads: [], assocs: [], distinct: nil, lock: nil, windows: [], with: nil]
```
josevalim: Should we also call the field `with_ctes`?

josevalim:
Thank you @feymartynov! this looks great, i have added some comments but they are minor. The only important bit is that we need to add tests for the parameter handler and make sure we recur query expressions in the planner.

lib/ecto/query/planner.ex
feymartynov: Didn't get it. `plan` doesn't work for `QueryExpr`.

josevalim:
You are correct. And you also implemented merge_cache, so we should be
good. Apologies!
-- 


feymartynov:
Updated the code.
Tests for parameter interpolation are in the ecto_sql's PR.

josevalim:
Thanks @feymartynov! Now that `ecto_sql` is in a separate repository, it would be nice if we assert the parameters on Ecto side of things. The planner suite already checks for that, can you please update the planner tests then? I have recently added the union parameters there, so if you can rebase and increment, it would be perfect: 8c1662aed1f822f8881c296f8ba2767cb0b7b9cf

lib/ecto/query/planner.ex
```diff
@@ -814,6 +849,32 @@ defmodule Ecto.Query.Planner do
...
+        {name, %Ecto.Query{} = query}, {queries, counter} ->
+          {query, counter} = traverse_exprs(query, operation, counter, fun)
+          {query, _} = normalize_select(query)
+          {_, select} = query |> subquery_select(adapter)
+          keys = %{select: select} |> subquery_types() |> Keyword.keys()
+          query = query.select.fields |> update_in(&Enum.zip(keys, &1))
+          {[{name, query} | queries], counter}
+
+        {name, %QueryExpr{expr: {:fragment, _, _} = fragment} = query_expr}, {queries, counter} ->
+          {fragment, counter} = prewalk_source(fragment, :cte, query, with_expr, counter, adapter)
```
```suggestion
          {fragment, counter} = prewalk_source(fragment, :with_cte, query, with_expr, counter, adapter)
```

lib/ecto/query/planner.ex
```diff
@@ -1285,7 +1346,7 @@ defmodule Ecto.Query.Planner do
 
   ## Helpers
 
-  @exprs [distinct: :distinct, select: :select, from: :from, join: :joins,
+  @exprs [with_ctes: :with_ctes, distinct: :distinct, select: :select, from: :from, join: :joins,
```
```suggestion
  @exprs [with_cte: :with_ctes, distinct: :distinct, select: :select, from: :from, join: :joins,
```

IIRC the left side is used for error reporting, so it needs to be the name of the macro/operation the user calls.

test/ecto/query/inspect_test.exs
```diff
@@ -69,6 +69,35 @@ defmodule Ecto.Query.InspectTest do
...
+    cte_query = initial_query |> union_all(^iteration_query)
+
+    query =
+      "products"
+      |> recursive_ctes(true)
+      |> with_cte("tree", as: ^cte_query)
+      |> join(:inner, [r], t in "tree", on: t.id == r.category_id)
+
+    assert i(query) ==
+      ~s{from p in "products", recursive_ctes: true, with_ctes: [tree: (from c in "categories",\n} <>
```
josevalim: This is tricky because we actually don't support this format.

Maybe we should emit this: `(from ...) |> with_cte(...) |> recursive_ctes(true)`?

feymartynov: Ok. What about `#Ecto.Query<…>` form? Should it also be `(#Ecto.Query<…>) |> with_cte("name", as: #Ecto.Query<…>)`?

josevalim: Yes, that's what I would do, i think. No need to wrap the `#Ecto.Query<...>` one in parens though.

feymartynov:
@josevalim please check out the update.

josevalim:
Hi @feymartynov! Right now we are focused on patch releases for Ecto. Once that is out of the way we will focus on merging non-maintenance work. :) Thanks!

kenken1mura:
close

icehaunter:
Hey, sorry to be a nuisance, but any idea on when this could be merged? CTEs would allow so many easy solutions for current problems!

josevalim:
Hi @icehaunter, it is no bother at all. There are some concerns about the limited approach in this PR (only pipe syntax) but more importantly, we are mostly focused on bug fixes since v3.0.

lib/ecto/query/inspect.ex
```diff
@@ -27,7 +27,22 @@ defimpl Inspect, for: Ecto.Query do
         string
     end)
 
-    surround_many("#Ecto.Query<", list, ">", opts, fn str, _ -> str end)
+    result = surround_many("#Ecto.Query<", list, ">", opts, fn str, _ -> str end)
+
+    case query.with_ctes do
+      %WithExpr{recursive: recursive, queries: [_] = queries} ->
```
josevalim:
```suggestion
      %WithExpr{recursive: recursive, queries: [_ | _] = queries} ->
```

test/ecto/query/planner_test.exs
josevalim: For example, we need to pass a query with parameters in `as: ^from_query_with_parameter` as well as test a fragment like this `fragment("SOME QUERY WITH ?", parameter)`.

lib/ecto/query/inspect.ex
```diff
@@ -27,7 +27,22 @@ defimpl Inspect, for: Ecto.Query do
...
+        with_ctes =
+          Enum.map(queries, fn {name, query} ->
+            cte = query |> __MODULE__.inspect(opts)
```
josevalim:
```suggestion
            cte = inspect(query, opts)
```

lib/ecto/query/planner.ex
```diff
@@ -593,6 +594,23 @@ defmodule Ecto.Query.Planner do
...
+  defp merge_cache(:with_cte, _query, with_expr, cache_and_params, operation, adapter) do
+    fun = &{&3, merge_cache(&1, &2, &3, &4, operation, adapter)}
+
+    Enum.reduce with_expr.queries, cache_and_params, fn
+      {_, %Ecto.Query{} = query}, acc ->
+        {_, acc} = traverse_exprs(query, operation, acc, fun)
+        acc
+
+      _, acc ->
+        acc
```
josevalim: In this branch we need to call `cast_and_merge_params`, like we do for other QueryExpr here: https://github.com/elixir-ecto/ecto/blob/3680d9b408672154c0ca16327b47ef42a89a2a8b/lib/ecto/query/planner.ex#L544-L556

josevalim: I have commented on the planner_test file about a test we should add that will expose the need for this.

lib/ecto/query/planner.ex
```diff
@@ -721,6 +739,23 @@ defmodule Ecto.Query.Planner do
...
+        {name, %Ecto.Query{} = query} ->
+          {planned_query, _params, _key} = plan(query, operation, adapter, counter)
+          planned_query = planned_query |> ensure_select(true)
+          {name, planned_query}
+
+        {name, other} ->
+          {name, other}
+      end
+
+    query.with_ctes.queries |> put_in(queries)
```
```suggestion
    put_in(query.with_ctes.queries, queries)
```

josevalim:
Hi @feymartynov. Sorry for the delay on this but things have finally calmed down in the Ecto front. I have dropped some final comments. Then we only need to rebase and we should be good to go. Thank you!

feymartynov:
Hello @josevalim! I'll do it on the weekend.

lib/ecto/query/inspect.ex
feymartynov: `__MODULE__` in necessary because:

```
** (CompileError) lib/ecto/query/inspect.ex:28: imported Kernel.inspect/2 conflicts with local function
```

feymartynov:
@josevalim Updated the PR.

josevalim:
Merged manually, thank you! 

chulkilee:
For the record: commit is ec47bb8483dd5206bd02b68c69c15eacf744be07

hauleth:
Just for the record - I still think that this is bad API and "following SQL/QUEL syntax" would be better choice.

PR #2758, opened 2018-10-20
Updated docs for cast_assoc/3

LostKobrakai:


zachdaniel:
Please undo the formatting that’s been done on the rest of the file to keep the diff clean. If you’d like to propose reformatting the file then use a different PR after this one to do so..

zachdaniel:
Looks like you took care of it before I finished my comment :D

lib/ecto/changeset.ex
```diff
@@ -680,6 +680,32 @@ defmodule Ecto.Changeset do
         end
       end
 
+  ## Partial changes for many-style associations
+
+  While the introductionary sentences for this function talk about `cast_assoc/3`
```
zachdaniel: I would open with something clearer like “if you’ve preloaded an association, that will confine the behavior of `cast_assoc` to the records you’ve preloaded.

LostKobrakai:
The example is now not that expressive anymore imho. I was going for the country of addresses, because all addresses of a country (over all users) is a place where editing just a subset of pages makes quite a lot of sense because the number is likely beyond something that should be shown as one full set.

We can certainly go for limiting the addresses of a single user to just a country, but then the example must be a changeset of a user not a country. Currently the example is saving a country, limiting addresses to that country, which is essentially editing the full set again – and not a subset.

The second point I have is that limiting the preload by ids is something, which is less likely to change below your feet between e.g. a edit and update request in phoenix. If I'm correct concurrent edits could only accidentally undelete items here.

`where: [country: ^edit_country]` on the other hand could delete stuff on concurrent edits.

Maybe that's nothing we should care too much here though. I mean both have issues with concurrent edits and it's not that `cast_assoc` without partial changes would behave any better :D

josevalim:
My concern with the previous approach is that, if you are changing just a single address, then it is better to flip the form around and possibly change the country with cast assoc for the parent. Also, the previous approach leaves it up in the air on how the IDs that are being changed should be retrieved, this one is a bit clearer.

If you don't have a strong preference for one or the other, then I slightly prefer the current approach, so I would rather go with it. :)

LostKobrakai:
Ah, it was rather meant like changing e.g. 50 addresses out of say hundreds. 

Yeah, I don't have a strong preference as long as the example is clear in the end. I'll go and update the changeset to be a user changeset similar to the first example then.

josevalim:

PR #2761, opened 2018-10-21
fix possible duplicate type

tcoopman:
@fertapric fixes: https://github.com/elixir-ecto/ecto/pull/2749#discussion_r226597676

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -271,7 +271,7 @@ defmodule Ecto.ChangesetTest do
     refute changeset.valid?
   end
 
-  test "cast/4: field has a custom invalid error message with a type parameter will be erased" do
+  test "cast/4: ignores the :type parameter in custom errors" do
     params = %{"custom_error_with_type" => :error}
     struct = %CustomErrorTest{}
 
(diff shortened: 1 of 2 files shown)


fertapric:
Thanks 💜!

I would add a test 🙂

josevalim:
Thanks! Although I actually think the previous approach is fine. The type
in the front should be given higher priority in all cases and we document
that we cant return it. :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

fertapric:
> The type in the front should be given higher priority in all cases 

`Enum.into(errors, %{})` puts in the map the last `:type` value.

I agree this is a minor change but I think it's worthy if we can prevent some undesired side effects 🙂

test/ecto/changeset_test.exs
```diff
@@ -262,6 +271,15 @@ defmodule Ecto.ChangesetTest do
     refute changeset.valid?
   end
 
+  test "cast/4: field has a custom invalid error message with a type parameter will be erased" do
```
fertapric: Maybe `cast/4: ignores the :type parameter in custom errors`?

josevalim:

PR #2773, opened 2018-10-24
Upgrade poison dependency

masonforest:


josevalim:
Unfortunately 4.0 removes the contract that Ecto uses, which is calling `encode_to_iodata`, so I am afraid we can't allow it. If you want to try it though, you can always specify it at the root of your project with override: true. Thanks!

PR #2776, opened 2018-10-29
Update repo.ex

arcz:
Fix typo


josevalim:
It is correct as is. The telemetry_prefix is at the global config. Per operation we can pass the event. :) Thanks!

arcz:
ops got confused because it is not anywhere is this codebase, but looks like ecto_sql uses it 👍

PR #2790, opened 2018-11-05
put_assoc still doesn't accept any options

CrowdHailer:
I'm pretty sure this is true. Although I only realized this after a lot of false starts.


josevalim:

PR #2798, opened 2018-11-07
Dump UTC datetime as a DateTime

### LINKED ISSUE #2796: Error when inserted_at/updated_at are timestamptz
(Ecto 3.0)

We have some timestamps in `timestamptz` format.

My definition looks like:

`timestamps(inserted_at: :created_at, type: :utc_datetime)`

But it seems the Ecto is inserting it as a NaiveDateTime, because I get this error:

```
** (DBConnection.EncodeError) Postgrex expected %DateTime{}, got ~N[2018-11-07 11:26:56]. Please make sure the value you are passing matches the definition in your table or in your query or convert the value accordingly.
```

josevalim:
Closes #2796.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -404,23 +404,14 @@ defmodule Ecto.Type do
   defp dump_naive_datetime_usec(%NaiveDateTime{microsecond: {_, 6}} = term), do: {:ok, term}
   defp dump_naive_datetime_usec(_), do: :error
 
-  defp dump_utc_datetime(%DateTime{time_zone: time_zone, microsecond: {0, 0}} = term) do
-    if time_zone != "Etc/UTC" do
-      message = ":utc_datetime expects the time zone to be \"Etc/UTC\", got `#{inspect(term)}`"
-      raise ArgumentError, message
-    end
+  defp dump_utc_datetime(%DateTime{microsecond: {0, 0}} = datetime),
+    do: {:ok, datetime |> check_utc_timezone!(:utc_datetime)}
 
-    {:ok, DateTime.to_naive(term)}
-  end
   defp dump_utc_datetime(_), do: :error
 
-  defp dump_utc_datetime_usec(%DateTime{time_zone: time_zone, microsecond: {_, 6}} = datetime) do
-    if time_zone != "Etc/UTC" do
-      message = ":utc_datetime_usec expects the time zone to be \"Etc/UTC\", got `#{inspect(datetime)}`"
-      raise ArgumentError, message
-    end
-    {:ok, DateTime.to_naive(datetime)}
-  end
+  defp dump_utc_datetime_usec(%DateTime{microsecond: {_, 6}} = datetime),
+    do: {:ok, datetime |> check_utc_timezone!(:utc_datetime_usec)}
+
   defp dump_utc_datetime_usec(_), do: :error
 
   defp dump_embed(%{cardinality: :one, related: schema, field: field},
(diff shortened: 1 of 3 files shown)


lib/ecto/type.ex
```diff
@@ -533,23 +524,43 @@ defmodule Ecto.Type do
...
+
   defp load_naive_datetime_usec(_), do: :error
 
-  defp load_utc_datetime(%DateTime{} = datetime),
-    do: {:ok, truncate_usec(datetime)}
+  # This is an upcast but because we assume the database
+  # is always in UTC, we can perform it.
   defp load_utc_datetime(%NaiveDateTime{} = naive_datetime),
-    do: naive_datetime |> truncate_usec() |> DateTime.from_naive("Etc/UTC")
+    do: {:ok, naive_datetime |> truncate_usec() |> DateTime.from_naive!("Etc/UTC")}
```
samphilipd: Does this mean that NaiveDateTime now always loads as a DateTime in UTC time zone?

josevalim: No, this is for `utc_datetime`. The behaviour here didn't change. I just rewrote the code to be in sync with everything else. :)

lib/ecto/type.ex
```diff
@@ -404,23 +404,14 @@ defmodule Ecto.Type do
   defp dump_naive_datetime_usec(%NaiveDateTime{microsecond: {_, 6}} = term), do: {:ok, term}
   defp dump_naive_datetime_usec(_), do: :error
 
-  defp dump_utc_datetime(%DateTime{time_zone: time_zone, microsecond: {0, 0}} = term) do
-    if time_zone != "Etc/UTC" do
-      message = ":utc_datetime expects the time zone to be \"Etc/UTC\", got `#{inspect(term)}`"
-      raise ArgumentError, message
-    end
+  defp dump_utc_datetime(%DateTime{microsecond: {0, 0}} = datetime),
+    do: {:ok, datetime |> check_utc_timezone!(:utc_datetime)}
```
qcam: perhaps we don't need to pipe since it's only one statement?

```suggestion
    do: {:ok, check_utc_timezone!(datetime, :utc_datetime)}
```

lib/ecto/type.ex
```diff
@@ -404,23 +404,14 @@ defmodule Ecto.Type do
...
 
-  defp dump_utc_datetime_usec(%DateTime{time_zone: time_zone, microsecond: {_, 6}} = datetime) do
-    if time_zone != "Etc/UTC" do
-      message = ":utc_datetime_usec expects the time zone to be \"Etc/UTC\", got `#{inspect(datetime)}`"
-      raise ArgumentError, message
-    end
-    {:ok, DateTime.to_naive(datetime)}
-  end
+  defp dump_utc_datetime_usec(%DateTime{microsecond: {_, 6}} = datetime),
+    do: {:ok, datetime |> check_utc_timezone!(:utc_datetime_usec)}
```
qcam: perhaps we don't need to pipe since it's only one statement?

```suggestion
    do: {:ok, check_utc_timezone!(datetime, :utc_datetime)}
```

lib/ecto/type.ex
wojtekmach: let's leave it as is, I'll add another check here in a separate PR.

josevalim:
Yup. I added it because I knew another check will be added. :)
-- 


lib/ecto/type.ex
```diff
@@ -1143,4 +1154,11 @@ defmodule Ecto.Type do
 
   defp pad_usec(%{microsecond: {microsecond, _}} = struct),
     do: %{struct | microsecond: {microsecond, 6}}
+
+  defp check_utc_timezone!(%{time_zone: "Etc/UTC"} = datetime, _kind), do: datetime
+
+  defp check_utc_timezone!(%{time_zone: time_zone} = datetime, kind) do
```
wojtekmach: `time_zone` is unused

josevalim:
```suggestion
  defp check_utc_timezone!(datetime, kind) do
```

lib/ecto/type.ex
```diff
@@ -1143,4 +1154,11 @@ defmodule Ecto.Type do
 
   defp pad_usec(%{microsecond: {microsecond, _}} = struct),
     do: %{struct | microsecond: {microsecond, 6}}
+
+  defp check_utc_timezone!(%{time_zone: "Etc/UTC"} = datetime, _kind), do: datetime
```
fertapric: Minor nitpick: should it be `check_utc_time_zone!` so we use “time zone” separated everywhere?

davydog187:
This is blocking us from upgrading to Phoenix 1.4, and I see people getting bit by this. https://elixirforum.com/t/ecto-3-0-app-with-utc-datetime-timestamps-gives-dbconnection-encodeerror-postgrex-expected-datetime-got-n-error/17879/2

Is https://github.com/xerions/mariaex/pull/245 blocking this from being merged?

Issue #2807, opened 2018-11-11
Using on_conflict: :replace_all does not infer column name when the name of the field is different from the name of the column


esdras:
Hi Team.

I think I found a bug, details below.

Thank you very much for your work in Ecto and Elixir in general ❤️ 

### Environment

* Elixir version (elixir -v): Elixir 1.7.1
* Database and version: PostgreSQL 11
* Ecto version (mix deps): ecto 3.0.1, ecto_sql 3.0.0
* Database adapter and version (mix deps): postgrex 0.14.0
* Operating system: macOS Sierra 10.12.6

### Current behavior
```elixir
# schema definition
defmodule Wallet do
  schema "wallets" do
    #... other fields
    field(:__rev__, :integer, source: :__gut_rev__) #source and field have different names
  end
end

cs = Wallet.changeset(%Wallet{}, params)
Repo.insert(cs, on_conflict: :replace_all, conflict_target: [:id])

```

```bash
# complains the database does not have a column named __rev__
21:51:17.563 link_id=1 [error] Task #PID<0.837.0> started from #PID<0.832.0> terminating
** (Postgrex.Error) ERROR 42703 (undefined_column) column excluded.__rev__ does not exist
```

### Expected behavior
When I remove the `source` option and rename the field to the name of the column, the error does not occur. I think this should be the expected behavior.

I think it has to do with this function:
https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L576

Thank you.

josevalim:
Yeah, I don't think we are normalizing any of the fields given to replace. We have to address this. Good catch!

qcam:
I can work on this if it has not been taken.

josevalim:
Please go ahead!
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

josevalim:
Closing in favor of #2808.

PR #2808, opened 2018-11-11
Take "source" into account in "replace_all_fields"

qcam:
Close https://github.com/elixir-ecto/ecto/issues/2807

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -549,7 +549,8 @@ defmodule Ecto.Repo.Schema do
         raise ArgumentError, ":conflict_target option is required when :on_conflict is replace"
 
       {:replace, keys} when is_list(keys) ->
-        {keys, [], conflict_target}
+        fields = Enum.map(keys, &field_source!(schema, &1))
+        {fields, [], conflict_target}
 
       :replace_all_except_primary_key ->
         fields = replace_all_fields!(:replace_all_except_primary_key, schema)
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @qcam! I think we also need to do it for the keys here: https://github.com/elixir-ecto/ecto/blob/e081f62297b7e1ff79d8776eb661f31e3c4ce1ab/lib/ecto/repo/schema.ex#L552 Thoughts?

josevalim:
Oops, pressed the wrong button. :)

qcam:
Yeah that's right, I'll fix that one as well.

lib/ecto/repo/schema.ex
```diff
@@ -573,7 +574,17 @@ defmodule Ecto.Repo.Schema do
...
 
   defp replace_all_fields!(_kind, schema) do
-    schema.__schema__(:fields)
+    Enum.map(schema.__schema__(:fields), &field_source!(schema, &1))
+  end
+
+  defp field_source!(schema, field) do
+    case schema.__schema__(:field_source, field) do
+      nil ->
+        raise ArgumentError, "unknown field for :on_conflict, got: #{inspect(field)}"
```
qcam: I'm not sure if this is desired but I think it would be helpful in case users do something like:

```ex
MyRepo.insert!(%MySchema{foo: :foo}, on_confict: {:replace, [:unknown]}, ...)
```

Instead of raising on `nil` field (returned by `__schema__(:field_source, field)`) not existing in lower layers, this raises more meaningful error IMO.

josevalim:

PR #2809, opened 2018-11-11
Update couple Filtering Associations examples

sysashi:
I think you will have to follow binding in your queries (although dynamic query expression is the new thing for me).

Usually `c` would be used instead of `comment` for binding, not sure which one is correct here.


josevalim:

PR #2816, opened 2018-11-14
Support passing Ecto.Query in Repo.insert_all

qcam:
Part of https://github.com/elixir-ecto/ecto_sql/pull/39.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -78,29 +78,67 @@ defmodule Ecto.Repo.Schema do
   defp extract_header_and_fields(rows, schema, dumper, autogen_id, adapter) do
     mapper = init_mapper(schema, dumper, adapter)
 
-    Enum.map_reduce(rows, %{}, fn fields, header ->
-      {fields, header} = Enum.map_reduce(fields, header, mapper)
-      autogenerate_id(autogen_id, fields, header, adapter)
-    end)
+    {rows, {header, has_query?}} =
+      Enum.map_reduce(rows, {%{}, false}, fn fields, acc ->
+        {fields, {header, has_query?}} = Enum.map_reduce(fields, acc, mapper)
+        {fields, header} = autogenerate_id(autogen_id, fields, header, adapter)
+        {fields, {header, has_query?}}
+      end)
+
+    if has_query? do
+      rows = plan_query_in_rows(rows, header, adapter)
+      {rows, header}
+    else
+      {rows, header}
+    end
   end
 
   defp init_mapper(nil, _dumper, _adapter) do
-    fn {field, _} = tuple, acc ->
-      {tuple, Map.put(acc, field, true)}
+    fn {field, _} = tuple, {header, has_query?} ->
+      {tuple, {Map.put(header, field, true), has_query?}}
     end
   end
 
   defp init_mapper(schema, dumper, adapter) do
-    fn {field, value}, acc ->
-      case dumper do
-        %{^field => {source, type}} ->
-          value = dump_field!(:insert_all, schema, field, type, value, adapter)
-          {{source, value}, Map.put(acc, source, true)}
-        %{} ->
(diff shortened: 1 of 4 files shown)


lib/ecto/repo/schema.ex
```diff
@@ -78,31 +78,69 @@ defmodule Ecto.Repo.Schema do
   defp extract_header_and_fields(rows, schema, dumper, autogen_id, adapter) do
     mapper = init_mapper(schema, dumper, adapter)
 
-    Enum.map_reduce(rows, %{}, fn fields, header ->
-      {fields, header} = Enum.map_reduce(fields, header, mapper)
```
qcam: @josevalim 

> What is a map? :) I think we normalize all of the headears and fields before sending them to the adapter, right?

Sorry for not being clear. I meant the `header` in this context. :)

The problem I encountered was that since we don't process items in each row the same order as when we [generate SQL](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/sql.ex#L509-L518), the query params counter in this step might not be in the same as when the adapter processes the params.

_I might be missing something very obvious (which happens very usually) so please bear with me._ :)

lib/ecto/repo/schema.ex
```diff
@@ -53,7 +53,7 @@ defmodule Ecto.Repo.Schema do
       on_conflict(on_conflict, conflict_target, schema_meta, counter, adapter)
 
     {count, rows} =
-      adapter.insert_all(adapter_meta, schema_meta, Map.keys(header), rows, on_conflict, return_sources, opts)
+      adapter.insert_all(adapter_meta, schema_meta, header, rows, on_conflict, return_sources, opts)
```
josevalim: Btw, we can't do this change, as it would be an adapter breaking change.

josevalim: Oh, ignore me. `header` is a different data structure now. :)

lib/ecto/repo/schema.ex
josevalim: It is expected that adapters will process the parameters in the order given in headers. We can even explicitly document it. Would it address this issue? :)

josevalim: Here is a suggestion to try to simplify this:

1. First we do a traversal, exactly like we do today, although we also check if any of the fields contain an ecto query

2. If any of the fields contain an ecto query, then we do another traversal. Except that on this new traversal we already have a list of headers and therefore we have the expected ordering. So now we can build the queries with the relevant counters.

This should make the common case as efficient as it is today and it should simplify the work we have to do when we do have a query. Thoughts?

qcam: Oh that indeed sounds better, I will give that a try later today 👍. Thanks for the help!

qcam:
@josevalim Could you help to give it another round of review?

lib/ecto/repo/schema.ex
```diff
@@ -78,31 +78,69 @@ defmodule Ecto.Repo.Schema do
...
+  defp plan_query_in_rows(rows, header, adapter) do
+    {rows, _counter} =
+      Enum.map_reduce(rows, 0, fn fields, counter ->
+        Enum.flat_map_reduce(header, counter, fn {key, _}, counter ->
+          case :lists.keyfind(key, 1, fields) do
+            {^key, %Ecto.Query{} = query} ->
+              {query, params, _} = Ecto.Query.Planner.plan(query, :all, adapter, counter)
+              {query, _} = Ecto.Query.Planner.normalize(query, :all, adapter, counter)
+
+              {[{key, {query, params}}], counter + length(params)}
```
josevalim: So the adapter will receive `{Ecto.Query, params}`, correct?

qcam: Yes you're right. Do you think we need to update the documentation somewhere?

josevalim: We should update the documentation for the insert_all @callback in the adapter. :)

josevalim:

Issue #2820, opened 2018-11-17
Ecto.Repo.Supervisor does not work with new-style child specs


evadne:
* Elixir version (elixir -v): 1.7.4
* Ecto version (mix deps): 3.x
* Operating system: macOS

### Current behavior

When creating a new adapter, this works:

```
defmodule Adapter do
  @behaviour Ecto.Adapter
  def init(config) do
    child_spec = Supervisor.Spec.worker(Worker, [])
    adapter_meta = %{}
    {:ok, child_spec, adapter_meta}
  end
end
```

This does not:

```
defmodule Adapter do
  @behaviour Ecto.Adapter
  def init(config) do
    child_spec = Worker.child_spec([])
    adapter_meta = %{}
    {:ok, child_spec, adapter_meta}
  end
end
```

I think this is due to `Ecto.Repo.Supervisor` calling `supervise/2` (old behaviour).

### Expected behaviour

Would it be worth considering whether to support the new style?

josevalim:
Definitely. We should call Supervisor.init/2 if available and fallback to
Supervisor.Spec.supervise/2 if not, as we still support older Elixir
versions.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
Thanks, new version released!

PR #2821, opened 2018-11-17
modifies Ecto.Repo.Supervisor to support new style child spec

evadne:
As per discussion in #2820, amends `Ecto.Repo.Supervisor` so once it has a wrapped child spec, it tries to use `Supervisor.init/2` instead of `Supervisor.Spec.supervise/2`.

The decision is currently implemented as an inlined function with the underlying call determined at compile time. I can change this so it happens at runtime if that is better.

### DIFF
--- a/lib/ecto/repo/supervisor.ex
+++ b/lib/ecto/repo/supervisor.ex
@@ -188,45 +188,33 @@ defmodule Ecto.Repo.Supervisor do
       {:ok, opts} ->
         {:ok, child, meta} = adapter.init([repo: repo] ++ opts)
         cache = Ecto.Query.Planner.new_query_cache(name)
-        child = wrap_start(child, [adapter, cache, meta])
-        perform_start(child)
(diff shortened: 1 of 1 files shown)


josevalim:
Let’s go with runtime because we don’t have to worry about performance here
and it will be cleaner. Thanks :)
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
We probably need a test too. :) Maybe we can set up a simple adapter that
returns a Task as child spec? Thoughts?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

evadne:
Thanks, I will make the changes. Of the 2 test environments on Travis, the older one is wonky and crashing when booting Elixir, not too sure why :(

josevalim:
Looks great! I have removed some of the added white space and reverted some of the style changes. Just a test and we are ready to ship it. :)

josevalim:

Issue #2822, opened 2018-11-18
Ecto.Repo.Supervisor.compile_config/2 should use ensure_compiled? instead of ensure_loaded?


evadne:
* Elixir version (elixir -v): 1.7.4
* Ecto version (mix deps): 3.x
* Operating system: macOS

### Current behaviour

Currently `Ecto.Repo.Supervisor.compile_config/2` uses `Code.ensure_loaded?/1` which is fine in most cases, when the adapter lives elsewhere, and would possibly have been compiled already as part of a dependency. But this can be wobbly when both the Repo and the Adapter it refers to are in the same OTP app. When I do this and repeatedly run `mix compile` on a machine with 8 cores, this fails 50% of the time but never fails during the first compilation pass after running `mix clean`.

Upon consulting [Elixir Documentation on code loading](https://hexdocs.pm/elixir/Code.html#ensure_loaded/1-code-loading-on-the-erlang-vm) I believe this call should be changed. Once this is changed the wobbliness is no more.

PR #2823, opened 2018-11-18
fixes race condition Ecto.Repo.Supervisor if adapter needs compiling

evadne:
- Follow-up of #2822

- Previously, `Ecto.Repo.Supervisor ` uses `Core.ensure_loaded?/1` which could cause errors when the adapter is not yet compiled.

- See https://hexdocs.pm/elixir/Code.html#ensure_loaded/1-code-loading-on-the-erlang-vm

  > When invoked, ensure_compiled/1 halts the compilation of the caller until the module given to `ensure_compiled/1` becomes available or all files for the current project have been compiled. If compilation finishes and the module is not available, an error tuple is returned.


josevalim:

Issue #2825, opened 2018-11-18
Ecto 3: union/union_all + order_by gives "(undefined_table)" Postgres error


eumend:
### Environment

* Elixir version (elixir -v): 1.7.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): psql (PostgreSQL) 10.6 (Ubuntu 10.6-0ubuntu0.18.04.1)
* Ecto version (mix deps):
{:ecto_sql, "~> 3.0.1"}
{:ecto, "~> 3.0.2"},
* Database adapter and version (mix deps):
{:postgrex, ">= 0.0.0"}
* Operating system:
Ubuntu 18.04

### Current behavior

I'm using Ecto 3's `union_all` in order to union and paginate multiple tables, sorted by timestamps, with a single query. It works quite well (thanks for this feature!) until I add the `order_by` clause:

```
def trades_query() do
    Trades.Trade
  end

  def orders_query() do
    Orders.Order
  end

  def big_select() do
    trades_select =
      trades_query()
      |> select([t], %{
        tag: "trade",
        timestamp: t.inserted_at,
        data: fragment("row_to_json(?)", t)
      })

    orders_select =
      orders_query()
      |> select([o], %{
        tag: "order",
        timestamp: o.inserted_at,
        data: fragment("row_to_json(?)", o)
      })

    trades_select
    |> union_all(^orders_select)
    # This breaks, as does order_by(:inserted_at)
    |> order_by(:timestamp)
    |> limit(50)
    |> offset(30)
  end
```

It gives the following error

```
** (Postgrex.Error) ERROR 42P01 (undefined_table) missing FROM-clause entry for table "t0"

    query: SELECT t0."inserted_at", row_to_json(t0) FROM "trades" AS t0 UNION ALL (SELECT o0."inserted_at", row_to_json(o0) FROM "orders" AS o0) ORDER BY t0."timestamp" LIMIT 50 OFFSET 30
    (ecto_sql) lib/ecto/adapters/sql.ex:590: Ecto.Adapters.SQL.raise_sql_call_error/1
    (ecto_sql) lib/ecto/adapters/sql.ex:523: Ecto.Adapters.SQL.execute/5
    (ecto) lib/ecto/repo/queryable.ex:147: Ecto.Repo.Queryable.execute/4
    (ecto) lib/ecto/repo/queryable.ex:18: Ecto.Repo.Queryable.all/3
```

[...]

PR #2828, opened 2018-11-19
Ensure no bindings when using combinations in Ecto.Query

qcam:
Spawned from https://github.com/elixir-ecto/ecto/issues/2825#issuecomment-439817232.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -771,7 +771,7 @@ defmodule Ecto.Query.Planner do
     case operation do
       :all ->
         assert_no_update!(query, operation)
-        assert_valid_combinations!(query, operation)
+        assert_valid_combinations!(query)
       :update_all ->
         assert_update!(query, operation)
         assert_only_filter_expressions!(query, operation)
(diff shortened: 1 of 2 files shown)


lib/ecto/query/planner.ex
```diff
@@ -1424,6 +1425,19 @@ defmodule Ecto.Query.Planner do
     end
   end
 
+  defp assert_valid_combinations!(query, operation) do
+    if query.combinations != [] do
+      has_bindings? =
+        Enum.any?(query.order_bys, fn order_by ->
+          Enum.any?(order_by.expr, &match?({_dir, {{:., [], [{:&, [], [0]}, :id]}, [], []}}, &1))
```
```suggestion
          Enum.any?(order_by.expr, &match?({_dir, {{:., [], [{:&, [], [_]}, _]}, [], []}}, &1))
```

lib/ecto/query/planner.ex
```diff
@@ -1424,6 +1425,19 @@ defmodule Ecto.Query.Planner do
...
+        end)
+
+      if has_bindings? do
+        error! query, "`#{operation}` does not allow bindings in `order_by` when using `union` or `union_all`"
```
josevalim: We also have intersect, intersect_all, except and except_all.

josevalim: If you don't want to list all of them, you can do this:

```elixir
case query.combinations do
  [] -> :ok
  [{kind, _} | _] -> ...
end
```

and then use `kind` in the error message. We also don't need the `operation` here, because this is never valid, regardless of the operation. I would go with something like this:

```
error!, query, """
cannot use bindings in `order_by` when using `#{kind}`.

That's because the `order_by` applies to the whole `#{kind}` and not \
an individual query. If you really want to order the results, you can wrap \
the existing query in a subquery and then order it:

    query = query1 |> #{kind}(query2)
    from q in subquery(query), order_by: q.field

josevalim:

Issue #2830, opened 2018-11-20
Not return the correct binary_id list with ecto 3


oo6:
### Environment

* Elixir version (elixir -v): 1.7.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 10.1
* Ecto version (mix deps): 3.0.2
* Database adapter and version (mix deps): postgrex 0.14.0
* Operating system: macOS Mojave

### Current behavior

```elixir
# lib/sandbox/account/user.ex
defmodule Sandbox.Account.User do
  use Ecto.Schema

  schema "users" do
    field :uuid, :binary_id
    field :uuids, {:array, :binary_id}
  end
end

# priv/repo/migrations/20181120033938_create_users.exs
defmodule Sandbox.Repo.Migrations.CreateUsers do
  use Ecto.Migration

  def change do
    create table(:users) do
      add :uuid, :binary_id
      add :uuids, {:array, :binary_id}
    end
  end
end
```
test code in `iex -S mix`:
```elixir
foo = Sandbox.Repo.insert!(%Sandbox.Account.User{uuid: Ecto.UUID.generate(), uuids: [Ecto.UUID.generate()]})
%Sandbox.Account.User{
  __meta__: #Ecto.Schema.Metadata<:loaded, "users">,
  id: 1,
  uuid: "80780db2-94ea-4216-9538-6f28cc25f018",
  uuids: ["e261cfbf-76da-42cc-a194-3587db155c84"]
}

bar = Sandbox.Repo.get!(Sandbox.Account.User, a.id)
%Sandbox.Account.User{
  __meta__: #Ecto.Schema.Metadata<:loaded, "users">,
  id: 1,
  uuid: "80780db2-94ea-4216-9538-6f28cc25f018",
  uuids: [<<226, 97, 207, 191, 118, 218, 66, 204, 161, 148, 53, 135, 219, 21, 92, 132>>]
}
```

`foo.uuids == bar.uuids` returns `false`

### Expected behavior

`bar.uuids` is the same as `foo.uuids`

Working correctly on 2.x.
I am not sure if this is the right behavior on 3.x, and not found the description in the changelog.

josevalim:
It is a bug. I will push a fix shortly. :)

PR #2831, opened 2018-11-20
fix load binary id array error

smartepsh:
Related issue: #2830 

Like the issue says, Ecto 3.0 can't transform binary_id in an array to the printable string.

### Reason

In Ecto 3.0, the type `:binary_id` is specially handled: 
1) Through adapter.loaders, the type `:binary_id` has becomes to `[Ecto.UUID, :bianry_id]` 
2) Then load twice for each `Ecto.UUID` and `:binary_id`

The result is correct because it calls the function `Ecto.UUID.load/1` during the first load.  But `{:array, :binary_id}` is not specially handled. The function `dump_binary/1` just returns what it receives.

### Solution

Like Ecto 2.x, pattern match the `{:array, :binary_id}` and handle the value of each of the array.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -496,6 +496,10 @@ defmodule Ecto.Type do
     {:ok, nil}
   end
 
+  def load({:array, :binary_id}, value, loader) do
+    array(value, &loader.(:binary_id, &1), [])
+  end
+
   def load(type, value, _loader) do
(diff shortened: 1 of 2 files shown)


josevalim:
Sorry, I did not see you were working on it. I have already pushed a fix to master. :)

kenken1mura:
delet

Issue #2835, opened 2018-11-22
Ecto.Query.API.type/2 with dynamic field


LostKobrakai:
### Environment

* Elixir version (elixir -v): 1.7.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): postgres 10
* Ecto version (mix deps): 3.0.3 (3.0.2 ecto_sql)
* Database adapter and version (mix deps): postgrex 0.14.0
* Operating system:

### Current behavior

Follow up to https://github.com/elixir-ecto/ecto/issues/2485#issuecomment-440874401

I've added a few failing tests: 
https://github.com/LostKobrakai/ecto_sql/commit/a9e22c1296065ac809c9e2b341b56e3aa96061c3

The current error message `** (Ecto.Query.CompileError) type/2 expects an alias, atom or source.field as second argument, got: field(x, ^field)` would suggest my test examples should work, as the type is defined via an alias. But using an alias errors with 

```
** (ArgumentError) you attempted to apply :type on {:__aliases__, [line: 101], [:ConnectAuxiliary, :Ecto, :Money_Composite]}. If you are using apply/3, make sure the module is an atom. If you are using the dot syntax, such as map.field or module.function, make sure the left side of the dot is an atom or a map
    :erlang.apply({:__aliases__, [line: 101], [:ConnectAuxiliary, :Ecto, :Money_Composite]}, :type, [])
    lib/ecto/type.ex:263: Ecto.Type.match?/2
    lib/ecto/query/builder.ex:546: Ecto.Query.Builder.assert_type!/3
    lib/ecto/query/builder.ex:360: Ecto.Query.Builder.escape/5
    lib/ecto/query/builder.ex:407: Ecto.Query.Builder.escape_with_type/5
    lib/ecto/query/builder/select.ex:104: anonymous fn/4 in Ecto.Query.Builder.Select.escape_pairs/4
    (elixir) lib/enum.ex:1418: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    lib/ecto/query/builder/select.ex:75: Ecto.Query.Builder.Select.escape/4
```

I'm not sure if `select: type(sum(field(x, ^name)), field(x, ^name))` is supposed to work. Having aliases work would be enough for my case for now.

### Expected behavior

Have some way to not need to hardcode a column name to be able to type things in an ecto query.

josevalim:
Closing in favor of the PR.

PR #2836, opened 2018-11-22
add support for field/2 as first element of type/2 and alias as second element of type/2

LostKobrakai:
I'm not sure if `validate_type!` is the correct place to expand the alias to a module atom, but it seems to work and not blow up any tests in ecto/ecto_sql

https://github.com/elixir-ecto/ecto/issues/2835

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -409,7 +409,7 @@ defmodule Ecto.Query.Builder do
   defp escape_with_type(expr, type, params_acc, vars, env) do
     type = validate_type!(type, vars, env)
     {expr, params_acc} = escape(expr, type, params_acc, vars, env)
-    {{:{}, [], [:type, [], [expr, type]]}, params_acc}
+    {{:{}, [], [:type, [], [expr, Macro.escape(type)]]}, params_acc}
   end
 
   defp wrap_nil(params, {:{}, _, [:^, _, [ix]]}), do: wrap_nil(params, ix, [])
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -85,8 +85,8 @@ defmodule Ecto.Query.Builder do
   end
 
   # tagged types
-  def escape({:type, _, [{:^, _, [arg]}, type]}, _type, {params, acc}, vars, _env) do
-    type = validate_type!(type, vars)
+  def escape({:type, _, [{:^, _, [arg]}, type]}, _type, {params, acc}, vars, env) do
+    type = validate_type!(type, vars, env)
     expr = {:{}, [], [:type, [], [{:{}, [], [:^, [], [length(params)]]}, type]]}
     params = [{arg, type} | params]
     {expr, {params, acc}}
@@ -103,6 +103,10 @@ defmodule Ecto.Query.Builder do
   end
 
   def escape({:type, _, [{:fragment, _, [_ | _]} = expr, type]}, _type, params_acc, vars, env) do
+    escape_with_type(expr, type, params_acc, vars, env)
+  end
+
+  def escape({:type, _, [{:field, _, [_ | _]} = expr, type]}, _type, params_acc, vars, env) do
     escape_with_type(expr, type, params_acc, vars, env)
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks! We need to add tests to builder_test.exs or planner_test.exs.

lib/ecto/query/builder.ex
```diff
@@ -555,23 +559,23 @@ defmodule Ecto.Query.Builder do
...
+    {composite, validate_type!(type, vars, env)}
   end
-  def validate_type!({:^, _, [type]}, _vars),
-    do: type
-  def validate_type!({:__aliases__, _, _} = type, _vars),
+  def validate_type!({:^, _, [type]}, _vars, _env),
     do: type
-  def validate_type!(type, _vars) when is_atom(type),
+  def validate_type!({:__aliases__, _, _} = type, _vars, {env, _}),
+    do: Macro.expand(type, env)
```
josevalim: Why do we need to expand the alias? It is fine to return it as an alias as it will be expanded by the compiler.

lib/ecto/query/builder.ex
```diff
@@ -106,6 +106,10 @@ defmodule Ecto.Query.Builder do
     escape_with_type(expr, type, params_acc, vars, env)
   end
 
+  def escape({:type, _, [{:field, _, [_ | _]} = expr, type]}, _type, params_acc, vars, env) do
+    escape_with_type(expr, type, params_acc, vars, env)
+  end
```
josevalim: It feels like this is the only change you need. The way `validate_type!` worked with aliases should be fine. Or are you running into something?

lib/ecto/query/builder.ex
LostKobrakai: Without expanding the alias it didn't work trying to call `.type()` on the AST instead of on the module.

LostKobrakai:
> Thanks! We need to add tests to builder_test.exs or planner_test.exs.

Yeah, I wasn't sure where/how to add them. I've used those in ecto_sql: https://github.com/LostKobrakai/ecto_sql/commit/a9e22c1296065ac809c9e2b341b56e3aa96061c3

lib/ecto/query/builder.ex
josevalim: Ah, I see. Good call then!

josevalim:
@LostKobrakai yeah. We should be able to write those as unit tests. One of the flaws we had from having both ecto+ecto_sql together is that we relied extensively in integration tests and now we should improve our coverage back.

LostKobrakai:
I've added unit tests for the things I added.

test/ecto/query/builder_test.exs
```diff
@@ -27,6 +27,9 @@ defmodule Ecto.Query.BuilderTest do
     assert {Macro.escape(quote do type(&0.y + &1.z, :decimal) end), []} ==
            escape(quote do type(x.y + y.z, :decimal) end, [x: 0, y: 1], __ENV__)
 
+    assert {Macro.escape(quote do type(&0.y, :decimal) end), []} ==
+          escape(quote do type(field(x, :y), :decimal) end, [x: 0], __ENV__)
```
josevalim: Shouldn't we test this with an alias? I think we have a Custom.Permalink type thingy.

LostKobrakai: I can add this as well, but even this didn't work previously.

josevalim: Let's add it because otherwise I can remove lines of code from your commit and all tests would still pass. :)

josevalim: Ping. :)

josevalim: Ping? :)

LostKobrakai: Sorry for the delay. I've not much time at hand at the moment and I was getting into trouble where the coe would return an alias in AST format, but `Macro.escape` in the test would either escape the AST or have the module name as atom.

josevalim:
Ping!

LostKobrakai:
So I got around my issue with my test, but I'm not sure if it's the right way to solve it. Especially as it seem to have worked without the change for my other tests (and me using this branch in prod for a few weeks already).

josevalim:

PR #2837, opened 2018-11-22
Add `force?` option to Ecto.Changeset.cast/4

qcam:
I would like to propose `force?` option in `Ecto.Changeset.cast/4`.

IMO this option would be very helpful to deal with default field value in upsert. Let's considering the cases below:

```ex
defmodule Post do
  use Ecto.Schema

  schema "posts" do
    field :title, :string
    field :published, :boolean, default: true
  end
end

import Ecto.Changeset

params = %{title: "foo", published: true}
changeset = cast(%Post{}, params, [:title, :published])
```

In the example above, `published` in params isn't casted into `changeset.changes` because there is no change for `published` (same as default value).

This in a bit inconvenient in a way that when we use "Upserts" feature with an `Ecto.Query`, like:

```ex
params = %{title: "foo", published: true}
changeset = cast(%Post{}, params, [:title, :published])
set_values = Keyword.new(changeset.changes)

options = [
  conflict_target: [:id],
  on_conflict: Post |> update(set: ^set_values) |> where([p], p.foo_condition...)
]
```

In this case, only `title` is sent to the database because `published` is not available in `Ecto.Changeset.changes`, as described above. To ensure every attribute in `params` ends up in the database, we need to iterate and use `force_change`.

This could be less obvious to notice when we have `nil` as the default value, for example `title` in `%{title: nil}` will also be ignored in the code snippet above.

I hope `:force?` options can make this scenario like this easier to handle. Thank you.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -420,6 +420,8 @@ defmodule Ecto.Changeset do
 
     * `:empty_values` - a list of values to be considered as empty when casting.
       Defaults to the changeset value, which defaults to `[""]`
+    * `:force?` - a boolean flag to enforce putting casted value to changes even
+      if the field has not been changed. Defaults to `false`.
 
   ## Examples
 
@@ -490,6 +492,7 @@ defmodule Ecto.Changeset do
 
   defp cast(%{} = data, %{} = types, %{} = changes, %{} = params, permitted, opts) when is_list(permitted) do
     {empty_values, _opts} = Keyword.pop(opts, :empty_values, @empty_values)
+    {force?, _opts} = Keyword.pop(opts, :force?, false)
     params = convert_params(params)
 
     defaults = case data do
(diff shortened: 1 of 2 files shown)


josevalim:
Can you use `force_change` instead?

qcam:
No, I don't think I can. `cast` gives some better benefits than `force_change` with casting and type checking.

I imagine `cast` combining with `force_change` like this snippet could help to achieve similar thing:

```elixir
changeset = cast(%Post{}, params, permitted_fields)

changeset =
  Enum.reduce(fields_we_need, changeset, fn field, changeset ->
    if Map.has_key?(changeset.changes, field) do
      changeset
    else
      force_change(changeset, field, Map.fetch!(changeset.data, field))
    end
  end)
```

But having an option to force change casted data right at `cast/4` would help doing this much easier IMO.

josevalim:
So for now I will propose for you to go with a custom helper as above and if this feature feels useful to others, they can chime in, so we gather more cases before adding it to Ecto forever and ever. What do you think? :)

qcam:
👍Yes that works. Thanks for considering this.

PR #2842, opened 2018-11-26
Add association typespecs to Schema

yordis:
I keep copying this all over the place where I have Ecto, so I figured that maybe some people will find it useful.

This is an example of the usage

```elixir
@type t :: %__MODULE__{
          address: Ecto.Schema.belongs_to(Address.t())
        }
```

It will help to have a better typespec, I notice that some folks do not add `| Ecto.Association.NotLoaded.t()` which is the actual type of the field.


michalmuskala:
Thank you! :heart:

PR #2849, opened 2018-12-03
Properly match keyword lists on query planner

leandrocp:
Fixes an error that happens when you try to query a list of custom ecto type which is defined as a tuple, but it's not a kw list. Eg:

```elixir
locations = [{"europe", "france"}, {"south_america", "brazil"}]
from(c in City, where: [location: locations])
```

In that case, `location` would be a custom ecto type that dumps to string, jsonb, etc. in the DB.

Let me know if that makes sense or needs more info about that use case.


leandrocp:
Wow that was fast! 😀
Thanks @josevalim and also @lucas-cegatti for pointing me this error.

Issue #2848, opened 2018-12-03
cast utc_datetime_usec: dropping microseconds for non-utc datetimes


hauptbenutzer:
### Environment

* Elixir version (elixir -v): Elixir 1.7.4 (compiled with Erlang/OTP 21)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): doesn't matter
* Ecto version (mix deps): 2.x and 3.x
* Operating system: macOS 10.14.1 (18B75)

### Current behavior

I stumbled upon the following behaviour using Ecto 2.2: 

```elixir
iex(1)> berlin = "2018-12-03 00:00:00.000001" |> NaiveDateTime.from_iso8601!() |> Timex.to_datetime("Europe/Berlin")
#DateTime<2018-12-03 00:00:00.000001+01:00 CET Europe/Berlin>
iex(2)> utc = "2018-12-02 23:00:00.000001" |> NaiveDateTime.from_iso8601!() |> Timex.to_datetime("Etc/UTC")
#DateTime<2018-12-02 23:00:00.000001Z>
iex(3)> Ecto.Type.cast(:utc_datetime, berlin)
{:ok, #DateTime<2018-12-02 23:00:00Z>}
iex(4)> Ecto.Type.cast(:utc_datetime, utc)
{:ok, #DateTime<2018-12-02 23:00:00.000001Z>}
```

Nevermind the Timex conversion, the important bit is, that the microseconds are lost when casting a non-UTC DateTime struct. I then went investigating and found that this stems from https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L936 , where microseconds are ignored. This function is also used in Ecto 3 for casting `utc_datetime_usec`, so the problem persists.

### Expected behavior

I would expect either 

* Ecto to error when casting a non-UTC DateTime struct (as it seems to do for `load` and `dump`)
* Or, preferably, keep the microseconds on cast

I've wrote some failing tests and added a fix proposition over here: https://github.com/hauptbenutzer/ecto/commits/fix/utc_datetime_usec_timezone 
I'd gladly open a PR for this, if desired. 

Cheers

wojtekmach:
I tried this on Ecto master and here are results:

```elixir
iex> berlin = %{DateTime.from_naive!(~N[2018-12-03 00:00:00.000001], "Etc/UTC") | zone_abbr: "CET", time_zone: "Europe/Berlin", utc_offset: 3600}
[...]

PR #2850, opened 2018-12-03
fix utc_datetime_usec cast with timezone

hauptbenutzer:
As discussed in https://github.com/elixir-ecto/ecto/issues/2848, the cast for `utc_datetime_usec` currently uses 

```elixir
datetime |> DateTime.to_unix() |> DateTime.from_unix()
```

for casting non-UTC DateTime structs to UTC. In this process, the microsecond data on the struct is lost. We can fix this by explicitly using microsecond precision when converting to/from unix timestamp.

### DIFF
--- a/test/ecto/type_test.exs
+++ b/test/ecto/type_test.exs
@@ -572,6 +572,20 @@ defmodule Ecto.TypeTest do
   @datetime_zero DateTime.from_unix!(1422057000, :second)
   @datetime_zero_usec DateTime.from_unix!(1422057000000000, :microsecond)
   @datetime_usec DateTime.from_unix!(1422057007008000, :microsecond)
+  @datetime_usec_tz %DateTime{
+    calendar: Calendar.ISO,
+    day: 24,
+    hour: 0,
+    microsecond: {8000, 6},
+    minute: 50,
+    month: 1,
+    second: 7,
+    std_offset: 0,
+    time_zone: "Europe/Berlin",
+    utc_offset: 3600,
+    year: 2015,
+    zone_abbr: "CET"
+  }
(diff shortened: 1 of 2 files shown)


test/ecto/type_test.exs
```diff
@@ -698,6 +713,14 @@ defmodule Ecto.TypeTest do
       }
       assert Ecto.Type.cast(:utc_datetime_usec, term) == {:ok, @datetime_zero_usec}
 
+      term = %DateTime{
+        calendar: Calendar.ISO, year: 2015, month: 1, day: 24,
+        hour: 9, minute: 50, second: 7, microsecond: {8000, 6},
+        std_offset: 0, utc_offset: 36000,
+        time_zone: "Etc/GMT-10", zone_abbr: "+10"
+      }
+      assert Ecto.Type.cast(:utc_datetime_usec, term) == {:ok, @datetime_usec}
```
hauptbenutzer: Both of these added tests basically check the same thing. I was unsure where to add them, as the test description further above says "cast from map", and yet in https://github.com/elixir-ecto/ecto/pull/2850/files#diff-af151a6ff25be431d4b33d60b963bffdR708 we have a DateTime struct. 

I guess either one will do? Maybe keep this one and get rid of the other, as the struct's a bit ugly as a module variable?

wojtekmach: I personally think it's fine either way and as is too. I slightly lean towards the other example as it's shorter. We could also make the module attribute less ugly by something like this:

```elixir
%{DateTime.from_naive!(~N[2018-12-03 00:00:00.000001], "Etc/UTC") | zone_abbr: "CET", time_zone: "Europe/Berlin", utc_offset: 3600}
```

but on the other hand it's ok to be super explicit in a test.

PR #2852, opened 2018-12-05
Allow specifying a query for associations

coop:
The [documentation suggests that you are allowed to specify a query instead of a schema](https://hexdocs.pm/ecto/Ecto.Schema.html#belongs_to/3-using-queries-as-associations) - when running this locally I always hit:

```elixir
def related_from_query(queryable, name) do
  raise ArgumentError, "association #{inspect name} queryable must be a schema or " <>
    "a {source, schema}. got: #{inspect queryable}"
end
```

I have changed the code to support the documentation. I tried writing a test but I couldn't get the integration tests to run:

```
$ mix test integration_test/cases/assoc.exs

== Compilation error in file integration_test/cases/assoc.exs ==
** (CompileError) integration_test/cases/assoc.exs:4: module Ecto.Integration.Case is not loaded and could not be found
    (elixir) expanding macro: Kernel.use/2
    integration_test/cases/assoc.exs:4: Ecto.Integration.AssocTest (module)
```

I'm probably doing something wrong - if someone could help me get the tests running I'll add a test showing the functionality seems to be currently broken.

Thanks.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -295,6 +295,7 @@ defmodule Ecto.Association do
   """
   def related_from_query(atom, _name) when is_atom(atom), do: atom
   def related_from_query({source, schema}, _name) when is_binary(source) and is_atom(schema), do: schema
+  def related_from_query(%Ecto.Query{} = queryable, _name), do: queryable
   def related_from_query(queryable, name) do
     raise ArgumentError, "association #{inspect name} queryable must be a schema or " <>
       "a {source, schema}. got: #{inspect queryable}"


coop:
Sorry for closing - I accidentally submitted without a description.

coop:
I'm guessing this isn't the full solution because I am getting

>Compilation failed because of a deadlock between files.
>The following files depended on the following modules:

errors when compiling my project (only after `rm -rf _build`).

A simplified version of my schemas:

```elixir
defmodule Post do
  schema "posts" do
    has_many :comments, Comment
    has_one :latest_comment, from(c in Comment, order_by: [desc: c.id])
  end
end

defmodule Comment do
  schema "comments" do
    belongs_to :post, Post
  end
end
```

I have a feeling the `has_one` is causing the deadlock but if I comment out `belongs_to :post, Post` and replace it with `field :post_id, :integer` I can also get around the problem.

josevalim:
We have recently tried to add this and ran into many issues, so we settled on supporting some keys individually. For example, today you can pass `where` to your associations. You can see the implementation for that feature here: https://github.com/elixir-ecto/ecto/pull/2520

coop:
I saw `where` and while it's a great addition it doesn't solve my problem. Would you considering a pull request if I added `order_by`?

josevalim:
Sure. Just keep in mind is a relatively high amount of work. Here is the first PR, which allowed you to pass `where` via query, before we changed it to pass `where` as a separate option: https://github.com/elixir-ecto/ecto/pull/2472/files

coop:
Maybe it would be easier is _someone else_ implemented it lol. I'm happy to give it a go because I need this functionality. Also `has_one` seems to have implicit ordering right now - which pg it seems to return the first result. I'm not sure if that is true on all databases though.

The alternative would be a virtual field on `Post` and I populate with a second query - it's not terrible but it's a bit of work that I'd prefer to not have to do in several spots.

josevalim:
Actually PG has no ordering guarantee, so it may be the first result but it is not guarenteed. MySQL orders by ID, IIRC.

PR #2853, opened 2018-12-05
Add support for has_many/has_one order_by

coop:
I originally opened https://github.com/elixir-ecto/ecto/pull/2852 in an effort to add support for allowing queries as associations (it was mentioned in the docs but not in code). After a quick discussion I realised that the problem was much larger than I thought. In https://github.com/elixir-ecto/ecto/pull/2852 I mentioned that I could also solve my problem if I was allowed to `order_by` when defining a `has_one` / `has_many` association - this PR is the code for that.

I have only looked at this from a querying point of view, specifically preloads and `Ecto.assoc`. I started looking through the code for building schemas via `Ecto.assoc` but realised it's probably not necessary (not like `where`). I was surprised by how simple the implementation was (based on https://github.com/elixir-ecto/ecto/pull/2852#issuecomment-444403862) which leads me to believe I haven't covered all the required cases to get this merged.

I'm also missing tests but I wanted to get an idea if I was on the right path before proceeding much further.

Any feedback would be appreciated.

Thanks.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -232,47 +232,27 @@ defmodule Ecto.Association do
     do: combine_assoc_query(assoc, queryable)
 
   def combine_assoc_query(assoc, queryable) do
-    do_combine_assoc_query(Map.to_list(assoc), queryable)
+    queryable
+    |> where_assoc_query(assoc)
+    |> order_by_assoc_query(assoc)
   end
 
-  def do_combine_assoc_query([], queryable), do: queryable
+  def where_assoc_query(queryable, %{where: nil}), do: queryable
+  def where_assoc_query(queryable, %{where: []}), do: queryable
 
-  def do_combine_assoc_query([{:where, nil} | rest], queryable) do
-    do_combine_assoc_query(rest, queryable)
-  end
-
-  def do_combine_assoc_query([{:where, []} | rest], queryable) do
-    do_combine_assoc_query(rest, queryable)
-  end
-
-  def do_combine_assoc_query([{:where, {module, function, args}} | rest], queryable) do
+  def where_assoc_query(queryable, %{where: {module, function, args}}) do
     where = apply(module, function, args)
-    queryable = Ecto.Query.where(queryable, _, ^where)
-
-    do_combine_assoc_query(rest, queryable)
+    Ecto.Query.where(queryable, _, ^where)
   end
 
-  def do_combine_assoc_query([{:where, where} | rest], queryable) do
-    queryable = Ecto.Query.where(queryable, _, ^where)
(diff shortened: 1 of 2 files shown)


josevalim:
@coop I don't think this is the correct way to go. You probably want to change `combine_assoc_query` instead. The `where` in the lines you are changing are not related to the "association default where" but rather the clauses from when joining the associations.

coop:
@josevalim I pushed a change that moves `order_by` to `combine_assoc_query`. I wasn't sure how best to pattern match against `where: nil or [] and order_by: nil or []` so I turned the map into a list and process the list recursively. I don't know if that is the best way to solve it.

To clean it up a little bit I could store the "allowable filters" in a module variable and only iterate through them. This would also allow me to remove the duplication when checking if the options are `[]` or `nil`:

```elixir
defp do_combine_assoc_query([{x, nil} | rest], q) when x in @allowable_filters do
   # ...
end
```

josevalim:
@coop i would split it into separate functions and call them all, something like: `query |> where_assoc_query |> order_by_assoc_query`.

josevalim:
Just a heads up, this is not going to work without major changes to the planner. There is a bug in how `where` works that will require us to reduce its scope a bit. I will write more information about it soon but I wanted to give you a heads up.

coop:
Thanks for the heads up.

josevalim:
Actually, it will work, but you will have to rebase and pretty much start again. :) The good news is that it should be less work, because there are less features to support.

josevalim:
Hi @coop! I have pushed changes here: https://github.com/elixir-ecto/ecto/commit/776a5238903dc0f7cb6353b49c7c7f07813bffe6

Those changes were necessary because the previous `:where` did not work correctly on "joins queries". For example, if you wrote `from p in Post, join: c in assoc(p, :comments)`, the `where` would apply to the wrong schema, so I had to rework things a bit.

What does it mean to you? It means that your patch no longer applies. I recommend you to start from scratch but you should be able to port your code without much work. Good news are:

1. I have already prepared the new code for your extensions :)

2. You don't have to worry about `joins_query` in your case since I don't think the `order_by` should apply in such situations (although we need to make sure to document those)

3. You no longer needs to support the tuple format `{Mod, :fun, [args]}`. This means applying order_by in your case should be as easy as:

    def combine_assoc_order_by(query, %{order_by: nil}), do: query
    def combine_assoc_order_by(query, %{order_by: order_by}), do: Ecto.Query.order_by(query, _, order_by)

Besides the notes above, please note that you should amend your code to consistenty support `:order_by` in has_many/one, belongs_to and many_to_many (the same places we support where). And then, of course, docs and tests.

That's it, please let me know if you have any questions!

coop:
Thanks!

PR #2855, opened 2018-12-05
Fix support for upserts with :replace without a schema

darinwilson:
Per conversation with @josevalim on the mailing list:

Calling `Repo.insert_all` with `on_conflict: {:replace, fields}` and no schema resulted in this error message:

```
** (UndefinedFunctionError) function nil.__schema__/2 is undefined
```

This was because the code was trying to look up field sources even when `insert_all` was called without a schema. This fix assures that the lookup only happens when a schema is present.


josevalim:

PR #2856, opened 2018-12-06
renaming "conditions" for clarity, so it isn't the same as the  macro

errkk:
I found this example a little confusing, as the condition that gets interpolated in the where clause has the same name as the macro that's used to build it.

Proposing to give different names to these things for clarity, as I was initially wondering how a boolean could be callable! 🤯

```elixir
dynamic = false
# ...
dynamic([p], p.is_public or ^dynamic)
```

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -412,23 +412,23 @@ defmodule Ecto.Query do
   For example, imagine you have a set of conditions you want to
   build your query on:
 
-      dynamic = false
+      conditions = false
 
-      dynamic =
+      conditions =
         if params["is_public"] do
-          dynamic([p], p.is_public or ^dynamic)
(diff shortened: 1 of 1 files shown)


lib/ecto/query.ex
```diff
@@ -439,18 +439,18 @@ defmodule Ecto.Query do
 
   ## `where`, `having` and a `join`'s `on'
 
-  `dynamic` can be interpolated at the root of a `where`, `having` or
+  `conditions` can be interpolated at the root of a `where`, `having` or
```
josevalim: I thin in this case it should be `dynamic` :) Or say "The result of the `dynamic` macro" or similar.

josevalim:
Good call. I have dropped just one comment!

PR #2865, opened 2018-12-14
Remove deprecated example

devonestes:
Now that the `returning` option is deprecated, we should probably remove
it from the documentation.


michalmuskala:
Thank you! :heart:

Issue #2862, opened 2018-12-13
(CaseClauseError) no case clause matching: {:error, :rollback}


beerlington:
### Environment

* Elixir version (elixir -v): 1.7.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6.2
* Ecto version (mix deps): 3.0.5
* Database adapter and version (mix deps): postgrex 0.14.1
* Operating system: Heroku Stack heroku-18

### Current behavior

Hello!

After upgrading to the newest versions of Ecto and Phoenix, we're seeing some errors on production and can't seem to pinpoint exactly what is causing it. The stack trace is:

```
Elixir.CaseClauseError: no case clause matching: {:error, :rollback}
  File "lib/ecto/repo/transaction.ex", line 16, in Ecto.Repo.Transaction.transaction/3
  File "lib/stone/my_context/my_module.ex", line 43, in anonymous fn/4 in Stone.MyModule.my_func1/3
    |> my_func2(arg2)
  File "lib/enum.ex", line 765, in Enum."-each/2-lists^foreach/1-0-"/2
  File "lib/enum.ex", line 765, in Enum.each/2
  File "lib/stone/server_action/worker.ex", line 134, in Stone.ServerAction.Worker.handle_cast/2
    MyModule.my_func1(id, template, opts)
  File "gen_server.erl", line 637, in :gen_server.try_dispatch/4
  File "gen_server.erl", line 711, in :gen_server.handle_msg/6
  File "proc_lib.erl", line 249, in :proc_lib.init_p_do_apply/3
```

I changed the name of the module and function in our app, but it's chaining a bunch of `Multi.run` functions with `Repo.transaction()` at the end:

```ex
  defp my_func2(arg1, arg2) do
    Multi.new()
    |> Multi.run(:name1, fn _, _ ->
      # removed
    end)
    |> Multi.run(:name2, fn _repo, %{name1: name1} ->
      # removed
    end)
    |> Multi.run(:name3, fn _, _ ->
      # removed
    end)
    |> Multi.run(:name4, fn _repo, %{name3: name3, name1: name1} ->
      # removed
    end)
    |> Repo.transaction()
  end
```

### Expected behavior

I'm not sure what the expected behavior is, but I would expect more information about what triggered the rollback.

I'm sort of able to reproduce it by explicitly triggering a rollback with the value `:rollback`, however, we're not doing any explicit rollbacks in our application:

```ex
  def test do
[...]

PR #2867, opened 2018-12-14
Add error clause result of rollback in multi trans

ramortegui:
This PR adds a clause to manage errors on a failing nested rollback.

 - https://github.com/elixir-ecto/ecto/issues/2862

### DIFF
--- a/lib/ecto/repo/transaction.ex
+++ b/lib/ecto/repo/transaction.ex
@@ -16,6 +16,7 @@ defmodule Ecto.Repo.Transaction do
     case Ecto.Multi.__apply__(multi, name, wrap, return) do
       {:ok, values} -> {:ok, values}
       {:error, {key, error_value, values}} -> {:error, key, error_value, values}
+      {:error, error_value} -> {:error, error_value}
     end
   end
 
--- a/test/ecto/multi_test.exs
+++ b/test/ecto/multi_test.exs
@@ -336,6 +336,16 @@ defmodule Ecto.MultiTest do
       assert :ok == data.inside_ok
     end
 
+    test "rollbacks on errors in nested function" do
+      multi =
+        Ecto.Multi.new()
+        |> Ecto.Multi.run(:foo, fn repo, _ ->
+          repo.rollback(:bar)
+        end)
+
+      assert {:error, :bar} = TestRepo.transaction(multi)
+    end
+
     test "does not allow repeated operations" do
       fun = fn _, _ -> {:ok, :ok} end
 


test/ecto/multi_test.exs
```diff
@@ -336,6 +336,16 @@ defmodule Ecto.MultiTest do
...
     end
 
+    test "rollbacks on errors in nested function" do
+      multi =
+        Ecto.Multi.new()
+        |> Ecto.Multi.run(:foo, fn repo, _ ->
+          repo.rollback(:bar)
+        end)
+
+      assert {:error, :bar} = TestRepo.transaction(multi)
```
wojtekmach: should this be the following instead?

```suggestion
      assert {:error, :foo, :bar, %{}} = TestRepo.transaction(multi)
```

this way it's gonna be similar to what happens if any previous multi steps failed.

Moreover, it would return previous steps that did succeed:

```elixir
multi =
  Ecto.Multi.new()
  |> Ecto.Multi.run(:foo, fn repo, _ ->
    {:ok, :good}
  |> Ecto.Multi.run(:bar, fn repo, _ ->
    repo.rollback(:bad)
  end)

{:error, :bar, :bad, %{foo: :good}} = TestRepo.transaction(multi)
```

What do you all think?

josevalim:
Unfortunately I don't think we should allow someone to rollback from inside a multi. The multi is about using explicit flow-control instead of non-local returns (throws). We could maybe catch the `{:error, _}` and raise a nice error message but I don't think we should support it. Thoughts?

michalmuskala:
Yeah, that would make sense too - provide a nice error message telling explicit rollbacks are not supported when using Multi - I'm fine with that.

ramortegui:
Hi guys!  The issue reported seems that the user wasn't doing an explicit `rollback`, but something inside a `Muli.run` was triggering it.  The test is trying to reflect the issue but explicitly.

josevalim:
@ramortegui I agree and they should stop doing that. :) Maybe a better error message can help them spot the error.

ramortegui:
Looking at the source code has been fun. I was trying to introspect the anonymous function to get something from it, but after reading a couple of blogs seems not possible. The latest commit is about raising an exception displaying the operation that doesn't match with the returns from `Multi.__apply__/4`.  The exception is `Ecto.LocalReturnsNotSupportedOnMulti`, but I'm not sure if this needs to be more general.

lib/ecto/repo/transaction.ex
```diff
@@ -16,7 +16,7 @@ defmodule Ecto.Repo.Transaction do
     case Ecto.Multi.__apply__(multi, name, wrap, return) do
       {:ok, values} -> {:ok, values}
       {:error, {key, error_value, values}} -> {:error, key, error_value, values}
-      {:error, error_value} -> {:error, error_value}
+      {:error, operation} -> raise Ecto.LocalReturnsNotSupportedOnMulti, operation
```
josevalim: Perfect, this is exactly it. I don't think we need a custom exception though. I would instead raise a RuntimeError right here:

```elixir
raise "operation #{inspect operation} is manually rolling back, which is not supported Ecto.Multi"
```

Can you please amend? Thanks!

ramortegui: Done! :D

qcam: @ramortegui I think what @josevalim wanted is to drop the custom exception and raise a RuntimeError here :).

PR #2869, opened 2018-12-16
Allow adapter developers to specify builtin custom type overrides so integration test can be reusable.

mjaric:
As per discussion in ecto_sql [issue 62](https://github.com/elixir-ecto/ecto_sql/issues/62)

Since Ecto repo do not use PoolRepo I found this as better naming for module that should help making integration schema reusable.

In adapters `test_helper.exs` I need to implement `Ecto.Integration.AdapterTypes` like in example below

```elixir
defmodule Ecto.Integration.AdapterTypes do
  def uuid, do: Tds.Types.UUID
end
``` 

Then later in `all_test.exs` I can include integration files as we did before

```elixir
ecto = Mix.Project.deps_paths[:ecto]
Code.require_file "#{ecto}/integration_test/cases/assoc.exs", __DIR__
Code.require_file "#{ecto}/integration_test/cases/interval.exs", __DIR__
Code.require_file "#{ecto}/integration_test/cases/joins.exs", __DIR__
Code.require_file "#{ecto}/integration_test/cases/preload.exs", __DIR__
Code.require_file "#{ecto}/integration_test/cases/repo.exs", __DIR__
Code.require_file "#{ecto}/integration_test/cases/type.exs", __DIR__

Code.require_file "../sql/alter.exs", __DIR__
Code.require_file "../sql/lock.exs", __DIR__
Code.require_file "../sql/logging.exs", __DIR__
Code.require_file "../sql/migration.exs", __DIR__
Code.require_file "../sql/migrator.exs", __DIR__
Code.require_file "../sql/sandbox.exs", __DIR__
Code.require_file "../sql/sql.exs", __DIR__
Code.require_file "../sql/stream.exs", __DIR__
Code.require_file "../sql/subquery.exs", __DIR__
Code.require_file "../sql/transaction.exs", __DIR__
```

There is another thing I'd like to add if we can accept this as solution, and it is related to `text` and `ntext` types. 
* `like` is not supported in filter statement for such column (I'm not sure about full text search, but anyways that is something requires writing as a fragment)
* `distinct` is not supported for such column types

To solve above, such columns must be `varchar` or `nvarchar`.

Let me know if this is ok and if I should create another PR to slightly change integration tests for `text` column type

### DIFF
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -6,7 +6,7 @@ Logger.configure(level: :info)
 
 defmodule Ecto.Integration.AdapterTypes do
   # Builtin uuid type is Ecto.UUID, many adapters will use it, but some may use
-  # it own implmentation.
+  # it own implementation.
   def uuid, do: Ecto.UUID
 end
 
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -4,12 +4,6 @@ Mix.shell(Mix.Shell.Process)
 System.put_env("ECTO_EDITOR", "")
 Logger.configure(level: :info)
 
-defmodule Ecto.Integration.AdapterTypes do
-  # Builtin uuid type is Ecto.UUID, many adapters will use it, but some may use
-  # it own implementation.
-  def uuid, do: Ecto.UUID
-end
-
 Code.require_file "support/test_repo.exs", __DIR__
 ExUnit.start()
 
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -4,6 +4,12 @@ Mix.shell(Mix.Shell.Process)
 System.put_env("ECTO_EDITOR", "")
 Logger.configure(level: :info)
 
+defmodule Ecto.Integration.AdapterTypes do
+  # Builtin uuid type is Ecto.UUID, many adapters will use it, but some may use
+  # it own implmentation.
+  def uuid, do: Ecto.UUID
+end
+
 Code.require_file "support/test_repo.exs", __DIR__
 ExUnit.start()
 
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -2,6 +2,7 @@ defmodule Ecto.Integration.Schema do
   defmacro __using__(_) do
     quote do
       use Ecto.Schema
+      alias Ecto.Integration.AdapterTypes
       type =
         Application.get_env(:ecto, :primary_key_type) ||
         raise ":primary_key_type not set in :ecto application"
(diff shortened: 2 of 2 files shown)


josevalim:
> Since Ecto repo do not use PoolRepo I found this as better naming for module that should help making integration schema reusable.

Oh, I am sorry. So let's move it this configuration and everything else to the `TestRepo`, could it be?

test/test_helper.exs
```diff
@@ -4,6 +4,12 @@ Mix.shell(Mix.Shell.Process)
 System.put_env("ECTO_EDITOR", "")
 Logger.configure(level: :info)
 
+defmodule Ecto.Integration.AdapterTypes do
+  # Builtin uuid type is Ecto.UUID, many adapters will use it, but some may use
+  # it own implementation.
+  def uuid, do: Ecto.UUID
+end
```
josevalim: I don't think this is ever loaded for adapters, so we can probably get rid of these changes. :)

mjaric:
> > Since Ecto repo do not use PoolRepo I found this as better naming for module that should help making integration schema reusable.
> 
> Oh, I am sorry. So let's move it this configuration and everything else to the `TestRepo`, could it be?

Should I then change namespace from `Ecto.TestRepo` to `Ecto.Integration.TestRepo`... and to follow same naming convention ... `Ecto.TestAdapter` to `Ecto.Integration.TestAdapter`?

josevalim:
If we rename it from `Ecto.Integration.TestRepo` to `Ecto.TestRepo` we will have to rename a lot of tests, so let's keep the same module name. Also note that the `Ecto.TestRepo` is only used in `test` but the integration one is in `Ecto.Integration.TestRepo`, so they shouldn't conflict.

mjaric:
So I should add only rename `Ecto.Integration.AdapterTypes` module in `test_helper.exs`?

mjaric:
sorry, ignore my last comment, I just realized that ecto is not using integration_test. I will remove that new module since it is not needed here

josevalim:
@mjaric sorry, I think there is some confusion here. :)

We have two testing structures in this project. Everything in "test", including "test/test_helper.exs", should not be related to this discussion at all, as they never load the schemas.

The schemas files are used by integration tests but this repository does not run integration tests. So there is no "test_helper.exs" to change in this repository regarding integration tests. They are all in the ecto_sql repository.

So the changes should be:

1. change the schemas.exs to get the type from `Ecto.Integration.TestRepo`
2. remove the changes to test/test_helper.exs

Does it make sense? :)

josevalim:

josevalim:
Nope, i can squash when merging :) thanks!

PR #2880, opened 2018-12-31
add Ecto.Query.reverse_order/1 to reverse query sort

bgentry:
Following [my proposal from today](https://groups.google.com/d/msgid/elixir-ecto/8402afe1-734c-429c-bb32-0d4ad84f4daf%40googlegroups.com), this PR adds a `reverse_order/1` function to `Ecto.Query`. This function reverses the order for each of the query's `order_bys`.

Most of this logic was already implemented in `last/2`, so I extracted that to a shared private function and used that from both places.

Let me know what changes I can make 😄:v:

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1852,8 +1852,10 @@ defmodule Ecto.Query do
   end
 
   @doc """
-  Reverses the ordering of the query, with ASC columns becoming DESC columns
-  (and vice-versa). If the query has no order_bys, this is a no-op.
+  Reverses the ordering of the query.
+  
+  ASC columns become DESC columns (and vice-versa). If the query
+  has no order_bys, this is a no-op.
 
   ## Examples
 
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1805,22 +1805,10 @@ defmodule Ecto.Query do
 
   def last(%Ecto.Query{} = query, nil) do
     query = %{query | limit: limit()}
-    update_in query.order_bys, fn
-      [] ->
-        [order_by_pk(query, :desc)]
-      order_bys ->
-        for %{expr: expr} = order_by <- order_bys do
-          %{order_by | expr:
-              Enum.map(expr, fn
-                {:desc, ast} -> {:asc, ast}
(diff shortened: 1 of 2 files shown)


lib/ecto/query.ex
```diff
@@ -1862,4 +1850,36 @@ defmodule Ecto.Query do
   def has_named_binding?(queryable, key) do
     has_named_binding?(Ecto.Queryable.to_query(queryable), key)
   end
+
+  @doc """
+  Reverses the ordering of the query.
```
josevalim: The first line of the doc should always be a short summary, so I amended it.

josevalim:

test/ecto/query_test.exs
```diff
@@ -715,4 +715,18 @@ defmodule Ecto.QueryTest do
...
+    test "reverses the order of a simple query" do
+      order_bys = [asc: :inserted_at, desc: :id]
+      reversed_order_bys = [desc: :inserted_at, asc: :id]
+      q = from(p in "posts")
+      assert reverse_order(order_by(q, ^order_bys)) == order_by(q, ^reversed_order_bys)
+    end
+
+    test "is a no-op on a query with no order" do
+      q = from(p in "posts")
+      assert reverse_order(q) == q
```
michalmuskala: I'm really not sure about this behaviour. It seems very weird to me that `Repo.all(query) == Repo.all(reverse(query))`. I wouldn't expect such a behaviour.

josevalim:
The counter argument to that is that list == reverse(list) for an empty
list too. :)
-- 


michalmuskala:
The result of the query is not empty, though. And we already have some similar behaviour in the `last` query modifier.

josevalim:
@michalmuskala oh, you are saying that we should reverse by primary key then? That is fine by me. So last becomes `reverse_order |> first` right?

michalmuskala:
Yeah, I think that would give less surprising results.

bgentry:
Well, that would make the code even more shared with `last` 🙂 can’t think of any scenarios where that would be problematic either.

josevalim:
Sweet, i pushed a change to master!

PR #2881, opened 2019-01-06
Add :primary_key option to embeds_* docs

toraritte:
Documenting `:primary_key` option explicitly in the "Options" sections  for [`embeds_one/3`](https://hexdocs.pm/ecto/Ecto.Schema.html#embeds_one/3) and [`embeds_many/3`](https://hexdocs.pm/ecto/Ecto.Schema.html#embeds_many/3). 

There is a sentence ("_When defining an inline embed, the :primary_key option may be given to customize the embed primary key type._") between the examples but it would be nice to see it listed in the options.

Ran `mix test`, no issues but couldn't find any contribution guidelines (or may have overlooked...) so please let me know if this isn't the right way to contribute. 

Thanks for all your efforts on this awesome tool!


josevalim:

PR #2882, opened 2019-01-08
Document stale errors opts for insert_or_update/2

maartenvanvliet:
https://hexdocs.pm/ecto/Ecto.Repo.html#c:update/2 shows how to deal with StaleEntryErrors when you want to catch them instead of letting it raise. `insert_or_update/2` was missing this information. 

I've added this and made a note that it only applies to updates, inserts should never have stale errors.


josevalim:

PR #2889, opened 2019-01-23
Update from/2 docs with note about bindings

cdfuller:
I believe this change better reflects how `from/2` works. The extra source references are optional in my testing so I thought it could be changed from "must" to "may".


josevalim:

PR #2890, opened 2019-01-24
Use non-deprecated version of Multi.run in Multi DocTests

EskiMag:
Use the version with arity 2 instead of 1 as last argument to prevent warnings.


josevalim:

PR #2891, opened 2019-01-24
Add documentation for pool_size

jsangilve:
`Ecto.Repo`'s documentation mentions `pool_size` parameter under `URL` section, but anywhere else. Considering this parameter is defaulted to `10` in `Ecto.Repo` module, I added it to the documentation.

I believe `timeout` parameter is a similar case. However, it's documented under the `Ecto.Adapters.Postgres` and `EctoAdapters.MySQL` modules.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -43,7 +43,7 @@ defmodule Ecto.Repo do
       Defaults to `:debug`
 
     * `:pool_size` - the size of the pool used by the connection module.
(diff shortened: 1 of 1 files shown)


AndrewDryga:
I believe `pool_size` is removed in Ecto 3, instead please replace example in URL section with other option that we currently support?

josevalim:
pool_size is still around in Ecto 3.

josevalim:

PR #2892, opened 2019-01-25
Remove excessive underscore for ignored variable

turboza:
Minor correction for the ignored variable which is supposed to have only one underscore.

I scanned and didn't find this somewhere else.


josevalim:

PR #2905, opened 2019-02-01
No need for validation query unless relevant fields were changed

### LINKED ISSUE #2901: Improvement: lazy `Ecto.Changeset.unsafe_validate_unique/4`
While working on a project I noticed that `Ecto.Changeset.unsafe_validate_unique/4` executes a uniqueness check on my repo even when the field(s) specified are not present in the changeset's changes. Perhaps I've missed something, but I think this could be done lazily by comparing the changeset's changes with the specified fields and save a database query if it is not necessary.

### Environment

* Elixir version (elixir -v): 1.8.0
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgres 11
* Ecto version (mix deps): 3.0.6
* Database adapter and version (mix deps): postgrex 0.14.1
* Operating system: Ubuntu 18.04

### Current behavior

As part of my `User` schema:
```Elixir
def general(user, attrs) do
    user
    |> cast(attrs, [:name, :email])
    |> unsafe_validate_unique([:email], MyApp.Repo)
    |> unique_constraint(:email)
end
```

IEx:
```Elixir
alias MyApp.Accounts.User
%User{email: "dave@domain.tld", name: "Dave"} |> User.changeset(%{"name" => "Peter"})
[debug] QUERY OK source="users" db=3.4ms
SELECT TRUE FROM "users" AS u0 WHERE (u0."email" = $1) LIMIT 1 ["dave@domain.tld"]
#Ecto.Changeset<
  action: nil,
  changes: %{name: "Peter"},
  errors: [],
  data: #MyApp.Accounts.User<>,
  valid?: true
>
```
So a query is executed for the email field while it would not be changed by applying the resulting changeset.

### Expected behavior
```Elixir
alias MyApp.Accounts.User
%User{email: "dave@domain.tld", name: "Dave"} |> User.changeset(%{"name" => "Peter"})
#Ecto.Changeset<
  action: nil,
  changes: %{name: "Peter"},
  errors: [],
  data: #MyApp.Accounts.User<>,
  valid?: true
>
```

### Proposal

Perhaps it could be improved by wrapping `unsafe_validate_unique/4` in a change-checker, something like:
```Elixir
def lazy_unsafe_validate_unique(changeset, fields, repo, opts \\ []) do
    changed_fields =
      changeset.changes
      |> Map.keys()
      |> MapSet.new()
      |> MapSet.intersection(fields |> MapSet.new())

    if MapSet.new() != changed_fields do
      unsafe_validate_unique(changeset, fields, repo, opts)
    else
      changeset
    end
  end
```
Or something more performant along the same lines, if somebody can come up with something :)

nathanl:
`unsafe_validate_unique/3` performs a query to ensure that the specified
fields represent a unique combination of values in the repo. For
example, perhaps the combination of city and state must be unique.
But there's no need to perform this query if the changeset is not making
any change to city or state.

Fixes #2901

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1719,8 +1719,13 @@ defmodule Ecto.Changeset do
       {field, get_field(changeset, field)}
     end
 
+    # No need to query if we haven't changed any of the fields in question
+    unrelated_changes? = fields -- Map.keys(changeset.changes) == fields
+
     # If we don't have values for all fields, we can't query for uniqueness
-    if Enum.any?(where_clause, &(&1 |> elem(1) |> is_nil())) do
+    any_nil_values_for_fields? = Enum.any?(where_clause, &(&1 |> elem(1) |> is_nil()))
+
+    if unrelated_changes? || any_nil_values_for_fields? do
       changeset
     else
       pk_pairs = pk_fields_and_values(changeset, struct)
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1316,6 +1316,22 @@ defmodule Ecto.ChangesetTest do
     Process.put(:test_repo_all_results, no_dup_result)
     changeset = unsafe_validate_unique(base_changeset, :title, TestRepo, prefix: "public")
     assert changeset.valid?
+
+    # The validation runs a query only when the changes are relevant to it
+    defmodule DetectionRepo do
+      def one(query) do
+        send(self(), [function: :one, query: query])
+      end
+    end
+    body_change = changeset(%Post{title: "Hello World", body: "hi"}, %{body: "ho"})
+    unsafe_validate_unique(body_change, :body, DetectionRepo)
+    assert_receive [function: :one, query: %Ecto.Query{}]
+
+    unsafe_validate_unique(body_change, [:body, :title], DetectionRepo)
+    assert_receive [function: :one, query: %Ecto.Query{}]
+
+    unsafe_validate_unique(body_change, :title, DetectionRepo)
+    refute_receive [function: :one, query: %Ecto.Query{}]
   end
 
   ## Locks


nathanl:
If you set the diff to hide white space changes, you'll see that the change here is very small. The function is now even longer and uglier with the additional nesting. 😆 This might be a nice place for someone to contribute a refactor.

lib/ecto/changeset.ex
```diff
@@ -1706,55 +1706,68 @@ defmodule Ecto.Changeset do
...
-    where_clause = for field <- fields do
-      {field, get_field(changeset, field)}
-    end
+    fields_to_validate =
+      fields
+      |> MapSet.new()
 
-    # If we don't have values for all fields, we can't query for uniqueness
-    if Enum.any?(where_clause, &(&1 |> elem(1) |> is_nil())) do
+    if MapSet.disjoint?(changed_fields, fields_to_validate) do
```
josevalim: I think this list is small enough that doing `fields -- Map.keys(changs) == fields` is going to be much more efficient.

lib/ecto/changeset.ex
```diff
@@ -1706,55 +1706,68 @@ defmodule Ecto.Changeset do
...
-            Ecto.Query.or_where(acc, [q], field(q, ^field) != ^value)
-          end)
+      {validations, struct} =
+        case changeset do
+          %Ecto.Changeset{validations: validations, data: %{__struct__: struct}} ->
+            {validations, struct}
+          %Ecto.Changeset{} ->
+            raise ArgumentError, "unsafe_validate_unique/4 does not work with schemaless changesets"
         end
+      changeset = %{changeset | validations: [{:unsafe_unique, fields} | validations]}
```
josevalim: We should do this always, even if the fields did not change, so we have to move those lines up. Maybe we can streamline the code to avoid nesting? Something like this:

    if fields -- Map.keys(changes) == fields or Enum.any?(where_clause, &(&1 |> elem(1) |> is_nil())) do
      changeset

?

josevalim:
Thank you so much @nathanl! I have dropped some comments. We just need some tests too and  we  should be good to go!

nathanl:
@josevalim Thanks for the feedback. I noticed that `mix test` doesn't run the integration tests. I think `make test` used to do that, but I don't see how now.

josevalim:
@nathanl the integration tests are now part of `ecto_sql`. I think we can probably unit test this change in particular (i.e. you can pass a fake Repo as it should never be invoked). Otherwise `ecto_sql` is the way to go!

nathanl:
Hmm. I'm confused because `integration_test/cases/repo.exs` is still in this project and has tests referring to this validation, but putting `assert true == false` in one of them doesn't cause a failure.

josevalim:
@nathanl the integration tests are still shipped as part of ecto, because they are shared across adapters, but  there is no adapter in ecto per se for us to  run them against.

test/ecto/changeset_test.exs
```diff
@@ -1316,6 +1316,22 @@ defmodule Ecto.ChangesetTest do
...
+    end
+    body_change = changeset(%Post{title: "Hello World", body: "hi"}, %{body: "ho"})
+    unsafe_validate_unique(body_change, :body, DetectionRepo)
+    assert_receive [function: :one, query: %Ecto.Query{}]
+
+    unsafe_validate_unique(body_change, [:body, :title], DetectionRepo)
+    assert_receive [function: :one, query: %Ecto.Query{}]
+
+    unsafe_validate_unique(body_change, :title, DetectionRepo)
+    refute_receive [function: :one, query: %Ecto.Query{}]
```
josevalim: I was going to say we should break this apart but given everything is together for now, then it is best to leave it for another PR. :)

nathanl: @josevalim Here's a PR to break apart the test: https://github.com/elixir-ecto/ecto/pull/2909

PR #2909, opened 2019-02-05
Break apart tests for unsafe_validate_unique/3

nathanl:
Refactor in response to https://github.com/elixir-ecto/ecto/pull/2905#discussion_r253647018

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1276,62 +1276,90 @@ defmodule Ecto.ChangesetTest do
 
   alias Ecto.TestRepo
 
-  test "unsafe_validate_unique/3" do
-    dup_result = {1, [true]}
-    no_dup_result = {0, []}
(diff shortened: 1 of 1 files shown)


test/support/mock_repo.exs
```diff
@@ -0,0 +1,5 @@
+defmodule Ecto.MockRepo do
```
nathanl: I wasn't sure about adding another `test/support` module, but it's very generic and this seemed cleaner to me than defining it in the test. I'm happy to move it to the test if that seems better to others.

josevalim: Yeah, I would prefer to keep it inside the describe in the test case, as it is not used anywhere else. Even it was, having local fixtures is often better for clarity. :)

josevalim:
This is fantastic, thank  you @nathanl! One comment and we should be good to go! :heart:

nathanl:
@josevalim Updated!

josevalim:

PR #2911, opened 2019-02-06
Improve handling of arrays with empty values

alizain:
The primary issue here is that Ecto removes `""` as empty, but doesn't filter `[""]` to `[]` as well. This PR changes that behaviour.

**Question** should `[]` be considered an "empty value" as well (ie. `@empty_values ["" []]`)?

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -581,7 +581,7 @@ defmodule Ecto.Changeset do
     end
   end
 
-  defp strip_empty_values_from_list?(value, {:array, _}, empty_values) do
+  defp strip_empty_values_from_list?(value, {:array, _}, empty_values) when is_list(value) do
     Enum.reject(value, &Enum.member?(empty_values, &1))
   end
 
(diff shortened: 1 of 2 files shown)


alizain:
I agree that explicit is better. But, a list of values is just that - a list of values. If a value inside that list would be considered "empty" when not in a list, it should be considered empty if inside a list. By considering them to be different, I think we're sacrificing predictability.

Instead of doing it automatically, would you consider another API for the same functionality? Maybe an option can be passed in alongside `empty_values`, or another changeset function?

alizain:
To clarify, this PR does not make `[]` an "empty value". It simply strips "empty values" from a list of values, like so ->

```elixir
%{"topics" => ["", "", "", ""]} -> %{topics: []}
```

josevalim:
It is something that can be trivially handled before calling cast, so I don't see a strong reason to support it in Ecto. As mentioned, it is not a very common use case and sometimes it is simpler to handle it  upstream. Thanks.

alizain:
Okay, thank you

OvermindDL1:
Just to add a voice on the not-touching-it side, I really *really* would not like 'empty' values being stripped from an array because the empty value is a valid value in an array and it's very presence is a marker that something is there but that something is just null or "" or so.

alizain:
@OvermindDL1 yeah, I understand, thanks for sharing.

I've thought about it some more, and agree with @josevalim that it's not a good idea to do automatically.

PR #2913, opened 2019-02-07
Expose the type that was validated in errors generated by `Changeset.validate_length/3`

timrogers:
Errors in an `Ecto.Changeset` have a key, a message and additional contextual information provided in the form of a keyword list.

`Ecto.Changeset` comes bundled with a number of validations, for example `validate_length/3`.

The built-in validations tend to include additional contextual information - for example, `validate_length/3` includes `count` and `kind`. So, for example, if the validation was configured to check that a field was "no more than 3" and it fails, then you will have `[count: 3, validation: :length, kind: :max]`.

The `validate_length` validation is interesting, because it can operate on strings, binaries and lists. However, it doesn't expose in the additional context what type of data was validated.

This context can be valuable, for example if you want your errors to be fully machine-readable.

This exposes `[type: :string]`, `[type: :list]` or `[type: :binary]`, as appropriate, on the error.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1953,27 +1953,27 @@ defmodule Ecto.Changeset do
 
   defp wrong_length(_type, value, value, _opts), do: nil
   defp wrong_length(:string, _length, value, opts), do:
-    {message(opts, "should be %{count} character(s)"), count: value, validation: :length, kind: :is}
+    {message(opts, "should be %{count} character(s)"), count: value, validation: :length, kind: :is, type: :string}
   defp wrong_length(:binary, _length, value, opts), do:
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1953,27 +1953,27 @@ defmodule Ecto.Changeset do
 
   defp wrong_length(_type, value, value, _opts), do: nil
   defp wrong_length(:string, _length, value, opts), do:
```
timrogers: We could potentially pattern-match this if preferred:

```elixir
defp wrong_length(:string = type, _length, value, opts), do:
  {message(opts, "should be %{count} character(s)"), count: value, validation: :length, kind: :is, type: :string}
end
```

Issue #2918, opened 2019-02-18
`prefix` option ignored for some Queryables


devonestes:
### Environment

* Elixir version (elixir -v): 1.8.1
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6
* Ecto version (mix deps): 3.0.7
* Database adapter and version (mix deps): Postgrex 0.14.1
* Operating system: Linux Mint 19

### Current behavior

In the documentation for `Ecto.Repo.all/2`, it [says that you can pass a `prefix:`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo.ex#L605)
option that will override the prefix in the given query. However, I don't see any
examples of this working in trying to make that work today, and I didn't see
any tests [here](https://github.com/elixir-ecto/ecto/blob/8c25abd2fcbff5c4d7f7eb9c8457e40c2c2c7df7/test/ecto/repo_test.exs) for any `Repo` functions other than `insert`, `update`,
`insert_or_update` and `delete`, so I'm not sure this is working properly.

Here's an example:

In my schema for `User` I have:

```elixir
defmodule MyApp.User do
  use MyApp.Schema

  @schema_prefix "filtered"

  schema "users" do
    # ...
   end
end
```

No, if we assume that I have one record in `public.users` and 0 records in
`filtered.users`. The following code returns 0 records:

`Repo.all(User, prefix: "public")`
`Repo.all(from(u in User), prefix: "public")`

In order to get the right record returned I need to do:

`Repo.all(from(u in User, prefix: "public"))`

### Expected behavior

I would expect all three of the calls to `Repo.all/2` above to have the same
behavior.

I also see this same behavior with all the other functions other than the ones
I listed above that are tested in Ecto's tests, even though pretty much
every function in `Ecto.Repo` lists `prefix:` as an available option to override
the prefix of the given query.

josevalim:
From Ecto 3.0, the @schema_prefix is per *schema* and no longer for the whole query. So we first lookup to the schema prefix and then fallback to query. Since you have defined a prefix for the schema, it will never fallback to the query one.

PR #2919, opened 2019-02-19
Add tests and better documentation for behavior around `prefix` option

devonestes:
There was some confusing documentation and very little test coverage
around the behavior of the `prefix` option in the `Ecto.Repo` behaviour.
I've added more tests to clarify the current expected behavior, and
documented that behavior more thoroughly so folks aren't surprised by
some of the edge cases that arrise when using that option.

Hopefully this will clear up some of the confusion I had that led me to
post #2918 yesterday.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -414,11 +414,20 @@ defmodule Ecto.Repo do
 
   ## Options
 
+    * `:prefix` - The prefix to run the query on (such as the schema path
+      in Postgres or the database in MySQL). This will be applied to all `from`
+      and `join`s in the query that did not have a prefix previously given
+      either via the `:prefix` option on `join`/`from` or via `@schema_prefix`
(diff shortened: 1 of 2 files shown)


lib/ecto/repo.ex
```diff
@@ -414,12 +414,18 @@ defmodule Ecto.Repo do
 
   ## Options
 
+    * `:prefix` - The prefix to run the query on (such as the schema path
+      in Postgres or the database in MySQL). This will only be applied if no
+      `@schema_prefix` has previously been set.
```
josevalim: I think the more precise definition is: "This will be applied to all from and joins in the query that did not have a prefix explicitly given (either via the :prefix option on join/from or via `@schema_prefix` in the schema)".

josevalim: Ping. :)

devonestes: :flushed: updated!

josevalim:

PR #2923, opened 2019-02-22
Add `not_equal_to` option for `validate_number`

hauleth:
This checks for inequality with given parameter. It is useful in
situations when we want to disallow one particular value. For example
imagine that we want to define table to store value income and outcome
in some account. It make sense to be positive (income) or negative
(outcome), but it makes no sense to store operation that has value equal
to `0`, with this change we can simply:

```elixir
validate_number(changeset, :diff, not_equal_to: 0)
```

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1095,6 +1095,13 @@ defmodule Ecto.ChangesetTest do
     assert changeset.errors == [upvotes: {"must be greater than %{number}", validation: :number, kind: :greater_than, number: 0}]
     assert validations(changeset) == [upvotes: {:number, [greater_than: 0]}]
 
+    # Non equality error
+    changeset = changeset(%{"upvotes" => 1})
+                |> validate_number(:upvotes, not_equal_to: 1)
+    refute changeset.valid?
+    assert changeset.errors == [upvotes: {"must be not equal to %{number}", validation: :number, kind: :not_equal_to, number: 1}]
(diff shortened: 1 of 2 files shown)


test/ecto/changeset_test.exs
```diff
@@ -1127,6 +1127,10 @@ defmodule Ecto.ChangesetTest do
                 |> validate_number(:decimal, equal_to: Decimal.new(-1))
     assert changeset.valid?
 
+    changeset = changeset(%{"decimal" => Decimal.new(0)})
+                |> validate_number(:decimal, not_equal_to: Decimal.new(-1))
+    assert changeset.valid?
```
josevalim: This test case is only checking things for decimals but we need a regular test as well, that checks that `validations(changeset)` returns the proper metadata and what happens in case of errors. Thanks!

josevalim:
I have added one comment and we shuold be good to go!

test/ecto/changeset_test.exs
hauleth: Added simple test to check whether metadata are correct. However in the process I have found possible bug, but I am not sure about that. To be exact, data set as `default` aren't checked by the validations, so this will pass:

```elixir
defmodule Foo do
  use Ecto.Schema

  schema "foos" do
    field :count, :integer, default: 0
  end

  def changeset(params) do
    %__MODULE__{}
    |> validate_number(:count, not_equal_to: 0)
  end
end

assert Foo.changeset(%{"count" => 0}).valid?
```

josevalim:

PR #2928, opened 2019-02-27
Add docs for proper testing env setup with mix

achempion:
I think for now the guide looks incomplete about how to start testing with ecto. We have a mix project but after we create `repo_case.ex` somewhere it wouldn't compile.

Also, it was unclear should we create either `.exs` or `.ex`.

I think the improvement will ease the way for proper tested software :)


josevalim:

PR #2938, opened 2019-03-11
Disable moduledoc for embedded schemas

akoutmos:
Firstly, thanks for Ecto! It's awesome!!! 

I make use of embedded schemas quite a bit, and I find that the generated ex_docs are a bit noisy from all the generated modules. This change makes it so that embedded schemas do not show up in the ex_doc generated module lists.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1836,6 +1836,8 @@ defmodule Ecto.Schema do
 
     block =
       quote do
+        @moduledoc false
+
         use Ecto.Schema
 
         @primary_key unquote(Macro.escape(pk))


Eiji7:
I suggest to use your own macro which calls `use Ecto.Schema` and adds `@moduledoc false` rather than force it to library. Please notice that for some use cases really strong documentation is just **must have**. For example think that client **requires** to include documentation for every part of code in order to understand every part properly even for newbie. Look how weird it would work in case like this one:

```
defmodule MyApp.MyContext.MyModel do
  @moduledoc """
  This documentation is going to be "magically" overwritten …
  """

  use Ecto.Schema
end
```

josevalim:
As @Eiji7 said, I don't think it is Ecto's job to disable documentation or not but you are welcome to encapsulate it in your app!

akoutmos:
Totally fair to pull this concern into my side of the code, but I think we may be talking about different things. This moduledoc disable is not for ecto schemas (definitely on the same page that disabling those docs magically is a terrible experience) but for embedded schemas. From what I understand, if you have a schema definition like so:

```elixir
defmodule Parent do
  use Ecto.Schema

  schema "parents" do
    field :name, :string

    embeds_many :children, Child do
      field :name, :string
      field :age,  :integer
    end
  end
end
```
2 modules will be created: `Parent` and the dynamically generated `Parent.Child`. When calling `Code.fetch_docs/1` on `Parent.Child`, the `doc_content` will currently always be `:none`. The proposed change makes the `doc_content` `:hidden` instead. 

Is there a way to disable the moduledoc without this change in the dynamically generated embedded schema module? My understanding is that once `Module.create/3` is called the module cannot be altered https://github.com/elixir-ecto/ecto/blob/a74bbec8fa74a552f14da0adc9d1baeb4675608c/lib/ecto/schema.ex#L1848

josevalim:
This should work:

```elixir
    embeds_many :children, Child do
      @moduledoc false
      field :name, :string
      field :age,  :integer
    end
```

akoutmos:
I could have sworn that I tried that and didn't get the result I wanted...but I tried it again and it works. Sorry for the confusion and thanks @josevalim!

PR #2941, opened 2019-03-12
Improves documentation on Ecto.Adapter.checkout/3;

### LINKED ISSUE #2935: Improve documentation on Elixir.Adapter.checkout/3 behaviour callback
Currently, it is very difficult to understand how [Ecto.Adapter.checkout/3](https://hexdocs.pm/ecto/Ecto.Adapter.html#c:checkout/3) should be implemented by an adapter.
Furthermore, it is difficult to see what is going on in the reference adapter implementations of  MySQL and Postgres because [the implementation in ecto_sql](https://github.com/elixir-ecto/ecto_sql/blob/04d9b75e491021dc3e2d5850ab65ae5573f26c68/lib/ecto/adapters/sql.ex#L478) [redirects in a roundabout, undocumented way to](https://github.com/elixir-ecto/ecto_sql/blob/04d9b75e491021dc3e2d5850ab65ae5573f26c68/lib/ecto/adapters/sql.ex#L869) [DBConnection.run](https://github.com/elixir-ecto/db_connection/blob/master/lib/db_connection.ex#L706)(?)

Adapter-implementers would be greatly helped if:

- There is more information on the kind of input/output that is expected for `checkout/3`.
- There is an explanation on how `checkout/3` is expected to make the handle of a pool that was checked out available to the inner function.
- There are some examples for `checkout/3`.
- There is an explanation of how to implement it if you do not (or cannot) use a pool.

### LINKED ISSUE #2936: Document how Ecto.Adapter.Queryable should be implemented
The documentation of [Ecto.Adapter.Queryable](https://hexdocs.pm/ecto/Ecto.Adapter.Queryable.html) is unclear of what is expected by the callbacks.

Especially if we have datastores for which only some types of queries make sense (like for instance simple key-value stores that only expose `get` and `insert`), it is not clear how this should be done using the given behaviour callbacks.

Qqwy:
Fixes #2935.

- Improves documentation on Ecto.Adapter.checkout/3.
- Renames 'options' to 'config' in callback to match naming with the other callbacks in this module.
- Also adds a short notice to the module documentation of
Ecto.Adapter.Queryable (fixes #2936).

### DIFF
--- a/lib/ecto/adapter.ex
+++ b/lib/ecto/adapter.ex
@@ -52,8 +52,9 @@ defmodule Ecto.Adapter do
 
   If the adapter provides a pool, it is supposed to 'check out'
   one of the pool connections for the duration of the function call.
-  Which connections is checked out is not passed to the calling function, so it should be done
-  using a stateful method like the calling process' pdict (Process Dictionary),
+  Which connections is checked out is not passed to the calling function,
+  so it should be done using a stateful method like using
+  the current process' pdict (Process Dictionary),
   Process Tracking, or some kind of other lookup method.
   (Of course, make sure that this stored connection is then used
   in the implementations of e.g. `Ecto.Adapter.Queryable` and `Ecto.Adapter.Schema`.)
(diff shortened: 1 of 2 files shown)


josevalim:

PR #2942, opened 2019-03-13
Provide examples with correct API usage in `Ecto.Multi` module.

fuelen:
It was confusing for me and I think it is better to provide valid examples.


josevalim:

PR #2947, opened 2019-03-19
work with dynamic multi tenant databases

bartekupartek:


josevalim:
Thanks @bartekupartek but the dynamic dispatch could be done in your application side too by doing `dynamic_repo().update_all`, `dynamic_repo().get`, etc and so on.

bartekupartek:
ok, @josevalim this is great feature I was looking for in https://github.com/elixir-ecto/ecto/issues/1964 and introduced in Ecto 3.0, but is there any plan to replace `__MODULE__` based Repo api in the future? It'd be optional feature only for multi tenant databases and it'd save overwriting Repo functions, database could be set by ie. plug based on subdomain. I'm going to merge over 40 app instances with independent databases into one app instance and leave databases untouched, so I need to configure Repo to handle multiple connections, schemas would be the same on each database also do you have idea how to run `Ecto.Migrator` over Repo with dynamic databases?

josevalim:
I see @bartekupartek!

I think migrations was one of the parts pending in the dynamic repo work. If your goal is to use dynamic repo, then my suggestion is to define your own `Ecto.DynamicRepo` that uses Ecto private APIs (like you did in this commit) and then post how you are using the feature in the dynamic repo thread (or write about it generally). We need precisely this sort of feedback to finish the feature. But if you have 40 known instances, then you can also define the repos dynamically.

bartekupartek:
@josevalim how about following change I'd like to apply to all Repo functions?
```
def insert(struct, opts \\ [], repo_name_or_pid \\ __MODULE__) do
  Ecto.Repo.Schema.insert(repo_name_or_pid, struct, opts)
end
```

josevalim:
I am thinking about a separate module that provides those. But yeah... we
need something along those lines. :)
-- 

PR #2948, opened 2019-03-22
Fix Ecto.Changeset.check_constraint/3 @doc example

esambo:
I just noticed that the example given from #2878 doesn't work and fixed it accordingly

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2307,14 +2307,14 @@ defmodule Ecto.Changeset do
   In order to use the check constraint, the first step is
   to define the check constraint in a migration:
 
-      create constraint("users", :price_must_be_positive, check: "price > 0")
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -2307,14 +2307,14 @@ defmodule Ecto.Changeset do
   In order to use the check constraint, the first step is
   to define the check constraint in a migration:
 
-      create constraint("users", :price_must_be_positive, check: "price > 0")
+      create constraint("users", :price, :price_must_be_positive, check: "price > 0")
```
wojtekmach: this was correct for _migrations_, could you revert this change? Good catch on the changeset function tho!

esambo: ups, thanks for catching that! I fixed it in 5df8138

wojtekmach:
Thank you!

PR #2953, opened 2019-03-29
lookup for repo_name_or_pid in Repo

bartekupartek:
make Repo flexible to use with named Repos for multi tenant :)


bartekupartek:
or pass it in opts? or duplicate `Ecto.Repo` code to `Ecto.DynamicRepo` just to handle `repo_name_or_pid` would be nicer? in `DynamicRepo` way defaults opts would remain optional. Other advantages/disadvantages?

bartekupartek:
IMHO Keyword way looks better.

josevalim:
Oh, I like the options approach to!

So we have:

1. Options
2. A separate module
3. Pdict

I will think about these options and let you know how we want to proceed!

bartekupartek:
FYI isolated `repo_name_or_pid` option directly in `Ecto.Repo.Registry.lookup/2` to have test in one place, I hope it's good

bartekupartek:
I've started using it in my app but it still needs to duplicate code into separable module just to pass extra option for repo, otherwise getting `defines defaults multiple times` in the same repo

bartekupartek:
But problem still exists, somewhere I have to wrap opts to add `repo_name_or_pid`, I'll try separable module approach

josevalim:
Just a quick heads up: we have decided to go with `Ecto.Repo.put_dynamic_repo/1`!

josevalim:
Hi @bartekupartek! I pushed 297823aad2bade23a8b5989960b65c4862b96faa to master with some docs and some other changes that streamline the dynamic repo usage. Thank you so much for exploring the different ideas and sending pull requests, without them this feature would not have been possible!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

bartekupartek:
Wow, even better, your documentation is far better then my and few things I haven't noticed, Thank you for finishing it ❤️

bartekupartek:
TIL: `@compile {:inline, get_dynamic_repo: 0}`

Issue #2954, opened 2019-03-29
Associated model updating updated_at timestamp when sibling is updated


epilgrim:
Elixir Version:
```
Erlang/OTP 21 [erts-10.1.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
Elixir 1.8.0 (compiled with Erlang/OTP 20)
```
Ecto version: master

### Current behavior
I'm in the middle of the upgrade from ecto 2 to ecto 3, and i'm seeing many updates in an associated table, only changing the updated_at field (generated through timestamps)

I've added this failing test case that shows the issue: https://github.com/epilgrim/ecto/commit/9e62d5e2f72f1901f195185592ffd55609117825

The idea is that if I have company with multiple offices, when updating one of the offices through the company changeset (with cast_assoc) only the updated office should change it's `updated_at` timestamp. That is the behavior i had with ecto 2.

```elixir
  test "does not update updated_at when the associated record did not change" do
    company = TestRepo.insert!(%Company{offices: [%Office{id: 1, name: "1"}, %Office{id: 2, name: "2"}]})
    [office_one, office_two] = company.offices

    changes = %{offices: [%{id: 1, name: "updated"}, %{id: 2, name: "2"}]}
    updated_company =
      company
      |> Ecto.Changeset.cast(changes, [])
      |> Ecto.Changeset.cast_assoc(:offices)
      |> TestRepo.update!()
    [updated_office_one, updated_office_two] = updated_company.offices
    assert updated_office_one.updated_at != office_one.updated_at
    assert updated_office_two.updated_at == office_two.updated_at
  end
```

I haven't seen anything related to this change in the changelog.

### Expected behavior

The updated_at timestamp only gets updated when the specific entity changed, not when one of it's siblings changed

josevalim:
> The idea is that if I have company with multiple offices, when updating one of the offices through the company changeset (with cast_assoc) only the updated office should change it's updated_at timestamp. That is the behavior i had with ecto 2.

Are you sure this is the behaviour in Ecto 2? I am almost sure that Ecto 2 also propagated the updates upstream but now I have to double check it. :D

epilgrim:
Positive. Just checked this same test in the latest tag of v2, and the test passes (in v3 it fails)

Bear in mind that it's not the propagation upstream, but the sibling the one that "wrongly" get's the timestamp updated

josevalim:
[...]

PR #2957, opened 2019-04-01
remove unused opts configuration in test

epilgrim:
small cleanup after #2954.
`@timestamp_opts` set in the schema is not used at all in the tests


josevalim:

PR #2963, opened 2019-04-04
Update CHANGELOG

wojtekmach:


josevalim:
I don't mind keeping a single history for now, especially as we don't plan to release frequently anyway and this can help those still on v2.x planning to jump to latest v3.x. Thoughts?

wojtekmach:
Yup, it's fine to keep single history. I just noticed we were missing v3.0.8 entry on master (as we cut off a v3.0 branch) so thought about this, but I think we can keep it manually in sync

Issue #2961, opened 2019-04-03
`Repo.transaction(multi)` doesn't work with dynamic repos


bartekupartek:
### Environment

* Ecto version (mix deps): 3.1.0

### Current behavior

`Repo.put_dynamic_repo(:hospital1)` and `Repo.transaction(multi)`  doesn't get the repo module but process name.

```
stacktrace:
       :hospital1.insert(#Ecto.Changeset<action: :insert, changes: %{amount: #Decimal<100>, amount_paid: 66, doctor: #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #SecretApp.Doctor<>, valid?: true>, payments: [#Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #SecretApp.Patient.Payment<>, valid?: true>, #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #SecretApp.Patient.Payment<>, valid?: true>], provider: #Ecto.Changeset<action: :update, changes: %{}, errors: [], data: #SecretApp.Billing.Provider<>, valid?: true>}, errors: [], data: #SecretApp.Patient.ClaimsApplication<>, valid?: true>, [])
       (ecto) lib/ecto/multi.ex:579: Ecto.Multi.apply_operation/5
       (elixir) lib/enum.ex:1925: Enum."-reduce/3-lists^foldl/2-0-"/3
       (ecto) lib/ecto/multi.ex:563: anonymous fn/5 in Ecto.Multi.apply_operations/5
       (ecto_sql) lib/ecto/adapters/sql.ex:875: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
       (db_connection) lib/db_connection.ex:1415: DBConnection.run_transaction/4
       (ecto) lib/ecto/repo/transaction.ex:16: Ecto.Repo.Transaction.transaction/3
       (secret_app) lib/secret_app/patient/claim_forms.ex:89: SecretApp.Patient.ClaimForms.apply_claim_forms_payments/1
       test/lib/finance/claim_forms_test.exs:191: (test)
```

josevalim:
Awesome! Can you please provide a failing test case and possibly a PR with
a fix? Thank you!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

iautom8things:
[...]

PR #2964, opened 2019-04-04
fix dynamic repo in transaction (#2961)

bartekupartek:


test/ecto/multi_test.exs
```diff
@@ -496,4 +496,19 @@ defmodule Ecto.MultiTest do
       end
     end
   end
+
+  describe "dynamic repo" do
```
josevalim: Let's please move this to `repo_test`, since we already have some dynamic_repo tests in there. WDYT?

josevalim:

PR #2969, opened 2019-04-05
update_change consistent with put_change

balexand:
This fixes a bug that is reproduced by the included test case. `update_change` was leaving a field in the `changes` map even though that field was not being changed. The new behavior is consistent with the behavior of `put_change`.

Thanks ✨


balexand:
This pull-request might not be something that can be merged because it is a breaking change. If it can't be merged then is there an alternative solution that you would be willing to merge? Maybe there could be a `clean_changes` function that removes changes that are identical to the original value?

josevalim:

PR #2971, opened 2019-04-09
fix.validate_confirmation

SumitPugalia:
If field is required, the code worked fine.
But when the field itself was optional, we had to make the confirmation field as optional (required: false) because if we make it required it will be invalidated when field is not provided.

Example - 

  @cast_key [
    :first_name,
    :last_name,
    :mobile_number,
    :email,
    :password,
    :password_hash
  ]
  
  @required_key @cast_key -- [:password, :password_hash]

  @primary_key {:id, :binary_id, autogenerate: true}
  schema "customers" do
    field :email, :string
    field :first_name, :string
    field :last_name, :string
    field :mobile_number, :string
    field :password, :string, virtual: true
    field :password_hash, :string
    timestamps()
  end

  def changeset(customer , attrs) do
    customer
    |> cast(attrs, @cast_key)
    |> validate_required(@required_key)
    |> validate_length(:password, min: 8, max: 26)
    |> validate_confirmation(:password, message: "does not match password", required: true)
    |> hash_password()
  end

Now In this case even if we don't pass password (as its optional ) the validation_confirmation adds error. (Fixing this issue)

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2057,6 +2057,11 @@ defmodule Ecto.Changeset do
   By calling `validate_confirmation(changeset, :email)`, this
   validation will check if both "email" and "email_confirmation"
   in the parameter map matches.
+  
+  Note that when the field is optional, it makes sure that field 
+  is present to validate the confirmation field. Incase the field is 
+  not present and confirmation field is required it will not add error.
+  It will add the error only when the field is present. 
 
   Note that if the confirmation field is nil or missing, by default this does
   not add a validation error. You can specify that the confirmation field is
@@ -2084,11 +2089,11 @@ defmodule Ecto.Changeset do
     param = Atom.to_string(field)
     error_param = "#{param}_confirmation"
     error_field = String.to_atom(error_param)
-    value = Map.get(params, param)
+    value = Map.fetch(params, param)
 
     errors =
       case Map.fetch(params, error_param) do
-        {:ok, ^value} ->
+        ^value ->
           []
         {:ok, _} ->
           [{error_field,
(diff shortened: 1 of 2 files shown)


josevalim:
@SumitPugalia thanks for the PR!

Quick question: it seems that the behaviour you want to achieve is the one you would get by not passing required: true, no?

SumitPugalia:
> @SumitPugalia thanks for the PR!
> 
> Quick question: it seems that the behaviour you want to achieve is the one you would get by not passing required: true, no?

If I do so, then validation_confirmation will not add error even if i don't pass confirmation field..
%{password: "password"} will not add error ..

josevalim:
Which is intentional. Confirmation is a UI based feature. You don’t want to
require confirmation for things like API, where the confirmation should not
be sent.

In a browser, if you include the field, it is always sent and therefore it
should always trigger!
-- 


SumitPugalia:
> Which is intentional. Confirmation is a UI based feature. You don’t want to require confirmation for things like API, where the confirmation should not be sent. In a browser, if you include the field, it is always sent and therefore it should always trigger! -- *José Valimwww.plataformatec.com.br <http://www.plataformatec.com.br/>Founder and Director of R&D*

Agree 👍 
Thanks for that.

josevalim:
Cool. So do you think this issue still applies? It seems removing the
required is the best option.
-- 


SumitPugalia:
> Cool. So do you think this issue still applies? It seems removing the required is the best option. -- *José Valimwww.plataformatec.com.br <http://www.plataformatec.com.br/>Founder and Director of R&D*

Yes I can use it as required: false. I was trying to achieve something that cannot occur in real scenario.
I realised confirmation field should always come with the field

Issue #2973, opened 2019-04-12
Cannot set type for result of the window function


fuelen:
### Environment
I've created a project for reproducing, here is a test with queries
https://github.com/fuelen/ecto_issue_with_over/blob/master/test/ecto_ussue_with_over/main_test.exs

### Current behavior
I have the following query
```elixir
query =
      from(users in User,
        join: posts in assoc(users, :posts),
        select: %{
          user: users,
          post_ids: over(fragment("ARRAY_AGG(?)", posts.id), :post_ids)
        },
        distinct: users.id,
        windows: [post_ids: [partition_by: users.id]]
)
```
the result of execution is
```elixir
[%{
    post_ids: [
      <<114, 24, 89, 122, 170, 38, 73, 180, 175, 160, 202, 12, 70, 25, 79, 108>>,
      <<13, 144, 66, 132, 16, 28, 74, 241, 128, 22, 228, 13, 63, 105, 15, 107>>,
      <<35, 162, 41, 35, 227, 224, 79, 204, 161, 226, 63, 87, 50, 60, 48, 228>>
    ],
    user: %EctoIssueWithOver.User{
      __meta__: #Ecto.Schema.Metadata<:loaded, "users">,
      id: "f6cc67ad-6cb9-4e0a-9eb3-da93a36c4496",
      name: "name2",
      posts: #Ecto.Association.NotLoaded<association :posts is not loaded>
    }
  }]
```
and I want to get an array of strings (id is UUID) not binaries.
I tried this

```elixir
     query =
       from(users in User,
         join: posts in assoc(users, :posts),
         select: %{user: users, post_ids: type(over(fragment("ARRAY_AGG(?)", posts.id), {:array, :binary_id}), :post_ids)},
         distinct: users.id,
         windows: [post_ids: [partition_by: users.id]]
   )
```
but got a compilation error
```
 == Compilation error in file test/ecto_ussue_with_over/main_test.exs ==
     ** (Ecto.Query.CompileError) `type(over(fragment("ARRAY_AGG(?)", posts.id()), {:array, :binary_id}), :post_ids)` is not a valid query expression.
    
     * If you intended to call a database function, please check the documentation
       for Ecto.Query to see the supported database expressions
    
     * If you intended to call an Elixir function or introduce a value,
 you need to explicitly interpolate it with ^
```

### Expected behavior
load the result of the `over` to custom types

josevalim:
[...]

PR #2974, opened 2019-04-12
Add support for type+over

fuelen:
https://github.com/elixir-ecto/ecto/issues/2973

@josevalim I've tested exactly with your suggestion.
It generates an invalid query
```
** (Postgrex.Error) ERROR 42601 (syntax_error) syntax error at or near "OVER"
     
  query: SELECT DISTINCT ON (u0."id") u0."id", u0."name", ARRAY_AGG(p1."id") OVER "post_ids" FROM "users" AS u0 INNER JOIN "posts" AS p1 ON p1."user_id" = u0."id" WINDOW "post_ids" AS (PARTITION BY u0."id")
```
So, this version
```
type(over(fragment("ARRAY_AGG(?)", posts.id), {:array, :binary_id}), :post_ids)
```
generates correct query now.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -110,6 +110,10 @@ defmodule Ecto.Query.Builder do
     escape_with_type(expr, type, params_acc, vars, env)
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:

PR #2975, opened 2019-04-16
Add explicit casting of non-base type primary keys

slapers:
This is a WIP to get some feedback on casting custom primary keys. 
I created a topic in [elixir-ecto](https://groups.google.com/forum/#!topic/elixir-ecto/7VnyVUU26c0) and a question on the [forum](https://elixirforum.com/t/postgres-composite-type-as-primary-key-in-ecto-schemas/21195) that describes the issue i face. 

I'm struggling with the preload `assoc_query` implementation, resorting to manually creating a BooleanExpr to properly add type info when joining. I only added this in the `Has` association for now.

I opened a complementary PR on [elixir-ecto/ecto_sql](https://github.com/elixir-ecto/ecto_sql/pull/106) that supports the casting needed in the filters.


josevalim:
Hi @slapers!

So if I understand the problem correctly, you are trying to associate two tables with have columns of different types? Why can't they be of the same type?

slapers:
Hi @josevalim, actually im trying to use primary keys that are [postgres composite types](https://www.postgresql.org/docs/10/rowtypes.html). 

For example a URN could be a composite type with a Ecto.Type implementation like this:

```
defmodule URN do
   @moduledoc "Represents a URN. See https://tools.ietf.org/html/rfc8141"
   @behaviour Ecto.Type
   defstruct [:nid, :nss]

   def type, do: :urn
   def new(nid, nss), do: {:ok, %__MODULE__{nid: nid, nss: nss}}
   def cast(%__MODULE__{} = urn ), do: {:ok, urn}
   def cast({nid, nss}), do: new(nid, nss)
   def load({nid, nss}), do: new(nid, nss)
   def dump(urn), do: {:ok, {urn.nid, urn.nss}}
end
```

The type would be created in postgres like so:
```
CREATE TYPE public.urn AS (nid varchar, nss varchar);
```

This could be a schema using the URN as primary key:

```
defmodule Book do
  use Ecto.Schema

  @primary_key {:id, URN, autogenerate: false}
  schema "books" do
    field :title
  end
end
```

Currently in ecto inserting is working correctly:

```
iex(4)> %Book{id: %URN{nid: "isbn", nss: "123465789"}} |> Repo.insert!
[debug] QUERY OK db=0.8ms decode=1.1ms queue=0.5ms
INSERT INTO "books" ("id") VALUES ($1) [{"isbn", "123465789"}]
```

Unfortunately fetching, deleting or updating a record fails:
```
iex(5)> Repo.get Book, %URN{nid: "isbn", nss: "123465789"}
[debug] QUERY ERROR source="books" db=1.8ms queue=1.4ms
SELECT b0."id", b0."title" FROM "books" AS b0 WHERE (b0."id" = $1) [{"isbn", "123465789"}]
** (FunctionClauseError) no function clause matching in Postgrex.DefaultTypes.encode_tuple/5 
```

This is because the query should cast the type, with the PR it looks like this:
```
iex(1)> EctoCompositeType.Repo.get Book, %URN{nid: "isbn", nss: "123465789"}
[debug] QUERY OK source="books" db=1.0ms decode=0.9ms queue=1.1ms
SELECT b0."id", b0."title" FROM "books" AS b0 WHERE (b0."id" = $1::urn) [{"isbn", "123465789"}]
```

I just wanted to reach out if this is worth considering to add to ecto.. We are working on a project where this would be really nice to have.

josevalim:
The error is coming from Postgrex. I wonder if a Postgrex type extension would actually help solve that? Because Postgres *knows* that `b0."id"` is a URN, doesn't it? The issue is just that Postgrex does not know what that looks like in-and-out of Elixir.

slapers:
We tried creating a postgrex type first but got stuck. I believe we got it somehow working with hardcoding oids in the extension but could not figure out where to get them dynamically.

josevalim:
Can you paste the code here for further discussions? In theory postgrex worries about the OIDs for you.

josevalim:
FWIW, your extension should be declared like this:

    use Postgrex.BinaryExtension, type: "urn"

And if you define the encode/decode to emit a %URN{...} struct directly then I believe it should work.

slapers:
Unfortunately `type: "urn"` does not cause the encode/decode to be called. 

It seems we only get called when we match with `send: "record_send"`, so we ended up with this: (which needs to also take care of regular records)

```
@behaviour Postgrex.SuperExtension
  import Postgrex.BinaryUtils, warn: false

  def init(_), do: nil
  def format(_), do: :super_binary
  def matching(_), do: [type: "urn", send: "record_send"]

  def oids(%Postgrex.TypeInfo{comp_elems: []}, _), do: nil
  def oids(%Postgrex.TypeInfo{comp_elems: comp_oids}, _), do: comp_oids

  def encode(_) do
    quote do
      %URN{nid: nid, nss: nss}, nil, _types ->
        # types and oids we get are not ok, so hardcoded inline

        case encode_tuple({nid, nss}, [1043, 1043], [
               Postgrex.Extensions.Raw,
               Postgrex.Extensions.Raw
             ]) do
          :error ->
            raise DBConnection.EncodeError, "unable to correctly encode urn"

          data ->
            [<<IO.iodata_length(data) + 4::int32, 2::int32>> | data]
        end

      tuple, oids, types when is_tuple(tuple) ->
        # encode_tuple/3 defined by TypeModule
        case encode_tuple(tuple, oids, types) do
          :error ->
            raise DBConnection.EncodeError,
                  "expected a tuple of size #{length(oids)}, got: #{inspect(tuple)}"

          data ->
            [<<IO.iodata_length(data) + 4::int32, tuple_size(tuple)::int32>> | data]
        end

      other, _, _ ->
        raise DBConnection.EncodeError, Postgrex.Utils.encode_msg(other, "a tuple")
    end
  end

  def decode(_) do
    quote do
      any, oids, types -> raise "not implemented"
    end
  end
end
```

Running a query now, unfortunately results in:
```
** (Postgrex.Error) ERROR 0A000 (feature_not_supported) input of anonymous composite types is not implemented
```

I suppose the outer oid is that of a plain record instead of the urn type ?

josevalim:
Hi Stefan, can you push a small repo where you are attempting these ideas?
Then I can try to jump in and play with things as well. :)
-- 


slapers:
Of course, just cleaned it up a bit and pushed it to: https://github.com/slapers/pg_ext_testing
There is 1 test that yields the error on anonymous types. Thx for jumping in on it :-)

josevalim:
> ** (Postgrex.Error) ERROR 0A000 (feature_not_supported) input of anonymous composite types is not implemented

Yeah, I think we need to send this one to PG mailing list and ask them the best way to encode it. Or if we can only encode it by explicitly casting it in the query.

josevalim:
Also, for future discussion, I got this working:

```elixir
defmodule PgExt.URN.PgExtension do
  @moduledoc false

  @behaviour Postgrex.Extension

  def init(_), do: nil

  def matching(_), do: [type: "urn", send: "record_send"]

  def format(_), do: :binary

  def encode(a) do
    quote location: :keep do
      %PgExt.URN{nid: nid, nss: nss} ->
        nid_size = byte_size(nid)
        nis_size = byte_size(nss)
        total_size = 4 + 8 + nid_size + 8 + nis_size

        IO.[
          <<total_size::int32, 2::int32>>,
          <<1043::int32, nid_size::int32>>, nid,
          <<1043::int32, nis_size::int32>>, nss
        ]
      other ->
        raise DBConnection.EncodeError, Postgrex.Utils.encode_msg(other, PgExt.URN)
    end
  end

  def decode(a) do
    quote location: :keep do
      other ->
        IO.inspect other
    end
  end
end
```

But the OIDs are also hardcoded, we should (maybe we do?) have an API for retrieving them.

slapers:
That looks a quite bit cleaner than what we came up with.. :-)  Also i noticed that leaving out `type: "urn"` does not affect the encode from being used, seems like `send: "record_send"` is the only relevant thing in the matching.

josevalim:
I guess type: "urn" is nececssary if you have other types?

josevalim:
Hi @slapers, were you able to reach out to the Postgres folks about the best way to handle this issue? In case you were not, please open up an issue in Postgrex so we provide an official API to retrieve the OID, as I think there would be the best place to tackle this. Thanks!

Issue #2979, opened 2019-04-17
Dynamically Providing key name to select with group_by returns wrong query


tanweerdev:
### Environment

* Elixir version (elixir -v):
    Erlang/OTP 20 [erts-9.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:10] [hipe] 
  [kernel-poll:false]

    Elixir 1.8.1 (compiled with Erlang/OTP 20)
* Ecto version (> 3 exact 3.0.4):
* Operating system: MacOS 10.14.3

### Current behavior
    query = from en in Data.InfectionRevision
    query = from q in query, select: map(q, [:is_complete, :is_active])
    query = from(q in query,select_merge: %{"$aggregate": %{"$max": %{"revision_id" => max(q.revision_id)}}})
    from(q in query,group_by: q.encounter_id,select_merge: %{"$group": %{"encounter_id" => q.encounter_id}})
will return you 

    #Ecto.Query<from i0 in Data.InfectionRevision, 
    group_by: [i0.encounter_id],
    select: merge(merge(
      map(i0, [:is_complete, :is_active]), 
      %{"$aggregate": %{"$max": %{"revision_id" => max(map(i0, [:is_complete, :is_active]).revision_id)}}}), 
      %{"$group": %{"encounter_id" => map(i0, [:is_complete, :is_active]).encounter_id}})>

But if one key field is dynamic, wrong query is returned `eg` if you try below code sample 

    query = from en in Data.InfectionRevision
    query = from q in query, select: map(q, [:is_complete, :is_active])
    key = "revision_id"
    query = from(q in query,select_merge: %{"$aggregate" => %{"$max" => %{^key => max(q.revision_id)}}})
    key = "encounter_id"
    from(q in query,group_by: q.encounter_id,select_merge: %{"$group" => %{^key => q.encounter_id}})
you will get 

    #Ecto.Query<from i0 in Data.InfectionRevision, group_by: [i0.encounter_id],
      select: merge(merge(
        map(i0, [:is_complete, :is_active]), 
        %{"$aggregate" => %{"$max" => %{^"revision_id" => max(map(i0, [:is_complete, :is_active]).revision_id)}}}), 
        %{"$group" => %{^"revision_id" => map(i0, [:is_complete, :is_active]).encounter_id}})>

[...]

PR #2980, opened 2019-04-17
The subsequent `select_merge` overrides expression bindings

am-kantox:
Steps to reproduce: see the test. Basically, it produces what is shown in red below.
```diff
select:
  merge(
    merge(map(p0, [:title, :body]), %{^"title" => map(p0, [:title, :body]).title}),
+    %{^"body" => map(p0, [:title, :body]).body})
-   %{^"title" => map(p0, [:title, :body]).body})
```
I was unable to introduce the concise solution without bringing many changes, but the below might be treated as a temporary workaround. It seems to happen on a consequtive merges only.

Related to #2979. Originally posted on https://stackoverflow.com/questions/55719195/elixir-ecto-multi-insert-all.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -354,25 +354,9 @@ defmodule Ecto.Query do
   The prefixes set in the query will be preserved when loading data.
   """
 
-  defstruct prefix: nil,
-            sources: nil,
-            from: nil,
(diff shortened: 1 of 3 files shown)


josevalim:
Beautiful, I just started looking into the app. :)

PR #2981, opened 2019-04-17
Exclude schema field in query

marucjmar:
Functionality that allows to exclude a field in the queries. In my opinion, helpful when a field in table  contains a large set of data to serialize


josevalim:

PR #2982, opened 2019-04-20
Add missing :uses_usec tags

DiodonHystrix:
I'm updating sqlite adapter to work with Ecto3.
I noticed that these tests are missing `@tag :uses_usec`, because they are failing as sqlite adapter doesn't support microseconds interval.

I hope it's a correct change.


josevalim:

PR #2984, opened 2019-04-24
Expose the enum that was validated against in errors from enum-based validations

timrogers:
Errors in an `Ecto.Changeset` have a key, a message and additional contextual information provided in the form of a keyword list.

`Ecto.Changeset` comes bundled with a number of validations, for example `validate_inclusion/4`.

The built-in validations tend to include additional contextual information - for example, `validate_length/3` includes `count` and `kind`. So, for example, if the validation was configured to check that a field was "no more than 3" and it fails, then you will have `[count: 3, validation: :length, kind: :max]`.

There are a number of validations that validate a field against an object that implements the `Enumerable` protocol - for example `validate_inclusion/4` which checks that the field's value is included in a supplied enumerable.

However, these validations don't expose in the additional context the enumerable that the value was validated against.

This context can be valuable, for example if you want your errors to be fully machine-readable.

This exposes `[enum: enum_that_the_field_was_validated_against]`
on errors generated by `validate_inclusion/4`, `validate_exclusion/4` and `validate_subset/4`.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1850,7 +1850,7 @@ defmodule Ecto.Changeset do
   def validate_subset(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:subset, data}, fn _, value ->
       case Enum.any?(value, fn(x) -> not(x in data) end) do
-        true -> [{field, {message(opts, "has an invalid entry"), [validation: :subset]}}]
+        true -> [{field, {message(opts, "has an invalid entry"), [validation: :subset, enum: data]}}]
         false -> []
       end
     end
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1828,7 +1828,7 @@ defmodule Ecto.Changeset do
     validate_change changeset, field, {:inclusion, data}, fn _, value ->
       if value in data,
         do: [],
-        else: [{field, {message(opts, "is invalid"), [validation: :inclusion]}}]
+        else: [{field, {message(opts, "is invalid"), [validation: :inclusion, enum: data]}}]
```
timrogers: All of these validations use the `Kernel.in/2` macro, which uses `Enum.member?/2` under the hood, so it only works on values that implement the `Enumerable` protocol - hence the key being called `enum`.

`data` could be a potential alternative name to match the parameter to the validator.

josevalim:

PR #2986, opened 2019-05-01
Fix examples in the docs

ibarchenkov:
Examples in the docs doesn't work because `put_assoc/4` expects a changeset as a first argument.


josevalim:

Issue #2985, opened 2019-04-30
Repo prefix is not being respected in many_to_many association


chobomuffin:
### Environment

* Elixir version (elixir -v):
Erlang/OTP 21 [erts-10.0.6] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.7.3 (compiled with Erlang/OTP 21)

* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):
PostgreSQL 10.5

* Ecto version (mix deps):
ecto: 3.1.2
ecto_sql: 3.1.1

* Database adapter and version (mix deps):
postgrex 0.14.2

* Operating system:
Mac OSX 10.13.6

### Current behavior
Repo prefix is not being respected in many_to_many association. Every other join in the query is being prefixed except the many to many join. It works if I pass in a prefix to the query but it only accepts compile time strings. My prefix is dynamic.

### Expected behavior
I expect it to prefix all of the table names/joins with the prefix I specify in my Repo.all() not just some of them.

josevalim:
Hi Daniel, can you please provide more information? How are you defining
your schemes and associations? How are you passing the prefix? On the
schema? On the query? To the repo? Which query are you running? What is the
Ecto query and what is the logged SQL query? Please try to include as much
info and code samples as possible.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

chobomuffin:
Hey sorry about that, I wasn't sure I was allowed to share the code, but its just a snippet, here it is.

**Query**
```
private_spaces = from s in Space,
      join: a in assoc(s, :assignments),
      join: r in assoc(a, :role),
      join: p in assoc(r, :permissions),
      where: s.sensitive == true and s.type == "global" and a.user_id == ^user_id and p.name == ^Permission.space_read_sensitive
```

**Many to many association on Role**
```
    many_to_many :permissions,
      Permission,
      join_through: "roles_permissions",
      on_replace: :delete
```

**Many to many association on Permission**
```
    many_to_many :roles,
      Role,
      join_through: "roles_permissions",
      on_replace: :delete
```

**SQL**
```
SELECT 
  s0."id", 
  s0."description", 
  s0."type", 
  s0."label", 
  s0."sensitive", 
  s0."status", 
  s0."created_by_id", 
  s0."parent_id", 
  s0."inserted_at", 
[...]

PR #2987, opened 2019-05-02
Fix: prefix assignment for many to many associations

304:
Based on this bug report: https://github.com/elixir-ecto/ecto/issues/2985

Hi, I saw this bug and decided to take a look on it. That's what I found.
`join_through` for many to many associations can be defined using a schema or a table name. 
The problem is that the `rewrite_join_prefix` method for the second case was not defined. Because of that the prefix assignment for many to many table was skipped.

It's my first PR to ecto, please let know if I'm missing something :)


josevalim:
Beautiful!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #2989, opened 2019-05-03
Improve the documentation around defining primary keys

LostKobrakai:
I had a conversion with @ccapndave on slack about primary keys in ecto and we found the last sentence of the `@primary_key` section to be confusing. We both though `false` would refer to `autogenerate: false`. Also the `unless composite keys are defined …` part didn't make matters clearer. 

I tried to make both parts cleaner by shifting the default to the end and only talking about "generation of a primary key" in the context of `@primary_key false` – skipping the part about `field/3` having an option to make an explicitly created field a primary key. 

In favor of removing the complexity there I added a few paragraphs to the "Primary key" section of the docs, which explains when to use `@primary_key` and/or `primary_key: true` on `field/3`.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -187,22 +187,22 @@ defmodule Ecto.Schema do
   auto-generation of the id. This is often the case for primary keys
   in relational databases which are auto-incremented.
 
-  Besides `:id` and `:binary_id`, which are often used by primary
-  and foreign keys, Ecto provides a huge variety of types to be used
-  by any column.
-  
-  There are two ways to define primary keys in ecto: Using the `@primary_key`
+  There are two ways to define primary keys in Ecto: using the `@primary_key`
   module attribute and using `primary_key: true` as option for `field/3` in
   your schema definition. They are not mutually exclusive and can be used
   together.
   
   Using `@primary_key` should be prefered for single field primary keys and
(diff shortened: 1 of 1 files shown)


LostKobrakai:
I'm also wondering if `@primary_key - configures the schema primary key.` is a good initial summary given that it's not the only way to configure the primary key of a schema.

ccapndave:
It might actually be worth adding something clarifying that `autogenerate: true` refers to autogenerating the id when creating a new record and isn't anything to do with generating fields in the schema.

lib/ecto/schema.ex
```diff
@@ -191,10 +190,19 @@ defmodule Ecto.Schema do
...
-  a `@primary_key`, as usual, and then passing the `primary_key: true` option
-  to any of the composite fields.
+  
+  There are two ways to define primary keys in ecto: Using the `@primary_key`
+  module attribute and using `primary_key: true` as option for `field/3` in
+  your schema definition. They are not mutually exclusive and can be used
+  together.
+  
+  Using `@primary_key` should be prefered for single field primary keys and
+  sharing primary key defintions between multiple schemas using macros.
```
fboyer: Small typo, should be "defin**i**tions" instead of "defintions". :tada:

josevalim:

Issue #2991, opened 2019-05-07
Changeset error fields no longer always have values that implement String.Chars


lpil:
This linked change makes it so that validate_inclusion and validate_subset return the enumerable as part of the error. https://github.com/elixir-ecto/ecto/pull/2984

Previously all error fields would implement the `String.Chars` protocol, but now it may not as not all enumerables do (i.e. MapSet in the stdlib).

This change breaks Phoenix's generated `DataCase.errors_on` function. https://github.com/phoenixframework/phoenix/blob/c6dd582ea0a8df321977ebf459afedaffc997689/installer/templates/phx_ecto/data_case.ex#L49

Is this a problem on the Ecto side or the Phoenix side?

It may be a little tricky to fix for Phoenix as there's no longer quite as easy a way to replicate the behaviour of gettext there. Perhaps the enumerable in the error field could be converted to a list?

PR #2992, opened 2019-05-07
Convert `enum` error field to list to always impl String.Chars

lpil:
Returning the enum as is means that the field may not always implement
`String.Chars`, a protocol that downstream code (such as Phoenix)
depended upon.

Closes https://github.com/elixir-ecto/ecto/issues/2991


josevalim:

Issue #2997, opened 2019-05-09
Selecting a map containing a string literal strips the string from select


TylerWitt:
### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 21 [erts-10.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.8.1 (compiled with Erlang/OTP 21)
```

* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):

PostgreSQL 11.2

* Ecto version (mix deps):

3.1.4

* Database adapter and version (mix deps):

Postgrex 0.14.3

* Operating system: 

OS X 10.14.4

### Current behavior

```
 query_1 =
      ProjectA
      |> select([sg], %{type: "project_group", value: sg.id, display: sg.name})

 ProjectB
 |> select([sg], %{type: "project_group", value: sg.id, display: sg.name})
 |> union(^query_1)
 |> Repo.all()
```

We are currently doing a union with a map containing a literal field in the select (similar to what is shown above), but in `planner.ex`, the inner `"project_group"` literal is being completely stripped out of the `ProjectB` select when being converted to a query, resulting in a non-matching number of columns. Example error below:

``` 
** (Postgrex.Error) ERROR 42601 (syntax_error) each UNION query must have the same number of columns
     
         query: SELECT m0."id", m0."name" FROM "project_a" AS m0 UNION (SELECT 'project_group', s0."id", s0."name" FROM "project_b" AS s0 )
```

I was digging around on how to fix this, but I don't think I am familiar enough with the internals.

The field is being removed in `remove_literals/1` within `normalize/4` because the `query.select.fields` entry for that column remains `"project_group"` without any other structure as opposed to the other fields that contain typing information.

I figure the fix is either `collect_fields` should be adding field metadata to the literal, or that `remove_literals` should be ignoring it (which seems wrong), especially because the union is preserving this field when converted to a query.

Restructuring the query as shown below produces no error.

```
val = "project_group"
 query_1 =
      ProjectA
      |> select([sg], %{type: ^val, value: sg.id, display: sg.name})

 ProjectB
 |> select([sg], %{type: ^val, value: sg.id, display: sg.name})
 |> union(^query_1)
 |> Repo.all()
```

### Expected behavior

The query below should either not produce an error and function with project_group being preserved, or an error should be surfaced before the sql is created stating to pass in an interpolated variable instead.

```
 query_1 =
[...]

PR #2998, opened 2019-05-09
Add test for union select fields

TylerWitt:
Adds a test for union select fields to verify that they are normalized equally.

I'm not quite sure if this is the cause of the missing column from unions in https://github.com/elixir-ecto/ecto/issues/2997, but I did confirm it is behaving differently from 3.0.0 (test passes) and 3.1.4 (test fails).


josevalim:
Great job! 👏  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #3001, opened 2019-05-13
Allow handling nil values in custom type

basbz:
I had a similar type as mentioned: in https://github.com/elixir-ecto/ecto/issues/2576
Although the documentation clearly states that `nil` is bypassed, it's not clear why?
It is clearly an edge case, but  I'd still like to propose an _opt in_ mechanism to allow custom types to handle nil values if they really want to.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -473,6 +473,10 @@ defmodule Ecto.Type do
     load_embed(embed, value, &load/2)
   end
 
+  def load(mod, nil) when is_atom(mod) do
+    if can_handle_nil?(mod), do: mod.load(nil), else: {:ok, nil}
+  end
+
   def load(_type, nil) do
     {:ok, nil}
   end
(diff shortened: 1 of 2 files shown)


josevalim:
The problem is that this change makes everything else slower since we are adding an extra check on all operations and loading all of the data is often the hotspot when loading thousands of entries at the same. It is unlikely we are going to change the behaviour regarding nil. Why does nil have to be handled in your case?

basbz:
> The problem is that this change makes everything else slower
Hm, ok that makes sense... I didn't think of that kind of scale

I really like how Ecto can transform data and safeguard the boundaries in my program.
I'd like to transform nil values into a maybe type
https://package.elm-lang.org/packages/elm/core/latest/Maybe
https://hexdocs.pm/algae/Algae.Maybe.html
So it  the have to be dealt with explicitly. 
Any pointers on getting there in another way, would be appreciated

josevalim:
Hrm, I see. Unfortunately the `nil` assumption is ingrained in Ecto. There are places in changesets and associations that check for `nil` and change this will break it and make Ecto do things it is not supposed to. So I don't think a custom type would be enough, sorry.

basbz:
No problem, at least I have a better understanding of the choices made now.

PR #3003, opened 2019-05-17
Support {:fragment, ...} in assoc :where

lnikkila:
Before 3.0.5, we were using dynamic expressions with the `:where` option for filtering out associations based on the current timestamp. In my opinion this was really useful for auto-expiring rows, soft deletes, etc. After updating to 3.0.5, I think it’s a little too easy to forget to filter those rows out at query/preload time when it’s almost always needed.

Based on comments in earlier PRs, I’m under the impression the feature was removed because of compilation time problems related to the MFA tuples. (Please correct me if I’m wrong!) I’d like to propose another way to define dynamic associations at compilation time using fragments instead of dynamic expressions:

```
  where: [expires_at: {:fragment, "? > now()"}]
```

I feel like this filter would be a fairly natural extension to the existing filters like `{:in, [...]}`. A single fragment argument should cover most use cases, and more complicated filtering logic could be pushed down to the database by defining functions. (Although I don’t think doing anything more complicated is necessarily a good idea.)

I’m open to doing the work if you feel like this could be worth exploring, however I could use some guidance with how this should be handled internally because I’m not super familiar with the codebase. I hacked together something that seems to work, but crossing the `Ecto.Query.Builder` boundary feels gross and I think this needs some type checks.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -293,6 +293,11 @@ defmodule Ecto.Association do
           expr = {:and, [], [expr, {:not, [], [{:is_nil, [], [to_field(binding, key)]}]}]}
           {expr, params, counter}
 
+        {key, {:fragment, frag}}, {expr, params, counter} when is_binary(frag) ->
+          pieces = Ecto.Query.Builder.fragment_pieces(frag, [to_field(binding, key)])
+          expr = {:and, [], [expr, {:fragment, [], pieces}]}
+          {expr, params, counter}
+
         {key, {:in, value}}, {expr, params, counter} when is_list(value) ->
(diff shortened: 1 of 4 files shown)


josevalim:
I like this a lot, implementation included. The builder is invoked from
different parts in the code base, so that’s fine IMO. This PR looks good to
be merged, we just need docs!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

lnikkila:
Well that was easier than I expected... :smile: I just added a mention to the docs.

josevalim:

Issue #2832, opened 2018-11-20
Dynamically select named binding


LostKobrakai:
### Environment

* Elixir version (elixir -v): 
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):
* Ecto version (mix deps): ecto 3.0.2, ecto_sql 3.0.1
* Database adapter and version (mix deps): postgrex 0.14.0
* Operating system:

### Current behavior

When trying to select a named binding dynamically for a `Ecto.Query.API` macro the only way to do it currently seems to be using the `[{as, position}]` syntax:

```elixir
position = Map.get(query.aliases, binding_name)
select_merge(query, [{x, position}], %{id => x.id})
```

```elixir
select_merge(query, [{x, binding_name}], %{id => x.id}) // doesn't work
# ** (UndefinedFunctionError) function :binding_atom.__schema__/2 is undefined
```

### Expected behavior

- Some way to just do e.g. `[{x, binding_name}]` and it doing the position detection internally.
- Better error message: 
`(Ecto.Query.CompileError) binding list should contain only variables or {as, var} tuples, got: {^binding, x}`  
`{as, var}` is not really descriptive if it's more of a `{as, position}`
- Documenting the `[{as, position}]` syntax. I only found out how it's supposed to work by reading the tests after getting the above error message, which hinted there's some form of not hardcoded keywork list.

josevalim:
For now I would go with a better error message. :)

kenken1mura:
delet

kenken1mura:
delet

kenken1mura:
delet

qcam:
@LostKobrakai Pardon me if I'm wrong but shouldn't it be?

```ex
select_merge(query, [{binding_name, x}], %{id => x.id})
```

LostKobrakai:
If you want to refer to `[binding_name: x]`, yes yours seems like the better idea. But ecto internally doesn't use a keyword list. It uses positional bindings (binding -> position) and a separate mapping for name -> position. So that's why `[{x, position}]` was already existing (even before ecto 3) as a way to get the binding at position `position` as `x`. So it's probably more appropiate to use `[{x, binding_name}]`, where binding_name is just another for of saying, which binding to use, as it's more in line with existing behavior.

qcam:
That's fair. But as you already pointed out `{x, position}` is an internal implementation and isn't publicly documented 🤔. So this can be changed without notice, depending on this is kinda risky, isn't it?

[...]

PR #3005, opened 2019-05-21
Implement dynamically selected named binding

LostKobrakai:
Regarding #2832

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -743,6 +743,9 @@ defmodule Ecto.Query.Builder do
     end
   end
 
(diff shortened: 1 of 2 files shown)


AndrewDryga:
Thank you, this simplifies few places for us 🎊.

LostKobrakai:
Yeah, the part when you accidentally work on an old master and only rebase at the end. Seems like this is already in ecto.

AndrewDryga:
@LostKobrakai I don't think that it's supported for all the cases. I believe it's still not supported for joins and/or preloads.

LostKobrakai:
At least https://github.com/elixir-ecto/ecto/pull/3005/files#diff-c04711315c01e29757522ea2764fd871R755 is pretty much what I added.

PR #3010, opened 2019-06-03
Add HOW-TO's and rearrange guides' structure

msaraiva:
This is the first suggestion for the new HOW-TO's section. We basically do two things:

  * Add the first 6 HOW-TO guides extracted/adapted from the "What's new in Ecto 2.1" book
  * Split the guides into 2 parts, "Introduction" and "How-to's"


guides/howtos/Multi tenancy with query prefixes.md
```diff
@@ -0,0 +1,253 @@
...
+    field :client_id, :integer
+    field :db_prefix
+    timestamps
+  end
+end
+```
+
+Now running `MyApp.Repo.all MyApp.Mapping` will by default run on the "main" prefix, regardless of the value configured globally on the `:after_connect` callback. Similar will happen to `insert`, `update`, and similar operations, the `@schema_prefix` is used unless the `:prefix` is explicitly changed via `Ecto.put_meta/2` or by passing the `:prefix` option to the repository operation.
+
+Keep in mind, however, that queries run on a single prefix. For example, if `MyApp.Mapping` on prefix "main" depends on a schema named `MyApp.Other` on prefix "another", a query starting with `MyApp.Mapping` will always run on the "main" prefix. By design it is not possible to perform query joins across prefixes. If data belongs to different prefixes, it is best to not couple them structurally nor via queries, in order to **keep data in different prefixes isolated**.
```
josevalim: This is no longer true. Ecto 3.0 allows you to do this:

```elixir
from p in Post, prefix: "foo",
  join: c in Comment, prefix: "bar"
```

We need to update this chapter to add a new section. Here are the docs: https://hexdocs.pm/ecto/Ecto.Query.html#module-query-prefix

guides/howtos/Multi tenancy with query prefixes.md
```diff
@@ -0,0 +1,253 @@
...
+
+Query prefixes may be useful in different scenarios. For example, multi tenant apps running on Postgres would define multiple prefixes, usually one per client, under a single database. The idea is that prefixes will provide data isolation between the different users of the application, guaranteeing either globally or at the data level that queries and commands act on a specific prefix.
+
+Prefixes may also be useful on high-traffic applications where data is partitioned upfront. For example, a gaming platform may break game data into isolated partitions, each named after a different prefix. A partition for a given player is either chosen at random or calculated based on the player information.
+
+While query prefixes were designed with the two scenarios above in mind, they may also be used in other circumstances, which we will explore throughout this chapter. All the examples below assume you are using Postgres. Other databases engines may require slightly different solutions.
+
+## Global prefixes
```
josevalim: Let's call this `Connection prefixes`.

josevalim: We will have to change the rest of the text accordingly.

guides/howtos/Schemas and changesets.md
```diff
@@ -0,0 +1,121 @@
+# Schemas and changesets
```
josevalim: I would suggest changing this chapter to: "Data mapping and validation"

mix.exs
```diff
@@ -93,4 +92,24 @@ defmodule Ecto.MixProject do
...
+      "guides/howtos/Aggregates and subqueries.md",
+      "guides/howtos/Multi tenancy with query prefixes.md",
+      "guides/howtos/Test factories.md"
+    ]
+  end
+
+  defp groups_for_extras do
+    [
+      "Introduction": ~r/guides\/introduction\/.?/,
+      "HOW-TO'S": ~r/guides\/howtos\/.?/
```
josevalim:
```suggestion
      "How-To's": ~r/guides\/howtos\/.?/
```

guides/howtos/Schemas and changesets.md
```diff
@@ -0,0 +1,121 @@
...
+
+We will take a look at the role schemas play when validating and casting data through changesets. As we will see, sometimes the best solution is not to completely avoid schemas, but break a large schema into smaller ones. Maybe one for reading data, another for writing. Maybe one for your database, another for your forms.
+
+## Schemas are mappers
+
+[The Ecto documentation](https://hexdocs.pm/ecto/Ecto.Schema.html) says:
```
josevalim: We should rewrite any links to Ecto itself as regular code links. For example, here we could say:

> The `Ecto.Schema` moduledoc says:

guides/howtos/Aggregates and subqueries.md
```diff
@@ -0,0 +1,96 @@
...
+#=> #Decimal<1743>
+```
+
+Behind the scenes, the query above translates to:
+
+```elixir
+MyApp.Repo.one(from p in MyApp.Post, select: avg(p.visits))
+```
+
+The `aggregate/3` function supports any of the aggregate operations listed in the [Ecto Query API](https://hexdocs.pm/ecto/Ecto.Query.API.html).
```
josevalim:
```suggestion
The `c:Ecto.Repo.aggregate/3` function supports any of the aggregate operations listed in the `Ecto.Query.API` module.
```

ExDoc will take care of linking.

josevalim:
Hi @msaraiva! All the guides with the exception of the query prefixes look good. I have added only minor comments. For the query prefix one, I think we need to a major overhaul. In particular, in Ecto 2.0 we only had a single prefix for the whole query, but that is no longer the case in Ecto 3.0. For each join/from, the prefix used is determined by this order:

1. If the `prefix` option is given exclusively to join/from
2. If the `@schema_prefix` is set
3. If the `:prefix` field is set on the query (i.e. `%{query | prefix: prefix}`) or to the repo operation (i.e. `Repo.all query, prefix: prefix`)
4. The connection prefix (which we call global prefix but I would change it to connection)

We can either:

1. Remove this page for now and merge the other guides
2. Fix this page and merge it altogether :)

msaraiva:
Let's try option 2 :)

josevalim:

PR #3014, opened 2019-06-06
Add many-to-many how-tos

msaraiva:
Guides extracted/adapted from "What's new in Ecto 2.1" ebook:

  * Many to many and casting
  * Many to many and upserts


guides/howtos/Many to many and casting.md
```diff
@@ -0,0 +1,383 @@
+# Many to many and casting
+
+Besides `belong_to`, `has_many`, `has_one` and `:through` associations, Ecto also includes `many_to_many`. `many_to_many` relationships, as the name says, allows a record from table X to have many associated entries from table Y and vice-versa. Although `many_to_many` associations can be written as `has_many :through`, using `many_to_many` may considerably simplify some workflows.
```
josevalim:
```suggestion
Besides `belongs_to`, `has_many`, `has_one` and `:through` associations, Ecto also includes `many_to_many`. `many_to_many` relationships, as the name says, allows a record from table X to have many associated entries from table Y and vice-versa. Although `many_to_many` associations can be written as `has_many :through`, using `many_to_many` may considerably simplify some workflows.
```

guides/howtos/Many to many and casting.md
```diff
@@ -0,0 +1,383 @@
...
+Differently from `has_many :through`, `many_to_many` associations are also writeable. This means we can send data through our forms exactly as we did at the beginning of this guide:
+
+```elixir
+%{"todo_list" => %{
+  "title" => "shipping list",
+  "todo_items" => %{
+    0 => %{"description" => "bread"},
+    1 => %{"description" => "eggs"},
+  }
+}}
```
josevalim: We should fix the nesting here so each  closing `}` is in its own line.

guides/howtos/Many to many and casting.md
```diff
@@ -0,0 +1,383 @@
+# Many to many and casting
```
josevalim: Perhaps a better name is:

* Polymorphic associations with many to many

or:

* Casting polymorphic associations with many to many

I prefer the first. What do you think?

guides/howtos/Many to many and upserts.md
```diff
@@ -0,0 +1,207 @@
...
+
+By adding the unique index and then using `get_by`  with a `insert!` to get or insert a tag, we introduced a potential error in our application. If two posts are submitted at the same time with a similar tag, there is a chance we will check if the tag exists at the same time, leading both submissions to believe there is no such tag in the database. When that happens, only one of the submissions will succeed while the other one will fail. That's a race condition: your code will error from time to time, only when certain conditions are met. And those conditions are time sensitive.
+
+Many developers have a tendency to think such errors won't happen in practice or, if they happened, they would be irrelevant. But in practice they often lead to very frustrating user experiences. I have heard a first-hand example coming from a mobile game company. In the game, a player is able to play quests and on every quest you have to choose a guest character from another player out of a short list to go on the quest with you. At the end of the quest, you have the option to add the guest character as a friend.
```
josevalim: I would remove this "story". Let's say "Luckily Ecto gives us a mechanism to handle constraint errors from the database." here and skip to the next section.

guides/howtos/Many to many and upserts.md
```diff
@@ -0,0 +1,197 @@
...
+  add :tag_id, references(:tags)
+end
+```
+
+Note we added a unique index to the tag name because we don't want to have duplicated tags in our database. It is important to add an index at the database level instead of using a validation since there is always a chance two tags with the same name would be validated and inserted simultaneously, passing the validation and leading to duplicated entries.
+
+Now let's also imagine we want the user to input such tags as a list of words split by comma, such as: "elixir, erlang, ecto". Once this data is received in the server, we will break it apart into multiple tags and associate them to the post, creating any tag that does not yet exist in the database.
+
+While the constraints above sound reasonable, that's exactly what put us in trouble with `cast_assoc/3`. The `cast_assoc/3` changeset function was designed to receive external parameters and compare them with the associated data in our structs.To do so correctly, Ecto requires tags to be sent as a list of maps. We saw an example of this in the [Many to many and casting guide](many-to-many-and-casting.html). However here we expect tags to be sent in a string separated by comma.
```
josevalim: If you change the other guide title, please fix the link here.

guides/howtos/Many to many and upserts.md
```diff
@@ -0,0 +1,197 @@
+# Many to many and upserts
```
josevalim: I have done some changes to this chapter to focus on Constraints and Upserts. Therefore I suggest to rename it exactly to "Constraints and Upserts" :) Thoughts?

mix.exs
```diff
@@ -101,6 +101,8 @@ defmodule Ecto.MixProject do
       "guides/howtos/Schemaless queries.md",
       "guides/howtos/Dynamic queries.md",
       "guides/howtos/Aggregates and subqueries.md",
+      "guides/howtos/Many to many and casting.md",
+      "guides/howtos/Many to many and upserts.md",
```
josevalim: Perhaps we should keep the How-Tos alphabetically ordered?

guides/howtos/Many to many and casting.md
msaraiva: I like the first one too. Especially because it's shorter. It still gets part of it cut off along with a couple of children, but I'm not sure there's much we can do about it.

image

guides/howtos/Many to many and upserts.md
msaraiva: Sounds good!

josevalim:

PR #3023, opened 2019-06-11
Add custom composite types

nicholasjhenry:
For example, it allows a schema with:

```
schema "my schema" do
  field :field_a, {CustomType, :string}
end
```

There was originally an issue for this https://github.com/elixir-ecto/ecto/issues/1871#issuecomment-302340748:

I'm looking for early feedback before I proceed any further to make sure that I'm heading in the right direction and that a PR for this feature would be accepted. Thank you!

### DIFF
--- a/integration_test/support/types.exs
+++ b/integration_test/support/types.exs
@@ -14,3 +14,15 @@ defmodule Custom.Permalink do
   def load(integer) when is_integer(integer), do: {:ok, integer}
   def dump(integer) when is_integer(integer), do: {:ok, integer}
 end
+
+defmodule Custom.Composite do
+  def type, do: :any
+
+  def cast(nil), do: {:ok, :none}
+  def cast(value), do: {:ok, {:some, value}}
+
+  def load(nil), do: {:ok, :none}
+  def load(value), do: {:ok, {:some, value}}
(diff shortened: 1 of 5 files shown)


josevalim:
Hi @nicholasjhenry!

We had discussions about this in the past and it happens that the composite type is likely a more specific version of a parameterized type (where the parameter is another type). So it would be preferable to add a parameterized type - which I would recommend as next step.

Then the next step is to make the current embeds parameterized types too and that's usually the part we have struggled with. I believe a fresh take from you would be really welcome but I just want to warn upfront this is potentially a long journey. :)

josevalim:
Hi @nicholasjhenry, I will go ahead and close this one, as there are other things we need to explore before adding a similar feature. If you have any new insights, please let us know!

nicholasjhenry:
Hello @josevalim, sorry for my tardy reply, I just got back from vacation. I'll review your comment (https://github.com/elixir-ecto/ecto/pull/3023#issuecomment-501035891) and I'll ask some follow up questions. When I'm ready, shall I open a new issue or continue the discussion here? Thank you.

josevalim:
Both options are fine. We can discuss it in here for now!
-- 

PR #3024, opened 2019-06-13
fix docs typo 'migrations' -> 'schemas'

vyachkonovalov:


wojtekmach:
The docs are correct, we dont declare this as type in schema because its already an embed.

Issue #3026, opened 2019-06-14
Ecto.Query.exclude doesn't remove aliases


fuelen:
### Precheck

Ecto.Query.exclude doesn't remove aliases, so I can't add new join with the same name.

### Environment

* Elixir version (elixir -v): 1.8.1
* Ecto version (mix deps): 3.1.5

### Current behavior
```
> import Ecto.Query
Ecto.Query

> query = from u in "users", where: u.id == "1", inner_join: p in "posts", on: p.user_id == u.id, as: :posts
#Ecto.Query<from u0 in "users", join: p1 in "posts", as: :posts,
 on: p1.user_id == u0.id, where: u0.id == "1">

> from [u] in exclude(query, :inner_join), inner_join: p in "posts", on: p.user_id == u.id, as: :posts                                 
** (Ecto.Query.CompileError) alias `:posts` already exists

> exclude(query, :inner_join).aliases                                                                   
%{posts: 1}
```
### Expected behavior
remove appropriate aliases

josevalim:
@fuelen can you please send a pull request? Thank you!

fuelen:
Yes, I'll do that

PR #3027, opened 2019-06-15
Remove named binding when excluding joins

fuelen:
https://github.com/elixir-ecto/ecto/issues/3026


josevalim:
Looks good! One change is necessary though. The “from” part can also have
an alias, so in the case we remove all joins, we need to keep the from
alias.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

fuelen:
@josevalim thanks, updated

PR #3030, opened 2019-06-20
Assign repo_opts to changeset to make it available down the chain

humancopy:
I'm using [Triplex](https://hexdocs.pm/triplex/readme.html) for multi-tenancy and I had an issue with the [Pow email confirmation extension](https://hexdocs.pm/pow/pow_email_confirmation.html) where it [tries to get the prefix using the repo_opts option on the changeset](https://github.com/danschultzer/pow/blob/e3f8a0638dd4ad32f61f89cf2522cc1bd81e6f5e/lib/extensions/email_confirmation/ecto/schema.ex#L77) for validating an email is unique but Ecto doesn't pass it down the line. This commit solves this.

Otherwise you need to do something like this:

```elixir
user
|> User.changeset(attrs)
|> Map.put(:repo_opts, [prefix: tenant])
|> Repo.update(prefix: tenant)
```

Instead of this
```elixir
user
|> User.changeset(attrs)
|> Repo.update(prefix: tenant)
```

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -231,7 +231,7 @@ defmodule Ecto.Repo.Schema do
 
     # On insert, we always merge the whole struct into the
     # changeset as changes, except the primary key if it is nil.
-    changeset = put_repo_and_action(changeset, :insert, repo)
+    changeset = put_repo_and_action(changeset, :insert, repo, opts)
     changeset = surface_changes(changeset, struct, fields ++ assocs)
 
     wrap_in_transaction(adapter, adapter_meta, opts, changeset, assocs, embeds, prepare, fn ->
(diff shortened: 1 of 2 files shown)


josevalim:
Note that this should work though:

```elixir
user
|> User.changeset(attrs)
|> Map.put(:repo_opts, [prefix: tenant])
|> Repo.update()
```

Have you tried it?

In any case, this looks good, we just need tests. :)

humancopy:
No I haven't tried but yeah I saw in the code that this should work actually 😄 

I wasn't sure about the tests because it's a private method. Could you give me an idea please how to test this?

josevalim:
In this test: https://github.com/elixir-ecto/ecto/blob/master/test/ecto/repo_test.exs#L704

They are already matching on action and repo, so you need to match on repo_opts too.

humancopy:
I added the tests, let me know if it's ok :)

josevalim:
Beautiful, thank you!

Issue #3021, opened 2019-06-11
where with a literal keyword list allows interpolating nils


wojtekmach:
### Environment

* Elixir version (elixir -v): master
* Ecto version (mix deps): master

### Current behavior

Ecto prevents this:

```
iex(3)> from(p in Package, where: [id: nil])
** (Ecto.Query.CompileError) nil given for :id. Comparison with nil is forbidden as it is unsafe. Instead write a query with is_nil/1, for example: is_nil(s.id)
```

and this:

```
iex(3)> from(p in Package, where: p.id == nil)
** (Ecto.Query.CompileError) comparison with nil is forbidden as it is unsafe. If you want to check if a value is nil, use is_nil/1 instead
```

and this:

```
iex(3)> id = nil; from(p in Package, where: p.id == ^id)
** (ArgumentError) comparison with nil is forbidden as it is unsafe. If you want to check if a value is nil, use is_nil/1 instead
```

and this:

```
iex(3)> kw = [id: nil]; from(p in Package, where: ^kw)
** (ArgumentError) nil given for :id. Comparison with nil is forbidden as it is unsafe. Instead write a query with is_nil/1, for example: is_nil(s.id)
```

but not this:

```
iex(4)> id = nil; from(p in Package, where: [id: ^id])
#Ecto.Query<from p in Hexpm.Repository.Package, where: p.id == ^nil>
```

### Expected behavior

nil should be prevented.

josevalim:
Since this can potentially break bad user code, I think we should postpone it to v3.2. The saving grace is that the query we actually emit is useless.

fsanggang:
I'd like to have a go at this if that's okay?

josevalim:
Yes please!
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

PR #3031, opened 2019-06-21
Don't allow interpolated nil in literal keyword list when building query

fsanggang:
#3021

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -78,6 +78,11 @@ defmodule Ecto.Query.Builder do
   end
 
   # param interpolation
+  def escape({:^, _, [nil]}, _type, {_params, _acc}, _vars, _env) do
+    error! "comparison with nil is forbidden as it is unsafe. " <>
+             "If you want to check if a value is nil, use is_nil/1 instead"
+  end
+
   def escape({:^, _, [arg]}, type, {params, acc}, _vars, _env) do
     expr = {:{}, [], [:^, [], [length(params)]]}
     params = [{arg, type} | params]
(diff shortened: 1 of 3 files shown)


fsanggang:
I believe this fixes the specific issue in #3021, namely interpolated nils in keyword lists. But as the tests show, there is a side-effect of causing another kind of nil comparison to raise a compilation error instead of an argument error, and I'm not sure that's acceptable? 🤔

michalmuskala:
I don't think this is all the way in a place where we'd want the check to happen, since it would only apply to actual `nil` literals - the test we'd want to check it with, should be something along the lines of:

```elixir
var = nil
from p in Post, where: [id: ^var]
```

That's probably need to expand into some sort of a runtime check for `nil` like we do in other places.

josevalim:
The proper solution is to change this clause:

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder/filter.ex#L26

to something like this:

```elixir
{field, in_value}, params when is_atom(field) ->
  {out_value, {params, :acc}} = Builder.escape(in_value, {binding, field}, {params, :acc}, vars, env)
  value = maybe_check_for_nils(in_value, out_value, field)
  {{:{}, [], [:==, [], [to_escaped_field(binding, field), value]]}, params}
```

where:

```
defp maybe_check_for_nils({:^, _, [_]}, out_value, field) do
  quote do
    Ecto.Query.Builder.Filter.not_nil!(unquote(out_value), field)
  end
end

defp maybe_check_for_nils(_, out_value, _field) do
  out_value
end
```

And the `not_nil` should check if the value is `nil` or not and raise a meaningful error message.

fsanggang:
Thank you for the tips and pointers!

josevalim:
Merged locally, thank you! 

PR #3034, opened 2019-06-21
Correct type signature of Changeset.cast

tomekowal:
The implementation of `Changeset.cast` explicitly checks that all keys in the `permitted` list are atoms https://github.com/elixir-ecto/ecto/blob/fb6bae393e7522d0e0d514fc2e8b2930cccd62dd/lib/ecto/changeset.ex#L489 and raise otherwise.
Type spec allowed either atoms or strings there which is incorrect. I got confused today because it appears in the generated documentation https://hexdocs.pm/ecto/Ecto.Changeset.html#cast/4
This PR fixes the type spec.


michalmuskala:
Thank you! ❤️

It seems it's a left over from a time when we did support both.

Issue #3028, opened 2019-06-18
Repo prefix is not being respected in union queries


eraf2135:
### Environment

* Elixir version: 1.7.3 (compiled with Erlang/OTP 21)
* Database and version: (PostgreSQL) 10.6
* Ecto version: 3.1.5 
* Database adapter and version: postgrex 0.14.3 
* Operating system: MacOS 10.14.4 

### Current behavior

Repo prefix is not being respected in union queries. e.g. given a query like:
```
from c in query,
    union: ^union_query
```
the prefix will be applied to the `query` portion but not to the `union_query` portion.
It works if I pass in a prefix to each join in `union_query` but this is only appropriate for compile time strings and my prefix is dynamic.

### Expected behavior
I expect it to prefix all of the tables in union subqueries with the prefix I specify in my Repo.all().

josevalim:
Closing in favor of #3039.

PR #3039, opened 2019-06-26
Fix: attach prefix for union queries

304:
Fix for https://github.com/elixir-ecto/ecto/issues/3028

The problem is that the prefix from `Repo.all(query, prefix: prefix)`
is applied only to the main query. Union queries have a different
implementation and the prefix attachment is ignored for them.

I'm not 100% sure if it is a proper solution, so a review is really welcome :)

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -743,7 +743,7 @@ defmodule Ecto.Query.Planner do
   defp plan_combinations(query, adapter) do
     combinations =
       Enum.map query.combinations, fn {type, combination_query} ->
-        {prepared_query, _params, _key} = plan(combination_query, :all, adapter)
+        {prepared_query, _params, _key} = plan(%{combination_query | prefix: query.prefix}, :all, adapter)
         prepared_query = prepared_query |> ensure_select(true)
         {type, prepared_query}
       end
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -743,13 +743,16 @@ defmodule Ecto.Query.Planner do
   defp plan_combinations(query, adapter) do
     combinations =
       Enum.map query.combinations, fn {type, combination_query} ->
-        {prepared_query, _params, _key} = plan(%{combination_query | prefix: query.prefix}, :all, adapter)
+        {prepared_query, _params, _key} = combination_query |> attach_prefix(query) |> plan(:all, adapter)
         prepared_query = prepared_query |> ensure_select(true)
         {type, prepared_query}
       end
 
     %{query | combinations: combinations}
   end
+
+  defp attach_prefix(%{prefix: nil} = query, %{prefix: prefix}), do: %{query | prefix: prefix}
+  defp attach_prefix(query, _), do: query
 
   defp plan_ctes(%Ecto.Query{with_ctes: nil} = query, _adapter), do: query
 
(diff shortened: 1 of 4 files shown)


michalmuskala:
Unfortunately that simple solution won't work due to some complexities in query planner - try adding two similar queries like that in a single test - it should blow up.

304:
@michalmuskala, thanks for review! 

I've tried several options, but still I was not able to reproduce the case you mentioned. 
Can you please clarify what do you mean by "adding two similar queries in a single test"?  
It would help me a lot to proceed looking into this issue, if I had a failing test as a starting point.

michalmuskala:
I meant a test case similar to this:
```elixir
    test "two queries" do
      query = from(MySchema, union: ^from(MySchema), union_all: ^from(MySchema))

      TestRepo.all(query, prefix: "public")
      assert_received {:all, _query}

      TestRepo.all(query, prefix: "private")
      assert_received {:all, _query}
    end
```

304:
@michalmuskala here is an implementation of a test you suggested 519c618, but there are still no failures in the test suite 🤔 

While working on that, I figured out that `assert_received` actually rebinds a variable provided to it.
Let's say, I define `query = from(...)` first and later use `assert_received {:all, query}`. Then after this call the `query` value won't be an original one anymore. And if I try to provide this *modified* `query` to `TestRepo.all` second time, it won't work as expected. Maybe that's the issue you was facing before?

304:
@josevalim can you please also take a look?

lib/ecto/repo/queryable.ex
```diff
@@ -128,7 +128,14 @@ defmodule Ecto.Repo.Queryable do
...
   defp attach_prefix(query, opts) do
     case Keyword.fetch(opts, :prefix) do
-      {:ok, prefix} -> %{query | prefix: prefix}
+      {:ok, prefix} ->
+        combinations =
+          Enum.map(query.combinations, fn {type, combination_query} ->
+            {type, %{combination_query | prefix: prefix }}
+          end)
+
+        %{query | prefix: prefix, combinations: combinations}
```
josevalim: If the combination itself already has a prefix though, we should most likely respect it. Also, this needs to be done recursively. The combination itself may have other combinations. I think the best place to apply this is inside Ecto.Query.Planner.

304: @josevalim thanks for the suggestion! I've moved logic of prefix attachment for combinations  into a query planner and I've added more tests to cover, for example, the recursive case you mentioned.

lib/ecto/query/planner.ex
```diff
@@ -743,7 +743,7 @@ defmodule Ecto.Query.Planner do
   defp plan_combinations(query, adapter) do
     combinations =
       Enum.map query.combinations, fn {type, combination_query} ->
-        {prepared_query, _params, _key} = plan(combination_query, :all, adapter)
+        {prepared_query, _params, _key} = plan(%{combination_query | prefix: query.prefix}, :all, adapter)
```
josevalim: If a prefix is set, we need to respect it though. it should be something like this:

```elixir
defp attach_prefix(%{prefix: nil} = query, %{prefix: prefix}), do: %{query | prefix: prefix}
defp attach_prefix(query, _), do: %{query | prefix: prefix}
```

And we should be good to go!

304: 👍 
I've made a requested change here 68f7c5b1540d06102f08ef306c9c48314ceaf78b

josevalim:

josevalim:
@304 would you also be so kind to submit a similar PR to `plan_cte`? The solution and tests should be very similar.

PR #3045, opened 2019-06-28
UUID cast error

kbaird:
We encountered the following error with UUID casting:
```elixir
[error] #PID<0.1267.0> running OurAppWeb.Endpoint (connection #PID<0.1266.0>, stream id 1) terminated
Server: localhost:4000 (http)
Request: GET /api/v1/redacted_resources
    ** (UndefinedFunctionError) function :uuid.cast/1 is undefined (module :uuid is not available)
        :uuid.cast("a72e8643-5f6a-4882-a419-88be227479d2")
        (ecto_sql) lib/ecto/adapters/sql.ex:554: anonymous fn/2 in Ecto.Adapters.SQL.load_embed/2
        (ecto) lib/ecto/type.ex:1159: Ecto.Type.map/4
        (ecto) lib/ecto/type.ex:800: Ecto.Type.process_loaders/3
        (ecto) lib/ecto/repo/queryable.ex:339: Ecto.Repo.Queryable.struct_load!/6
        (ecto) lib/ecto/repo/queryable.ex:201: anonymous fn/5 in Ecto.Repo.Queryable.preprocessor/3
        (elixir) lib/enum.ex:1327: Enum."-map/2-lists^map/1-0-"/2
        (ecto) lib/ecto/repo/queryable.ex:158: Ecto.Repo.Queryable.execute/4
        (ecto) lib/ecto/repo/queryable.ex:18: Ecto.Repo.Queryable.all/3
        (scrivener_ecto) lib/scrivener/paginater/ecto/query.ex:28: Scrivener.Paginater.Ecto.Query.paginate/2
...
```

This PR appears to fix it.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -730,6 +730,7 @@ defmodule Ecto.Type do
   defp cast_fun({:in, type}), do: &array(&1, cast_fun(type), [])
   defp cast_fun({:array, type}), do: &array(&1, cast_fun(type), [])
   defp cast_fun({:map, type}), do: &map(&1, cast_fun(type), %{})
+  defp cast_fun(:uuid), do: &Ecto.UUID.cast(&1)
   defp cast_fun(mod) when is_atom(mod), do: &mod.cast(&1)
 
   defp cast_integer(term) when is_binary(term) do
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -80,19 +80,33 @@ defmodule Ecto.Type do
   import Kernel, except: [match?: 2]
 
   @typedoc "An Ecto type, primitive or custom."
-  @type t         :: primitive | custom
+  @type t :: primitive | custom
 
   @typedoc "Primitive Ecto types (handled by Ecto)."
   @type primitive :: base | composite
 
   @typedoc "Custom types are represented by user-defined modules."
-  @type custom    :: module
+  @type custom :: module
 
-  @typep base      :: :integer | :float | :boolean | :string | :map |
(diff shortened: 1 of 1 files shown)


lib/ecto/type.ex
```diff
@@ -677,6 +697,9 @@ defmodule Ecto.Type do
       iex> cast(:string, [1, 2, 3])
       :error
 
+      iex> cast(:uuid, "240d1a54-2f3d-43a4-85b9-3ef5ac702feb")
+      {:ok, "240d1a54-2f3d-43a4-85b9-3ef5ac702feb"}
+
```
kbaird: Every other change in this file appears to stem from `mix format` config changes.

Lines 700-702 above are the only intended changes for this commit.

josevalim:
Ecto doesn’t have  a :uuid type as a built-in. Ecto.UUID must be used as the the type instead.

josevalim:

kbaird:
OK. Just as a heads-up, we are using `Ecto.UUID` as the type in our app.

Specifically, we're using `Ecto.UUID` as a schema field type, and I suspect it's getting converted to `:uuid` inside the stack via `Ecto.Type.type(Ecto.UUID)`. I can make this error go away by using `:binary_id` as the schema field type, but would prefer `Ecto.UUID`.

I will keep investigating and try to make a different PR that is acceptable.

josevalim:
Btw, my previous response was innacurate, `:uuid`is a primitive Ecto.Type, but it works at the byte level.

kbaird:
Precursor to https://github.com/elixir-ecto/ecto/issues/3046. Nothing further on this.

Issue #3046, opened 2019-06-29
Ecto.UUID type() return causes a composite schema type casting error.


joeyrosztoczy:
### Precheck [Done]

We've [also submitted a PR here](https://github.com/elixir-ecto/ecto/pull/3047) describing what we believe to be the most appropriate fix.

### Environment

* Elixir version (elixir -v): 1.8.1 (but bug is fairly version independent)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 11.2 (but also version independent)
* Ecto version (mix deps): 3.1.7
* Database adapter and version (mix deps): postgrex
* Operating system: mac osx

### Current behavior

When trying to cast an Ecto.UUID inside of a {:map, :inner_type} schema type, ecto throws the following error: 

`:uuid.cast/1 is undefined (module :uuid is not available)`

Here is a minimal reproducible error case which can be triggered by installing deps and running `$ mix test`:
https://github.com/joeyrosztoczy/uuid_cast_error

The relevant schema is here: https://github.com/joeyrosztoczy/uuid_cast_error/blob/master/lib/minimal_uuid_cast_error.ex#L10-L24

I believe it is because Ecto.UUID is returning a type of :uuid, which gets interpreted as a custom module type since :uuid is no longer a primitive type.

### Expected behavior
Inner type successfully cast.

PR #3047, opened 2019-06-29
Fix Ecto.UUID type() return

joeyrosztoczy:
Co-authored-by: Kevin Baird <kevincbaird@gmail.com>

I opened the following issue in case our fix was in the wrong direction, I do believe this is a bug though: https://github.com/elixir-ecto/ecto/issues/3046

This PR fixes the linked issue.

Main changes:

[uuid.ex ln 21.](https://github.com/elixir-ecto/ecto/compare/master...joeyrosztoczy:update-uuid-type-return?expand=1#diff-8e368fa7e9c6f93fcd82a15e705fc08cL21)

I believe that Ecto.UUID.type() should return the desired schema type (either custom or primitive). It was returning a non-custom, non-primitive value which could throw runtime errors for composite type casting.

I updated the type() doc tests to reflect that :uuid is no longer returned.

And I changed [this doc test to match](https://github.com/elixir-ecto/ecto/compare/master...joeyrosztoczy:update-uuid-type-return?expand=1#diff-68b1275ffa052929bf96ca32ea92092dR267), which should have previously failed since :uuid wasn't a primitive type.

I apologize for the noise - I checked a couple times and believe I just ran the formatter as configured, but if you'd like me to strip down to just the relevant lines or it seems like the formatting is in error please let me know.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -80,19 +80,33 @@ defmodule Ecto.Type do
   import Kernel, except: [match?: 2]
 
   @typedoc "An Ecto type, primitive or custom."
-  @type t         :: primitive | custom
+  @type t :: primitive | custom
 
   @typedoc "Primitive Ecto types (handled by Ecto)."
   @type primitive :: base | composite
 
   @typedoc "Custom types are represented by user-defined modules."
-  @type custom    :: module
+  @type custom :: module
 
-  @typep base      :: :integer | :float | :boolean | :string | :map |
-                      :binary | :decimal | :id | :binary_id |
-                      :utc_datetime | :naive_datetime | :date | :time | :any |
-                      :utc_datetime_usec | :naive_datetime_usec | :time_usec
-  @typep composite :: {:array, t} | {:map, t} | {:embed, Ecto.Embedded.t} | {:in, t}
+  @typep base ::
+           :integer
+           | :float
+           | :boolean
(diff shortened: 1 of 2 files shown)


josevalim:
Can you please submit the changes without the formatter modifications?
Otherwise it is complicated to review it. We also need a test to avoid
regressions in the future.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
Btw, the test should be an "integration test"  that reproduces the failure in #3046. My suspicion is that the bug is not in `Ecto.UUID` but rather in the typed map logic which is converting `Ecto.UUID` to `:uuid` when it should not. You can look at `test/ecto/type_test.exs` for example. We need an assertion like:

    assert cast({:map, Ecto.UUID}, %{"x" => "foo"}) == {:ok, %{"x" => ...}}

That fails with the error you mentioned in the parent issue. Thanks!

josevalim:
Ok, i found the root cause of the bug! :D Thanks for the minimal app reproducing it, it makes our lives much easier! :D

joeyrosztoczy:
Ah dangit. Haha. You bet for the minimal case. Sorry for the formatter / wrong test will make sure not to use if ever find a chance to try and contribute! Thanks for taking care of the real issue in ecto_sql <3

PR #3051, opened 2019-07-05
do not override explictly set autoupdate fields

ericsullivan:
(For a schema using timestamps) if the updated_at field isn't set, Ecto will
autoupdate it. If it's part of the changeset, Ecto will use that value.
The issue is that when putting changes, Ecto will not add identical values.
If during an update, the updated_at field is set to the models (original)
updated_at value, it will not be part of the changeset. If another value is
changed (so the schema has changes), the updated_at field will (potentially)
have a different value (depending on the timing of the update call). As the
field was being explicitly set to a specific value, this was surprising.

This fix forces the updated_at field to be added to the changeset, even when
it's identical to the original value. I assumed this logic would apply to
all autoupdated fields, and not just timestamps.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1143,7 +1143,7 @@ defmodule Ecto.Changeset do
 
   defp put_change(data, changes, errors, valid?, key, value, type) do
     cond do
-      not Ecto.Type.equal?(type, Map.get(data, key), value) ->
+      not Ecto.Type.equal?(type, Map.get(data, key), value) || autoupdate_field?(data, key) ->
         {Map.put(changes, key, value), errors, valid?}
 
       Map.has_key?(changes, key) ->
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks for the PR!

Changesets work on a simple contract, __changeset__ and we wouldn’t want to
augment it and couple it with schemas any more than necessary.

Also, the behavior can be achieved with force_change, so please use said
API instead.


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

ericsullivan:
Thanks for the quick review. I noticed this on a schema that is sometimes populated from external data, and was surprised by this behaviour of `cast` (I'd explicitly set updated_at in my tests, and the persisted value would occasionally be a second later). But the argument against any further coupling make sense. I thought this approach might have been rejected due to the possibility of unnecessary DB write, so I'm glad I proposed this PR before I tried a more complex solution. Thanks again

Issue #3052, opened 2019-07-05
CTE in update_all


mbuhot:
### Precheck

* Do not use the issues tracker for help or support requests (try Elixir Forum, Stack Overflow, IRC or mailing lists, etc).
* For proposing a new feature, please start a discussion on [elixir-ecto](https://groups.google.com/forum/#!forum/elixir-ecto).
* For bugs, do a quick search and make sure the bug has not yet been reported.
* Finally, be nice and have fun!

Apologies if this is more feature request than bug report. I tried to access the google groups link above, but got this error: 

```
This group either doesn't exist, or you don't have permission to access it. If you're sure this group exists, contact the Owner of the group and ask them to give you access.
```

### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 21 [erts-10.2.3] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.8.1 (compiled with Erlang/OTP 21)
```

* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): postgres:9.6
* Ecto version (mix deps): ecto 3.2.0-dev at 8fa9df5 ecto_sql 3.2.0-dev  at 542fe4d
* Database adapter and version (mix deps): postgrex 0.14.3
* Operating system: MacOS 10.14.5

### Current behavior

I'd like to express the following update statement using Ecto's recently added CTE support:

```sql
WITH available_jobs AS (
    SELECT id 
    FROM jobs
    WHERE jobs.state = 'AVAILABLE'
    ORDER BY jobs.id ASC
    LIMIT 10
    FOR UPDATE SKIP LOCKED
)
UPDATE jobs
SET state = 'RESERVED'
FROM available_jobs
WHERE jobs.id = available_jobs.id
RETURNING jobs.*
```

The equivalent ecto code:

```elixir
available_jobs = 
  from(
    job in Jobs,
    where: job.state == "AVAILABLE",
    order_by: [asc: job.id],
    limit: 10,
    lock: "FOR UPDATE SKIP LOCKED",
    select: %{id: job.id}
  )

Job
|> with_cte("available_jobs", as: ^available_jobs)
|> join(:inner, [job], a in "available_jobs", on: job.id == a.id) 
|> select([job, _available], job)
|> Repo.update_all(set: [state: "RESERVED"])
```

Fails with an error: 

```
[...]

PR #3053, opened 2019-07-06
Allow CTEs in update_all queries

mbuhot:
This PR is to enable #3052 such that CTEs may be used in `update_all` queries.

The initial implementation is a minimal hack, hopefully someone can assist in getting to a clean solution.

The changes are:

1) Relax the validation in `Ecto.Query.Planner.assert_only_filter_expressions!` to allow non-nil `with_ctes` in a query.

2) The `Ecto.Query.Planner.traverse_exprs` function processes all query expressions as well as the existing update expressions. Without this change, the `params` list returned from `Ecto.Query.Planner.plan_cache` would not include any parameters intended to be used in the CTE.

Based on my limited understand of the query planner module, maybe something needs to be modified in [merge_cache/6](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L607-L625) ?

Corresponding changes to ecto_sql in https://github.com/elixir-ecto/ecto_sql/pull/125

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -604,16 +604,16 @@ defmodule Ecto.Query.Planner do
     cache_and_params
   end
 
-  defp merge_cache(:with_cte, query, with_expr, cache_and_params, operation, adapter) do
+  defp merge_cache(:with_cte, query, with_expr, cache_and_params, _operation, adapter) do
     %{queries: queries, recursive: recursive} = with_expr
     key = if recursive, do: :recursive_cte, else: :non_recursive_cte
-    fun = &{&3, merge_cache(&1, &2, &3, &4, operation, adapter)}
+    fun = &{&3, merge_cache(&1, &2, &3, &4, :all, adapter)}
 
     # In here we add each cte as its own entry in the cache key.
     # We could group them to avoid multiple keys, but since they are uncommon, we keep it simple.
     Enum.reduce queries, cache_and_params, fn
       {name, %Ecto.Query{} = query}, {cache, params} ->
-        case traverse_exprs(query, operation, {[], params}, fun) do
+        case traverse_exprs(query, :all, {[], params}, fun) do
           {_, {:nocache, _} = acc} -> acc
           {_, {inner_cache, params}} -> {[{key, name, inner_cache} | cache], params}
         end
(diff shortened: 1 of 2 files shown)


lib/ecto/query/planner.ex
```diff
@@ -1420,7 +1420,7 @@ defmodule Ecto.Query.Planner do
     exprs =
       case operation do
         :all -> @all_exprs
-        :update_all -> @update_all_exprs
```
josevalim: There is no reason to change this line AFAIK? Update all already includes with_cte.

josevalim: Or do you mean that without this change ctes are not processed at all? If so, I think what we should do is, when we process ctes, the operation should always be :all. The same for subwueries.

josevalim: To be clear, pass :all instead of operation here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L761

Please do the same for when we plan subqueries and combinators. :)

josevalim: Actually, I figured out other improvements in this area, so I push a commit to master. You should rebase and remove those changes and everything should still work. :heart:

mbuhot: Thanks @josevalim, I've rebased and added a failing test to demonstrate the problem I'm getting.

mbuhot: After taking another look, I think I found the problem. `merge_cache` and `validate_and_increment` both had to be updated to use `:all` when calling `traverse_exprs` on a CTE query expression.

test/ecto/query/planner_test.exs
```diff
@@ -465,6 +465,27 @@ defmodule Ecto.Query.PlannerTest do
     assert [:all, {"comments", Comment, _, nil}, {:recursive_cte, "cte", ^expr}] = cache
   end
 
+  test "prepare: Prepare Update CTE" do
```
mbuhot: This test was to help reproduce the issue I was having with params missing from CTE expressions.
Not sure if it is worth keeping?

josevalim: Yes, it is worth keeping a test with CTE but we should rewrite it to do similar assertions to the test above. also please fix the casing in the test title. :)

josevalim:
One comment and we should be good to go!

test/ecto/query/planner_test.exs
mbuhot: 👍 I've added assertions for the structure of the query and cache key.

josevalim:

PR #3057, opened 2019-07-10
Consider reordered embedded children as change

jayjun:
**Problem**
Reordering embedded children (with no other changes) does not get persisted. For example,

```elixir
order = Repo.get!(Order, 42)
order.items
# => [%Item{id: 1, title: "Soap"}, %Item{id: 2, title: "Shampoo"}]

new_order =
  order
  |> Ecto.Changeset.change()
  |> Ecto.Changeset.put_embed(:items, Enum.reverse(order.items))
  |> Repo.update!()

new_order.items
# => [%Item{id: 1, title: "Soap"}, %Item{id: 2, title: "Shampoo"}]
```

This behaviour is understandable for relational associations, but can be surprising when you know your children are stored in a JSON column.

The simplest workaround is to add a `:position` field to every child, which feels rather extra when JSON arrays are implicitly ordered.

**Solution**
This pull request prevents strictly ordinal changes of embeds from getting flagged as `:ignore`. 

It has the added benefit of recognising the implicit order of input to `cast_embed`. So one can sort children merely by rearranging the order of DOM elements, which should make a killer Phoenix LiveView demo.

For example, casting
```elixir
order_params = %{
  "items" => [
    %{"id" => 2},
    %{"id" => 1}
  ]
}
```
will reorder existing `items` accordingly.

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -264,11 +264,26 @@ defmodule Ecto.Changeset.Relation do
     {:ok, [], true}
   end
 
-  defp cast_or_change(%{cardinality: :many, unique: unique}, value, current, current_pks, new_pks, fun) when is_list(value) do
-    map_changes(value, new_pks, fun, process_current(current, current_pks), [], true, true, unique && %{})
+  defp cast_or_change(%relation{cardinality: :many, unique: unique}, value, current, current_pks, new_pks, fun) when is_list(value) do
+    skip? = not embedded_and_reordered?(relation, valid_new_pks(value, new_pks, []), Enum.map(current, current_pks))
+    map_changes(value, new_pks, fun, process_current(current, current_pks), [], true, skip?, unique && %{})
   end
 
   defp cast_or_change(_, _, _, _, _, _), do: :error
+
+  defp embedded_and_reordered?(Ecto.Embedded, new_pks, current_pks) when new_pks != [],
+    do: new_pks != current_pks
+  defp embedded_and_reordered?(_relation, _new_pks, _current_pks),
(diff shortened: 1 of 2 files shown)


josevalim:
Quick question, do you know upfront they have been reordered or is this mostly happening due to external factors?

jayjun:
My intention is without knowing upfront. My endgame is, imagine a form with `inputs_for` tags like

```html
<input id="order_items_0_id" name="order[items][0][id]" type="hidden" value="1">
<input id="order_items_0_title" name="order[items][0][title]" type="text" value="Soap">

<input id="order_items_1_id" name="order[items][1][id]" type="hidden" value="2">
<input id="order_items_1_title" name="order[items][1][title]" type="text" value="Shampoo">
```

I want to reorder them in the database merely by changing that to

```html
<input id="order_items_0_id" name="order[items][0][id]" type="hidden" value="2">
<input id="order_items_0_title" name="order[items][0][title]" type="text" value="Shampoo">

<input id="order_items_1_id" name="order[items][1][id]" type="hidden" value="1">
<input id="order_items_1_title" name="order[items][1][title]" type="text" value="Soap">
```

Similarly, the RESTful equivalent
```json
PUT /orders/:id
{
  "items": [
    {"id": 2, "title": "Shampoo"},
    {"id": 1, "title": "Soap"}
  ]
}
```
without adding a `:position` field. Currently, Ecto does not consider the changes above as a change.

Granted this only works for embeds, but as mentioned, I love the simplicity that comes from leveraging implicit orders.

josevalim:
I agree, asking for curiosity. Thanks @jayjun!

PR #3070, opened 2019-07-24
Fix migration example in constraints guide

gjastrab:
There is no `Ecto.Migration.add/1`


josevalim:

PR #3075, opened 2019-07-26
Fix missing prefix for CTE queries

304:
It is a follow-up for https://github.com/elixir-ecto/ecto/pull/3039#issuecomment-515116883.

CTE queries are not respecting a query prefix. The solution is to fix them in a similar manner as it was done for union queries.


josevalim:
Excellent, thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #3088, opened 2019-08-15
Allow for `Repo.transaction` to receive a arity one function, which is supplied the repo

LostKobrakai:
I feel this would align the behaviour of `Repo.transaction` functions to the ones of `Ecto.Multi.run` functions. I was just talking to @benwilson512 about an API in his fable library, which would return a function or ecto.multi based on the inputs and would always be piped into `Repo.transaction`. Currently only the `Ecto.Multi` based API wouldn't need to know the actual repo used, while the one returning a functions can't work the same way.

Edit: I'm aware that test are missing, maybe a issue would've been good before the pr :)

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1386,6 +1386,7 @@ defmodule Ecto.Repo do
         MyRepo.update!(change(bob, balance: bob.balance + 10))
       end)
       
+      # When passing a function of arity 1, it receives the repository itself
       MyRepo.transaction(fn repo -> 
         repo.insert!(%Post{})
       end)
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1340,6 +1340,11 @@ defmodule Ecto.Repo do
 
   ## Use with function
 
+  `c:transaction/2` can be called with both a function of arity
+  zero or one. The arity zero function will just be executed as is,
+  while the arity one function will receive the repo of the transaction
+  as it's first argument, similar to `Ecto.Multi.run`.
+
   If an unhandled error occurs the transaction will be rolled back
   and the error will bubble up from the transaction function.
   If no error occurred the transaction will be committed when the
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks. Yes, let's go ahead with it!

LostKobrakai:
Seems like there's actually no test covering the existing functionality yet. Am I missing something here?

josevalim:
@LostKobrakai they  are under integration_test, the README has instructions on how to get started on those, but I think this feature will be fine with unit tests. We definitely have some unit tests using transactions but we likely don't have unit tests for Repo.transactioni directly. A great time to get sstarted!

lib/ecto/repo.ex
```diff
@@ -1376,9 +1381,9 @@ defmodule Ecto.Repo do
 
       import Ecto.Changeset, only: [change: 2]
 
-      MyRepo.transaction(fn ->
-        MyRepo.update!(change(alice, balance: alice.balance - 10))
-        MyRepo.update!(change(bob, balance: bob.balance + 10))
+      MyRepo.transaction(fn repo ->
```
LostKobrakai: I'm not sure about this: Edit the existing example or add a new one or just skip an example for the new option?

lib/ecto/repo.ex
```diff
@@ -1340,6 +1340,11 @@ defmodule Ecto.Repo do
 
   ## Use with function
 
+  `c:transaction/2` can be called with both a function of arity
+  zero or one. The arity zero function will just be executed as is,
+  while the arity one function will receive the repo of the transaction
+  as it's first argument, similar to `Ecto.Multi.run`.
```
```suggestion
  as its first argument, similar to `Ecto.Multi.run`.
```

lib/ecto/repo.ex
josevalim: Let's show two separate examples!

test/ecto/repo_test.exs
```diff
@@ -1169,4 +1169,19 @@ defmodule Ecto.RepoTest do
       refute function_exported?(ReadOnlyRepo, :delete_all, 2)
     end
   end
+
+  describe "transaction" do
+    test "a arity zero function will be executed any it's value returned" do
```
josevalim:
```suggestion
    test "an arity zero function will be executed any it's value returned" do
```

lib/ecto/repo.ex
```diff
@@ -1380,6 +1385,10 @@ defmodule Ecto.Repo do
         MyRepo.update!(change(alice, balance: alice.balance - 10))
         MyRepo.update!(change(bob, balance: bob.balance + 10))
       end)
+      
+      MyRepo.transaction(fn repo ->
```
```suggestion
      # When passing a function of arity 1, it receives the repository itself
      MyRepo.transaction(fn repo -> 
```

josevalim:

PR #3092, opened 2019-08-27
Dynamic SelectMerge

isaacsanders:
This is something that I am trying to do in an application. It didn't work in my application, but this is a minimal test that contains the feature I am trying to add.


isaacsanders:
I think there is work that needs to be done in `ecto_sql` to support this, too.

PR #3096, opened 2019-08-31
Fix `escape/5` @spec

joaothallis:
The `params_acc` param is a a tuple with a list and term, the result also is a tuple with a list and term.


josevalim:

PR #3100, opened 2019-09-06
Tweak join docs to use on: versus where:

codinghombre:
Change Ecto.Query join doc examples to use "on:" instead of ":where"


josevalim:

PR #3101, opened 2019-09-09
fix link to callback

dnsbty:
There is a broken link from `Type.embed_as/2` to the `embed_as/1` callback. See the broken link here: https://hexdocs.pm/ecto/Ecto.Type.html#embed_as/2


wojtekmach:
Thank you!

Issue #3103, opened 2019-09-10
replace default incremeter in optimistic_lock to avoid exception raised when it exceeds integer limit


bruteforcecat:
https://github.com/elixir-ecto/ecto/blob/d8ae5564a17f5b0a40962d81ef71360560210f12/lib/ecto/changeset.ex#L2235

If the lock_version is specified as :integer in the migration, it will throw exception when the lock_version exceed its db limit(2147483647 for both Postgres and MySQL). So should we change the default incrementer to cycle it back to default(1) when it reaches `2147483647`?

I can create a PR for this if most of you think it's a good idea.

josevalim:
I would love to see a PR for this, yes, thank you.

josevalim:
Closing in favor of PR.

PR #3105, opened 2019-09-10
add rollover logic for default incremeter in optimistic_lock

bruteforcecat:
As discussed in the previous [issue](https://github.com/elixir-ecto/ecto/issues/3103), this PR is for avoiding an exception when the lock version exceed the PostgresSQL and MySQL signed integer size limit.


josevalim:
Beautiful. Can you please add a test?

bruteforcecat:
I added the test. Do you think it's sufficient?

I was thinking of adding like an integration test for actual DB operation to assert not exception raise but then realize there should be no DB adapter library required in Ecto repo itself. So I didn't add the integration test.

josevalim:

PR #3108, opened 2019-09-11
Fix docs. Replace explicit calls in `Ecto.Multi` to `repo` from function header.

Hajto:
Currently:
```
|> Ecto.Multi.run(:post, fn _repo, _changes ->
     {:ok, MyApp.Repo.get(Post, 1) || %Post{}}
   end)
```

I think the `repo` argument intention was to enable injecting repo module into callback function. I think it is nice feature that complies with `explicit contract` methodology and should be enforced in docs.  So I changed references of `MyApp.Repo` to `repo`.

```
|> Ecto.Multi.run(:post, fn repo, _changes ->
     {:ok, repo.get(Post, 1) || %Post{}}
   end)
```

If this is idea is welcome I can also format this file, becase it does not compy with mix format.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -324,7 +324,7 @@ defmodule Ecto.Multi do
 
       Ecto.Multi.new()
       |> Ecto.Multi.run(:post, fn _repo, _changes ->
-           {:ok, MyApp.Repo.get(Post, 1) || %Post{}}
+           {:ok, repo.get(Post, 1) || %Post{}}
          end)
       |> Ecto.Multi.insert_or_update(:update, fn %{post: post} ->
            Ecto.Changeset.change(post, title: "New title")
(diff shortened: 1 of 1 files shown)


josevalim:

lib/ecto/multi.ex
```diff
@@ -324,7 +324,7 @@ defmodule Ecto.Multi do
 
       Ecto.Multi.new()
       |> Ecto.Multi.run(:post, fn _repo, _changes ->
```
jakub-gonet: `_repo` -> `repo`, as we now use this variable

lib/ecto/multi.ex
```diff
@@ -361,7 +361,7 @@ defmodule Ecto.Multi do
...
```
jakub-gonet: `_repo` -> `repo`, as we now use this variable

PR #3114, opened 2019-09-17
Document that the function passed to prepare_changes/2 only runs when changeset is valid

thojanssens:


josevalim:

PR #3124, opened 2019-09-26
Allow :with to be given as a {module, function} tuple

ccapndave:
I've been building an app where various forms contain multiple liveviews for particularly interactive sections. I've been passing the form into the session, and using normal form functions inside the liveview which has worked fine.

This afternoon I added `cast_assoc` to a changeset and suddenly my liveviews exploded with `** (ArgumentError) cannot deserialize #Function<15.129553521/2 in Ecto.Changeset.on_cast_default/2>, the term is not safe for deserialization` errors, because as it turns out the `:with` option of `cast_assoc` implicitly creates an anonymous function.

Therefore this pull request circumvents the problem by allowing the `:with` option to be given as a tuple of `{module_name, changeset_function_name}`, thereby making the changeset serializable if necessary.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -822,8 +822,8 @@ defmodule Ecto.Changeset do
               which is used by cast_#{type}/3. You need to either:
 
                 1. implement the #{type}.changeset/2 function
-                2. pass the :with option to cast_#{type}/3 with an anonymous function that expects 2 args
-                   or an MFA.
+                2. pass the :with option to cast_#{type}/3 with an anonymous
+                   function that expects 2 args or an MFA tuple
 
               When using an inline embed, the :with option must be given
               """
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -709,8 +709,10 @@ defmodule Ecto.Changeset do
 
     * `:with` - the function to build the changeset from params.
       Defaults to the changeset/2 function in the association module.
-      This can either be an anonymous function, or a `{module, function}`
-      tuple.
+      This can either be an anonymous function, or an MFA.  If using an MFA, the
+      passed arguments are added to the default changeset and parameters arguments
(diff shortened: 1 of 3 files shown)


josevalim:
Hi Dave, can you please use {module, function, args} as the tuple where the
changeset is prepended? That’s the most common format. Also, can you please
add a test? Thank you.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

ccapndave:
Hi Jose,

Just to check, do you mean to specifically pass in the changeset and params at the point that `cast_assoc` is used, like this?

```
def changeset(todo_list, params \\ %{}) do
  before_cast_changeset =
    todo_list
    |> cast(params, [:body])

  before_cast_changeset
  |> cast_assoc(:todo_items, with: {TodoItem, :changeset, [before_cast_changeset, params]})
end
```

wojtekmach:
@ccapndave the usage should be like this:

```elixir
before_cast_changeset
|> cast_assoc(:todo_items, with: {TodoItem, :changeset, [params]})
```

but under the hood it would be called like that: `TodoItem.changeset(before_cast_changeset, params)`

ccapndave:
I'm just looking at this now, and Ecto does some cleverness with extracting the relevant bit of the attributes to pass into the associated changeset function.  I guess we don't want the user to have to duplicate this logic, so shall I instead prepend both the `changeset` and the `attrs` to the passed parameters?

That means that in general the third argument would be an empty list, unless the changeset function actually takes extra arguments.

josevalim:
Yes, sounds good. We would do something like `{module, fun, args}` and call it as `apply(module, fun, [changeset, attrs | args])`.

ccapndave:
I never knew you could prepend multiple items to a list with that syntax.  How about that :)

ccapndave:
Requested changes implemented.

josevalim:

PR #3128, opened 2019-09-30
[WIP] Add a failing subquery test

futpib:


josevalim:
See #3129 for why the behaviour is correct. :)

Issue #3134, opened 2019-10-02
Misleading error in Ecto.Changeset.put_change/7


syfgkjasdkn:
👋

* Elixir version (elixir -v):

```
Erlang/OTP 22 [erts-10.5.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.9.1 (compiled with Erlang/OTP 22)
```

* Ecto version (mix deps): 3.2.2
* Operating system: macos 10.14.6 (18G95)

Example: https://github.com/syfgkjasdkn/ecto-error

### Current behavior

```elixir
iex(2)> Demo.error
** (ArgumentError) unknown field `something` in %Demo.Schema{__meta__: #Ecto.Schema.Metadata<:built, "asd">, id: nil, something: nil}
    (ecto) lib/ecto/changeset.ex:1141: Ecto.Changeset.put_change/7
    (stdlib) maps.erl:232: :maps.fold_1/3
    (ecto) lib/ecto/changeset.ex:384: Ecto.Changeset.change/2
```

### Expected behavior

No errors. Or an error saying that some piece of metadata is missing like https://github.com/elixir-ecto/ecto/blob/v3.2/lib/ecto/changeset.ex#L1169.

Forum discussion: https://elixirforum.com/t/misleading-ecto-error-unknown-field-password-in-app-user-password-nil/25814

PR #3135, opened 2019-10-03
Improve error message on non-atom keys to change/put_change

wojtekmach:
Ref: #3134

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1192,8 +1192,12 @@ defmodule Ecto.Changeset do
     end
   end
 
-  defp put_change(data, _changes, _errors, _valid?, key, _value, nil) do
+  defp put_change(data, _changes, _errors, _valid?, key, _value, nil) when is_atom(key) do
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1192,10 +1192,14 @@ defmodule Ecto.Changeset do
     end
   end
 
-  defp put_change(data, _changes, _errors, _valid?, key, _value, nil) do
+  defp put_change(data, _changes, _errors, _valid?, key, _value, nil) when is_atom(key) do
     raise ArgumentError, "unknown field `#{inspect(key)}` in #{inspect(data)}"
   end
 
+  defp put_change(_data, _changes, _errors, _valid?, key, _value, nil) when not is_atom(key) do
+    raise ArgumentError, "field names given to change/put_change must be atoms, got: `#{inspect(key)}`"
```
wojtekmach: we discussed augmenting the existing error message, however if we put it at the end of the error it can easily get lost:

image

wojtekmach:
Backported to v3.2.

PR #3141, opened 2019-10-16
Won't autogenerated value like timestamps

wandarkaf:
If the migration of tags is set with timestamps(), it will throw:
*ERROR 23502 (not_null_violation) null value in column "inserted_at" violates not-null constraint*
Because insert_all/3  won't autogenerated value like timestamps by definition.

In order to this guide work, we need to remove timestamps(). And let


josevalim:
Excellent find, @wandarkaf! What if instead of removing the timestamps, we also set the timestamps when generating the insert_all? I think this would be more helpful because people often run into this issue, so having an example would be great. It would be a matter of doing something like:

    timestamp = NaiveDateTime.utc_now
    maps = Enum.map(names, &%{name: &1, inserted_at: timestamp, updated_at: timestamp})

WDYT? Can you please update accordingly?

wandarkaf:
Done! I added a quick explanation on the next paragraph, so is more clear the purpose of adding timestamps. 

What do you think?

PR #3142, opened 2019-10-17
Fixing issues in code examples

wandarkaf:
On `insert_and_get_all_tags()`, inside the case, for consistency, it should expect `names`. If you put `tags`, it will give an error (is not expecting it). Or you can change all to tags, could make more sense semantically speaking.

For some reason, which I don't know why yet, `timestamps()` expects that you truncate the second off the `NativeDateTime` given format. If not, it will give you an error.


josevalim:
Thanks for helping keeping the guides consistent. 

josevalim:

PR #3143, opened 2019-10-17
Issue with timestamp native format

wandarkaf:


wandarkaf:
Maybe is an overkill this patch and it should be together with patch-4.

Is about the timestamp I commented previously on patch-4. Sorry if it is an overkill and I'm mistaken. I'm just trying to learn with scrutiny with the help of these guides.

josevalim:

Issue #3139, opened 2019-10-15
Poor error message on invalid binding


josevalim:
```
"organizations"
|> join(:inner, [o, p] in subquery(some_subquery), on: o.id == p.org_id)
```

Raises:

```
** (FunctionClauseError) no function clause matching in Ecto.Query."MACRO-join"/6    
    
    The following arguments were given to Ecto.Query."MACRO-join"/6:
```

michalmuskala:
Closed by #3145

PR #3145, opened 2019-10-18
fix: better error message on invalid join binding

hauptbenutzer:
Hi! 

So #3139 was a bit vague in terms of what was the expected behaviour 😅
But I thought, I'd give it a shot.

Cheers!


josevalim:

PR #3150, opened 2019-10-26
Fix: prefix assignment for through: [x, y] associations

hkrutzer:


hkrutzer:
Associations with `through: [x, y]` don't currently get a prefix assigned.

However I'm not sure `source` should be `nil` here at all. @josevalim can you shed some light on this?

If this is the correct fix I'll add some tests to cover other prefix types.

josevalim:
Hi @hkrutzer!

I went to validate if your patch is correct and it turned out that:

1. I didn't know the answer
2. Finding out the answer required solving the problem

So I went ahead and pushed a fix. The test case was extremely helpful though, thank you!

hkrutzer:
@josevalim thank you for your very fast response :)

PR #3151, opened 2019-10-27
fix example code in guides for stopping dynamic repo

irisTa56:
The original example code results in `** (ErlangError) Erlang error: :timeout_value` because `Ecto.Repo.stop/1` accepts `timeout` not `pid`.
I think calling `Supervisor.stop/3` is one of the ways to stop the dynamic repo here.


josevalim:
Good catch! I have pushed a slightly different solution, then I realized that your solution is indeed better, so I have reverted my changes. :D Thank you!

josevalim:

PR #3156, opened 2019-11-06
Remove @actions list in Ecto.Changeset.apply_action/2. Update spec, t…

ckoch-cars:
…ests, and type.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -293,7 +293,7 @@ defmodule Ecto.Changeset do
 
   @type t :: t(Ecto.Schema.t | map | nil)
   @type error :: {String.t, Keyword.t}
-  @type action :: nil | atom
+   @type action :: nil | :insert | :update | :delete | :replace | :ignore | atom
   @type constraint :: %{type: :check | :exclusion | :foreign_key | :unique,
                         constraint: String.t, match: :exact | :suffix | :prefix,
                         field: atom, error_message: String.t, error_type: atom}
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -293,7 +293,7 @@ defmodule Ecto.Changeset do
 
   @type t :: t(Ecto.Schema.t | map | nil)
   @type error :: {String.t, Keyword.t}
-   @type action :: nil | :insert | :update | :delete | :replace | :ignore | atom
+  @type action :: nil | :insert | :update | :delete | :replace | :ignore | atom
   @type constraint :: %{type: :check | :exclusion | :foreign_key | :unique,
                         constraint: String.t, match: :exact | :suffix | :prefix,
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -293,7 +293,7 @@ defmodule Ecto.Changeset do
 
   @type t :: t(Ecto.Schema.t | map | nil)
   @type error :: {String.t, Keyword.t}
-  @type action :: nil | :insert | :update | :delete | :replace | :ignore
+  @type action :: nil | atom
```
ckoch-cars: I wasn't clear about the role that `:ignore` plays in this type, especially given that it was not in `@actions` list.

josevalim: Those are the types used by Ecto, so I would keep them in the list for explicitness and add the catch all, like this:

```suggestion
   @type action :: nil | :insert | :update | :delete | :replace | :ignore | atom
```

josevalim:
Just one comment and we can ship it!

lib/ecto/changeset.ex
```diff
@@ -1498,17 +1497,16 @@ defmodule Ecto.Changeset do
       iex> {:error, changeset} = apply_action(changeset, :update)
       %Ecto.Changeset{action: :update}
   """
-  @spec apply_action(t, :insert | :update | :delete | :replace) ::
-          {:ok, Ecto.Schema.t() | data} | {:error, t}
-  def apply_action(%Changeset{} = changeset, action) when action in @actions do
+  @spec apply_action(t, atom) :: {:ok, Ecto.Schema.t() | data} | {:error, t}
```
wojtekmach:
```suggestion
  @spec apply_action(t, action) :: {:ok, Ecto.Schema.t() | data} | {:error, t}
```

ckoch-cars: Thanks for the suggestion. This seems like it would be functionally equivalent (although it would allow for the possibility to rollaback the change to allow any atom in actions.) Either way, I'm happy to add your suggested change.

lib/ecto/changeset.ex
ckoch-cars: @josevalim I must admit, I'm still unclear about the role the `:ignore` plays in a changeset, esp given the exception that gets raised in Schema. Under what circumstances would :ignore be used?

josevalim:
`:ignore` is used to discard changesets generated in cast_assoc, put_assoc and friends.

lib/ecto/changeset.ex
josevalim: I think I will go with atom just because `action` has both `nil` and `:ignore` that we wouldn't place here. It is kind of a weak argument, because those can be passed as `atom` anyway, but maybe as atom it calls less attention to them.

josevalim:

PR #3157, opened 2019-11-07
Adds `storage_status/1` callback to `Ecto.Adapters.Storage` behaviour

mracos:
Returns the status of a given `storage` for the given `options`.
Can be `:up`, `:down` or `{:error, term()}` in case anything goes wrong.

### Motivation
At least for now, it's planned to be used on a (future) Plug on `phoenix_ecto` that shows an error page when the database for a repo is not created.

Related to: elixir-ecto/ecto_sql#164

cc @josevalim


wojtekmach:
Thank you!

PR #3160, opened 2019-11-11
Add test for union returning wrong result

samphilipd:


integration_test/cases/unions.exs
```diff
@@ -0,0 +1,25 @@
...
+  test "ecto bug test" do
+    TestRepo.insert!(%Post{})
+    TestRepo.insert!(%Post{})
+
+    # Insert one role
+    query1 = from(p in Post, select: %{n: 1})
+    query2 = from(p in Post, select: %{n: 2})
+
+    # This assertion fails
+    assert TestRepo.all(union(query1, ^query2)) == [%{n: 1}, %{n: 2}]
```
Eiji7: Please correct me if I'm wrong, but even if there is any bug in `select` then after solving it this test would fail anyway as we are inserting 2 `Post` while having a union of 2 queries (each would give those 2 posts), so expected result should be:

```elixir
[%{n: 1}, %{n: 1}, %{n: 2}, %{n: 2}]
```

PR #3161, opened 2019-11-13
Document one-arity function for Repo.transaction/2

sorentwo:
The ability to pass a zero or one arity function to Repo.transaction was added in https://github.com/elixir-ecto/ecto/pull/3088, but it wasn't documented in the CHANGELOG.

I learned about the change from this error report on Oban: https://github.com/sorentwo/oban/issues/96


josevalim:

PR #3176, opened 2019-12-02
Allow aggregating over `*`

### LINKED ISSUE #3175: Allow passing `:*` as a column name in the `Repo.aggregate/{3,4}`
Reasoning the same as in #2595.

hauleth:
This is done to provide similar API to `Ecto.Query.API.count/0` as
counting over `*` is faster in some DB than counting over column, even
when column is `NOT NULL` and indexed.

Close #3175


josevalim:

PR #3182, opened 2019-12-11
Fix naive datetime casting

aptinio:
for maps with either all date fields or all time fields blank


josevalim:

PR #3195, opened 2019-12-25
Support `filter/2` in WindowAPI

fuelen:


josevalim:
Excellent. Can you please add a basic test here? https://github.com/elixir-ecto/ecto/blob/master/test/ecto/query/builder/windows_test.exs - please make sure the tree at is failing before the fix and happy holidays!

fuelen:
@josevalim I don't know what test should be in `test/ecto/query/builder/windows_test.exs` file because I can't find similar tests there, so I added it to `test/ecto/query/builder_test.exs`

josevalim:

PR #3197, opened 2020-01-02
Run the formatter on all the files

Adzz:
Is this an okay thing to do?

I was writing a separate PR locally and I have 'format on save' going in my editor which meant every time I saved there were lots of unrelated file changes.

Shall I turn it off when contributing to Ecto, or is this okay to do?


josevalim:
Hi @Adzz! We haven't run the formatter on purpose. We would need to check file by file, as someone a slight refactor may be necessary, and it would mess up the git history. So we prefer to do it gradually as we modify things. Thank you.

PR #3201, opened 2020-01-09
Respect child schema prefix in cast_assoc

### LINKED ISSUE #3200: Additional schema prefix not being used in cast_assoc
### Environment

* Elixir version (elixir -v): 1.8
* Database and version PostgreSQL 9.6
* Ecto version (mix deps): 3.3
* Database adapter and version (mix deps): Ecto.Adapters.Postgres
* Operating system: MacOs 10.14

### Current behavior

I have run into an issue when using cast_assoc when the child association is a different schema prefix then the parent schema and I am receiving this error

```

defmodule ParentSchema do
  use Ecto.Schema
  import Ecto.Changeset

  @schema_prefix "parent"
  schema "table_1" do
    belongs_to(:child, ChildSchema)
  end

  def changeset(data, params) do
    data
    |> cast(params, __schema__(:fields))
    |> cast_assoc(:child)
  end
end

defmodule ChildSchema do
  use Ecto.Schema
  import Ecto.Changeset

  @schema_prefix "child"
  schema "table_2" do
    field(:text)
  end

  def changeset(data, params) do
    cast(data, params, __schema__(:fields))
  end
end

defmodule Test do
  def update do
    params = %{
      id: 1,
      child: %{
        id: 1,
        text: "Updated Text"
      }
    }

    ParentSchema
    |> MyRepo.get(1)
    |> My.Repo.preload([:child])
    |> ParentSchema.changeset(params)
    |> My.Repo.update()
  end
end
```

```
iex(28)> Test.update

** (Postgrex.Error) ERROR 42P01 (undefined_table) relation "parent.table_2" do
es not exist

    query: UPDATE "parent"."table_2" SET "text" = $1 WHERE "id" = $2
```

### Expected behavior

In theory, this should be attempting to update "child.table_2" rather then "parent.table_2" based on the schema prefix provided

Based on @josevalim 's response to [this stack overflow](https://stackoverflow.com/questions/56348193/argumenterror-could-not-generate-inputs-for-association?noredirect=1) from May '19 it seems like `it assumes all schemas are in the same prefix.` - I was wondering if this was still the case, if it was something that would be fixable in the short term and what the reasoning behind this decision was.

Really appreciate all of your hard work and would be happy to dig into this if the change is reasonably scoped!

camilleryr:
Resolves #3200

### DIFF
--- a/test/ecto/repo_test.exs
+++ b/test/ecto/repo_test.exs
@@ -8,6 +8,18 @@ defmodule Ecto.RepoTest do
   defmodule MyParent do
     use Ecto.Schema
 
+    schema "my_parent" do
+      field :n, :integer
+    end
+
+    def changeset(struct, params) do
+      Ecto.Changeset.cast(struct, params, [:n])
+    end
+  end
+
+  defmodule MyParentWithPrefix do
+    use Ecto.Schema
+
     @schema_prefix "private"
 
     schema "my_parent" do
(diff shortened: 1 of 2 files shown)


test/ecto/repo_test.exs
```diff
@@ -656,7 +656,7 @@ defmodule Ecto.RepoTest do
       assert schema.__meta__.prefix == "private"
     end
 
-    test "insert, update and insert_or_update parent schema_prefix overrides children schema_prefix" do
+    test "insert, update and insert_or_update parent schema_prefix does not override children schema_prefix" do
```
josevalim: Do we have a test that the parent does override the child if the child has no prefix?

camilleryr: I have added two additional test cases so we now cover

josevalim:

PR #3206, opened 2020-01-15
Removing prefixed table name from constraints (see ecto_sql #175)

kuffel:
Hello,

i have reported this issue here: https://github.com/elixir-ecto/ecto_sql/issues/175 and i am not sure if this is the correct way of submitting my suggested fix for that in ecto.

Nevertheless, this fixes the issue in a backward compatible way. 

I have checked the documentation for mysql and postgres and "." seems to be a character which is not allowed for naming indexes.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -677,6 +677,8 @@ defmodule Ecto.Repo.Schema do
   defp constraints_to_errors(%{constraints: user_constraints, errors: errors} = changeset, action, constraints) do
     constraint_errors =
       Enum.map constraints, fn {type, constraint} ->
+        constraint = constraint |> String.split(".") |> Enum.at(-1)
+
         user_constraint =
           Enum.find(user_constraints, fn c ->
             case {c.type, c.constraint,  c.match} do


lib/ecto/repo/schema.ex
```diff
@@ -677,6 +677,8 @@ defmodule Ecto.Repo.Schema do
   defp constraints_to_errors(%{constraints: user_constraints, errors: errors} = changeset, action, constraints) do
     constraint_errors =
       Enum.map constraints, fn {type, constraint} ->
+        constraint = constraint |> String.split(".") |> Enum.at(-1)
```
josevalim: Hi @kuffel, this would change the code for all adapters. I think we want to do this change in the ecto_sql repository, in particular only to the MyXQL adapter. Can you please move it there? Thank you.

wojtekmach: the problem is the constraint name may have a dot inside and if that's the case this logic would give the wrong result. These are totally possible today:

```elixir
def change do
  create unique_index(:posts, [:uuid], name: "foo.bar"
end
```

```
mysql> CREATE UNIQUE INDEX `foo.bar` ON posts (uuid);
```

to be honest I'm not sure if people actually do that but it may be important for integrating with DBs that were not created with ecto migrations. WDYT?

josevalim: There is said issue as well. Ideally, if we have the table name in the constraint error, then we remove the table name plus the dot only. If not, we discard up to the first dot.

kuffel: Hi,

thanks for the feedback. I have found this function where the tablename seems to be in the `sql` parameter:
https://github.com/elixir-ecto/ecto_sql/blob/aad616d965492531b39750833c885b44143ce486/lib/ecto/adapters/sql.ex#L663

It has a format which looks useful:
```
[
  "UPDATE ",
  [[96, "oops", 96], 46, [96, "posts", 96]],
  " SET ",
  [[[], [[96, "title", 96], " = ?"], ", "], [96, "updated_at", 96], " = ?"],
  " WHERE ",
  [],
  [96, "id", 96],
  " = ?"
]
```

Can this be a place where we could extract the tablename and pass it to the conn.to_constraints function?

wojtekmach: Yeah, I think the solution would be to change to_constraints(error) callback to to_constraints(error, opts) and we’d pass opts = [source: source]. Wdyt @josevalim?

josevalim: @wojtekmach sounds good to me. We will need a way to keep it backwards compatible though. :/

kuffel: @josevalim @wojtekmach thanks for the ideas. I created a pull request here:

https://github.com/elixir-ecto/ecto_sql/pull/178

Is this going into the right direction? I still need to check why CI is failing.

kuffel: Ok, the CI seems to fails for quite some time, i have tested it locally with and it seeems fine.

PR #3217, opened 2020-01-29
Fix dynamic through assertion for has_many through assoc in integration tests

tspenov:
Since dynamic through assertion is not matched to query - it results in two
same assertions and not testing the dynamic case.


josevalim:

PR #3219, opened 2020-02-01
Fix select_merge on map with outer join

### LINKED ISSUE #3218: select_merge + map + join(:left) fails by returning nil instead of map
My colleague @iog-kc originally posted this on Elixir Forum [yesterday](https://elixirforum.com/t/select-merge-with-dynamic-fields-on-a-left-outer-join/28655), but the more I think about it, the more that it feels like there’s something not quite right with Ecto’s handling of this. I’d be happy to tackle this issue if it is an Ecto issue, but I don’t know where to begin.

We’re on Elixir 1.8.2, PostgreSQL 9.6, Postgrex 0.15, Ecto 3.2 (we will be on Ecto 3.3, etc. soon). We’re testing this on MacOS but expect this to happen on any OS. The code example included is a _deliberately_ simplified version of the problem; the fields in question _are_ dynamic as we’re building a translations strategy where different tables may have different translatable fields and we will be adding virtual fields to the parent schema for this to work.

### Current behavior

We have a query that does a left outer join on a child table and then uses select_merge to merge dynamically selected fields from the child row into the parent. However, when there is no child row, the call to map fails with cannot merge because the right side is not a map, got: nil. Is there a way to get the call to map to use an empty map for merging when there isn’t a row to merge?

As a simplified example (assume that Product exists and has virtual fields for the translated fields in Product.Translation), we have the function below where the translated_fields are provided contextually (for a product list, you might just want the name, but on a product display page, you want the description as well):

```elixir
def product_with_translations(sku_code, locale, translated_fields \\ ~w(name)a) do
  Product
  |> from(as: :product)
  |> where(sku_code: ^sku_code)
  |> join(
    :left, [product: p],
    t in Product.Translation,
    on: t.product_id == p.id and t.locale == ^locale,
    as: :translation
  )
  |> select([product: p], p)
  |> select_merge([translation: t], map(t, ^translated_fields))
end
```

As long as the associated Product.Translation exists, this works beautifully. However, if we were to ask for locale `de-DE` and don’t have it we receive an `ArgumentError`:

```elixir
iex> Repo.one(product_with_translations("123", "fr-CA", ~w(name description)a))
** (ArgumentError) cannot merge because the right side is not a map, got: nil
```

### Expected Behaviour

No exception should be thrown and the `name` and `description` fields should be `nil`. The `SELECT` works.

Can we dynamically fill a `map` in `select_merge` from a left outer join in some way that we’re not seeing, or is this a currently unhandled case in Ecto? I think that the behaviour of `map(t, ^translated_fields)` should result in the equivalent of `Map.take(t || %{}, translated_fields)`.

halostatue:
Resolves #3218

Three cases have been added to merge processing:

- If the merge argument (`right`) is `nil`, return `left` unmodified. This can
  happen with a `:left_join` or `:full_join`. This is fully tested.

- If the merge source (`left`) is `nil`, return `right` unmodified. This can
  happen with a `:right_join` or `:full_join`. This is not tested.

- If both the source and argument are `nil`, return `nil`. This _probably_
  shouldn’t happen, but is currently included for completeness. This is not
  tested.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -5,6 +5,7 @@ defmodule Ecto.Integration.RepoTest do
   import Ecto.Query
 
   alias Ecto.Integration.Post
+  alias Ecto.Integration.Post.Translation
   alias Ecto.Integration.Order
   alias Ecto.Integration.User
   alias Ecto.Integration.Comment
@@ -1237,7 +1238,7 @@ defmodule Ecto.Integration.RepoTest do
     end
 
     test "merge" do
-      %Post{} = TestRepo.insert!(%Post{title: "1", counter: nil})
+      %Post{id: post_id} = TestRepo.insert!(%Post{title: "1", counter: nil})
 
       # Merge on source
       assert [%Post{title: "2"}] =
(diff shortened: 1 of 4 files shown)


josevalim:
Thanks @halostatue, the PR looks great. Just one comment:

We generally like to avoid adding new fixtures/schemas/tables to integration tests, otherwise they would grow very unwieldy with time. Perhaps a simpler to add a `title` field to Permalink? This way Post in your tests become Permalink and Post.Translation becomes Post. You can use the current url field to be "1" or "2" and use it to make the join return a result or not. Could you please amend it accordingly?

halostatue:
Let me see what I can do. It will be much later today before I can get to it.

halostatue:
Done. I did add one field to `permalink` so that the distinction between overwriting an existing field and a virtual field are considered in the tests here.

josevalim:
Thank you, fantastic work! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #3223, opened 2020-02-07
Adopted integration tests for mssql adapter usage

mjaric:
Since there are technical differences how mssql handles types and implicit conversion between them, certain integration test can not pass when mssql adapter is set in ecto repo. With this in mind, this PR should allow reuse of integration test even with mssql adapter [will be](https://github.com/elixir-ecto/ecto_sql/pull/184) integrated in ecto_sql library.


integration_test/cases/assoc.exs
```diff
@@ -75,6 +75,7 @@ defmodule Ecto.Integration.AssocTest do
     assert [^u2, ^u1] = TestRepo.all(query)
   end
 
+  @tag :on_replace_nulify
```
josevalim:
```suggestion
  @tag :on_replace_nilify
```

integration_test/cases/assoc.exs
```diff
@@ -191,6 +192,7 @@ defmodule Ecto.Integration.AssocTest do
     assert [0] == TestRepo.all(from(p in Permalink, select: count(p.id)))
   end
 
+  @tag :on_replace_nulify
```
josevalim:
```suggestion
  @tag :on_replace_nilify
```

integration_test/cases/repo.exs
```diff
@@ -324,6 +324,7 @@ defmodule Ecto.Integration.RepoTest do
   end
 
   @tag :unique_constraint
+  @tag :unique_constraint_conflict
```
josevalim: The tests below also trigger conflict on unique_constraint. So is there a particular reason why this test does not pass on MSSQL?

integration_test/cases/repo.exs
```diff
@@ -763,8 +763,25 @@ defmodule Ecto.Integration.RepoTest do
     # With order_by and limit
     query = from Post, order_by: [asc: :visits], limit: 2
     assert TestRepo.aggregate(query, :max, :visits) == 12
+  end
+
+  @tag :db_engine_can_maintain_precision
```
josevalim:
```suggestion
  @tag :average_with_precision
```

WDYT?

integration_test/cases/type.exs
```diff
@@ -348,6 +364,7 @@ defmodule Ecto.Integration.TypeTest do
   end
 
   @tag :decimal_type
+  @tag :decimal_type_cast
   test "decimal type" do
```
josevalim: Should we break this test in two? Unless the last two assertions perform casting on select. Then we can call it `test "decimal type cast" do`.

integration_test/cases/type.exs
```diff
@@ -365,6 +382,7 @@ defmodule Ecto.Integration.TypeTest do
   end
 
   @tag :decimal_type
+  @tag :decimal_type_cast
   test "typed aggregations" do
```
josevalim:
```suggestion
  test "decimal typed aggregations" do
```

integration_test/cases/type.exs
```diff
@@ -91,13 +104,15 @@ defmodule Ecto.Integration.TypeTest do
     assert [^datetime] = TestRepo.all(query)
   end
 
+  @tag :coalesce_type
   test "coalesce type when default" do
```
josevalim: This test also has coalesce:

https://github.com/elixir-ecto/ecto/pull/3223/files#diff-32bcb9b1e35982eb56032061dfab4aa9L378

But it doesn't have a `@tag :coalesce_type`. So do you know why this test fails but that one does not? MSSQL doesn't support coalesce?

mjaric: trick is with types and how ms threat `CAST(value as DECIMAL)` for instance. Decimal without precision is `DECIMAL(X, 0) ` so in case `COALESCE(t1.coumn_name, CAST(1.0 AS DECIMAL))` will not work  since assert expects `#Decimal<1.0>` but in this case we have binary that will preserve all bytes and result is for sure `<<0, 1>>`. 

So coalesce work but CAST is the problem in some cases where precision is required, like in case of Decimal where default precision is 0.

It is worth mentioning here, tds will encode decimal correctly with correct precision, but adapter have no clue about parameter precision or scale used in `select` since `Ecto.Query.Api.type/2`  only pass parameter type that should be used in cast but not other info that is kept in migrations :(

mjaric: I should probably use different tag eg :decimal_casting

josevalim: Both `:decimal_type_cast` and `:db_engine_can_maintain_precision` seem to be related to decimal precision, so what if we have a `@tag :decimal_precision` for all of them?

Do the coalesce tests pass when they don't use decimal?

mjaric: I’m not at my desk but as I can recall, money and decimal types suffer from this in all casting scenarios in select statement

integration_test/cases/repo.exs
```diff
@@ -763,8 +762,25 @@ defmodule Ecto.Integration.RepoTest do
     # With order_by and limit
     query = from Post, order_by: [asc: :visits], limit: 2
     assert TestRepo.aggregate(query, :max, :visits) == 12
+  end
+
+  @tag :average_with_precision
```
josevalim: Following our discussion, this can probably be called `@tag :decimal_precision` too!

integration_test/cases/type.exs
josevalim: Both :decimal_type_cast and :db_engine_can_maintain_precision seem to be related to decimal precision, so what if we have a @tag :decimal_precision for all of them?

Do the coalesce tests pass when they don't use decimal?

mjaric: coalesce_type is issue where we map string to varbinary. This is part that I could not fix with [migration override](https://github.com/mjaric/ecto_sql/blob/tds-support/integration_test/mssql/migration.exs) 

1. If I don not override migration this MSSQL will complain about implicit conversion nvarchar to varbinary in about 20 test cases.  binary <<0, 2>> parameter tagged as type :string  (thar is nvarchar) and column is varbinary
2. if I override migration and set column to nvarchar (:string) in migration, then this 20 test passes except these 2 with tag :coalesce_type. And here problem is that decoder needs to conver ucs2 to utf-8 (fixed length to variable length) and that is how we get <<2>> in return, padding zero is removed during decoding.

Maybe if I rename tag to :string_encoding?

josevalim: Good point. :string_encoding or :varchar_encoding works. Your call!

integration_test/cases/type.exs
mjaric: renamed

integration_test/cases/type.exs
mjaric: renamed

integration_test/cases/repo.exs
mjaric: renamed

integration_test/cases/type.exs
josevalim: Also, remember to rename the one below too. :)

integration_test/cases/type.exs
```diff
@@ -381,6 +403,7 @@ defmodule Ecto.Integration.TypeTest do
     assert [^decimal] = TestRepo.all(from p in Post, select: coalesce(p.cost, 0))
   end
 
+  @tag :union_with_literals
```
josevalim: Btw, why does this test fail? Is it because of number precision? :)

mjaric: that could be some odd bug. line 420 fails since somehow %{n: 1.0} is decoded into decimal. 

```
  1) test unions with literals (Ecto.Integration.TypeTest)
     c:/prj/elixir-ecto/ecto/integration_test/cases/type.exs:407
     ** (ArgumentError) cannot load `#Decimal<1.0>` as type :float
     code: assert TestRepo.all(union_all(query1, ^query2)) ==
     stacktrace:
       (ecto) lib/ecto/repo/queryable.ex:382: Ecto.Repo.Queryable.process/4
       (ecto) lib/ecto/repo/queryable.ex:367: anonymous fn/4 in Ecto.Repo.Queryable.process_kv/4
       (elixir) lib/enum.ex:1440: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (ecto) lib/ecto/repo/queryable.ex:326: Ecto.Repo.Queryable.process/4
       (ecto) lib/ecto/repo/queryable.ex:246: anonymous fn/3 in Ecto.Repo.Queryable.postprocessor/4
       (elixir) lib/enum.ex:1336: Enum."-map/2-lists^map/1-0-"/2
       (ecto) lib/ecto/repo/queryable.ex:198: Ecto.Repo.Queryable.execute/4
       (ecto) lib/ecto/repo/queryable.ex:17: Ecto.Repo.Queryable.all/3
       c:/prj/elixir-ecto/ecto/integration_test/cases/type.exs:420: (test)

```

josevalim:

PR #3225, opened 2020-02-08
Use explicit match on fetch_change!/2

qcam:
Since `Ecto.Changeset.fetch_change/2` only returns `{:ok, term()} | :error`.


josevalim:

PR #3227, opened 2020-02-10
Add Ecto.Query.API.json_extract_path/2

wojtekmach:
Ref https://github.com/elixir-ecto/ecto/issues/3216

Ref: https://github.com/elixir-ecto/ecto_sql/pull/187

I plan to update this PR (or create a new one) with support for a similar `embed_extract_path/2` that works with embeds (we'd use it like this: `from o in Order, select: o.item.price`.)

### DIFF
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -281,6 +281,43 @@ defmodule Ecto.Integration.TypeTest do
   end
 
   @tag :map_type
+  @tag :json_extract_path
+  test "json_extract_path with primitive values" do
+    post = %Post{meta: %{:id => 123, :time => ~T[09:00:00], "'single quoted'" => "bar", "\"double quoted\"" => "baz"}}
+    TestRepo.insert!(post)
+
+    assert TestRepo.one(from p in Post, select: p.meta["id"]) == 123
+
+    assert TestRepo.one(from p in Post, select: p.meta["bad"]) == nil
+    assert TestRepo.one(from p in Post, select: p.meta["bad"]["bad"]) == nil
+
+    assert TestRepo.one(from p in Post, select: type(p.meta["id"], :string)) == "123"
+
+    field = "id"
+    assert TestRepo.one(from p in Post, select: p.meta[^field]) == 123
+
+    assert TestRepo.one(from p in Post, select: p.meta["time"]) == "09:00:00"
+
+    assert TestRepo.one(from p in Post, select: p.meta["'single quoted'"]) == "bar"
+    assert TestRepo.one(from p in Post, select: p.meta["';"]) == nil
+    assert TestRepo.one(from p in Post, select: p.meta["\"double quoted\""]) == "baz"
+  end
+
+  @tag :map_type
+  @tag :json_extract_path
+  test "json_extract_path with arrays and objects" do
+    post = %Post{meta: %{tags: [%{name: "red"}, %{name: "green"}]}}
+    TestRepo.insert!(post)
+
+    assert TestRepo.one(from p in Post, select: p.meta["tags"][0]["name"]) == "red"
+
+    index = 1
+    assert TestRepo.one(from p in Post, select: p.meta["tags"][^index]["name"]) == "green"
+
+    assert TestRepo.one(from p in Post, select: p.meta["tags"][99]["name"]) == nil
+  end
+
+  @tag :map_type
   @tag :map_type_schemaless
   test "embeds one with custom type" do
     item = %Item{price: 123, reference: "PREFIX-EXAMPLE"}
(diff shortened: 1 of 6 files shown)


lib/ecto/query/builder.ex
```diff
@@ -815,6 +831,34 @@ defmodule Ecto.Query.Builder do
   def field!(other),
     do: error!("expected atom in field/2, got: `#{inspect other}`")
 
+  defp escape_json_path(path) do
+    Enum.map(path, &quoted_json_path_element!/1)
+  end
+
+  defp quoted_json_path_element!({:^, _, [expr]}),
+    do: quote(do: Ecto.Query.Builder.json_path_element!(unquote(expr)))
```
wojtekmach: currently we simply interpolate the value into the query - if the value comes from user input it might be an SQL injection. (FWIW this is similar to how `field/2` works but given that one works on atoms, sql injections are less likely/impossible.)

 I've documented it as such, but I'm planning to see if I can handle this properly: interpolated values would be sent as params.

lib/ecto/query/builder.ex
```diff
@@ -815,6 +831,36 @@ defmodule Ecto.Query.Builder do
   def field!(other),
     do: error!("expected atom in field/2, got: `#{inspect other}`")
 
+  defp escape_json_path(path) do
```
josevalim: You want to explicitly assert it is a list at compile time, raise otherwise.

lib/ecto/query/builder.ex
```diff
@@ -815,6 +831,36 @@ defmodule Ecto.Query.Builder do
...
+  end
+
+  defp quoted_json_path_element!({:^, _, [expr]}),
+    do: quote(do: Ecto.Query.Builder.json_path_element!(unquote(expr)))
+
+  defp quoted_json_path_element!(binary) when is_binary(binary),
+    do: binary
+
+  defp quoted_json_path_element!(integer) when is_integer(integer),
+    do: integer
```
josevalim: Good call on keeping it as an integer here. We probably want to convert this to binary on the adapter.

test/ecto/query/builder_test.exs
```diff
@@ -38,6 +38,14 @@ defmodule Ecto.Query.BuilderTest do
            |> elem(0)
            |> Code.eval_quoted([], __ENV__)
            |> elem(0)
+
+    expected = {Macro.escape(quote do: json_extract_path(&0.y(), ["a", "b"])), []}
+    actual = escape(quote do x.y["a"]["b"] end, [x: 0], __ENV__)
+    assert actual == expected
```
josevalim: Also include tests for json_extract_path being called explicitly.

test/ecto/query/inspect_test.exs
```diff
@@ -270,6 +270,14 @@ defmodule Ecto.Query.InspectTest do
       ~s{from p0 in Inspect.Post, where: fragment(title: [foo: ^"foobar"])}
   end
 
+  test "json_extract_path" do
+    assert i(from(x in Post, select: x.meta["author"])) ==
+             ~s{from p0 in Inspect.Post, select: p0.meta[\"author\"]}
+
+    assert i(from(x in Post, select: x.meta["author"]["name"])) ==
+             ~s{from p0 in Inspect.Post, select: p0.meta[\"author\"][\"name\"]}
```
josevalim: Also include tests for json_extract_path being called explicitly.

lib/ecto/query/builder.ex
michalmuskala: AFAIK that's not true with `field` - we always quote the field names - https://github.com/elixir-ecto/ecto_sql/blob/wm-json-extract-path/lib/ecto/adapters/myxql/connection.ex#L463-L467

wojtekmach: Yup, good point. I've added proper escaping.

josevalim:
Let's wait for MSSQL to get in and then we can merge this.

integration_test/cases/type.exs
```diff
@@ -280,6 +280,55 @@ defmodule Ecto.Integration.TypeTest do
...
+  test "json_extract_path" do
+    post = %Post{meta: %{:id => "123", :time => ~T[09:00:00], "'single quoted'" => "bar", "\"double quoted\"" => "baz"}}
+    TestRepo.insert!(post)
+
+    assert TestRepo.one(from p in Post, select: p.meta["id"]) == "123"
+
+    assert TestRepo.one(from p in Post, select: p.meta["bad"]) == nil
+    assert TestRepo.one(from p in Post, select: p.meta["bad"]["bad"]) == nil
+
+    # assert TestRepo.one(from p in Post, select: type(p.meta["id"], :integer)) == 123
```
josevalim: Should we uncomment this?

PR #3231, opened 2020-02-16
Some minor touch-ups for queryable

qcam:


lib/ecto/repo/queryable.ex
```diff
@@ -307,10 +307,10 @@ defmodule Ecto.Repo.Queryable do
   defp process(row, {:struct, struct, args}, from, adapter) do
     {fields, row} = process_kv(args, row, from, adapter)
 
-    case Map.merge(struct.__struct__(), Map.new(fields)) do
+    case struct(struct, fields) do
```
qcam: Perhaps we could go stricter and use `Kernel.struct!/2`?

josevalim: So the reason why it was written like above is because `struct` does a bunch of checking and in this case we are sure all of the fields belong to the struct, so we can skip the checking. It is done for performance reasons. :)

lib/ecto/repo/queryable.ex
```diff
@@ -307,10 +307,10 @@ defmodule Ecto.Repo.Queryable do
...
       %{__meta__: %Ecto.Schema.Metadata{state: state} = metadata} = struct
       when state != :loaded ->
-        {Map.put(struct, :__meta__, %{metadata | state: :loaded}), row}
+        {Map.replace!(struct, :__meta__, %{metadata | state: :loaded}), row}
```
josevalim: Good call!

lib/ecto/repo/queryable.ex
qcam: I see! I will remove this then.

PR #3233, opened 2020-02-16
Simplify and optimize uniqueness validation

qcam:


lib/ecto/changeset.ex
```diff
@@ -1786,42 +1786,43 @@ defmodule Ecto.Changeset do
...
-    # If we don't have values for all fields, we can't query for uniqueness
-    any_nil_values_for_fields? = Enum.any?(where_clause, &(&1 |> elem(1) |> is_nil()))
+          :error ->
+            # If any field is not present, we cannot query for uniqueness.
+            {:halt, {true, []}}
+        end
+      end)
 
-    if unrelated_changes? || any_nil_values_for_fields? do
+    if skip_validation? do
```
josevalim: I find the previous approach much more readable, so I would propose to keep this part the same. Perhaps simply rewrite:

```
unrelated_changes? = fields -- Map.keys(changeset.changes) == fields
```

to 

```
unrelated_changes? = Enum.all?(fields, &not Map.has_key?(changeset.changes, &1))
```

I am not super worried about performance because fields is going to be 5 elements at worse, so the multiple traversals do not really matter. :)

lib/ecto/changeset.ex
```diff
@@ -1786,42 +1786,43 @@ defmodule Ecto.Changeset do
...
-        else
-          Enum.reduce(pk_pairs, struct, fn {field, value}, acc ->
-            Ecto.Query.or_where(acc, [q], field(q, ^field) != ^value)
-          end)
-        end
-
       query =
-        pk_query
+        schema
+        |> maybe_exclude_itself(changeset)
```
josevalim: I like this one a lot though, good job! :)

lib/ecto/changeset.ex
qcam: yeah, I don't feel very strong about that too.

josevalim:

PR #3239, opened 2020-02-28
Show meaningful error in case of doubled schema

hauleth:
Earlier it returned information about `id` being defined twice, which could be confusing for the users as in most cases this field wasn't defined anywhere. Now it will raise `RuntimeError` with meaningful error that contain line with first schema.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -453,7 +453,7 @@ defmodule Ecto.Schema do
   `@primary_key` attribute.
   """
   defmacro embedded_schema([do: block]) do
-    schema(nil, false, :binary_id, block)
+    schema(__ENV__, nil, false, :binary_id, block)
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/ecto/schema.ex
```diff
@@ -453,7 +453,7 @@ defmodule Ecto.Schema do
   `@primary_key` attribute.
   """
   defmacro embedded_schema([do: block]) do
-    schema(nil, false, :binary_id, block)
+    schema(__ENV__, nil, false, :binary_id, block)
```
josevalim: This should be `__CALLER__`. I recommend you to check the line in your test as it would have caught this error. :)

lib/ecto/schema.ex
```diff
@@ -464,12 +464,20 @@ defmodule Ecto.Schema do
...
 
-  defp schema(source, meta?, type, block) do
+  defp schema(caller, source, meta?, type, block) do
     prelude =
       quote do
+        if Module.get_attribute(__MODULE__, :ecto_schema_defined) do
+          raise """
+          Schema already defined for #{inspect(__MODULE__)} on line #{@ecto_schema_defined}
+          """
+        end
```
josevalim:
```suggestion
        if line = Module.get_attribute(__MODULE__, :ecto_schema_defined) do
          raise "schema already defined for #{inspect(__MODULE__)} on line #{line}"
        end
```

lib/ecto/schema.ex
hauleth: I cannot make `Code.compile_quoted` to return proper lines so I fall back to compiling string instead.

josevalim:

hauleth:
It probably should be expanded to the `embedded_schema` as well, but I do not use them, so I have forgot about it.

PR #3241, opened 2020-03-07
adding missing insert

mjaric:
During previous test case adoption for mssql adapter, mistakenly one insert is left out during refactoring, so test is failing for pg and myxql adapters.

This PR should fix this test


wojtekmach:
Thanks!

PR #3242, opened 2020-03-08
CTE: allow dynamic table name + better error messages.

azizk:
Hi! I found it a bit restrictive that you can't define a CTE's table name with the contents of a variable. Looks like it works nicely with my changes. What do you think? :-)


josevalim:

PR #3250, opened 2020-03-20
Support default options (such as a default prefix) for a repo

ryanbjones:
Summary:

We'd like to be able to connect to a shard similar to `put_dynamic_repo` where all queries in the process will hit a particular schema without having to manually pass in the prefix on all possible queries. This also removes the need to have a repo specifically connected to the the schema (say through a `after_connect` function), which would increase the number of connections to the DB. 

Original post below: 

This is a POC (unit tests to be added if like idea). 

Problem trying to solve is not having to message pass the organization_id through the process and to not have to manually write it into every query in the application

image

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -214,7 +214,7 @@ defmodule Ecto.Repo do
 
       def with_default_opts(opts) do
         default_opts = Process.get({__MODULE__, :default_opts}, [])
-        Keyword.merge(opts, default_opts)
+        Keyword.merge(default_opts, opts)
       end
       ## Transactions
 
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -200,6 +200,10 @@ defmodule Ecto.Repo do
 
       @compile {:inline, get_dynamic_repo: 0}
 
+      def default_opts(opts \\ []) do
+        Process.put({__MODULE__, :default_opts}, opts)
+      end
+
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @ryanbjones!

I totally understand why this is necessary but I am afraid of promoting more process dictionary usage out of the box. We did it for dynamic repo but we really had no other option in there. However, I do want to make this possible.

One very straight-forward suggestion is for us to call this `default_options/1`. It will receive the name of the operation as argument and return a keyword list. Then, in your apps you can do:

    def default_options(_), do: Process.get({__MODULE__, :opts}, [])

And you can also add the functions for managing the process dict directly to your repo. I feel this is a preferable solution as it allows static configuration for those who only need something static but if you need the pdict or something more complex, you can implement it too.

Thoughts?

ryanbjones:
So if I understand correctly, in `Ecto` we'd have

```elixir
def default_options(_operation_name), do: []
defoverridable default_options: 1


def with_default_opts(operation_name, opts) do
  Keyword.merge(default_opts(operation_name), opts)
end
```
and then in the queries we'd see something like

```elixir
def all(queryable, opts \\ []) do
  Ecto.Repo.Queryable.all(get_dynamic_repo(), queryable, with_default_opts(:all, opts))
end
```

Is that correct?

josevalim:
Correct!

ryanbjones:
Updated

josevalim:

PR #3255, opened 2020-03-25
Improve warning when a struct or similar is used with constraints

JonRowe:
When you accidentally use a schema (or struct) with a constraint you get a cryptic pattern match error, this adds a second fallback pattern match to raise an `ArgumentError` with a more descriptive message.

I looked for a place to add a test, but the other fall back isn't tested that I could find? Happy for suggested improvements...

This originated from someone issue in the ecto channel on the elixir slack...


josevalim:

PR #3256, opened 2020-03-25
Also expand_alias on Schema field

lasseebert:
Ref the original PR that introduces expand_alias: https://github.com/elixir-ecto/ecto/pull/1670

When the type is a custom Ecto.Type, this will now create a runtime dependency on the type module instead of a compile time dependency.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -617,6 +617,7 @@ defmodule Ecto.Schema do
 
   """
   defmacro field(name, type \\ :string, opts \\ []) do
+    type = expand_alias(type, __CALLER__)
     quote do
       Ecto.Schema.__field__(__MODULE__, unquote(name), unquote(type), unquote(opts))
     end


michalmuskala:
I think this would be incorrect - the type module is used at compile-time to run some checks. Removing the compile-time dependency could lead to issues. 

https://github.com/elixir-ecto/ecto/blob/84ac50ffabf27420ccf39ada52fd33c6b815ef3b/lib/ecto/schema.ex#L2033

josevalim:
Good catch. We would need to move those checks to the `after_compile` hook too.

lasseebert:
Ah, I understand. So if my custom ecto type module changes to no longer be a valid ecto type, the schema should be recompiled to notice it, right?

michalmuskala:
Yes, exactly

lasseebert:
I'm afraid I don't understand how moving the compile-time check to `after_compile` would solve recompilation of the schema when the type changes. I tried to read through some of the after_compile code for associations, and I'm not getting it :P 

Could you provide some rough guidelines on how to solve it? Or should it not be touched, because it is exactly the way it should be?

josevalim:
Actually, I went to give it a try and we need to be aware of something. If we stop tracking types at compile time and for some reason you make the type invalid, then we won't catch it at compile-time.

For associations, it kinda works because the validations were introduced as warnings. It means that sometimes you may get the warning, sometimes you may not. But since we raise for types, it means that compilation may work when partial recompiling but fail when compiling the whole project. So for now I would prefer to not merge this. Thanks!

lasseebert:
Ok, thanks for taking your time for this :)

Issue #3263, opened 2020-04-03
Support where in (subquery)  


thenrio:
Summary
---

New expression `where([t], t.id in subquery(s))`, that yield sql like `where id in (s)`. 

> syntax MAY change.

Motivation
---
Main driver is `where x in (subquery)` MAY perform better than `join (select distinct x from subquery) s`, on postgres at least.
With this change, we can choose to build an expression that performs better on database.

Detailed design
---
Build a new expression in `Ecto.Queries.wheres`.
Make a PR and have feedback?

See [discussion](https://groups.google.com/forum/#!topic/elixir-ecto/ZRinU3ElLmg) for details and figures.

Drawbacks
---
Need to change ecto_sql also, to act on new expression and output desired sql.
See https://github.com/elixir-ecto/ecto_sql/issues/218.

azizk:
I would love to see this accepted. Is there a good reason why the subquery function is necessary? I would prefer `where([t], t.id in ^query)`.

Besides being a cool feature, in my case it would tremendously ease a pain point I have with filtering and paginating/sorting queries which I haven't solved properly yet. If I use 1:n-joins, I get duplicate rows. Using :distinct prevents sorting and forces ordering by id. Using :group_by becomes awkward fast and I have to give up. Last option I see may be a CTE which selects the filtered rows first and then I continue with that result set... Anyway, this feature would elegantly solve my issues I think.

thenrio:
I chose `subquery(s)` because:

* subquery is a term in postgres and mysql documentation in this context. It's "where x in (subquery)" rather than "where x in (query)". (see syntaxes on https://github.com/elixir-ecto/ecto_sql/issues/218).
* `in subquery(s)` is already a valid ecto expression for join (see https://github.com/elixir-ecto/ecto/blob/5e5662babc05853c1fdaf8a61f6590e4891b8974/lib/ecto/query.ex#L887).
* change in builder was minimal? (YMMV)

azizk:
Okay, I see what you mean. But I'm not sure if it should automatically apply to "ecto's language". In the docs for `join()` you'll see an example where a query is interpolated directly in the join clause without the `subquery()` function.
```elixir
posts =
  if params["drafts"] do
    from p in Post, where: [drafts: true]
  else
    from p in Post, where: [public: true]
  end

from c in Comment,
  join: p in ^posts, on: [id: c.post_id],
  select: {p.title, c.text}
```
[...]

PR #3264, opened 2020-04-03
support where in (subquery)

thenrio:
Proposal for #3263 .

Cheers!

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -318,7 +318,7 @@ defmodule Ecto.Query.Builder do
     rtype = {:in, quoted_type(left, vars)}
 
     {left, params_acc} = escape(left, ltype, params_acc, vars, env)
-    {right, params_acc} = escape(right, rtype, params_acc, vars, env)
+    {right, params_acc} = escape_subquery(right, rtype, params_acc, vars, env)
 
     # Remove any type wrapper from the right side
     right =
@@ -449,6 +449,13 @@ defmodule Ecto.Query.Builder do
     type = validate_type!(type, vars, env)
     {expr, params_acc} = escape(expr, type, params_acc, vars, env)
     {{:{}, [], [:type, [], [expr, Macro.escape(type)]]}, params_acc}
+  end
+
+  defp escape_subquery({:subquery, _, [expr]}, _, params, _vars, _env) do
+    {quote(do: Ecto.Query.subquery(unquote(expr))), params}
+  end
+  defp escape_subquery(expr, type, params, vars, env) do
+    escape(expr, type, params, vars, env)
   end
 
   defp wrap_nil(params, {:{}, _, [:^, _, [ix]]}), do: wrap_nil(params, ix, [])
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -451,8 +451,11 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:type, [], [expr, Macro.escape(type)]]}, params_acc}
   end
 
-  defp escape_subquery({:subquery, _, [expr]}, _, params, _vars, _env) do
-    {quote(do: Ecto.Query.subquery(unquote(expr))), params}
+  defp escape_subquery({:subquery, _, [expr]}, _, {params, subqueries}, _vars, _env) do
+    subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
+    index = length(subqueries)
+    expr = {:subquery, index}
+    {expr, {[expr | params], subqueries ++ [subquery]}}
   end
   defp escape_subquery(expr, type, params, vars, env) do
     escape(expr, type, params, vars, env)
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -454,8 +454,8 @@ defmodule Ecto.Query.Builder do
   defp escape_subquery({:subquery, _, [expr]}, _, {params, subqueries}, _vars, _env) do
     subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
     index = length(subqueries)
-    expr = {:subquery, index}
-    {expr, {[expr | params], subqueries ++ [subquery]}}
+    expr = {:subquery, index} # used both in ast and in parameters, as a placeholder.
+    {expr, {[expr | params], [subquery | subqueries]}}
   end
   defp escape_subquery(expr, type, params, vars, env) do
     escape(expr, type, params, vars, env)
--- a/lib/ecto/query/builder/filter.ex
+++ b/lib/ecto/query/builder/filter.ex
@@ -12,7 +12,8 @@ defmodule Ecto.Query.Builder.Filter do
   or a keyword list of field names and values. In a keyword
   list multiple key value pairs will be joined with "and".
 
-  Returned is {expression, {params, subqueries}} which is a valid escaped expression, see Macro.escape/2
+  Returned is {expression, {params, subqueries}} which is a valid escaped expression, see Macro.escape/2.
+  Both params and subqueries are reversed.
   """
   @spec escape(:where | :having | :on, Macro.t, non_neg_integer, Keyword.t, Macro.Env.t) :: {Macro.t, {list, list}}
   def escape(_kind, [], _binding, _vars, _env) do
@@ -61,7 +62,9 @@ defmodule Ecto.Query.Builder.Filter do
   def build(kind, op, query, binding, expr, env) do
     {query, binding} = Builder.escape_binding(query, binding, env)
     {expr, {params, subqueries}} = escape(kind, expr, 0, binding, env)
-    params = Builder.escape_params(params)
+
+    params = Enum.reverse(params)
+    subqueries = Enum.reverse(subqueries)
 
     expr = quote do: %Ecto.Query.BooleanExpr{
                         expr: unquote(expr),
--- a/lib/ecto/query/builder/filter.ex
+++ b/lib/ecto/query/builder/filter.ex
@@ -11,35 +11,36 @@ defmodule Ecto.Query.Builder.Filter do
   It allows query expressions that evaluate to a boolean
   or a keyword list of field names and values. In a keyword
   list multiple key value pairs will be joined with "and".
+
+  Returned is {expression, {params, subqueries}} which is a valid escaped expression, see Macro.escape/2
   """
-  @spec escape(:where | :having | :on, Macro.t, non_neg_integer, Keyword.t, Macro.Env.t) :: {Macro.t, []}
+  @spec escape(:where | :having | :on, Macro.t, non_neg_integer, Keyword.t, Macro.Env.t) :: {Macro.t, {list, list}}
   def escape(_kind, [], _binding, _vars, _env) do
-    {true, []}
+    {true, {[], []}}
   end
 
   def escape(kind, expr, binding, vars, env) when is_list(expr) do
-    {parts, params} =
-      Enum.map_reduce(expr, [], fn
-        {field, nil}, _params ->
+    {parts, params_subqueries} =
+      Enum.map_reduce(expr, {[], []}, fn
+        {field, nil}, _params_subqueries ->
           Builder.error! "nil given for `#{field}`. Comparison with nil is forbidden as it is unsafe. " <>
                          "Instead write a query with is_nil/1, for example: is_nil(s.#{field})"
-        {field, value}, params when is_atom(field) ->
+        {field, value}, params_subqueries when is_atom(field) ->
           value = check_for_nils(value, field)
-          {value, {params, :acc}} = Builder.escape(value, {binding, field}, {params, :acc}, vars, env)
-          {{:{}, [], [:==, [], [to_escaped_field(binding, field), value]]}, params}
-        _, _params ->
+          {value, params_subqueries} = Builder.escape(value, {binding, field}, params_subqueries, vars, env)
+          {{:{}, [], [:==, [], [to_escaped_field(binding, field), value]]}, params_subqueries}
+        _, _params_subqueries ->
           Builder.error! "expected a keyword list at compile time in #{kind}, " <>
                          "got: `#{Macro.to_string expr}`. If you would like to " <>
                          "pass a list dynamically, please interpolate the whole list with ^"
       end)
 
     expr = Enum.reduce parts, &{:{}, [], [:and, [], [&2, &1]]}
-    {expr, params}
+    {expr, params_subqueries}
   end
 
   def escape(_kind, expr, _binding, vars, env) do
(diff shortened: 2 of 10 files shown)


thenrio:
Just updated the PR to solve conflict.

Cheers.

lib/ecto/query/planner.ex
```diff
@@ -540,6 +547,17 @@ defmodule Ecto.Query.Planner do
     end
   end
 
+  @spec plan_wheres(Ecto.Query.t, atom) :: Ecto.Query.t
+  defp plan_wheres(q, adapter) do
+    wheres = q.wheres |> Enum.map(fn
+      %BooleanExpr{expr: {:in, [], [left, %Ecto.SubQuery{} = s]}} = where ->
+        %{where | expr: {:in, [], [left, plan_subquery(s, s.query.prefix || q.prefix, adapter)]}}
```
josevalim, line 554: Unfortunately this will only work if the `in` expression is at the root. So if you do this: `where: foo.bar and foo.baz in subquery(...)`, then it won't work. We would need to traverse the whole `where` to find subqueries, but this could be very expensive.

Perhaps a better option here is to add a new field to BooleanExpr called `:subqueries`. So `foo in subquery(...)` becomes at the AST level something like `foo in {:subquery, 0}` where `0` is the index of the subquery in the  `:subqueries` field.

lib/ecto/query/planner.ex
```diff
@@ -646,6 +664,16 @@ defmodule Ecto.Query.Planner do
   defp expr_to_cache(%QueryExpr{expr: expr}), do: expr
   defp expr_to_cache(%SelectExpr{expr: expr}), do: expr
 
+  defp cast_and_merge_params(:where, _query, %BooleanExpr{expr: {:in, [], [_left, %Ecto.SubQuery{params: p}]}}, params, _adapter) do
```
josevalim, line 667: Similarly, this only works if the `:in` is at the root. We also need to return the subquery cache key.

josevalim:
Hi @thenrio! I have dropped some comments. The general structure is good but we need to make it recursive - however making it recursive can be expensive. So my recommendation is to store subqueries in an index in the BooleanExpr field and we put them back in the expression when doing the normalization (i.e. during prewalk).

josevalim:
I had an idea on how to make this work without a specialized subqueries field. I will follow up shortly. :)

josevalim:
> I had an idea on how to make this work without a specialized subqueries field. I will follow up shortly. :)

Nope, it didn't work. We will need a subquery field on BooleanExpr and find a way to pass the subquery when escaping to the caller.

thenrio:
Hello @josevalim. Here is an attempt using a subqueries field in BooleanExpr.

As mentioned in last commit message, I used a placeholder in parameters, so that they are in order in depth first traversal order. Then maybe a `in ^s` or `in ^subquery(s)` would fit better in current code?

I'm all ears :).

And by the way, what was the morning idea?

Cheers.

lib/ecto/query/planner.ex
thenrio: Addressed by c87008e and 2c7fd0fe.

lib/ecto/query/builder/filter.ex
```diff
@@ -59,13 +61,16 @@ defmodule Ecto.Query.Builder.Filter do
 
   def build(kind, op, query, binding, expr, env) do
     {query, binding} = Builder.escape_binding(query, binding, env)
-    {expr, params} = escape(kind, expr, 0, binding, env)
-    params = Builder.escape_params(params)
+    {expr, {params, subqueries}} = escape(kind, expr, 0, binding, env)
+
+    params = Enum.reverse(params)
```
josevalim, line 66:
```suggestion
    params = Builder.escape_params(params)
```

lib/ecto/query/builder.ex
```diff
@@ -451,6 +451,16 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:type, [], [expr, Macro.escape(type)]]}, params_acc}
   end
 
+  defp escape_subquery({:subquery, _, [expr]}, _, {params, subqueries}, _vars, _env) do
+    subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
+    index = length(subqueries)
+    expr = {:subquery, index} # used both in ast and in parameters, as a placeholder.
+    {expr, {[expr | params], [subquery | subqueries]}}
```
josevalim, line 458: Why do we need to store it in params too?

josevalim: Oh, we need to know the other we have to inject the subquery parameters. :)

lib/ecto/query/planner.ex
```diff
@@ -642,21 +659,40 @@ defmodule Ecto.Query.Planner do
     end
   end
 
-  defp expr_to_cache(%BooleanExpr{op: op, expr: expr}), do: {op, expr}
+  defp expr_to_cache(%BooleanExpr{op: op, expr: expr, subqueries: []}), do: {op, expr}
+  defp expr_to_cache(%BooleanExpr{op: op, expr: expr, subqueries: subqueries}) do
+    # Alternate implementation could be replace {:subquery, i} expression in expr.
+    # Current strategy appends [{:subquery, i, cache}], where cache is the cache key for this subquery.
+    {op, expr, subqueries |> Enum.with_index() |> Enum.map(fn {%{cache: cache}, i} -> {:subquery, i, cache} end)}
```
josevalim, line 666: I don't think we need an index. :)

```suggestion
    {op, expr, Enum.map(subqueries, fn %{cache: cache} -> {:subquery, cache} end)}
```

lib/ecto/query/inspect.ex
```diff
@@ -268,6 +268,21 @@ defimpl Inspect, for: Ecto.Query do
...
+    # Macro.to_string/2? we do not receive {a, b} in callback but {:{}, [], [a, b]}...
+    #
+    # iex(9)> Macro.to_string({1, 2}, fn ast, s -> IO.inspect(ast); s end)
+    # 1
+    # 2
+    # {:{}, [], [1, 2]}
+    # "{1, 2}"
+    # see https://github.com/elixir-lang/elixir/blob/27bd9ffcc607b74ce56b547cb6ba92c9012c317c/lib/elixir/lib/macro.ex#L932
+    #
+    inspect_source(Enum.at(subqueries, i))
```
josevalim, line 283:
```suggestion
    inspect_source(Enum.fetch!(subqueries, i))
```

lib/ecto/query/planner.ex
```diff
@@ -1022,6 +1058,12 @@ defmodule Ecto.Query.Planner do
     {{:in, in_meta, [left, right]}, acc}
   end
 
+  defp prewalk({:in, in_meta, [left, {:subquery, i}]}, kind, query, expr, acc, adapter) do
+    {left, acc} = prewalk(left, kind, query, expr, acc, adapter)
+    {right, acc} = prewalk_source(Enum.at(expr.subqueries, i), kind, query, expr, acc, adapter)
```
josevalim, line 1063:
```suggestion
    {right, acc} = prewalk_source(Enum.fetch!(expr.subqueries, i), kind, query, expr, acc, adapter)
```

test/ecto/query/planner_test.exs
```diff
@@ -229,6 +233,78 @@ defmodule Ecto.Query.PlannerTest do
...
+    q = from(c in Comment, where: c.post_id in subquery(p))
+
+    q = q |> plan() |> elem(0)
+
+    assert [%{expr: expr, subqueries: [subquery]}] = q.wheres
+    assert {:in, [], [{{:., [], [{:&, [], [0]}, :post_id]}, [], []}, {:subquery, 0}]} = expr
+    assert %Ecto.SubQuery{} = subquery
+  end
+
+  test "plan: where in subquery, params (TODO)" do
```
josevalim, line 247: You can merge this with the test above.

test/ecto/query/planner_test.exs
```diff
@@ -229,6 +233,78 @@ defmodule Ecto.Query.PlannerTest do
...
+  test "plan: where x and in subquery, params (TODO)" do
+    p = from(p in Post, select: p.id, where: p.id in ^[2, 3])
+    q = from(c in Comment, where: c.text == ^"1" and c.post_id in subquery(p))
+
+    params = q |> plan() |> elem(1)
+
+    assert params == ["1", 2, 3]
+  end
+
+  test "plan: where in subquery and x, params (TODO)" do
```
josevalim, line 265: You can merge this with the test above with parameters before and after.

lib/ecto/query/planner.ex
```diff
@@ -642,21 +659,40 @@ defmodule Ecto.Query.Planner do
...
 
+  defp cast_and_merge_param(kind, query, expr, adapter, v, type, acc, cacheable?) do
+    case cast_param(kind, query, expr, v, type, adapter) do
+      {:in, v} ->
+        {Enum.reverse(v, acc), false}
+
+      v ->
+        {[v | acc], cacheable?}
+    end
+  end
```
josevalim, line 694:
```suggestion
    Enum.reduce expr.params, {params, true}, fn
      {:subquery, i}, {acc, cacheable?} ->
        # This is the place holder to intersperse subquery parameters.
        %Ecto.SubQuery{params: subparams, cache: cache} = Enum.fetch!(expr.subqueries, i)
        {Enum.reverse(subparams, acc), cacheable? and cache != :nocache}

      {v, type}, {acc, cacheable?} ->
        case cast_param(kind, query, expr, v, type, adapter) do
          {:in, v} -> {Enum.reverse(v, acc), false}
          v -> {[v | acc], cacheable?}
        end
    end
```

josevalim:
Awesome job @thenrio! I have left some minor comments. Also, please finish the tests and move most of them to ecto/query/subquery_test.exs and we should be good to go.

lib/ecto/query/inspect.ex
```diff
@@ -268,6 +268,21 @@ defimpl Inspect, for: Ecto.Query do
...
+    #
+    # Macro.to_string/2? we do not receive {a, b} in callback but {:{}, [], [a, b]}...
+    #
+    # iex(9)> Macro.to_string({1, 2}, fn ast, s -> IO.inspect(ast); s end)
+    # 1
+    # 2
+    # {:{}, [], [1, 2]}
+    # "{1, 2}"
+    # see https://github.com/elixir-lang/elixir/blob/27bd9ffcc607b74ce56b547cb6ba92c9012c317c/lib/elixir/lib/macro.ex#L932
+    #
```
josevalim, line 282:
```suggestion
    # We were supposed to match on {:subquery, i} but Elixir incorrectly
    # translates those to `:{}` when converting to string.
    # See https://github.com/elixir-lang/elixir/blob/27bd9ffcc607b74ce56b547cb6ba92c9012c317c/lib/elixir/lib/macro.ex#L932
```

josevalim:
Awesome progress! Please let me know once it is good to go. :)

thenrio:
Added a runtime check at normalize step.
I think we could do it at compile time, I tried and failed once :)
What do you think?

josevalim:
@thenrio checks on normalize is fine, because they are cached. :)

lib/ecto/query/planner.ex
```diff
@@ -1682,6 +1718,31 @@ defmodule Ecto.Query.Planner do
...
+
+  @spec types!(Ecto.Query.t, expr :: Macro.t | Ecto.SubQuery.t) :: [atom]
+  defp types!(query, {{:., [], [{:&, [], [ix]}, field]}, [], []} = expr) do
+    [type!(:all, query, expr, ix, field)]
+  end
+  defp types!(_query, %Ecto.SubQuery{select: select}) do
+    case select do
+      {:map, fields} -> for {_field, {_kind, type}} <- fields, do: type
+      {:struct, _schema, fields} -> for {_field, {_kind, type}} <- fields, do: type
+    end
```
josevalim, line 1743: What happens if the subquery returns multiple fields? Such as `%{id: post.id, name: post.name}`? Or `map(Post, [:id, :name])`? Shouldn't we validate that only a single field is allowed?

test/ecto/query/subquery_test.exs
```diff
@@ -310,4 +375,33 @@ defmodule Ecto.Query.SubqueryTest do
       normalize(from(p in subquery(subquery), select: [:title]))
     end
   end
+
+  defp macro_arg_at(m, i) do
+    m |> elem(2) |> Enum.at(i)
+  end
```
josevalim, line 381: It is best to pattern match instead of elem(2) |> Enum.at. Pattern matching always asserts more about the shape. :)

lib/ecto/query/planner.ex
josevalim: Oh, I see your code raises if there are multiple fields... but I think this code would be better:

```elixir
   case select do
      {:map, [{_field, {_kind, type}}]} -> type
      {:struct, _schema, [{_field, {_kind, type}}]} -> type
      _ ->
        error!(subquery, "subquery must return a single field in order to be used on the right-side of `in`")
    end
```

WDYT?

test/ecto/query/subquery_test.exs
```diff
@@ -310,4 +375,33 @@ defmodule Ecto.Query.SubqueryTest do
...
+  test "normalize: where in subquery with too many selected expressions" do
+    # Otherwise we face database error, such as
+    # (Postgrex.Error) ERROR 42601 (syntax_error) subquery has too many columns
+    #
+    # Alternate design MAY be a expansion time, Ecto.Query.CompileError, such as 
+    # "only one select expression is allowed in query"
+    assert_raise Ecto.QueryError, ~r/^incorrect type of expressions in subquery/, fn ->
+      p = from(p in Post)
+      from(c in Comment, where: c.post_id in subquery(p)) |> normalize()
+    end
```
josevalim, line 404: We need to break this test in two:

1. One about actual mismatched types (i.e. `c.post_id in subquery(that_returns_a_string_column)`)
2. One about returning too many fields :)

lib/ecto/query/planner.ex
```diff
@@ -1682,6 +1718,31 @@ defmodule Ecto.Query.Planner do
     end
   end
 
+  @spec assert_same_types!(Ecto.Query.t, left :: Macro.t, right :: Ecto.SubQuery.t) :: true
+  defp assert_same_types!(query, left, right) do
```
josevalim, line 1722: Since this is subquery specific, let's call it `assert_in_subquery_type!` or similar.

josevalim:
I have added some notes, I will take them directly though! Thank you very much! Can you please send a PR to ecto_sql that adds coverage to `test/ecto/adpaters/*_test.exs`?

josevalim:

thenrio:
:tada:, Thanks @josevalim.

ckruse:
Woah, great news! Thanks for your work, both of you!

PR #3268, opened 2020-04-06
Emit telemetry event upon Repo start

### LINKED ISSUE #3266: Submit a telemetry event upon Repo start
One of the challenges for an auto-instrumentation library like New Relic is providing a zero-configuration APM experience.

To do that, the agent currently has to do some gymnastics to discover Repos as they start, then detect their configuration & `telemetry` event name. Right now this is done via erlang tracers & reading Application config.

This could be achieved in a much friendlier way if Ecto itself submitted a `telemetry` event whenever a Repo is started that contained all the configuration, including the `telemetry` event prefix that it will use. An event handler could then trigger the subscription to further events. The only thing we'd need is to make sure the event name for the "start" events is known up front!

Something roughly like this:

```elixir
:telemetry.execute(
  [:ecto, :repo, :start],
  %{system_time: System.system_time()},
  %{
    repo: MyRepo,
    config: %{
      telemetry_prefix: [:my_repo, :query],
      hostname: hostname,
      # ...
    }
  }
)
```

Is this something we might be able to add?

binaryseed:
This PR enables `telemetry` handlers to listen to an event that signals that a new `Repo` has been started. This feature enables automatic instrumentation for Repos, whose telemetry events are not knowable ahead of time.

@josevalim I tried this out in the New Relic agent and it works quite nicely. The only other thought I had was that perhaps this is a more general pattern that could be recommended by `telemetry`, which led to the thought that the event name itself could be something like `[:telemetry, :broadcast]` or something. I'm not totally sure about that, but I do like this pattern.

closes #3266

### DIFF
--- a/lib/ecto/repo/supervisor.ex
+++ b/lib/ecto/repo/supervisor.ex
@@ -169,6 +169,12 @@ defmodule Ecto.Repo.Supervisor do
   def init({name, repo, otp_app, adapter, opts}) do
     case runtime_config(:supervisor, repo, otp_app, opts) do
       {:ok, opts} ->
+        :telemetry.execute(
+          [:ecto, :repo, :start],
+          %{system_time: System.system_time()},
+          %{repo: repo, opts: opts}
+        )
+
         {:ok, child, meta} = adapter.init([repo: repo] ++ opts)
         cache = Ecto.Query.Planner.new_query_cache(name)
         meta = Map.merge(meta, %{repo: repo, cache: cache})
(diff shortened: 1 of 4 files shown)


lib/ecto/repo/supervisor.ex
```diff
@@ -169,6 +169,12 @@ defmodule Ecto.Repo.Supervisor do
   def init({name, repo, otp_app, adapter, opts}) do
     case runtime_config(:supervisor, repo, otp_app, opts) do
       {:ok, opts} ->
+        :telemetry.execute(
+          [:ecto, :repo, :start],
```
josevalim, line 173: I wouldn't use start in here because people may think this is a trace and that eventually we will have a stop/exception. What do you think about init or boot?

binaryseed: Agreed, I'll change to `init` to match the action that is actually happening

josevalim:

binaryseed:
Awesome! Looking forward to the next release to leverage this!

Issue #3269, opened 2020-04-06
Repo.embedded_load/3 convenience


narrowtux:
When I write queries that filter within embedded fields that are backed by JSON, I see no simple way to select the filtered embeds so they're correctly loaded using the ecto type system.

My example goes like this:

```elixir
      limits_query = from l in License,
        select: %{
          limit: fragment("jsonb_array_elements(?)", l.limits),
          valid_from: l.valid_from,
          valid_until: l.valid_until
        }

      query = from l in subquery(limits_query),
        select: l.limit,
        where: fragment("? >= now()", l.valid_until),
        where: fragment("? <= now()", l.valid_from),
        where: fragment("(? -> 'category') = ?::jsonb", l.limit, ^category)
```

The `License` schema has `embeds_many :limits, Limit` which is backed by a `jsonb` field.

What I tried to load the resulting json documents into actual `%Limit{}` structs was to map the results through the `Repo.load/2` function like this:

```elixir 
Repo.all(query)
|> Enum.map(&Repo.load(Limit, &1))
```

However, since the primary key type is `Ecto.UUID`, it won't actually work with `Ecto.UUID.load` since that one expects a binary UUID and not a UUID in text form.

So I propose a new convenience function `embedded_load(schema, data, format)` that uses `Ecto.Type.embedded_load/3` internally so I can fix my example like this:

```elixir 
Repo.all(query)
|> Enum.map(&Repo.embedded_load(Limit, &1, :json))
```

If there's any other way to prevent having to load manually I'd be glad to hear it :)

josevalim:
A PR would be welcome!

narrowtux:
I can try, I'm stuck now anyway :D

josevalim:
If you want to get unstuck, you can return this `select: map(l, :limits)` and then extract them out in Elixir land.

narrowtux:
It was actually pretty easy, in `repo/schema.ex`:


```elixir
  def embedded_load(_adapter, schema_or_types, data, format) do
    do_load(schema_or_types, data, &Ecto.Type.embedded_load(&1, &2, format))
  end
```

PR will follow shorltly

josevalim:
Closing in favor of PR!

PR #3270, opened 2020-04-06
Added Repo.embedded_load/3 function

narrowtux:
as described in #3269

### DIFF
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -602,7 +602,6 @@ defmodule Ecto do
     end
   end
 
-
   @doc """
   Loads previously dumped `data` in the given `format` into a schema.
 
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -625,7 +625,7 @@ defmodule Ecto do
   ## Examples
 
       iex> result = Ecto.Adapters.SQL.query!(MyRepo, "SELECT users.settings FROM users", [])
-      iex> Enum.map(result.rows, fn [settings] -> MyRepo.embedded_load(Setting, Jason.decode!(settings), :json) end)
+      iex> Enum.map(result.rows, fn [settings] -> Ecto.embedded_load(Setting, Jason.decode!(settings), :json) end)
       [%Setting{...}, ...]
   """
   @spec embedded_load(
--- a/lib/ecto.ex
+++ b/lib/ecto.ex
@@ -605,8 +605,8 @@ defmodule Ecto do
   @doc """
   Loads previously dumped `data` in the given `format` into a schema.
 
-  The first argument can be a a schema module, or a map (of types) and determines
-  the return value: a struct or a map, respectively.
+  The first argument can be a an embedded schema module, or a map (of types) and
+  determines the return value: a struct or a map, respectively.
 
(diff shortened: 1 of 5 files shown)


josevalim:
Oh, interesting. Given this function doesn't use the adapter, maybe it doesn't belong on Repo but elsewhere. The question is where? `Ecto.Schema`? `Ecto`?

josevalim:
Btw, really good job. The question is only where to place the code.

narrowtux:
Yeah that's true. I think it belongs in Schema then.

josevalim:
@wojtekmach @michalmuskala any thoughts?

wojtekmach:
> maybe it doesn't belong on Repo but elsewhere. The question is where? Ecto.Schema? Ecto?

Yeah, neither option seems optimal to me. Given embedded_load/3 is implemented and documented almost identically to Repo.load/2, I think I'd keep it on the Repo regardless of the caveat you mentioned (which I totally agree with). I think putting it on `Ecto.Schema` would be fine too but then let's mention both functions in their respective docs imho.

michalmuskala:
Yes, I agree. We'll need at least a lot of cross-linking between them since they are related. I think there's another question and that is loading data that is not from database into the schema. Since, as far as I remember, for embeds load and cast is exactly the same this is generally already possible with `Changeset.cast` followed by `Changeset.apply_changes`. What should be the interaction there? It seems to me this might be a bit confusing as to what function you're supposed to use in what situation.

narrowtux:
> for embeds load and cast is exactly the same this is generally already possible with Changeset.cast followed by Changeset.apply_changes. What should be the interaction there? It seems to me this might be a bit confusing as to what function you're supposed to use in what situation.

One advantage over the changeset calls is that you don't need to repeat the list of fields.

The changeset approach offers more flexibility of course.

josevalim:
>  I think there's another question and that is loading data that is not from database into the schema.

Good call. In those cases we may not have a repo, so putting it on a repo is a no-go. I looked at Ecto.Schema and it only has the schema definition functionality, functions that work on the schema at runtime are all in Ecto, so I propose to add it as `Ecto.embedded_load` and work on cross-linking them in the docs.

Can you please amend it acccordingly @narrowtux ? Thanks everyone!

narrowtux:
Okay

narrowtux:
@josevalim I moved the embedded_load/3 function to the Ecto module, and moved the test to `embedded_test.exs`.

lib/ecto.ex
```diff
@@ -601,4 +601,40 @@ defmodule Ecto do
       true
     end
   end
+
+
```
josevalim, line 605:
```suggestion
```

josevalim:

josevalim:
It feels like Ecto.embedded_dump/3 would really complement this feature, but this is good enough for getting started. Thanks @narrowtux !

narrowtux:
Yeah, an `embedded_dump` makes sense! It probably also works with the `Ecto.Repo.Schema.do_load` function.

narrowtux:
Also, I noticed that this function doesn't load the data into recursive schemas. I'll continue to work on this a bit

PR #3271, opened 2020-04-07
Added Ecto.embedded_dump/2 and fix recursion for deep embeds

narrowtux:
This is a continuation to #3270 - it adds the counterpart to `Ecto.embedded_load/3` and enables loading / dumping recursively for embeds.

I was able to (ab)use `Ecto.Repo.Schema.do_load` to dump, but since that one adds back the `__struct__` key with the schema and the `__meta__` key with the metadata, I removed them after that again, because I don't think dumped data should contain these.

While writing the code for recursive assocs/embeds I stumbled across `do_load` not calling for assocs, only for embeds. 

I think there needs to be a new function that's similar to `do_load` but will also call for assocs (when they're loaded / when the key is present) and use that for both `embedded_dump` and `embedded_load`.

What do you think? Should assocs also be loaded and dumped recursively? 

With that, I think the dump function would have great potential when writing APIs with Phoenix:

```elixir
def index(conn, _) do
  blog_posts = 
    Repo.all(BlogPost, preload: [comments: [:author], :author])
    |> Enum.map(&Ecto.embedded_dump(&1, :json))
    |> Jason.encode!()
  
  send_resp(conn, 200, blog_posts)
end
```

## TODO

 - [ ] Docs for `Ecto.embedded_dump/2`
 - [ ] Maybe recursively dump/load assocs as well

### DIFF
--- a/guides/howtos/Dynamic queries.md
+++ b/guides/howtos/Dynamic queries.md
@@ -185,13 +185,13 @@ end
 
 # 2. Returned dynamic with join binding
 def filter_order_by("published_at_desc"),
-  do: dynamic([p], desc: p.published_at)
+  do: [desc: dynamic([p], p.published_at)]
 
 def filter_order_by("published_at"),
   do: dynamic([p], p.published_at)
 
 def filter_order_by("author_name_desc"),
-  do: dynamic([authors: a], desc: a.name)
+  do: [desc: dynamic([authors: a], a.name)]
 
 def filter_order_by("author_name"),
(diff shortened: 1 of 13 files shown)


josevalim:

josevalim:
@narrowtux i have pushed some fix in master, can you please let me know if it still works as you expect? Thanks!

narrowtux:
Yeah, for my usage in our app, everything still works. Cheers!

PR #3282, opened 2020-04-16
Fix dialyzer warnings

albertored:
This PR fixes most of the dialyzer warnings in ecto, only 5 of them are still present.

I split the PR in more commits so you can decide to omit some of them.

The commit that fixes most of the warnings is [this one](https://github.com/elixir-ecto/ecto/commit/91f967b4aa50d3b8b6fbe321fb549ddfb1e6648d).

I hope it helps the wonderful ecto package and elixir ecosystem!


lib/ecto.ex
```diff
@@ -551,7 +551,7 @@ defmodule Ecto do
   """
   @spec put_meta(Ecto.Schema.schema, meta) :: Ecto.Schema.schema
         when meta: [source: Ecto.Schema.source, prefix: Ecto.Schema.prefix,
-                    context: Ecto.Schema.Metadata.context, state: Ecto.Schema.Metadata.state]
+                    context: any, state: Ecto.Schema.Metadata.state]
```
michalmuskala, line 554: Let's fix this by exporting the `context` type from the `Metadata` module.

lib/ecto/association.ex
```diff
@@ -883,14 +884,9 @@ defmodule Ecto.Association.BelongsTo do
...
-    queryable = Keyword.fetch!(opts, :queryable)
-    related   = Ecto.Association.related_from_query(queryable, name)
-
-    unless is_atom(related) do
-      raise ArgumentError, "association queryable must be a schema, got: #{inspect related}"
-    end
-
+    ref        = if ref = opts[:references], do: ref, else: :id
+    queryable  = Keyword.fetch!(opts, :queryable)
+    related    = Ecto.Association.related_from_query(queryable, name)
```
michalmuskala, line 889: If the lines are going to be changed anyway, I'd prefer to remove the alignment instead of adding more of it

lib/ecto/changeset/relation.ex
```diff
@@ -4,13 +4,15 @@ defmodule Ecto.Changeset.Relation do
...
-               related: atom,
-               field: atom}
+  @type t :: %{required(:cardinality) => :one | :many,
+               required(:on_replace) => :raise | :mark_as_invalid | atom,
+               required(:relationship) => :parent | :child,
+               required(:ordered) => boolean,
+               required(:owner) => atom,
+               required(:related) => atom,
+               required(:field) => atom,
+               optional(:__struct__) => atom(),
```
michalmuskala, line 14: What are the cases where this is optional?

lib/ecto/type.ex
```diff
@@ -1298,10 +1294,6 @@ defmodule Ecto.Type do
     {:ok, acc}
   end
 
-  defp map(_, _, _, _) do
-    :error
-  end
```
michalmuskala, line 1303: I'm not sure removing this clause is correct. I believe it handles a case where we get something that's not a map or keyword from adapter

lib/ecto/type.ex
```diff
@@ -828,7 +828,6 @@ defmodule Ecto.Type do
   defp cast_decimal(term) when is_binary(term) do
     case Decimal.parse(term) do
       {:ok, decimal} -> check_decimal(decimal, false)
-      {decimal, ""} -> check_decimal(decimal, false)
```
michalmuskala, line 831: This is a forward compatibility layer for `Decimal` 2.0

albertored: Oh, I checked only version 1.x, restored

lib/ecto/changeset/relation.ex
albertored: No cases, just a typo, now it's required

josevalim:
Thank you, this is very appreciated! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:

PR #3290, opened 2020-04-30
Add notes that Ecto 3.4 requires Elixir 1.7

joshuataylor:
As the other versions show when the next elixir version is required, it should also be noted Ecto 3.4 requires 1.7, as per https://github.com/elixir-ecto/ecto/commit/e5908c0ed7b988c8c230502d31e278a85f3d2a68#diff-6023be6004fce4718dad3dafb576d258


josevalim:

PR #3293, opened 2020-05-06
Add example of getting only the association as a map

nathanl:
This is something I've wondered how to do and have seen others ask about. I finally figured it out and wanted to document it.


josevalim:

PR #3294, opened 2020-05-08
Fix table names of the join tables in polymorphic assoc guide

zillou:
In this guide, the name of the join tables in the migration examples are using pluralizes convention, `projects_items`, `todo_lists_items`, while in the schema definition examples are using a different names: `project_items`, `todo_list_item`.

This PR changes the table names in the migration examples to match the table names with the schema definition examples.

ref: 

1. https://hexdocs.pm/ecto/polymorphic-associations-with-many-to-many.html#summary


josevalim:

PR #3299, opened 2020-05-15
Allow custom error key in unsafe_validate_unique

qcam:
I think it would help to distinguish the error faster when we already have another validation for the first field.

I can update the docs if this change is accepted.


josevalim:
Sounds great, we just need docs. :)

josevalim:

PR #3301, opened 2020-05-16
Raise on select single atom

### LINKED ISSUE #3297: `select` should accept single atom
Example use:

```elixir
from q in Qwerty, select: :id
```

16:36 < josevalim> both `u.id` or `[:id]` will work
16:36 < josevalim> so the question is if we should support `select: :id` or raise
16:37 < josevalim> given we allow `order_by: :title`, i don't see a reason to now allow `select: :id`
16:37 < josevalim> *not allow

qcam:
Closes #3297

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -24,7 +24,8 @@ defmodule Ecto.Query.Builder.Select do
 
   """
   @spec escape(Macro.t, Keyword.t, Macro.Env.t) :: {Macro.t, {list, %{}}}
-  def escape(atom, _vars, _env) when is_atom(atom) and not is_boolean(atom) do
+  def escape(atom, _vars, _env)
+      when is_atom(atom) and not is_boolean(atom) and atom != nil do
     Builder.error! """
     #{inspect(atom)} is not a valid query expression, :select expects a query expression or a list of fields
     """
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -24,6 +24,12 @@ defmodule Ecto.Query.Builder.Select do
 
   """
   @spec escape(Macro.t, Keyword.t, Macro.Env.t) :: {Macro.t, {list, %{}}}
+  def escape(atom, _vars, _env) when is_atom(atom) and not is_boolean(atom) do
+    Builder.error! """
+    #{inspect(atom)} is not a valid query expression, :select expects a query expression or a list of fields
+    """
+  end
+
   def escape(other, vars, env) do
     if take?(other) do
       {{:{}, [], [:&, [], [0]]}, {[], %{0 => {:any, other}}}}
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder/select.ex
```diff
@@ -140,14 +140,18 @@ defmodule Ecto.Query.Builder.Select do
     end
   end
 
-  defp take?(fields) do
-    is_list(fields) and Enum.all?(fields, fn
-      {k, v} when is_atom(k) -> take?(List.wrap(v))
-      k when is_atom(k) -> true
-      _ -> false
+  defp take?(boolean) when is_boolean(boolean), do: false
```
josevalim, line 143: Interesting... what is the reason to avoid booleans?

qcam: I was thinking of cases like `[:foo, [true: false]]` being passed 🤔.

josevalim: I mean... if they have a field named true, there is nothing we can do. :)

qcam: that's true :D. I'll remove this check then.

qcam: oh but then there exists this: https://github.com/elixir-ecto/ecto/blob/87ce8e3223a1c8a92d40536cf56ac83c0270d711/lib/ecto/changeset.ex#L1857

However, I could not find where `select(true)` is documented 🤔.

qcam: I changed that to `select(1)` like what we did on `Repo.exists?`

josevalim: @qcam ah, but that may be a breaking change... I think we finally understand why we expect it to be a list then. With this new information, I think it may be better to improve the error message after all. WDYT?

qcam: Yeah, maybe someone already depends on this behavior. Shall we document this?

I think currently if we pass `select(:foo)`, the error message is:

```
** (Ecto.Query.CompileError) `:foo` is not a valid query expression
```

I think that's reasonably clear. How do you think we would improve the message otherwise?

josevalim: It would be nice if we could say how to fix it. Maybe we could detect if something is atom and not nil/boolean, and say: "`:foo` is not a valid query expression. `:select` expects a query expression or a list of fields"?

qcam: 👍 I changed the implementation.

lib/ecto/query/builder/select.ex
```diff
@@ -24,6 +24,12 @@ defmodule Ecto.Query.Builder.Select do
 
   """
   @spec escape(Macro.t, Keyword.t, Macro.Env.t) :: {Macro.t, {list, %{}}}
+  def escape(atom, _vars, _env) when is_atom(atom) and not is_boolean(atom) do
```
qcam, line 27: Unfortunately I could not find any better place than putting the check here. Would we prefer to extracting this into a helper function instead?

josevalim: Looks great but we also need to check it is not nil.

qcam: Oops right!

josevalim:

PR #3302, opened 2020-05-18
Update parent_as example

alexfreska:
Hello all, I noticed the new `parent_as` feature was added - really awesome!

I was playing with it via ecto_sql, I could be wrong but I think the example doesn't work / make sense. I think it should be a lateral join?


josevalim:
The wiki is maintained by the community. So if there aren't any up to date instructions, we recommend you to explore the solution yourself and hopefully contribute your findings back!

josevalim:
Oops, wrong message. Here is the correct one: :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart: :D

PR #3305, opened 2020-05-20
Update examples for `Repo.one/2`

mathieuprog:
Examples for `Repo.one/2` are misleading:

```
Repo.one(Post)
```

In the example above, the reader will expect the queryable to return multiple results, but a call to `Repo.one/2` will raise an error if it does.

```
Repo.one(from p in Post, where: p.like_count > 10)
```

In this other example, we are looking for the posts that have over 10 likes. Again, the reader will expect this will likely return multiple results.

The examples lead the reader to think that `Repo.one/2` will actually filter out the first result in case of multiple results.

What I suggest as example:

A query that has a where clause on something unique, like an id. But with a join, otherwise the reader might wonder why not use `get/3`/`get_by/3`.

In the example suggested in this PR, a post is retrieved based on its id only if it contains comments. It's not the best example; I think the `Repo.one/2` function is used for rather specific cases.

Also note that the documentation for `Repo` doesn't contain queries with `join`, `assoc`, ... and that I introduce that with this PR.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -767,11 +767,9 @@ defmodule Ecto.Repo do
 
   ## Examples
 
-      Repo.one(Post)
+      Repo.one(from p in Post, join: c in assoc(p, :comments), where: p.id == ^post_id)
 
-      Repo.one(from p in Post, where: p.like_count > 10)
-
-      query = from p in Post, where: p.like_count > 10
+      query = from p in Post, join: c in assoc(p, :comments), where: p.id == ^post_id
       Repo.one(query, prefix: "private")
   """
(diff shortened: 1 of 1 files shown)


mathieuprog:
I also considered, more simply:
```
Repo.one(from p in Post, where: p.id == ^post_id)
```
better use `get/3` in this case, but up to the user to understand that `one/2` is more suited when need to pass a query.

or (didn't test it), but this kinda completes my example to make more sense out of it, where we want to fetch a single post by id and preload the comments if any.
```
Repo.one(from p in Post, left_join: c in assoc(p, :comments), where: p.id == ^post_id, preload: [comments: c])
```

josevalim:

PR #3306, opened 2020-05-20
Fix documentation where wrong variable is used

cheerfulstoic:
I was confused when reading this documentation, though I might not be understanding what the aim was here.  I see that `mary` and `john` also aren't defined, but I think those are supposed to be defined before this block.


josevalim:

PR #3313, opened 2020-05-24
Proposal: add support for EXPLAIN statement

leandrocp:
During development a very common process is to optimize complex queries by using the `EXPLAIN` sql statement, which as for now requires to play with `to_sql`, copy & paste, manually building the query and calling it on a SQL client. This PR proposes to add `Repo.explain/2` function to ease this process and improve the developer experience.

Of course It's just a WIP to gather feedback of the core team, as it's missing:

- [ ] Maybe improve the function return? current output is just a string like `"Seq Scan on barebones b0  (cost=0.00..30.40 rows=2040 width=4)"`
- [ ] Call EXPLAIN inside a transaction to avoid side effects
- [ ] Support other statements besides SELECT
- [ ] Pass options to the EXPLAIN statement
- [ ] Implement callback for all adapters
- [ ] Improve tests

ecto-sql implementation: https://github.com/elixir-ecto/ecto_sql/pull/231

Please note I'm not an expert in Ecto internals so it may not be the best possible implementation.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -876,6 +876,9 @@ defmodule Ecto.Query.Planner do
       :all ->
         assert_no_update!(query, operation)
         assert_valid_combinations!(query)
+      :explain ->
+        assert_no_update!(query, operation)
+        assert_valid_combinations!(query)
       :update_all ->
         assert_update!(query, operation)
         assert_only_filter_expressions!(query, operation)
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @leandrocp! Thanks so much for the PR. I think however we can implement this directly in Ecto.SQL. We probably don't need to change the planner and instead we can prepare it using :all and then append the EXPLAIN instructions. WDYT?

leandrocp:
The only reason I changed the planner was because I wasn't sure the best way to implement this and it looked right :)
So I'm fine with implementing this on ecto_sql side, maybe similar to `to_sql`?

josevalim:
>  So I'm fine with implementing this on ecto_sql side, maybe similar to to_sql?

Yes, exactly!

josevalim:
I will go ahead and close this but ping me if you feel like this is necessary. :)

PR #3320, opened 2020-06-05
Add Ecto.Query.main_schema/1 function

albertored:
This function returns the main schema of a given Queryable.
All cases are covered but for the bitstring query one, in that case the function returns `nil`.

I'm contributing to a library for building filters on top of ecto queries from some input data structure and I need to obtain the main schema of a given query (I need to call `__schema__` on it). I think this can be useful also for other libraries/projects so I created this PR.

If you think it can be merged I will add some tests.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2019,6 +2019,23 @@ defmodule Ecto.Query do
   end
 
   @doc """
+  Returns the main Ecto.Schema of a query.
+
+  When the input schema is a bitstring `nil` is returned.
+
+  ## Examples
+
+      main_schema(Post) == Post
+
+      User |> join(:inner, [u], p in assoc(u, :posts)) |> main_schema() == User
+
+      main_schema("from u in users") == nil
+  """
+  defp main_schema(%Ecto.Query{from: %{source: {_, module}}}), do: module
+
+  defp main_schema(queryable), do: main_schema(Ecto.Queryable.to_query(queryable))
+
+  @doc """
   Reverses the ordering of the query.
 
   ASC columns become DESC columns (and vice-versa). If the query


josevalim:
Hi @albertored! 

I understand the use case but I am not sure returning just the schema is general enough. What if someone wants the bitstring if the schema is not there? I think returning the whole `from.source` would be better. At the same time, I also wouldn't want to put pollute `Ecto.Query` with such helper functions, because `Ecto.Query` is mostly used by people writing queries.

albertored:
Hi @josevalim! Thanks for your prompt reply!

We can always return a `{source, schema}` tuple (at the moment it is exactly the `query.from.source` value) in order to handle possible future changes of the internal structure of `%Ecto.Query{}`. If I understood correctly the `Ecto.Query` struct is internal and libraries should avoid pattern matching on its internal fields.

I put this function on the `Ecto.Query` because I think it is similar to `Ecto.Query.has_named_binding/2`: it can be used to extract info from a query. Are you suggesting moving it somewhere else or do you think it's not worth adding it?

josevalim:
> Are you suggesting moving it somewhere else or do you think it's not worth adding it?

I am not sure yet. :) Note that the source field can also point to a subquery, so we would need to document those cases too. Are you ok with us developing this feature a bit more even if it is not guaranteed we will merge it? I would like to see if we can define a proper API for it that does not feel too convoluted.

albertored:
Yes sure!

I don't know very well the ecto internals but I opened the PR because sometime 10 lines of code are better than 20 lines of words for explaining what one is suggesting ;-)

josevalim:
Sorry for delay. I have one follow up question. If from is a subquery, what would you do in your app? Would you traverse and fetch the schema inside the subquery?

albertored:
No problem

Yes, for my needs traversing and fetching the schema would be ideal.

I don't know if in general it can be useful to know that it is indeed a subquery (returning for instance a tagged tuple `{:subquery, result}`). For my use case it's not needed.

josevalim:
Yeah, I can see some apps wanting to traverse subqueries. For example, that's what we do in Ecto internals itself. Given it is probably hard to come up with a general purpose approach, I don't think we should add this one for now. Thank you for the discussion! <3

PR #3321, opened 2020-06-06
Test factories.md: Consistent piping

henrik:
Seems more consistent with surrounding code, and perhaps more idiomatic.


josevalim:
I am particularly not a fan of single pipes, so I would prefer to keep things as is. Tks!

henrik:
Interesting, and no worries :) Thanks for looking!

henrik:
Come to think of it, `factory_name |> build(attributes) |> Repo.insert!` would be even more consistent with the preceding function, but obviously not important. If you want me to PR that one, just let me know, otherwise I'll leave it :)

PR #3322, opened 2020-06-06
Test factories.md: Use `struct!` to check for key validity

henrik:
I made a typo in an attribute name and this would have helped catch it.

And it's probably just a good pattern here. If someone wants to support passing in non-attributes, they can use `Keyword.delete` before passing to `struct!`.


josevalim:
Sounds good! Let's also make the other one a complete pipeline as you proposed.

henrik:
Thanks, José! Made another PR.

PR #3325, opened 2020-06-07
Hierarchical relationships example swaps columns

kaidadragonfly:
I was having issues when testing the hierarchical relationship in a Phoenix app I was building.  

It turns out doing `Repo.preload(parent, :parent)` would have the parent load the child, and having `Repo.preload(child, :parent)` would load `nil`.  (Where `parent` has `parent_id` set to `nil` and `child` has `parent_id` set to `parent`'s id.)

Swapping the columns like I did here makes things function correctly.


josevalim:

PR #3337, opened 2020-06-20
Add Ecto.Type.Atom for safely storing atoms

TheFirstAvenger:
`Ecto.Type.Atom` provides a simple way to store atoms using the underlying `:string` ecto type. This is done in a manner that is safe with respect to atom overflow as only specified existing atoms can be loaded.

To define a game schema with status as an atom field:

```elixir
      defmodule MyApp.Games.Game do
        use MyApp.Schema

        # First define the Status atom type:
        defmodule Status do
          use Ecto.Type.Atom, values: [:joining, :running, :ended]
        end

        schema "games" do
          ...
          # Then use the Status module as the field type:
          field :status, Status
          ...
        end
      end
```

A helper function `values/0` is generated on the module, e.g. `Status.values()` with the above module will return `[:joining, :running, :ended]`.

The `else:` option allows custom behaviour in the case that the values being loaded are not in the list of allowed values.

`cast` takes the atom or a string version of the atom. The string version is useful for things like Phoenix requests, and is safe because it validates that the string represents an atom specified in `values` _before_ calling `String.to_existing_atom/1`.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -479,6 +479,7 @@ defmodule Ecto.Schema do
         @after_compile Ecto.Schema
         Module.register_attribute(__MODULE__, :changeset_fields, accumulate: true)
         Module.register_attribute(__MODULE__, :struct_fields, accumulate: true)
+        Module.register_attribute(__MODULE__, :enum_fields, accumulate: true)
 
         meta?  = unquote(meta?)
         source = unquote(source)
@@ -563,6 +564,12 @@ defmodule Ecto.Schema do
         for clauses <- Ecto.Schema.__schema__(fields, field_sources, assocs, embeds),
             {args, body} <- clauses do
           def __schema__(unquote_splicing(args)), do: unquote(body)
+        end
+        
+        for {name, values} <- @enum_fields do
+          defmodule name do
+            use Ecto.Type.Atom, values: values
+          end
         end
       end
 
@@ -1746,6 +1753,12 @@ defmodule Ecto.Schema do
     define_field(mod, name, type, opts)
   end
 
+  defp define_field(mod, name, :enum, opts) do
+    camelized_name = Module.concat(mod, Macro.camelize("#{name}" <> "Enum"))
+    Module.put_attribute(mod, :enum_fields, {camelized_name, opts[:values]})
+    define_field(mod, name, camelized_name, Keyword.drop(opts, [:values]))
+  end
+
   defp define_field(mod, name, type, opts) do
     virtual? = opts[:virtual] || false
     pk? = opts[:primary_key] || false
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @TheFirstAvenger!

My understanding is that this can be completely implemented as a library, so that would be our vote. :)

I would love to have enums in Ecto but I would like to support them directly without a module:

    field :foo, enum([:foo, :bar, :baz])

Or similar. However, our current ecto types are not powerful enough to support this. I have discussed possible ways to solve this with @wojtekmach in the past, perhaps generalizing our embeds code a bit, but we never came up with anything concrete.

wojtekmach:
Yes, the idea was to parameterize types: `{type, params}`, e.g.:

    field :foo, {MyApp.Enum, [:a, :b, :c]}

and ship :enum as part of Ecto:

    field :foo, {:enum, [:a, :b, :c]}

and yes, our litmus test was to be able to implement embeds using this new mechanism which would mean its quite powerful.
On Sat, Jun 20, 2020 at 3:42 PM GitHub <notifications@github.com> wrote:

“Hi @TheFirstAvenger!

My understanding is that this can be completely implemented as a library, so that would be our vote. :)

I would love to have enums in Ecto but I would like to support them directly without a module:

field :foo, enum([:foo, :bar, :baz])

Or similar. However, our current ecto types are not powerful enough to support this. I have discussed possible ways to solve this with @wojtekmach in the past, perhaps generalizing our embeds code a bit, but we never came up with anything concrete.”

TheFirstAvenger:
@josevalim Thanks for the reply! I definitely agree that an enum as you mentioned would be cleaner. As far as a Library, there are already a couple of ecto atom libraries out there, but as external libraries have never caught on. My thoughts are that since atoms are so central to the Elixir language, the ability to store them via Ecto should exist in the core asap. I wonder if this approach could be the core approach until the time that enums as you have discussed are implemented.

josevalim:
@TheFirstAvenger given Ecto doesn't plan on a new major version, I don't want to support something that I know there is a better solution for, because once the better solution lands, we would need to support the non-optimal solution forever. So having it as an external solution suffices for now. :)

TheFirstAvenger:
@josevalim that makes sense, it seemed like how you were explaining it, :enum might be a major version away. If it is coming soon, then that's great!

josevalim:
I mean it is not coming soon, but it also isn't a major version away. :P

TheFirstAvenger:
@josevalim @wojtekmach In thinking about this further over the weekend, I wonder if we might not be as far away from the final format of specifying an `:enum` type as I thought. I implemented a first pass at supporting this format:

```elixir
field "status", :enum, values: [:joining, :running, :ended]
```
which still uses the Ecto.Type.Atom under the hood, but is much closer to the format you were looking for (I prefer `values` as an opt of this format as it is more explicit, but a tuple format could be used too), and eliminates the step of manually creating the custom type beforehand. The custom type still exists in the background, so if you had the above field on `MyApp.Games.Game`, there would be a `MyApp.Games.Game.StatusEnum` helper module with a `values/0` function generated which would be available for other application code to use and keeps the list of possible values DRY.

Keep in mind this was my first dive into the schema `using` macro, so there is probably some cleanup that could be done, but if we are _close_ to `:enum`, I would love to see it put across the finish line.

josevalim:
@TheFirstAvenger my main concern with the implementation was not necessarily the API but having to define a module in the first place. I would like to not define any module whatsoever. The module definition is a limitation that only exists because our types cannot be parameterized.

We could add `{:enum, [....]}` as a built-in Ecto type but my point is exactly that we should support parameterized types in general, as other projects would benefit from this feature set.

TheFirstAvenger:
@josevalim that makes sense. I will iterate on the idea of parameterized types and see what I can come up with.

PR #3341, opened 2020-06-24
use field-defined primary_key for default has_ reference

dallincrane:
Currently, when using `has_many` or `has_one`, the default for the `references:` option is determined entirely by the `@primary_key` value. This doesn't work for primary_key(s) that are defined by the `primary_key:` field option.

Example of when a default `references:` could default to the schema primary_key but isn't.

```
defmodule MyApp.Schema do
  defmacro __using__(_) do
    quote do
      use Ecto.Schema
      @primary_key false
    end
  end
end

defmodule MyApp.Post do
  use MyApp.Schema
  schema "posts" do
    field :id, Ecto.UUID, autogenerate: true, primary_key: true
    has_many :comments, MyApp.Comment
    # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ requires^, references: :id
  end
end

defmodule MyApp.Comment do
  use MyApp.Schema
  schema "posts" do
    field :id, Ecto.UUID, autogenerate: true, primary_key: true
    belongs_to :post, MyApp.Post
  end
end
```

There is also an example in the tests, modified in this PR where the `has_many` `references:` is defaulted and maybe shouldn't be, given the schema has multiple primary keys. Luckily, since it's likely not a very frequent scenario and it's also a compile time check, upgrading Ecto and it "breaking" would be unlikely, a quick find, a quick fix, and have no possibility of effecting released code.

My apologies that I'm not quite sure of the preferred way to add a test for a compile time check to this PR

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -587,15 +587,14 @@ defmodule Ecto.Association.Has do
     cardinality = Keyword.fetch!(opts, :cardinality)
     related = Ecto.Association.related_from_query(queryable, name)
 
-    ref =
-      module
-      |> Module.get_attribute(:primary_key)
-      |> get_ref(opts[:references], name)
+    ref_option = opts[:references]
 
-    unless Module.get_attribute(module, :ecto_fields)[ref] do
-      raise ArgumentError, "schema does not have the field #{inspect ref} used by " <>
+    if ref_option && Module.get_attribute(module, :ecto_fields)[ref_option] == nil do
+      raise ArgumentError, "schema does not have the field #{inspect ref_option} used by " <>
         "association #{inspect name}, please set the :references option accordingly"
     end
+
+    ref = ref_option || get_ref(module, name)
 
     if opts[:through] do
       raise ArgumentError, "invalid association #{inspect name}. When using the :through " <>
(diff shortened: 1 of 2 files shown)


lib/ecto/association.ex
```diff
@@ -587,16 +587,15 @@ defmodule Ecto.Association.Has do
...
 
-    ref =
-      module
-      |> Module.get_attribute(:primary_key)
-      |> get_ref(opts[:references], name)
+    ref_option = opts[:references]
 
-    unless Module.get_attribute(module, :ecto_fields)[ref] do
-      raise ArgumentError, "schema does not have the field #{inspect ref} used by " <>
+    if ref_option && Module.get_attribute(module, :ecto_fields)[ref_option] == nil do
```
dallincrane, line 592: Moved this check up before the call to get the module's primary_key since what is returned by `Module.get_attribute(module, :ecto_primary_keys)` will for sure be in `:ecto_fields`

test/ecto/changeset_test.exs
```diff
@@ -50,8 +50,8 @@ defmodule Ecto.ChangesetTest do
       field :source, :map
       field :permalink, :string, source: :url
       belongs_to :category, Ecto.ChangesetTest.Category, source: :cat_id
-      has_many :comments, Ecto.ChangesetTest.Comment, on_replace: :delete
-      has_one :comment, Ecto.ChangesetTest.Comment
+      has_many :comments, Ecto.ChangesetTest.Comment, on_replace: :delete, references: :id
```
dallincrane, line 53: This is the scenario mentioned in the PR description where one of multiple primary keys is a assumed and maybe shouldn't be.

The two primary keys
1) hidden via the default `@primary_key {:id, :id, autogenerate: true}`
2) defined by the field call `field :token, :integer, primary_key: true`

Current implementation only looks at `@primary_key` and ignores the field-defined primary key

josevalim:
Hi @dallincrane, thanks for the PR! :heart:

This difference is intentional as the approach you propose has some issues. For example, it is ordering dependent. If the primary key fields are defined after, then it won't affect relationships. Another issue, as you noted, is composite keys. Therefore I have pushed a commit that makes this explicit in the docs. If there are other places in the docs you would rather change, a PR in the area is also welcome.

PR #3347, opened 2020-06-26
Add Ecto.Changeset.put_change_if_exists/4 function

Jmeggesto:
This function puts a change `new_key` only if a value exists in the changeset for an existing `key`.

I have noticed that `Ecto.Changeset.get_change/3` is used very frequently in codebases to test for the presence of an existing change in a changeset, and if the change is present, to put some _other_ change, often conditional on the value of the existing change.

In practice this looks like:
```elixir
case get_change(changeset, :key) do
    nil -> changeset
    value -> put_change(changeset, :new_key, new_value)
end
```

A convenient function to put a new change if an existing one is present would be a small but highly useful addition.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -601,12 +601,12 @@ defmodule Ecto.Changeset do
       |> Enum.reduce(nil, fn
         {key, _value}, nil when is_binary(key) ->
           nil
-  
+
         {key, _value}, _ when is_binary(key) ->
           raise Ecto.CastError, type: :map, value: params,
                                 message: "expected params to be a map with atoms or string keys, " <>
                                          "got a map with mixed keys: #{inspect params}"
-  
+
         {key, value}, nil when is_atom(key) ->
           [{Atom.to_string(key), value}]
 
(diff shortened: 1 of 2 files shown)


wojtekmach:
We already have [`update_change/3`](https://hexdocs.pm/ecto/Ecto.Changeset.html#update_change/3) that works very similarly so I don't believe we need to add another function. Thank you for the PR nonetheless!

Jmeggesto:
@wojtekmach `update_change/3` only updates the existing value for the given key -- there's no easy facility for putting a change for `key1` _only_ if there exists a change for a different key `key2`. Oftentimes it's necessary to have the existence of one change be dependent on the presence of a different change, which is what this PR addresses. 

With `update_change/3`, to achieve similar behaviour you'd need to do:
```elixir
update_change(changeset, existing_key, fn existing_value ->
  put_change(changeset, new_key, new_value)
  existing_value
end)
```

josevalim:
@Jmeggesto that's a good point. Although I would say that I would prefer to write the case you originally mentioned. That's because adding a new function increases the API and the bigger the API the less likely folks will use auxiliary functions such as this one. If you find yourself doing that a lot, then you create a helper in your app, but I think the odds that other developers besides the three of us will use this function is most likely low. :) In any case, thanks for the PR! :heart:

PR #3348, opened 2020-06-27
Maybe unbreak MySQL 8 specs failing due to implicit order

henrik:
I saw test failures in ecto_sql's CI such as https://github.com/elixir-ecto/ecto_sql/runs/814700353 that seem to be due to tests assuming that records are returned in a given order, but the query does not specify any order as far as I can tell.

I believe this should fix those tests, but I didn't want to go through all the hoops to get the MySQL integration tests running locally, so I haven't fully verified. Perhaps someone with everything set up could do so.


josevalim:

PR #3352, opened 2020-07-02
Accept fun on multi functions

thiamsantos:
The functions `insert_all/5`, `update_all/5` and `delete_all/4` do not accept a fun as argument. So we are not able to build the params from the previous changes of the multi. These changes were proposed at the mailing list https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/elixir-ecto/cr7IREEpdgU/nbXN_Gn2CAAJ.

With the changes in the PR the following combinations are possible:

```elixir
Ecto.Multi.insert_all(multi, name, schema_or_source, entries)
Ecto.Multi.insert_all(multi, name, schema_or_source, fn changes -> entries end)
Ecto.Multi.insert_all(multi, name, fn changes -> schema_or_source end, entries)
Ecto.Multi.insert_all(multi, name, fn changes -> schema_or_source end, fn changes -> entries end)

Ecto.Multi.delete_all(multi, name, queryable)
Ecto.Multi.delete_all(multi, name, fn changes -> queryable end)

Ecto.Multi.update_all(multi, name, queryable, updates)
Ecto.Multi.update_all(multi, name, fn changes -> queryable end, updates)
Ecto.Multi.update_all(multi, name, queryable, fn changes -> updates end)
Ecto.Multi.update_all(multi, name, fn changes -> queryable end, fn changes -> updates end)
```

With these multi functions become more flexible, and in more parity with the other multi function like `insert/3` or `delete/3`.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -469,6 +469,23 @@ defmodule Ecto.Multi do
       |> Ecto.Multi.insert_all(:insert_all, Post, posts)
       |> MyApp.Repo.transaction()
 
+      Ecto.Multi.new()
+      |> Ecto.Multi.run(:post, fn repo, _changes ->
+        case repo.get(Post, 1) do
+          nil -> {:error, :not_found}
+          post -> {:ok, post}
+        end
+      end)
+      |> Ecto.Multi.insert_all(:insert_all, Comment, fn %{post: post} ->
+        # Others validations
+
+        entries
+        |> Enum.map(fn comment ->
+          Map.put(comment, :post_id, post.id)
+        end)
+      end, [])
+      |> MyApp.Repo.transaction()
+
   """
   @spec insert_all(t, name, schema_or_source | fun(schema_or_source), [map | Keyword.t] | fun([map | Keyword.t]), Keyword.t) :: t
   def insert_all(multi, name, schema_or_source_or_fun, entries_or_fun, opts \\ [])
@@ -490,6 +507,19 @@ defmodule Ecto.Multi do
 
       Ecto.Multi.new()
       |> Ecto.Multi.update_all(:update_all, Post, set: [title: "New title"])
+      |> MyApp.Repo.transaction()
+
+      Ecto.Multi.new()
+      |> Ecto.Multi.run(:post, fn repo, _changes ->
+        case repo.get(Post, 1) do
+          nil -> {:error, :not_found}
+          post -> {:ok, post}
+        end
+      end)
+      |> Ecto.Multi.update_all(:update_all, fn %{post: post} ->
+        # Others validations
+        from(c in Comment, where: c.post_id == ^post.id, update: [set: [title: "New title"]])
+      end, [])
       |> MyApp.Repo.transaction()
 
   """
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -470,8 +470,8 @@ defmodule Ecto.Multi do
       |> MyApp.Repo.transaction()
 
   """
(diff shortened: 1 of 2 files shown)


lib/ecto/multi.ex
```diff
@@ -505,9 +505,28 @@ defmodule Ecto.Multi do
...
+        end
+      end)
+      |> Ecto.Multi.delete_all(:delete_all, fn %{post: post} ->
+        # Others validations
+        from(c in Comment, where: c.post_id == ^post.id)
+      end)
+      |> MyApp.Repo.transaction()
+
   """
   @spec delete_all(t, name, Ecto.Queryable.t, Keyword.t) :: t
```
sam-levy, line 522: 💅 
```suggestion
  @spec delete_all(t, name, Ecto.Queryable.t | fun(Ecto.Queryable.t), Keyword.t) :: t
```

josevalim:
Hi @thiamsantos! This looks like a positive direction but I believe we are adding too many functions. I don't believe we need to make the `schema_or_source` a function nor the updates (especially because in recent Ecto you can include the update as part of the queryable. Can you please simplify the changes accordingly? Thank you!

thiamsantos:
@josevalim Thanks for the review :)

I removed the function for `schema_or_source` on `insert_all/5` and for `updates` on `update_all/5`. WDYT?

test/ecto/multi_test.exs
```diff
@@ -196,6 +196,18 @@ defmodule Ecto.MultiTest do
     assert [{:comments, {:insert_all, Comment, [[x: 2]], []}}] = multi.operations
   end
 
+  test "insert_all fun" do
+    fun_schema = fn _changes -> {:ok, Comment} end
+    fun_entries = fn _changes -> {:ok, [[x: 2]]} end
```
josevalim, line 201: This is definitely no longer a function but the tests are passing. This probably means we are missing some guards in the code.

test/ecto/multi_test.exs
```diff
@@ -196,6 +196,18 @@ defmodule Ecto.MultiTest do
     assert [{:comments, {:insert_all, Comment, [[x: 2]], []}}] = multi.operations
   end
 
+  test "insert_all fun" do
+    fun_schema = fn _changes -> {:ok, Comment} end
```
josevalim, line 200: This should return only `Comment`. Same thing in the other tests. :)

josevalim, line 200: This is no longer a function. :)

test/ecto/multi_test.exs
josevalim, line 201: It should return just a list, no? I wonder if we can have a more precise test, because this one and the ones below are not catching much. :)

thiamsantos: > It should return just a list, no? 

I think it is correct. We have a list of entries, each entry can be a keyword list as `[x: 2]`

> I wonder if we can have a more precise test, because this one and the ones below are not catching much. :)

Yes, the scenario is clearly wrong, but the tests are still passing. WDYT of running the multi with `TestRepo.transaction` to have more cover of the functionality?

josevalim: > Yes, the scenario is clearly wrong, but the tests are still passing. WDYT of running the multi with TestRepo.transaction to have more cover of the functionality?

Sounds good!

thiamsantos: > It should return just a list, no?

I had misunderstood. You were right. Most of the examples in the added tests had this same problem.

> Sounds good!

I just updated the tests to run the multi. Could you take another look?

josevalim:

josevalim:

PR #3353, opened 2020-07-02
don't validate uniqueness if there is a prior error

slashmili:
There is no need to validate uniqueness when there is an error on the expected unique fields

Example:

    user
    |> cast(attrs, [:name, :email])
    |> validate_format(:email, ~r/@/)
    |> unsafe_validate_unique([:email], MyApp.Repo)

if email doesn't contain at sign, we can save few database query until the email is in the correct format.

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1391,6 +1391,29 @@ defmodule Ecto.ChangesetTest do
       assert changeset.valid?
     end
 
+    test "does not validate uniqueness if there is any prior error on a field", context do
+      Process.put(:test_repo_all_results, context.dup_result)
+      changeset =
+        context.base_changeset
+        |> validate_length(:title, max: 3)
+        |> unsafe_validate_unique(:title, TestRepo)
+
+      refute changeset.valid?
+      assert changeset.errors == [title: {"should be at most %{count} character(s)", [count: 3, validation: :length, kind: :max, type: :string]}]
+    end
+
+    test "does not validate uniqueness if there is any prior error on a combination of fields", context do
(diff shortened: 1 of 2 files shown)


test/ecto/changeset_test.exs
```diff
@@ -1391,6 +1391,29 @@ defmodule Ecto.ChangesetTest do
...
+      changeset =
+        context.base_changeset
+        |> validate_length(:title, max: 3)
+        |> unsafe_validate_unique([:title, :body], TestRepo)
+
+      refute changeset.valid?
+      assert changeset.errors == [title: {"should be at most %{count} character(s)", [count: 3, validation: :length, kind: :max, type: :string]}]
+    end
+
+
```
josevalim, line 1416:
```suggestion
```

josevalim:

nathanl:
I thought I had implemented this, but I remembered wrong - I was thinking of https://github.com/elixir-ecto/ecto/pull/2905

Nice catch and very nice PR! 👏 😄 💯

michalmuskala:
I'm not convinced this is a correct change. In general validators try to add as many errors as possible at one go - e.g. we run subsequent validations even in a previous one failed. 

This has significant impact on the final UI presented to the user. Imagine a registration situation. The user will either get information that the password they chose is not long enough and the username is already taken at the same time, or they have to submit the form twice. 

This change will lead to the poorer user experience of having to submit the form multiple times and getting a trickle of errors one-by-one. This issue of one-by-one errors is the main advantage of `unsafe_validate_unique` over constraints. With this change I don't really see a reason to use this function over constraints.

PR #3354, opened 2020-07-07
Documents `:source` option for embeds_many/one

tiagonbotelho:
`:source` is [an accepted option](https://github.com/elixir-ecto/ecto/blob/v3.4.5/lib/ecto/schema.ex#L1864) to use in embeds but it wasn't
present in the documentation.

This commit makes it clearer that these options are supported.


josevalim:

Issue #3017, opened 2019-06-08
embeds_one requires id field on cast_embed


Hermanverschooten:
### Environment

* Elixir version (elixir -v): 1.8.1
* Database and version: PostgreSQL 11.3
* Ecto version (mix deps): 3.1.4
* Database adapter and version (mix deps): postgrex 0.14.3
* Operating system: OSX 10.14.5

### Current behavior
I have this schema:

```
schema "suppliers" do
  field :name, string
  embeds_one :info, Info, on_replace: :delete do
    field :account, :string
  end
end

def changeset(supplier, attrs) do
  supplier
  |> cast(attrs, [:name])
  |> cast_embed(:info, with: &info_changeset/2)
end

def info_changeset(schema, attrs) do
  schema
  |> cast(attrs, [:account])
end
attrs = %{ "name" => "Herman", "info" => %{ "account" => "1234" }}

%Supplier
|> Suppliers.Supplier.changeset(attrs)
|> Repo.insert()
```
fails with
```
** (Ecto.NoPrimaryKeyValueError) struct `%Suppliers.Supplier.Info{account: nil, id: nil}` is missing primary key value
```

### Expected behavior

It should insert the record, not complain about a missing id.

if I  add an `id` field to the embedded `info`, it works, uses a generated `id` though.
Or if I force the primary_key to false in the `embeds_one`'s options then it works.

I see 2 issues with this:

1.  Why should I need to add an `id` field to my `info` to make it work out of the box. I do not need to do this with `embeds_many`
2. Why is there an `id` in `embeds_one` in the first place, there can be only one.

josevalim:
Can you provide the complete stacktrace? Is it raising in the insert line?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

Hermanverschooten:
```
[error] #PID<0.1069.0> running OvlInvWeb.Endpoint (connection #PID<0.1068.0>, stream id 1) terminated
Server: localhost:4000 (http)
Request: PUT /api/suppliers/12
** (exit) an exception was raised:
    ** (Ecto.NoPrimaryKeyValueError) struct `%OvlInv.Suppliers.Supplier.Info{account: nil, client_ref: nil, id: nil}` is missing primary key value
        (ecto) lib/ecto/embedded.ex:156: anonymous fn/3 in Ecto.Embedded.autogenerate_id/5
[...]

PR #3358, opened 2020-07-08
Nested embeds are missing autogenerated ids

tomconroy:
This is really an issue, I've added a failing test here demonstrating, but I haven't tried to fix in ecto.

It looks related to https://github.com/elixir-ecto/ecto/issues/3017 but I'm not sure if it's the same issue exactly.

FWIW: I worked around the issue by making my code deal only with changesets rather than structs like this, but this was surprising behavior from ecto.

### Current behavior

When using `put_change` (rather than casting) to insert an `embeds_one` inside another `embeds_one`, the `id` is not automatically generated on the inner embed (it is correctly generated on the outer one). That creates issues discussed in #3017, I had to fix production data with a migration since it was missing the ids.

Inspecting the output from `TestRepo.insert` looks like this:

```elixir
{:ok,
 %Ecto.Changeset.EmbeddedTest.Author{
   __meta__: #Ecto.Schema.Metadata<:loaded, "authors">,
   id: 1,
   inline_posts: [],
   inline_profile: nil,
   invalid_posts: [],
   invalid_profile: nil,
   name: nil,
   nested_example: %Ecto.Changeset.EmbeddedTest.NestedExample{
     id: "094b4869-cab2-41a9-8083-19fb7c4e5ef7",
     profile: %Ecto.Changeset.EmbeddedTest.Profile{id: nil, name: "tom"}
   },
   posts: [],
   profile: nil,
   raise_posts: [],
   raise_profile: nil,
   update_profile: nil
 }}
```

### Expected behavior

Expect the id to be autogenerated in this case

### DIFF
--- a/test/ecto/changeset/embedded_test.exs
+++ b/test/ecto/changeset/embedded_test.exs
@@ -9,6 +9,7 @@ defmodule Ecto.Changeset.EmbeddedTest do
   alias __MODULE__.Author
   alias __MODULE__.Profile
   alias __MODULE__.Post
+  alias __MODULE__.NestedExample
 
   defmodule Author do
     use Ecto.Schema
@@ -22,6 +23,7 @@ defmodule Ecto.Changeset.EmbeddedTest do
       embeds_one :inline_profile, Profile do
         field :name, :string
       end
+      embeds_one :nested_example, NestedExample
       embeds_many :posts, Post, on_replace: :delete
       embeds_many :raise_posts, Post, on_replace: :raise
       embeds_many :invalid_posts, Post, on_replace: :mark_as_invalid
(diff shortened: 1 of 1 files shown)


josevalim:
Fixed in #3360. But we will probably wait a while before merging it as to not step on the toes of @TheFirstAvenger work in #3339.

Btw @TheFirstAvenger, while working on this patch, I just remembered how different embeds and associations are, so please don't try to include associations on your work in #3339. :)

josevalim:
Btw @tomconroy , thanks for the small test case, it makes my life much easier. :)

TheFirstAvenger:
@josevalim Noted. I have added this test to my branch locally so it should be fixed in my work.

josevalim:
@TheFirstAvenger Sweet. Note I just pushed another commit to fix the integration tests. In order to run them, you will have to clone ecto_sql and run a command like this:

    ECTO_ADAPTER=pg ECTO_PATH=../ecto mix test

Issue #3361, opened 2020-07-17
id in ^subquery does not work


AndrewDryga:
### Environment

* Elixir version (elixir -v): 1.10.2
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PG 11
* Ecto version (mix deps): 3.4.5
* Database adapter and version (mix deps): 3.4.5
* Operating system: Mac OS

### Current behavior

Fro the documentation I expected that Ecto is able to do `id in ^subquery` kind of queries but this does not seem to work:

```
** (Ecto.Query.CastError) query.ex:113: value `#Ecto.Query<from p0 in Domain.Mes
saging.Conversation.Topic.Participant, as: :participants, where: p0.assignment_id == ^"8876497c-4bf7-463b-b5a9-db3a250592dc", select: p0.
topic_id>` in `where` cannot be cast to type {:in, :binary_id} in query:

     from c0 in Domain.Messaging.Conversation,
       as: :conversation,
       left_join: t1 in Domain.Messaging.Conversation.Topic,
       on: t1.id == c0.topic_id,
       join: m2 in Domain.Messaging.Message,
       on: m2.id == c0.last_message_id,
       where: c0.type != "topic" or t1.is_public == true or t1.id in ^#Ecto.Query<from p0 in Domain.Messaging.Conversation.Topic
.Participant, as: :participants, where: p0.assignment_id == ^"8876497c-4bf7-463b-b5a9-db3a250592dc", select: p0.topic_id>,
       order_by: [desc: coalesce(m2.sent_at, m2.inserted_at)],
       order_by: [desc: c0.id],
       limit: ^51,
       select: c0,
       preload: [topic: t1, last_message: m2]

     code: assert {:ok, [], _cursor} = list_conversations_for_account(account, subject)
     stacktrace:
       (elixir 1.10.2) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
       (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir 1.10.2) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir 1.10.2) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
       (ecto 3.4.5) lib/ecto/repo/queryable.ex:176: Ecto.Repo.Queryable.execute/4
[...]

PR #3362, opened 2020-07-18
Improve exception message for Ecto.Query as right-side value of in operator

Eiji7:
For consistency I did not want to break into cases this part:
https://github.com/elixir-ecto/ecto/blob/f5c2de8b30a33138bcedf09356aedc0908fb35b7/lib/ecto/exceptions.ex#L31-L35

Instead I changed code to pass optional `hint` message which is added at end of `exception` message:
>`%Ecto.Query{}` struct is not supported as right-side value of `in` operator.
Did you mean to use `subquery(query)` instead?

cc @josevalim and @AndrewDryga 
Closes: #3361

### DIFF
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -48,7 +48,7 @@ defmodule Ecto.QueryError do
 
     message =
       case hint do
-        {:ok, text} -> message <> "\n\n" <> text
+        {:ok, text} -> message <> "\n" <> text
         _ -> message
       end
 
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -28,6 +28,7 @@ defmodule Ecto.QueryError do
   def exception(opts) do
     message = Keyword.fetch!(opts, :message)
     query   = Keyword.fetch!(opts, :query)
+    hint    = Keyword.fetch(opts, :hint)
     message = """
     #{message} in query:
 
@@ -43,6 +44,12 @@ defmodule Ecto.QueryError do
         Exception.format_file_line(relative, line) <> " " <> message
       else
         message
+      end
+
+    message =
+      case hint do
+        {:ok, text} -> message <> "\n\n" <> text
+        _ -> message
       end
 
     %__MODULE__{message: message}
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks @Eiji7! This is definitely in the right direction and I believe there is an even simpler solution. You can add two new clauses around here:

https://github.com/elixir-ecto/ecto/blob/903c00479cfeaadbcd9d09511fa25cbdeb626df9/lib/ecto/query/planner.ex#L730

Something like:

```
  defp cast_param(_kind, query, expr, %Query{}, {:in, _}, _value) do
    error! query, expr, "..."
  end
```

You mentioned consistency but I think this is fine and consistent with the other error handling we have right now. :)

Eiji7:
@josevalim Ah, what a mistake. I was focused on the matched clause and completely ignored rest.

I have combined this with `hint` option and got such example error:
```elixir
  1) test plan: Ecto.Query struct as right-side value of in operator (Ecto.Query.PlannerTest)
     test/ecto/query/planner_test.exs:166
     ** (Ecto.QueryError) test/ecto/query/planner_test.exs:170: `%Ecto.Query{}` struct is not supported as right-side value of `in` operator on in query:
     
     from p0 in Ecto.Query.PlannerTest.Post,
       prefix: "my_prefix",
       where: p0.id in ^#Ecto.Query<from p0 in Ecto.Query.PlannerTest.Post, prefix: "my_prefix">
     
     Did you mean to use `subquery(query)` instead?
     
     code: plan(Post |> where([p], p.id in ^query))
     stacktrace:
       (elixir 1.10.4) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
       (elixir 1.10.4) lib/enum.ex:1520: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir 1.10.4) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
       test/ecto/query/planner_test.exs:170: (test)
```

I really like how it looks now, but I have one question …

I made this error like others, so it ends with " on" suffix, but later `Ecto.QueryError` is adding: `in query` after it which gives: `(…) on in query:(…)`. I'm not enough good in English to say what's wrong and what's not, but it looks weird for me. Should I change those error messages or we are fine with it?

Also pattern-matched clauses raises `Ecto.QueryError` instead of `Ecto.CastError` which is a bit inconsistent with behaviour of last clause (and fact that it raises inside `cast_param` i.e. casting). Is it expected?

josevalim:

Issue #3238, opened 2020-02-26
Can't preload embedded_schema struct


ZhengQingchen:
### Environment
* Elixir version (elixir -v):
Elixir 1.9.4 (compiled with Erlang/OTP 22)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):
PostgreSQL 10.6
* Ecto version (mix deps):
  ecto_sql 3.3.3 (Hex package) (mix)
  locked at 3.3.3 (ecto_sql) 3d5b8e14
* Database adapter and version (mix deps):
  postgrex 0.15.3 (Hex package) (mix)
  locked at 0.15.3 (postgrex) 4737ce62
* Operating system:
MacOS
### Current behavior
There is an embedded_schema which has a belongs_to relationship with other database schema. When I try to preload the embedded_schema. It's got a match error

```elixir
iex(102)> Repo.preload completion_cache, :completed_by
** (MatchError) no match of right hand side value: CompletionCache{completed_at: nil, completed_by: #Ecto.Association.NotLoaded<association :completed_by is not loaded>, completed_by_uuid: "30d8ab0d-0cc5-4130-ad18-bc5c944bd03c"}
```

The preload schema is.
```
defmodule CompletionCache do
  @derive Jason.Encoder
  use Airbase.Schema, primary_key: false

  embedded_schema do
    field :completed_at, :naive_datetime
    belongs_to :completed_by, User, foreign_key: :completed_by_uuid, references: :uuid 
  end
end
```
The user schema just a normal database schema. 
It's cause by this patch match

https://github.com/elixir-ecto/ecto/blob/e25cb0c05ed6c8923b8c4186fdf285322e67af78/lib/ecto/repo/preloader.ex#L92

So Can I use the `belongs_to` in an `embedded_schema`? 
Or could I just ignore the prefix when we do preload the embedded_schema struct like this?
```
  defp preload_prefix(opts, sample) do
    case Keyword.fetch(opts, :prefix) do
      {:ok, prefix} ->
        prefix
      :error ->
        case sample do
          %{__meta__: %{prefix: prefix}} -> prefix
          _ -> nil
        end
    end
  end
```

josevalim:
Embedded schemas are always loaded when you load entries. You don't have to preload them as they are not in separate tables.

erlandsona:
@josevalim I'm trying to do the same thing and running into a different error, but what do you mean by...
> Embedded schemas are always loaded when you load entries

[...]

Issue #1837, opened 2016-12-07
Support associations inside embeds


michalmuskala:
When embeds were introduced it was never considered that an association could be inside an embed. Because of this, it silently fails.

There are couple places where changes are needed:
* casting - as attempted by https://github.com/elixir-ecto/ecto/pull/1833
* persisting - amending the `surface_changes` function in https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L491-L528
* possibly other, that I haven't considered.

While it makes little sense to use associations inside embeds in SQL databases, the main use case would be in various NoSQL ones, where this seems like a completely reasonable thing to do.

jfayad:
I would also add that embeds should also work inside an embed.

I think this type of recursion shouldn't be limited by Ecto because each design is not right or wrong it always depends on how it will be used *at max show a warning if the recursion is really going deep*

michalmuskala:
Embeds inside embeds should work correctly. This is a use case that is extensively tested.

merqlove:
Looks likes it worked in version 1.0.0, because i'm have this pattern in my app & its tests. After upgrade to 2.0 this pattern is failed.
Example app which uses same pattern & fails on PG:
https://github.com/merqlove/some_ecto/blob/master/test/some_ecto_test.exs

merqlove:
Error is because embed structs don't have `__meta__`.
It was in 1.0:
https://github.com/elixir-ecto/ecto/blob/v1.0.0/lib/ecto/schema.ex#L319
But not in 2.0:
https://github.com/elixir-ecto/ecto/blob/v2.0/lib/ecto/schema.ex#L337

merqlove:
Aha, i found workaround for embeds :)
assert Repo.preload(pts, :party, prefix: nil)

josevalim:
I am closing this for now. If someone is interested on this feature, a pull request is welcome!

linjunpop:
+1 for associations inside embeds.

PR #3368, opened 2020-08-04
Support preloading associations in embedded schemas

sebastianseilund:
Embedded schema structs do not have a `__meta__` key, so the old code failed with:

```
no match of right hand side value: %MyApp.EmbeddedThing{...
```

If the struct(s) given to preload do not have `__meta__`, it seems safe to use `nil` as `prefix`.

----

Contrived use case: `Post` _embeds_ many `Comment`s. `Comment` (an embedded schema) belongs to an author `User`.

```elixir
defmodule User do
  schema "users" do
    field :name, :string
  end
end

defmodule Post do
  schema "posts" do
    embeds_many :comments, Comment
  end
end

defmodule Comment do
  embedded_schema "posts" do
    field :message, :string
    belongs_to :author, User
  end
end

# If I have a `Comment` struct, I'd like to be able to preload its `author
comment = Repo.get(Comment, comment_id) |> Repo.preload(:author)
```

----

Related issues: #3238, #1837

### DIFF
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -6,6 +6,7 @@ defmodule Ecto.Integration.PreloadTest do
 
   alias Ecto.Integration.Post
   alias Ecto.Integration.Comment
+  alias Ecto.Integration.Item
   alias Ecto.Integration.Permalink
   alias Ecto.Integration.User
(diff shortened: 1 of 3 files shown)


josevalim:

Issue #3237, opened 2020-02-22
Ecto.Schema and typespec issues in Elixir 1.10 


davelively14:
### Environment

* Elixir version (elixir -v): 1.10.1
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): N/A, but Postgres 12.2
* Ecto version (mix deps): 3.3.3
* Database adapter and version (mix deps): N/A, but ecto_sql, 3.3.4
* Operating system: Mac OS Mojave

### Current behavior

I think this is an Ecto issue with the newest minor version of Elixir, but in Elixir 1.10.1, I can't use the schema types in my typespecs. This is not an issue on previous versions of Elixir.

Using this code (note the spec on what would be line 12): 

```elixir
defmodule MyApp.User do
  use Ecto.Schema
  import Ecto.Changeset
  alias __MODULE__

  schema "users" do
    field :email, :string
    field :password, :string
    timestamps()
  end

  @spec changeset(User.t(), map()) :: Ecto.Changeset.t()
  def changeset(%User{} = user \\ %User{}, attrs) do
    user
    |> cast(attrs, [:email, :password])
  end
end
```

Paste in an iex session with Elixir 1.10.1 (erlang 22.2.6, Ecto 3.3.3), and you'll get this error:
```
** (CompileError) iex:12: type t/0 undefined (no such type in MyApp.User)
    (elixir 1.10.1) lib/kernel/typespec.ex:898: Kernel.Typespec.compile_error/2
    (stdlib 3.11.2) lists.erl:1354: :lists.mapfoldl/3
    (elixir 1.10.1) lib/kernel/typespec.ex:950: Kernel.Typespec.fn_args/5
    (elixir 1.10.1) lib/kernel/typespec.ex:936: Kernel.Typespec.fn_args/6
    (elixir 1.10.1) lib/kernel/typespec.ex:377: Kernel.Typespec.translate_spec/8
    (stdlib 3.11.2) lists.erl:1354: :lists.mapfoldl/3
    (elixir 1.10.1) lib/kernel/typespec.ex:229: Kernel.Typespec.translate_typespecs_for_module/2
    (elixir 1.10.1) src/elixir_erl_compiler.erl:12: anonymous fn/3 in :elixir_erl_compiler.spawn/2
```

Paste that same code in an iex session with Elixir 1.9.1 (erlang 22.1.3, Ecto 3.1.7), and it compiles:
```
{:module, MyApp.User,
[...]

PR #3369, opened 2020-08-06
Document typespecs in Schema

sabiwara:
Hi :wave: 

I found myself facing the same issue as described here: https://github.com/elixir-ecto/ecto/issues/3237#issuecomment-589992045. The comment answered my question, but I thought it could be nice to mention it directly in the documentation to increase discoverability, since it is an honest question and that typespecs are part of the language? WDYT?

Screenshot from 2020-08-04 22-20-05


josevalim:

PR #3370, opened 2020-08-07
Add ParameterizedType

TheFirstAvenger:
This is v1 of `ParameterizedType` as outlined in [this comment](https://github.com/elixir-ecto/ecto/pull/3339#issuecomment-670081678) on a previous iteration of this work. `ParameterizedType` is added, and `Ecto.Type.Enum` is added in `test` as the validation implementation of ParameterizedType (to be promoted to `lib` in a follow up PR).

@josevalim I modified the Embedded changes in the previous commit slightly, as all of the ParameterizedType callbacks have the options/config as the final argument, not the first (making them read left-to-right closer to the Ecto.Type callbacks).

### DIFF
--- a/lib/ecto/parameterized_type.ex
+++ b/lib/ecto/parameterized_type.ex
@@ -1,0 +1,99 @@
+defmodule Ecto.ParameterizedType do
+  @moduledoc """
+  Parameterized types are Ecto types that can be customized per field.
+
+  Paramterized types allow a set of options to be specified in the schema
+  which are initialized on compilation and passed to the callback functions
+  as the last argument. 
+
+  For example, `field :foo, :string` behaves the same for every field.
+  On the other hand, `field :foo, Ecto.Enum, values: [:foo, :bar, :baz]`
+  will likely have a different set of values per field.
+
+  Note that options are specified as a keyword, but it is idiomatic to
+  convert them to maps inside `c:init/1` for easier pattern matching in
+  other callbacks.
+
+  Parameterized types are a superset of regualr types. In other words,
+  with parameterized types you can do everything a regular type does,
+  and more. For example, parameterized types can handle `nil` values
+  in both `load` and `dump` callbacks, they can customize `cast` behavior
+  per query and per changeset, and also control how values are embedded.
+  
+  However, parameterized types are also more complex. Therefore, if
+  everything you need to achieve can be done with basic types, they
+  should be preferred to parameterized ones.
+
+  ## Examples
+
+  To create a parameterized type, create a module as shown below:
+
+    defmodule MyApp.MyType do
+        use Ecto.ParameterizedType
+
+        def type(_params), do: :string
+
+        def init(opts) do
+          validate_opts(opts)
+          Enum.into(opts, %{})
+        end
+
+        def cast(data, params) do
+          ...
+          cast_data
+        end
+
+        def load(data, _loader, params) do
+          ...
+          {:ok, loaded_data}
+        end
+
+        def dump(data, dumper, params) do
+          ...
+          {:ok, dumped_data}
+        end
+
+        def equal?(a, b, _params) do
+          a == b
+        end
+      end
+
+  To use this type in a schema field, specify the type and parameters like this:
+
+      schema "foo" do
+        field "bar", MyApp.MyType, opt1: :baz, opt2: :boo
+      end
+
+  """
+
+  @type opts :: keyword()
+
+  @type params :: term()
+
+  @callback init(opts :: opts()) :: params()
+
+  @callback cast(data :: term, params :: params()) ::
+              {:ok, term} | {:error, keyword()} | :error
+
+  @callback load(value :: any(), loader :: function(), params :: params()) :: {:ok, value :: any()} | :error
+
+  @callback dump(value :: any(), dumper :: function(), params :: params()) :: {:ok, value :: any()} | :error
+
+  @callback type(params :: params()) :: Ecto.Type.t()
+
+  @callback equal?(value1 :: any(), value2 :: any(), params :: params()) :: boolean()
+
+  @callback embed_as(format :: atom(), params :: params()) :: :self | :dump
+
+  @doc false
+  defmacro __using__(_) do
+    quote location: :keep do
+      @behaviour Ecto.ParameterizedType
+      # TODO: Make both equal? and embed_as specific only to parameterized types
+      def embed_as(_, _), do: :self
+      # TODO: evaluate if we should keep this once we add cast/3 and change/3
+      def equal?(term1, term2, _params), do: term1 == term2
+      defoverridable embed_as: 2, equal?: 3
+    end
+  end
+end
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1771,7 +1771,7 @@ defmodule Ecto.Schema do
 
   @doc false
   def __field__(mod, name, type, opts) do
-    check_field_type!(name, type, type, opts)
+    type = check_field_type!(mod, name, type, opts)
     Module.put_attribute(mod, :changeset_fields, {name, type})
     define_field(mod, name, type, opts)
   end
@@ -2035,20 +2035,21 @@ defmodule Ecto.Schema do
     end
   end
 
-  defp check_field_type!(name, :datetime, _, _opts) do
+  defp check_field_type!(_mod, name, :datetime, _opts) do
     raise ArgumentError, "invalid type :datetime for field #{inspect name}. " <>
                            "You probably meant to choose one between :naive_datetime " <>
                            "(no time zone information) or :utc_datetime (time zone is set to UTC)"
   end
 
-  defp check_field_type!(name, type, full_type, opts) do
+  defp check_field_type!(mod, name, type, opts) do
     cond do
       type == :any and !opts[:virtual] ->
         raise ArgumentError, "only virtual fields can have type :any, " <>
                              "invalid type for field #{inspect name}"
 
-      inner_type = inner_from_composite(type, name) ->
-        check_field_type!(name, inner_type, type, opts)
+      composite?(type, name) ->
+        {outer_type, inner_type} = type
+        {outer_type, check_field_type!(mod, name, inner_type, opts)}
 
       Ecto.Type.base?(type) ->
         type
@@ -2056,27 +2057,30 @@ defmodule Ecto.Schema do
       is_atom(type) and Code.ensure_compiled(type) == {:module, type} and function_exported?(type, :type, 0) ->
         type
 
+      is_atom(type) and Code.ensure_compiled(type) == {:module, type} and function_exported?(type, :type, 1) ->
+        {:parameterized, type, type.init(Keyword.merge(opts, field: name, schema: mod))}
+
       is_atom(type) and function_exported?(type, :__schema__, 1) ->
         raise ArgumentError,
-          "schema #{inspect full_type} is not a valid type for field #{inspect name}." <>
+          "schema #{inspect type} is not a valid type for field #{inspect name}." <>
           " Did you mean to use belongs_to, has_one, has_many, embeds_one, or embeds_many instead?"
 
       true ->
-        raise ArgumentError, "invalid or unknown type #{inspect full_type} for field #{inspect name}"
+        raise ArgumentError, "invalid or unknown type #{inspect type} for field #{inspect name}"
     end
   end
 
-  defp inner_from_composite({composite, inner_type} = type, name) do
+  defp composite?({composite, _} = type, name) do
     if Ecto.Type.composite?(composite) do
-      inner_type
+      true
     else
       raise ArgumentError,
-        "invalid or unknown composite #{inspect type} for field #{inspect name}" <>
-        " Did you mean to use array or map as first element of tuple instead?"
+        "invalid or unknown composite #{inspect type} for field #{inspect name}. " <>
+        "Did you mean to use array or map as first element of tuple instead?"
     end
   end
 
-  defp inner_from_composite(_type, _name), do: false
+  defp composite?(_type, _name), do: false
 
   defp store_mfa_autogenerate!(mod, name, type, mfa) do
     if autogenerate_id(type) do
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -230,7 +230,7 @@ defmodule Ecto.Type do
 
   See `c:embed_as/1`.
   """
-  def embed_as({:parameterized, module, config}, format), do: module.embed_as(config, format)
+  def embed_as({:parameterized, module, params}, format), do: module.embed_as(format, params)
   def embed_as({composite, _}, _format) when composite in @composite, do: :self
   def embed_as(base, _format) when base in @base, do: :self
   def embed_as(mod, format) do
@@ -298,7 +298,7 @@ defmodule Ecto.Type do
   """
   @spec type(t) :: t
   def type(type)
-  def type({:parameterized, module, config}), do: module.type(config)
+  def type({:parameterized, type, params}), do: type.type(params)
   def type({:array, type}), do: {:array, type(type)}
   def type({:map, type}), do: {:map, type(type)}
 
@@ -383,8 +383,8 @@ defmodule Ecto.Type do
 
   """
   @spec dump(t, term) :: {:ok, term} | :error
-  def dump({:parameterized, module, config}, value) do
-    module.dump(config, value, &dump/2)
+  def dump({:parameterized, module, params}, value) do
+    module.dump(value, &dump/2, params)
   end
 
   def dump(_type, nil) do
@@ -409,8 +409,8 @@ defmodule Ecto.Type do
   the given `dumper` function is used.
   """
   @spec dump(t, term, (t, term -> {:ok, term} | :error)) :: {:ok, term} | :error
-  def dump({:parameterized, module, config}, value, dumper) do
-    module.dump(config, value, dumper)
+  def dump({:parameterized, module, params}, value, dumper) do
+    module.dump(value, dumper, params)
   end
 
   def dump(_type, nil, _dumper) do
@@ -517,8 +517,8 @@ defmodule Ecto.Type do
 
   """
   @spec load(t, term) :: {:ok, term} | :error
-  def load({:parameterized, module, config}, value) do
-    module.load(config, value, &load/2)
+  def load({:parameterized, module, params}, value) do
+    module.load(value, &load/2, params)
   end
 
   def load(_type, nil) do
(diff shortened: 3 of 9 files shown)


lib/ecto/parameterized_type.ex
```diff
@@ -0,0 +1,83 @@
...
+      use Ecto.ParameterizedType
+
+      def type(_opts), do: :string
+
+      def init(opts) do
+        ...
+        Enum.into(opts, %{})
+      end
+
+      def cast(data, current, opts) do
```
josevalim, line 22: Instead of calling the third element `opts`, let's call it `config` or something else. `opts` kind of implies keyword. :)

lib/ecto/parameterized_type.ex
```diff
@@ -0,0 +1,83 @@
...
+
+    schema "foo" do
+      field "bar", {MyApp.MyType, opt1: :baz, opt2: :boo}
+    end
+
+  """
+
+  @type opts :: keyword() | map()
+
+  @callback init(opts :: keyword()) :: opts()
```
josevalim, line 52: I believe the field name would also be given on init, correct?

lib/ecto/parameterized_type.ex
```diff
@@ -0,0 +1,83 @@
...
+  @callback equal?(value :: any(), value :: any(), opts :: opts()) :: boolean()
+
+  @callback embed_as(format :: atom(), opts :: opts()) :: :self | :dump
+
+  @doc false
+  defmacro __using__(_) do
+    quote location: :keep do
+      @behaviour Ecto.ParameterizedType
+      def load(term, _loader, opts), do: load(term, opts)
+      def dump(term, _dumper, opts), do: dump(term, opts)
```
josevalim, line 76: We definitely don't need both `load/2` and `load/3`, so my suggestion is to keep it simply as `load/3` and `dump/3` and simplify the API.

lib/ecto/parameterized_type.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Ecto.ParameterizedType do
+  @moduledoc """
+  Parameterized types are similar to Ecto Custom Types, however they allow a set of options to be
+  specified in the schema which are initialized on compilation and passed to the callback functions
+  as a final argument. Note that options should be specified as a keyword, but it is idiomatic to
+  convert them to maps inside init for easier pattern matching in other callbacks.
+
```
josevalim, line 7: We should also a paragraph that explains that, opposite to regular types, custom types are responsible for handling `nil` in both `load` and `dump` callbacks.

test/ecto/type/enum_test.exs
```diff
@@ -0,0 +1,98 @@
+defmodule Ecto.Type.EnumTest do
```
josevalim, line 1: I am thinking this test is a bit too high-level for now. It probably makes sense to unit test those through the Ecto.Type API. We also need tests for equal?.

josevalim, line 1: Those tests are good but they are too high-level. We also need to unit test parameterized types directly through the Ecto.Type API.

lib/ecto/parameterized_type.ex
josevalim: Oh, I see you are passing it as an option. I am thinking it may be better to make this become: `init(schema, field, value)` and we don't put any restriction on the value. Thoughts?

lib/ecto/type.ex
```diff
@@ -714,19 +718,24 @@ defmodule Ecto.Type do
       :error
 
   """
-  @spec cast(t, term) :: {:ok, term} | {:error, keyword()} | :error
-  def cast({:parameterized, module, config}, value), do: module.cast(config, value)
-  def cast({:in, _type}, nil), do: :error
-  def cast(_type, nil), do: {:ok, nil}
+  @spec cast(t, term, term) :: {:ok, term} | {:error, keyword() | Ecto.Changeset.t()} | :error | :ignore
+  def cast(_, _, current \\ nil)
```
josevalim, line 722: So this is a bit complicated... `cast` is called both by `Ecto.Changeset` and also by `Ecto.Query`. Once we pass `current` here, we are lumping both of them together, but in one of them we will never actually have a `current`. To make matters further apart, an embed would never really cast for a query, only for a changeset.

So it may be that, for parameterized types, we will have two cast functions, one receives the current value and the other does not. We can also give different names to them. 

For `cast+current`, we don't need to add it to Ecto.Type, we can call it directly from the changeset. So in a nutshell, my suggestion is to remove the extra `current` argument from here for now. We will add a new `cast` function once we work on the changeset.

lib/ecto/parameterized_type.ex
Eiji7: Sorry, maybe I do not see something, but what is the purpose of passing `schema` and `field` to any callback? I don't see a use case and therefore I don't believe anybody would need those. Also anything dynamic like `enum` values should be passed to `attrs`/`config`/`opts`/or however we name it, right?

josevalim:
Awesome job @TheFirstAvenger, I have added some comments. I did not comment too much on the Enum module for now - but since it does touch some of the Ecto.Schema API, I would like to discuss the API for specifying parameterized modules.

I am not a big fan of the current approach where `{Mod, opts}` get translated to `{:parameterized, Mod, opts}`. If we want to go the translation approach, then I would rather do this:

```elixir
field :foo, Ecto.Enum, values: [:foo, :bar, :baz]
```

Then we basically need to check if the type is a parameterized type or basic type based on its implementation. The benefit of this approach is that for a user, there is no distinction between a parameterized API and a regular one, so an implementation can even swap between them without breaking user compatibility.

However, if we don't want to have translation, then I would rather have users specify the whole thing:

```elixir
field :foo, {:parameterized, Ecto.Enum, values: [:foo, :bar, :baz]}
```

It is a bit too verbose, so users could wrap it as:

```elixir
field :foo, enum([:foo, :bar, :baz])
```

But this feels like pointless indirection.

Thoughts?

lib/ecto/parameterized_type.ex
josevalim: @Eiji7 enum definitely needs them, but we get to cheat because we have a macro for passing them in. So I can see others needing it in the future. The whole point of `init` is to receive all information so you decide what to keep. What you keep is what becomes `config`.

Eiji7:
> ```elixir
> field :foo, Ecto.Enum, values: [:foo, :bar, :baz]
> ```

I thought that there is already decision about API, but I was wrong. I really like this idea - that looks definitely better than previous one.

TheFirstAvenger:
> ```elixir
> field :foo, Ecto.Enum, values: [:foo, :bar, :baz]
> ```

Yes, I much prefer this. I had similarly thought the decision was to go with the 2-tuple, but I will update it to this much cleaner format.

lib/ecto/parameterized_type.ex
TheFirstAvenger: @josevalim Good point. It initially was always a keyword. Now that the standard will be to convert it to a map in init, opts doesn't work. How about `params` since it is a ParameterizedType?

josevalim: To me the params are the arguments given on `init` and not the inner state itself. That said, I am also fine with `params`. Anything besides `opts` is game for me. :)

lib/ecto/parameterized_type.ex
josevalim: Btw, if we decide to go with `field :foo, Ecto.Enum, values: [...]`, then I think this signature is perfect, as it will always be a keyword list. We can also pass the `:schema` and `:field` as an option too, the same as today. So I would just keep it as is. You can consider this particular comment resolved. :)

lib/ecto/schema.ex
```diff
@@ -1771,11 +1771,35 @@ defmodule Ecto.Schema do
 
   @doc false
   def __field__(mod, name, type, opts) do
+    type = maybe_to_parameterized(type, name, mod)
```
josevalim, line 1774: Instead of doing an early pass, change `check_field_type!` to return the type and we do all traversal there. Especially if we are going with `field :foo, Ecto.Enum, values: [...]`, since `check_field_type!` already has logic to validate custom types.

lib/ecto/parameterized_type.ex
TheFirstAvenger: Sounds good. I am thinking that `opts` will remain the name of the keyword that is passed into init (including what we call it in schema as we are parsing the field macro), and params is the map that is returned from init, and everywhere that is used in subsequent callbacks. As an example, ParameterizedType would have these:

```elixir
  @type opts :: keyword()

  @type params :: map()

  @callback init(opts :: opts()) :: params()

  @callback type(params :: params()) :: Ecto.Type.t()
```

lib/ecto/type.ex
TheFirstAvenger: Sounds good. Yeah, `current` came into play with `Embed`, but isn't necessary in v1.

lib/ecto/parameterized_type.ex
josevalim: Params should be any `term`. There is no reason for us to impose it being a map. :)

TheFirstAvenger:
@josevalim I have implemented all of the PR feedback, with the exception of adding tests specifically for ParameterizedType, which I will tackle tomorrow. Let me know if there is anything else missing or incorrect.

lib/ecto/parameterized_type.ex
```diff
@@ -0,0 +1,82 @@
...
+  @callback embed_as(format :: atom(), params :: params()) :: :self | :dump
+
+  @doc false
+  defmacro __using__(_) do
+    quote location: :keep do
+      @behaviour Ecto.ParameterizedType
+      def embed_as(_, _), do: :self
+      def equal?(term1, term2, _params), do: term1 == term2
+      def init(opts), do: Enum.into(opts, %{})
```
josevalim, line 78: Let's remove the default implementation for `init`. I don't think it is a good idea for the default to simply collecting everything. Typically, you will want to validate options and keep only what you need. You will also want to precompute as much as you can to avoid compile time work. For example, for enums, you would do something like this:

```elixir
def init(opts) do
  values = Keyword.fetch!(opts, :values)
  user_to_db = Map.new(values, fn k -> {k, Atom.to_string(v)} end)
  db_to_user = Map.new(values, fn k -> {Atom.to_string(v), k} end)
  %{user_to_db: user_to_db, db_to_user: db_to_user}
end
```

So load and dump happens as efficiently as possible. For `Ecto.Embedded`, you can make it so:

```elixir
def init(_opts), do: raise "TODO: invoke this directly"
```

lib/ecto/parameterized_type.ex
```diff
@@ -0,0 +1,82 @@
...
+  @callback equal?(value1 :: any(), value2 :: any(), params :: params()) :: boolean()
+
+  @callback embed_as(format :: atom(), params :: params()) :: :self | :dump
+
+  @doc false
+  defmacro __using__(_) do
+    quote location: :keep do
+      @behaviour Ecto.ParameterizedType
+      def embed_as(_, _), do: :self
+      def equal?(term1, term2, _params), do: term1 == term2
```
josevalim, line 77: I have a hunch we won't need `equal?/3` once we add `cast/3` and `change/3` because those callbacks can simply return `ignore` instead. So can you please add a TODO?

```suggestion
      # TODO: evaluate if we should keep this once we add cast/3 and change/3
      def equal?(term1, term2, _params), do: term1 == term2
```

lib/ecto/repo/schema.ex
```diff
@@ -899,9 +899,13 @@ defmodule Ecto.Repo.Schema do
       {:ok, value} ->
         value
       :error ->
+        type_str = case type do
+          {:parameterized, t, _} -> t
+          _ -> type
+        end
```
josevalim, line 905: I would revert this change simply because in all other errors it will show up as `{:paramterized, ..., ...}`, so I wouldn't sweat about it for now.

lib/ecto/schema.ex
```diff
@@ -1771,11 +1771,15 @@ defmodule Ecto.Schema do
...
   @doc false
   def __field__(mod, name, type, opts) do
-    check_field_type!(name, type, type, opts)
+    # type = convert(type)
+    type = check_field_type!(mod, name, type, type, opts)
     Module.put_attribute(mod, :changeset_fields, {name, type})
     define_field(mod, name, type, opts)
   end
 
+  # defp convert(:enum), do: Ecto.Type.Enum
```
josevalim, line 1780: Btw, in the future we will simply call it `Ecto.Enum` and we won't need to convert this. We already do this for `Ecto.UUID`, so it is fine. :) So you can remove the commented out code!

lib/ecto/parameterized_type.ex
TheFirstAvenger: Done. I really like this mapping method, it makes cast/load/dump much nicer as declarative instead of imperative.

lib/ecto/schema.ex
TheFirstAvenger: Done. `Ecto.Type.Enum` and `:enum` are both now `Ecto.Enum`

TheFirstAvenger:
@josevalim All feedback addressed and pushed, starting on ParameterizedType tests now.

TheFirstAvenger:
@josevalim ParameterizedTypeTest added. Took a bit longer because in comparing with Ecto.Type tests, I realized that we haven't addressed :parameterized in :array, :map, or :maybe composite types. I wrote tests which I believe reflect how it _should_ work, however they don't currently work, and my initial pass at trying to get them to work got messy, so I pulled back to just testing what we have currently working, and added `:skip` to the :array, :map, and :maybe tests.

lib/ecto/type.ex
```diff
@@ -439,6 +439,10 @@ defmodule Ecto.Type do
     array(value, type, dumper, [])
   end
 
+  def dump({:parameterized, type, params}, value, dumper) do
+    type.dump(value, dumper, params)
+  end
+
```
josevalim, line 445: I think we can remove this. It is a duplicate of the same clause on line 412:

```suggestion
```

test/ecto/parameterized_type_test.exs
```diff
@@ -0,0 +1,143 @@
+defmodule Ecto.ParameterizedTypeTest do
+  use ExUnit.Case, async: true
+
+
```
josevalim, line 4:
```suggestion
```

test/ecto/type/enum_test.exs
```diff
@@ -0,0 +1,120 @@
+defmodule Ecto.EnumTest do
+  use ExUnit.Case
```
josevalim, line 2:
```suggestion
  use ExUnit.Case, async: true
```

Also please rename the test file to `test/ecto/enum_test.exs`.

test/ecto/parameterized_type_test.exs
```diff
@@ -0,0 +1,143 @@
+defmodule Ecto.ParameterizedTypeTest do
```
josevalim, line 1: Great job!

test/support/enum.ex
```diff
@@ -0,0 +1,64 @@
+defmodule Ecto.Enum do
```
josevalim, line 1: I think we can move this to lib/ecto/enum.ex to be quite honest. It is almost ready, we just really need docs.

test/support/enum.ex
```diff
@@ -0,0 +1,64 @@
...
+      raise ArgumentError, "Ecto.Enum types must have a values option specified as a list of atoms, e.g. field :my_field, Ecto.Enum, values: [:foo, :bar]"
+    end
+
+    user_to_db = Map.new(values, &{&1, Atom.to_string(&1)})
+    db_to_user = Map.new(values, &{Atom.to_string(&1), &1})
+    %{user_to_db: user_to_db, db_to_user: db_to_user}
+  end
+
+  def values({:parameterized, __MODULE__, params}), do: values(params)
+  def values(%{user_to_db: user_to_db}), do: Map.keys(user_to_db)
```
josevalim, line 29: I don't think we should make this public. :)

test/support/enum.ex
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def load(data, _loader, %{db_to_user: db_to_user}) do
+    case db_to_user do
+      %{^data => as_atom} -> {:ok, as_atom}
+      _ -> :error
+    end
+  end
+
+  def dump(data, _dumper, %{user_to_db: user_to_db}) do
```
josevalim, line 50: We should probably handle `nil` explicitly here too, WDYT?

test/ecto/parameterized_type_test.exs
```diff
@@ -0,0 +1,143 @@
...
+    def params, do: @params
+
+    def init([some_opt: :some_opt_value, field: :my_type, schema: _]), do: @params
+    def type(@params), do: :custom
+    def load(_, _, @params), do: {:ok, :load}
+    def dump( _, _, @params),  do: {:ok, :dump}
+    def cast( _, @params),  do: {:ok, :cast}
+    def equal?(true, _, @params), do: true
+    def equal?(_, _, @params), do: false
+    def embed_as(_, @params), do: :dump
```
josevalim, line 18: Since `Ecto.Enum` has embed_as `:dump`, let's make this one `:self` so we can test different scenarios.

josevalim:
@TheFirstAvenger I have dropped some comments. I will look into the failing tests tomorrow morning and hopefully we can merge this. Can you give me access to your fork just in case?

test/ecto/parameterized_type_test.exs
TheFirstAvenger: Done. I also added tests for the error paths (I originally added it for :maybe, but that isn't working yet).

TheFirstAvenger:
@josevalim Done, you have access. Heads up that I squashed all the commits and force-pushed to clean it up a bit.

josevalim:

wojtekmach:
@TheFirstAvenger and others involved: Thank you, this is really exciting!

josevalim:
I think we are good for the next step, which is to convert the callbacks in Relation - starting with the ones used by changesets - to Ecto.ParameterizedType. This is probably when we will introduce the specific `cast` and `change` callbacks. :)

josevalim:
Oh, we also need to document the callbacks on "Ecto.ParamterizedType", right now none of them are documented.

lib/ecto/parameterized_type.ex
```diff
@@ -0,0 +1,99 @@
...
+
+  For example, `field :foo, :string` behaves the same for every field.
+  On the other hand, `field :foo, Ecto.Enum, values: [:foo, :bar, :baz]`
+  will likely have a different set of values per field.
+
+  Note that options are specified as a keyword, but it is idiomatic to
+  convert them to maps inside `c:init/1` for easier pattern matching in
+  other callbacks.
+
+  Parameterized types are a superset of regualr types. In other words,
```
brianbowden, line 17: Got a typo here! ("regualr")

josevalim:
I have created an issue so we can continue the discussion of next steps: #3380 - i think the work in the previous PR will be relevant for us to move forward :)

PR #3372, opened 2020-08-08
Fix check_field_type! for invalid composite types

### LINKED ISSUE #3371: The schema validation type for `{:array, inner_type} `  is not working as the query one
### Environment

* Elixir version: 1.10.1

* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): postgres (PostgreSQL) 12.2

* Ecto version: 3.4.4

* Database adapter and version (mix deps): Postgrex 0.15.5

* Operating system: Mac OS and Alpine Linux

### Current behavior

The schema check for `{:array, inner_type} `  is not working as the query one.

### Expected behavior

The schema validation for types for `{:array, inner_type} ` is not working as the query one.

When we create a field with `{:array, :jsonb}`  in the schema and this column is empty in the database we don't receive any error on compilation time. 

But when we run a query in the table that has this column and has a value we receive an exception telling us that the `:jsonb` does not have a type definition.

```
    ** (UndefinedFunctionError) function :jsonb.type/0 is undefined (module :jsonb is not available)
        :jsonb.type()
        (ecto 3.4.4) lib/ecto/type.ex:307: Ecto.Type.type/1
        (ecto 3.4.4) lib/ecto/type.ex:916: Ecto.Type.adapter_load/3
        (ecto 3.4.4) lib/ecto/repo/queryable.ex:381: Ecto.Repo.Queryable.struct_load!/6
        (ecto 3.4.4) lib/ecto/repo/queryable.ex:213: anonymous fn/5 in Ecto.Repo.Queryable.preprocessor/3
        (elixir 1.10.1) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
        (ecto 3.4.4) lib/ecto/repo/queryable.ex:200: Ecto.Repo.Queryable.execute/4
        (ecto 3.4.4) lib/ecto/repo/queryable.ex:17: Ecto.Repo.Queryable.all/3
        (ecto 3.4.4) lib/ecto/repo/queryable.ex:112: Ecto.Repo.Queryable.one/3
```

I would expect to have the exception also on compilation time, but it is not happening because we are checking it differently. 
At the Schema level, we check a tuple, and at the Query level, we check only the type.

```
iex(1)> Ecto.Type.primitive?(:jsonb)
false
iex(2)> Ecto.Type.primitive?({:array, :jsonb})
true
```

**Schema**

[https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L2050-L2051](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L2050-L2051)

**Query**

[https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L300-L311](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L300-L311)

### Application Example

[ecto_bug.zip](https://github.com/elixir-ecto/ecto/files/5044145/ecto_bug.zip)

Eiji7:
closes #3371

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -2047,11 +2047,13 @@ defmodule Ecto.Schema do
         raise ArgumentError, "only virtual fields can have type :any, " <>
                              "invalid type for field #{inspect name}"
 
-      is_tuple(type) and tuple_size(type) == 2 and type |> elem(0) |> Ecto.Type.composite?() ->
-        inner_type = elem(type, 1)
+      inner_type = inner_from_composite(type, name) ->
         check_field_type!(name, inner_type, type, opts)
 
-      Ecto.Type.primitive?(type) ->
+      parameterized?(type, name) ->
+        type
+
+      Ecto.Type.base?(type) ->
         type
 
       is_atom(type) and Code.ensure_compiled(type) == {:module, type} and function_exported?(type, :type, 0) ->
@@ -2066,6 +2068,37 @@ defmodule Ecto.Schema do
         raise ArgumentError, "invalid or unknown type #{inspect full_type} for field #{inspect name}"
     end
   end
+
+  defp inner_from_composite({composite, inner_type} = type, name) do
+    if Ecto.Type.composite?(composite) do
+      inner_type
+    else
+      raise ArgumentError,
+        "invalid or unknown composite #{inspect type} for field #{inspect name}" <>
+        " Did you mean to use array or map as first element of tuple instead?"
+    end
+  end
+
+  defp inner_from_composite(_type, _name), do: false
+
+  defp parameterized?({:parameterized, module, _opts} = type, name) do
+    cond do
+      Code.ensure_compiled(module) != {:module, module} ->
+        raise ArgumentError,
+          "schema #{inspect type} is not a valid parameterized type for field #{inspect name}." <>
+          " Module from parameterized type cannot be compiled."
+
+      not function_exported?(module, :type, 1) ->
+        raise ArgumentError,
+          "schema #{inspect type} is not a valid parameterized type for field #{inspect name}." <>
+          " Module from parameterized type does not implements Ecto.ParameterizedType behaviour."
+
+      true ->
+        type
+    end
+  end
+
+  defp parameterized?(_type, _name), do: false
 
   defp store_mfa_autogenerate!(mod, name, type, mfa) do
     if autogenerate_id(type) do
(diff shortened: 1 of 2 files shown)


lib/ecto/schema.ex
```diff
@@ -2041,12 +2041,18 @@ defmodule Ecto.Schema do
                            "(no time zone information) or :utc_datetime (time zone is set to UTC)"
   end
 
-  defp check_field_type!(name, type, opts) do
+  defp check_field_type!(name, type, opts, full_type \\ nil) do
+    full_type = if is_nil(full_type), do: type, else: full_type
```
josevalim, line 2045: Instead of passing it as `nil`, maybe we should change the caller to also pass `type` here and we never have `nil`?

lib/ecto/schema.ex
```diff
@@ -2041,12 +2041,18 @@ defmodule Ecto.Schema do
...
+
     cond do
       type == :any and !opts[:virtual] ->
         raise ArgumentError, "only virtual fields can have type :any, " <>
                              "invalid type for field #{inspect name}"
 
+      is_tuple(type) and tuple_size(type) == 2 and type |> elem(0) |> Ecto.Type.composite?() ->
+        inner_type = elem(type, 1)
+        check_field_type!(name, inner_type, opts, type)
```
josevalim, line 2054: If it is a two element tuple, it is always composite:

```suggestion
      is_tuple(type) and tuple_size(type) == 2 ->
        check_field_type!(name, elem(type, 1), opts, type)
```

We will also need a three element tuple check for parameterized types. :D

Eiji7: @josevalim 
> If it is a two element tuple, it is always composite:

Sorry, I do not understand that. If so then why we have a `Ecto.Type.composite?` function and how would let's say `{:jsonb, :jsonb}` work?

> We will also need a three element tuple check for parameterized types. :D

Ah right, I forgot about that

lib/ecto/schema.ex
```diff
@@ -2035,18 +2035,22 @@ defmodule Ecto.Schema do
...
   end
 
-  defp check_field_type!(name, type, opts) do
+  defp check_field_type!(name, type, full_type, opts) do
     cond do
       type == :any and !opts[:virtual] ->
         raise ArgumentError, "only virtual fields can have type :any, " <>
                              "invalid type for field #{inspect name}"
 
+      is_tuple(type) and tuple_size(type) == 2 and type |> elem(0) |> Ecto.Type.composite?() ->
```
josevalim, line 2050: Ah, good point. Can we at least move this to a private function? so we just call it as `composite?(type) ->` or, even better, `inner_type = inner_from_composite(type) ->`. WDYT?

josevalim:

PR #3373, opened 2020-08-08
add recommended Elixir and Erlang/OTP combinations to CI

adrianomitre:
This pull request makes sure all recommended Elixir and Erlang/OTP combinations are run on CI.

Quoting José Valim:
> I would still advise folks to go with:
> * For each Elixir version, add a run with earliest supported Erlang/OTP
> * For the last Elixir version, also test the latest supported Erlang/OTP

Please refer to the original discussion on https://github.com/dashbitco/broadway/pull/188 for the full rationale.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,24 +1,24 @@
 name: CI
 on: [pull_request, push]
 jobs:
- mix_test:
-   name: mix test (Elixir ${{ matrix.elixir }} OTP ${{ matrix.otp }})
-   strategy:
-     matrix:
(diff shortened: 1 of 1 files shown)


josevalim:

adrianomitre:
@josevalim I did not get the reason for https://github.com/elixir-ecto/ecto/pull/3373/commits/0d9bf879360807653c7bf9912f18c07f46e92d13. Would you mind explaining?

josevalim:
I would like to talk to other maintainers before enabling the flag. Ecto is more complex than your usual project, so there may be more scenarios to consider.

PR #3374, opened 2020-08-08
Add mix format --check-formatted to CI and ensure it passes

adrianomitre:
This pull request builds upon https://github.com/elixir-ecto/ecto/pull/3373, so please merge this one after it.


josevalim:
Ecto is not formatted and I don't think we should blindly format a codebase, as sometimes the result for very nested code can be unreadable. It needs to be done carefully and even then it has the dowside of messing history, so a :-1: from me for now. Thanks.

adrianomitre:
> I don't think we should blindly format a codebase, as sometimes the result for very nested code can be unreadable. It needs to be done carefully and even then it has the dowside of messing history

I wholeheartedly agree with you that automated formatting does not necessarily lead to the most beautiful or readable code and it can indeed hurt instead of helping. Having said that, most of the changes `mix format` produced did not mess with very nested code (which [should be rare](https://blog.jetbrains.com/idea/2017/08/code-smells-deeply-nested-code/) anyway) nor produced code that is harder to read. In many instances it just added parenthesis.

I tend to believe that manually revising the auto-formatted code could make it easy to read, beautiful etc while compliant to `mix format --check-formatted` in most, if not all, cases. Given that there is no workaround, however, to the "messing history" concern (i.e., making old code look like new on first-order inspection with, e.g., git blame), I understand the rejection of the pull request instead of requesting improvements.

In spite of that, maybe https://github.com/elixir-ecto/ecto/pull/3374/commits/c6cad55d9a4f4cd69828591cfc9ba8f403ae7426, which fixes `.formatter.exs`, and https://github.com/elixir-ecto/ecto/pull/3374/commits/c6cad55d9a4f4cd69828591cfc9ba8f403ae7426, which make the test position-indenpendent, could be cherry-picked for a separate pull request. What are you thoughts on that?

josevalim:
>  Having said that, most of the changes mix format produced did not mess with very nested code (which should be rare anyway) nor produced code that is harder to read. In many instances it just added parenthesis.

If you look at the formatted query/planner.ex and under the tests in test/ecto/query/builder, you will find many occurrences of multiple line function definitions, multiple line function invocations, and similar that could become a two-line call if the code was broken apart.

I don't think we should merge c6cad55 either, because it will cause people to automatically format the codebase and similar when submitting pull requests. I need to look carefully if cda0f95 is indeed the best approach for the problem too. There are other places where the line could be stripped in the same file but we don't.

PR #3377, opened 2020-08-11
Provide guidance for `:telemetry_prefix` when you have multiple dbs

davydog187:
Speaking with @bryannaegele, we were incorrectly setting telemetry prefixes for our app that uses multiple databases. This should help push others in the right direction.


josevalim:

Issue #3380, opened 2020-08-12
Improvements to paramterized types


josevalim:
* [x] Add docs to all Ecto.ParameterizedType callbacks

* [ ] Discuss if and how to move the functionality in Ecto.Changeset to Ecto.ParameterizedType (cast, change, apply_changes)

* [ ] Discuss if and how to move the functionality in Ecto.Repo.Schema to Ecto.ParameterizedType (Embedded.prepare, surface_changes)

Eiji7:
@josevalim I just want to confirm … If we agree on 2nd and 3rd point we would be able to write such code:

```elixir
defmodule MyApp.Comment do
  use Ecto.Schema
  use MyLib

  schema "comments" do
    belongs_to :author, MyApp.User
    exactly_one_of :source, :belongs_to do
      polym :post, MyApp.Post
      # …
    end
    # helpful macro which would generate something like:
    # belongs_to :post, MyApp.Post
    # …
    # field :source, MyLib.ExactlyOneOf, polym_fields: [:post, …]
  end
end
```
which would allow to work on `:source` field just like on `:post` association (in case `%MyApp.Post{…}` would be used or in this example `post_id` would be the only non-null `column` in database) just by `Ecto.ParameterizedType` callbacks without any workarounds and/or rewriting any part of `ecto`/`ecto_sql` code, right?

I believe that true polymorphic associations would be a good example for a complex real world use case of all callbacks. Here ideally if in callbacks we would receive full `schema` instead of only a field `value`. Maybe it would be good to consider if `Ecto.ParameterizedType` would receive full `schema` when `virtual` option is set to `true`? Look that for example callback `load` (from database) of `virtual` (field) does not makes sense unless it have access to full `schema` or at least a specified list of `fields` and looks like the same applies to other (all?) callbacks.

WDYT?

josevalim:
Associations have other complexities that are not taken into account here, such as tracking transactions. Plus there are many other association specific callbacks. So what you suggest above is a non-goal, at the moment.

I mean, we haven't even solved embeds yet, talking about associations is definitely **too early**.

Eiji7:
> @josevalim Associations have other complexities that are not taken into account here, such as tracking transactions. Plus there are many other association specific callbacks. So what you suggest above is a non-goal, at the moment.

[...]

PR #3382, opened 2020-08-12
Parameterized type change

Eiji7:
* [x] Added `change/3` callback for `Ecto.ParameterizedType`.
* [x] Added `change/3` callback implementation for `Ecto.Enum`.
* [x] Added tests for new callback.

**TODO**:
* [x] `Ecto.Embedded` support and tests for it
* [x] Define default implementation for `change/3` on `use Ecto.ParameterizedType`
* [ ] Implement `change_fun` for arrays and maps. For arrays, we need to recursively call `change` on pair-wise. If the new_value has more entries than old_value, we need to consider the remaining entries are nil. For maps, we need similar logic and compare them based on the keys.

Related issue: #3380 

cc @josevalim @TheFirstAvenger 

Does it looks good? How should `Ecto.Embedded` work?

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1221,9 +1221,15 @@ defmodule Ecto.Changeset do
     old_value = Map.get(data, key)
 
     case type.change(old_value, new_value, params) do
-      {:ok, change} -> {Map.put(changes, key, change), errors, valid?}
-      {:error, error} -> {changes, [{key, error} | errors], false}
-      :skip -> {Map.delete(changes, key), errors, valid?}
+      {:ok, change} ->
+        {Map.put(changes, key, change), errors, valid?}
+
+      {:error, error} ->
+        {message, opts} = Keyword.pop(error, :message, "cannot be changed")
+        {changes, [{key, {message, opts}} | errors], false}
+
+      :skip ->
+        {Map.delete(changes, key), errors, valid?}
     end
   end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1217,6 +1217,19 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "field names given to change/put_change must be atoms, got: `#{inspect(key)}`"
   end
 
+  defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
+    old_value = Map.get(data, key)
+
+    if type.equal?(old_value, new_value, params) do
+      {Map.delete(changes, key), errors, valid?}
+    else
+      case type.change(old_value, new_value, params) do
+        {:ok, change} -> {Map.put(changes, key, change), errors, valid?}
+        {:error, error} -> {changes, [{key, error} | errors], false}
+      end
+    end
+  end
+
   defp put_change(data, changes, errors, valid?, key, value, type) do
     if not Ecto.Type.equal?(type, Map.get(data, key), value) do
       {Map.put(changes, key, value), errors, valid?}
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1220,13 +1220,13 @@ defmodule Ecto.Changeset do
   defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
     old_value = Map.get(data, key)
 
-    if type.equal?(old_value, new_value, params) do
-      {Map.delete(changes, key), errors, valid?}
-    else
+    if type.equal?(old_value, new_value, params) in [:skip, false] do
       case type.change(old_value, new_value, params) do
         {:ok, change} -> {Map.put(changes, key, change), errors, valid?}
         {:error, error} -> {changes, [{key, error} | errors], false}
       end
+    else
+      {Map.delete(changes, key), errors, valid?}
     end
   end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1217,10 +1217,10 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "field names given to change/put_change must be atoms, got: `#{inspect(key)}`"
   end
 
-  defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
+  defp put_change(data, changes, errors, valid?, key, new_value, type) do
     old_value = Map.get(data, key)
 
-    case type.change(old_value, new_value, params) do
+    case Ecto.Type.change(type, old_value, new_value) do
       {:ok, change} ->
         {Map.put(changes, key, change), errors, valid?}
 
@@ -1230,14 +1230,6 @@ defmodule Ecto.Changeset do
 
       :skip ->
         {Map.delete(changes, key), errors, valid?}
-    end
-  end
-
-  defp put_change(data, changes, errors, valid?, key, value, type) do
-    if not Ecto.Type.equal?(type, Map.get(data, key), value) do
-      {Map.put(changes, key, value), errors, valid?}
-    else
-      {Map.delete(changes, key), errors, valid?}
     end
   end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1228,7 +1228,7 @@ defmodule Ecto.Changeset do
         {Map.put(changes, key, change), errors, false}
 
       {:ok, change, true} ->
-        {Map.put(changes, key, change), errors, true}
+        {Map.put(changes, key, change), errors, valid?}
 
       {:ok, change} ->
         {Map.put(changes, key, change), errors, valid?}
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1193,8 +1193,11 @@ defmodule Ecto.Changeset do
     %{changeset | changes: changes, errors: errors, valid?: valid?}
   end
 
-  defp put_change(data, changes, errors, valid?, key, value, {tag, relation})
-       when tag in @relations do
+  defp put_change(data, changes, errors, valid?, key, value, {:embed, relation}) do
+    put_change(data, changes, errors, valid?, key, value, {:parameterized, Ecto.Embedded, relation})
+  end
+
+  defp put_change(data, changes, errors, valid?, key, value, {:assoc, relation}) do
     original = Map.get(data, key)
     current = Relation.load!(data, original)
 
@@ -1221,6 +1224,12 @@ defmodule Ecto.Changeset do
     old_value = Map.get(data, key)
 
     case Ecto.Type.change(type, old_value, new_value) do
+      {:ok, change, false} ->
+        {Map.put(changes, key, change), errors, false}
+
+      {:ok, change, true} ->
+        {Map.put(changes, key, change), errors, true}
+
       {:ok, change} ->
         {Map.put(changes, key, change), errors, valid?}
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1230,9 +1230,6 @@ defmodule Ecto.Changeset do
       {:ok, change, true} ->
         {Map.put(changes, key, change), errors, valid?}
 
-      {:ok, change} ->
-        {Map.put(changes, key, change), errors, valid?}
-
       {:error, error} ->
         {message, opts} = Keyword.pop(error, :message, "is invalid")
         {changes, [{key, {message, opts}} | errors], false}
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1220,13 +1220,10 @@ defmodule Ecto.Changeset do
   defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
     old_value = Map.get(data, key)
 
-    if type.equal?(old_value, new_value, params) in [:skip, false] do
-      case type.change(old_value, new_value, params) do
-        {:ok, change} -> {Map.put(changes, key, change), errors, valid?}
-        {:error, error} -> {changes, [{key, error} | errors], false}
-      end
-    else
-      {Map.delete(changes, key), errors, valid?}
+    case type.change(old_value, new_value, params) do
+      {:ok, change} -> {Map.put(changes, key, change), errors, valid?}
+      {:error, error} -> {changes, [{key, error} | errors], false}
+      :skip -> {Map.delete(changes, key), errors, valid?}
     end
   end
 
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -41,7 +41,7 @@ defmodule Ecto.Enum do
   def change(_old_value, new_value, params) do
     case params do
       %{on_dump: %{^new_value => _}} -> {:ok, new_value}
-      _ -> {:error, {"unknown enum value", value: new_value}}
+      _ -> {:error, [message: "unknown enum value", value: new_value]}
    end
   end
 
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -38,6 +38,14 @@ defmodule Ecto.Enum do
   end
 
   @impl Ecto.ParameterizedType
+  def change(_old_value, new_value, params) do
+    case params do
+      %{on_dump: %{^new_value => _}} -> {:ok, new_value}
+      _ -> {:error, {"unknown enum value", value: new_value}}
+   end
+  end
+
+  @impl Ecto.ParameterizedType
   def cast(data, params) do
     case params do
       %{on_load: %{^data => as_atom}} -> {:ok, as_atom}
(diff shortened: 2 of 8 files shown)


lib/ecto/changeset.ex
```diff
@@ -1217,6 +1217,19 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "field names given to change/put_change must be atoms, got: `#{inspect(key)}`"
   end
 
+  defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
+    old_value = Map.get(data, key)
+
+    if type.equal?(old_value, new_value, params) do
```
josevalim, line 1223: My plan is to not have `equal?` for parameterized types. Instead, they can simply return `:ignore` or similar. That's because sometimes checking if they are the same or not is non-straight-forward. 

For the rest, I will only know if this is good once we make it work with Ecto.Embedded. :)

lib/ecto/type.ex
```diff
@@ -972,6 +972,23 @@ defmodule Ecto.Type do
     end
   end
 
+  @spec change(t, term, term) :: {:ok, term} | {:error, Ecto.Changeset.error()}
+  def change(type, term1, term2) do
```
josevalim, line 976: I don't think we need to add this function here. We can simply match and call change directly from `Ecto.Changeset`. :)

lib/ecto/parameterized_type.ex
```diff
@@ -99,6 +99,9 @@ defmodule Ecto.ParameterizedType do
   """
   @callback init(opts :: opts()) :: params()
 
+  @callback change(old_value :: any(), new_value :: any(), params :: params()) ::
+              {:ok, new_value :: any()} | {:error, Ecto.Changeset.error()}
```
josevalim, line 103: Perhaps it is best to make the last argument be a keyword list where you extract the message, similar to how we handle errors from types on `cast` here:

```elixir
      {:invalid, custom_errors} ->
        {message, new_errors} =
          custom_errors
          |> Keyword.put_new(:validation, :cast)
          |> Keyword.put(:type, type)
          |> Keyword.pop(:message, "is invalid")
        {changes, [{key, {message, new_errors}} | e
```

lib/ecto/type.ex
josevalim: Actually, ignore me. We need this function because we need to make it recursive to handle parameterized types inside arrays/maps. So that's what we should actually look into next.

lib/ecto/type.ex
```diff
@@ -972,6 +972,23 @@ defmodule Ecto.Type do
...
+    else
+      {:ok, term2}
+    end
+  end
+
+  defp change_fun({:parameterized, mod, params}) do
+    &mod.change(&1, &2, params)
+  end
+
+  defp change_fun(mod) when is_atom(mod), do: &mod.change/2
```
josevalim, line 988: We are not adding `change/2` to basic custom types, so we can remove this one.

lib/ecto/parameterized_type.ex
Eiji7: @josevalim I understand that you are talking about `{:error, Ecto.Changeset.error()}`, right? If so then please take a look at that type definition:

```elixir
@type error() :: {String.t(), Keyword.t()}
```
I used that `typespec`, because that's how the changesets works. Look at last line you quoted and how I'm handling errors:
> ```elixir
>       {:error, error} ->
>         {changes, [{key, error} | errors], false}
> ```
> Source: https://github.com/Eiji7/ecto/blob/4e0838d98c037ea49e468ae8c48c7999d2d3630d/lib/ecto/changeset.ex#L1204-L1205

Or maybe by last argument you mean `params :: params()`, so I should (for now) pass something like `[type: {:parameterized, type, params}]` instead of just `params` (of parameterized type) to the callback?

Eiji7:
@josevalim 2 commit == 2 outdated, for 1 I was not sure so asked
Rest is related to embeds - I would need to add implementation for them at end.

lib/ecto/type.ex
Eiji7: > I don't think we need to add this function here. We can simply match and call change directly from Ecto.Changeset. :)

I added it only for tests purposes, so I'm ok with moving it.

> Actually, ignore me. We need this function because we need to make it recursive to handle parameterized types inside arrays/maps. So that's what we should actually look into next.

So instead of `Ecto.Type` it should be in `Ecto.ParameterizedType`, right?

lib/ecto/changeset.ex
```diff
@@ -1217,6 +1217,19 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "field names given to change/put_change must be atoms, got: `#{inspect(key)}`"
   end
 
+  defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
+    old_value = Map.get(data, key)
+
+    if type.equal?(old_value, new_value, params) in [:skip, false] do
```
josevalim, line 1223: Sorry, I mean to say that `type.change` will return `:skip`, not `type.equal?`. :) We would skip this call altogether!

lib/ecto/enum.ex
```diff
@@ -35,6 +37,15 @@ defmodule Ecto.Enum do
     %{on_load: on_load, on_dump: on_dump}
   end
 
+  @impl Ecto.ParameterizedType
+  def change(_old_value, new_value, params) do
+    case params do
+      %{on_dump: %{^new_value => _}} -> {:ok, new_value}
+      _ -> {:error, {"unknown enum value", value: new_value}}
```
josevalim, line 44:
```suggestion
      _ -> {:error, [message: "unknown enum value", value: new_value]}
```

lib/ecto/parameterized_type.ex
josevalim: > I used that typespec, because that's how the changesets works. Look at last line you quoted and how I'm handling errors:

I am aware but we already have a precedent for returning errors from `Ecto.Type` and they are not of the type `Ecto.Changeset.error`. Instead they are a keyword list. Please look at Ecto.Type.cast return type. :)

lib/ecto/type.ex
josevalim: No, sorry for the confusion. This function should actually be here. Here is what it would look like:

    def change(type, old_value, new_value) do
      change_fun(type).(old_value, new_value)
    end

    defp change_fun({:parameterized, mod, params}), do: &mod.change(&1, &2, params)
    defp change_fun(t) do
      fn old_value, new_value ->
        if equal?(type, old_value, new_value) do
          :skip
        else
          {:ok, value}
        end
      end
    end

And then in `put_change`, you will use the same code for both regular
types and paramterized ones:

    old_value = Map.get(data, key)

    case Ecto.Type.change(type, old_value, new_value) do
      {:ok, change} ->
        {Map.put(changes, key, change), errors, valid?}

      {:error, error} ->
        {message, opts} = Keyword.pop(error, :message, "cannot be changed")
        {changes, [{key, {message, opts}} | errors], false}

      :skip ->
        {Map.delete(changes, key), errors, valid?}
    end

josevalim: We also need to implement it for change_fun for arrays and maps, but it is a bit more complex. For example, if you have an array of enums, but one array is bigger than the other, what do you do? Similar for maps.

lib/ecto/parameterized_type.ex
```diff
@@ -135,7 +138,7 @@ defmodule Ecto.ParameterizedType do
   @doc """
   Checks if two terms are semantically equal.
   """
-  @callback equal?(value1 :: any(), value2 :: any(), params :: params()) :: boolean()
+  @callback equal?(value1 :: any(), value2 :: any(), params :: params()) :: boolean() | :ignore
```
josevalim, line 141: The  `:ignore` (or `:skip`) should be in `change`. Please see the comment I posted on Ecto.Type for a reference implementation. :)

lib/ecto/changeset.ex
```diff
@@ -1217,6 +1217,16 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "field names given to change/put_change must be atoms, got: `#{inspect(key)}`"
   end
 
+  defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
+    old_value = Map.get(data, key)
+
+    case type.change(old_value, new_value, params) do
+      {:ok, change} -> {Map.put(changes, key, change), errors, valid?}
+      {:error, error} -> {changes, [{key, error} | errors], false}
```
josevalim, line 1225: Please do the Keyword.pop bit in the original implementation too. :)

lib/ecto/parameterized_type.ex
```diff
@@ -99,6 +99,9 @@ defmodule Ecto.ParameterizedType do
   """
   @callback init(opts :: opts()) :: params()
 
+  @callback change(old_value :: any(), new_value :: any(), params :: params()) ::
+              {:ok, new_value :: any()} | {:error, Ecto.Changeset.error()} | :skip
+
```
josevalim, line 104: I believe we can define a default implementation for change that simply does:

    if equal?(old_value, new_value, params), do: :skip, else: {:ok, new_value}

lib/ecto/changeset.ex
```diff
@@ -1221,9 +1221,15 @@ defmodule Ecto.Changeset do
...
 
     case type.change(old_value, new_value, params) do
-      {:ok, change} -> {Map.put(changes, key, change), errors, valid?}
-      {:error, error} -> {changes, [{key, error} | errors], false}
-      :skip -> {Map.delete(changes, key), errors, valid?}
+      {:ok, change} ->
+        {Map.put(changes, key, change), errors, valid?}
+
+      {:error, error} ->
+        {message, opts} = Keyword.pop(error, :message, "cannot be changed")
```
josevalim, line 1228: I just checked and for cast we do this:

```suggestion
        {message, opts} = Keyword.pop(error, :message, "is invalid")
```

I guess that's better than introducing a new option? WDYT?

lib/ecto/changeset.ex
Eiji7: This is resolved, but removing callback is not part of this PR :-)

lib/ecto/parameterized_type.ex
Eiji7: done

lib/ecto/changeset.ex
Eiji7: done

lib/ecto/enum.ex
Eiji7: done

lib/ecto/parameterized_type.ex
Eiji7: done

lib/ecto/changeset.ex
Eiji7: I'm always trying to do 1 reasonable thing (with passing tests) for one commit and well describe what I do in commit message.

Updated as in previous comment even before saw this one. :-)

lib/ecto/parameterized_type.ex
```diff
@@ -100,7 +100,7 @@ defmodule Ecto.ParameterizedType do
   @callback init(opts :: opts()) :: params()
 
   @callback change(old_value :: any(), new_value :: any(), params :: params()) ::
-              {:ok, new_value :: any()} | {:error, Ecto.Changeset.error()} | :skip
+              {:ok, new_value :: any()} | {:error, opts} | :skip
```
josevalim, line 103:
```suggestion
              {:ok, new_value :: any()} | {:error, keyword()} | :skip
```

lib/ecto/changeset.ex
josevalim: The callback will still stay, so no worries! :)

lib/ecto/changeset.ex
```diff
@@ -1217,6 +1217,22 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "field names given to change/put_change must be atoms, got: `#{inspect(key)}`"
   end
 
+  defp put_change(data, changes, errors, valid?, key, new_value, {:parameterized, type, params}) do
+    old_value = Map.get(data, key)
+
+    case type.change(old_value, new_value, params) do
```
josevalim, line 1223: You should use this branch for all types:

```suggestion
  defp put_change(data, changes, errors, valid?, key, new_value, type) do
    old_value = Map.get(data, key)

    case Ecto.Type.change(type, old_value, new_value) do
```

lib/ecto/parameterized_type.ex
Eiji7: Also already done in: https://github.com/elixir-ecto/ecto/pull/3382/commits/dc286435cd22afaacd503763e05ef37f50038c01

lib/ecto/parameterized_type.ex
Eiji7: Ah, what a mistake … `opts` here is for a different purpose - I made mistake as it's really general name, sorry :-)

lib/ecto/changeset.ex
Eiji7: Right, previously I was not sure if `Ecto.Type.change/3` would survive :-)

Eiji7:
@josevalim All suggestion done - I still need to make callback implementation for `embeds`

Should I wait for @TheFirstAvenger to see how things would go with `cast` and proposed by you `cast_change` (as it's related to `embeds`)?

josevalim:
>  All suggestion done - I still need to make callback implementation for embeds

It looks great. Besides embeds, we only need to define a default implementation for `change/3` on `use Ecto.ParamterizedType`.

> What we are going to do with embeds? You have mentioned cast_change before, so I'm unsure what to do.

You should be able to move the logic specific to embeds into `Ecto.Embedded.change/3`. You don't need to wait for `cast_change`, as this wouldn't dynamically depend on it. In order words, you should be able to add a clause like this:

```elixir
  defp put_change(data, changes, errors, valid?, key, value, {:embed, relation}) do
    put_change(data, changes, errors, valid?, key, value, {:parameterized, Ecto.Embedded, relation})
  end
```

And it seems you will also need to change the `c:Ecto.ParameterizedType.change/3` definition so it accepts `{:ok, value, valid?}` as return type. It is for when a change is accepted but some of its inner values are actually invalid.

lib/ecto/parameterized_type.ex
```diff
@@ -100,7 +100,7 @@ defmodule Ecto.ParameterizedType do
   @callback init(opts :: opts()) :: params()
 
   @callback change(old_value :: any(), new_value :: any(), params :: params()) ::
-              {:ok, new_value :: any()} | {:error, keyword()} | :skip
+              {:ok, new_value :: any(), valid? :: boolean()} | {:ok, new_value :: any()} | {:error, keyword()} | :skip
```
josevalim, line 103:
```suggestion
              {:ok, new_value :: any(), valid? :: boolean()} | {:error, keyword()} | :skip
```

lib/ecto/changeset.ex
```diff
@@ -1221,6 +1224,12 @@ defmodule Ecto.Changeset do
     old_value = Map.get(data, key)
 
     case Ecto.Type.change(type, old_value, new_value) do
+      {:ok, change, false} ->
+        {Map.put(changes, key, change), errors, false}
+
+      {:ok, change, true} ->
+        {Map.put(changes, key, change), errors, true}
```
josevalim, line 1231:
```suggestion
        {Map.put(changes, key, change), errors, valid?}
```

lib/ecto/changeset.ex
```diff
@@ -1221,6 +1224,12 @@ defmodule Ecto.Changeset do
...
+
       {:ok, change} ->
         {Map.put(changes, key, change), errors, valid?}
```
```suggestion
```

lib/ecto/parameterized_type.ex
Eiji7: @josevalim Why do you want to do that? The `valid?` was added only for `:embeds` purpose whereas other types still should return `{:ok, new_value}`, right?

Eiji7:
@josevalim As you saw I have just added support for `embeds`. As you said I would implement `change_fun` for arrays and maps in separate PR, so for me it's good to go.

lib/ecto/parameterized_type.ex
josevalim: The other types can return `{:ok, new_value, true}`. I think it is easier for everyone if the contracts are as minimal as they can be, instead of allowing many different returns.

josevalim:
Note I have revamped the discussion on #3380. My suggestion for now is to not merge the latest changes we worked on and keep them around until someone has an effective use case, as we won't really be able to move all of embeds to parameterized types.

lib/ecto/parameterized_type.ex
Eiji7: ah, so the purpose is to keep contracts minimal - no problem :slightly_smiling_face:

josevalim:
Closing this per #3380 but in case we get back to it, we will have a starting point. :) Thanks a lot! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #3384, opened 2020-08-13
Support casting Embeds via Changeset.cast()

TheFirstAvenger:
Continued work on ParameterizedType (#3380). This PR allows embeds to be supported through regular `Changeset.cast`, in addition to the already supported `Changeset.cast_embed`.

I duplicated the existing `embedded_test.exs`, moving the previous version unmodified to `embedded_via_cast_embed_test.exs`, and modifying `embedded_test.exs` to use `cast` directly instead of `cast_embed`. This allows us to use the `embedded_test.exs` file to review/validate the differences between `cast_embed` and `cast`.

There are two failing tests which are currently skipped in `type_test.exs`. I may be mistaken, but it looks like the existing test had reversed parameters in the changeset function. I swapped them, but the tests are still failing.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -536,6 +536,8 @@ defmodule Ecto.Changeset do
     case types do
       %{^key => {:assoc, _}} ->
         raise "casting assocs with cast/4 for #{inspect key} field is not supported, use cast_assoc/3 instead"
+      %{^key => {:embed, embedded}} ->
+        {:parameterized, Ecto.Embedded, embedded}
       %{^key => type} ->
         type
       _ ->
@@ -556,7 +558,7 @@ defmodule Ecto.Changeset do
     case params do
       %{^param_key => value} ->
         value = if value in empty_values, do: Map.get(defaults, key), else: value
-        case Ecto.Type.cast(type, value, current) do
+        case Ecto.Type.cast_change(type, value, current) do
           {:ok, value} ->
             if Ecto.Type.equal?(type, current, value) do
               :missing
@@ -800,7 +802,7 @@ defmodule Ecto.Changeset do
         {changeset, false}
       end
 
-    on_cast  = Keyword.get_lazy(opts, :with, fn -> on_cast_default(type, related) end)
+    on_cast  = Keyword.get_lazy(opts, :with, fn -> Relation.on_cast_default(type, related) end)
     original = Map.get(data, key)
 
     changeset =
@@ -814,7 +816,8 @@ defmodule Ecto.Changeset do
               changeset = %{force_update(changeset, opts) | changes: changes, valid?: valid?}
               missing_relation(changeset, key, current, required?, relation, opts)
 
-            {:error, {message, meta}} ->
+            {:error, message_and_meta} ->
+              {message, meta} = Keyword.pop(message_and_meta, :message)
               meta = [validation: type] ++ meta
               error = {key, {message(opts, :invalid_message, message), meta}}
               %{changeset | errors: [error | changeset.errors], valid?: false}
@@ -830,32 +833,6 @@ defmodule Ecto.Changeset do
 
     update_in changeset.types[key], fn {type, relation} ->
       {type, %{relation | on_cast: on_cast}}
-    end
-  end
-
-  def on_cast_default(type, module) do
-    fn struct, params ->
-      try do
-        module.changeset(struct, params)
-      rescue
-        e in UndefinedFunctionError ->
-          case __STACKTRACE__ do
-            [{^module, :changeset, args_or_arity, _}] when args_or_arity == 2
-                                                      when length(args_or_arity) == 2 ->
-              raise ArgumentError, """
-              the module #{inspect module} does not define a changeset/2 function,
-              which is used by cast_#{type}/3. You need to either:
-
-                1. implement the #{type}.changeset/2 function
-                2. pass the :with option to cast_#{type}/3 with an anonymous
-                   function that expects 2 args or an MFA tuple
-
-              When using an inline embed, the :with option must be given
-              """
-            stacktrace ->
-              reraise e, stacktrace
-          end
-      end
     end
   end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -534,8 +534,8 @@ defmodule Ecto.Changeset do
 
   defp type!(types, key) do
     case types do
-      %{^key => {tag, _}} when tag in @relations ->
-        raise "casting #{tag}s with cast/4 for #{inspect key} field is not supported, use cast_#{tag}/3 instead"
+      %{^key => {:assoc, _}} ->
+        raise "casting assocs with cast/4 for #{inspect key} field is not supported, use cast_assoc/3 instead"
       %{^key => type} ->
         type
       _ ->
@@ -556,13 +556,23 @@ defmodule Ecto.Changeset do
     case params do
       %{^param_key => value} ->
         value = if value in empty_values, do: Map.get(defaults, key), else: value
-        case Ecto.Type.cast(type, value) do
+        case Ecto.Type.cast(type, value, current) do
           {:ok, value} ->
             if Ecto.Type.equal?(type, current, value) do
               :missing
             else
               {:ok, value, valid?}
             end
+
+          {:ok, value, valid2?} ->
+            if Ecto.Type.equal?(type, current, value) do
+              :missing
+            else
+              {:ok, value, valid? and valid2?}
+            end
+
+          :ignore ->
+            :missing
 
           :error ->
             {:invalid, []}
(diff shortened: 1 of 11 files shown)


lib/ecto/type.ex
```diff
@@ -644,45 +644,48 @@ defmodule Ecto.Type do
...
-  @spec cast(t, term) :: {:ok, term} | {:error, keyword()} | :error
-  def cast({:parameterized, type, params}, value), do: type.cast(value, params)
-  def cast({:in, _type}, nil), do: :error
-  def cast(_type, nil), do: {:ok, nil}
-
-  def cast({:maybe, type}, value) do
+  @spec cast(t, term) :: {:ok, term} | {:ok, term, boolean} | {:error, keyword()} | :error
+  def cast(type, value, current \\ nil)
+  def cast({:parameterized, type, params}, value, current), do: type.cast_with_current(value, current, params)
+  def cast({:embed, embedded}, value, current), do: Ecto.Embedded.cast_with_current(value, current, embedded)
```
josevalim, line 650: This should be handled in `Ecto.Changeset`, as there is no such thing as `{:embed, embedded}` from the `Ecto.Type` perspective.

josevalim: For example, in `Ecto.Changeset.type!`, you can rewrite {:embed, embedded} into the parameterized one. :)

lib/ecto/schema.ex
```diff
@@ -1880,15 +1880,15 @@ defmodule Ecto.Schema do
     Module.put_attribute(mod, :changeset_fields, {name, {:assoc, struct}})
   end
 
-  @valid_embeds_one_options [:strategy, :on_replace, :source]
+  @valid_embeds_one_options [:strategy, :on_replace, :source, :on_cast]
```
josevalim, line 1883: I would prefer to not introduce this option. `cast_embed` by itself is not going away... we are just allowing it to be handled on regular cast too.

lib/ecto/type.ex
```diff
@@ -644,45 +644,48 @@ defmodule Ecto.Type do
...
 
   """
-  @spec cast(t, term) :: {:ok, term} | {:error, keyword()} | :error
-  def cast({:parameterized, type, params}, value), do: type.cast(value, params)
-  def cast({:in, _type}, nil), do: :error
-  def cast(_type, nil), do: {:ok, nil}
-
-  def cast({:maybe, type}, value) do
+  @spec cast(t, term) :: {:ok, term} | {:ok, term, boolean} | {:error, keyword()} | :error
+  def cast(type, value, current \\ nil)
```
josevalim, line 648: As I mentioned in the previous discussion, we need to introduce a new `cast` function. We cannot repurpose this one because we do not always have a `current` and assuming that `current` is nil will lead to the wrong behaviour in places like using `cast` to cast a query.

Here is a more concrete proposal: what if introduce a new function and call it `cast_change`? It can be a contrast to `put_change`, which @Eiji7 is working on.

lib/ecto/changeset.ex
```diff
@@ -823,7 +833,7 @@ defmodule Ecto.Changeset do
     end
   end
 
-  defp on_cast_default(type, module) do
+  def on_cast_default(type, module) do
```
josevalim, line 836: Move this function to Ecto.Changeset.Relation so we can share it from both modules. :)

lib/ecto/embedded.ex
```diff
@@ -121,6 +121,51 @@ defmodule Ecto.Embedded do
...
+    else
+      value
+    end
+    on_cast = case on_cast do
+      nil -> on_cast_default(related)
+      _ -> on_cast
+    end
+    case Relation.cast(embedded, nil, value, current, on_cast) do
+      {:ok, change, valid?} -> {:ok, change, valid?}
+      {:error, {message, _meta}} -> {:error, message: message}
```
josevalim, line 138:
```suggestion
      {:error, {message, _meta}} -> {:error, [message: message] ++ meta}
```

In fact, I would even change `Relation.cast` to return `{:error, [message: message] ++ meta}` directly and then we do `Keyword.pop` in `cast_relation` to get the message.

lib/ecto/changeset.ex
```diff
@@ -556,14 +556,24 @@ defmodule Ecto.Changeset do
...
+        case Ecto.Type.cast(type, value, current) do
           {:ok, value} ->
             if Ecto.Type.equal?(type, current, value) do
               :missing
             else
               {:ok, value, valid?}
             end
 
+          {:ok, value, valid2?} ->
+            if Ecto.Type.equal?(type, current, value) do
```
josevalim, line 568: I was ideally thinking we get rid of `equal?` and have it return `:ignore` if it is the same but that would mean adding a branch to `Ecto.Changeset` only to handle parameterized types... which is not ideal. 🤔

josevalim:
Thank you @TheFirstAvenger for the PR! I will be very honest and say I am not completely sure we should merge this functionality. I have pointed out some improvements and some open questions... but perhaps, given cast_embed and cast_assoc are already their own function, maybe it makes sense to keep their own functions after all. And if someone wants to provide especially casting behaviour too, they can introduce their own `cast_whatever`.

My suggestion for now is to let this one sit for now while we explore other parts of the puzzle. We still have apply_changes to explore and a couple others from schema. Thanks! :heart:

TheFirstAvenger:
> My suggestion for now is to let this one sit for now while we explore other parts of the puzzle. We still have apply_changes to explore and a couple others from schema. Thanks! ❤️

@josevalim Sounds like a plan. I will work through the suggestions and then we can let it sit for now. I really would love to see embeds work seamlessly in cast, because I know it is a source of much confusion and frustration with developers (especially newer ones) having to figure out, code for, and maintain cast_embed logic in changesets.

lib/ecto/type.ex
Eiji7: @josevalim Whatever you decide. However here are my 2 cents:

`cast_change` definitely would be confusing as `cast` was previously for `validation` and `change` (per documentation) for internal application data only.

> Note the value is directly stored in the changeset with no validation whatsoever. For this reason, this function is meant for working with data internal to the application.

Also I wonder if we even need `change/2` callback and simply add support for `parameterized` inside `Ecto.Changeset.change/2`, so the only callback parameterized types needs to implement would be `cast`.

Thoughts?

josevalim: The issue is that cast and change for embeds behave very differently. Furthermore, cast with current and without current also behave differently. So we do need all of them as distinct things, the question is which ones we want to expose.

Eiji7: Hmm … `cast` for me is obvious, but as `change` documentation says we should just put new data without any validation (so I suppose also no `raise` for `:on_replace` option etc.) as this is for internal application data only, so developer should know what he/she is changing. If I would like to let's say update (in context of `embeds`) I would simply use `Map.merge/2` before even calling `change`. Of course that's a breaking change, but if we are going with "new" `Ecto.Type` (i.e. `Ecto.ParameterizedType`) we can always add something like `Ecto.ParameterizedEmbed` just to make things simpler if current implementation is too complicated …

josevalim: If we don't want to support embeds (or we want to special case anyway), then I would just rather not do any of this, since the whole point of adding these new callbacks is to unify embeds with parameterized types.

Eiji7: @josevalim ok, now after I look at new code I see how `cast_change` is going work … I can see that 

I only wonder why `cast` is always raising on `cast/3` :thinking: 

```elixir
  @impl Ecto.ParameterizedType
  def cast(_, %{field: field, owner: owner}) do
    raise "cannot cast embed `#{field}` from `#{owner}` in queries"
  end
```

Should I wait for this PR and do same for `change` or maybe I'm just confused with `cast_change` name and `change` itself should do something else?

josevalim: > change itself should do something else?

`change` does something else, so no worries!

lib/ecto/changeset.ex
```diff
@@ -1902,6 +1888,7 @@ defmodule Ecto.Changeset do
       value when is_binary(value) and trim -> String.trim_leading(value) == ""
       value when is_binary(value) -> value == ""
       nil -> true
+      [] -> true
```
josevalim, line 1891: I am worried about the implication of this change for the general usage of `{:array, any_type}`. 🤔

TheFirstAvenger: Agreed, if I am reading it correctly, there is a difference in functionality between "required: true" on `cast_embed` (which fails on `[]`) and `validate_required` on an `{:array, any_type}`, which does not fail if it is a `[]`. If I remove the `[] -> true` clause, [this line fails](https://github.com/elixir-ecto/ecto/blob/master/test/ecto/changeset/embedded_test.exs#L467)

josevalim:
Closing this per #3380 but in case we get back to it, we will have a starting point. :) Thanks a lot! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #3385, opened 2020-08-14
Add redacted: true to field macro. Changeset inspect hides these fields

### LINKED ISSUE #3307: Add config :ecto, :filter_changes, [:password]
Those changes will be filtered whenever the changeset is inspected or an error that shows fields is raised. The default will be the password field given above.

TheFirstAvenger:
First pass at a solution for #3307. Adds `redacted: true` to field macro. When present, field value is set to `**redacted**` in changes inside Changeset inspect. `:password` is automatically redacted unless overridden with `redacted: false`.

Fixes #3307

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -255,6 +255,11 @@ defmodule Ecto.Changeset do
     * `filters`
     * `prepare`
 
+  ### Redacting fields in inspect
+
+  To hide a fields value from the inspect protocol of `Ecto.Changeset`, mark the field as `redacted: true`
+  in the schema, and it will display with the value `**redacted**`.
+
   """
 
   require Ecto.Query
@@ -601,12 +606,12 @@ defmodule Ecto.Changeset do
       |> Enum.reduce(nil, fn
         {key, _value}, nil when is_binary(key) ->
           nil
-  
+
         {key, _value}, _ when is_binary(key) ->
           raise Ecto.CastError, type: :map, value: params,
                                 message: "expected params to be a map with atoms or string keys, " <>
                                          "got a map with mixed keys: #{inspect params}"
-  
+
         {key, value}, nil when is_atom(key) ->
           [{Atom.to_string(key), value}]
 
@@ -2929,14 +2934,19 @@ end
 defimpl Inspect, for: Ecto.Changeset do
   import Inspect.Algebra
 
-  def inspect(changeset, opts) do
+  def inspect(%Ecto.Changeset{data: data} = changeset, opts) do
     list = for attr <- [:action, :changes, :errors, :data, :valid?] do
       {attr, Map.get(changeset, attr)}
     end
 
+    redacted_fields = case data do
+      %type{} -> type.__schema__(:redacted_fields)
+      _ -> []
+    end
+
     container_doc("#Ecto.Changeset<", list, ">", opts, fn
       {:action, action}, opts   -> concat("action: ", to_doc(action, opts))
-      {:changes, changes}, opts -> concat("changes: ", to_doc(changes, opts))
+      {:changes, changes}, opts -> concat("changes: ", changes |> filter(redacted_fields) |> to_doc(opts))
       {:data, data}, _opts      -> concat("data: ", to_struct(data, opts))
       {:errors, errors}, opts   -> concat("errors: ", to_doc(errors, opts))
       {:valid?, valid?}, opts   -> concat("valid?: ", to_doc(valid?, opts))
@@ -2945,4 +2955,14 @@ defimpl Inspect, for: Ecto.Changeset do
 
   defp to_struct(%{__struct__: struct}, _opts), do: "#" <> Kernel.inspect(struct) <> "<>"
   defp to_struct(other, opts), do: to_doc(other, opts)
+
+  defp filter(changes, redacted_fields) do
+    Enum.reduce(redacted_fields, changes, fn redacted_field, changes ->
+      if Map.has_key?(changes, redacted_field) do
+        Map.put(changes, redacted_field, "**redacted**")
+      else
+        changes
+      end
+    end)
+  end
 end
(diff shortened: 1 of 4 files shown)


Eiji7:
@josevalim Do we really want to do this in `ecto`? Look that for every library like `ecto` and even for every `struct` with such data we would need to basically copy-paste inspect logic which for me does not makes sense.

How about adding it in `elixir` like now we have `@enforce_keys`, so for example:
```elixir
defmodule MyApp.MyContext.MySchema do
  use Ecto.Schema

  @redacted_fields [:password, :password_repeat]

  schema "…" do
    # …
  end
end

defmodule MyApp.MyStruct do
  @redacted_fields […]
  defstruct […]
end

defmodule MyApp.MyCustomSchema do
  use MyLib.Schema

  @redacted_fields […]

  my_custom_schema_macro(…)
end
```

josevalim:
@TheFirstAvenger  this looks great but I would prefer if we did not special case the password field. This may give the impression that it somehow magically works but in truth as soon as they call the field something else, we will run into issues. 

@Eiji7 Yes, we want to do this in Ecto. We should make it as simple as possible for you to not leak this information - if it was only inspects it would be easy, but we need to consider changesets too. So it should be as close to one-liners as possible.

lib/ecto/changeset.ex
```diff
@@ -2929,7 +2929,11 @@ end
 defimpl Inspect, for: Ecto.Changeset do
   import Inspect.Algebra
 
-  def inspect(changeset, opts) do
+  def inspect(%Ecto.Changeset{changes: changes, data: data} = changeset, opts) do
+    changeset = case data do
+      %type{} -> %Ecto.Changeset{changeset | changes: filter(changes, type.__schema__(:redacted_fields))}
+      _ -> changeset
+    end
```
josevalim, line 2936: Instead of applying the filter here, maybe we can apply it a couple lines below where we match on {:changes, changes}?

lib/ecto/schema.ex
```diff
@@ -552,6 +555,11 @@ defmodule Ecto.Schema do
         assocs = @ecto_assocs |> Enum.reverse
         embeds = @ecto_embeds |> Enum.reverse
         loaded = Ecto.Schema.__loaded__(__MODULE__, @struct_fields)
+        redacted_fields = if @ecto_redact_password do
+          [:password | @ecto_redacted_fields || []]
```
josevalim, line 559: `@ecto_redacted_fields` should always return an empty list since it is marked as accumulate, so you can skip the ` || []` parts. :)

lib/ecto/changeset.ex
Eiji7: We want to redact `data` as well, right?

josevalim: `data` will be redacted as part of the inspect protocol though. I will look into a separate PR that automatically derives inspect based on redacted if one was not derived yet.

TheFirstAvenger: Yes, I went through all the work to write tests and redact values in data also, and then the tests verifying that it didn't redact when not set failed because the password still didn't show 😂

josevalim:
This looks great! :heart: I was also thinking we could automatically derive inspect. Here is the idea:

1. If any of the fields are redacted
2. and we are not deriving inspect (i.e. ` List.keymember?(@derive, Inspect, 0)` is false)
3. then we will call `@derive {Inspect, except: redacted_fields}`

We should also have an option on `schema` to disable this behaviour, such as `schema "foo", derive_inspect_for_redacted_fields: false do`.

We can do this as part of this PR or a separate one. WDYT?

TheFirstAvenger:
@josevalim TIL deriving inspect 😁 . I pushed what I think should work in https://github.com/elixir-ecto/ecto/pull/3385/commits/66acd2105f7ca2f4527c891e8083e82d377a8578, however the value still appears in inspect, so the test fails.

lib/ecto/schema.ex
```diff
@@ -584,6 +599,10 @@ defmodule Ecto.Schema do
             {args, body} <- clauses do
           def __schema__(unquote_splicing(args)), do: unquote(body)
         end
+
+        if !List.keymember?(@derive, Inspect, 0) and @derive_inspect_for_redacted_fields and @ecto_redacted_fields != [] do
+          @derive {Inspect, except: @ecto_redacted_fields}
+        end
```
josevalim, line 605: You need to call this before we call `defstruct`. :)

lib/ecto/schema.ex
```diff
@@ -472,22 +483,23 @@ defmodule Ecto.Schema do
...
   @doc """
   Defines a schema struct with a source name and field definitions.
 
   An additional field called `__meta__` is added to the struct for storing
   internal Ecto state. This field always has a `Ecto.Schema.Metadata` struct
   as value and can be manipulated with the `Ecto.put_meta/2` function.
   """
-  defmacro schema(source, [do: block]) do
-    schema(__CALLER__, source, true, :id, block)
+  defmacro schema(source, opts \\ [], [do: block]) do
```
josevalim, line 496: Oops... I forgot we do not accept any options today. What if instead of an option we use a module attribute then? `@ecto_derive_inspect_for_redacted_fields` which defaults to true? Sorry for misleading...

TheFirstAvenger:
@josevalim All set. Added a note about derive being 1.8+ and excluded it from tests on lower versions.

josevalim:
> All set. Added a note about derive being 1.8+ and excluded it from tests on lower versions.

Ah, good call. Good opportunity to require Elixir v1.8+ then. 👍

josevalim:

kipcole9:
Probably in the "nitpick" category, but I think it would be more correct to say `redact: true` rather than `redacted: true`. The difference is that `redact: true` say "apply the redact action on this field" whereas `redacted: true` implies the data is already redacted which could lead to the mistaken belief that the data stored in the schema struct is itself redacted.

PR #3388, opened 2020-08-25
Update repo.ex

ktec:
Reduce the emotive language used to describe the return behaviour of `c:insert_all/3` when using the `:on_conflict` option.


josevalim:

PR #3390, opened 2020-08-28
Parse decimal floats with Decimal.from_float instead of Decimal.new

alfredbaudisch:
- v2.2.* currently uses `Decimal.new/1` to cast and dump all decimal numbers, including float, which triggers a deprecation warning:
```
 warning: passing float to Decimal.new/1 is deprecated as floats have inherent inaccuracy. Use Decimal.from_float/1 or Decimal.cast/1 instead
```
- v3.* already uses `Decimal.from_float/1`
- This pull request adds the usage of `Decimal.from_float/1` to v2.2.*

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -319,6 +319,10 @@ defmodule Ecto.Type do
     end
   end
 
+  def dump(:decimal, term, _dumper) when is_float(term) do
+    {:ok, Decimal.from_float(term)}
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @alfredbaudisch, we no longer maintain the v2.2 branch, except for security patches, so we don’t be merging these changes. We recommend everyone to upgrade. Thanks!

PR #3398, opened 2020-09-02
Add documentation for dynamic binding selection

LostKobrakai:


lib/ecto/query.ex
```diff
@@ -236,6 +236,22 @@ defmodule Ecto.Query do
...
+      
+  ### Dynamic binding selection
+  
+  When composing functions a common requirement is selecting a binding 
+  on the query. Previous sections showed means of hardcoding that selection,
+  but there are also ways of doing it dynamically.
+  
+      # Knowing the position of the binding
+      def sort(query, position, field) do
+        from [{x, position}] in query, order_by: field(x, ^field)
```
josevalim, line 248: I believe this API is private!

LostKobrakai: I was wondering about that as the test I added when adding the named equivalent seems to no longer exist. I'm wondering about the implications of making it public though. Especially for query builders tracking the position might be the simpler option than needing to create names for bindings just to be able to refer to them again later.

josevalim: We have added the names exactly so we no longer need to depend on positional bindings. If there is a use case that can not be done with names, then we can discuss approaches, otherwise i would prefer to keep the API private for now. :)

LostKobrakai: Sounds fine. I've reduced this pr to only add to the named binding section.

josevalim:

PR #3402, opened 2020-09-08
Add Ecto.Enum.values/2 helper

TheFirstAvenger:
Add Ecto.Enum.values/2 helper. This provides a public way to access the values defined for an Ecto.Enum field without the end user having to parse through the schema field type parameters. Allows simple DRY access to the list of values, preventing duplication for things like Phoenix field drop-downs

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -13,6 +13,24 @@ defmodule Ecto.Enum do
   and only if the atom exists in the list (otherwise an error will be raised).
   Attempting to load any string not represented by an atom in the list will be
   invalid.
+
+  The helper function `values/2` returns the values for a given schema and
+  field, which can be used in places like form drop-downs. For example,
+  given the following schema:
+
+      defmodule EnumSchema do
+        use Ecto.Schema
+
+        schema "my_schema" do
+          field :my_enum, Ecto.Enum, values: [:foo, :bar, :baz]
+        end
+      end
+
+  you can call `values/2` like this:
+
+      > Ecto.Enum.values(EnumSchema, :my_enum)
+      [:foo, :bar, :baz]
+
   """
 
   use Ecto.ParameterizedType
@@ -34,7 +52,7 @@ defmodule Ecto.Enum do
 
     on_load = Map.new(values, &{Atom.to_string(&1), &1})
     on_dump = Map.new(values, &{&1, Atom.to_string(&1)})
-    %{on_load: on_load, on_dump: on_dump}
+    %{on_load: on_load, on_dump: on_dump, values: values}
   end
 
   @impl true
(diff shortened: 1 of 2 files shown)


lib/ecto/enum.ex
```diff
@@ -71,4 +89,16 @@ defmodule Ecto.Enum do
 
   @impl true
   def embed_as(_, _), do: :self
+
+  def values(schema, field) do
+    try do
+      case schema.__schema__(:type, field) do
+        {:parameterized, Ecto.Enum, %{values: values}} -> values
+        {_, {:parameterized, Ecto.Enum, %{values: values}}} -> values
+        nil -> raise ArgumentError, "#{field} is not an Ecto.Enum field"
```
josevalim, line 98: This will be rescued and reraised. Seee suggestion below on a different way to write this!

lib/ecto/enum.ex
```diff
@@ -71,4 +89,16 @@ defmodule Ecto.Enum do
...
+      end
+    rescue
+      _ in UndefinedFunctionError -> raise ArgumentError, "#{inspect schema} is not an Ecto schema"
+    end
```
josevalim, line 102:
```suggestion
    try do
      schema.__schema__(:type, field)
    rescue
      _ in UndefinedFunctionError -> raise ArgumentError, "#{inspect schema} is not an Ecto schema"
    else
      {:parameterized, Ecto.Enum, %{values: values}} -> values
      {_, {:parameterized, Ecto.Enum, %{values: values}}} -> values
      nil -> raise ArgumentError, "#{field} is not an Ecto.Enum field"
    end
```

josevalim:
Thanks @TheFirstAvenger! I have dropped a comment.

I have one question about the API: I can see someone asking in the future a way to return the dump values. How would you call this extra function? `dump_values/2`? We don't need too add an API for it now, I just want to make sure we are not painting ourselves into a corner right now. :)

lib/ecto/enum.ex
```diff
@@ -13,6 +13,24 @@ defmodule Ecto.Enum do
   and only if the atom exists in the list (otherwise an error will be raised).
   Attempting to load any string not represented by an atom in the list will be
   invalid.
+
+  The helper function `values/2` exists to get the allowed values for a given
+  schema and field, allowing you to DRY the list in things like form drop-downs
+  in Phoenix. For example, given the following schema:
```
josevalim, line 19:
```suggestion
  The helper function `values/2` returns the values for a given schema and
  field, which can be used in places such as form drop-downs. For example,
  given the following schema:
```

TheFirstAvenger:
@josevalim An earlier version of the PR had values/2 and values_as_strings/2. I can add that second one in too.

josevalim:
@TheFirstAvenger i want to support values as integers in the future too so that's why i proposed `dump_values` but we can definitely skip them for now. Sounds good?

TheFirstAvenger:
@josevalim that makes sense. My initial thought is that from a user perspective, `dump` as a phrase might not connect with users as well as some other term indicating that these are the values as they are stored in the database. I also think a `values_as_*` format would be helpful if we can pull it off so it appears right next to `values` in the documentation and auto-complete, increasing awareness.

On another note, one thought I had which might have implications in ParameterizedType was this:

If one uses of this values function is in something like a Phoenix view, it seems a bit off to me to reference  `Ecto.Enum` in the web layer template. It would be cleaner in my mind for the values function to be accessible on the schema itself. This had me thinking about modifying the field/schema macros to inject a function on the schema module itself to access these values (e.g. something like `User.enum_values(:role)`). Not wanting to give Ecto.Enum special treatment, I thought that this might be one of a number of possible future use cases for parameterized types being able to inject helper functions on the schema module. Not sure how it would work or even if it is a good idea yet, but figured I'd mention the idea to see if there was something there.

Either way, this PR is squashed and good to go from my end.

josevalim:

PR #3404, opened 2020-09-10
Fix empty string converted to non-empty default

balexand:
I came across the following surprising behavior in Ecto. My schema looked like:

```elixir
defmodule MyApp.Redirect do
  use Ecto.Schema
  import Ecto.Changeset

  schema "redirects" do
    field :path, :string, default: "/"
  end

  @doc false
  def changeset(redirect, attrs) do
    redirect
    |> cast(attrs, [:path])
    |> validate_required([:path])
  end
end
```

The following worked as expected:

```elixir
changeset = MyApp.Redirect.changeset(%MyApp.Redirect{}, %{path: nil})
refute changeset.valid?
assert changeset.errors == [path: {"can't be blank", [validation: :required]}]
```

But mysteriously, `cast` was not setting the field when the value was an empty string:

```elixir
changeset = MyApp.Redirect.changeset(%MyApp.Redirect{}, %{path: ""})
assert changeset.valid?
assert Ecto.Changeset.get_field(changeset, :path) == "/"
```

It appears that Ecto is converting values from `changeset.empty_values` to the field's default value. But in this case, the field's default value does not represent an empty value. Having a default value that does not represent "empty" seems like a common use case.

This pull-request ensures that the conversion from empty value to default value will only occur if the default value is an empty value.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -105,11 +105,13 @@ defmodule Ecto.Changeset do
   Many times, the data given on cast needs to be further pruned, specially
   regarding empty values. For example, if you are gathering data to be
   cast from the command line or through an HTML form or any other text-based
-  format, it is likely those means cannot express nil values. For
+  format, it is likely those means cannot express `nil` values. For
   those reasons, changesets include the concept of empty values, which are
-  values that will be automatically converted to the field's default value
-  on `cast/4`. Those values are stored in the changeset `empty_values` field
-  and default to `[""]`.
+  stored in the changeset `empty_values` field with a default of `[""]`. Empty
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @balexand, unfortunately I am not sure we can consider this to be a general assumption and I can see it being a breaking change for some folks. It also makes the logic more complex, it is easier to say "those values are always considered empty" rather than saying "those values are always considered empty as long as they are a default, which changes per field".

My suggestion is for you to call `cast` twice, each time with different value for empty_values:

    |> cast(params, [:path], empty_values: [])
    |> cast(params, [:path], empty_values: [""])

Thank you!

PR #3410, opened 2020-09-13
Add docs for Ecto.Type on it's usage with preloads

doorgan:
I recently had an issue when using a custom Ecto.Type as primary key and trying to preload associations.
I posted the details [in this thread](https://elixirforum.com/t/ecto-custom-type-issue-value-list-of-ids-in-where-cannot-be-cast-to-type-in-id-in-query/34325) of the elixir forums while I was trying to figure it out what was the cause.

The issue was that, when using a custom Ecto.Type as primary key, it needs to be used also in the `:type` option of `Ecto.Schema.belongs_to`, otherwise Ecto will fail to cast the list of ids and will fail with an error that's not very obvious:

```elixir
[error] Task #PID<0.11545.0> started from #PID<0.11541.0> terminating
** (Ecto.Query.CastError) deps/ecto/lib/ecto/association.ex:623: value `["KV3A8", "Ja344", "8n43V", "304Ko"]` in `where` cannot be cast to type {:in, :id} in query:

from r0 in Embers.Reactions.Reaction,
  where: r0.post_id in ^["KV3A8", "Ja344", "8n43V", "304Ko"],
  order_by: [asc: r0.post_id],
  select: {r0.post_id, r0}
```

This wasn't obvious to me, mainly because documentation on custom type definition and usage is scattered between the `Ecto.Type` docs and the `Ecto.Schema` module attributes and `Ecto.Schema.belongs_to` options and it can be hard to figure out how everything fits together.

This PR adds a short example of this kind of usage to the `Ecto.Type` docs to make the relationship a little bit clearer.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -80,6 +80,79 @@ defmodule Ecto.Type do
 
   Note: `nil` values are always bypassed and cannot be handled by
   custom types.
+  
+  ## Custom types and preloads
+  
+  Imagine the id of your schema is an auto incrementing number.
+  Since it's auto incrementing nature could allow them to enumerate
+  the contents of your service, you don't may not want your
+  users to know it, so you send an encoded version of the id to
+  them.
+  
+  An Ecto type could handle the conversion between the encoded
+  version of the id and it's representation in the database.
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3411, opened 2020-09-14
Adds support for map on update_all

### LINKED ISSUE #3391: Repo.update_all should accept a map
At present `Repo.insert_all` accept either a map or a keyword list, for example, we can do:

```elixir
attrs = %{"col" => "data"}
Repo.insert_all("some_table", [attrs])
```

But, we cannot do:

```elixir
attrs = %{"col" => "newdata"}
id = 1

from(row in "some_table", where: row.id == ^id) 
|> Repo.update_all(set: attrs)
```

This is enforced by https://github.com/elixir-ecto/ecto/blob/a33f79fb577ac765f5b6c70b7f4be9fab94c9ebb/lib/ecto/query/builder/update.ex#L90

This is an issue with dynamic table mapping. I have an app with dynamic tables where each user can create its own schema, and I don't want to convert all user defined column name to atoms. This is a minor issue, but I think it's good hygiene to not `to_atom` user data.

v0idpwn:
Closes #3391

### DIFF
--- a/test/ecto/query/builder/update_test.exs
+++ b/test/ecto/query/builder/update_test.exs
@@ -19,9 +19,19 @@ defmodule Ecto.Query.Builder.UpdateTest do
       [compile] = query.updates
       assert compile.expr == [set: [foo: {:==, [], [{{:., [], [{:&, [], [0]}, :foo]}, [], []}, {:^, [], [0]}]}]]
       assert compile.params == [{1, {0, :foo}}]
+
+      query = "foo" |> update([p], set: %{foo: p.foo == ^1})
+      [compile] = query.updates
+      assert compile.expr == [set: [foo: {:==, [], [{{:., [], [{:&, [], [0]}, :foo]}, [], []}, {:^, [], [0]}]}]]
+      assert compile.params == [{1, {0, :foo}}]
+
+      query = "foo" |> update([p], set: %{"foo" => p.foo == ^1})
+      [compile] = query.updates
+      assert compile.expr == [set: [{"foo", {:==, [], [{{:., [], [{:&, [], [0]}, :foo]}, [], []}, {:^, [], [0]}]}}]]
+      assert compile.params == [{1, {0, :foo}}]
     end
 
-    test "raises on non-keyword lists" do
+    test "raises on non keyword/string-key lists" do
       assert_raise Ecto.Query.CompileError,
                    ~r"malformed update `\[1\]` in query expression", fn ->
         escape(quote do [1] end, [], __ENV__)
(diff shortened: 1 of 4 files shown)


v0idpwn:
Do you think this can/should work with compile time interpolation, @josevalim? I ported for runtime only because I don't see string keys being useful in compile time.

Eiji7 (approved):

v0idpwn:
This PR isn't actually done. The current approach won't cover:
- interpolation
- dynamic

I'm trying to finish to port to those cases and will follow up here.

v0idpwn:
Just finished, ready to review

test/ecto/query/builder/update_test.exs
```diff
@@ -19,9 +19,19 @@ defmodule Ecto.Query.Builder.UpdateTest do
...
+      assert compile.params == [{1, {0, :foo}}]
+
+      query = "foo" |> update([p], set: %{"foo" => p.foo == ^1})
+      [compile] = query.updates
+      assert compile.expr == [set: [{"foo", {:==, [], [{{:., [], [{:&, [], [0]}, :foo]}, [], []}, {:^, [], [0]}]}}]]
+      assert compile.params == [{1, {0, :foo}}]
     end
 
-    test "raises on non-keyword lists" do
+    test "raises on non keyword/string-key lists" do
```
Eiji7, line 34: I'm not sure here … First of all personally I would avoid making such change in case you are not updating anything else (here test body).

You are removing `-` character, but using it in other part. Maybe I'm wrong, but it looks like a small inconsistency (?) … Shouldn't it be `non keyword/string key` or `non-keyword/string-key` instead? Personally I would name it like: `non keyword or string key`.

Also I do not see a `string key` usage. There are only two checks:
1. If `List` which is not `Keyword`
2. If `:set` option is `List` which is not `Keyword`
Did you forgot to add an extra check here?

Eiji7 (approved):
> @v0idpwn Thanks for the review, @Eiji7, but this PR isn't actually done. The current approach won't cover:
> 
>     * interpolation
> 
>     * dynamic
> 
> 
> I'm trying to finish adapting to those cases and will follow up here.

Made review, because:
> (…) v0idpwn  marked this pull request as ready for review (…)

Except one comment everything looks really well. Big :heavy_plus_sign: for writing an useful enhancement with just few extra pattern-matching clauses. Your changes are written in same way as other parts of existing code which definitely makes checking changes clear.

josevalim:
Great first step! :heart:

v0idpwn:
Hi, @josevalim!
This works in all the cases tested, am I missing anything?

josevalim:
I am not sure if this is what you are asking but for the PR to be considered complete we would need to support the interpolation of maps with string keys. Supporting literal maps do not change much since in those scenarios you could pass atoms with no problems. :)

v0idpwn:
I believe it supports, but maybe I didn't write enough tests

josevalim:
I see, you are correct. We just need integration tests for “set: ^%{...}”. Also please check in the README the steps for running integration tests, as they do not run by default. This may require adapter changes (or not).

Finally, it seems we are converting maps to lists and having tuples of string keys. Maybe we should keep them as maps?

v0idpwn:
@josevalim I've tested on postgres only, I'll setup to test with the other adapters today and add tests for this case you said.

As for the lists vs maps, I ain't sure. I'm still learning about metaprogramming in Elixir, but I may try.

josevalim:
Testing only Postgres is fine as this is not a complex change.

Regarding the map issue? It is being converted to a list when you traverse it, so you need to convert it back to a map but you are correct it may cause issues on the meta programming side as we would need to escape its representation, so let’s keep it as is and we just add some dynamic tests to the integration suite.

v0idpwn:
Just added the dynamic test and it is passing just fine. Unfortunately, for now dynamic won't support something like

```elixir
dynamic([p], p["visits"] + 1)
```

Which makes it rather useless with maps. Are we planning to better support dynamic table mapping in the future?

Eiji7:
> Just added the dynamic test and it is passing just fine. Unfortunately, for now dynamic won't support something like
> 
> ```elixir
> dynamic([p], p["visits"] + 1)
> ```
> 
> Which makes it rather useless with maps. Are we planning to better support dynamic table mapping in the future?

Hmm … isn't that a mistake in your sample?
`p` in `p["visits"]` is a row where `"visits"` would be a column, but I don't remember such API. However we have [Ecto.Query.field/2](https://hexdocs.pm/ecto/Ecto.Query.API.html#field/2), so it would look like:
```elixir
dynamic([p], field(p, ^column) + 1)
```

You wrote it in context of `maps` (`jsonb` support), so I guess you forgot about [Ecto.Query.API.json_extract_path/2](https://hexdocs.pm/ecto/Ecto.Query.API.html#json_extract_path/2) …

> Returns value from the json_field pointed to by path.
> ```elixir
> from(post in Post, select: json_extract_path(post.meta, ["author", "name"]))
> ```
> The query can be also rewritten as:
> ```elixir
> from(post in Post, select: post.meta["author"]["name"])
> ```
> Path elements can be integers to access values in JSON arrays:
> ```elixir
> from(post in Post, select: post.meta["tags"][0]["name"])
> ```
> Any element of the path can be dynamic:
> ```elixir
> field = "name"
> from(post in Post, select: post.meta["author"][^field])
> ```

So in your case it would be something like:
```elixir
# missing stats (or similar) column
dynamic([p], p.stats["visits"] + 1)
```

josevalim:
Sorry, there has been some confusion. By `dynamic` I do not mean the `dynamic(...)` API, but rather interpolating the map completely: `set: ^%{....}`.

In any case, I am glad you did explore `dynamic` because you bring a good point: there is no way to refer to existing columns as the column syntax expects atoms. I was already on the fence but now I am really thinking the original reporter should most likely issue regular SQL commands to the database.

v0idpwn:
I think that, at this point, I agree. Wanna close this? Maybe support maps with atom keys only? (seems more coherent with the ecto api)

josevalim:
Yes, let’s close this. Maps themselves without binary keys do not offer any benefit, so let’s not change it. Thanks a lot for your PR though!

Issue #3412, opened 2020-09-15
Querying a Ecto.Enum without a pin operator in the last rc


sabiwara:
### Environment

* Elixir version (elixir -v): 1.10.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 12.1
* Ecto version (mix deps): `3.5.0-rc.0`
```
      {:phoenix_ecto, "~> 4.1", override: true},
      {:ecto, "~> 3.5.0-rc.0", override: true},
      {:ecto_sql, "~> 3.5.0-rc.0", override: true},
      {:postgrex, ">= 0.0.0"},
```
* Operating system: Ubuntu 18.04

### Current behavior

Reproduced a mininal example:

Schema:
```elixir
  schema "my_resources" do
    field :barbaz, Ecto.Enum, values: [:bar, :baz]

    timestamps()
  end
```

Failing example 1:

```elixir
Ecto.Query.where(MyResource, barbaz: :bar)
```

fails with a compilation error:
```
== Compilation error in file lib/foo/my_context/my_resource.ex ==
** (Ecto.Query.CompileError) `:bar` is not a valid query expression
    (ecto 3.5.0-rc.0) expanding macro: Ecto.Query.where/2
    (foo 0.1.0) lib/foo/my_context/my_resource.ex:26: Foo.MyContext.MyResource.atom_query/0
    (elixir 1.10.3) lib/kernel/parallel_compiler.ex:304: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/7
```

Failing example 2:

```elixir
Ecto.Query.where(MyResource, barbaz: "bar")
```
Compiles but fails at runtime (which is understandable) with:
```
** (Ecto.QueryError) lib/foo/my_context/my_resource.ex:22: value `"bar"` cannot be dumped to type {:parameterized, Ecto.Enum, %{on_dump: %{bar: "bar", baz: "baz"}, on_load: %{"bar" => :bar, "baz" => :baz}}}. Or the value is incompatible or it must be interpolated (using ^) so it may be cast accordingly in query:
```

Working example (pinning the atom):

```elixir
Ecto.Query.where(MyResource, barbaz: ^:bar)
```

### Expected behavior

Either one of the following would be great:
- modify the `where` macro to be able to query directly with an atom, given it is now a legit value (example 1. above)
- if not possible, maybe complete `Ecto.Enum`'s documentation with examples how they should be queried?

[...]

PR #3414, opened 2020-09-17
Allow querying with atoms in Ecto.Enum

sabiwara:
Solves #3412

I think I've got the basic scenarios working as expected.
I'm still getting familiar with the codebase though, so there are probably missing pieces and edge cases I haven't covered.

Thank you for letting me try it!

PS: atoms are a bit tricky because they are used for built-in like booleans and nil as well, so I'm a bit afraid of a potential unintended side effect.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -267,6 +267,12 @@ defmodule Ecto.Query.Builder do
   def escape(nil, _type, params_acc, _vars, _env),
     do: {nil, params_acc}
 
+  def escape(atom, type, {params, acc}, vars, _env) when is_atom(atom) do
+    expr = literal(atom, type, vars)
+    params = [{atom, type} | params]
+    {expr, {params, acc}}
+  end
+
   # comparison operators
   def escape({comp_op, _, [left, right]} = expr, type, params_acc, vars, env)
       when comp_op in ~w(== != < > <= >=)a do
@@ -1013,6 +1019,7 @@ defmodule Ecto.Query.Builder do
   def quoted_type(literal, _vars) when is_float(literal),   do: :float
   def quoted_type(literal, _vars) when is_binary(literal),  do: :string
   def quoted_type(literal, _vars) when is_boolean(literal), do: :boolean
+  def quoted_type(literal, _vars) when is_atom(literal) and not is_nil(literal), do: :atom
   def quoted_type(literal, _vars) when is_integer(literal), do: :integer
 
   # Tuples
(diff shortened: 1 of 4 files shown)


lib/ecto/query/builder.ex
```diff
@@ -267,6 +267,12 @@ defmodule Ecto.Query.Builder do
   def escape(nil, _type, params_acc, _vars, _env),
     do: {nil, params_acc}
 
+  def escape(atom, type, {params, acc}, vars, _env) when is_atom(atom) do
+    expr = literal(atom, type, vars)
+    params = [{atom, type} | params]
+    {expr, {params, acc}}
+  end
```
josevalim, line 274: Hi @sabiwara!

I don't think we should convert it to a parameter, because parameters are passed as interpolation. I believe you can rewrite this code as:

```suggestion
  def escape(atom, type, params_acc, vars, _env) when is_atom(atom) do
    {atom, params_acc}
  end
```

Have you tried this?

test/ecto/query/planner_test.exs
```diff
@@ -251,6 +252,35 @@ defmodule Ecto.Query.PlannerTest do
     assert params == []
   end
 
+  test "plan: casts atom values" do
+    {_query, params, _key} = plan(Post |> where([p], p.status == :draft))
+    assert params == ["draft"]
```
josevalim, line 257: So this example in particular should not have any parameter, instead, we will convert :draft into something directly in the query.

josevalim:
This is a great first step! I have added two comments, if for some reason doing the changes I mentioned are too complex, let me know and I will take a look at it!

lib/ecto/query/builder.ex
sabiwara: Hi @josevalim, thanks a lot for the review!

Yes this was actually my first attempt, but I wasn't able to have the following case raise without adding it as a param:
https://github.com/elixir-ecto/ecto/blob/6dfc2f512e462f292f84542ed5a4d49648f15534/test/ecto/query/planner_test.exs#L262

From what I understood, the `Builder` stage isn't able to detect a type mismatch yet, because it is unaware of the schema?
So I tried to understand how it was done for other similar errors, and ended up making it a param, so that `Planner` can check it when running `cast_and_merge_params`.

It might have been kind of a hack, telling Ecto to "treat `:foo` as you would have treated `^:foo`".
But this was exactly the behavior I had in mind: always cast atoms.

Any pointer for where to look without using params? I would like to to keep trying a bit if it is OK :slightly_smiling_face:

josevalim: If that approach doesn't work, then I don't know *why* it doesn't work, so I need to jump in and take a look. :) The fix may be simple but finding it may be much harder. :D

josevalim:
So it was a simply fix after error: the errors are raised on `normalize`, no `plan`. Changed that fixed it. I have pushed it to master, thank you!

sabiwara:
Thanks @josevalim ! :purple_heart: 
I totally missed `normalize`, this is a much more satisfying answer indeed.

PR #3416, opened 2020-09-20
Add Repo.reload/2 and Repo.reload!/2

### LINKED ISSUE #3389: Add Repo.reload and Repo.reload!
It receives a schema (or a list of schemas) and uses the meta information + primary key to reload the same entry from the database. The non bang variant returns nil (or removes the element from the list), the bang variant raises.

v0idpwn:
Closes #3389 
I'm still polishing, got to fix some things regarding code quality. Still, the API proposal is this. Do I need to change something?

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -338,6 +338,14 @@ defmodule Ecto.Repo do
           Ecto.Repo.Queryable.get_by!(get_dynamic_repo(), queryable, clauses, opts)
         end
 
+        def reload(queryable, opts \\ []) do
+          Ecto.Repo.Queryable.reload(get_dynamic_repo(), queryable, opts)
+        end
+
+        def reload!(queryable, opts \\ []) do
+          Ecto.Repo.Queryable.reload!(get_dynamic_repo(), queryable, opts)
+        end
+
         def one(queryable, opts \\ []) do
           Ecto.Repo.Queryable.one(get_dynamic_repo(), queryable, opts)
         end
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -669,19 +669,9 @@ defmodule Ecto.Repo do
   @doc """
   Reloads a given schema or schema list from the database.
 
-  When using with lists, ordering is guaranteed to be kept. Results not found in
+  When using with lists, it is expected that all of the structs in the list belong
+  to the same schema. Ordering is guaranteed to be kept. Results not found in
   the database will be returned as `nil`.
-  
-  ## Options
-
-    * `:prefix` - The prefix to run the query on (such as the schema path
-      in Postgres or the database in MySQL). This will be applied to all `from`
-      and `join`s in the query that did not have a prefix previously given
-      either via the `:prefix` option on `join`/`from` or via `@schema_prefix`
-      in the schema. For more information see the "Query Prefix" section of the
-      `Ecto.Query` documentation.
-
-  See the "Shared options" section at the module documentation for more options.
 
   ## Example
 
@@ -690,6 +680,9 @@ defmodule Ecto.Repo do
 
     MyRepo.reload([post1, post2])
     [%Post{}, %Post{}]
+
+    MyRepo.reload([deleted_post, post1])
+    [nil, %Post{}]
   """
   @callback reload(
               (schema :: Ecto.Schema.t()) | (schemas :: [Ecto.Schema.t()]),
@@ -698,18 +691,9 @@ defmodule Ecto.Repo do
 
 
   @doc """
-  Similar to `c:reload/2`, but raises when something is not found
+  Similar to `c:reload/2`, but raises when something is not found.
 
   When using with lists, ordering is guaranteed to be kept. 
-
-  ## Options
-
-    * `:prefix` - The prefix to run the query on (such as the schema path
-      in Postgres or the database in MySQL). This will be applied to all `from`
-      and `join`s in the query that did not have a prefix previously given
-      either via the `:prefix` option on `join`/`from` or via `@schema_prefix`
-      in the schema. For more information see the "Query Prefix" section of the
-      `Ecto.Query` documentation.
 
   ## Example
 
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -551,8 +551,9 @@ defmodule Ecto.Repo do
 
   ## Ecto.Adapter.Queryable
 
-  @optional_callbacks get: 3, get!: 3, get_by: 3, get_by!: 3, aggregate: 3, aggregate: 4, exists?: 2,
-                      one: 2, one!: 2, preload: 3, all: 2, stream: 2, update_all: 3, delete_all: 2
+  @optional_callbacks get: 3, get!: 3, get_by: 3, get_by!: 3, reload: 2, reload!: 2, aggregate: 3, 
+                      aggregate: 4, exists?: 2, one: 2, one!: 2, preload: 3, all: 2, stream: 2, 
+                      update_all: 3, delete_all: 2
 
   @doc """
   Fetches a single struct from the data store where the primary key matches the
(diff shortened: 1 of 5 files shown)


lib/ecto/repo/queryable.ex
```diff
@@ -428,6 +459,26 @@ defmodule Ecto.Repo.Queryable do
...
 
+  defp query_for_reload([head| _] = structs, opts) do
+    assert_structs!(structs)
+
+    schema = head.__struct__
+
+    case schema.__schema__(:primary_key) do
+      [pk] ->
+        keys = Enum.map(structs, &get_pk!(&1, pk))
+        preloads = Keyword.get(opts, :preload, [])
```
josevalim, line 470: Honestly, I don't think we need to support this option. Doing a `|> Repo.preload(...)` after is only very few characters more verbose than `, preload: `.

lib/ecto/repo/queryable.ex
```diff
@@ -74,6 +74,37 @@ defmodule Ecto.Repo.Queryable do
...
+    query = query_for_reload(structs, opts)
+    result = all(name, query, opts)
+    result_count = Enum.count(result)
+    expected_count = Enum.count(structs)
+
+    if result_count < expected_count do
+      raise Ecto.TooFewResultsError, queryable: query, min: expected_count, count: result_count
+    end
+
+    result
```
josevalim, line 95: Unfortunately there is no guarantee `result` will be in the same order as `structs`. So you need to do something like this:

    for struct <- structs do
      struct_pk = Map.fetch!(struct, pk)
      Enum.find(result, &Map.fetch!(&1, pk) == struct_pk) || raise "could not find"
    end

This means you don't need the Enum.count nor the exception above, as the above will guarantee all entries have a match.

lib/ecto/repo/queryable.ex
```diff
@@ -485,4 +536,28 @@ defmodule Ecto.Repo.Queryable do
...
       message: "expected a from expression with a schema"
   end
+
+  defp assert_structs!(structs) when is_list(structs) do
+    case Enum.reject(structs, &is_struct?/1) do
+      [] ->
+        :ok
+
+      rejected ->
+        raise ArgumentError, "Expected a struct or a list of structs, received #{inspect(rejected)}"
```
josevalim, line 546: Messages should start in lowercase.

lib/ecto/repo/queryable.ex
```diff
@@ -485,4 +536,28 @@ defmodule Ecto.Repo.Queryable do
...
+    end
+  end
+
+  defp is_struct?(%{__struct__: _, __meta__: _}), do: true
+  defp is_struct?(_), do: false
```
josevalim, line 551: We typically use `is_` or `?` but not both.  In this case, since it is not a guard, `?` works better. :)

```suggestion
  defp schema?(%{__struct__: _, __meta__: _}), do: true
  defp schema?(_), do: false
```

josevalim:
It is looking great! I have dropped some comments.

v0idpwn:
Thanks, @josevalim, I'll fix those ASAP :)

lib/ecto/repo/queryable.ex
qcam: If we guarantee same ordering in `reload!/2`, should we also do the same for `reload/2`?

lib/ecto/repo/queryable.ex
```diff
@@ -485,4 +536,28 @@ defmodule Ecto.Repo.Queryable do
       query: query,
       message: "expected a from expression with a schema"
   end
+
+  defp assert_structs!(structs) when is_list(structs) do
+    case Enum.reject(structs, &is_struct?/1) do
```
qcam, line 541: 🤔, did we mean `Enum.filter` actually? Or did I misunderstand :)?

qcam: ah right I misunderstood. 🙈

lib/ecto/repo/queryable.ex
```diff
@@ -428,6 +459,26 @@ defmodule Ecto.Repo.Queryable do
     Query.where(queryable, [], ^Enum.to_list(clauses))
   end
 
+  defp query_for_reload([head| _] = structs, opts) do
```
qcam, line 462:
```suggestion
  defp query_for_reload([%schema{} | _] = structs, opts) do
```

maybe?

lib/ecto/repo/queryable.ex
```diff
@@ -485,4 +536,28 @@ defmodule Ecto.Repo.Queryable do
       query: query,
       message: "expected a from expression with a schema"
   end
+
+  defp assert_structs!(structs) when is_list(structs) do
```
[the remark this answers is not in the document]
v0idpwn: I'll cover this case

lib/ecto/repo/queryable.ex
polvalente: @josevalim wouldn't `|> Repo.preload()` imply another trip to the database while `from(..., preload: ...)` does it in one pass? Or they're always equivalent?

lib/ecto/repo/queryable.ex
polvalente: @josevalim I have a conceptual question: does `Map.put(%{}, :__struct__, ...)` imply that the result is a struct?

In my conception, matching on `%_{__meta__: _}` would be a little bit more semantic in this regard, but I really don't know if there is a practical difference

lib/ecto/repo/queryable.ex
josevalim: Preload always issues separates queries, that’s what it is about, so both forms are equivalent!

lib/ecto/repo/queryable.ex
josevalim: Yes, it should be for both!

lib/ecto/repo/queryable.ex
josevalim: This is a subset match, so you can match as above, but you shouldn’t build the struct manually as you may miss some fields. In any case your syntax is definitely nicer!

lib/ecto/repo/queryable.ex
polvalente: That's nice to know! I thought the form in `Ecto.Query` would optimize into a join of some kind

v0idpwn:
Hi! I just fixed most of the suggestions. I think there is room for improvement in the code, but I didn't figure out how to improve it.

v0idpwn:
@josevalim let me know if you wanna merge without a squash, I may improve the commits to make a better history :)

v0idpwn:
Bump :)

(just in case you lost the notification)

lib/ecto/repo.ex
```diff
@@ -657,6 +666,64 @@ defmodule Ecto.Repo do
...
+  the database will be returned as `nil`.
+  
+  ## Options
+
+    * `:prefix` - The prefix to run the query on (such as the schema path
+      in Postgres or the database in MySQL). This will be applied to all `from`
+      and `join`s in the query that did not have a prefix previously given
+      either via the `:prefix` option on `join`/`from` or via `@schema_prefix`
+      in the schema. For more information see the "Query Prefix" section of the
+      `Ecto.Query` documentation.
```
josevalim, line 682: I don't think we should accept this option, rather, we should use the prefix from the structs themselves. You can find it under the `__meta__` field.

lib/ecto/repo.ex
```diff
@@ -657,6 +666,64 @@ defmodule Ecto.Repo do
               opts :: Keyword.t()
             ) :: Ecto.Schema.t()
 
+  @doc """
+  Reloads a given schema or schema list from the database.
+
+  When using with lists, ordering is guaranteed to be kept. Results not found in
+  the database will be returned as `nil`.
```
josevalim, line 673:
```suggestion
  When using with lists, it is expected that all of the structs in the list belong
  to the same schema. Ordering is guaranteed to be kept. Results not found in the
  database will be returned as `nil`.
```

lib/ecto/repo/queryable.ex
```diff
@@ -74,6 +74,43 @@ defmodule Ecto.Repo.Queryable do
...
+
+  def reload(name, struct, opts) do
+    one(name, query_for_reload(struct), opts)
+  end
+
+  def reload!(name, [head | _] = structs, opts) when is_list(structs) do
+    query = query_for_reload(structs)
+    results = all(name, query, opts)
+
+    [pk] = head.__struct__.__schema__(:primary_key)
```
josevalim, line 96: We should support multiple primary keys but we can do that in a later PR.

lib/ecto/repo/queryable.ex
```diff
@@ -485,4 +541,31 @@ defmodule Ecto.Repo.Queryable do
...
+    with  [] <- Enum.reject(structs, &schema?/1),
+         true <- Enum.all?(structs, &(&1.__struct__ == head.__struct__)) do
+      :ok
+    else
+      rejected when is_list(rejected) ->
+        raise ArgumentError, "expected a struct or a list of structs, received #{inspect(rejected)}"
+
+      false ->
+        raise ArgumentError, "expected an homogenous list, received different struct types"
+    end
```
josevalim, line 555: This can be clearer as conditionals:

```elixir
if not Enum.all?(structs, &schema?/1) do
  raise ArgumentError, "expected a struct or a list of structs, received #{inspect(rejected)}"
end

if not Enum.all?(structs, &(&1.__struct__ == head.__struct__)) do
  raise ArgumentError, "expected an homogenous list, received different struct types"
end

:ok
end
```

lib/ecto/repo/queryable.ex
```diff
@@ -428,6 +465,25 @@ defmodule Ecto.Repo.Queryable do
...
+        Query.from(x in schema, where: field(x, ^pk) in ^keys)
+
+      pks ->
+        raise ArgumentError,
+              "Ecto.Repo.reload/2 requires the schema #{inspect(schema)} " <>
+                "to have exactly one primary key, got: #{inspect(pks)}"
+    end
+  end
+
+  defp query_for_reload(struct), do: query_for_reload([struct])
```
josevalim, line 485: I would remove this clause and call it as `query_for_reload([struct])` in the `reload`/`reload!` functions. :)

lib/ecto/repo.ex
josevalim: You can test this by using `Ecto.put_meta(struct, :prefix, "another")` in test/ecto/repo_test.exs and then using assert_receive to check to which prefix the query went to. :)

If that doesn't work, have an integration test that sets `Ecto.put_meta(struct, :prefix, "invalid")` and try to reload it and see that it will fail.

josevalim:
Awesome work @v0idpwn! I have added some minor comments and a note about prefix handling. In particular, I think you need to do this:

    %{query | prefix: head.__meta__.prefix}

I have added some notes on how to write the test for it.

v0idpwn:
Just pushed those fixes :)

josevalim:

PR #3418, opened 2020-09-22
[WIP] Initial `Queryable.insert_all/4` to support 'INSERT INTO SELECT'

gjaldon:
refs: https://github.com/elixir-ecto/ecto/issues/3400

This is for supporting insert queries like:

```sql
INSERT INTO blog_sums ( blog_id, date, total_comments)
    SELECT blog_id, '2016-09-22', count(comment_id) as total_comments_update
    FROM blog_comments
    WHERE date = '2016-09-22'
    GROUP BY blog_id
```

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -116,6 +116,12 @@ defmodule Ecto.Query.Planner do
   The cache value is the compiled query by the adapter
   along-side the select expression.
   """
+  def query({%Ecto.Query{} = query, _prefix, _table, _on_conflict, _return_sources} = tuple, :insert_all, cache, adapter, counter) do
+    {query, params, key} = plan(query, :insert_all, adapter)
+    tuple = put_elem(tuple, 0, query)
+    query_with_cache(key, tuple, :insert_all, cache, adapter, counter, params)
+  end
+
   def query(query, operation, cache, adapter, counter) do
     {query, params, key} = plan(query, operation, adapter)
     query_with_cache(key, query, operation, cache, adapter, counter, params)
@@ -173,6 +179,13 @@ defmodule Ecto.Query.Planner do
   defp cache_reset(cache, key, prepared) do
     _ = :ets.update_element(cache, key, [{2, :cache}, {4, prepared}])
     :ok
+  end
+
+  defp query_without_cache({%Ecto.Query{} = query, _prefix, _table, _on_conflict, _return_sources} = tuple, :insert_all, adapter, counter) do
+    {query, select} = normalize(query, :insert_all, adapter, counter)
+    tuple = put_elem(tuple, 0, query)
+    {cache, prepared} = adapter.prepare(:insert_all, tuple)
+    {cache, select, prepared}
   end
 
   defp query_without_cache(query, operation, adapter, counter) do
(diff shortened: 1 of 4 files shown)


lib/ecto/repo/queryable.ex
```diff
@@ -124,6 +125,75 @@ defmodule Ecto.Repo.Queryable do
...
+      %{select: select, sources: sources, preloads: preloads} ->
+        %{
+          preprocess: preprocess,
+          postprocess: postprocess,
+          take: take,
+          assocs: assocs,
+          from: from
+        } = select
+
+        preprocessor = preprocessor(from, preprocess, adapter)
```
gjaldon, line 186: @josevalim - do I need to call this preprocessor/3 and postprocessor/4 for Queryable.insert_all or should I just call Schema.post_process/5?

lib/ecto/query/planner.ex
```diff
@@ -116,6 +116,12 @@ defmodule Ecto.Query.Planner do
   The cache value is the compiled query by the adapter
   along-side the select expression.
   """
+  def query({%Ecto.Query{} = query, _prefix, _table, _on_conflict, _return_sources} = tuple, :insert_all, cache, adapter, counter) do
```
josevalim, line 119: Instead of passing the whole tuple here, pass only the query and handle the tuple upstream. :)

lib/ecto/repo.ex
```diff
@@ -295,10 +295,32 @@ defmodule Ecto.Repo do
         def delete!(struct, opts \\ []) do
           Ecto.Repo.Schema.delete!(__MODULE__, get_dynamic_repo(), struct, with_default_options(:delete, opts))
         end
+      end
+
+      ## `insert_all/3` is defined in both Schema and Queryable
+
+      if Ecto.Adapter.Schema in behaviours and Ecto.Adapter.Queryable not in behaviours and not @read_only do
```
josevalim, line 302: I believe you can have simplify it like this:

```elixir
      if Ecto.Adapter.Schema in behaviours or Ecto.Adapter.Queryable in behaviours and not @read_only do
        def insert_all(schema_or_source, entries, opts \\ [])
      end        

      if Ecto.Adapter.Schema in behaviours and not @read_only do
        def insert_all(schema_or_source, entries, opts) do ...
      end

      if Ecto.Adapter.Queryable in behaviours and not @read_only do
        def insert_all(schema_or_source, queryable, opts) do ...
      end
```

WDYT?

lib/ecto/repo/queryable.ex
josevalim: Hrm... now that I think about it, you are correct. We don't want to call this preprocessor/postprocessor at all. Rather, we want to depend on everything the Ecto.Repo.Schema does about the return type. In fact, we may even want to use the same adapter.insert_all instructions, except that we pass a query instead of `rows` here: https://github.com/elixir-ecto/ecto/pull/3418/files#diff-053c85c1728be8c0624c7e24d6eddc83R54

Given we ended-up invoking a bunch of functions in Ecto.Repo.Schema, I wonder if it would be better to move this functionality to Ecto.Repo.Schema altogether.

The other thing is that we need to guarantee that the select returns a map and the map keys must match the fields in the schema. Those are going to be used as the INTO headers.

josevalim:
Hi @gjaldon! I took another look and a stab at some of our discussions and I have to say that unfortunately this feature is way too complex to justify its addition to Ecto, so I will go ahead and close this. In any case, I want to thank you for your work, as we wouldn't be able to reach this conclusion if it was not for your PR. :heart:

gjaldon:
Ok sounds good, @josevalim. I agree implementing this is more complicated than for the other queryable cases like for `update_all`, `delete_all` and `all` and may not be worth it considering that the given example for its use is probably better covered with a materialized view.

PR #3425, opened 2020-09-28
Bump decimal and jason

v0idpwn:
Ecto was supporting both decimal 1.6 and 2.0, this was causing it to eventually generate warnings (from the deprecated function Decimal.cmp/2) when compiled as a dependency. This bumps both Decimal and Jason.


josevalim:
We want to support earlier versions so we don't force users to upgrade for now, so a warning is fine and users can always `mix deps.update` to get rid of the warning themselves. Thanks!

PR #3431, opened 2020-10-05
Fix deprecation warning on build

bruno-b-martins:
Hello there,

Just a quick fix to a deprecation warning coming from Decimal.
```
warning: Decimal.cmp/2 is deprecated. Use compare/2 instead
  lib/ecto/changeset.ex:2152: Ecto.Changeset.validate_number/6
```
Looking at the [docs](https://hexdocs.pm/decimal/readme.html#comparisons), it looks like only the name changed.


josevalim:
We want to keep compatibility for now to not force decimal users to update. In the future we shall remove it. :) Thanks!

Issue #3434, opened 2020-10-06
belongs_to with Ecto.ParameterizedType


elielhaouzi:
### Environment

* Elixir version (elixir -v): 1.10.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): 
* Ecto version (mix deps): 3.5
* Database adapter and version (mix deps): postgrex
* Operating system: ubuntu

### Current behavior

```elixir
  defmodule MyParameterizedType do
    use Ecto.ParameterizedType

    def params(embed), do: %{embed: embed}
    def init([some_opt: :some_opt_value, field: :my_type, schema: _]), do: :init
    def type(_), do: :id
    def load(_, _, _), do: {:ok, :load}
    def dump( _, _, _),  do: {:ok, :dump}
    def cast( _, _),  do: {:ok, :cast}
    def equal?(true, _, _), do: true
    def equal?(_, _, _), do: false
    def embed_as(_, %{embed: embed}), do: embed
  end

  defmodule Schema1 do
    use Ecto.Schema

    @primary_key {:id, MyParameterizedType, some_opt: :some_opt_value, autogenerate: true}
    schema "" do
    end
  end

  defmodule Schema2 do
    use Ecto.Schema

    schema "" do
      belongs_to(:schema_1, Schema1, type: MyParameterizedType, some_opt: :some_opt_value)
    end
  end
```
(ArgumentError) invalid option :some_opt for belongs_to/3

In principle, the code support the parameterized type and init it right with the options received ([here](https://github.com/elixir-ecto/ecto/blob/v3.5.0/lib/ecto/schema.ex#L2086)) the problem is the belongs_to function limit the options to its keywords ([here](https://github.com/elixir-ecto/ecto/blob/v3.5.0/lib/ecto/schema.ex#L1881)) via the `@valid_belongs_to_options`

elielhaouzi:
After reflexion, the problem with my solution is that we need to re-write in the belonged field all the options. 
Maybe is it a way to get the options of the `related_key` via the the `related` module ?
The question is is there a way to get the `related` module attribute from the current module ?

josevalim:
In this case, it is better for you to manually define the field and pass `define_field: false` to `belongs_to`. Although I would be careful with using both belongs_to and parameterized types, as each of those may make different assumptions about how the type should behave.

PR #3435, opened 2020-10-06
belongs_to with ParameterizedType

elielhaouzi:
#3434 

makes belongs_to function accepts options for ParameterizedType other that the default ones.

I assume that once the options passed to the belongs_to fields are valid, the field is right initiated as we see in the `check_field_type!` function.

### DIFF
--- a/integration_test/support/types.exs
+++ b/integration_test/support/types.exs
@@ -32,3 +32,13 @@ defmodule WrappedInteger do
   def load(integer), do: {:ok, {:int, integer}}
   def dump({:int, integer}), do: {:ok, integer}
 end
+
+defmodule ParameterizedPrefixedString do
+  use Ecto.ParameterizedType
+  def init(opts), do: Enum.into(opts, %{})
+  def type(_), do: :string
+  def cast(string, %{prefix: _}), do: {:ok, string |> String.split("-") |> List.last()}
+  def dump(string, _, %{prefix: _prefix} = opts), do: cast(string, opts)
+  def load(string, _, %{prefix: prefix}), do: {:ok, prefix <> "-" <> string}
+  def embed_as(_, _), do: :dump
+end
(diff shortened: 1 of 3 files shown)


josevalim:
Good work! As mentioned in #3434, I would still be careful about mixing both but given the PR effectively allows them to be used together, we can at least explore how it goes and maybe revert it if indeed there complications and the whole pattern should be discouraged. :) Thanks! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

elielhaouzi:
My bad, maybe the ParameterizedPrefixedString module that I added to types.exs may be confusing/wrong.

[There](https://github.com/elixir-ecto/ecto/blob/master/integration_test/support/types.exs#L36), the dump function returns the result of the cast function as though the cast need to return the Ecto native type but it is not true, the cast function need to return the runtime data as mentioned in the Ecto.Type doc.

I did the same error in [ParameterizedTypePrefixedUUID](https://github.com/elixir-ecto/ecto/blob/master/test/ecto/repo/autogenerate_test.exs#L74) and in [ParameterizedTypePrefixedID](https://github.com/elixir-ecto/ecto/blob/master/test/ecto/repo/autogenerate_test.exs#L92) 

This is the fix for the ParameterizedPrefixedString:
```elixir
defmodule ParameterizedPrefixedString do
  use Ecto.ParameterizedType
  def init(opts), do: Enum.into(opts, %{})
  def type(_), do: :string

  def cast(value, %{prefix: prefix}) do
    if String.starts_with?(value, [prefix <> "-"] do
      {:ok, value}
    else
      {:ok, prefix <> "-" <> value}
    end
  end

  def dump(string, _, %{prefix: _prefix} = opts), do: {:ok, string |> String.split("-") |> List.last()}
  def load(string, _, %{prefix: prefix}), do: {:ok, prefix <> "-" <> string}
  def embed_as(_, _), do: :dump
end
```
it is not critical because in all cases it returns a string but can conduct to error to future dev that see the code.

To do an another PR to fix this?

josevalim:
Good find. Yes, please do send a PR! :heart:

PR #3436, opened 2020-10-07
Fix - cast/2 implementation in module for tests

elielhaouzi:
cast/2 implementation of Ecto.ParameterizedType needs to return the runtime data and not the data to dump


josevalim:

PR #3437, opened 2020-10-07
Add :array example to Ecto.Enum docs

frankdugan3:


lib/ecto/enum.ex
```diff
@@ -9,6 +9,10 @@ defmodule Ecto.Enum do
 
       field :status, Ecto.Enum, values: [:foo, :bar, :baz]
 
+  More complex types, e.g. `:array`, are also supported:
```
josevalim, line 12:
```suggestion
  Composite types, such as `:array`, are also supported:
```

lib/ecto/enum.ex
```diff
@@ -9,6 +9,10 @@ defmodule Ecto.Enum do
...
+
+    field :roles, {:array, Ecto.Enum}, values: [:Author, :Editor, :Admin]
```
josevalim, line 14:
```suggestion
      field :roles, {:array, Ecto.Enum}, values: [:Author, :Editor, :Admin]
```

josevalim:

PR #3440, opened 2020-10-10
Fix "__schema__/1 is undefined or private" error while inspecting a schemaless changeset

ruudk:
This fixes an issue introduced in v3.5.0 (https://github.com/elixir-ecto/ecto/pull/3385) when inspecting a changeset that is based on a struct without a schema.

```elixir
defmodule NoSchemaPost do
  defstruct [:title, :upvotes]
end

data   = {%NoSchemaPost{title: "hello"}, %{title: :string, upvotes: :integer}}
params = %{"title" => "world", "upvotes" => "0"}

data
|> cast(params, ~w(title upvotes)a)
|> IO.inspect

# got UndefinedFunctionError with message "function Ecto.ChangesetTest.NoSchemaPost.__schema__/1 is 
# undefined or private" while inspecting %{__struct__: Ecto.Changeset, action: nil, 
# changes: %{title: "world", upvotes: 0}, constraints: [], data: %Ecto.ChangesetTest.NoSchemaPost{title: "hello", upvotes: nil},
# empty_values: [""], errors: [], filters: %{}, params: %{"title" => "world", "upvotes" => "0"}, prepare: [],
# repo: nil, repo_opts: [], required: [], types: %{title: :string, upvotes: :integer}, valid?: true, validations: []}\"}
```

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2941,7 +2941,7 @@ defimpl Inspect, for: Ecto.Changeset do
     end
 
     redacted_fields = case data do
-      %type{} -> if function_exported?(type, :__schema__, 2), do: type.__schema__(:redact_fields), else: []
+      %type{__meta__: _} -> type.__schema__(:redact_fields)
       _ -> []
     end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2941,7 +2941,7 @@ defimpl Inspect, for: Ecto.Changeset do
     end
 
     redacted_fields = case data do
-      %type{} -> type.__schema__(:redact_fields)
+      %type{} -> if function_exported?(type, :__schema__, 2), do: type.__schema__(:redact_fields), else: []
       _ -> []
     end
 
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -2941,7 +2941,7 @@ defimpl Inspect, for: Ecto.Changeset do
     end
 
     redacted_fields = case data do
-      %type{} -> type.__schema__(:redact_fields)
+      %type{} -> if function_exported?(type, :__schema__, 2), do: type.__schema__(:redact_fields), else: []
```
ruudk, line 2944: I tried to use pattern matching but couldn't get it to work. There is probably a much better way to solve this.

josevalim: You can check if it also has a `__meta__: ...` field instead of function_exported?. WDYT?

josevalim:
```suggestion
      %type{__meta__: _} -> type.__schema__(:redact_fields)
```

josevalim:

PR #3444, opened 2020-10-14
Improve default error message when cast fails

lexmag:
This helps to distinguish it from other validations.


josevalim:
Casting is not always about being an invalid type. For example, if you can send a date as a string but the string has to be in the correct format. So the string type itself is correct but its format is invalid. Also, leaking about types in the UI is probably not recommended as it is a programming concern which may make little sense to users. :) In any case,. thanks!

lexmag:
I thought it might be the reason of not going with a similar message at first place.

PR #3445, opened 2020-10-16
Align behaviour cast for { :array, :string } with :string

danj3:
In the refactor of type.ex for 3.5 a breaking change and
inconsistency was introduced:

Prior to the Type refactor:
```
3.4> Ecto.Type.cast( {:array, :string} , ["hello" , nil ])
{:ok, ["hello", nil]}
```

After the Type refactor:
```
3.5> Ecto.Type.cast( {:array, :string} , ["hello" , nil ])
:error
```

It's also an inconsistency with a cast invocation like this:
```
3.5> Ecto.Type.cast( :string, nil )
{:ok, nil}
```

After amending type.ex with this head:
```
 defp array([nil | t], fun, acc ),
    do: array( t, fun, [nil | acc ] )
```

which would resolve the `{ :array, :string }` problem,
a few test cases fail, an :integer array [3,nil] which
is a 3.5 `:error` now returns [3,nil]. The parameterized
test exhibits similar behaviour. The `:integer` behaviour
mirror the :string change.

I do believe the above array() head is the correct
and consistent behaviour given `cast( :integer, nil ) -> { :ok, nil }`.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -765,8 +765,8 @@ defmodule Ecto.Type do
   defp cast_fun(:utc_datetime), do: &maybe_truncate_usec(cast_utc_datetime(&1))
   defp cast_fun(:utc_datetime_usec), do: &maybe_pad_usec(cast_utc_datetime(&1))
   defp cast_fun({:param, :any_datetime}), do: &cast_any_datetime(&1)
-  defp cast_fun({:in, type}), do: &array(&1, cast_fun(type), [])
-  defp cast_fun({:array, type}), do: &array(&1, cast_fun(type), [])
+  defp cast_fun({:in, type}), do: &cast_array(&1, cast_fun(type), type, [])
+  defp cast_fun({:array, type}), do: &cast_array(&1, cast_fun(type), type, [])
   defp cast_fun({:map, type}), do: &map(&1, cast_fun(type), %{})
   defp cast_fun({:parameterized, mod, params}), do: &mod.cast(&1, params)
   defp cast_fun(mod) when is_atom(mod) do
@@ -1171,22 +1171,22 @@ defmodule Ecto.Type do
   defp of_base_type?(:date, value), do: Kernel.match?(%Date{}, value)
   defp of_base_type?(_, _), do: false
 
-  defp array([nil | t], fun, acc ),
-    do: array( t, fun, [nil | acc ] )
+  defp cast_array([nil | t], fun, type, acc ) when type in @base,
+    do: cast_array( t, fun, type, [nil | acc] )
 
-  defp array([h | t], fun, acc) do
+  defp cast_array([h | t], fun, type, acc) do
     case fun.(h) do
-      {:ok, h} -> array(t, fun, [h | acc])
+      {:ok, h} -> cast_array(t, fun, type, [h | acc])
       :error -> :error
       {:error, _custom_errors} -> :error
     end
   end
 
-  defp array([], _fun, acc) do
+  defp cast_array([], _fun, _type, acc) do
     {:ok, Enum.reverse(acc)}
   end
 
-  defp array(_, _, _) do
+  defp cast_array(_, _, _, _) do
     :error
   end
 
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks @danj3!

The parameterized type is responsible for handling nil and this change makes it so it no longer handles them. So that would be an inconsistency on the other side. Given we have already shipped this, I am thinking of rather updating the CHANGELOG and require users to use a custom type if they want to have arrays with nils or to do a pre-filter or similar.

Can you talk a bit more about your use case?

danj3:
Good point on the non-base type handling. I guarded the nil handling to base types only. I think it is important and logically predictable that casting a scalar should behave the same way as casting an array of the same scalar.

This came about because I have embedded types stored in a db and some { :array, :string } already have nils, and when testing the ecto upgrade loading those broke. I think it's not desirable to have a stored type dump that will not not load because of a version change. Compounding, it may not reveal itself as a problem until some point in the future when some random dump fails to load.

josevalim:
> This came about because I have embedded types stored in a db and some { :array, :string } already have nils

That's a good point. Unfortunately the fix is a bit more complicated though, since we also need to fix dump/load, so I push a commit to master. Can you please give it a try? If it works I can cut a new release later. Thanks!

danj3:
I've tried the commit pushed to master but it doesn't fix the case:

```
** (ArgumentError) cannot load `["123 main street", nil]` as type {:array, :string} for field `address` in schema Redacted
    (ecto 3.5.2) lib/ecto/schema/loader.ex:76: anonymous fn/5 in Ecto.Schema.Loader.unsafe_load/4
    (elixir 1.10.4) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto 3.5.2) lib/ecto/embedded.ex:60: Ecto.Embedded.load/3
    (ecto 3.5.2) lib/ecto/type.ex:870: Ecto.Type.process_loaders/3
    (ecto 3.5.2) lib/ecto/repo/queryable.ex:411: Ecto.Repo.Queryable.struct_load!/6
    (ecto 3.5.2) lib/ecto/repo/queryable.ex:243: anonymous fn/5 in Ecto.Repo.Queryable.preprocessor/3
    (ecto 3.5.2) lib/ecto/repo/assoc.ex:23: anonymous fn/3 in Ecto.Repo.Assoc.query/4
    (elixir 1.10.4) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
```

The pull request patch allows this to succeed.

Is there something else I can provide to help resolve this?

josevalim:
Can you provide a mechanism to reproduce this? Both cast and loading works for me:

```elixir
iex(3)> Ecto.Type.load {:array, :string}, ["123 main street", nil]
{:ok, ["123 main street", nil]}
iex(4)> Ecto.Type.cast {:array, :string}, ["123 main street", nil]
{:ok, ["123 main street", nil]}
```

If you try the lines above in your IEx, does it work for you?

danj3:
This reproduces:
```
defmodule EmbedTest do
  use Ecto.Schema

  @primary_key false

  embedded_schema do

    embeds_one :ship_addr, ShipAddr, on_replace: :update, primary_key: false do
      field :address, { :array, :string }
      field :city, :string
      field :postal_code, :string
      field :state, :string
    end

  end

  def fail_sample_data do
    %{
      "ship_addr" =>
      %{
        "address" => ["123 Main St", nil],
        "city" => "Memphis",
        "postal_code" => "12345",
        "state" => "TN"
      }
    }
  end

  def ok_sample_data do
    %{
      "ship_addr" =>
      %{
        "address" => ["123 Main St", "Apt 1a"],
        "city" => "Memphis",
        "postal_code" => "12345",
        "state" => "TN"
      }
    }
  end

  def load_fail do
    Ecto.embedded_load( EmbedTest, fail_sample_data(), :json )
  end

  def load_ok do
    Ecto.embedded_load( EmbedTest, ok_sample_data(), :json )
  end
end
```

danj3:
I also tried the above with my pull and it doesn't fail.

josevalim:
I copied the file above as is to ecto root and then I did:

     mix run -r foo.exs -e "IO.inspect EmbedTest.load_fail"

And it printed it was loaded successfully. are you sure you are running on Ecto master? What is your commit sha shown by `mix deps`?

danj3:
I followed your instructions and it passes. I thought I ran the test against master, had it checked out in a path, but apparently I had something wrong. Thanks for your patience, and thanks for the fix!

Issue #3452, opened 2020-10-22
Changeset's typespec specifies required keys for maps that can be empty


duzzifelipe:
### Environment

* Elixir version: 1.9.4-otp-22
* Ecto version: 3.4.6

### Current behavior

While working on a new feature to replace all of our testing mocks from [mox](https://github.com/dashbitco/mox) to [hammox](https://github.com/msz/hammox), our team faced some type warnings on hammox related to ecto:

```
** (Hammox.TypeMatchError) 
     Returned value {:error, #Ecto.Changeset<action: nil, changes: %{}, errors: [], data: nil, valid?: false>} does not match type {:error, Ecto.Changeset.t()}
...
(some more text)
...
Map value %{} for key :changes does not match map value type %{required(atom()) => term()}
```

@polvalente opened a PR for hammox here https://github.com/msz/hammox/pull/63, but according to the discussion, it seemed to be a problem more related to Ecto than hammox itself, since the changeset's `changes` key was `%{}` and Ecto specifies a spec `%{required(...) => ...}`.

### Expected behavior
Since `changes` key can be an empty map, the typespec should be changed from required map keys to optional keys. The same applies to `repo_changes` (doesn't exist on master) and `filters`.

Additional info: according to the [documentation](https://hexdocs.pm/elixir/typespecs.html#literals), a typespec of `%{key_type => value_type}` is the same as `%{required(key_type) => value_type}`.

PR #3453, opened 2020-10-22
refactor: changeset typespec

duzzifelipe:
Resolves the issue explained here: https://github.com/elixir-ecto/ecto/issues/3452.

Just an additional comment: I had this problem using the version 3.4.6 and there is a third key that needs to be changed (repo_changes). But since I'm targeting to master, it doesn't exist on this PR.


josevalim:
Sorry, I have already fixed in master when I saw the issue. :) Next time, feel free to skip the issue and jump straight into the PR. :)

Issue #1479, opened 2016-06-11
Support where exists(subquery) as well as where left in subquery


josevalim:

josevalim:
Unfortunately implementing this is way trickier than expected if we want to allow the bindings in the parent query to be accessed from the subquery. For now, the best solution for this will be to use fragments still.

AndrewDryga:
I guess we can support it by allowing interpolation of Ecto.Query in `fragment/1`? Right now it throws exception, but as I understand it's should not be possible when developer is building another query.

```
to prevent sql injection, only a keyword list may be interpolated as the first argument to `fragment/1` with the `^` operator, got `#Ecto.Query<from s in Ecto.Paging.Schema, where: s.id == ^7700, select: s.id>`
```

AndrewDryga:
Example:
```
    sub_query = %Ecto.Query{from: query.from}
    |> select([c], :id)
    |> where([c], :id == ^starting_after)

    query
    |> where([c], :inserted_at > fragment(^sub_query))
```

aaronjensen:
We just hit this as well. It'd be great if we could at least build queries to use in fragments rather than having to use strings.

bcardarella:
@AndrewDryga @josevalim was there ever movement on query interpolation within `fragment` ?

kenips:
@josevalim is it possible to revisit this? Back then it was too much to implement but as Ecto matures this should be reconsidered? Forcing join only subquery or writing fragment is not the best solution in some cases (performance / maintainability) and it would be great if Ecto can support this.

fuelen:
@josevalim it seems like with introducing parent_as https://github.com/elixir-ecto/ecto/commit/7633ac0acb0b65f55d57c51ddea6c741b98e0439 this issue can be reopened, right?

josevalim:
Yup, good call! and we can even look at the “in query” support as an example of how to implement this. PRs are welcome!

fuelen:
Could you update labels and milestone?

GabrielAlacchi:
I believe this can be closed. @josevalim is there anything missing that's covered in this issue?

PR #3465, opened 2020-11-02
Add Support for Subqueries and Quantifiers in Comparison, and Exists

GabrielAlacchi:
This is my first contribution to the elixir ecosystem as a whole, and it likely won't be my last! 

This work builds on #3264 which introduced the `expr IN (subquery)` syntax to `Ecto.Query`. This laid out the groundwork for also supporting comparison operators with subqueries, including ANY and ALL [quantifiers](https://www.sqlshack.com/sql-server-universal-comparison-quantified-predicates/). I will also submit a PR to `ecto_sql` to implement the SQL in the adapters if this is merged.

Postgres, MySQL, and most others (I believe) support these operations.

### My Proposed DSL Syntax

Currently we can perform an IN subquery operation as follows
```
from(p in Post, where: p.id in subquery(some_query_expression))
```

I simply implemented comparison with quantifiers as

```
from(p in Post, where: p.rating >= all(some_query_expression))
from(p in Post, where: p.rating >= any(some_query_expression))
```
and of course `>=` can be substituted for any comparison operator! You can also re-use the same subquery syntax if you are expecting a single row to be returned from the subquery

```
avg_subquery = from(p in Post, select: avg(p.rating))
from(p in Post, where: p.rating >= subquery(avg_subquery))
```
### Implementation

I implemented this by including a `:quantifier` field in the `Ecto.SubQuery` struct. This signals to the adapter that it should include this quantifier when converting the subquery to SQL, and if the underlying database doesn't support quantifiers then it can raise an exception.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -359,6 +359,11 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:over, [], [aggregate, window]]}, params_acc}
   end
 
+  def escape({quantifier, meta, [subquery]}, type, params_acc, vars, env) when quantifier in [:all, :any, :exists] do
+    {subquery, params_acc} = escape_subquery({:subquery, meta, [subquery]}, type, params_acc, vars, env)
+    {{:{}, [], [quantifier, [], [subquery]]}, params_acc}
+  end
+
   def escape({:=, _, _} = expr, _type, _params_acc, _vars, _env) do
     error! "`#{Macro.to_string(expr)}` is not a valid query expression. " <>
             "The match operator is not supported: `=`. " <>
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1117,6 +1117,27 @@ defmodule Ecto.Query.Planner do
     {{:in, in_meta, [left, right]}, acc}
   end
 
+  defp prewalk({quantifier, meta, [{:subquery, i}]}, kind, query, expr, acc, adapter) when quantifier in [:exists, :any, :all] do
+    subquery = Enum.fetch!(expr.subqueries, i)
+    {subquery, acc} = prewalk_source(subquery, kind, query, expr, acc, adapter)
+
+    case {quantifier, subquery.query.select.fields} do
+      {:exists, _} ->
+        :ok
+
+      {_, [_]} ->
+        :ok
+
+      _ ->
+        error!(
+          query,
+          "subquery must return a single field in order to be used with #{quantifier}"
+        )
+    end
+
+    {{quantifier, meta, [subquery]}, acc}
+  end
+
   defp prewalk({{:., dot_meta, [left, field]}, meta, []},
                kind, query, expr, acc, _adapter) do
     {ix, ix_expr, ix_query} = get_ix!(left, query)
--- a/test/ecto/query/planner_test.exs
+++ b/test/ecto/query/planner_test.exs
@@ -1275,4 +1275,47 @@ defmodule Ecto.Query.PlannerTest do
       from(p in Post, order_by: p.title) |> normalize(:delete_all)
     end
   end
+
+  describe "normalize: subqueries in boolean expressions" do
+    test "replaces {:subquery, index} with an Ecto.SubQuery struct" do
+      subquery = from(p in Post, select: p.visits)
+
+      %{wheres: [where]} =
+        from(p in Post, where: p.visits in subquery(subquery))
+        |> normalize()
+
+      assert {:in, _, [_, %Ecto.SubQuery{}] } = where.expr
+
+      %{wheres: [where]} =
+        from(p in Post, where: p.visits >= all(subquery))
+        |> normalize()
+
+      assert {:>=, _, [_, {:all, _, [%Ecto.SubQuery{}] }]} = where.expr
+
+      %{wheres: [where]} =
+        from(p in Post, where: exists(subquery))
+        |> normalize()
+
+      assert {:exists, _, [%Ecto.SubQuery{}]} = where.expr
+    end
+
+    test "raises a runtime error if more than 1 field is selected" do
+      s = from(p in Post, select: [p.visits, p.id])
+
+      assert_raise Ecto.QueryError, fn ->
+        from(p in Post, where: p.id in subquery(s))
+        |> normalize()
+      end
+
+      assert_raise Ecto.QueryError, fn ->
+        from(p in Post, where: p.id > any(s))
+        |> normalize()
+      end
+
+      assert_raise Ecto.QueryError, fn ->
+        from(p in Post, where: p.id > all(s))
+        |> normalize()
+      end
+    end
+  end
 end
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -6,6 +6,7 @@ defmodule Ecto.Query.API do
     * Arithmetic operators: `+`, `-`, `*`, `/`
     * Boolean operators: `and`, `or`, `not`
     * Inclusion operator: `in/2`
+    * Subquery operators: `any`, `all` and `exists`
     * Search functions: `like/2` and `ilike/2`
     * Null check functions: `is_nil/1`
     * Aggregates: `count/0`, `count/1`, `avg/1`, `sum/1`, `min/1`, `max/1`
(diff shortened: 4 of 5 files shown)


josevalim:
Hi @GabrielAlacchi! Welcome and thanks for the PR!

I think the PR is moving in the correct direction but I don't think we need to make the quantifier part of the subquery. Instead, you want to do something similar to `:in`, where the quantifier is kept in the AST:

```elixir
  def escape({quantifier, _, [subquery]} = expr, type, params_acc, vars, env) when quantifier in [:all, :any] do
    {subquery, params_acc} = escape_subquery(right, type, params_acc, vars, env)
    {{:{}, [], [quantifier, [], [subquery]]}, params_acc}
  end
```

This should be enough (or close enough) to make it work! Plus you don't have to worry about validating the `left in any(...)` syntax because the database will complain anyway. If you want to validate it, you can validate it in the builder, while we are building the query, but IMO you can skip it.

Once you do this, we just need docs on Ecto.Query.API and the second test you added to filter_test.exs :)

GabrielAlacchi:
Hey @josevalim thanks for the follow up. It definitely comes down to the philosophy of whether you'd like Ecto to enforce correct queries at compile time or just let the database complain if any of these statements are used in the wrong place. I was calling `escape_quantifier_or_subquery` from the escape clause in builder for comparison operators. This enforced correct usage of `any` and `all` at compile time, but if this is not a concern simply treated them as functions and putting them directly into the AST actually avoids us needing to change the adapters. This also made it trivial to also support `exists` which means this also might be enough to close https://github.com/elixir-ecto/ecto/issues/1479.

I still needed to modify `prewalk` in planner since the adapters don't seem to handle `{:subquery, index}` in the AST, but rather expect an `Ecto.SubQuery`, so I made sure that these also get replaced when prewalk reaches an `all`, `any` or `exists` node. I also took the chance to validate that there's only 1 field in the query at that point, to be consistent with the implementation of `left in subquery`.

lib/ecto/query/builder.ex
```diff
@@ -281,7 +281,7 @@ defmodule Ecto.Query.Builder do
     rtype = quoted_type(left, vars)
 
     {left,  params_acc} = escape(left, ltype, params_acc, vars, env)
-    {right, params_acc} = escape(right, rtype, params_acc, vars, env)
+    {right, params_acc} = escape_subquery(right, rtype, params_acc, vars, env)
```
josevalim, line 284: I don't think you need this change. :)

lib/ecto/query/planner.ex
```diff
@@ -1105,16 +1105,47 @@ defmodule Ecto.Query.Planner do
     {{:in, in_meta, [left, right]}, acc}
   end
 
-  defp prewalk({:in, in_meta, [left, {:subquery, i}]}, kind, query, expr, acc, adapter) do
+  # Case when a subquery is at the Right Hand Side of either an in or quantifier based expression
+  defp prewalk({in_or_comp, op_meta, [left, {:subquery, i}]}, kind, query, expr, acc, adapter)
```
josevalim, line 1109: I think you can reverse these changes too to what they were! (i.e. check only for `:in`)

lib/ecto/query/builder.ex
GabrielAlacchi: This is needed to support the case where we don't want to use a quantifier. For example:

```
avg_rating = from(p in Post, select: avg(p.rating))
from(p in Post, where: p.rating > subquery(avg_rating))
```

Without this we get
```
** (Ecto.Query.CompileError) `subquery(query)` is not a valid query expression.
```

test/ecto/query/planner_test.exs
```diff
@@ -1275,4 +1275,53 @@ defmodule Ecto.Query.PlannerTest do
...
+        from(p in Post, where: p.visits >= all(subquery))
+        |> normalize()
+
+      case where.expr do
+        {:>=, _, [_, {:all, _, [%Ecto.SubQuery{}] }]} ->
+          :ok
+
+        _ ->
+          raise "planner did not replace {:subquery, index} with an Ecto.SubQuery struct in quantified comparison"
+      end
```
josevalim, line 1293:
```suggestion
      assert {:>=, _, [_, {:all, _, [%Ecto.SubQuery{}] }]} = where.expr
```

josevalim: Please do similar change below. :)

test/ecto/query/planner_test.exs
```diff
@@ -1275,4 +1275,53 @@ defmodule Ecto.Query.PlannerTest do
...
+    test "raises a runtime error if more than 1 field is selected" do
+      s = from(p in Post, select: [p.visits, p.id])
+
+      assert_raise Ecto.QueryError, fn ->
+        from(p in Post, where: p.id in subquery(s))
+        |> normalize()
+      end
+
+      assert_raise Ecto.QueryError, fn ->
+        from(p in Post, where: p.id > subquery(s))
```
josevalim, line 1317:
```suggestion
        from(p in Post, where: p.id > any(s))
```

?

lib/ecto/query/planner.ex
GabrielAlacchi: I'm also letting comparison operators go into this block for the same reasons as above. It's not necessarily to use a quantifier when comparing with a subquery. That's why I'm handling the case where a subquery without a quantifier is at the RHS of an "in" or comparison operator the same way.

lib/ecto/query/builder.ex
josevalim: Oh, you also want to support row constructors. I see. :) I would skip it for now since we don't support the `ROW` operation built-in. We can always add it later in the future though if someone has a proper use case. :)

lib/ecto/query/planner.ex
josevalim: See the above. I missed you wanted to support row constructors too but I would like to skip the row constructors one for now. :)

test/ecto/query/planner_test.exs
GabrielAlacchi: This is valid SQL assuming the subquery returns a single column and a single row. If the database finds that the subquery returns more than 1 row it'll throw a cardinality exception.

josevalim: Right, in this case I would rather have folks using all/any with proper limit, rather than support a syntax designed for row constructs which we don't properly handle. :D

lib/ecto/query/builder.ex
GabrielAlacchi: I see what you're referring too, yes row constructors are related to this work. I didn't know you could do this type of comparison with multiple fields using row constructors https://dev.mysql.com/doc/refman/8.0/en/row-subqueries.html

For now I'm just dealing with the [scalar case](https://dev.mysql.com/doc/refman/8.0/en/comparisons-using-subqueries.html) where the subquery must return a single column and be compared to a scalar expression. I use this a lot for use cases like finding "above average" rows, rows that are in certain percentiles and stuff like that.

By the way, I also tested all these queries locally with an old PHX project and they generate valid SQL without any adapter changes =).

GabrielAlacchi: So I don't want to work on supporting ROW constructors yet, I've never seen this used constructively so it's best to just use a fragment for that kind of thing, but the scalar comparison is not too difficult to support with the groundwork that's already been laid out.

josevalim: If those queries have to return a single column and a single row, then it is equivalent `foo >= all(sub)` anyway, no?

GabrielAlacchi: Good point. I guess databases support both syntaxes for backwards compatibility, we don't necessarily need to. Since it's one less thing to maintain and worry about I will drop the `>= subquery` syntax.

GabrielAlacchi:
@josevalim Ok I dropped the row constructor syntax. I will add documentation next and squash the branch. Should the docs for `any` and `all` be added to `Ecto.Query.API`?

josevalim:
> I will add documentation next and squash the branch. Should the docs for any, all and exists be added to Ecto.Query.API?

Yes, and exists too!

GabrielAlacchi:
Added some docs and squashed.

josevalim:
Awesome job as your first contribution @GabrielAlacchi, thank you! :heart:

josevalim:

GabrielAlacchi:
Obrigado @josevalim! You went to USP right? My GF is from Brazil and studied CS there =)

josevalim:
Yes, I am from Poli-USP! Please send her regards. :)

GabrielAlacchi:
She sends regards as well and respectfully says #racacaaso #chupapoli

augnustin:
@GabrielAlacchi Thanks for the PR.

I'm trying to use the `exists` method, without success. :( 

I would love to write:

```
from(m in Model, where: exists(a in assoc(m, :assoc))
```

but I can't, can I?

I also tried:

```
from(m in Model, where: exists(from a in Assoc, where: a.model_id == parent_as(:assoc).id))
```

but I get:

```
** (Ecto.SubQueryError) the following exception happened when compiling a subquery.

    ** (Ecto.QueryError) could not find named binding `parent_as(:assoc)` in query:
    
    from u0 in MyApp.Assoc,
      where: u0.model_id == parent_as(:assoc).id,
      select: u0
    

The subquery originated from the following query:

from d0 in MyApp.Model,
  where: exists(subquery(from u0 in MyApp.Assoc,
  where: u0.model_id == parent_as(:assoc).id,
  select: u0)),
  select: d0

    (elixir) lib/enum.ex:1940: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:1940: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto) lib/ecto/repo/queryable.ex:208: Ecto.Repo.Queryable.execute/4
    (ecto) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
    (iex) lib/iex/evaluator.ex:257: IEx.Evaluator.handle_eval/5
    (iex) lib/iex/evaluator.ex:237: IEx.Evaluator.do_eval/3
    (iex) lib/iex/evaluator.ex:215: IEx.Evaluator.eval/3
    (iex) lib/iex/evaluator.ex:103: IEx.Evaluator.loop/1
```

Can I get more hints on this?

Cheers

josevalim:
Where have you defined the named binding? See more info here: https://hexdocs.pm/ecto/Ecto.Query.html#module-named-bindings

augnustin:
Thanks @josevalim for your reactivity!

Indeed it does work with:

```
from(m in Model, as: :model, where: exists(from a in Assoc, where: a.model_id == parent_as(:model).id))
```

Any chance to make it work with [`Ecto.assoc`](https://hexdocs.pm/ecto/Ecto.html#assoc/2)?

Otherwise I need to fetch association module, foreign and related key, which doesn't feel DRY at all...

Cheers

augnustin:
Besides, I'm trying to pass the `:model` atom as a parameter in a function and I get:

```
  def get_assoc_data(query, assoc) do
    case query_to_module(query).__schema__(:association, assoc) do
      %{
        queryable: queryable,
        related_key: related_key,
        owner_key: owner_key
      } -> {queryable, owner_key, related_key}
    end
  end

  def build_exists_subquery(query, assoc, as) do
    {assoc_struct, owner_key, foreign_key} = get_assoc_data(query, assoc)
    from(a in assoc_struct, where: field(a, ^foreign_key) == field(parent_as(query), ^owner_key))
  end


# usage:

    from(
      m in query,
      as: :query,
      where: not exists(build_exists_subquery(Model, :assoc, :query))
    )
```

```
== Compilation error in file lib/utils.ex ==
** (Protocol.UndefinedError) protocol String.Chars not implemented for {:^, [line: 24], [{:owner_key, [line: 24], nil}]}
    (elixir) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir) lib/string/chars.ex:22: String.Chars.to_string/1
    lib/ecto/query/builder.ex:588: Ecto.Query.Builder.escape_field!/3
    lib/ecto/query/builder.ex:75: Ecto.Query.Builder.escape/5
    lib/ecto/query/builder.ex:291: Ecto.Query.Builder.escape/5
    lib/ecto/query/builder/filter.ex:67: Ecto.Query.Builder.Filter.build/6
    expanding macro: Ecto.Query.where/3
    lib/utils.ex:24: ExQueb.Utils.build_exists_subquery/3

```

PR #3467, opened 2020-11-03
Add `cast` clause for nil values on Ecto.Enum

### LINKED ISSUE #3466: Ecto.Enum in embedded schema can't load nil correctly
### Environment

* Elixir version (elixir -v): Elixir 1.11.1 (compiled with Erlang/OTP 23)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): postgresql: stable 13.0 (bottled), HEAD
* Ecto version (mix deps): 3.5.4
* Database adapter and version (mix deps): postgrex 0.15.6
* Operating system: macOS Catalina

### Current behavior

If the value of enum field in embedded schema is null, it will raise `ArgumentError` when I load data.

```elixir
** (ArgumentError) cannot load `nil` as type {:parameterized, Ecto.Enum, ......} for field `xxx` in schema
``` 

But it works fine in normal schema.

### Expected behavior

Load `nil` correctly in embedded schema.

v0idpwn:
Embedded schemas rely on `cast/2` instead of `load/3`, which implies the
need for this clause.

Closes #3466 

I'm still not 100% sure this is the correct approach, but it is up for review :)

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -60,6 +60,8 @@ defmodule Ecto.Enum do
   end
 
   @impl true
+  def cast(nil, _params), do: {:ok, nil}
+
   def cast(data, params) do
     case params do
       %{on_load: %{^data => as_atom}} -> {:ok, as_atom}


oo6:
Maybe change the return value of `Ecto.Enum.embed_as/1` to `:dump`?

josevalim:

josevalim:
@oo6 I think it is better to keep the enum higher level representation (i.e. an atom) in embeds in case some storage wants to make that distinction. :)

PR #3478, opened 2020-11-12
Allow for string decimals

zamith:
When using an inline embedded schema, backed by a map (jsonb column), decimal will be cast as string, but even though the Decimal libraries supports creating from strings, ecto's type does not.

This allows casting from a string for decimals

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -850,6 +850,7 @@ defmodule Ecto.Type do
   defp same_map(_), do: :error
 
   defp same_decimal(term) when is_integer(term), do: {:ok, Decimal.new(term)}
+  defp same_decimal(term) when is_bitstring(term), do: {:ok, Decimal.new(term)}
   defp same_decimal(term) when is_float(term), do: {:ok, Decimal.from_float(term)}
   defp same_decimal(%Decimal{} = term), do: check_decimal(term, true)
   defp same_decimal(_), do: :error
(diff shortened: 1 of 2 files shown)


jfranciscosousa (approved):

lib/ecto/type.ex
```diff
@@ -850,6 +850,7 @@ defmodule Ecto.Type do
   defp same_map(_), do: :error
 
   defp same_decimal(term) when is_integer(term), do: {:ok, Decimal.new(term)}
+  defp same_decimal(term) when is_bitstring(term), do: {:ok, Decimal.new(term)}
```
josevalim, line 853:
```suggestion
  defp same_decimal(term) when is_binary(term), do: {:ok, Decimal.new(term)}
```

josevalim:
Can you please look into the test failures? Thank you.

josevalim:
I have been looking at the code and we already handle strings when casting: https://github.com/elixir-ecto/ecto/blob/9603fe4168d92cdb324915c569b9ef4f4a973182/lib/ecto/type.ex#L816

And JSON calls cast when loading the data. So what is missing here?

lukaszsamson:
Isn't that related to the problem fixed by my (reverted) PR https://github.com/elixir-ecto/ecto/pull/3215?

zamith:
@josevalim done. Had to handle version 1.6 and 2.0 of Decimal, at least from the mix file it seems like it needs it.

zamith:
@josevalim true. I updated to the lastest version of ecto and it works 🤦

PR #3479, opened 2020-11-12
Improved issue template file.

azizk:
Hey, I'd like to improve the issue template a little bit. Using $() style placeholders after the colon makes it a bit more readable and easier to replace/fill in I think. I also changed to the British spelling of "behaviour". Can change it back if you don't like it. It's used in code documentation more often than "behavior". Also added a line for an optional Docker image... :grinning:


josevalim:
Hi @azizk! I think we are fine without these changes. $() can be confusing for those not familiar with bash and things like Docker is unlikely to matter for a project like Ecto. Thank you!

PR #3481, opened 2020-11-14
Remove Unnecessary INNER JOIN in Many To Many assoc_query

GabrielAlacchi:
This PR closes https://github.com/elixir-ecto/ecto_sql/issues/282 which points out an unnecessary INNER JOIN which is generated when preloading (or loading with `Ecto.assoc`) a many to many association.

### Approach

I simply just changed the `assoc_query` of `Ecto.Association.ManyToMany` to join the related table to the join table. I then specify in `preload_info` that the filtering should be done using `join_owner_key` on the join table.

### Non Primary Key Associations

In the related issue I remarked that this optimization would not work if `owner_key` was not the primary key of the owner table. I realized during my implementation that this wasn't true. My assumption when I said this was that we'd need to use the primary key of the owner table to perform the preloading, but the default is actually to use the key provided in `join_keys`. To sanity check this I tested the following scenario:

```
schema "products" do
  field :sku, :string

  many_to_many :locations,
    Sample.Location,
    join_through: "skus_locations",
    join_keys: [sku: :sku, location_id: :id]

  timestamps()
end

schema "locations" do
  field :name, :string

  timestamps()
end
```

The idea here is that a SKU can be listed separately so it is not unique in the products table, but a SKU is unique in the context of inventory management. So sku is used as the join key to products to avoid redundancy. I've seen this type of modelling out in the wild in eCommerce systems.

The preload query that is generated here looks correct to me:
**Before My Change**
```
[debug] QUERY OK source="locations" db=2.2ms queue=2.5ms idle=414.8ms
SELECT l0."id", l0."name", l0."inserted_at", l0."updated_at", p1."sku" FROM "locations" AS l0 INNER JOIN "products" AS p1
ON p1."sku" = ANY($1) INNER JOIN "skus_locations" AS s2 ON s2."sku" = p1."sku" 
WHERE (s2."location_id" = l0."id") ORDER BY p1."sku" [["sku", "ski", "bro"]]
```

**With My Change**
```
[debug] QUERY OK source="locations" db=2.6ms queue=3.0ms idle=777.4ms
SELECT l0."id", l0."name", l0."inserted_at", l0."updated_at", s1."sku" FROM "locations" AS l0 INNER JOIN "skus_locations" AS s1
ON l0."id" = s1."location_id" WHERE (s1."sku" = ANY($1)) ORDER BY s1."sku" [["sku", "ski", "bro"]]
```

In an extreme case some skus may be listed 20-30+ times. The original query would be loading every single product listing for that sku unnecessarily even if the actual set we're preloading against only contains a single one of these product records.

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -134,7 +134,7 @@ defmodule Ecto.Association do
   Returns information used by the preloader.
   """
   @callback preload_info(t) ::
-              {:assoc, t, {integer, atom}} | {:through, t, [atom]}
+              {:assoc, t, {integer, atom}} | {:through, t, [atom]} | {:assoc, t, {integer, atom, Ecto.Type.t()}}
 
   @doc """
   Performs the repository change on the association.
@@ -1139,14 +1139,16 @@ defmodule Ecto.Association.ManyToMany do
 
   @impl true
   def assoc_query(assoc, query, values) do
-    %{queryable: queryable, join_through: join_through, join_keys: join_keys} = assoc
-    [{join_owner_key, _}, {join_related_key, related_key}] = join_keys
+    %{queryable: queryable, join_through: join_through, join_keys: join_keys, owner: owner} = assoc
+    [{join_owner_key, owner_key}, {join_related_key, related_key}] = join_keys
+
+    owner_key_type = owner.__schema__(:type, owner_key)
 
     # We only need to join in the "join table". Preload and Ecto.assoc expressions can then filter
     # by &1.join_owner_key in ^... to filter down to the associated entries in the related table.
     from(q in (query || queryable),
       join: j in ^join_through, on: field(q, ^related_key) == field(j, ^join_related_key),
-      where: field(j, ^join_owner_key) in ^values
+      where: field(j, ^join_owner_key) in type(^values, {:in, ^owner_key_type})
     )
     |> Ecto.Association.combine_assoc_query(assoc.where)
     |> Ecto.Association.combine_joins_query(assoc.join_where, 1)
@@ -1160,10 +1162,12 @@ defmodule Ecto.Association.ManyToMany do
   end
 
   @impl true
-  def preload_info(%{join_keys: [{join_owner_key, _}, {_, _}]} = refl) do
+  def preload_info(%{join_keys: [{join_owner_key, owner_key}, {_, _}], owner: owner} = refl) do
+    owner_key_type = owner.__schema__(:type, owner_key)
+
     # When preloading use the last bound table (which is the join table) and the join_owner_key
     # to filter out related entities to the owner structs we're preloading with.
-    {:assoc, refl, {-1, join_owner_key}}
+    {:assoc, refl, {-1, join_owner_key, owner_key_type}}
   end
 
   @impl true
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -134,7 +134,7 @@ defmodule Ecto.Association do
   Returns information used by the preloader.
   """
   @callback preload_info(t) ::
-              {:assoc, t, {integer, atom}} | {:through, t, [atom]} | {:assoc, t, {integer, atom, Ecto.Type.t()}}
+              {:assoc, t, {integer, atom} | {integer, atom, Ecto.Type.t()}} | {:through, t, [atom]}
 
   @doc """
   Performs the repository change on the association.
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -1139,18 +1139,17 @@ defmodule Ecto.Association.ManyToMany do
 
   @impl true
   def assoc_query(assoc, query, values) do
-    %{queryable: queryable, join_through: join_through, join_keys: join_keys, owner: owner} = assoc
-    [{join_owner_key, owner_key}, {join_related_key, related_key}] = join_keys
+    %{queryable: queryable, join_through: join_through, join_keys: join_keys} = assoc
+    [{join_owner_key, _}, {join_related_key, related_key}] = join_keys
 
-    # We need to go all the way using owner and query so
-    # Ecto has all the information necessary to cast fields.
-    # This also helps validate the associated schema exists all the way.
+    # We only need to join in the "join table". Preload and Ecto.assoc expressions can then filter
+    # by &1.join_owner_key in ^... to filter down to the associated entries in the related table.
     from(q in (query || queryable),
-      join: o in ^owner, on: field(o, ^owner_key) in ^values,
-      join: j in ^join_through, on: field(j, ^join_owner_key) == field(o, ^owner_key),
-      where: field(j, ^join_related_key) == field(q, ^related_key))
+      join: j in ^join_through, on: field(q, ^related_key) == field(j, ^join_related_key),
+      where: field(j, ^join_owner_key) in ^values
+    )
     |> Ecto.Association.combine_assoc_query(assoc.where)
-    |> Ecto.Association.combine_joins_query(assoc.join_where, 2)
+    |> Ecto.Association.combine_joins_query(assoc.join_where, 1)
   end
 
   @impl true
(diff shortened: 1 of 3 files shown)


josevalim:
Thanks!

Did you run ecto_sql suite as specified in the README? You can set ECTO_PATH pointing to your checkout.

josevalim:
Btw, awesome job in this PR too! I am really impressed by your ability to jump into the complex parts of the code ade and make clean contributions!

GabrielAlacchi:
Thanks @josevalim this one took some head scratching. I am getting an integration testing error let me investigate it. Have you tried making a github workflow to run these tests? I can take a crack at that later if you'd like.

```
1) test many_to_many changeset assoc with self-referential binary_id (Ecto.Integration.AssocTest)
     /home/gabriel/oss/ecto/integration_test/cases/assoc.exs:470
     ** (DBConnection.EncodeError) Postgrex expected a binary of 16 bytes, got "6379f124-0984-410d-b127-81d7e8fbdbc2". Please make sure the value you are passing matches the definition in your table or in your query or convert the value accordingly.
     code: custom = Custom |> TestRepo.get!(custom.bid) |> TestRepo.preload(:customs)
     stacktrace:
       (postgrex 0.15.5) lib/postgrex/type_module.ex:897: Postgrex.DefaultTypes.encode_list/3
       (postgrex 0.15.5) lib/postgrex/extensions/array.ex:82: Postgrex.Extensions.Array.encode/4
       (postgrex 0.15.5) lib/postgrex/extensions/array.ex:52: Postgrex.Extensions.Array.encode/3
       (postgrex 0.15.5) lib/postgrex/type_module.ex:897: Postgrex.DefaultTypes.encode_params/3
       (postgrex 0.15.5) lib/postgrex/query.ex:75: DBConnection.Query.Postgrex.Query.encode/3
       (db_connection 2.2.2) lib/db_connection.ex:1148: DBConnection.encode/5
       (db_connection 2.2.2) lib/db_connection.ex:1246: DBConnection.run_prepare_execute/5
       (db_connection 2.2.2) lib/db_connection.ex:1342: DBConnection.run/6
       (db_connection 2.2.2) lib/db_connection.ex:539: DBConnection.parsed_prepare_execute/5
       (db_connection 2.2.2) lib/db_connection.ex:532: DBConnection.prepare_execute/4
       (ecto_sql 3.6.0-dev) lib/ecto/adapters/sql.ex:682: Ecto.Adapters.SQL.execute!/4
       (ecto_sql 3.6.0-dev) lib/ecto/adapters/sql.ex:674: Ecto.Adapters.SQL.execute/5
       (ecto 3.5.4) lib/ecto/repo/queryable.ex:229: Ecto.Repo.Queryable.execute/4
       (ecto 3.5.4) lib/ecto/repo/queryable.ex:17: Ecto.Repo.Queryable.all/3
       (elixir 1.11.1) lib/enum.ex:1399: Enum."-map/2-lists^map/1-0-"/2
       /home/gabriel/oss/ecto/integration_test/cases/assoc.exs:474: (test)
```

josevalim:
Is that the only failure you get? So the reason why it happens is that, because we are not going through the scheme, we are no long dumping the type when join_through is a binary. We may need to write the query like this: where: field in type(^ids, {:in, type_of_pk}) or similar. Not sure how easy it is, I can take a look later.

GabrielAlacchi:
Trying this out, the query passes type checking but the preload is failing to load the associated record in this case.

So the `assoc_query` now looks like

```
    from(q in (query || queryable),
      join: j in ^join_through, on: field(q, ^related_key) == field(j, ^join_related_key),
      where: field(j, ^join_owner_key) in type(^values, {:in, ^owner_key_type})
    )
    |> Ecto.Association.combine_assoc_query(assoc.where)
    |> Ecto.Association.combine_joins_query(assoc.join_where, 1)
```

````
1) test many_to_many changeset assoc with self-referential binary_id (Ecto.Integration.AssocTest)
     /home/gabriel/oss/ecto/integration_test/cases/assoc.exs:470
     match (=) failed
     code:  assert [_] = custom.customs
     left:  [_]
     right: []
     stacktrace:
       /home/gabriel/oss/ecto/integration_test/cases/assoc.exs:476: (test)
```

josevalim:
You can enable logging in the test_helper.exs file of each adapter. Just comment the first line.

GabrielAlacchi:
@josevalim My latest commit seems to have solved the problem with regards to the type checking with the binary id, although I'm not sure if it's the best solution. My approach was to use `type(^values, {:in, owner_key_type})` as you suggested. I also needed to tell the preloading query that the `select` of the related key has the same type as `owner_key_type`, and so I modified the definition for `related_key_to_field` to specifically add a `type` call to the AST to tell it that the related field has that type.

There remains some integration testing errors for `myxql` but these don't seem to be related to this PR? I ran against master and these still showed up.

```
  1) test decimal type cast (Ecto.Integration.TypeTest)
     /home/gabriel/oss/ecto/integration_test/cases/type.exs:424
     Assertion with == failed
     code:  assert TestRepo.all(from(p in Post, select: type(2 + ^"2", p.cost))) == [Decimal.new("4")]
     left:  [#Decimal<4.000000000000000000000000000000>]
     right: [#Decimal<4>]
     stacktrace:
       /home/gabriel/oss/ecto/integration_test/cases/type.exs:428: (test)

..............................................................................................................................................................................................................................................

  2) test bit as boolean (Ecto.Integration.MyXQLTypeTest)
     integration_test/myxql/myxql_type_test.exs:23
     ** (MyXQL.Error) (1210) (ER_WRONG_ARGUMENTS) Incorrect arguments to mysqld_stmt_execute
     code: TestRepo.insert_all("bits", [[bit: <<1::1>>], [bit: <<0::1>>]])
     stacktrace:
       (ecto_sql 3.6.0-dev) lib/ecto/adapters/sql.ex:741: Ecto.Adapters.SQL.raise_sql_call_error/1
       (ecto_sql 3.6.0-dev) lib/ecto/adapters/sql.ex:650: Ecto.Adapters.SQL.insert_all/8
       (ecto 3.5.4) lib/ecto/repo/schema.ex:54: Ecto.Repo.Schema.do_insert_all/6
       integration_test/myxql/myxql_type_test.exs:24: (test)



  3) test bit (Ecto.Integration.MyXQLTypeTest)
     integration_test/myxql/myxql_type_test.exs:14
     ** (MyXQL.Error) (1210) (ER_WRONG_ARGUMENTS) Incorrect arguments to mysqld_stmt_execute
     code: TestRepo.insert_all("bits", [[bit: <<1::1>>], [bit: <<0::1>>]])
     stacktrace:
       (ecto_sql 3.6.0-dev) lib/ecto/adapters/sql.ex:741: Ecto.Adapters.SQL.raise_sql_call_error/1
       (ecto_sql 3.6.0-dev) lib/ecto/adapters/sql.ex:650: Ecto.Adapters.SQL.insert_all/8 
       (ecto 3.5.4) lib/ecto/repo/schema.ex:54: Ecto.Repo.Schema.do_insert_all/6
       integration_test/myxql/myxql_type_test.exs:15: (test)
```

lib/ecto/repo/preloader.ex
```diff
@@ -272,6 +272,19 @@ defmodule Ecto.Repo.Preloader do
     We expected a tuple but we got: #{inspect(entry)}
     """
 
+  defp related_key_to_field(%Ecto.Association.ManyToMany{} = assoc, query, {pos, key}) do
```
josevalim, line 275: We shouldn't directly access any association here because this code should be association agnostic. Instead, you should most likely change `preload_info` to return an expanded contract that looks like this:

    {:assoc, refl, {-1, join_owner_key, type}}

If the three element tuple is returned, you wrap it with the given type.

GabrielAlacchi: Agreed this is a much cleaner approach. I just implemented this and the tests are passing (aside from those myxql ones).

lib/ecto/association.ex
```diff
@@ -134,7 +134,7 @@ defmodule Ecto.Association do
   Returns information used by the preloader.
   """
   @callback preload_info(t) ::
-              {:assoc, t, {integer, atom}} | {:through, t, [atom]}
+              {:assoc, t, {integer, atom}} | {:through, t, [atom]} | {:assoc, t, {integer, atom, Ecto.Type.t()}}
```
josevalim, line 137:
```suggestion
              {:assoc, t, {integer, atom} | {integer, atom, Ecto.Type.t()}} | {:through, t, [atom]}
```

josevalim:

tcrossland:
Just to let you know, this change breaks certain expression queries. For example, the following no longer works:

```elixir
parent
|> Ecto.assoc(:children)
|> join(:inner, [child], grandchild in assoc(child, :children))
|> where([_parent, _child, rel, _grandchild], rel.type_id == 0)
|> Repo.all()
```

... where the parent schema has a `many_to_many` relation with itself via a `join_through` schema.

PR #3482, opened 2020-11-15
Document that decimal must be used for numeric database types

iloveitaly:
This was not obvious to me. I had to dig through `postgrex` to understand that `:numeric` is mapped to a `Decimal` object.


josevalim:

Issue #3483, opened 2020-11-16
Ensure field default values match type


wojtekmach:
Currently we don't do that:

```elixir
defmodule Person do
  use Ecto.Schema

  schema "people" do
    field :age, :integer, default: "10"
  end
end

defmodule PersonTest do
  use ExUnit.Case, async: true

  test "it works" do
    Ecto.TestRepo.insert!(%Person{})
    # ** (Ecto.ChangeError) value `"10"` for `Person.age` in `insert` does not match type :integer
  end
end
```

I think we should raise an error during compilation instead.

Similar to https://github.com/dashbitco/nimble_options/pull/50.

josevalim:
Sounds good. We should call dump on it to see if it can be dumped.

v0idpwn:
Can I take this one?

wojtekmach:
Yup, got for it!

PR #3487, opened 2020-11-20
Raise ArgumentError when default has invalid type

v0idpwn:
This is achieved through a dump on the default value. We ignore nil
values and empty lists. Parameterized types use their own dump
implementation.

I added a failing test so we can discuss something: I noticed some
parameterized types will raise on `dump` if the input is not of the
expected type. What shall we do in this scenario? Catch those
errors and build the default argument error?

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1809,6 +1809,7 @@ defmodule Ecto.Schema do
     virtual? = opts[:virtual] || false
     pk? = opts[:primary_key] || false
     put_struct_field(mod, name, Keyword.get(opts, :default))
+    validate_default!(type, opts[:default])
 
     if Keyword.get(opts, :redact, false) do
       Module.put_attribute(mod, :ecto_redact_fields, name)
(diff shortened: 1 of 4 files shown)


v0idpwn:
Oh, and this closes #3483

lib/ecto/schema.ex
```diff
@@ -2061,6 +2062,30 @@ defmodule Ecto.Schema do
     Module.put_attribute(mod, :struct_fields, {name, assoc})
   end
 
+  defp validate_default!(_, nil), do: :ok
+  defp validate_default!(_, []), do: :ok
+
+  defp validate_default!({:parameterized, mod, _} = type, value) do
```
josevalim, line 2068: Do we need this clause? Ecto.Type will already do the same for us, no?

lib/ecto/schema.ex
```diff
@@ -2061,6 +2062,30 @@ defmodule Ecto.Schema do
     Module.put_attribute(mod, :struct_fields, {name, assoc})
   end
 
+  defp validate_default!(_, nil), do: :ok
+  defp validate_default!(_, []), do: :ok
```
josevalim, line 2066: Do we need these clauses? They should also be handled by Ecto.Type.dump?

josevalim:
I have added some comments. Can you please look at the tests too? Thank you!

v0idpwn:
Hi, José!
About the failing test, I added it because I wanted an opinion. Some
parameterized types will raise instead of returning an error in the dump/2
function when receiving invalid input. The failing test reproduces this
scenario. Should I catch those errors?

About the clauses, I'll try to improve them




Em sex, 20 de nov de 2020 05:29, José Valim <notifications@github.com>
escreveu:

josevalim:
I think we can just let them raise for now, yes.

lib/ecto/schema.ex
v0idpwn: You're correct! :) 
Removed it.

lib/ecto/schema.ex
v0idpwn: So, I'm in doubt on this one: 

Without the `nil` clause, some custom types will raise on dump. An example is: Ecto.Repo.AutogenerateTest.ParameterizedTypePrefixedID

Without the `[]` clause, I get failures with test scenarios that rely on `embeds_many SchemaDefinedInTestFilesOnly`, as the compiler can't seem to find the module for the Ecto.Embedded parameterized type.

v0idpwn: I believe something could be done in this direction, but I probably need help to find the best approach

josevalim: Then it this case, you either need to pass a default or you need to change `Ecto.Repo.AutogenerateTest.ParameterizedTypePrefixedID` to accept a `nil` on dump. I.e. we are actually catching a bug. 👍 

For embeds, you should probably move the place where the check happens. Check it on `__field__` directly so we don't run it for embeds and the like.

v0idpwn: Managed to remove both of them, had to change some test schemas.

v0idpwn:
I believe it is ready now. If there is any other issue, let me know :)

josevalim:

PR #3491, opened 2020-11-20
Fixes in howto

thojanssens:
Note the change from "references" to "referenced". Please review.

There are two things I didn't understand:

1.
Purpose of `create unique_index(:posts, [:id, :org_id])`. Since `:id` is the primary key, it is always unique. Why then create a unique index for `:id` with `:org_id`, if `:id` can anyway never be repeated?

2. 
> If by any chance you have schemas that are not tied to an org_id, you can even consider keeping them in a separate query prefix or in a separate database altogether, so you keep non-tenant data completely separated from tenant-specific data.

Why even separate tables scoped to org_id with tables that are not scoped? Only to avoid passing the option `:skip_org_id`? Intention/advantages could be explained.

### DIFF
--- a/guides/howtos/Multi tenancy with foreign keys.md
+++ b/guides/howtos/Multi tenancy with foreign keys.md
@@ -216,11 +216,11 @@ create table(:comments) do
   add :org_id, :integer, null: false
 
   add :post_id,
-      references(:posts, with: [org_id: :org_id]),
+      references(:posts, with: [org_id: :org_id], match: :full),
       null: false
 
   add :user_id,
-      references(:users, with: [org_id: :org_id]),
+      references(:users, with: [org_id: :org_id], match: :full),
       null: false
 
   timestamps()
--- a/guides/howtos/Multi tenancy with foreign keys.md
+++ b/guides/howtos/Multi tenancy with foreign keys.md
@@ -24,7 +24,7 @@ defmodule MyApp.Repo do
 
   require Ecto.Query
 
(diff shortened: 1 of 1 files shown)


josevalim:
1. Composite foreign keys require a unique index with all keys in the composite FK

2. Just so it is harder to accidentally query something that you should not because it isn’t unscoped

Feel free to add these clarifications to the text. :)

thojanssens:
Thank you! I'll add one more update tomorrow; sleeping 😴

thojanssens:
With this technique, `unsafe_validate_unique/4` won't be able to run its query, it seems that a custom option such as `:skip_tenant_id` won't be passed to the `Repo` function.

josevalim:
We can support `:repo_opts` in unsafe_validate_unique. A PR is welcome!

guides/howtos/Multi tenancy with foreign keys.md
```diff
@@ -216,11 +216,11 @@ create table(:comments) do
   add :org_id, :integer, null: false
 
   add :post_id,
-      references(:posts, with: [org_id: :org_id]),
+      references(:posts, with: [org_id: :org_id], match: :full),
```
josevalim, line 219: This is not supported by all DBs, so we should revert it?

mathieuprog: Reverted

josevalim:

PR #3492, opened 2020-11-22
Support `:repo_opts` in `unsafe_validate_unique`

mathieuprog:
Not sure how to test it

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -101,7 +101,7 @@ defmodule Ecto.Changeset do
   transaction.
 
   ## Empty values
-  
+
   Many times, the data given on cast needs to be further pruned, specially
   regarding empty values. For example, if you are gathering data to be
   cast from the command line or through an HTML form or any other text-based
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1863,7 +1864,7 @@ defmodule Ecto.Changeset do
           query
         end
 
-      if repo.one(query) do
+      if repo.one(query, repo_opts) do
```
wojtekmach, line 1867: Out of curiosity, what's your use case, which options do you need to pass?

Instead of adding another argument, maybe we should use the existing `opts` argument like this:

```elixir
repo_opts = Keyword.take(opts, [:prefix, :timeout, :log, :telemetry_event, :telemetry_prefix])`

if repo.one(query, opts) do
```

this way we could also remove lines 1859-1864. It would be easier to document too, we could just say see "Shared options" in Repo, same as we do in https://hexdocs.pm/ecto/Ecto.Repo.html#c:one/2-options

mathieuprog: See how we get the tenant ID in `prepare_query` through a 'custom' option passed to Repo in this howto:
https://github.com/elixir-ecto/ecto/blob/5904da3073832b533e87ed12a75bfbccfe7aeb66/guides/howtos/Multi%20tenancy%20with%20foreign%20keys.md

https://github.com/elixir-ecto/ecto/pull/3491#issuecomment-731763071

lib/ecto/changeset.ex
```diff
@@ -1822,7 +1822,8 @@ defmodule Ecto.Changeset do
 
   """
   @spec unsafe_validate_unique(t, atom | [atom, ...], Ecto.Repo.t, Keyword.t) :: t
-  def unsafe_validate_unique(changeset, fields, repo, opts \\ []) when is_list(opts) do
+  def unsafe_validate_unique(changeset, fields, repo, opts \\ [], repo_opts \\ [])
```
josevalim, line 1825: Instead of passing it as a separate argument, pass it as part of `opts`. So here you do:

    repo_opts = Keyword.pop(opts, :repo_opts, [])

:)

josevalim:

PR #3499, opened 2020-12-04
Improve the error message of Relation.on_replace/2

Ljzn:
When I trying to update the children table with `cast_assoc` changeset. I saw this error message:

```
 If you have a relationship with many children, at least the same N children must be given on update.
```

It's hard to see intuitively that the `N` is the number of the current amount of children. The following message is more friendly for debugging.

```
If you have a relationship with many children, all children must be given on update.
```


josevalim:

PR #3503, opened 2020-12-08
Remove deprecated method call

pawurb:
This PR removes a deprecation warning. 

image

Decimal 1.6 which is the lowest supported version also has `compare` method available so there shouldn't be any issues.


josevalim:
Unfortunately it broke the build, so I am afraid we can't use it yet. :'(

Issue #3470, opened 2020-11-05
Decimal.cmp/2 is deprecated. Use compare/2 instead


inoas:
### Environment

* Elixir version (elixir -v): `Erlang/OTP 23 [erts-11.1.1] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe] [dtrace]`

```
warning: Decimal.cmp/2 is deprecated. Use compare/2 instead
  lib/ecto/changeset.ex:2155: Ecto.Changeset.validate_number/6
```

josevalim:
Correct. We are keeping it for backwards compatibility but we will drop it soon. Thanks!

v0idpwn:
This should be closed due to #3504

wojtekmach:
Thanks @v0idpwn!

PR #3504, opened 2020-12-08
Remove deprecated method call

pawurb:
This PR removes the Decimal deprecation warning. 

image

It looks like `compare` has a different API between 1.6 and 2.0 so I had to add this conversion method. It can be removed if Ecto bumps Decimal dependency to 2.0.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2155,7 +2155,7 @@ defmodule Ecto.Changeset do
   end
 
   defp validate_number(field, %Decimal{} = value, message, spec_key, _spec_function, target_value) do
-    result = Decimal.cmp(value, decimal_new(target_value))
+    result = Decimal.compare(value, decimal_new(target_value)) |> normalize_compare()
     case decimal_compare(result, spec_key) do
       true  -> nil
       false -> [{field, {message, validation: :number, kind: spec_key, number: target_value}}]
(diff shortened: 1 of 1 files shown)


v0idpwn:
Hi, @pawurb
There's an issue for that: https://github.com/elixir-ecto/ecto/issues/3470
Ecto is still using Decimal.cmp/2 so it doesn't force anyone to update to decimal 2.0 for now.

There were other pull requests doing the same thing as yours: #3431, #3425,

pawurb:
https://github.com/elixir-ecto/ecto/pull/3425#issuecomment-699818342 but updating the dependencies is what will generate the warning now. `cmp` is deprecated in 2.0. My PR removes the warning for both versions without forcing anyone to update or change dependencies.

v0idpwn:
Oh, I see! Sorry for the confusion. LGTM!

v0idpwn (approved):

lib/ecto/changeset.ex
```diff
@@ -2169,6 +2169,16 @@ defmodule Ecto.Changeset do
     end
   end
 
+  # Support mismatch between API for Decimal.compare/2 for versions 1.6 and 2.0
```
josevalim, line 2172:
```suggestion
  # TODO: Remove me once we support Decimal 2.0 only
  # Support mismatch between API for Decimal.compare/2 for versions 1.6 and 2.0
```

josevalim:

PR #3505, opened 2020-12-09
Fix Ecto.Changeset.t(data_type) typespec

### LINKED ISSUE #3502: Typespec for `types` field of `Changeset.t()` is incomplete
### Environment

* Elixir version (elixir -v): Elixir 1.10.0 (compiled with Erlang/OTP 22)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 11.6
* Ecto version (mix deps): 3.5.5
* Database adapter and version (mix deps): postgrex 0.15.7
* Operating system: MacOS 10 Mojave

### Current behavior

Using [hammox](https://github.com/msz/hammox) to add typechecking for my Mox mocks, I'm getting a type mismatch error in `Ecto.Changeset.t()` on the `types` field.  I assume Dialyzer would report the same issue, but we're not currently using it on this project, so I don't have an easy way to check that.

The most relevant message from the output (slightly anonymized) is:

```
 Map value {:assoc, %Ecto.Association.BelongsTo{cardinality: :one, defaults: [], field: :account, on_cast: nil, on_replace: :raise, ordered: false, owner: User, owner_key: :account_id, queryable: Account, related: Account, related_key: :id, relationship: :parent, unique: true, where: []}} for key :account does not match map value type Ecto.Type.t().
```

The typespec for `Ecto.Changeset.t()` says that the `types` field has type `nil | %{atom => Ecto.Type.t}`, but the `{:assoc, <association>}` tuple (and from looking at the code, I think also the `{:embed, <???>}` tuple don't match `Ecto.Type.t()`.

### Expected behavior

A changeset should correctly match the `Ecto.Changeset.t()` typespec.

### Analysis

I'm not terribly familiar with this codebase, but I don't think the `:assoc`/`:embed` tuples should be part of `Ecto.Type.t()`'s definition.  Instead, I think the typespec for the `types` field should be something like:

```elixir
nil | %{atom => Ecto.Type.t | {:assoc, Ecto.Association.t()} | {:embed, ???}
```

I'm not sure what the correct type for the `:embed` tuple should be.

If this sounds like the right direction to go, I'd be happy to work on a PR, but would need some guidance on the correct typespec for the `:embed` tuple.

Thanks so much for a great library!

v0idpwn:
The `types` field on Ecto.Changeset.t(data_type) may contain
Ecto.Changeset.Relation.t(), not only Ecto.Type.t()

Closes #3502

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -288,7 +288,7 @@ defmodule Ecto.Changeset do
                         validations: [{atom, term}],
                         filters: %{optional(atom) => term},
                         action: action,
-                        types: nil | %{atom => Ecto.Type.t | Relation.t}}
+                        types: nil | %{atom => Ecto.Type.t | {:assoc, Ecto.Association.t()} | {:embed, Ecto.Embedded.t()}}
 
   @type t :: t(Ecto.Schema.t | map | nil)
   @type error :: {String.t, Keyword.t}
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -288,7 +288,7 @@ defmodule Ecto.Changeset do
                         validations: [{atom, term}],
                         filters: %{optional(atom) => term},
                         action: action,
-                        types: nil | %{atom => Ecto.Type.t}}
+                        types: nil | %{atom => Ecto.Type.t | Relation.t}}
 
   @type t :: t(Ecto.Schema.t | map | nil)
   @type error :: {String.t, Keyword.t}


lib/ecto/changeset.ex
```diff
@@ -288,7 +288,7 @@ defmodule Ecto.Changeset do
                         validations: [{atom, term}],
                         filters: %{optional(atom) => term},
                         action: action,
-                        types: nil | %{atom => Ecto.Type.t}}
+                        types: nil | %{atom => Ecto.Type.t | Relation.t}}
```
randycoulman, line 291: I tested this change and it doesn't solve the issue. However, in investigating (and testing) further, I think I found the right typespec for both the `:embed` and `:assoc` tuples:

```suggestion
                        types: nil | %{atom => Ecto.Type.t | {:assoc, Ecto.Association.t()} | {:embed, Ecto.Embedded.t()}}
```

I don't have any embedded schemas to test with, but I believe that this is correct. The `:assoc` tuple definitely solves my issue.

randycoulman (changes requested):
This isn't quite right, but see my suggestion below.

josevalim:

PR #3507, opened 2020-12-09
Fix broken build

v0idpwn:


v0idpwn:
I'm still a bit skeptical about the last PR. Please do not merge this one yet, maybe I'll change it a bit :)

josevalim:
I will merge it only to fix the build, sorry. :) Regarding the last PR, i think we can also simply make the value be `Ecto.Type.t | term` but it is fine as is too.

josevalim:

randycoulman:
Ah, whoops, that was my fault. Sorry! My editor had re-formatted the code and put the trailing `}` on the last line and I missed that when I put in the suggested change.  🤦

PR #3508, opened 2020-12-10
Raise when using invalid expressions on join :on

### LINKED ISSUE #3500: Adding a subquery to join's `:on` option returns cryptic error
### Environment

* Elixir version (elixir -v): 1.10.2 (OTP 21)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): 11
* Ecto version (mix deps): 3.5.5
* Database adapter and version (mix deps): postgrex, 0.15.7
* Operating system: Windows 10

### Current behavior

A subquery in a join's `:on` option returns a cryptic error. Example:

```
subset =
  from acl in Acl,
    where: acl.grantor_id == parent_as(:users).id,
    where: acl.grantee_id == ^100,
    select: acl.grantor_id

query =
  from users in User, as: :users,
    left_join: articles in Article, on: articles.author_id == users.id and users.id in subquery(subset),
    preload: [:authored_articles]
```

Error:

> == Compilation error in file test/foo_test.exs ==
** (MatchError) no match of right hand side value: {{:{}, [], [:and, [], [{:{}, [], [:==, [], [{:{}, [], [{:{}, [], [:., [], [{:{}, [], [:&, [], [1]]}, :author_id]]}, [], []]}, {:{}, [], [{:{}, [], [:., [], [{:{}, [], [:&, [], [0]]}, :id]]}, [], []]}]]}, {:{}, [], [:in, [], [{:{}, [], [{:{}, [], [:., [], [{:{}, [], [:&, [], [0]]}, :id]]}, [], []]}, {:subquery, 0}]]}]]}, {[subquery: 0], [{{:., [], [{:__aliases__, [alias: false], [:Ecto, :Query]}, :subquery]}, [], [{:subset, [line: 132], nil}]}]}}
    (ecto 3.5.5) lib/ecto/query/builder/join.ex:206: Ecto.Query.Builder.Join.build_on/7
    (ecto 3.5.5) lib/ecto/query/builder/join.ex:185: Ecto.Query.Builder.Join.build/10
    (ecto 3.5.5) lib/ecto/query.ex:844: Ecto.Query.from/5
    (ecto 3.5.5) expanding macro: Ecto.Query.from/2

v0idpwn:
Closes #3500. Still, I think it would be nice if we supported those expressions that are currently invalid.

### DIFF
--- a/lib/ecto/query/builder/join.ex
+++ b/lib/ecto/query/builder/join.ex
@@ -203,23 +203,28 @@ defmodule Ecto.Query.Builder.Join do
   end
 
   def build_on(on, join, as, query, binding, count_bind, env) do
-    {on_expr, {on_params, []}} = Ecto.Query.Builder.Filter.escape(:on, on, count_bind, binding, env)
-    on_params = Builder.escape_params(on_params)
+    case Ecto.Query.Builder.Filter.escape(:on, on, count_bind, binding, env) do
+      {on_expr, {on_params, []}}  ->
+        on_params = Builder.escape_params(on_params)
 
-    join =
-      quote do
-        %JoinExpr{
-          unquote_splicing(join),
-          on: %QueryExpr{
-            expr: unquote(on_expr),
-            params: unquote(on_params),
-            line: unquote(env.line),
-            file: unquote(env.file)
-          }
-        }
-      end
+        join =
+          quote do
+            %JoinExpr{
+              unquote_splicing(join),
+              on: %QueryExpr{
+                expr: unquote(on_expr),
+                params: unquote(on_params),
+                line: unquote(env.line),
+                file: unquote(env.file)
+              }
+            }
+          end
 
-    Builder.apply_query(query, __MODULE__, [join, as, count_bind], env)
+        Builder.apply_query(query, __MODULE__, [join, as, count_bind], env)
+
+      _pattern ->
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder/join.ex
```diff
@@ -203,23 +203,28 @@ defmodule Ecto.Query.Builder.Join do
...
+                file: unquote(env.file)
+              }
+            }
+          end
+
+        Builder.apply_query(query, __MODULE__, [join, as, count_bind], env)
 
-    Builder.apply_query(query, __MODULE__, [join, as, count_bind], env)
+      _pattern ->
+        raise ArgumentError, "invalid expression for join `:on`, only simple expressions are supported"
```
josevalim, line 226: Would it be better to explicitly say that subqueries are not allowed? Is there anything else that is “forbidden”?

v0idpwn:
"Exists" and others are also forbidden

josevalim:
Exists is also a subquery though, so I think we can be explicit about subqueries? :)

v0idpwn:
Agreed. Just amended it :)

josevalim:
Btw, I just want to say "muitissimo obrigado" for all of your contributions, both here and in the Elixir repo, they are very appreciated! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

v0idpwn:
eu que agradeço pela oportunidade de contribuir 💜

PR #3511, opened 2020-12-14
Support for integer values in Ecto.Enum

joerichsen:
Currently Ecto.Enum has support for mapping string database values to atoms like this:

```field :status, Ecto.Enum, values: [:foo, :bar, :baz]```

I have added support for mapping integer database values to atoms in with a syntax like this:

```field :status, Ecto.Enum, values: [foo: 1, bar: 2, baz: 5]```

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -50,31 +50,36 @@ defmodule Ecto.Enum do
   def init(opts) do
     values = Keyword.get(opts, :values, nil)
 
-    unless (is_list(values) and
-              Enum.all?(values, &is_atom/1)) or
-             (Keyword.keyword?(values) and Enum.all?(Keyword.values(values), &is_integer/1)) do
-      raise ArgumentError, """
-      Ecto.Enum types must have a values option specified as a list of atoms or a keyword list with a mapping from atoms to values.
+    values =
+      cond do
+        is_list(values) and
+            Enum.all?(values, &is_atom/1) ->
+          Enum.map(values, fn atom -> {atom, to_string(atom)} end)
 
-      For example:
+        Keyword.keyword?(values) and Enum.all?(Keyword.values(values), &is_integer/1) ->
+          values
 
-          field :my_field, Ecto.Enum, values: [:foo, :bar]
+        Keyword.keyword?(values) and Enum.all?(Keyword.values(values), &is_binary/1) ->
+          values
 
-      or
+        true ->
+          raise ArgumentError, """
+          Ecto.Enum types must have a values option specified as a list of atoms or a
+          keyword list with a mapping from atoms to either integer or string values.
 
-          field :my_field, Ecto.Enum, values: [foo: 1, bar: 2, baz: 5]
-      """
-    end
+          For example:
 
-    if Enum.all?(values, &is_atom/1) do
-      on_load = Map.new(values, &{Atom.to_string(&1), &1})
-      on_dump = Map.new(values, &{&1, Atom.to_string(&1)})
-      %{on_load: on_load, on_dump: on_dump, values: values}
-    else
-      on_load = Map.new(values, fn {key, val} -> {val, key} end)
-      on_dump = Enum.into(values, %{})
-      %{on_load: on_load, on_dump: on_dump, values: Keyword.keys(values)}
-    end
+              field :my_field, Ecto.Enum, values: [:foo, :bar]
+
+          or
+
+              field :my_field, Ecto.Enum, values: [foo: 1, bar: 2, baz: 5]
+          """
+      end
+
+    on_load = Map.new(values, fn {key, val} -> {val, key} end)
+    on_dump = Enum.into(values, %{})
+    %{on_load: on_load, on_dump: on_dump, values: Keyword.keys(values)}
   end
 
   @impl true
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -4,19 +4,23 @@ defmodule Ecto.Enum do
 
   `Ecto.Enum` must be used whenever you want to keep atom values in a field.
   Since atoms cannot be persisted to the database, `Ecto.Enum` converts them
-  to string when writing to the database and converts them back to atoms when
-  loading data. It can be used in your schemas as follows:
+  to a string or an integer when writing to the database and converts them back
+  to atoms when loading data. It can be used in your schemas as follows:
 
       field :status, Ecto.Enum, values: [:foo, :bar, :baz]
+
+  or
+
+      field :status, Ecto.Enum, values: [foo: 1, bar: 2, baz: 5]
 
   Composite types, such as `:array`, are also supported:
 
       field :roles, {:array, Ecto.Enum}, values: [:Author, :Editor, :Admin]
 
-  `:values` must be a list of atoms. String values will be cast to atoms safely
-  and only if the atom exists in the list (otherwise an error will be raised).
-  Attempting to load any string not represented by an atom in the list will be
-  invalid.
+  `:values` must be a list of atoms or a keyword list. Values will be cast to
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by wojtekmach** [comment]
I think this is a welcome change and I believe we discussed it as a use case in previous conversations.

I’d make sure that we can’t mix atom and integer values, e.g: `[:a, b: 2, c: 3]` raises with a good error message.

File: lib/ecto/enum.ex [hunk 1]
```diff
@@ -46,17 +50,25 @@ defmodule Ecto.Enum do
...
-    on_dump = Map.new(values, &{&1, Atom.to_string(&1)})
-    %{on_load: on_load, on_dump: on_dump, values: values}
+    if Enum.all?(values, &is_atom/1) do
+      on_load = Map.new(values, &{Atom.to_string(&1), &1})
+      on_dump = Map.new(values, &{&1, Atom.to_string(&1)})
+      %{on_load: on_load, on_dump: on_dump, values: values}
+    else
+      on_load = Map.new(values, fn {key, val} -> {val, key} end)
+      on_dump = Enum.into(values, %{})
+      %{on_load: on_load, on_dump: on_dump, values: values}
```
josevalim, line 70: Please normalize `values` to always be a map so we don't need to check for `Keyword.keyword?` later on.

lib/ecto/enum.ex
```diff
@@ -102,9 +114,14 @@ defmodule Ecto.Enum do
     rescue
       _ in UndefinedFunctionError -> raise ArgumentError, "#{inspect schema} is not an Ecto schema"
     else
-      %{^field => {:parameterized, Ecto.Enum, %{values: values}}} -> values
-      %{^field => {_, {:parameterized, Ecto.Enum, %{values: values}}}} -> values
-      %{} -> raise ArgumentError, "#{field} is not an Ecto.Enum field"
+      %{^field => {:parameterized, Ecto.Enum, %{values: values}}} ->
+        (Keyword.keyword?(values) && Keyword.keys(values)) || values
```
josevalim, line 118: See note above about normalizing `values`.

josevalim: To be clear, this should always return the atom keys. :)

thojanssens:
Looks great:) Can I ask uses cases for this (storing an integer and mapped as an atom), instead of just storing the atom as a string in the db?

v0idpwn:
@thojanssens faster lookups seems like a good selling point for me

lib/ecto/enum.ex
```diff
@@ -4,19 +4,23 @@ defmodule Ecto.Enum do
 
   `Ecto.Enum` must be used whenever you want to keep atom values in a field.
   Since atoms cannot be persisted to the database, `Ecto.Enum` converts them
-  to string when writing to the database and converts them back to atoms when
-  loading data. It can be used in your schemas as follows:
+  to a string or an integer when writing to the database and converts them back
+  to atoms when loading data. It can be used in your schemas as follows:
```
v0idpwn, line 8: Maybe we could generalize and support anything?

For instance, a list of [:a, :b, :c] is equal to [a: "a", b: "b", c: "c"], and maybe one could use something like [a: "b", b: "c", c:   "d"]. Seems like a more coherent API for me.

joerichsen:
> Looks great:) Can I ask uses cases for this (storing an integer and mapped as an atom), instead of just storing the atom as a string in the db?

@thojanssens Sure :-)

I have a legacy database where the values are stored as integers and because of other systems using the same database, I can't convert the integers to the corresponding string values. But I would like to have all the nice things of Ecto.Enum - hence this PR :-)

lib/ecto/enum.ex
joerichsen: That's a good point. I have changed it in 8f0b227f5db0eb7347c4601cac4f8377b13abd66 - was that what you had in mind?

lib/ecto/enum.ex
joerichsen: Fixed in 8f0b227f5db0eb7347c4601cac4f8377b13abd66 I think :-)

lib/ecto/enum.ex
```diff
@@ -46,17 +50,25 @@ defmodule Ecto.Enum do
...
-    on_dump = Map.new(values, &{&1, Atom.to_string(&1)})
-    %{on_load: on_load, on_dump: on_dump, values: values}
+    if Enum.all?(values, &is_atom/1) do
+      on_load = Map.new(values, &{Atom.to_string(&1), &1})
+      on_dump = Map.new(values, &{&1, Atom.to_string(&1)})
+      %{on_load: on_load, on_dump: on_dump, values: values}
+    else
+      on_load = Map.new(values, fn {key, val} -> {val, key} end)
+      on_dump = Enum.into(values, %{})
+      %{on_load: on_load, on_dump: on_dump, values: Keyword.keys(values)}
```
josevalim, line 70: This is good! But now that I think about it.... maybe we should stop storing `values` and use `Map.keys(on_dump)` instead? :D

lib/ecto/enum.ex
josevalim: Yes, let's go down this route. It may simplify the code too. Basically, if a list of atom is given, convert it to: `Enum.map(list, fn {k, to_string(k)} end)` and write the rest of the code assuming `values` is always a keyword. Note we don't need to validate the values, let's just assume they are valid.

josevalim:
@joerichsen you also need to change this line:

https://github.com/elixir-ecto/ecto/pull/3511/files#diff-025069044568b480ef6f86f025cb46a7064695e6df1e7008ede689e09a9c7836R47

We should possibly create a new key called :type and store it on the enum metadata. We should infer the type based on the keyword values.

lib/ecto/enum.ex
joerichsen: That is worth investigating. I just tried changing it to use ```Map.keys(on_dump)``` instead but had failing test, because the ordering of the values (sometimes) changed?

josevalim: Ah, converting to a map will change the order indeed. Ok, let's keep it then. :D

lib/ecto/enum.ex
```diff
@@ -46,17 +50,31 @@ defmodule Ecto.Enum do
   def init(opts) do
     values = Keyword.get(opts, :values, nil)
 
-    unless is_list(values) and Enum.all?(values, &is_atom/1) do
+    unless (is_list(values) and
+              Enum.all?(values, &is_atom/1)) or
+             (Keyword.keyword?(values) and Enum.all?(Keyword.values(values), &is_integer/1)) do
```
leandro, line 55: I believe these chained conditions might compromise a bit the code reading, so I'd suggest something a bit different instead:
```suggestion
    list_of_atoms? = &(is_list(&1) && Enum.all?(&1, &is_atom/1))
    keywords_of_integers? = &(Keyword.keyword?(&1) && Enum.all?(Keyword.values(&1), &is_integer/1))
    
    unless list_of_atoms?.(values) or keywords_of_integers?.(values) do
```

lib/ecto/enum.ex
joerichsen: Generalizing things and allowing for custom string mappings is a great idea.

I have added support for things like ```[a: "b", b: "c", c: "d"]``` in a00cef6b0d2d40e4f0d4aa6c9e20b202e7ed3105

@josevalim I have changed the list of atoms to a keyword list as you suggested and it cleaned up the code nicely :-)

josevalim:

thojanssens:
@joerichsen I suspected the legacy database thing:-) I'm actually surprised this made it into the library; was expecting maintainers to encourage you into making a library for such specific case.
Nice contribution anyway 👍

v0idpwn:
@thojanssens I think this implementation is better than the original one, tbh, and if we have Ecto.Enum I don't see why Ecto.Enum shouldn't be the best possible enum

PR #3513, opened 2020-12-16
fix query preload doc - query has to be pinned

elielhaouzi:
The code without the pin operator does not compile.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1910,7 +1910,7 @@ defmodule Ecto.Query do
   the comments on those posts:
 
       posts_query = from p in Post, where: p.state == :published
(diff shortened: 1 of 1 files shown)


lib/ecto/query.ex
```diff
@@ -1910,7 +1910,7 @@ defmodule Ecto.Query do
   the comments on those posts:
 
       posts_query = from p in Post, where: p.state == :published
-      Repo.all from a in Author, preload: [posts: {posts_query, [:comments]}]
+      Repo.all from a in Author, preload: [posts: ^{posts_query, [:comments]}]
```
josevalim, line 1913:
```suggestion
      Repo.all from a in Author, preload: [posts: {^posts_query, [:comments]}]
```

josevalim:

PR #3515, opened 2020-12-17
Add placeholder option to insert_all

Un3qual:
This PR contains the necessary changes to ecto core to make elixir-ecto/ecto_sql#290 work. Details are in that PR


lib/ecto/repo/schema.ex
```diff
@@ -91,28 +93,33 @@ defmodule Ecto.Repo.Schema do
...
+      case dumper do
+        %{^field => {source, type}} ->
+          case value do
+            %Ecto.Query{} = query ->
+              {{source, query}, {Map.put(header, source, true), true}}
+
+            {:placeholder, placeholder_key} ->
+              replaced_value = Map.fetch!(placeholder_map, placeholder_key)
+              value = dump_field!(:insert_all, schema, field, type, replaced_value, adapter)
+              {{source, value}, {Map.put(header, source, true), has_query?}}
```
josevalim, line 113: I believe you want to return `{:source, {:placeholder, placeholder_key}}` and instead update the dumped value in the placeholder map itself.

lib/ecto/repo/schema.ex
```diff
@@ -41,7 +41,9 @@ defmodule Ecto.Repo.Schema do
       |> returning(opts)
       |> fields_to_sources(dumper)
 
-    {rows, header} = extract_header_and_fields(rows, schema, dumper, autogen_id, adapter)
+    placeholder_map = Keyword.get(opts, :placeholders, %{})
+
+    {rows, header} = extract_header_and_fields(rows, schema, dumper, autogen_id, adapter, placeholder_map)
```
josevalim, line 46: Nitpick: the adapter is always the last argument so please keep it consistent. :)

lib/ecto/repo/schema.ex
josevalim: To be clear, I think it will be simpler if we update the to the placeholder map in here and pass the updated placeholder map to the `adapter.insert_all` call here: https://github.com/elixir-ecto/ecto/pull/3515/files#diff-91ed3e8c550063fcf5e2df36020d92405a303c5f5745c0bf643c7dc4c538befbR56

Then there is less reconstruction to do on the `ecto_sql` side of things.

integration_test/cases/repo.exs
```diff
@@ -953,6 +953,73 @@ defmodule Ecto.Integration.RepoTest do
     assert custom.bid == bid2
   end
 
+  @tag :ph
```
josevalim: Please replace :ph by :placeholders in the tags and in the descriptions. :) Also please put all placeholders tests in a single describe block.

```
describe "placeholders" do
  @describetag :placeholders
```

integration_test/cases/repo.exs
```diff
@@ -953,6 +953,73 @@ defmodule Ecto.Integration.RepoTest do
...
+    assert [{"Title", 111, now}, {"Title", 222, now}] == TestRepo.all(query)
+  end
+
+  @tag :ph
+  test "Repo.insert_all fills in placeholders" do
+    TestRepo.insert_all(Post, [%{title: {:placeholder, :foo}}], placeholders: %{foo: "Title"})
+
+    query = from(p in Post, select: p.title)
+    assert ["Title"] == TestRepo.all(query)
+  end
```
josevalim, line 962: You can remove this test, as it is the same as above but simpler?

integration_test/cases/repo.exs
```diff
@@ -953,6 +953,73 @@ defmodule Ecto.Integration.RepoTest do
...
+
+  @tag :ph
+  test "Repo.insert_all throws when placeholder key is used with invalid types" do
+    placeholders = %{string_key: "foo"}
+    entries = [%{visits: {:placeholder, :string_key}}]
+
+    assert_raise Ecto.ChangeError, fn ->
+      TestRepo.insert_all(Post, entries, placeholders: placeholders)
+    end
+  end
```
josevalim, line 1010: Tests that check it raises should not be integration tests. Instead, please move them to test/ecto/repo_test.exs (our unit tests).

lib/ecto/repo.ex
```diff
@@ -1204,7 +1205,7 @@ defmodule Ecto.Repo do
   Passing in a map for the `:placeholders` allows you to send less data over 
   the wire when you have many entries with the same value for a field. To use 
   a placeholder, replace its value in each of your entries with `{:placeholder, :key}`, 
-  where `:key` is the key you are using in the `:placeholders` option map.
+  where `:key` is the key you are using in the `:placeholders` option map. Placeholders require a database that supports index parameters, so it is not compatible with MySQL.
```
```suggestion
  where `:key` is the key you are using in the `:placeholders` option map.
  Placeholders require a database that supports index parameters, so it is
  not compatible with MySQL.
```

lib/ecto/repo/schema.ex
```diff
@@ -75,46 +75,66 @@ defmodule Ecto.Repo.Schema do
...
             {:placeholder, placeholder_key} ->
-              replaced_value = Map.fetch!(placeholder_map, placeholder_key)
-              value = dump_field!(:insert_all, schema, field, type, replaced_value, adapter)
-              {{source, value}, {Map.put(header, source, true), has_query?}}
+              placeholder_map = case placeholder_map do
+                %{^placeholder_key => {^type, _}} = map ->
+                  map
+
+                %{^placeholder_key => {_, _}} ->
+                  raise ArgumentError, "A placeholder key can only be used with columns of the same type. The key #{inspect(placeholder_key)} has already been dumped as a #{inspect(type)}."
```
josevalim:
```suggestion
                  raise ArgumentError,
                        "a placeholder key can only be used with columns of the same type. " <>
                          "The key #{inspect(placeholder_key)} has already been dumped as a #{inspect(type)}."
```

lib/ecto/repo/schema.ex
```diff
@@ -75,46 +75,66 @@ defmodule Ecto.Repo.Schema do
...
+
+                %{^placeholder_key => placeholder_val} = map ->
+                  value = dump_field!(:insert_all, schema, field, type, placeholder_val, adapter)
+                  Map.put(map, placeholder_key, {type, value})
+
+                map ->
+                  raise KeyError, "Placeholder key #{inspect(placeholder_key)} not found in #{inspect(map)}"
```
josevalim, line 128:
```suggestion
                  raise KeyError, "placeholder key #{inspect(placeholder_key)} not found in #{inspect(map)}"
```

lib/ecto/repo/schema.ex
```diff
@@ -889,8 +909,10 @@ defmodule Ecto.Repo.Schema do
   end
 
   defp dump_field!(action, schema, field, type, value, adapter) do
+
     case Ecto.Type.adapter_dump(adapter, type, value) do
       {:ok, value} ->
+        # Can the same input value diff types result in different output values from Ecto.Type.adapter_dump
```
josevalim:
```suggestion
    case Ecto.Type.adapter_dump(adapter, type, value) do
      {:ok, value} ->
```

lib/ecto/repo/schema.ex
```diff
@@ -75,46 +75,66 @@ defmodule Ecto.Repo.Schema do
...
             %Ecto.Query{} = query ->
-              {{source, query}, {Map.put(header, source, true), true}}
+              {{source, query}, {Map.put(header, source, true), true, placeholder_map}}
 
             {:placeholder, placeholder_key} ->
-              replaced_value = Map.fetch!(placeholder_map, placeholder_key)
-              value = dump_field!(:insert_all, schema, field, type, replaced_value, adapter)
-              {{source, value}, {Map.put(header, source, true), has_query?}}
+              placeholder_map = case placeholder_map do
+                %{^placeholder_key => {^type, _}} = map ->
```
josevalim, line 116: As it is implemented today, you can run into issues if for some reason there is a type that actually supports two element tuple as arguments. For example, imagine someone does this:

    insert_all([...], placeholders: %{foo: {:my_custom_type, :value}})

So we would need to keep the types and the values separately. I am also thinking that when we do this pass, we should already return the placeholders as a list and the entries in the insert all fields should already be `{:placeholder, index}`. So when we call the adapter, the simply pass a list with all placeholders.

So to be clear, we will call:

    adapter.insert_all(adapter_meta, schema_meta, Map.keys(header), rows, on_conflict, return_sources, placeholders, opts)

where placeholders is already a list:

    [Date.today, dumped_foo, dumped_bar, ...]

The rows will have entries like this:

    {:placeholder, 0}, {:placeholder, 1}, ... # starting from zero

Then in the Ecto.Adapters.SQL code you won't need to change unzip_inserts anymore, you will simply do this:

    placeholders ++ Enum.reverse(params) ++ on_conflict_values

And in Connection.insert you will receive all placeholders and you can compute the relative starting index by calling `length(placeholders)`.

This approach will make sure most of the placeholder logic is handled here, which is the shared code, and the placeholders logic in adapters is kept to a minimum.

josevalim:

PR #3519, opened 2020-12-21
Add CHANGELOG.md to docs extras

kianmeng:
- Add changelog to html doc
- Add additional badges


wojtekmach:
Thank you for this and or your other recent contributions! We don't use the extra badges in other projects so I went ahead and removed them from the PR.

PR #3523, opened 2020-12-24
Fix docs: `:on_replace` supports `:update` for `embeds_one`

mathieuprog:
Source: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/embedded.ex#L19


josevalim:

PR #3524, opened 2020-12-25
Add preload_order in associations

### LINKED ISSUE #3509: Support `order_by` in associations
Given PostgreSQL is not ordered by default, it would be convenient if we could easily specify how we want to order all our associations. The feature should be fairly straight-forward, the biggest issue here is the naming. `order_by` is not going to work with `assoc(parent, :association)` in queries. Only with preloaded queries and Ecto.assoc queries. We could call it `preload_order_by` but that doesn't encompass the `Ecto.assoc` case. Suggestions?

thiamsantos:
Closes #3509

### DIFF
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -552,6 +552,30 @@ defmodule Ecto.Integration.PreloadTest do
     assert p2.id == c4.post.id
   end
 
+  test "custom preload_order" do
+    p1 = TestRepo.insert!(%Post{title: "1"})
+
+    %Comment{id: cid1} = TestRepo.insert!(%Comment{text: "2", post_id: p1.id})
+    %Comment{id: cid2} = TestRepo.insert!(%Comment{text: "1", post_id: p1.id})
+
+    %User{id: uid1} = TestRepo.insert!(%User{name: "bar"})
+    %User{id: uid2} = TestRepo.insert!(%User{name: "foo"})
+
+    TestRepo.insert_all("posts_users", [[post_id: p1.id, user_id: uid1], [post_id: p1.id, user_id: uid2]])
+
+    p1 = TestRepo.preload(p1, [:ordered_comments, :ordered_users])
+
+    # asc
+    assert [c1, c2] = p1.ordered_comments
+    assert c1.id == cid2
+    assert c2.id == cid1
+
+    # desc
+    assert [u1, u2] = p1.ordered_users
+    assert u1.id == uid2
+    assert u2.id == uid1
+  end
+
   ## Others
 
   @tag :invalid_prefix
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -553,27 +553,18 @@ defmodule Ecto.Integration.PreloadTest do
   end
 
   test "custom preload_order" do
-    p1 = TestRepo.insert!(%Post{title: "1"})
+    post = TestRepo.insert!(%Post{users: [%User{name: "bar"}, %User{name: "foo"}], title: "1"})
 
-    %Comment{id: cid1} = TestRepo.insert!(%Comment{text: "2", post_id: p1.id})
-    %Comment{id: cid2} = TestRepo.insert!(%Comment{text: "1", post_id: p1.id})
+    TestRepo.insert!(%Comment{text: "2", post_id: post.id})
+    TestRepo.insert!(%Comment{text: "1", post_id: post.id})
 
-    %User{id: uid1} = TestRepo.insert!(%User{name: "bar"})
-    %User{id: uid2} = TestRepo.insert!(%User{name: "foo"})
-
-    TestRepo.insert_all("posts_users", [[post_id: p1.id, user_id: uid1], [post_id: p1.id, user_id: uid2]])
-
-    p1 = TestRepo.preload(p1, [:ordered_comments, :ordered_users])
+    post = TestRepo.preload(post, [:ordered_comments, :ordered_users])
 
     # asc
-    assert [c1, c2] = p1.ordered_comments
-    assert c1.id == cid2
-    assert c2.id == cid1
+    assert [%{text: "1"}, %{text: "2"}] = post.ordered_comments
 
     # desc
-    assert [u1, u2] = p1.ordered_users
-    assert u1.id == uid2
-    assert u2.id == uid1
+    assert [%{name: "foo"}, %{name: "bar"}] = post.ordered_users
   end
 
   ## Others
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -46,6 +46,7 @@ defmodule Ecto.Integration.Post do
     field :intensities, {:map, :float}
     field :posted, :date
     has_many :comments, Ecto.Integration.Comment, on_delete: :delete_all, on_replace: :delete
+    has_many :ordered_comments, Ecto.Integration.Comment, preload_order: [asc: :text]
     # The post<->permalink relationship should be marked as uniq
     has_one :permalink, Ecto.Integration.Permalink, on_delete: :delete_all, on_replace: :delete
     has_one :update_permalink, Ecto.Integration.Permalink, foreign_key: :post_id, on_delete: :delete_all, on_replace: :update
@@ -53,6 +54,7 @@ defmodule Ecto.Integration.Post do
     belongs_to :author, Ecto.Integration.User
     many_to_many :users, Ecto.Integration.User,
       join_through: "posts_users", on_delete: :delete_all, on_replace: :delete
+    many_to_many :ordered_users, Ecto.Integration.User, join_through: "posts_users", preload_order: [desc: :name]
     many_to_many :unique_users, Ecto.Integration.User,
       join_through: "posts_users", unique: true
     many_to_many :constraint_users, Ecto.Integration.User,
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -46,7 +46,7 @@ defmodule Ecto.Integration.Post do
     field :intensities, {:map, :float}
     field :posted, :date
     has_many :comments, Ecto.Integration.Comment, on_delete: :delete_all, on_replace: :delete
-    has_many :ordered_comments, Ecto.Integration.Comment, preload_order: [asc: :text]
+    has_many :ordered_comments, Ecto.Integration.Comment, preload_order: [:text]
     # The post<->permalink relationship should be marked as uniq
     has_one :permalink, Ecto.Integration.Permalink, on_delete: :delete_all, on_replace: :delete
     has_one :update_permalink, Ecto.Integration.Permalink, foreign_key: :post_id, on_delete: :delete_all, on_replace: :update
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -563,7 +563,7 @@ defmodule Ecto.Association.Has do
   @has_one_on_replace_opts @on_replace_opts ++ [:update]
   defstruct [:cardinality, :field, :owner, :related, :owner_key, :related_key, :on_cast,
              :queryable, :on_delete, :on_replace, where: [], unique: true, defaults: [],
-             relationship: :child, ordered: false]
+             relationship: :child, ordered: false, preload_order: []]
 
   @impl true
   def after_compile_validation(%{queryable: queryable, related_key: related_key}, env) do
@@ -627,6 +627,12 @@ defmodule Ecto.Association.Has do
       raise ArgumentError, "expected `:where` for #{inspect name} to be a keyword list, got: `#{inspect where}`"
     end
 
+    preload_order = opts[:preload_order] || []
+
+    unless is_list(preload_order) do
+      raise ArgumentError, "expected `:preload_order` for #{inspect name} to be a keyword list, got: `#{inspect preload_order}`"
+    end
+
     %__MODULE__{
       field: name,
       cardinality: cardinality,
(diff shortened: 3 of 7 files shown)


integration_test/cases/preload.exs
```diff
@@ -552,6 +552,30 @@ defmodule Ecto.Integration.PreloadTest do
     assert p2.id == c4.post.id
   end
 
+  test "custom preload_order" do
```
thiamsantos, line 555: This is enough for tests? Makes sense to add unit tests for this new option?

lib/ecto/association.ex
```diff
@@ -627,6 +627,12 @@ defmodule Ecto.Association.Has do
       raise ArgumentError, "expected `:where` for #{inspect name} to be a keyword list, got: `#{inspect where}`"
     end
 
+    preload_order = opts[:preload_order] || []
+
+    unless is_list(preload_order) do
```
thiamsantos, line 632: The format of the `preload_order` should be validated?

thiamsantos, line 632: As the has_one and has_many use the same struct, is necessary to validate that the preload_order option is accepted only for `cardinality: :many`?

lib/ecto/schema.ex
```diff
@@ -1910,7 +1910,7 @@ defmodule Ecto.Schema do
     Module.put_attribute(mod, :changeset_fields, {name, {:assoc, struct}})
   end
 
-  @valid_many_to_many_options [:join_through, :join_defaults, :join_keys, :on_delete, :defaults, :on_replace, :unique, :where, :join_where]
+  @valid_many_to_many_options [:join_through, :join_defaults, :join_keys, :on_delete, :defaults, :on_replace, :unique, :where, :join_where, :preload_order]
```
thiamsantos, line 1913: The line seems to be too long, should I break in two lines?

lib/ecto/repo/preloader.ex
```diff
@@ -272,6 +272,14 @@ defmodule Ecto.Repo.Preloader do
     We expected a tuple but we got: #{inspect(entry)}
     """
 
+  defp preload_order(assoc, query, related_field) do
+    custom_order_by = assoc.preload_order |> Enum.map(fn {direction, field} ->
```
thiamsantos, line 276: It is safe to assume that if cardinality is `many`, it will either be a `Ecto.Association.ManyToMany` or `Ecto.Association.Has`?

integration_test/support/schemas.exs
```diff
@@ -46,13 +46,15 @@ defmodule Ecto.Integration.Post do
     field :intensities, {:map, :float}
     field :posted, :date
     has_many :comments, Ecto.Integration.Comment, on_delete: :delete_all, on_replace: :delete
+    has_many :ordered_comments, Ecto.Integration.Comment, preload_order: [asc: :text]
```
Eiji7, line 49: I would use it like:
```suggestion
    has_many :ordered_comments, Ecto.Integration.Comment, preload_order: [:text]
```
because `:text` should work as `[asc: :text]`, so we would test 2 things (short notation + ascending order) at same time.

Maybe we want to strictly test all 3 cases i.e. `[:text]`, `[asc: :text]` and `[desc: :text]`, but at least 2 of them (`[:text]` and `[desc: :text]`) should be tested.

thiamsantos: @Eiji7 I added support for the short notation. WDYT?

integration_test/cases/preload.exs
```diff
@@ -552,6 +552,30 @@ defmodule Ecto.Integration.PreloadTest do
...
+    %User{id: uid2} = TestRepo.insert!(%User{name: "foo"})
+
+    TestRepo.insert_all("posts_users", [[post_id: p1.id, user_id: uid1], [post_id: p1.id, user_id: uid2]])
+
+    p1 = TestRepo.preload(p1, [:ordered_comments, :ordered_users])
+
+    # asc
+    assert [c1, c2] = p1.ordered_comments
+    assert c1.id == cid2
+    assert c2.id == cid1
```
Eiji7, line 571: I believe this one is not only shorter, but also clearly shows ordered results:
```suggestion
    assert [%{text: "1"}, %{text: "2"}] = p1.ordered_comments
```
Also with that change we do not need `cid1` and `cid2`.

integration_test/cases/preload.exs
```diff
@@ -552,6 +552,30 @@ defmodule Ecto.Integration.PreloadTest do
     assert p2.id == c4.post.id
   end
 
+  test "custom preload_order" do
+    p1 = TestRepo.insert!(%Post{title: "1"})
```
Eiji7, line 556:
```suggestion
    p = TestRepo.insert!(%Post{title: "1"})
```
We have only one `p*`, so how about using `p` or `post` instead of `p1`?

integration_test/cases/preload.exs
Eiji7: I think so

integration_test/cases/preload.exs
```diff
@@ -552,6 +552,30 @@ defmodule Ecto.Integration.PreloadTest do
...
+
+    # asc
+    assert [c1, c2] = p1.ordered_comments
+    assert c1.id == cid2
+    assert c2.id == cid1
+
+    # desc
+    assert [u1, u2] = p1.ordered_users
+    assert u1.id == uid2
+    assert u2.id == uid1
```
Eiji7, line 576: Same as above:
```suggestion
    assert [%{name: "foo"}, %{name: "bar"}] = p1.ordered_users
```

integration_test/cases/preload.exs
```diff
@@ -552,6 +552,30 @@ defmodule Ecto.Integration.PreloadTest do
...
+  test "custom preload_order" do
+    p1 = TestRepo.insert!(%Post{title: "1"})
+
+    %Comment{id: cid1} = TestRepo.insert!(%Comment{text: "2", post_id: p1.id})
+    %Comment{id: cid2} = TestRepo.insert!(%Comment{text: "1", post_id: p1.id})
+
+    %User{id: uid1} = TestRepo.insert!(%User{name: "bar"})
+    %User{id: uid2} = TestRepo.insert!(%User{name: "foo"})
+
+    TestRepo.insert_all("posts_users", [[post_id: p1.id, user_id: uid1], [post_id: p1.id, user_id: uid2]])
```
Eiji7, line 564: Please correct me if I'm wrong (writing from memory), but we can write it much shorter:
```suggestion
    TestRepo.insert!(%PostUser{post_id: p1.id, user: %{name: "bar"}})
    TestRepo.insert!(%PostUser{post_id: p1.id, user: %{name: "foo"}})
```
or even:
```suggestion
    p1 = TestRepo.insert!(%Post{users: [%{name: "bar"}, %{name: "foo"}], title: "1"})
```

lib/ecto/association.ex
Eiji7: Good question … of course it's up to maintainers, but I guess that if we have `where` validation looks same then it's all ok.

Eiji7: Good question … of course it's up to maintainers, but I guess that if `where` validation looks same then it's all ok.

Eiji7: Good question … of course it's up to maintainers, but I guess that if `where` validation looks same then it's all ok.

integration_test/cases/preload.exs
Eiji7: I think so

lib/ecto/schema.ex
josevalim: No, it is fine, it can be done later.

lib/ecto/association.ex
josevalim: It should. It should be a list of atoms and 2-elem tuples where the first element is either desc or asc.

integration_test/cases/preload.exs
josevalim: We generally avoid integration tests. Is it possible for this feature to be tested with unit tests?

lib/ecto/association.ex
josevalim: If you can validate the field exists, then it is even better!

thiamsantos: > element is either desc or asc

@josevalim The other directions should also be supported,right? Like `asc_nulls_first` and `desc_nulls_last`.

> If you can validate the field exists, then it is even better!

Any direction on how we could implement this? My first instinct was to call `related.__schema__(:fields)`, but the `__schema__` seems to be available only at runtime.

josevalim: > @josevalim The other directions should also be supported,right? Like asc_nulls_first and desc_nulls_last.

Yes, good call!

> Any direction on how we could implement this? My first instinct was to call related.__schema__(:fields), but the __schema__ seems to be available only at runtime.

Hrm, I misremembered. We actually don't validate fields yet. So don't worry about it, just validate the format. :)

lib/ecto/association.ex
```diff
@@ -555,6 +555,7 @@ defmodule Ecto.Association.Has do
     * `on_replace` - The action taken on associations when schema is replaced
     * `defaults` - Default fields used when building the association
     * `relationship` - The relationship to the specified schema, default is `:child`
+    * `preload_order` - Default `order_by` of the association, used by preload
   """
```
v0idpwn, line 559: I think it would be good to specify that this is used ONLY by preload (will it be used by assoc? if yes, it should be mentioned here too) and not by other queries. Or maybe link to the longer documentation snippet.

lib/ecto/association.ex
v0idpwn: Maybe we can do it in the __after_compile__ calback? Not sure about it, though.

v0idpwn: Maybe we can do it in the after_compile calback? Not sure about it, though.

integration_test/cases/preload.exs
thiamsantos: Thanks for the suggestion @Eiji7. 
I refactored the tests based on your suggestions, it seems to be much cleaner now. 
WDYT?

integration_test/cases/preload.exs
thiamsantos: Changed to use `post`. Thanks @Eiji7

integration_test/cases/preload.exs
thiamsantos: @Eiji7 Updated to use the second option. Thanks for the suggestion.

lib/ecto/association.ex
thiamsantos:
```diff
-     * `preload_order` - Default `order_by` of the association, used by preload
+     * `preload_order` - Default `order_by` of the association, used only by preload
``
@v0idpwn Do you think adding the `only` would be enough, or we should be even more explicit about that?

Comment (v0idpwn) [reply lib/ecto/association.ex:592]: only lgtm

File: integration_test/cases/preload.exs (hunk 6 above)
Comment (Eiji7) [reply integration_test/cases/preload.exs]: Definitely better!

File: lib/ecto/association.ex (hunk 2 above)
Comment (Eiji7) [reply lib/ecto/association.ex]: @josevalim Do you think we should add tests for `asc_nulls_first` and `desc_nulls_first` as well?

File: lib/ecto/association.ex (hunk 10 above)
Comment (thiamsantos) [reply lib/ecto/association.ex:592]: Done, thanks @v0idpwn

File: lib/ecto/association.ex [hunk 11]
```diff
@@ -382,6 +383,38 @@ defmodule Ecto.Association do
...
+            "got: `#{inspect preload_order}` " <>
+            "#{inspect direction} is invalid"
+        end
+
+        {direction, field}
+
+      item ->
+        raise ArgumentError,
+          "expected `:preload_order` for #{inspect name} to be a keyword list, " <>
+            "got: `#{inspect preload_order}` " <>
```
v0idpwn, line 407: Maybe "keyword list or a list of atoms/fields (?)"

lib/ecto/association.ex
```diff
@@ -382,6 +383,38 @@ defmodule Ecto.Association do
               "expected defaults for #{inspect name} to be a keyword list " <>
                 "or a {module, fun, args} tuple, got: `#{inspect defaults}`"
 
+  @doc """
+  Validates `preload_order` for association named `name`.
+  """
+  def validate_preload_order!(name, preload_order) when is_list(preload_order) do
+    Enum.map(preload_order, fn
+      field when is_atom(field) ->
```
v0idpwn, line 391: I think it would be great if we managed to validate these fields on `__after_compile__`, if possible, but I'm not sure if it is.

lib/ecto/association.ex
```diff
@@ -382,6 +383,38 @@ defmodule Ecto.Association do
...
+        field
+
+      {direction, field} when is_atom(direction) and is_atom(field) ->
+        unless OrderBy.valid_direction?(direction) do
+          raise ArgumentError,
+          "expected `:preload_order` for #{inspect name} to be a keyword list, " <>
+            "got: `#{inspect preload_order}` " <>
+            "#{inspect direction} is invalid"
```
v0idpwn, line 399:
```suggestion
            "#{inspect direction} is not a valid direction"
```

v0idpwn (approved):
Overall, this looks really good for me.  Dropped some minor comments.

lib/ecto/association.ex
thiamsantos: I think would be nice to validate the fields. But maybe we should introduce to other places as well, like the `related_key` option. WDYT @v0idpwn ?

@josevalim There is any plans to introduce field validation at compile time?

thiamsantos:
@josevalim @v0idpwn @Eiji7 There is anything do you think would be a good addition to this PR? Anything missing?

Eiji7 (approved):

josevalim:
Beautiful work, thank you!

josevalim:

PR #3525, opened 2020-12-26
[do not merge]  Loading nil value from DB for array of parameterized type should resolve into empty list

mathieuprog:
Just like loading a `nil` value from DB for an `embeds_many` field is resolving into an empty list, I expect arrays of parameterized types to behave the same.

However this test seems to contradict that :
https://github.com/elixir-ecto/ecto/blob/v3.5.5/test/ecto/parameterized_type_test.exs#L107
even though my addition will not make the tests fail.

I noticed this discrepancy when working on my lib `polymorphic_embed` where I want to comply with the behavior of `embeds_one`/`embeds_many`. That is, if a `nil` value from the DB is loaded for an `embeds_many`, it resolves into `[]`. But for an array of polymorphic embeds, the `nil` value from DB just stays `nil`, and that is a difference I cannot solve in the library.

This would solve it:
```
  def adapter_load(adapter, {:array, {:parameterized, _, _}}, nil) do
    {:ok, []}
  end
  ```

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -870,6 +870,9 @@ defmodule Ecto.Type do
   def adapter_load(adapter, {:parameterized, module, params} = type, value) do
     process_loaders(adapter.loaders(module.type(params), type), {:ok, value}, adapter)
   end
+  def adapter_load(adapter, {:array, {:parameterized, _, _}}, nil) do
+    {:ok, []}
+  end
   def adapter_load(_adapter, _type, nil) do
     {:ok, nil}
   end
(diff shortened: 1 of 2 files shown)


josevalim:
It is by design. Parameterized types are responsible for handling nil, so if is up to them to implement embeds_many semantics or not.

mathieuprog:
Ok but the `load` callback will never be called in case the field is `nil` for an array of parametrized types. Even if it would, I can't know if the field is an array or not in the `load` callback, to return `[]` or `nil` respectively.

josevalim:
The array is nil or is there a nil inside the array?

mathieuprog:
The array is nil in db. So yeah it makes sense that `load` is not called, but I can't mimic the behavior of `embeds_many`, where if the array is `nil` in db, it resolves to `[]`.

josevalim:
I am not sure that’s all bad. We wanted to fix it for embeds_many but it would be a breaking change, so we didn’t. To solve this we would need a new callback.

mathieuprog:
Yah probably not a big deal anyway. Thank you for your attention 🙏

PR #3528, opened 2020-12-28
Add bang UUID conversion methods

pawurb:
This PR adds bang versions of `load` and `dump` methods for converting UUID string to binary and other way around. I found them missing when doing UUID queries in console. I.e:

```
Repo.all(from a in Area, where: a.user_id == ^(Ecto.UUID.dump("6bfb65d2-a539-44b7-866b-53f764da3a91") |> elem(1)))
```

instead of a simpler:

```
Repo.all(from a in Area, where: a.user_id == ^(Ecto.UUID.dump!("6bfb65d2-a539-44b7-866b-53f764da3a91"))
```

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -134,6 +134,17 @@ defmodule Ecto.UUID do
   defp d(_),  do: throw(:error)
 
   @doc """
+  Same as `dump/1` but raises `Ecto.ArgumentError` on invalid arguments.
+  """
+  @spec dump!(t | raw | any) :: t
+  def dump!(value) do
+    case dump(value) do
+      {:ok, uuid} -> uuid
+      :error -> raise ArgumentError, "cannot dump given UUID to binary: #{inspect(value)}"
+    end
+  end
+
+  @doc """
   Converts a binary UUID into a string.
   """
   @spec load(raw | any) :: {:ok, t} | :error
@@ -145,6 +156,17 @@ defmodule Ecto.UUID do
                          "Maybe you wanted to declare :uuid as your database field?"
   end
   def load(_), do: :error
+
+  @doc """
+  Same as `load/1` but raises `Ecto.ArgumentError` on invalid arguments.
+  """
+  @spec load!(t | raw | any) :: t
+  def load!(value) do
+    case load(value) do
+      {:ok, uuid} -> uuid
+      :error -> raise ArgumentError, "cannot load given binary as UUID: #{inspect(value)}"
+    end
+  end
 
   @doc """
   Generates a version 4 (random) UUID.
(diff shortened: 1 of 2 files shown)


lib/ecto/uuid.ex
```diff
@@ -133,6 +133,17 @@ defmodule Ecto.UUID do
...
   defp d(_),  do: throw(:error)
 
+  @doc """
+  Same as `dump/1` but raises `Ecto.ArgumentError` on invalid arguments.
+  """
+  @spec dump!(t | raw | any) :: t
+  def dump!(value) do
+    case dump(value) do
+      {:ok, uuid} -> uuid
+      :error -> raise ArgumentError, "Cannot convert provided argument to binary."
```
josevalim, line 143:
```suggestion
      :error -> raise ArgumentError, "cannot dump given UUID to binary"
```

lib/ecto/uuid.ex
```diff
@@ -146,6 +157,17 @@ defmodule Ecto.UUID do
...
   def load(_), do: :error
 
+  @doc """
+  Same as `load/1` but raises `Ecto.ArgumentError` on invalid arguments.
+  """
+  @spec load!(t | raw | any) :: t
+  def load!(value) do
+    case load(value) do
+      {:ok, uuid} -> uuid
+      :error -> raise ArgumentError, "Cannot convert provided argument to UUID string."
```
josevalim, line 167:
```suggestion
      :error -> raise ArgumentError, "cannot load given binary as UUID"
```

josevalim:

wojtekmach:
I think this is a useful change but note that if you've specified `:user_id` as `Ecto.UUID`, you should be able to simply do:

```elixir
Repo.all(from a in Area, where: a.user_id == ^"6bfb65d2-a539-44b7-866b-53f764da3a91")
```

and it should have the same effect. (and if the given uuid is invalid, you'd get a proper error)

pawurb:
@wojtekmach I'm afraid it is not the case: 

While:

```
Repo.get_by(Area, %{id: "49dc38b3-504a-4807-abea-da1f3b5fc2f9"})
Repo.get(Area, "49dc38b3-504a-4807-abea-da1f3b5fc2f9")
```
both work.

but

```
Repo.all(from a in Area, where: a.id == "6bfb65d2-a539-44b7-866b-53f764da3a91")
```

raises:
```
** (EXIT from #PID<0.427.0>) shell process exited with reason: an exception was raised:
```

Maybe this case could also be fixed?

wojtekmach:
> `where: a.id == "6bfb65d2-a539-44b7-866b-53f764da3a91"`

yeah, this only works when interpolated:

```elixir
where: a.id == ^"6bfb65d2-a539-44b7-866b-53f764da3a91"
```

And yeah we should at least fix the error message to not be terrible.

pawurb:
Right, missed the `^`

michalmuskala:
So, do we really need this if it seems the function is not needed for the original use case that motivated the PR?

pawurb:
`cast` supports bang version. I'd leave it for API consistency.

PR #3531, opened 2020-12-30
Human readable UUID logs

pawurb:
Currently UUID values displayed in logs are far from useful without manual conversion:

image

this PR displays them in the correct format:

image

However it does not work correctly now when applied to a real project. It ends up in an endless loop. Any tips on how to implement it correctly would be welcome. Maybe adding a custom binary type for UUID values could do the trick?

I'm not sure how to proceed but I think this change would be really useful for all the projects using UUIDs.

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -228,3 +228,12 @@ defmodule Ecto.UUID do
   defp e(14), do: ?e
   defp e(15), do: ?f
 end
+
+defimpl Inspect, for: BitString do
+  def inspect(bit_string, _opts) do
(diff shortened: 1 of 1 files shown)


josevalim:
We should definitely not override the inspect implementation for bitstring. The proper way to implement this would be to pass some information to the log function about the type of parameters... so we can convert it back to UUID, as by the time logging happens, the original value (the UUID) has been long gone.

tlvenn:
@josevalim can we re open this please ? I believe this is something that needs to be addressed, right now dealing with UUIDs in the log is pretty painful and time consuming.

PR #3533, opened 2020-12-31
Allow nil values on typed maps

v0idpwn:
Array has a skip_nil? argument, but I don't see why it is needed, so
I didn't add it to map. If wanted, it could be done pretty easily.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -493,7 +493,7 @@ defmodule Ecto.Type do
 
   def dump({:array, {_, _, _} = type}, value, dumper), do: array(value, type, dumper, false, [])
   def dump({:array, type}, value, dumper), do: array(value, type, dumper, true, [])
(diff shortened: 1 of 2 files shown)


v0idpwn:
ok, just found a problem

lib/ecto/type.ex
```diff
@@ -1216,6 +1216,10 @@ defmodule Ecto.Type do
     :error
   end
 
+  defp map_each([{key, nil} | t], fun, acc) do
+    map_each(t, fun, Map.put(acc, key, nil))
```
josevalim, line 1220: If the type is parameterized, we should still call the type's callback, that's why we have skip_nil?, so we need similar here.

v0idpwn: You're absolutely correct. Just fixed it. Happy new year :)

josevalim:

Issue #3138, opened 2019-10-11
Ecto.Changeset.validate_subset enumerates on the wrong variable


mindreader:
Elixir version: 1.9.0
Ecto version: 3.1.7 (but the code is the same in master)

This line in deps/ecto/lib/ecto/changeset.ex looks really wrong to me.
```
case Enum.any?(value, fn(x) -> not(x in data) end) do
```

It should be
```
case value not in data do
```

Which fixes the error.  This code has been this way a long time.  I'm surprised that I'm the first to discover this, maybe I'm doing something wrong, but it doesn't seem like it.

```
defmodule Marker do
  schema "markers" do  
    field(:type, :string)
  end
end

%Marker{}
 |> Ecto.Changeset.cast(%{type: "pre"}, [:type])
 |> Ecto.Changeset.validate_subset(:type, ["pre", "post", "mid"])
```

### Current behavior

```
** (Protocol.UndefinedError) protocol Enumerable not implemented for "pre" of type BitString. This protocol is implemented for the following type(s): Ecto.Adapters.SQL.Stream, Postgrex.Stream, DBConnection.Stream, DBConnection.PrepareStream, Timex.Interval, Stream, List, Range, MapSet, Date.Range, IO.Stream, HashDict, File.Stream, HashSet, GenEvent.Stream, Map, Function
    (elixir) lib/enum.ex:1: Enumerable.impl_for!/1
    (elixir) lib/enum.ex:141: Enumerable.reduce/3
    (elixir) lib/enum.ex:366: Enum.any?/2
    (ecto) deps/ecto/lib/ecto/changeset.ex:1860: anonymous fn/5 in Ecto.Changeset.validate_subset/4
    (ecto) deps/ecto/lib/ecto/changeset.ex:1578: Ecto.Changeset.validate_change/3
```

### Expected behavior
```
#Ecto.Changeset<
  action: nil,
  changes: %{type: "pre"},
  errors: [],
  data: #Simple.Schema.Marker<>,
  valid?: true
>
```

wojtekmach:
> ```elixir
> defmodule Marker do
>   schema "markers" do  
>     field(:type, :string)
>   end
> end
>
> %Marker{}
>  |> Ecto.Changeset.cast(%{type: "pre"}, [:type])
>  |> Ecto.Changeset.validate_subset(:type, ["pre", "post", "mid"])
> ```

[...]

PR #3542, opened 2021-01-18
Improve docs at Ecto.Changeset.validate_subset/4

v0idpwn:
Although `subset` seems clear enough, it may be easy to mistake the
use-case for this function with `validate_inclusion`, as shown in #3138.
This commit removes the unnecessary reference of `validate_inclusion/4`
and adds a tip about the `validate_inclusion/4` use-case.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1970,8 +1970,10 @@ defmodule Ecto.Changeset do
   end
 
   @doc ~S"""
-  Validates a change, of type enum, is a subset of the given enumerable. Like
-  `validate_inclusion/4` for lists.
+  Validates a change, of type enum, is a subset of the given enumerable. 
+
+  If you need to validate if a value is inside a given enumerable, you
+  should use `validate_inclusion/4` instead.
 
   ## Options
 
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -1970,8 +1970,10 @@ defmodule Ecto.Changeset do
   end
 
   @doc ~S"""
-  Validates a change, of type enum, is a subset of the given enumerable. Like
-  `validate_inclusion/4` for lists.
+  Validates a change, of type enum, is a subset of the given enumerable. 
+
+  If you need to validate if a value is inside a given enumerable, you
```
josevalim, line 1975:
```suggestion
  This validates a list of values against a list of values. If you need
  to validate if a single value is inside a given enumerable, you
```

lib/ecto/changeset.ex
```diff
@@ -1970,8 +1970,10 @@ defmodule Ecto.Changeset do
...
+  should use `validate_inclusion/4` instead.
```
josevalim, line 1976:
```suggestion
  This validates if a list of values belongs to the given enumerable.
  If you need to validate if a single value is inside the given enumerable,
  you should use `validate_inclusion/4` instead.
```

josevalim:

PR #3546, opened 2021-01-19
Support file and line interpolation on ECTO_EDITOR

### LINKED ISSUE #3355: Make ECTO_EDITOR similar to ELIXIR_EDITOR and PLUG_EDITOR
If we happen to use all these variables, we might have this:

```
export ELIXIR_EDITOR="vim +__LINE__ __FILE__"
export PLUG_EDITOR=$ELIXIR_EDITOR
export ECTO_EDITOR="vim"
```

I think it would be good to unify, so we could do:

```
export ECTO_EDITOR=$ELIXIR_EDITOR
```

and we can do it in a backwards compatible way. Besides unification, a minor side benefit is we could set the `__LINE__` so e.g. when generating the migration we put the cursor below `def change`.

thiamsantos:
Closes #3355

### DIFF
--- a/lib/mix/ecto.ex
+++ b/lib/mix/ecto.ex
@@ -121,7 +121,7 @@ defmodule Mix.Ecto do
           "#{editor} #{inspect(file)}:#{line}"
         end
 
-      :os.cmd(String.to_charlist(command))
+      Mix.shell().cmd(command)
       true
     else
       false
--- a/lib/mix/ecto.ex
+++ b/lib/mix/ecto.ex
@@ -86,12 +86,42 @@ defmodule Mix.Ecto do
 
   @doc """
   Asks if the user wants to open a file based on ECTO_EDITOR.
+
+  By default, it attempts to open the file and line using the
+  `file:line` notation. For example, if your editor is called
+  `subl`, it will open the file as:
+
+      subl path/to/file:line
+
+  It is important that you choose an editor command that does
+  not block nor that attempts to run an editor directly in the
+  terminal. Command-line based editors likely need extra
+  configuration so they open up the given file and line in a
+  separate window.
+
+  Custom editors are supported by using the `__FILE__` and
+  `__LINE__` notations, for example:
+
+      ECTO_EDITOR="my_editor +__LINE__ __FILE__"
+
+  and Elixir will properly interpolate values.
+
   """
-  @spec open?(binary) :: boolean
-  def open?(file) do
+  @spec open?(binary, non_neg_integer) :: boolean
+  def open?(file, line \\ 1) do
     editor = System.get_env("ECTO_EDITOR") || ""
+
     if editor != "" do
-      :os.cmd(to_charlist(editor <> " " <> inspect(file)))
+      command =
+        if editor =~ "__FILE__" or editor =~ "__LINE__" do
+          editor
+          |> String.replace("__FILE__", inspect(file))
+          |> String.replace("__LINE__", Integer.to_string(line))
+        else
+          "#{editor} #{inspect(file)}:#{line}"
+        end
+
+      :os.cmd(String.to_charlist(command))
       true
     else
       false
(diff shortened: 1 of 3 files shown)


lib/mix/ecto.ex
```diff
@@ -86,12 +86,42 @@ defmodule Mix.Ecto do
...
+
     if editor != "" do
-      :os.cmd(to_charlist(editor <> " " <> inspect(file)))
+      command =
+        if editor =~ "__FILE__" or editor =~ "__LINE__" do
+          editor
+          |> String.replace("__FILE__", inspect(file))
+          |> String.replace("__LINE__", Integer.to_string(line))
+        else
+          "#{editor} #{inspect(file)}:#{line}"
```
thiamsantos, line 121: I believe that would be a breaking change, should I remove the line from to behave the same way as before?

lib/mix/ecto.ex
```diff
@@ -86,12 +86,42 @@ defmodule Mix.Ecto do
...
+        end
+
+      :os.cmd(String.to_charlist(command))
```
thiamsantos, line 124: I couldn't find a simple way to test it.

wojtekmach: I'd do it the same way IEx.Helpers.open/1 is tested: https://github.com/elixir-lang/elixir/blob/b662d32fc3171537778aa896a0b4fd65af06d354/lib/iex/test/iex/helpers_test.exs#L147-L156

lib/mix/ecto.ex
wojtekmach: Good question, not sure. I'd go with what you have right now because it matches Elixir behaviour.

lib/mix/ecto.ex
thiamsantos: I tried using something similar but it doesn't seem to work. I don't know if I'm missing something. 

```
ExUnit.CaptureIO.capture_io(fn ->
  :os.cmd('echo "something"')
end)
``` 
image

I tried a different approach by creating a file in tmp directory. WDYT?

azizk: Hey, I was just curious and looked this up. So `:os.cmd()` already captures the output and returns it as a string. That's why `capture_io()` doesn't return anything.

You could actually write the string to stdout, like so: `:os.cmd(to_charlist(command)) |> IO.write()`
That should make the test pass :+1:. The echo command will return a trailing newline btw, e.g.: `"something\n"`

wojtekmach, line 124:
```suggestion
      Mix.shell().cmd(command)
```

test/mix/ecto_test.exs
```diff
@@ -34,4 +34,34 @@ defmodule Mix.EctoTest do
...
+    path = Path.join(@tmp_path, to_string(path))
+
+    try do
+      File.rm_rf!(path)
+      File.mkdir_p!(path)
+      File.cd!(path, fun)
+    after
+      File.rm_rf!(path)
+    end
+  end
```
wojtekmach, line 66:
```suggestion
  describe "open?/2" do
    @editor System.get_env("ECTO_EDITOR")

    test "opens __FILE__ and __LINE__" do
      System.put_env("ECTO_EDITOR", "echo -n __FILE__:__LINE__")
      open?("lib/some/file.ex", 4)
      assert_received {:mix_shell, :run, ["-n lib/some/file.ex:4\n"]}
    after
      System.put_env("ECTO_EDITOR", @editor)
    end
  end
```

wojtekmach (approved):
Thank you!

I went ahead and changed :os.cmd to Mix.shell().cmd so that we can more easily test it.
Once this land, we should make similar changes to ecto_sql for `mix ecto.gen.migration`.

lib/mix/ecto.ex
josevalim: I would keep it as is, yes. I think it should be fine.

PR #3548, opened 2021-01-25
Many to many

f0rest8:
Hi,

I am submitting this pull request with an example for Ecto's `many_to_many` function docs on self-referencing `many_to_many` associations.

Thank you ♥

-Mark


guides/howtos/Polymorphic associations with many to many.md
```diff
@@ -401,3 +401,167 @@ end
 Overall our code looks structurally the same as `has_many` would, although at the database level our relationships are expressed with join tables.
 
 While in this guide we changed our code to cope with the parameter format required by `cast_assoc`, in [Constraints and Upserts](Constraints and Upserts.md) we drop `cast_assoc` altogether and use `put_assoc` which brings more flexibilities when working with associations.
+
+## Polymorphism with a self-referencing many_to_many
```
josevalim, line 405: I think this can be a separate guide called "Self-referencing many to many.md". What do you think? :)

josevalim:
This looks great, thank you! I have added just one comment!

guides/howtos/Polymorphic associations with many to many.md
Eiji7: We have `## Summary` section below, so if doing this change I suggest also to change mentioned section to `### Summary`

josevalim: If this is a separate guide, then `## Summary` is correct, as all other guides. :)

f0rest8: That sounds great! I'll make it a separate guide :)

f0rest8:
Okay, I separated it into another guide and gave it a `#Self-referencing many to many` title with `##People relationships `subheading. 😊

josevalim:

PR #3549, opened 2021-01-25
Update Self-referencing many to many.md

f0rest8:
Hi,

I think the way the example is, and how we've implemented showcasing the separation of concerns, we would need to `alias MyApp.Accounts.Person` because we have the Person module separated from the Relationships context by being under the Accounts module. 😊

♥


josevalim:
Oops, my bad. Good catch!

josevalim:

PR #3553, opened 2021-02-06
Ecto.Query.dynamic/2 breaks bindings order when subquery present and binded values are not literals

AndrewDryga:
I've added a test to showcase a bug, it was discovered during refactoring where we started to use dynamic to filter our queries and tests failed because `dynamic(^left and ^right)` produced incorrect value:
```
# Test code
IO.puts(" * left #{inspect(left)} \n * right #{inspect(right)}")
merged = dynamic(^left and ^right)
IO.puts("merged = #{inspect(merged)}")

# Result 
 * left dynamic([conference: conference], conference.id in {:subquery, 0}) 
 * right dynamic([conference: conference], conference.has_active_calls == ^false)
merged = dynamic([conference: conference], conference.id in {:subquery, 0} and conference.has_active_calls == ^:subquery)
```

^ here we can see that `has_active_calls` which had a boolean value is off and started to use subquery from the left part. Initially, I felt like it's an inspect protocol bug but looking further here is a `BoolExpr` from a query that was built:

```
%{
    __struct__: Ecto.Query.BooleanExpr,
    expr: {:and, [],
    [
        {:and, [],
        [
        {:in, [],
            [{{:., [], [{:&, [], [0]}, :id]}, [], []}, {:subquery, 0}]},
        {:==, [],
            [
            {{:., [], [{:&, [], [0]}, :has_active_calls]}, [], []},
            {:^, [], [0]}
            ]}
        ]},
        {:not, [],
        [
        {:in, [],
            [
            {{:., [], [{:&, [], [0]}, :owner_assignment_id]}, [], []},
            {:^, [], [1]}
            ]}
        ]}
    ]},
    op: :and,
    params: [
        {:subquery, 0},
        {false, {0, :has_active_calls}},
        {["7984183b-4f64-4a08-b8a6-36788957c22c"],
        {:in, {0, :owner_assignment_id}}}
    ],
    subqueries: [
        %{
            __struct__: Ecto.SubQuery,
            ...
        }
    ]
}
```

^ here we can see that `params` in `Ecto.Query` struct is a merged result from params and subqueries, where subqueries took first place. It shifted positions for rest of bindings but in `expr` we have an AST that has position-dependent bindings that are off now. 

And an exception that would be raised if such bug happens:
```
       :erlang.length(false)
       (elixir 1.11.2) lib/enum.ex:1521: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir 1.11.2) lib/enum.ex:1521: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir 1.11.2) lib/enum.ex:1521: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir 1.11.2) lib/enum.ex:2181: Enum."-reduce/3-lists^foldl/2-0-"/3
       (elixir 1.11.2) lib/enum.ex:2181: Enum."-reduce/3-lists^foldl/2-0-"/3
       (ecto 3.5.5) lib/ecto/adapter/queryable.ex:110: Ecto.Adapter.Queryable.prepare_query/3
       (ecto_sql 3.5.3) lib/ecto/adapters/sql.ex:235: Ecto.Adapters.SQL.to_sql/3
```

P.S.: I've added much more tests than needed but because I've already written them decided to keep them if you would like to merge them too.

### DIFF
--- a/lib/ecto/query/builder/dynamic.ex
+++ b/lib/ecto/query/builder/dynamic.ex
@@ -59,6 +59,7 @@ defmodule Ecto.Query.Builder.Dynamic do
           {%Ecto.Query.DynamicExpr{binding: new_binding} = dynamic, _} ->
             binding = if length(new_binding) > length(binding), do: new_binding, else: binding
             expand(query, dynamic, {binding, params, subqueries, count})
+          
           param ->
             {{:^, meta, [count]}, {binding, [param | params], subqueries, count + 1}}
         end
@@ -66,7 +67,7 @@ defmodule Ecto.Query.Builder.Dynamic do
       {:subquery, i}, {binding, params, subqueries, count} ->
         subquery = Enum.fetch!(dynamic_subqueries, i)
         ix = length(subqueries)
-        {{:subquery, ix}, {binding, [{:subquery, ix} | params], [subquery | subqueries], count}}
+        {{:subquery, ix}, {binding, [{:subquery, ix} | params], [subquery | subqueries], count + 1}}
 
       expr, acc ->
         {expr, acc}
--- a/lib/ecto/query/inspect.ex
+++ b/lib/ecto/query/inspect.ex
@@ -289,7 +289,7 @@ defimpl Inspect, for: Ecto.Query do
     json_expr_path_to_expr(expr, path) |> expr(names, part)
   end
 
-  defp expr_to_string({:{}, [], [:subquery, i]}, _string, _names, %BooleanExpr{subqueries: subqueries}) do
+  defp expr_to_string({:{}, [], [:subquery, i]}, _string, _names, %{subqueries: subqueries}) do
     # We were supposed to match on {:subquery, i} but Elixir incorrectly
     # translates those to `:{}` when converting to string.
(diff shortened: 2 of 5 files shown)


AndrewDryga:
I might fix it myself but please show where the legs grow from, I've spent a lot of time trying to digest macro magic and wasn't able to find which exact place triggers that.

AndrewDryga:
In the end it looks like it's `or` operator itself misbehaves because this also fails `inspect(dynamic(posts.id in subquery(subquery) or posts.is_public == true))`

AndrewDryga:
If we stop appending `{subquery, ix}` to params here https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder/dynamic.ex#L69 then the bug disappears but two other tests fail:
1. test/ecto/query/subquery_test.exs:408
2. test/ecto/query/subquery_test.exs:428

AndrewDryga:
I think I managed to fix it and also fixed the inspect protocol implementation. Please tell me if anything else needs to be done (like removing obsolete tests, squashing commits, etc).

josevalim:

PR #3561, opened 2021-02-21
Implement Join Optimization for HasThrough Queries

### LINKED ISSUE #3559: `Ecto.assoc/2` returns incorrect query for `many_to_many` with `through`
### Environment

* Erlang/OTP 23 [erts-11.1.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]
* Elixir 1.11.3 (compiled with Erlang/OTP 23)
* PostgreSQL 12.5
* Ecto version (mix deps): elixir-ecto/ecto {:ecto, github: "elixir-ecto/ecto", override: true}
* Database adapter and version (mix deps): postgrex "0.15.8"
* Operating system: Ubuntu 20.04.2 LTS

### Current behavior

exist the following schemas
```elixir

defmodule EctoDemo.Device do
  use Ecto.Schema
  schema "devices" do
    field(:name, :string)
    many_to_many(
      :interruptions,
      EctoDemo.Interruption,
      join_through: EctoDemo.InterruptionDevice,
      join_keys: [device_id: :id, interruption_id: :id],
      on_replace: :delete
    )

    has_many(:interruption_logs, through: [:interruptions, :interruption_logs])

    timestamps()
  end
end


defmodule EctoDemo.Interruption do
  use Ecto.Schema
  schema "interruptions" do
    field(:name, :string)
    has_many(:device_interruptions, EctoDemo.InterruptionDevice)
    has_many(:interruption_logs, EctoDemo.InterruptionLog)
    timestamps()
  end
end

defmodule EctoDemo.InterruptionDevice do
  use Ecto.Schema
  schema "device_interruptions" do
    belongs_to(:device, EctoDemo.Device)
    belongs_to(:interruption, EctoDemo.Interruption)
  end
end

defmodule EctoDemo.InterruptionLog do
  use Ecto.Schema
  schema "interruption_logs" do
    field(:name, :string)
    belongs_to(:interruption, EctoDemo.Interruption, on_replace: :nilify)
    timestamps()
  end
end

```

and trying to get an interruption_logs via `Repo.all(Ecto.assoc(device, [:interruption_logs]))` we will have error of sql:

```
   ** (Postgrex.Error) ERROR 42P01 (undefined_table) missing FROM-clause entry for table "i2"
     
         query: SELECT DISTINCT i0."id", i0."name", i0."interruption_id", i0."inserted_at", i0."updated_at" FROM "interruption_logs" AS i0 INNER JOIN "device_interruptions" AS d1 ON i2."id" = d1."interruption_id" INNER JOIN "interruptions" AS i2 ON d1."device_id" = ANY($1) WHERE (i0."interruption_id" = i2."id")

```

I found that error appear after this commit: https://github.com/elixir-ecto/ecto/commit/9fabc769cfd74b6227488facb1c59cfc0eca532a (https://github.com/elixir-ecto/ecto/pull/3481)

here simple app for reproduce error:  
https://github.com/parallel588/ecto_demo1
https://github.com/parallel588/ecto_demo1/blob/main/test/assoc_test.exs

GabrielAlacchi:
This PR Fixes #3559

### The problem and solution

The query generated by the author of the issue is as follows:
```
SELECT DISTINCT i0."id", i0."name", i0."interruption_id", i0."inserted_at", i0."updated_at" 
FROM "interruption_logs" AS i0
INNER JOIN "device_interruptions" AS d1 ON i2."id" = d1."interruption_id" 
INNER JOIN "interruptions" AS i2 ON d1."device_id" = ANY($1) 
WHERE (i0."interruption_id" = i2."id")
```

The query that is generated references the binding for the 2nd join table in the on expression of the 1st join table. Therefore Postgres errors out. My solution is to rearrange the ON expressions in `Ecto.Query.Planner.query_to_joins/4` so that they are placed in the correct join expression. The resulting query is this:

```
SELECT DISTINCT i0."id", i0."name", i0."interruption_id", i0."inserted_at", i0."updated_at" FROM "interruption_logs" AS i0
INNER JOIN "device_interruptions" AS d1 ON d1."device_id" = ANY($1) 
INNER JOIN "interruptions" AS i2 ON i2."id" = d1."interruption_id"
WHERE (i0."interruption_id" = i2."id")
```

This query is not perfect. We can technically avoid joining `interruptions` by joining ON `interruption_logs.interruption_id = device_interruptions.interruption_id`. Making this work would probably involve significant refactoring of the way associations are chained together, so this solution is sufficient for now IMO.

### DIFF
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -158,6 +158,14 @@ defmodule Ecto.Integration.User do
     belongs_to :custom, Ecto.Integration.Custom, references: :bid, type: :binary_id
     many_to_many :schema_posts, Ecto.Integration.Post, join_through: Ecto.Integration.PostUser
     many_to_many :unique_posts, Ecto.Integration.Post, join_through: Ecto.Integration.PostUserCompositePk
+
+    has_many :related_2nd_order_posts, through: [:posts, :users, :posts]
+    has_many :users_through_schema_posts, through: [:schema_posts, :users]
+
+    has_many :v2_comments, Ecto.Integration.Comment, foreign_key: :author_id, where: [lock_version: 2]
+    has_many :v2_comments_posts, through: [:v2_comments, :post]
+    has_many :co_commenters, through: [:comments, :post, :comments_authors]
+
     timestamps(type: :utc_datetime)
   end
 end
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -1,4 +1,4 @@
-import Ecto.Query, only: [from: 2, join: 4, distinct: 3]
+import Ecto.Query, only: [from: 1, from: 2, join: 4, join: 5, distinct: 3, where: 3]
 
 defmodule Ecto.Association.NotLoaded do
   @moduledoc """
@@ -40,7 +40,6 @@ defmodule Ecto.Association do
                required(:unique) => boolean,
                optional(atom) => any}
 
-  alias Ecto.Query.{BooleanExpr, QueryExpr, FromExpr}
   alias Ecto.Query.Builder.OrderBy
 
   @doc """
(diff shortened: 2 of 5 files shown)


josevalim:
Thank you @GabrielAlacchi! We have historically avoided doing traversals on the query to lookup information. It can be expensive but also brittle. :(

I would prefer if we could emit the proper query out of the box. Maybe we can change how has_through works to emit the missing join (or similar) when it sees a many to many?

GabrielAlacchi:
Hey @josevalim hope you're doing well! No worries I see your point.

If we can somehow construct a sequence of joins from the list of `throughs` in a more systematic way it would be a good opportunity to optimize as well. If we go through 2 many to many associations we can do the join exclusively through the join tables. Would have to think through the details of how to write this without loss of generality. Also the logic needs to be shared for both `HasThrough` and `Ecto.assoc/2` since you can supply a list of through associations in the latter as well.

I think I'll try writing something which traverses the list of `throughs` and determines the keys that will be joined in the process of building the query by introspection. Using the keys we can decide the minimum set of tables that need to be joined then emit the query in one shot. The current system is not looking ahead at all when constructing the association queries, so it's missing opportunities to cut out unnecessary joins.

GabrielAlacchi:
For example consider we generated the list 
```
[
  {:interrupt_id, InterruptLog},
  {:id, Interrupt},
  {:interrupt_id, DeviceInterrupt},
  {:device_id, DeviceInterrupt},
  {:id, Device}
]
```
using the list of throughs. We know by the `HasMany` from interrupt to interrupt log that `{:interrupt_id, InterruptLog}` is the same as `{:id, Interrupt}` and then we also know by the `ManyToMany` that `{:interrupt_id, DeviceInterrupt}` is the same as `{:id, Interrupt}` as well. So we can deduce that we can join directly to `DeviceInterrupt`.

josevalim:
@GabrielAlacchi keep in mind though we have recently added features like where/join_where and similar to assiciations, which means you need keep the intermediate stuff to filter them. At first, I would focus on correctness and the current issue and then find ways to optimize. :)

GabrielAlacchi:
@josevalim I just pushed a draft commit. Please see `Ecto.Association.chain_through/3`. I've implemented this idea of expanding the joins and I'm looking for feedback. I haven't actually integrated it with the rest of Ecto yet since I'd like to hear your thoughts first! It handles optimization of the joins but also accounts for `join_where` and `where`. It won't eliminate a table from the chain of joins if it has a where clause on it.

josevalim:
@GabrielAlacchi the general approach looks good to me! We would be able to get rid of the changes in the planner, right?

lib/ecto/association.ex
```diff
@@ -175,6 +175,85 @@ defmodule Ecto.Association do
...
+      next =  join(query, :inner, [{src, counter}], dest in ^related_queryable, on: field(src, ^out_key) == field(dest, ^in_key))
+        |> combine_joins_query(where, counter + 1)
+
+      {curr_rel, next, counter + 1, dest_out_key}
+    end)
+
+    final_bind = Ecto.Query.Builder.count_binds(query) - 1
+    values = List.wrap(values)
+    query
+    |> where([{dest, final_bind}], field(dest, ^dest_out_key) in ^values)
```
josevalim, line 254: check if values is a list or not and emit different queries. Some DBs cannot cache `in` queries.

GabrielAlacchi:
> @GabrielAlacchi the general approach looks good to me! We would be able to get rid of the changes in the planner, right?

Yes I'll finalize tomorrow night. I can drop the planner changes and remove `Ecto.Assocation.assoc_query/4` and use `chain_through` instead for both `assoc_query` and `joins_query` for `HasThrough` and for `Ecto.assoc/2`. I'll also ensure that we have integration test coverage for `HasThrough` with `join_where` and `where` to make sure they work correctly.

GabrielAlacchi:
Ok @josevalim I integrated `chain_through` with the rest of Ecto. In doing so I had to modify it slightly to handle going "forwards" or "backwards" depending on whether we're preloading or joining. There is plenty of test coverage which covers joining using `assoc/2` for `HasThrough` and my changes to those tests shows how we're able to save joins with this work! This is a major change which could have unintended side effects of course so please let me know if you'd like to see more test coverage before moving forward with this.

lib/ecto/association.ex
```diff
@@ -176,97 +175,163 @@ defmodule Ecto.Association do
...
   """
-  def assoc_query(refl, through, query, values)
-
-  def assoc_query(%{owner: owner, through: [h|t], field: field}, extra, query, values) do
-    refl = owner.__schema__(:association, h) ||
-            raise "unknown association `#{h}` for `#{inspect owner}` (used by through association `#{field}`)"
-    assoc_query refl, t ++ extra, query, values
+  def filter_through_chain(owner, through, values) do
+    chain_through(owner, through, nil, values)
+    |> distinct([x], true)
```
GabrielAlacchi, line 184: I'm actually confused as to why this `distinct` clause needs to be here. It's used at the end of `assoc_query/4` which is being replaced by this function. Without it certain tests fail due to duplicate records. Yet `ManyToMany` does not add on a distinct clause, and so there are integration tests which use `Ecto.assoc/2` to query the data but expect duplicate records, for example [this test case](https://github.com/elixir-ecto/ecto/blob/master/integration_test/cases/assoc.exs#L137).

I'm unclear on when and when not we should be using distinct in association queries. Perhaps this is something to make more consistent in a follow up PR?

integration_test/support/schemas.exs
```diff
@@ -53,9 +53,20 @@ defmodule Ecto.Integration.Post do
...
+
     belongs_to :author, Ecto.Integration.User
     many_to_many :users, Ecto.Integration.User,
       join_through: "posts_users", on_delete: :delete_all, on_replace: :delete
+
+    # Tests a has through with a many to many and a has many
+    has_many :posts_through_users, through: [:users, :posts]
+
+    # Tests a has through with two many to many associations
+    has_many :posts_through_posts, through: [:users, :schema_posts]
```
josevalim, line 68: This is schema is already super large. Any chance we can write this test by moving this additional associations to the User schema instead? 🤔

lib/ecto/association.ex
josevalim: My understanding is that it only applies to through associations and it was there since the beginning (I added it more than 5 years ago - so I can't quite remember). :+1: to revisiting later.

lib/ecto/association.ex
```diff
@@ -176,97 +175,163 @@ defmodule Ecto.Association do
...
+      {_, 0} -> true
+
+      {rel, _} ->
+        condition = Keyword.get(rel, :where, [])
+        in_key = Keyword.get(rel, :in_key)
+        out_key = Keyword.get(rel, :out_key)
+
+        # If the condition is not empty we need to join to the table. Otherwise if the in_key and out_key is the same
+        # then this join is redundant.
+        in_key != out_key || condition != []
```
josevalim, line 224: Since both sides are booleans:

```suggestion
        in_key != out_key or condition != []
```

lib/ecto/association.ex
```diff
@@ -176,97 +175,163 @@ defmodule Ecto.Association do
...
+    # a list of atomic associations to join through.
+    {_, through} = flatten_through_chain(owner, through, [])
+
+    # If we're joining then we're going forward from the owner table to the destination table.
+    # Otherwise we're going backward from the destination table then filtering by values.
+    chain_direction = if(join_to != nil, do: :forward, else: :backward)
+
+    # This stage produces a list of joins represented as a keyword list with the following structure:
+    # [
+    #   [schema: (The Schema), in_key: (The key used to join into the table), out_key: (The key used to join with the next), where: (The condition KW list)]
```
josevalim, line 207: Suggestion: let's make it a list of maps and make all of the keys required (for example, where can always be an empty list). This will simplify later code a lot. For example, the code below can go from this:

```elixir
        condition = Keyword.get(rel, :where, [])
        in_key = Keyword.get(rel, :in_key)
        out_key = Keyword.get(rel, :out_key)

        # If the condition is not empty we need to join to the table. Otherwise if the in_key and out_key is the same
        # then this join is redundant.
        in_key != out_key || condition != []
```

to this:

```elixir
      rel.in_key != rel.out_key or rel.where != []
```

josevalim:
Hi @GabrielAlacchi! This looks great, thank you! :heart: I have added only one minor comment, to move from keyword lists to maps, which should help clean up the code a little, and a bit of test cleaning it up too. Other than that, this looks good to go!

integration_test/support/schemas.exs
GabrielAlacchi: Sure I pivoted to queries on user

lib/ecto/association.ex
GabrielAlacchi: Yes this is definitely the best way to go!

GabrielAlacchi:
Ok I've refactored. Please check it out @josevalim! I'm interested in creating a GitHub workflow to run the integration tests by the way. I'd feel a lot more sane about these changes if they ran automatically on a matrix of DB and elixir versions. I've written workflows that run services in docker containers on the host for these kinds of tests.

How exhaustive your matrix is definitely depends on your budget on GitHub actions though.

josevalim:
@GabrielAlacchi it is tricky because we have the integration tests running on ecto_sql and we likely don't want to duplicate efforts... ideas?

josevalim:

GabrielAlacchi:
> @GabrielAlacchi it is tricky because we have the integration tests running on ecto_sql and we likely don't want to duplicate efforts... ideas?

I see, so you're running the integration tests in the CI for ecto_sql against the latest master of ecto? If that's the case then there's no need to duplicate as long as integration test regressions are checked before releasing a new version 😄

josevalim:
I think we need a mix deps.update Ecto there and that should push it to latest. :)

tcrossland:
This change doesn't fix the case described [here](https://github.com/elixir-ecto/ecto/pull/3481#issuecomment-814269916):

```elixir
parent
|> Ecto.assoc(:children)
|> join(:inner, [child], grandchild in assoc(child, :children))
|> where([_parent, _child, rel, _grandchild], rel.type_id == 0)
|> Repo.all()
```

I'll try to put together a simple test to reproduce the issue.

PR #3565, opened 2021-02-25
Add Ecto.Multi.put/3

mikestok:
This allows values to be put into the values so far in a multi
so that they don't have to be passed in to functions in the multi
pipeline.

Discussed in
https://groups.google.com/g/elixir-ecto/c/dKfkizcO73Y/m/z4yZ0INZBQAJ

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -125,6 +125,7 @@ defmodule Ecto.Multi do
   @typep schema_or_source :: binary | {binary | nil, binary} | atom
   @typep operation :: {:changeset, Changeset.t, Keyword.t} |
                       {:run, run} |
+                      {:put, any} |
                       {:merge, merge} |
                       {:update_all, Ecto.Query.t, Keyword.t} |
                       {:delete_all, Ecto.Query.t, Keyword.t} |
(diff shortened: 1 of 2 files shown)


lib/ecto/multi.ex
```diff
@@ -602,6 +602,18 @@ defmodule Ecto.Multi do
...
+  @doc """
+  Adds a value to the changes so far under the given name.
+
+  ## Example
+
+      Ecto.Multi.put(multi, :password, "sghbj15")
+  """
+  @spec put(t, name, any) :: t
+  def put(multi, name, value) do
+    Ecto.Multi.run(multi, name, fn _, _ -> {:ok, value} end)
```
josevalim, line 614: We can probably do an optimized implementation that stored the value directly. Can you please take a look?

test/ecto/multi_test.exs
```diff
@@ -327,6 +327,19 @@ defmodule Ecto.MultiTest do
...
+  test "put" do
+    name = :halo
+    value = "statue"
+
+    multi =
+      Multi.new()
+      |> Multi.put(name, value)
+
+    assert multi.names == MapSet.new([name])
+    assert {:run, fun} = multi.operations[name]
```
josevalim, line 339: This is testing implementation details and it will break once we refactor. It is better to test that we get the “put” value on subsequent ops and as a result of transaction. :)

lib/ecto/multi.ex
```diff
@@ -602,6 +602,18 @@ defmodule Ecto.Multi do
   defp format_operation(other),
     do: other
 
+  @doc """
+  Adds a value to the changes so far under the given name.
+
+  ## Example
+
+      Ecto.Multi.put(multi, :password, "sghbj15")
```
v0idpwn, line 610:
```suggestion
      Ecto.Multi.new()
      |> Ecto.Multi.put(:params, params)
      |> Ecto.Multi.run()
      |> Ecto.Multi.insert(:user, fn changes -> &User.changeset(changes.params) end)
      |> Ecto.Multi.insert(:person, fn changes -> &Person.changeset(changes.user, changes.params) end)
      |> MyApp.Repo.transaction()
```

I think this could show how it would be useful

lib/ecto/multi.ex
v0idpwn: Maybe we could also support keyword lists/maps?

josevalim:

lib/ecto/multi.ex
```diff
@@ -602,6 +603,23 @@ defmodule Ecto.Multi do
...
     do: other
 
+  @doc """
+  Adds a value to the changes so far under the given name.
+
+  ## Example
+
+      Ecto.Multi.new()
+      |> Ecto.Multi.put(:params, params)
+      |> Ecto.Multi.run()
```
halostatue, line 613: This looks like a stray bit from a previous revision that might want to be cleaned up in a small change.

halostatue:
💯

lib/ecto/multi.ex
wojtekmach: done on master, thank you for the PR!

PR #3566, opened 2021-02-26
Add Multi.inspect/1,2

v0idpwn:
Adds Multi.peek/2,3 as specified in the proposal:
https://groups.google.com/g/elixir-ecto/c/17MKpYXcEDU/m/zsR-BHpWAAAJ

Renamed from `inspect` to `peek` to avoid name-clashing with
`Kernel.inspect/2`

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -630,15 +630,30 @@ defmodule Ecto.Multi do
 
   All options for IO.inspect/2 are supported, it also support the following ones:
 
-    * `:only` - A list of fields to inspect, will print the entire map by
-      default.
+    * `:only` - A field or a list of fields to inspect, will print the entire
+      map by default.
 
-  ## Example
+  ## Examples
 
       Ecto.Multi.new()
-      |> Ecto.Multi.insert(:person, changeset)
+      |> Ecto.Multi.insert(:person_a, changeset)
+      |> Ecto.Multi.insert(:person_b, changeset)
       |> Ecto.Multi.inspect()
       |> MyApp.Repo.transaction()
+
+  Prints:
+      %{person_a: %Person{...}, person_b: %Person{...}}
+
+  We can use the `:only` option to limit which fields will be printed:
+
+      Ecto.Multi.new()
+      |> Ecto.Multi.insert(:person_a, changeset)
+      |> Ecto.Multi.insert(:person_b, changeset)
+      |> Ecto.Multi.inspect(only: :person_a)
+      |> MyApp.Repo.transaction()
+
+  Prints:
+      %{person_a: %Person{...}}
 
   """
   @spec inspect(t, name, keyword()) :: t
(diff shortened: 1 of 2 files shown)


v0idpwn:
Maybe we should also add an option to show a result only? Will do it, if its not desired we may merge just the first commit

josevalim:
It is fine to call it `inspect`.  Just make sure you do `Kernel.inspect/2` instead of `inspect/2` elsewhere in this file. AFter all, we do have `IO.inspect/2` and all is fine. :)

v0idpwn:
@josevalim sure, but isn't calling it `inspect` improper? For me, at least, the word relates to the protocol

josevalim:
I don't think it is. Especially because we call things like `insert`, `update`... they are all proxies to things that happen later.

v0idpwn:
Will rename it to inspect then :)

v0idpwn:
WDYT about the :only option?

josevalim:
`:only` doesn't hurt. :)

v0idpwn:
There it is :)

lib/ecto/multi.ex
```diff
@@ -619,6 +619,31 @@ defmodule Ecto.Multi do
...
+  ## Example
+
+      Ecto.Multi.new()
+      |> Ecto.Multi.insert(:person, changeset)
+      |> Ecto.Multi.inspect()
+      |> MyApp.Repo.transaction()
+
+  """
+  @spec inspect(t, name, keyword()) :: t
+  def inspect(multi, name, opts \\ []) do
```
josevalim, line 644: Let's remove the `name` as argument. Instead, you can set internally it to always be `:inspect` (we do similar for `merge`). Then we show no label by default and leave it to be customized via the label option.

lib/ecto/multi.ex
```diff
@@ -659,14 +684,26 @@ defmodule Ecto.Multi do
     end
   end
 
+  defp apply_operation({name, {:inspect, opts}}, _repo, _wrap_, _return, {acc, names}) do
+    opts = Keyword.merge([label: Atom.to_string(name)], opts)
+
+    if opts[:only] do
+      acc |> Map.take(opts[:only]) |> IO.inspect(opts)
```
josevalim, line 692: For convenience, please make sure this also works: `inspect(only: :foo)`:

```suggestion
      acc |> Map.take(List.wrap(opts[:only])) |> IO.inspect(opts)
```

Will also need a test. :)

v0idpwn:
Just pushed all the requested changes.

@josevalim while I agree that inspect is a more clear name, I'm concerned with the following:
```elixir
iex(1)> defmodule MyMod do 
...(1)>   import Ecto.Multi
...(1)>   def my_fun, do: inspect("bar")
...(1)> end
```

Which will raise a compile error:
```
** (CompileError) iex:3: function inspect/1 imported from both Ecto.Multi and Kernel, call is ambiguous
    (elixir 1.11.1) src/elixir_dispatch.erl:129: :elixir_dispatch.expand_import/6
    (elixir 1.11.1) src/elixir_dispatch.erl:91: :elixir_dispatch.dispatch_import/5
    iex:3: (module)
```

josevalim:
Right, then don't import Multi or if you have to, use the fully qualified name. That's fine. :)

josevalim:

Issue #2680, opened 2018-09-06
Validations don't clear existing errors


bencates:
### Environment

* Elixir version: 1.7.2
* Ecto version: 2.2.8

### Current behavior

```elixir
defmodule TestSchema do
  use Ecto.Schema
  embedded_schema do
    field :key, :string
  end
end

alias Ecto.Changeset

changeset =
  %TestSchema{}
  |> Changeset.change(%{})
  |> Changeset.validate_required(:key) # <= Changeset marked invalid here
  |> Changeset.put_change(:key, "value")
  |> Changeset.validate_required(:key)

changeset.valid? == false
changeset.errors == [key: {"can't be blank", [validation: :required]}]
```

### Expected behavior

Validations like `validate_required/3` should remove the previous error and, assuming there are no other errors, mark the changeset as valid again.

fertapric:
In the given example, could be an option to first set the value of `:key` conditionally and then perform the `validate_required/3` validation? Having two `validate_required/3` over the same field in a changeset pipeline feels a bit weird to me. Semantically, it would mean that the value is not “required” after all 🙂

josevalim:
This would be a breaking change too. There is nothing today requiring validation_required to come before all other validations, so removing any error once the field is there would certainly lead to bugs. The best solution in this case is to guarantee the data is in the proper shape before validating it.

bencates:
The listed code snippet is drastically simplified, of course. The actual issue in our application is that when `TestSchema.changeset/2` returns an invalid changeset I am unable to repair it in the calling context.

A less minimized example:

```elixir
def create_purchase_order(attrs) do
  # `attrs` is raw user input. It may specify a PO Number or allow the accounting
  # system to generate one automatically. It may have either atom or string keys.

  # `:po_number` is a required field in the changeset (and a `NOT NULL column
  # in the underlying database). If the user did not specify one then the resulting
  # changeset will be invalid at this stage. Either way we want to extract an in
  # memory record to use as a parameter for the following cleanup work.
  changeset = PurchaseOrder.changeset(%PurchaseOrder{}, attrs)
  purchase_order = Changeset.apply_changes(changeset)

  # We then have to make an external API call to register the new PO with
  # our accounting system.
  {:ok, res} = AccountingSystem.create_purchase_order(purchase_order)

  # 
  changeset
  |> Changeset.put_change(:po_number, res["poNumber"])
  |> Changeset.validate_required(:po_number)
  |> Repo.insert!
end
```

[...]

PR #3567, opened 2021-02-26
Add remove_errors/2-3 to Ecto.Changeset

tverlaan:
This feature was previously suggested in #2680 but never implemented. I think the feature is easily added to your own codebase, but at the same time it seems to make sense to add it to `Ecto.Changeset` as a counterpart to `add_error`.

I came across the original issue because I was looking to invalidate specific earlier validations while working in Phoenix Liveview. In the examples of Liveview the changeset is generated on each validation, even though the existing changeset with existing data & errors is stored in the socket:

```elixir
def handle_event("validate", %{"user" => params}, socket) do
  changeset =
    %User{}
    |> Accounts.change_user(params)
    |> Map.put(:action, :insert)

  {:noreply, assign(socket, changeset: changeset)}
end
```

In my specific case part of the changeset is generated in the first step, after which I only want to continue validating an additional field and I'd like to keep the other data intact. Removing the error for that field and validating only that field again seems to be the easiest approach.

I added remove_errors/3 to give the possibility for further customization for custom validations. An example to invalidate a specific error could be the following:

```elixir
@my_validation make_ref()
add_error(changeset, :title, "custom", [ref: @my_validation])
remove_errors(changeset, :title, &(&1 == [ref: @my_validation])
# or
remove_errors(changeset, :title, &Enum.any?(&1, fn additional -> additional == [ref: @my_validation] end))
```

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1656,6 +1656,51 @@ defmodule Ecto.Changeset do
   end
 
   @doc """
+  Removes error(s) from the changeset that match the given key.
+
+  This removes all errors that match given key. If no errors are
+  left the changeset is marked as valid again.
+
+  You can pass a `fun` as third argument which will receive the additional keyword
+  list and only removes those errors for which `fun` returns a truthy value.
+
+  ## Examples
+
+      iex> changeset = change(%Post{}, %{title: ""})
+      iex> changeset = add_error(changeset, :title, "empty")
+      iex> changeset = remove_errors(changeset, :title)
+      iex> changeset.errors
+      []
+      iex> changeset.valid?
+      true
+
+      iex> changeset = change(%Post{}, %{title: "", author: ""})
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @tverlaan! Thanks for the PR!

I have really mixed things about this feature. Removing errors doesn't meant the data suddenly became valid. For example, if a name is invalid, then you remove the error, the invalid name still won't be marked as a change, but maybe you wanted it to now become a change? Or did you remove the error due to UI concerns?

There are so many consequences from removing the errors that it is almost sure it will bite someone and it will be used outside of the original and small scope you are using it for. It will also lead to a style where we generate a wrong changeset, then we fix it, then it gets wrong again, which will be hard to reason about.

For all of those reasons, I don't want to merge the feature, even though it was elegantly implemented by you here. Luckily others can pick it up from here if they need it.

Thank you and have a great weekend!

Issue #450, opened 2015-02-12
Support put_new_change in changesets


josevalim:
/cc @whatyouhide

whatyouhide:
I think this can be closed :)

PR #3573, opened 2021-03-04
add changeset put_new_change function

lohayon:
A little PR to allow putting new changes in the changeset as Map.put_new does.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1262,38 +1262,9 @@ defmodule Ecto.Changeset do
     type = Map.get(types, key)
 
     {changes, errors, valid?} =
-      put_new_change(
-        data,
-        changeset.changes,
-        changeset.errors,
-        changeset.valid?,
-        key,
-        value,
-        type
-      )
+      put_new_change(data, changeset.changes, changeset.errors, changeset.valid?, key, value, type)
 
     %{changeset | changes: changes, errors: errors, valid?: valid?}
-  end
-
-  defp put_new_change(data, changes, errors, valid?, key, value, {tag, relation})
-       when tag in @relations do
-    if Map.has_key?(changes, key) do
-      {changes, errors, valid?}
-    else
-      put_change(data, changes, errors, valid?, key, value, {tag, relation})
-    end
-  end
-
-  defp put_new_change(data, _changes, _errors, _valid?, key, _value, nil) when is_atom(key) do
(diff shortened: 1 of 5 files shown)


lib/ecto/changeset.ex
```diff
@@ -1237,6 +1237,73 @@ defmodule Ecto.Changeset do
...
+          }`"
+  end
+
+  defp put_new_change(data, changes, errors, valid?, key, value, type) do
+    unless Map.has_key?(changes, key) do
+      put_change(data, changes, errors, valid?, key, value, type)
+    else
+      {changes, errors, valid?}
+    end
+  end
```
josevalim, line 1305: I think it should work if we keep only this clause and let `put_change` do the raising. WDYT? :)

josevalim (commented):
This looks great! I have dropped one comment for you consideration, thank you!

wojtekmach:
We actually briefly had this function (https://github.com/elixir-ecto/ecto/issues/450) and removed it (https://github.com/elixir-ecto/ecto/commit/5515298e1f30133f9c593ba8475d1578436f4ef1) so let's carefully consider that history before moving forward with it.

lib/ecto/changeset.ex
lohayon: oh yes it is better.
I will modify the PR

josevalim:
Oh, good call @wojtekmach. I remember now why we removed it because it was unclear if change should consider the current value on the changeset, so we ask folks to either get_change or get_field explicitly.

@lohayon i am sorry but we should probably no go ahead with this :(

lohayon:
The last pr was in 2015. Now, we have put_change that considers the fields of the changeset. The function put_new_change is explicitly named as it will add only new changes, considering the changes of the changeset. WDYT?

josevalim:
We also had `put_change` back then. The issue is not really what change means but rather what "new" means. :(

wojtekmach:
@lohayon thank you for the PR. I think we shouldn't move forward with this because of the reason we reverted it back then.

There's inherent ambiguity whether "put_new" refers to `changeset.data` or `changeset.changes` (in truth it refers to both) and the best way to solve this ambiguity is to explicitly use `get_field` or `get_change` and then `put_change` afterwards. We'd be happy to accept a PR that adds documentation to `put_change/2` that shows how to do that, then users can easily add such function to their projects. just to be clear, I think the PR is perfect and I think it'd be a good addition but I feel we sometimes shouldn't make choices for users so that the users can make the choices explicitly themselves.

PR #3574, opened 2021-03-05
Allow insert_all to be called with a query instead of rows

narrowtux:
There's a matching PR over in ecto_sql: https://github.com/elixir-ecto/ecto_sql/pull/308

This feature allows the usage of `Repo.insert_all` with a source query instead of rows with already materialized values:

```elixir
source = from p in Post, 
  select: %{
    author_id: p.author_id,
    posts: count(p.id)
  },
  group_by: p.author_id

Repo.insert_all(AuthorStatistics, source, on_conflict: [:author_id], on_conflict: :replace_all)
```

which is roughly equivalent to this SQL Query (in PostgreSQL syntax):

```sql
INSERT INTO 
  author_statistics 
  (author_id, posts) 
  (
    SELECT p1.author_id AS author_id, count(p1.id) AS posts 
    FROM posts p1
    GROUP BY p1.author_id
  )
ON CONFLICT 
  (author_id) 
DO UPDATE SET
  author_id = EXCLUDED.author_id,
  posts = EXCLUDED.posts
```

## TODO

 - [x] Postgres adapter (sql gen and tests)
 - [x] MyXQL adapter
 - [x] MSSQL Adapter
 - [x] Move planning from ecto_sql to ecto
 - [x] Documentation in `Repo.insert_all`
 - [x] Throw helpful errors for edge cases:
   - [x] source query does not have a select, or the select is not a map or a keyword list


test/ecto/repo_test.exs
```diff
@@ -480,6 +480,21 @@ defmodule Ecto.RepoTest do
...
+      import Ecto.Query
+
+      threshold = 10
+
+      query = from s in MySchema,
+        where: s.x > ^threshold
+
+      TestRepo.insert_all(MySchema, query)
+
+      assert_received {:insert_all, %{source: "my_schema"}, %Ecto.Query{} = query}
```
narrowtux, line 493: I think I will add checks here whether the query is planned

lib/ecto/repo/schema.ex
```diff
@@ -76,6 +81,40 @@ defmodule Ecto.Repo.Schema do
...
+        """
+
+      %Ecto.Query.SelectExpr{expr: {op, _ctx, args}} when op in [:%{}, :"[]"] ->
+        header =
+          Keyword.keys(args)
+          |> Enum.reduce(%{}, &Map.put(&2, &1, true))
+
+        query = Map.update!(query, :select, fn select ->
+          Map.update!(select, :expr, fn {op, ctx, args} ->
+            args = Enum.sort_by(args, &elem(&1, 0))
```
narrowtux, line 108: The sort is necessary since the `VALUES` clause in SQL is fed by a map, where the keys are alphabetically sorted.

josevalim:
Yes, sorting is good. Also we don’t really support keywords on select, so
you can remove that. And we should raise if you pass anything other than a
map.

test/ecto/repo_test.exs
narrowtux: Done

lib/ecto/repo/schema.ex
```diff
@@ -31,7 +31,7 @@ defmodule Ecto.Repo.Schema do
     end
   end
 
-  defp do_insert_all(name, schema, prefix, source, rows, opts) when is_list(rows) do
+  defp do_insert_all(name, schema, prefix, source, rows, opts) when is_list(rows) or is_struct(rows, Ecto.Query) do
```
narrowtux, line 34: Is there a way to have this when clause in elixir 1.8 at all? Otherwise I'd say we should remove the when clause entirely

josevalim: Not possible. You can validate the arguments separately and raise accordingly.

narrowtux: Latest push addresses this

lib/ecto/repo/schema.ex
```diff
@@ -31,7 +31,12 @@ defmodule Ecto.Repo.Schema do
     end
   end
 
-  defp do_insert_all(name, schema, prefix, source, rows, opts) when is_list(rows) do
+  defp do_insert_all(name, schema, prefix, source, rows, opts) do
+    case rows do
+      list when is_list(list) -> :ok
+      %Ecto.Query{} -> :ok
+      _ -> raise ArgumentError, message: "insert_all called with invalid rows_or_query: #{inspect rows}"
+    end
```
josevalim, line 39: Btw, something else you can do, and perhaps cleaner, is to check for this on `extract_header_and_fields`.

lib/ecto/repo/schema.ex
```diff
@@ -76,6 +86,43 @@ defmodule Ecto.Repo.Schema do
     end
   end
 
+  defp extract_header_and_fields(query = %Ecto.Query{}, _schema, _dumper, _autogen_id, _placeholder_map, adapter) do
+    case query.select do
+      %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}} ->
+        header =
+          Keyword.keys(args)
+          |> Enum.reduce(%{}, &Map.put(&2, &1, true))
```
josevalim, line 94:
```suggestion
        header = Map.new(args, fn {k, _} -> {k, true} end)
```

lib/ecto/repo/schema.ex
```diff
@@ -76,6 +86,43 @@ defmodule Ecto.Repo.Schema do
...
+
+        query = Map.update!(query, :select, fn select ->
+          Map.update!(select, :expr, fn {op, ctx, args} ->
+            args = Enum.sort_by(args, &elem(&1, 0))
+            {op, ctx, args}
+          end)
+        end)
```
josevalim, line 101:
```suggestion
        query = update_in query.select.expr, fn {op, ctx, args} ->
            args = Enum.sort_by(args, &elem(&1, 0))
            {op, ctx, args}
        end
```

lib/ecto/repo/schema.ex
```diff
@@ -76,6 +86,43 @@ defmodule Ecto.Repo.Schema do
...
+
+        from x in Source,
+          ...,
+          select: %{
+            field_a: x.bar,
+            field_b: x.foo
+          }
+
+        The keys must exist in the schema that is being inserted into.
+        """
```
josevalim, line 121:
```suggestion
        raise ArgumentError, """
        Cannot generate a fields list for insert_all from the given source query
        because it does not have a select clause that uses a map:

        #{inspect query}

        Please add a select clause that selects into a map, like this:

            from x in Source,
              ...,
              select: %{
                field_a: x.bar,
                field_b: x.foo
              }

        The keys must exist in the schema that is being inserted into
        """
```

lib/ecto/repo/schema.ex
```diff
@@ -76,6 +86,41 @@ defmodule Ecto.Repo.Schema do
...
+    case query.select do
+      %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}} ->
+        header = Map.new(args, fn {k, _} -> {k, true} end)
+
+        query = Map.update!(query, :select, fn select ->
+          Map.update!(select, :expr, fn {op, ctx, args} ->
+            args = Enum.sort_by(args, &elem(&1, 0))
+            {op, ctx, args}
+          end)
+        end)
```
josevalim, line 99: Quick question: is this sorting going to work if the Map has more than 32 elements? Then Map.keys/1 and this function may return different orders... perhaps it is best if this function return this `headers` as a list so we can just go with the order in `{:%{}, ..., args}`?

narrowtux: Are maps not sorted when they have over 32 elements? That's good to know. I agree, this is kind of sketchy. I would also much rather use a list that has the same order as the select expression to generate the fields list in the INSERT INTO.

josevalim: Yeah, so let's return headers a list here from this function and remove the Map.keys from the caller.

lib/ecto/repo/schema.ex
narrowtux: yes, excellent idea

lib/ecto/repo/schema.ex
```diff
@@ -42,21 +42,26 @@ defmodule Ecto.Repo.Schema do
...
-    {rows, header, placeholder_values} =
-      extract_header_and_fields(rows, schema, dumper, autogen_id, placeholder_map, adapter)
+    {rows_or_query, header, placeholder_values} =
+      extract_header_and_fields(rows_or_query, schema, dumper, autogen_id, placeholder_map, adapter)
+
+    counter =
+      case rows_or_query do
+        rows when is_list(rows) -> fn -> Enum.reduce(rows, 0, &length(&1) + &2) end
+        {%Ecto.Query{}, params} -> fn -> length(params) end
+      end
```
josevalim, line 52: we should likely move the counter to `extract_header_and_fields` and keep the logic all encapsulated in one place. :)

lib/ecto/repo/schema.ex
```diff
@@ -101,6 +108,35 @@ defmodule Ecto.Repo.Schema do
...
+  defp extract_header_and_fields(%Ecto.Query{select: %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}}} = query, _schema, _dumper, _autogen_id, _placeholder_map, adapter) do
+    header = Enum.map(args, &elem(&1, 0))
+
+    query_and_params = Ecto.Adapter.Queryable.plan_query(:all, adapter, query)
+
+    {query_and_params, header, []}
+  end
+  defp extract_header_and_fields(%Ecto.Query{} = query, _schema, _dumper, _autogen_id, _placeholder_map, _adapter) do
+    raise ArgumentError, message: """
+    Cannot generate a fields list for insert_all from the given source query
```
josevalim, line 120: Nitpick: error messages start in lowercase and we can skip the message bit. :)

```suggestion
    raise ArgumentError, """
    cannot generate a fields list for insert_all from the given source query
```

lib/ecto/repo/schema.ex
```diff
@@ -101,6 +108,35 @@ defmodule Ecto.Repo.Schema do
...
+        select: %{
+          field_a: x.bar,
+          field_b: x.foo
+        }
+
+    The keys must exist in the schema that is being inserted into
+    """
+  end
+  defp extract_header_and_fields(rows_or_query, _schema, _dumper, _autogen_id, _placeholder_map, _adapter) do
+    raise ArgumentError, message: "expected a list of rows or a query, but got #{inspect rows_or_query} as rows_or_query argument in insert_all"
```
josevalim, line 138:
```suggestion
    raise ArgumentError, "expected a list of rows or a query, but got #{inspect rows_or_query} as rows_or_query argument in insert_all"
```

lib/ecto/repo/schema.ex
narrowtux: ok!

lib/ecto/repo/schema.ex
```diff
@@ -96,11 +99,42 @@ defmodule Ecto.Repo.Schema do
...
+      {rows, header, placeholder_vals_list, counter}
     else
-      {rows, header, placeholder_vals_list}
+      {rows, header, placeholder_vals_list, counter}
     end
   end
+  defp extract_header_and_fields(%Ecto.Query{select: %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}}} = query, _schema, _dumper, _autogen_id, _placeholder_map, adapter) do
+    header = Enum.map(args, &elem(&1, 0))
+
+    {query, params} = Ecto.Adapter.Queryable.plan_query(:all, adapter, query)
```
josevalim, line 110: I wonder if this should be `:insert_all`. Looking at ecto/repo/queryable.ex, we may also want to call both

    {query, opts} = repo.prepare_query(operation, query, opts)
    query = attach_prefix(query, opts)

(perhaps we should add these two calls inside `Ecto.Adapter.Queryable.plan_query`).

josevalim:
Hi @narrowtux! I have added my last batch of reviews. We need to call repo.prepare_query for the query and potentially attach any prefix given as an option. The rest of the comments here are nitpicks. On the ecto_sql side, it is mostly a matter of structuring the tests. :) Thank you!

lib/ecto/repo/schema.ex
narrowtux: and then call attach_prefix only if `operation == :insert_all`? Also, how do we get the `repo` in there?

narrowtux: Oh, I was looking at the wrong file

josevalim: Good point. If we can't get the repo there we will need to do this dance before calling plan_adapter. Note you will have to call `Ecto.Queryable.to_query(queryable)` before calling prepare_query, so you might as well skip calling plan_adapter altogether and have all the logic here.

josevalim: There is another approach to take here, which is to call `Planner.query(query, operation, cache, adapter, 0)`. The advantage of calling it is that queries can be prepared and cached but that comes with other difficulties in the case of insert (since the cache key needs to include options such as conflict_target and so on - which is probably better to skip).

narrowtux: Since attach_prefix was a `defp` in Ecto.Repo.Queryable and I need it in Ecto.Repo.Schema, I extracted the function into a new module Ecto.Repo.Common

josevalim: Please move it to Ecto.Query.Planner, as it has the shared functionality already. :)

narrowtux: I've addressed this now. I have read the postgres insert into spec and there was no mention of limited functionality for the source query. So I've copied all the behaviour specific to `:all` queries in Ecto.Query.Planner to the new `:insert_all` operation.

narrowtux: I was able to loop the repo from `Ecto.Repo.Schema.insert_all` through `do_insert_all` and finally into `extract_header_and_fields`. Only question is: do we prepare the query before we plan or the other way around?

lib/ecto/repo/schema.ex
```diff
@@ -96,11 +101,45 @@ defmodule Ecto.Repo.Schema do
...
+      {rows, header, placeholder_vals_list, counter}
     end
   end
+  defp extract_header_and_fields(repo, %Ecto.Query{select: %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}}} = query, _schema, _dumper, _autogen_id, _placeholder_map, adapter, opts) do
+    header = Enum.map(args, &elem(&1, 0))
+
+    {query, opts} = repo.prepare_query(:insert_all, query, opts)
+    query = attach_prefix(query, opts)
+
+    {query, params} = Ecto.Adapter.Queryable.plan_query(:insert_all, adapter, query)
```
josevalim, line 115: This is the correct order. Although you may want to extract and validate the select.expr after the planning!

lib/ecto/repo/schema.ex
```diff
@@ -96,11 +101,45 @@ defmodule Ecto.Repo.Schema do
...
+      {rows, header, placeholder_vals_list, counter}
     else
-      {rows, header, placeholder_vals_list}
+      {rows, header, placeholder_vals_list, counter}
     end
   end
+  defp extract_header_and_fields(repo, %Ecto.Query{select: %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}}} = query, _schema, _dumper, _autogen_id, _placeholder_map, adapter, opts) do
+    header = Enum.map(args, &elem(&1, 0))
+
+    {query, opts} = repo.prepare_query(:insert_all, query, opts)
```
josevalim, line 112: Remember to update the docs/typespec of the prepare_query callback in Ecto.Repo!

integration_test/cases/repo.exs
```diff
@@ -881,6 +881,39 @@ defmodule Ecto.Integration.RepoTest do
...
+    source = from p in Post,
+      select: %{
+        title: fragment("concat(?, ?, ?)", p.title, type(^" suffix ", :string), p.id)
+      }
+
+    opts = case adapter do
+      "pg" -> [conflict_target: [:id], on_conflict: :replace_all, returning: [:id, :title]]
+      "myxql" -> [on_conflict: :replace_all]
+      "tds" -> [returning: [:id, :title]]
+    end
```
josevalim, line 901: Sorry, I probably gave unclear advice. Those should be distinct tests:

1. one test for insert_all with query and conflict target
2. one test for insert_all with query and returning
3. one test for insert_all with query and on_conflict
4. one test for insert_all with query

Then mark each test with its appropriate tags so they can be skipped from ecto_sql. :) We should not depend on the adapter here.

lib/ecto/repo.ex
```diff
@@ -900,8 +900,8 @@ defmodule Ecto.Repo do
   before it is transformed and sent to the database.
 
   This callback is invoked for all query APIs, including the `stream`
-  function, but it is not invoked for `insert_all` nor any of the
-  schema functions.
+  functions and insert_all (if a source query was given instead of a list of
+  rows), but it is not invoked for any of the other schema functions.
```
josevalim, line 904:
```suggestion
  functions. It is also invoked for `insert_all` if a source query is
  given. It is not invoked for any of the other schema functions.
```

integration_test/cases/repo.exs
josevalim: Feel free to put all of them in a shared describe block called "insert_all with query". :)

josevalim (commented):
Awesome job on the docs. I have dropped one comment on the tests and once they are broken apart we can ship this!

integration_test/cases/repo.exs
narrowtux: I see. I think the latest push is how you wanted it to work.

josevalim:

PR #3575, opened 2021-03-05
Add @impl true to Repo callbacks in guide

aaronrenner:
It seems like it would be nice if we added `@impl true` to the Repo callbacks in the Multi Tenancy with Foreign Keys guide. This not only ensures the arity and spelling is correct, but also indicates this function is a callback.

Awesome guide, btw!


josevalim:

PR #3576, opened 2021-03-06
Added apply_query() + other specs; fixed build_cte() spec.

azizk:
Also refactored apply_query() a little. The escape_query() function will always receive a Query struct afaics so I removed the second clause.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -1103,7 +1103,7 @@ defmodule Ecto.Query.Builder do
   given `module`, otherwise, it delegates the call to runtime.
 
   It is important to keep in mind the complexities introduced
-  by this function. In particular, a %Query{} is mixture of escaped
+  by this function. In particular, a %Query{} is a mixture of escaped
   and unescaped expressions which makes it impossible for this
   function to properly escape or unescape it at compile/runtime.
   For this reason, the apply function should be ready to handle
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder.ex
```diff
@@ -1133,21 +1133,24 @@ defmodule Ecto.Query.Builder do
...
 
-    case unescape_query(query) do
-      %Query{} = unescaped ->
-        apply(module, :apply, [unescaped|args]) |> escape_query
-      _ ->
+      runtime_query ->
         quote do
-          query = unquote(query) # Unquote the query for any binding variable
+          # Unquote the query before `module.apply()` for any binding variable.
+          query = unquote(runtime_query)
```
josevalim, line 1146: If we unquote the unescaped query, it will be invalid AST. We should revert this change. :)

lib/ecto/query/builder.ex
```diff
@@ -1163,10 +1166,8 @@ defmodule Ecto.Query.Builder do
   end
 
   # Escapes an `Ecto.Query` and associated structs.
-  defp escape_query(%Query{} = query),
-    do: {:%{}, [], Map.to_list(query)}
-  defp escape_query(other),
-    do: other
+  @spec escape_query(Query.t) :: Macro.t
+  defp escape_query(%Query{} = query), do: {:%{}, [], Map.to_list(query)}
```
josevalim, line 1170: Because this is the request of apply, there is no guarantee apply won't emit an already escaped AST, so we should not change this either. :)

lib/ecto/query/builder.ex
josevalim: Oh, ignore me, we don't actually unescape it in those case.

josevalim:

PR #3577, opened 2021-03-06
Allow shorthand assoc join via dot operator.

azizk:
I'd like to propose a shorthand syntax for joining associations to a query.

For example: `Post |> join(:inner, [p], p.comments)`

If this is something you'd also like to see, I'd be excited and I would write some tests and documentation to complete the pull request. :+1:

### DIFF
--- a/lib/ecto/query/builder/join.ex
+++ b/lib/ecto/query/builder/join.ex
@@ -38,6 +38,18 @@ defmodule Ecto.Query.Builder.Join do
       iex> escape(quote(do: c in assoc(p, :comments)), [p: 0], __ENV__)
       {:c, nil, {0, :comments}, []}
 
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @azizk, I don't want to support this syntax for two main reasons:

1. Our syntax sugar for cases like this is to pass atoms/keyword lists, for example: `from Post, select: :title`. So this should rather be `join: :comments` instead of `join: p.comments`. However, the issue with the atom is that it only applies to the root binding, and you often do association across bindings, so the atom approach is less useful here and we don't want to add new syntax sugar

2. A lot of times the preload is the best way to bring in associations and that does support a convenience syntax. Having a very convenient syntax may make it easier for users to do the wrong thing

In any case, thanks for reaching out!

PR #3579, opened 2021-03-08
Add extra info on multi operation order

### LINKED ISSUE #3578: Ecto.Multi operations not always executed in order as added (Documentation Issue?)
### Environment

* Elixir version (elixir -v): 1.10.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): psql (PostgreSQL) 11.0
* Ecto version (mix deps): 

```
    {:ecto_sql, "~> 3.5", override: true},
    {:phoenix_ecto, "~> 4.2.1"},
```

* Database adapter and version (mix deps):  `{:postgrex, ">= 0.0.0"}`
* Operating system: Ubuntu 20.4

### Current behavior

Consider this basic example

```Elixir
    multi =
      Multi.new()
      |> Multi.run(:token, fn _, _ -> {:error, :invalid_token} end)
      |> Multi.insert(:user, %Ecto.Changeset{valid?: false})
      |> Multi.run(:invitations, fn _, _ -> {:ok, 1} end)
      |> Multi.run(:user_token, fn _, _ -> {:ok, "foo"} end)

    multi |> Multi.to_list() |> IO.inspect(label: "order")

    multi |> Repo.transaction() |> IO.inspect(label: "result")
```

Order will bi printed out as
```
order: [
  token: {:run, #Function<0.131121918/2 in Graphotate.Auth.RegisterUser.call/3>},
  user: {:insert,
   #Ecto.Changeset<action: :insert, changes: %{}, errors: [], data: nil,
    valid?: false>, []},
  invitations: {:run,
   #Function<1.131121918/2 in Graphotate.Auth.RegisterUser.call/3>},
  user_token: {:run,
   #Function<2.131121918/2 in Graphotate.Auth.RegisterUser.call/3>}
]
```

Result, however, is 

```
result: {:error, :user,
 #Ecto.Changeset<action: :insert, changes: %{}, errors: [], data: nil,
  valid?: false>, %{}}
```

### Expected behavior

Based on documentation, expected result would be 

```
result: {:error, :token, :invalid_token, %{}}
```

From the code, it seems like this is intentional. There is a precheck for all operations being valid, and one of the conditions is checking that all passed-in changesets are valid. This also makes sense in most cases as it saves a db trip.

However, the documentation does say

```
All operations will be executed in the order they were added.
```

So should this be updated, or is this behavior not intentional?

begedin:
Not sure if this was worded most appropriately. Open to any criticism. Fixes #3578

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -8,7 +8,9 @@ defmodule Ecto.Multi do
   is given a name that is unique and will identify its result in case of
   success or failure.
 
-  All operations will be executed in the order they were added.
+  All operations will be executed in the order they were added, with the
+  possible exception of a multi which contains operations that accept
+  [changesets](#module-changesets).
 
   The `Ecto.Multi` structure should be considered opaque. You can use
   `%Ecto.Multi{}` to pattern match the type, but accessing fields or
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -8,9 +8,8 @@ defmodule Ecto.Multi do
   is given a name that is unique and will identify its result in case of
   success or failure.
 
-  All operations will be executed in the order they were added, with the
-  possible exception of a multi which contains operations that accept
-  [changesets](#module-changesets).
+  If a multi is valid (i.e. all the changesets in it are valid),
+  all operations will be executed in the order they were added.
 
   The `Ecto.Multi` structure should be considered opaque. You can use
   `%Ecto.Multi{}` to pattern match the type, but accessing fields or


lib/ecto/multi.ex
```diff
@@ -8,7 +8,9 @@ defmodule Ecto.Multi do
   is given a name that is unique and will identify its result in case of
   success or failure.
 
-  All operations will be executed in the order they were added.
+  All operations will be executed in the order they were added, with the
+  possible exception of a multi which contains operations that accept
+  [changesets](#module-changesets).
```
josevalim, line 13:
```suggestion
  If a multi is valid (i.e. all the changesets in it are valid), all operations
  will be executed in the order they were added.
```

WDYT?

begedin: Definitely the more direct approach of explaining it. 👍

josevalim:

PR #3582, opened 2021-03-10
Use Decimal.equal? instead of matching

warmwaffles:
While I am building exqlite I am running into issues with how sqlite3
handles decimals when summing.

```
sqlite> create table foo (id integer primary key, cost decimal(10, 2));
sqlite> insert into foo (id,cost) values (1, '1.00');
sqlite> select sum(foo.cost) + 0 from foo;
1
```

As you can see sqlite normalizes the decimal which results in
`Decimal.new(1)` which does not match

```
Decimal.new(1) != Decimal.new("1.00") #=> true
```

When there are values after the decimal it does what you expect.

```
sqlite> create table foo (id integer primary key, cost decimal);
sqlite> insert into foo (id,cost) values (1, '1.00'), (2, '1.2'), (3, '1.3');
sqlite> select sum(foo.cost) from foo;
3.5
```

### DIFF
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -409,14 +409,20 @@ defmodule Ecto.Integration.TypeTest do
     decimal = Decimal.new("1.0")
     TestRepo.insert!(%Post{cost: decimal})
 
-    assert [^decimal] = TestRepo.all(from p in Post, where: p.cost == ^decimal, select: p.cost)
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks for working on exqlite! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #3585, opened 2021-03-12
Add microsecond precision tags in interval tests

dozoyo:
Working on an [SQLite adapter](https://github.com/warmwaffles/exqlite) for Ecto :) 

[SQLite doesn't support microsecond precision](https://sqlite.org/lang_datefunc.html), only millisecond. So in order to integrate the new exqlite adapter there needs to be tags excluding microsecond operations.

This PR includes those tags, but also includes some duplicated tests. Tests that were using microsecond operations were duplicated and modified to use millisecond operations. With this duplication, datetime data types can be validated and recognized in SQLite.

I've tested these changes with the postgrex adapter, so I'm hoping there won't be any issues with the other adapters.

### DIFF
--- a/integration_test/cases/interval.exs
+++ b/integration_test/cases/interval.exs
@@ -391,7 +391,6 @@ defmodule Ecto.Integration.IntervalTest do
            TestRepo.all(from u in Usec, select: datetime_add(u.utc_datetime_usec, ^dec, "millisecond"))
   end
 
-  @tag :decimal_precision
   test "datetime_add with naive_datetime_usec in milliseconds" do
     TestRepo.insert!(%Usec{naive_datetime_usec: ~N[2014-01-01 02:00:00.001000]})
     datetime = ~N[2014-01-01 02:00:00.151000]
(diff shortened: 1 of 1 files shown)


integration_test/cases/interval.exs
```diff
@@ -355,4 +360,61 @@ defmodule Ecto.Integration.IntervalTest do
...
+    {:ok, datetime} = DateTime.from_naive(~N[2014-01-01 02:00:00.151000], "Etc/UTC")
+    dec = Decimal.new(150)
+
+    assert [^datetime] =
+           TestRepo.all(from u in Usec, select: datetime_add(u.utc_datetime_usec, ^150.0, "millisecond"))
+    assert [^datetime] =
+           TestRepo.all(from u in Usec, select: datetime_add(u.utc_datetime_usec, ^dec, "millisecond"))
+  end
+
+  @tag :decimal_precision
```
josevalim, line 394: We don't use decimal on this test. :)

josevalim:
Just one comment and it looks good to me!

integration_test/cases/interval.exs
dozoyo: Thanks! Removed it

dozoyo:
Great! Thanks :)

josevalim:

PR #3586, opened 2021-03-14
use Decimal.equal? for coalesce on mixed type

kevinlang:
Followup to #3582 . I also added two tags we needed locally, if you don't mind.


josevalim:

PR #3591, opened 2021-03-20
Replace gender example in validate_inclusion

### LINKED ISSUE #3590: Replace gender example in validate_inclusion
I would propose to change the example of `validate_inclusion` ([see here](https://github.com/elixir-ecto/ecto/blob/v3.5.8/lib/ecto/changeset.ex#L1956)) to a more neutral one. Using a gender and having it set to a default set of values advertises a way that will for sure discriminate against some people. There are so many other more neutral examples:

* Cardinal direction
* List of colors
* Preferred interface Languages
* And some more

I also would create a PR for this if this is wanted, and we decided what example is more neutral.

bitboxer:
closes #3590

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1963,7 +1963,7 @@ defmodule Ecto.Changeset do
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3592, opened 2021-03-23
Support dynamic hints

ukutaht:
See discussion starting here https://github.com/elixir-ecto/ecto/issues/2392#issuecomment-802706968


josevalim:

ukutaht:
I'm trying to integrate this into my app now but running into some trouble.

I think what's happening is that different values in the `hints` field do not invalidate the query cache entry. What I'm seeing is that whatever dynamic `SAMPLE` hint I've used for the first time gets used for every subsequent query.

Is there a way to manually override the cache when writing an Ecto query? Or would it make sense to add `hints` into the cache key?

josevalim:
@ukutaht it has been fixed in master. Can you please give it a try? If it works, I will ship a new release.

ukutaht:
Works great!

That was so quick, thank you so much @josevalim  🙏🏻

PR #3594, opened 2021-03-26
Add Ecto.Changeset.state/1

nathanl:
Provides a public API for determining a changeset's state in data
storage.

I suspect there is a reason this function doesn't already exist, but in
case it was an oversight, here it is. 😄 

Context: when a record has a field that should not change, I have
used `changeset.data.__meta__.state` in a validation function
to ensure that it doesn't. For example:

```
  defp ensure_hostname_is_permanent(changeset) do
    state = changeset.data.__meta__.state

    if get_change(changeset, :hostname) && state != :built do
      add_error(changeset, :hostname, "Cannot change hostname after persisting")
    else
      changeset
    end
  end
```

Using `changeset.data.__meta__.state` seems hacky, since the `__meta__`
seems to imply "this is a private implementation detail".

If there is a better way to approach this problem, I'd be interested to
hear it.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -103,7 +103,7 @@ defmodule Ecto.Integration.RepoTest do
     assert %Post{__meta__: ^loaded_meta} = TestRepo.insert!(post)
 
     post = TestRepo.one(Post)
-    assert post.__meta__.state == :loaded
+    assert Ecto.Changeset.state(post) == :loaded
     assert post.inserted_at
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
`Ecto.get_meta(changeset.date, :state)`. It is not very readable though, but it does what you want. :)

nathanl:
I see that you meant `Ecto.get_meta(changeset.data, :state)`. I didn't know about that. Thanks! :)

Should this PR be closed? Or be changed to define something like this?

```
def get_meta(%Ecto.Changeset{} = cs, key), do: Ecto.get_meta(cs.data, key)
```

josevalim:
Yeah, I don't want to add it to the changeset because not all changeset.data is a struct-backed by a schema. :)

PR #3600, opened 2021-04-01
add 2 error messages to catch invalid parameters being passed to joins

mayel:
After spending longer than I wanted to debugging my code by adding inspects to Ecto, here are a couple error messages I added to make it easier next time. Happy to refactor if necessary...

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1071,7 +1071,7 @@ defmodule Ecto.Query do
                            "list of variables, got: #{Macro.to_string(binding)}"
   end
 
-  defmacro join(_query, _qual, binding, _expr, opts) when is_list(binding) do
+  defmacro join(_query, _qual, binding, _expr, opts) do
     raise ArgumentError, "invalid opts passed to Ecto.Query.join/5, should be " <>
                            "list, got: #{Macro.to_string(opts)}"
   end
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1071,6 +1071,11 @@ defmodule Ecto.Query do
                            "list of variables, got: #{Macro.to_string(binding)}"
   end
 
+  defmacro join(_query, _qual, binding, _expr, opts) when is_list(binding) do
+    raise ArgumentError, "invalid opts passed to Ecto.Query.join/5, should be " <>
+                           "list, got: #{Macro.to_string(opts)}"
+  end
+
   @doc ~S'''
   A common table expression (CTE) also known as WITH expression.
 
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1071,7 +1071,7 @@ defmodule Ecto.Query do
                            "list of variables, got: #{Macro.to_string(binding)}"
   end
 
-  defmacro join(_query, _qual, binding, _expr, opts) do
+  defmacro join(_query, _qual, _binding, _expr, opts) do
     raise ArgumentError, "invalid opts passed to Ecto.Query.join/5, should be " <>
                            "list, got: #{Macro.to_string(opts)}"
   end
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -572,12 +572,12 @@ defmodule Ecto.Query.Planner do
     with {:ok, source} <- Enum.reverse(sources) |> Enum.fetch(ix) do
       source
     else _ ->
-      raise ArgumentError, message: "could not find a source with index `#{ix}` in `#{inspect sources}"
+      raise ArgumentError, "could not find a source with index `#{ix}` in `#{inspect sources}"
     end
   end
 
   defp fetch_source!(_, ix) do
-    raise ArgumentError, message: "invalid binding index: `#{inspect ix}` (check if you're binding using a valid :as atom)"
+    raise ArgumentError, "invalid binding index: `#{inspect ix}` (check if you're binding using a valid :as atom)"
   end
 
   defp schema_for_association_join!(query, join, source) do
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -434,60 +434,69 @@ defmodule Ecto.Query.Planner do
 
   defp plan_joins([%JoinExpr{assoc: {ix, assoc}, qual: qual, on: on, prefix: prefix} = join|t],
                      query, joins, sources, tail_sources, counter, offset, adapter) do
-    schema = schema_for_association_join!(query, join, Enum.fetch!(Enum.reverse(sources), ix))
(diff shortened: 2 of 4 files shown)


v0idpwn:
I'd recommend adding some tests to these scenarios so it doesn't become hard to maintain later on :)

lib/ecto/query/planner.ex
```diff
@@ -434,60 +434,69 @@ defmodule Ecto.Query.Planner do
...
-    schema = schema_for_association_join!(query, join, Enum.fetch!(Enum.reverse(sources), ix))
-    refl = schema.__schema__(:association, assoc)
 
-    unless refl do
-      error! query, join, "could not find association `#{assoc}` on schema #{inspect schema}"
-    end
+    with r_sources = Enum.reverse(sources),
+         true <- is_integer(ix),
+         {:ok, source} <- Enum.fetch(r_sources, ix) do
+
```
v0idpwn, line 441: Instead using `with`, you could write a private function that validates & fetches ix or raises otherwise, as you're raising on else anyway

josevalim: Yes, agreed. Instead of nesting all code, you can call `validate_sources!` and keep the logic only in there. In any case, great PR! :+1:

lib/ecto/query.ex
```diff
@@ -1071,6 +1071,11 @@ defmodule Ecto.Query do
                            "list of variables, got: #{Macro.to_string(binding)}"
   end
 
+  defmacro join(_query, _qual, binding, _expr, opts) when is_list(binding) do
```
kelvinst, line 1074: Cool! But what about if both `binding` and `opts` are not lists? Shouldn't we throw better exceptions. As far as I understand here, in that case we would throw a `FunctionClauseError` 🤔

lib/ecto/query/planner.ex
```diff
@@ -434,60 +434,69 @@ defmodule Ecto.Query.Planner do
 
   defp plan_joins([%JoinExpr{assoc: {ix, assoc}, qual: qual, on: on, prefix: prefix} = join|t],
                      query, joins, sources, tail_sources, counter, offset, adapter) do
-    schema = schema_for_association_join!(query, join, Enum.fetch!(Enum.reverse(sources), ix))
-    refl = schema.__schema__(:association, assoc)
```
kelvinst, line 437: Empty line after function header

kelvinst: I guess `mix format` should remove it right?

lib/ecto/query/planner.ex
kelvinst: What about `fetch_source!(sources, ix)` that actually does this with inside and just returns `source`? It would be used like: `source = fetch_source!(sources, ix)`

kelvinst (approved):
Really liked it, just had some questions.

lib/ecto/query/planner.ex
josevalim: We are not formatting Ecto files though, yet.

lib/ecto/query/planner.ex
mayel: Sounds good!

lib/ecto/query.ex
mayel: There's already a check for that.

lib/ecto/query/planner.ex
mayel: Refactored and added tests :)

mayel:
Refactored as suggested and added tests, please review again.

lib/ecto/query.ex
kelvinst: Are you sure? I couldn't find it, the other clauses match for when both `binding` and `opts` are lists or when only `opts` is a list. You added a clause for when only `binding` is a list. If both are not lists, no clause will match and one will get a `FunctionClauseError`

mayel: Ah gotcha! Latest commit should cover it (dev would get one error to fix, and then the other).

lib/ecto/query/planner.ex
```diff
@@ -567,6 +568,18 @@ defmodule Ecto.Query.Planner do
 
   defp rewrite_param_ix(param, _, _, _, _), do: param
 
+  defp fetch_source!(sources, ix) when is_integer(ix) do
+    with {:ok, source} <- Enum.reverse(sources) |> Enum.fetch(ix) do
+      source
+    else _ ->
+      raise ArgumentError, message: "could not find a source with index `#{ix}` in `#{inspect sources}"
```
josevalim, line 575:
```suggestion
      raise ArgumentError, "could not find a source with index `#{ix}` in `#{inspect sources}"
```

lib/ecto/query/planner.ex
```diff
@@ -567,6 +568,18 @@ defmodule Ecto.Query.Planner do
...
+    end
+  end
+
+  defp fetch_source!(_, ix) do
+    raise ArgumentError, message: "invalid binding index: `#{inspect ix}` (check if you're binding using a valid :as atom)"
```
josevalim, line 580:
```suggestion
    raise ArgumentError, "invalid binding index: `#{inspect ix}` (check if you're binding using a valid :as atom)"
```

test/ecto/query/planner_test.exs
```diff
@@ -340,6 +340,15 @@ defmodule Ecto.Query.PlannerTest do
            "&5.id() == &3.special_long_comment_id() and fragment({:raw, \"LEN(\"}, {:expr, &5.text()}, {:raw, \") > 100\"})"
   end
 
+  test "plan: raises on invalid binding index in join" do
+    query   = from(p in Post, as: :posts)
+    |> join(:left, [{p, :foo}], assoc(p, :comments))
```
josevalim, line 345:
```suggestion
    query =
      from(p in Post, as: :posts)
      |> join(:left, [{p, :foo}], assoc(p, :comments))
```

lib/ecto/query/planner.ex
```diff
@@ -567,6 +568,18 @@ defmodule Ecto.Query.Planner do
 
   defp rewrite_param_ix(param, _, _, _, _), do: param
 
+  defp fetch_source!(sources, ix) when is_integer(ix) do
+    with {:ok, source} <- Enum.reverse(sources) |> Enum.fetch(ix) do
+      source
+    else _ ->
+      raise ArgumentError, "could not find a source with index `#{ix}` in `#{inspect sources}"
+    end
```
josevalim, line 576:
```suggestion
    case Enum.reverse(sources) |> Enum.fetch(ix) do
      {:ok, source} -> source
      :error -> raise ArgumentError, "could not find a source with index `#{ix}` in `#{inspect sources}"
    end
```

josevalim:

PR #3602, opened 2021-04-04
Tag tests that use CONCAT fragments

warmwaffles:
In SQLite3 we do not have access to a CONCAT function. Instead we have
to use `||` with an example like the following:

    SELECT p0.foo || ' thing ' || p0.bar from posts as p0

Not ideal, but it will allow us downstream to ignore these tests and
we'll make a test for the same functionality within the ecto_sqlite3
library itself.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -884,6 +884,7 @@ defmodule Ecto.Integration.RepoTest do
   describe "insert_all with source query" do
     @tag :upsert
     @tag :with_conflict_target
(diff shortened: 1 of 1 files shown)


josevalim:
Awesome! If you want, we should probably implement `<>` in Ecto.Query and translate that accordingly, but this works great too!

josevalim:

warmwaffles:
@josevalim I'll explore that option. Seems like it would be nicer syntactically too.

Issue #3604, opened 2021-04-06
apply_changes try to access `id` field in relations when it's `nil`


rockneurotiko:
### Environment

* Elixir version (elixir -v): 1.11.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 11.9
* Ecto version (mix deps): 3.6.0
* Database adapter and version (mix deps): Postgrex, 0.15.8
* Operating system: Linux

### Current behavior

When creating an optional relation (with belongs_to) in a changeset as `nil`, when trying to apply the changes it raises an error trying to access the field `id` of the non existing relation:

Example of the issue with a test case

``` elixir

defmodule TestRelation do
  use ExUnit.Case, async: true

  import Ecto.Changeset

  defmodule Inner do
    use Ecto.Schema

    schema "inner" do
      field(:n, :integer)
    end

    def changeset(schema \\ %__MODULE__{}, attrs) do
      schema |> Ecto.Changeset.cast(attrs, [:n])
    end
  end

  alias __MODULE__.Inner

  defmodule Outer do
    use Ecto.Schema

    schema "outer" do
      belongs_to(:inner, Inner)

      field(:m, :integer)
    end

    def changeset(schema \\ %__MODULE__{}, attrs) do
      schema
      |> Ecto.Changeset.cast(attrs, [:m])
    end
  end

  alias __MODULE__.Outer

  test "with put_change" do
    inner = %Inner{n: 3}

    outer = Outer.changeset(%{"m" => 2}) |> put_assoc(:inner, inner) |> apply_changes()

    assert outer.m == 2
    assert outer.inner.n == 3

    outer = Outer.changeset(%{"m" => 2}) |> put_assoc(:inner, nil) |> apply_changes()

    assert outer.m == 2
    assert outer.inner == nil
  end

  test "with cast_assoc" do
    outer =
      Outer.changeset(%{m: 2, inner: %{n: 3}})
      |> cast_assoc(:inner)
      |> apply_changes()

    assert outer.m == 2
    assert outer.inner.n == 3

    outer =
      Outer.changeset(%{m: 2, inner: nil})
      |> cast_assoc(:inner)
      |> apply_changes()

    assert outer.m == 2
    assert outer.inner == nil
  end
end
``` 

When executing this tests, the following error raises in both tests:

```
[...]

PR #3605, opened 2021-04-06
Add the relation id in apply_changes if the id exists

rockneurotiko:
This PR fix the issue in #3604

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2970,8 +2970,8 @@ defmodule Ecto.Changeset do
 
     acc = Map.put(acc, key, relation_changed)
 
-    case relation do
-      %Ecto.Association.BelongsTo{} -> Map.put(acc, relation.owner_key, relation_changed.id)
+    case {relation, relation_changed} do
+      {%Ecto.Association.BelongsTo{}, %{id: id}} -> Map.put(acc, relation.owner_key, id)
       _ -> acc
     end
   end
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2970,8 +2970,10 @@ defmodule Ecto.Changeset do
 
     acc = Map.put(acc, key, relation_changed)
 
-    case {relation, relation_changed} do
-      {%Ecto.Association.BelongsTo{}, %{id: id}} -> Map.put(acc, relation.owner_key, id)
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -2970,8 +2970,8 @@ defmodule Ecto.Changeset do
 
     acc = Map.put(acc, key, relation_changed)
 
-    case relation do
-      %Ecto.Association.BelongsTo{} -> Map.put(acc, relation.owner_key, relation_changed.id)
+    case {relation, relation_changed} do
+      {%Ecto.Association.BelongsTo{}, %{id: id}} -> Map.put(acc, relation.owner_key, id)
```
josevalim, line 2974: Unfortunately we cannot assume the primary key is id. However, in this case, your point is that `relation_changed` is nil, right? I think checking only for a map should be enough then?

```suggestion
      {%Ecto.Association.BelongsTo{}, %{}} -> Map.put(acc, relation.owner_key, id)
```

v0idpwn: @josevalim wasn't the code assuming that the primary key was id before? How would it behave with different keys? got me thinking

rockneurotiko: You are right, the original code assumed that the primary key is id, and I didn't thought about it.

Maybe it makes sense to use the `related_key` of the BelongsTo information?

``` elixir
    case {relation, relation_changed} do
      {%Ecto.Association.BelongsTo{related_key: related_key}, %{} = related} when is_map_key(related, related_key) ->
        Map.put(acc, relation.owner_key, Map.get(related, related_key))
      _ -> acc
    end
```

josevalim: Ah, yes, it was. We should probably add a test then for custom belongs to in this function. In this case, we could rewrite it to:

    case {relation, relation_changed} do
      {%Ecto.Association.BelongsTo{related_key: related_key}, %{^related_key => id}} ->
        Map.put(acc, relation.owner_key, id)
      _ -> acc
    end


You may need to write it as a `with` though.

rockneurotiko: I wrote it using `with`, I tried the match before with the case and didn't work because the variable is not defined before, but I didn't realized to use with.

I also modified the test so it checks that the related_key is being used, if you prefer the test in other way let me know (other belongs_to instead of using the existing Category, for example)

josevalim:

PR #3611, opened 2021-04-12
Support overriding base query for unsafe_validate_unique/4

### LINKED ISSUE #3610: changing the base query of unsafe_validate_unique
I'm thinking of creating a `unsafe_validate_unique` that accepts an option that changes the base query. Our usecase is the use of soft deletes, we want to exclude rows that are marked as `deleted`. I think it's a relatively small change and I can create a PR if needed.

dvic:
Closes #3610

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1832,12 +1832,17 @@ defmodule Ecto.Changeset do
 
     * `:repo_opts` - the options to pass to the `Ecto.Repo` call.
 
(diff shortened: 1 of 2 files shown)


dvic:
I had to change `maybe_exclude_itself` to use a dynamic expression because it was using `or_where` before, but with a custom base query this won't work. The current structure of the query is `base_query AND NOT (matches_primary_keys) AND matches_unique_fields`. The `matches_primary_keys` is only included if non of the values are `nil`. I tried my best to test all these cases.

josevalim:

PR #3613, opened 2021-04-13
Clarify empty_values description

Apelsinka223:
Hi! Current description is a little bit misleading, and sounds like fields with empty values would be discarded at all if there is no default value for the field. I propose to clarify a description

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -427,7 +427,7 @@ defmodule Ecto.Changeset do
   ## Options
 
     * `:empty_values` - a list of values to be considered as empty when casting.
-      All empty values are replaced with default values on cast if default values are specified. Defaults to `[""]`
+      Empty values are always replaced by the default value of the respective key. Defaults to `[""]`
 
   ## Examples
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -427,7 +427,7 @@ defmodule Ecto.Changeset do
   ## Options
 
     * `:empty_values` - a list of values to be considered as empty when casting.
-      All empty values are discarded on cast. Defaults to `[""]`
+      All empty values are replaced with default values on cast if default values are specified. Defaults to `[""]`
 
   ## Examples
 


lib/ecto/changeset.ex
```diff
@@ -427,7 +427,7 @@ defmodule Ecto.Changeset do
   ## Options
 
     * `:empty_values` - a list of values to be considered as empty when casting.
-      All empty values are discarded on cast. Defaults to `[""]`
+      All empty values are replaced with default values on cast if default values are specified. Defaults to `[""]`
```
josevalim, line 430: I am not sure this is clearer. Can you please provide an example of what this sentence is meant to clarify? Thank you.

Apelsinka223: ### Current behavior without defaults:

```
defmodule User do
  schema "users" do
      field :first_name, :string
      field :last_name, :string
  end   
  
  def changeset(user, attrs) do
      user
      |> cast(attrs, [:first_name, :last_name], empty_values: [""])
  end
end

user = %User{first_name: "some name"} # user has non-empty first_name
User.changeset(user, %{first_name: ""}) 
=> 
#Ecto.Changeset<
  action: nil,
  changes: %{first_name: ""}, # first_name no discarded from changeset and set to ""
  errors: [],
  data: #User<>,
  valid?: true
> 
```

### Current behavior with defaults:

```
defmodule User do
  schema "users" do
      field :first_name, :string, default: "first_name"
      field :last_name, :string
  end   
  
  def changeset(user, attrs) do
      user
      |> cast(attrs, [:first_name, :last_name], empty_values: [""])
  end
end

user = %User{first_name: "some name"} # user has non-empty first_name
User.changeset(user, %{first_name: ""}) 
=> 
#Ecto.Changeset<
  action: nil,
  changes: %{first_name: "first_name"}, # first_name no discarded from changeset and set to default value "first_name"
  errors: [],
  data: #User<>,
  valid?: true
> 
```

### Expected behavior regardless of whether there is default value for field or not:
```
user = %User{first_name: "some name"} # user has non-empty first_name
User.changeset(user, %{first_name: ""}) 
=> 
#Ecto.Changeset<
  action: nil,
  changes: %{}, # no changes, because fields with empty_values where discarded
  errors: [],
  data: #User<>,
  valid?: true
> 
```

I apologize If there is any issue with construction of the phrase, feel free to correct it.

josevalim: Thank you for the write up! I get a different result for your first code snippet though:

```elixir
iex(2)> user = %User{first_name: "some name"} # user has non-empty first_name
%User{
  __meta__: #Ecto.Schema.Metadata<:built, "users">,
  first_name: "some name",
  id: nil,
  last_name: nil
}
iex(3)> User.changeset(user, %{first_name: ""})
#Ecto.Changeset<
  action: nil,
  changes: %{first_name: nil},
  errors: [],
  data: #User<>,
  valid?: true
>
```

At least on Ecto v3.6, the empty value is always discarded and the default value is used instead. So maybe we should say:

```suggestion
      Empty values are always discarded and the default value for the respective key is used instead. Defaults to `[""]`
```

?

Apelsinka223: Oh, you are right about first example, I made an error. 
Maybe I am wrong in interpretation of "discarded" meaning, but for me it sounds like the change wouldn't be passed to the changeset at all. And I'd rather change this word on something more related to "nullifying".

josevalim:

PR #3616, opened 2021-04-14
Fix documentation typo

leonzalion:
I think it should be `where: p.id == any` instead of `where: p.id = any` if I'm not mistaken :)


josevalim:

PR #3620, opened 2021-04-21
Allows *_many relations to be considered empty if they are nil.

Adzz:
Right now you can set a changeset to have an `nil` for a has_many / embeds_many relation, but if you do that `Ecto.Changeset.Relation.empty?/2` errors. I think it should instead return true.

I don't know if this is a good idea so I thought I'd open a PR to see.

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -32,9 +32,9 @@ defmodule Ecto.Changeset.Relation do
   Checks if the container can be considered empty.
   """
   def empty?(%{cardinality: _}, %NotLoaded{}), do: true
+  def empty?(%{cardinality: _}, nil), do: true
   def empty?(%{cardinality: :many}, []), do: true
   def empty?(%{cardinality: :many}, changes), do: filter_empty(changes) == []
-  def empty?(%{cardinality: :one}, nil), do: true
   def empty?(%{}, _), do: false
 
   @doc """


josevalim:
> Right now you can set a changeset to have an nil for a has_many / embeds_many relation

How are you setting to nil? What do you mean that you can set it? :) I am not sure if `nil` is actually valid generally and you might run into further issues.

Adzz:
Sorry I should have explained!

```elixir
Mix.install([:ecto])

defmodule Mate do
  use Ecto.Schema

  embedded_schema do
    field(:name, :string)
  end
end

defmodule Thing do
  use Ecto.Schema

  embedded_schema do
    field(:name, :string)
    has_many(:mates, Mate)
  end
end


data = %{name: "Ted Danson"}
(
  changeset = Ecto.Changeset.cast(%Thing{mates: nil}, data, [:name]) 
  |> Ecto.Changeset.cast_assoc(:mates)
)

{_, relation} =  Map.get(Thing.__changeset__(), :mates)

relation_in_changes = Map.get(changeset.changes, :mates)
relation_in_data = Map.fetch!(changeset.data, :mates)

# These both error:
Ecto.Changeset.Relation.empty?(relation, relation_in_changes)
Ecto.Changeset.Relation.empty?(relation, relation_in_data)
```

josevalim:
Nothing fails because we are likely checking and/or normalizing nils before.  :)

Keep in mind `Ecto.Changeset.Relation.empty?` are private functions, so I still don't understand which error you are running into. Unless the error is in calling the function directly, which you should not!

Adzz:
Yes the error is calling them directly the last two lines in the snippet above. 

Ah okay i didn’t know they were private! I can work around it instead.

PR #3624, opened 2021-04-25
Recursively search for parent_as binding

v0idpwn:
Work in progress attempt to fix #3619.

PR #3627, opened 2021-04-26
fix: parent_as when used at right side of a combining expression

dgvncsz0f:
This patch fixes a bug in which is impossible to compile a query using
`parent_as` at the right side of a union expression (or any other type
of combination). As an example, consider the following query:

```
  subquery = from(c in Comments, where: parent_as(:posts).id == c.post_id)
  from(p in Posts, where: exists(union_all(subquery, ^subquery)))
```

This query should fail to compile with the following error:

```
** (Ecto.SubQueryError) the following exception happened when compiling a subquery.
    ** (Ecto.QueryError) `parent_as(:post)` can only be used in subqueries in query:
```

The problem, as it seems, is that the `plan_subquery` does not include
the required alias on the combination queries -- the right side query
of the union operator in this case. Thus, when the times come to
resolve the alias, it fails with the above message.

To fix this issue, the function `plan_subquery` has been modified to
include the `@parent_as` alias for every elemen in the `combination`
field. This is done recursevily because the query may have two or more
combining queries.

There is also a required change on ~ecto_adapters~ ecto_sql repository to make
this solution works, which I'll be submiting patches for as well.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -258,13 +258,12 @@ defmodule Ecto.Query.Planner do
        do: error!(query, expr, "cannot set prefix: #{inspect(prefix)} option for fragment joins")
 
   defp plan_subquery(subquery, query, prefix, adapter, source?) do
-    %{query: inner_query} = subquery
-
-    inner_query = %{
-      inner_query
-      | prefix: prefix || subquery.query.prefix || query.prefix,
-        aliases: Map.put(inner_query.aliases, @parent_as, query)
-    }
+    inner_query =
+      modify_subquery_for_plan(subquery.query, fn subquery ->
+        subquery
+        |> subquery_put_prefix(prefix, query)
+        |> subquery_put_parent_as_alias(query)
+      end)
 
     {inner_query, params, key} = plan(inner_query, :all, adapter)
     assert_no_subquery_assocs!(inner_query)
@@ -274,7 +273,8 @@ defmodule Ecto.Query.Planner do
       |> ensure_select(true)
       |> normalize_subquery_select(adapter, source?)
 
-    {_, inner_query} = pop_in(inner_query.aliases[@parent_as])
+    inner_query = modify_subquery_for_plan(inner_query, &subquery_drop_parent_as_alias/1)
+
     %{subquery | query: inner_query, params: params, cache: key, select: select}
   rescue
     e -> raise Ecto.SubQueryError, query: query, exception: e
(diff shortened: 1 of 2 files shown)


test/ecto/query/planner_test.exs
```diff
@@ -839,6 +839,16 @@ defmodule Ecto.Query.PlannerTest do
     assert Macro.to_string(hd(hd(query.joins).source.query.wheres).expr) =~ "in %Ecto.SubQuery{"
   end
 
+  test "normalize: parent_as used at right side term of union expression" do
```
dgvncsz0f, line 842: it is not clear to me what kind of tests are necessary to test this change. Please advise.

lib/ecto/query/planner.ex
```diff
@@ -1079,10 +1079,10 @@ defmodule Ecto.Query.Planner do
   end
   defp prewalk_source(%Ecto.SubQuery{query: inner_query} = subquery, kind, query, _expr, counter, adapter) do
     try do
-      inner_query = put_in inner_query.aliases[@parent_as], query
+      inner_query = modify_subquery_for_plan(inner_query, & subquery_put_parent_as_alias(&1, query))
```
josevalim, line 1082:
```suggestion
      inner_query = modify_subquery_for_plan(inner_query, &subquery_put_parent_as_alias(&1, query))
```

josevalim:
Is this still necessary after #3624? I would prefer an approach as in that PR, where we traverse up, rather than changing all combinations as we traverse down, which is bound to be expensive.

dgvncsz0f:
@josevalim I've tested a similar query using the [master branch](https://github.com/elixir-ecto/ecto/commit/6982d4802233cdd8f83d4352edbaee86b7b31090) and also it triggers the error. As the PR you mention is merged, perhaps this is something new?

dgvncsz0f:
> Is this still necessary after #3624? I would prefer an approach as in that PR, where we traverse up, rather than changing all combinations as we traverse down, which is bound to be expensive.

yeah, it makes sense. At the moment. it is not clear to me how to implement this fix the way you are describing, but I can give it another go. If someone more familiar with the codebase could give me some pointers it would be highly appreciated.

dgvncsz0f:
the ecto_sql related changes: https://github.com/elixir-ecto/ecto_sql/pull/322

dgvncsz0f:
👋🏽 any updates on this?

josevalim:
Hi @dgvncsz0f, I would still prefer a mechanism that traverses up, to reduce the cost. So that needs to be investigated, before we can merge this, I haven’t been able to do that yet.

josevalim:
Closing for now. Glad to reopen if the problem still exists after PRs such as: https://github.com/elixir-ecto/ecto_sql/pull/338

PR #3629, opened 2021-05-02
Add `doc false` to parameterised types functions

wojtekmach:
Otherwise, they show up in docs: https://hexdocs.pm/ecto_map_set/EctoMapSet.html#functions


josevalim (approved):
Should we use `@impl true` instead? Anyway, it looks good to go. :)

josevalim:
Ah, the reason is that if the user overwrites it, we want them to set @impl true in case they are setting it on other functions. So yeah, this is good to go as is! :+1:

PR #3636, opened 2021-05-14
fix: insert_all on_conflict with conditional query using placeholders

lmarlow:
The parameter index was off for `insert_all` when using placeholders (#3515) and a conditional `on_conflict` query with a parameter. This fixes the counter function by not counting placeholder values since their parameters are reused.

Before this it would generate errors like the following:

```
** (Postgrex.Error) ERROR 42P18 (indeterminate_datatype) could not determine data type of parameter $6

query: INSERT INTO "posts" AS p0 ("posted","title","uuid") VALUES ($1,$2,$3),($1,$4,$5) ON CONFLICT ("uuid") DO UPDATE SET "title" = 'updated' WHERE (p0."title" != $7)
````

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1070,6 +1070,7 @@ defmodule Ecto.Integration.RepoTest do
       assert [100] == TestRepo.all(query)
     end
 
+    @tag :upsert_all
     @tag :with_conflict_target
     test "Repo.insert_all upserts and fills in placeholders with conditioned on_conflict query" do
       do_not_update_title = "don't touch me"
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1073,16 +1073,38 @@ defmodule Ecto.Integration.RepoTest do
     @tag :with_conflict_target
     test "Repo.insert_all upserts and fills in placeholders with conditioned on_conflict query" do
       do_not_update_title = "don't touch me"
-      on_conflict = from p in Post, update: [set: [title: "updated"]], where: p.title != ^do_not_update_title
+
+      on_conflict =
+        from p in Post, update: [set: [title: "updated"]], where: p.title != ^do_not_update_title
+
       placeholders = %{posted: Date.utc_today(), title: "title"}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by lmarlow** [comment]
The build failures are because `tds` only supports `on_conflict: :raise`. Should the `tds` tests be run excluding `:with_conflict_target` tags, like `myxql` does?

File: integration_test/cases/repo.exs [hunk 1]
```diff
@@ -1069,6 +1069,45 @@ defmodule Ecto.Integration.RepoTest do
       query = from(b in Barebone, select: b.num)
       assert [100] == TestRepo.all(query)
     end
+
+    @tag :with_conflict_target
+    test "Repo.insert_all upserts and fills in placeholders with conditioned on_conflict query" do
```
v0idpwn, line 1074: I think its failing on tds because it is missing the `:upsert` tag.

v0idpwn (commented):
Nice!

integration_test/cases/repo.exs
```diff
@@ -1069,6 +1069,45 @@ defmodule Ecto.Integration.RepoTest do
       query = from(b in Barebone, select: b.num)
       assert [100] == TestRepo.all(query)
     end
+
+    @tag :with_conflict_target
```
josevalim, line 1073:
```suggestion
    @tag :upsert
    @tag :with_conflict_target
```

lib/ecto/repo/schema.ex
```diff
@@ -89,7 +89,15 @@ defmodule Ecto.Repo.Schema do
...
     header = Map.keys(header)
 
-    counter = fn -> Enum.reduce(rows, 0, &length(&1) + &2) end
+    placeholder_size = map_size(placeholder_dump)
+
+    counter = fn ->
+      Enum.reduce(
+        rows,
+        placeholder_size,
+        &(Enum.count(&1, fn {_, val} -> !match?({:placeholder, _}, val) end) + &2)
```
josevalim, line 98:
```suggestion
        &(Enum.count(&1, fn {_, val} -> not match?({:placeholder, _}, val) end) + &2)
```

josevalim:

lmarlow:
Awesome! Thanks for putting the final touches on it, @josevalim ❤️

Issue #3351, opened 2020-07-02
Assoc composite foreign key support


perezd:
I have the following tables:

```sql
CREATE TABLE locations (
  state VARCHAR(2) NOT NULL,
  zip VARCHAR(5) NOT NULL,
  id INT8 NOT NULL,
  name VARCHAR(255) NOT NULL,
  CONSTRAINT "primary" PRIMARY KEY (state ASC, zip ASC, id ASC),
  FAMILY "primary" (state, zip, id, name)
)
```

```sql
CREATE TABLE addresses (
  id UUID NOT NULL DEFAULT gen_random_uuid(),
  line_1 VARCHAR(255) NOT NULL,
  line_2 VARCHAR(255) NULL,
  line_3 VARCHAR(255) NULL,
  city VARCHAR(255) NOT NULL,
  state VARCHAR(2) NOT NULL,
  zip VARCHAR(5) NOT NULL,
  inserted_at TIMESTAMP(0) NOT NULL,
  updated_at TIMESTAMP(0) NOT NULL,
  CONSTRAINT "primary" PRIMARY KEY (id ASC),
  FAMILY "primary" (id, line_1, line_2, line_3, city, state, zip, inserted_at, updated_at)
)
```

```sql
CREATE TABLE locations_addresses (
  state VARCHAR(2) NOT NULL,
  zip VARCHAR(5) NOT NULL,
  id INT8 NOT NULL,
  address_id UUID NOT NULL,
  CONSTRAINT "primary" PRIMARY KEY (state ASC, zip ASC, id ASC, address_id ASC),
  CONSTRAINT fk_locations FOREIGN KEY (state, zip, id) REFERENCES locations(state, zip, id),
  CONSTRAINT fk_addresses FOREIGN KEY (address_id) REFERENCES addresses(id),
  INDEX locations_addresses_auto_index_fk_addresses (address_id ASC),
  FAMILY "primary" (state, zip, id, address_id)
) INTERLEAVE IN PARENT locations (state, zip, id)
```
I need to be able to designate a many_to_many relationship where one side of the relationship is a composite foreign key. In my `Location` schema, I expected to be able to define the following:

```elixir
    many_to_many :addresses, Address,
      join_through: "locations_addresses",
      join_keys: [location_id: [:state, :zip, :id], address_id: :id]
```
but this returns the following error:

```sh
[...]

PR #3638, opened 2021-05-18
Support associations with composite foreign keys

soundmonster:
This PR introduces support for associations on composite foreign keys.

### Motivation
It's possible to define composite foreign keys in migrations using `refences(..., with: [...])`, but Ecto doesn't currently support multi-column references out-of-the-box. Splitting the primary key into multiple columns allows using a part of the primary key for partitioning, constraints/validation, etc.   

### Rough idea

* Allow lists of atoms (in addition to just an atom) for `:reference`, `:foreign_keys`, `:join_keys` and `:type` in schema macros, and
* Allow only lists of atoms `:owner_key` and`:related_key` in association structs

### Status
This is a work in progress with very limited functionality. Early feedback is highly appreciated! What has been worked on so far:

- [x] Schema support
  - [x] has_one
  - [x] has_many
  - [x] has_through
  - [x] belongs_to
  - [x] many_to_many
- [x] Changeset support
- [x] Integration tests
  - [x] has_one
  - [x] has_many
  - [x] has_through
  - [x] belongs_to
  - [x] many_to_many
- [x] Update docs

### DIFF
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -433,7 +433,8 @@ defmodule Ecto.Integration.PreloadTest do
       TestRepo.all(
         from c in CompositePk,
              join: cc in "composite_pk_composite_pk",
-             where: cc.a_1 in ^composite_ids_a and cc.b_1 in ^composite_ids_b and cc.a_2 == c.a and cc.b_2 == c.b,
+             on: cc.a_2 == c.a and cc.b_2 == c.b,
+             where: cc.a_1 in ^composite_ids_a and cc.b_1 in ^composite_ids_b,
              order_by: [c.a, c.b],
              select: map(c, [:name])
       )
@@ -449,7 +450,8 @@ defmodule Ecto.Integration.PreloadTest do
       TestRepo.all(
         from c in CompositePk,
              join: cc in "composite_pk_composite_pk",
-             where: cc.a_1 in ^composite_ids_a and cc.b_1 in ^composite_ids_b and cc.a_2 == c.a and cc.b_2 == c.b,
+             on: cc.a_2 == c.a and cc.b_2 == c.b,
+             where: cc.a_1 in ^composite_ids_a and cc.b_1 in ^composite_ids_b,
              order_by: [c.a, c.b],
              select: {[cc.a_1, cc.b_1], map(c, [:name])}
       )


soundmonster:
Might be related to #3351

Earthfile
```diff
@@ -60,7 +60,7 @@ integration-test-base:
         apk del .build-dependencies && rm -f msodbcsql*.sig mssql-tools*.apk
     ENV PATH="/opt/mssql-tools/bin:${PATH}"
 
-    GIT CLONE https://github.com/elixir-ecto/ecto_sql.git /src/ecto_sql
+    GIT CLONE --branch composite_foreign_keys https://github.com/soundmonster/ecto_sql.git /src/ecto_sql
```
soundmonster, line 63: This branch needs a few changes to the base migration used for the integration tests; I've opened a [PR](https://github.com/elixir-ecto/ecto_sql/pull/324) for this so I hope to be able to drop this line soon.

integration_test/support/schemas.exs
```diff
@@ -54,6 +54,8 @@ defmodule Ecto.Integration.Post do
     has_one :update_permalink, Ecto.Integration.Permalink, foreign_key: :post_id, on_delete: :delete_all, on_replace: :update
     has_many :comments_authors, through: [:comments, :author]
     belongs_to :author, Ecto.Integration.User
+    belongs_to :composite, Ecto.Integration.CompositePk,
+      foreign_key: [:composite_a, :composite_b], references: [:a, :b], type: [:integer, :integer], on_replace: :nilify
```
soundmonster, line 58: the key names here can probably be inferred from the models themselves; I'm not sure if this is something I should target in this PR, too?

integration_test/cases/repo.exs
```diff
@@ -152,6 +152,24 @@ defmodule Ecto.Integration.RepoTest do
     assert TestRepo.all(PostUserCompositePk) == []
   end
 
+  @tag :composite_pk
+  # TODO this needs a better name
+  test "insert, update and delete with associated composite pk #2" do
```
josevalim, line 157: I think we should remove the `associated` from the test before and then this one will be good. :)

lib/ecto/association.ex
```diff
@@ -35,7 +35,7 @@ defmodule Ecto.Association do
                required(:cardinality) => :one | :many,
                required(:relationship) => :parent | :child,
                required(:owner) => atom,
-               required(:owner_key) => atom,
+               required(:owner_key) => atom | list(atom),
```
josevalim, line 38: We should probably normalize it to always be a list? 🤔

lib/ecto/schema.ex
```diff
@@ -1959,7 +1959,15 @@ defmodule Ecto.Schema do
     end
 
     if Keyword.get(opts, :define_field, true) do
-      __field__(mod, opts[:foreign_key], foreign_key_type, opts)
+      foreign_keys = List.wrap(opts[:foreign_key])
+      foreign_key_types = if is_list(foreign_key_type) do
+        foreign_key_type
+      else
+        List.duplicate(foreign_key_type, length(foreign_keys))
```
josevalim, line 1966: We need a test for this branch. :)

lib/ecto/association.ex
josevalim: This will make sure we consider owner_key everywhere possible. Possibly the same change needs to be done for related_key!

josevalim:
Ping. If this is ready for another review, please let me know. :)

ttloves2 (approved):

soundmonster:
I'll have time to work on this next week. Thank you for the reminder!

soundmonster:
Two days of work later, the PR is somewhat further into the woods but quite some work is still needed around supporting composite keys for many-to-many associations. The most puzzling thing so far has been finding a good way to join tables on a variable list of columns. Tried and works (at least with Postgres):
* Join on just the first column, then add the rest in a where clause (e.g. with `Enum.reduce/3`). Postgres' planner doesn't care if the join condition is defined in the JOIN clause or in a WHERE clause, but I'm not sure if that's the case with other databases.
*  Use a recursive function together with `Ecto.Query.dynamic/2` to generate the `on:` option for the join. I'm not sure about the runtime overhead of doing this; it's also hard to generalize this because the bindings used by `dynamic` are often in different positions ("first and last" vs. "the last two"). However, it is much easier to understand what's happening and there's no "`hd(keys)` goes into the join, `tl(keys)` goes into a reduce with `where`" shenanigans.

I'm still very interested in getting this finished and would appreciate your thoughts on:
* is this going in the right direction, at all?
* what's the best way to handle a slow contributor? 😅 I'm committed to eventually finishing this up, but I'm not sure on timelines.

josevalim:
@soundmonster my suggestion is to not support composite foreign keys on many_to_many for now. What do you think? it will allow us to merge and review part of the work and you can take your own pace. :)

soundmonster:
Thank you for your reply, José. I've tried backtracking on ManyToMany, but the amount of changes needed to get the query planner and the HasThrough associations to work with both atoms and lists of atoms is more work than what I think is needed to complete support in ManyToMany. The current state of this branch passes all unit tests and only fails 6 integration tests (only checked with Postgres so far). I would have liked to split it into better reviewable chunks but I'm afraid that would be 2x more work 😅

braulio:
we look forward to completion!

soundmonster:
Status update:
* composite keys work for all associations 🥳 
* unit tests are green (on my machine, workflows need approval)
* integration tests are green with Postgres; I haven't yet gotten Earthly to run properly to test MySQL and MSSQL, maybe it's worth trying it out on CI?
* there's still quite a few TODOs for cleanup, esp. removing `List.wrap/1` everywhere
* docs are still open.

Where I need help:
* I'm using dynamic fragments to generate fields for association `join`s and `where`s. This doesn't only affect composite keys, but _all_ associations.
  * Is this expected to have a significantly negative performance impact?
  * Is there a better/other way to do this, apart from returning plain AST?
* It looks to me like all detailed association docs are in `association.ex` and there's no dedicated guide, but I might be missing something. Could you please point me to documentation / examples on associations _outside_ or `association.ex`?
* Probably workflow approval here and maybe for the [companion PR in ecto_sql](https://github.com/elixir-ecto/ecto_sql/pull/324).

isavita:
I am looking forward to completion as well!

josevalim:
That’s great! There are many ways you can help:

* review the code and give feedback
* run it locally. Do all tests pass too?
* use it in a project and give it a try! Is the behavior what you expect?

Looking forward to your thoughts!

-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

lib/ecto/changeset.ex
```diff
@@ -3071,6 +3071,12 @@ defmodule Ecto.Changeset do
     |> merge_keyword_keys(msg_func, changeset)
     |> merge_related_keys(changes, types, msg_func, &traverse_validations/2)
   end
+
+  defp atom_concat(atoms) do
+    atoms
+    |> Enum.map(&to_string/1)
```
josevalim, line 3077:
```suggestion
    |> Enum.map(&Atom.to_string/1)
```

lib/ecto/repo/preloader.ex
```diff
@@ -170,15 +170,23 @@ defmodule Ecto.Repo.Preloader do
         acc
       struct, {fetch_ids, loaded_ids, loaded_structs} ->
         assert_struct!(module, struct)
-        %{^owner_key => id, ^field => value} = struct
+        %{^field => value} = struct
+        ids = Enum.map(owner_key, fn key -> 
+          %{^key => id} = struct
+          id
+        end)
```
josevalim, line 177:
```suggestion
        ids = Enum.map(owner_key, &Map.fetch!(struct, &1))
```

lib/ecto/repo/preloader.ex
```diff
@@ -170,15 +170,23 @@ defmodule Ecto.Repo.Preloader do
...
         loaded? = Ecto.assoc_loaded?(value) and not force?
 
-        if loaded? and is_nil(id) and not Ecto.Changeset.Relation.empty?(assoc, value) do
+        if loaded? and Enum.any?(ids, &is_nil/1) and not Ecto.Changeset.Relation.empty?(assoc, value) do
+          {key_noun, key_verb_present, key_verb_past, key_list} = case owner_key do
+            [single_key] -> {"key", "is",  "was", single_key}
+            many_keys -> {"keys", "are", "were", "[#{Enum.join(many_keys, ", ")}]"}
+          end
```
josevalim, line 184: I think we can simplify it and leave only the plural form. Something like:

    its associations keys (`foo_id`) are nil.

lib/ecto/repo/preloader.ex
```diff
@@ -237,11 +245,20 @@ defmodule Ecto.Repo.Preloader do
   defp fetched_records_to_tuple_ids([], _assoc, _related_key),
     do: []
 
-  defp fetched_records_to_tuple_ids([%{} | _] = entries, _assoc, {0, key}),
-    do: Enum.map(entries, &{Map.fetch!(&1, key), &1})
+  defp fetched_records_to_tuple_ids([%{} | _] = entries, _assoc, {0, [single_key]}) do
+    Enum.map(entries, &{[Map.fetch!(&1, single_key)], &1})
+  end
```
josevalim, line 250: I think we can get rid of this clause?

lib/ecto/schema.ex
```diff
@@ -1999,15 +1999,25 @@ defmodule Ecto.Schema do
   def __belongs_to__(mod, name, queryable, opts) do
     check_options!(opts, @valid_belongs_to_options, "belongs_to/3")
 
-    opts = Keyword.put_new(opts, :foreign_key, :"#{name}_id")
+    opts = Keyword.update(opts, :foreign_key, :"#{name}_id", &List.wrap/1)
```
josevalim, line 2002:
```suggestion
    opts = Keyword.update(opts, :foreign_key, [:"#{name}_id"], &List.wrap/1)
```

The function is not called if there is no value. The fact the above is wrong and there is no error makes me think the foreign key is always set? If so, use `Keyword.update!` instead and see what breaks (if anything at all).

lib/ecto/schema.ex
```diff
@@ -1999,15 +1999,25 @@ defmodule Ecto.Schema do
...
     foreign_key_type = opts[:type] || Module.get_attribute(mod, :foreign_key_type)
 
-    if name == Keyword.get(opts, :foreign_key) do
+    if [name] == Keyword.get(opts, :foreign_key) do
       raise ArgumentError, "foreign_key #{inspect name} must be distinct from corresponding association name"
     end
 
     if Keyword.get(opts, :define_field, true) do
-      __field__(mod, opts[:foreign_key], foreign_key_type, opts)
+      foreign_keys = List.wrap(opts[:foreign_key])
```
josevalim, line 2010: Ah, this is why the line above works... yeah, we should clean up a bit.

lib/ecto/schema.ex
```diff
@@ -1999,15 +1999,25 @@ defmodule Ecto.Schema do
   def __belongs_to__(mod, name, queryable, opts) do
     check_options!(opts, @valid_belongs_to_options, "belongs_to/3")
 
-    opts = Keyword.put_new(opts, :foreign_key, :"#{name}_id")
+    opts = Keyword.update(opts, :foreign_key, :"#{name}_id", &List.wrap/1)
     foreign_key_type = opts[:type] || Module.get_attribute(mod, :foreign_key_type)
 
-    if name == Keyword.get(opts, :foreign_key) do
+    if [name] == Keyword.get(opts, :foreign_key) do
```
josevalim, line 2005: Maybe `name in Keyword.get(opts, :foreign_key)`?

lib/ecto/association.ex
```diff
@@ -261,6 +272,65 @@ defmodule Ecto.Association do
...
+  def transpose_values(values) do
+    values
+    |> Enum.zip()
+    |> Enum.map(&Tuple.to_list/1)
+  end
+
+  def strict_zip([], []), do: []
+  def strict_zip([_ | _], []), do: raise ArgumentError, "lists should be of equal length"
+  def strict_zip([], [_ | _]), do: raise ArgumentError, "lists should be of equal length"
+  def strict_zip([l | ls], [r | rs]), do: [{l, r} | strict_zip(ls, rs)]
```
josevalim, line 285:
```suggestion
  def strict_zip([l | ls], [r | rs]), do: [{l, r} | strict_zip(ls, rs)]
  def strict_zip([], []), do: []
  def strict_zip(_, _), do: raise ArgumentError, "lists should be of equal length"
```

lib/ecto/association.ex
```diff
@@ -261,6 +272,65 @@ defmodule Ecto.Association do
     combine_assoc_query(query, source.where || [])
   end
 
+  def transpose_values([[_]] = values), do: values
```
josevalim, line 275: We should `@doc false` this and the functions below.

lib/ecto/association.ex
```diff
@@ -247,12 +250,20 @@ defmodule Ecto.Association do
...
+        |> Enum.reduce(query, fn {dest_out_key_field, value}, query ->
+          query
+          |> where([{dest, final_bind}], field(dest, ^dest_out_key_field) == ^value)
+        end)
 
       {nil, values} ->
-        query
-        |> where([{dest, final_bind}], field(dest, ^dest_out_key) in ^values)
+        dest_out_key
+        |> Enum.zip(transpose_values(values))
```
josevalim, line 262: Can you please add a code comment explaining why we need `transpose_values`? TY!

lib/ecto/association.ex
```diff
@@ -235,8 +236,10 @@ defmodule Ecto.Association do
     # for the final WHERE clause with values.
     {_, query, _, dest_out_key} = Enum.reduce(joins, {source, query, counter, source.out_key}, fn curr_rel, {prev_rel, query, counter, _} ->
       related_queryable = curr_rel.schema
-
-      next = join(query, :inner, [{src, counter}], dest in ^related_queryable, on: field(src, ^prev_rel.out_key) == field(dest, ^curr_rel.in_key))
+      next = query
+        # join on the foreign key
+        |> join(:inner, [{src, counter}], dest in ^related_queryable, on: ^on_fields(prev_rel.out_key, curr_rel.in_key))
```
josevalim, line 241: Unfortunately I don't think this is correct. The `src` is not always going to be the one immediately before the `dest`. You need to pass the `counter` to `on_fields` in order to build the proper source. Using `dynamic` is not ideal either. We will need to find ways to optimize it but that can be a later concern.

gosseti:
Just to say I am hugely looking forward to this feature and very happy to test it out!

lib/ecto/association.ex
soundmonster: the function was bogus, actually. I've managed to remove it 😅 thank you for pointing me at it.

soundmonster:
Thank you very much for the review! It's been super helpful!

A quick update:
* CI flagged that the pre-2022 state of this brunch didn't work on MySQL as neither the DB engine nor the driver supports the `array` data type. I've put in some work into making it run on both MySQL _and_ Postgres, at least to the point where all integration and unit tests pass, including the new ones written for this specific feature. I'll update on MSSQL when I have it running, too.
* I have produced quite a heap of spaghetti while trying to get everything to work. I'll have to refactor large parts of `association.ex` with these goals in mind:
  * address the issue with binding positions being incorrect
  * either get rid of `dynamic` by generating AST directly, or at least not call `dynamic` for the "normal" case of single-field associations. 
* I'll also go over the other places I've changed and try to minimize all the "wrap/unwrap" dances I've produced.

Hence explicitly not asking for a review at this point. This is very much a work in progress still.

soundmonster:
Another quick update: I've finally managed to run the whole test matrix with Earthly locally and it is passiing all tests so far 😌

gosseti:
I’ve recently been testing this out with some great results so far!

Some things I noticed when introducing this branch as a dependency in my project:

- `owner_key` and `related_key` on `Ecto.Association.BelongsTo`, `Ecto.Association.Has` are always lists 
  -- it looks like in the current ecto version they are atoms
  -- this caused a regression in absinthe’s `dataloader`, which expects atoms

Was the decision to always force `owner_key` and `related_key` to be lists intentional? 

My thinking is that if using default-style (single) primary keys:

- they should keep the existing behaviour (exposed as atoms), 
- and then when it’s a composite primary key it uses lists

Great work on this, it’s been a huge unlock and let me know how I can help to test things out more.

lib/ecto/association.ex
gosseti: Worth noting that this changing to a list caused a regression in absinthe’s dataloader in my project. I’ve [set up an initial PR](https://github.com/absinthe-graphql/dataloader/pull/138) there to address it for when this feature lands

v0idpwn: __probably__ should be kept as `atom() | list(atom)`, then? To avoid breaking compatibility?

soundmonster:
I've managed to get rid of `dynamic` for constructing association queries. Will work on docs next. As always, testing is very welcome!

MarthinL:
> I've managed to get rid of `dynamic` for constructing association queries. Will work on docs next. As always, testing is very welcome!

Hey @soundmonster, yeah, me too. I've arrived at a point in my project where I have a very clear use-case for a  composite primary key in all several associations and I'd love to help get your PR code tested so it can get marged into the main branch. 

Unfortunately, I've extremely limited bandwidth and have never even built any version of erlang, OTP, Elixir, Phoenix or such on my in my own environment, let alone dabbled with alternative versions. I'm willing to help and offer my user-case as real-world validation of the need and implementation. In short, please talk me through what I'd need to do to get your version on my server so I can put it to the test in a way that helps you and your case get this important adaptation into the main-stream  code. 

I'm close to taking a first version of my new service live and I could conceivably do that without composite keys since I'd be running it in a single region whereas my use-case comes into play where I have parts of the data replicating glocally using the database itself and parts of the data remaining mastered in a specific region unless it becomes cached as such in one or more other regions. If possile I'd love to build the region login into the first release already to save on migration implications down the line so my first objective is to get a feel for how close to "production quality" this PR is in terms of time and otherwise.

greg-rychlewski:
Hi @MarthinL,

To test his branch you can point your Ecto and Ecto SQL dependencies at his branches in `mix.exs` and then update your deps:

> {:ecto, git: "https://github.com/soundmonster/ecto.git", branch: "composite_foreign_keys"}
> {:ecto_sql, git: "https://github.com/soundmonster/ecto_sql.git", branch: "composite_foreign_keys"}

As a general comment, I'm not sure you would need to block features in your app because of this. You could always have a unique id column that's used for the association.

MarthinL:
> Hi @MarthinL,
> 
> To test his branch you can point your Ecto and Ecto SQL dependencies at his branches in `mix.exs` and then update your deps:
> 
> > {:ecto, git: "https://github.com/soundmonster/ecto.git", branch: "composite_foreign_keys"}
> > {:ecto_sql, git: "https://github.com/soundmonster/ecto_sql.git", branch: "composite_foreign_keys"}
> 
> As a general comment, I'm not sure you would need to block features in your app because of this. You could always have a unique id column that's used for the association.

Thanks, I was unaware that the deps format in mix.exs allowed for such syntax

However, when I try to use your two lines as is, I run into a series of problems starting with the inconvenient fact defp deps do in mix.exs has 
      {:phoenix_ecto, "~> 4.4"},
      {:ecto_sql, "~> 3.6"},

i.e. :ecto isn't a direct dependency but brought in by :ecto_sql but :phoenix_ecto is a direct dependency that is incompatible with the :ecto direct dependency.

If I substitute your two lines for the two I've listed above, the result of running mix deps.get is:

Dependencies have diverged:
* ecto (https://github.com/soundmonster/ecto.git - origin/composite_foreign_keys)
  the dependency ecto in mix.exs is overriding a child dependency:

  > In mix.exs:
    {:ecto, [env: :prod, git: "https://github.com/soundmonster/ecto.git", branch: "composite_foreign_keys"]}

  > In deps/ecto_sql/mix.exs:
    {:ecto, [env: :prod, path: "../ecto"]}

  Ensure they match or specify one of the above in your deps and set "override: true"
** (Mix) Can't continue due to errors on dependencies

If I only substitute the :ecto_sql lines mix deps.get reports:

Dependencies have diverged:
* ecto (../ecto)
  different specs were given for the ecto app:

  > In deps/ecto_sql/mix.exs:
    {:ecto, [env: :prod, path: "../ecto"]}

  > In deps/phoenix_ecto/mix.exs:
    {:ecto, "~> 3.3", [env: :prod, hex: "ecto", repo: "hexpm", optional: false]}

  Ensure they match or specify one of the above in your deps and set "override: true"
** (Mix) Can't continue due to errors on dependencies

LIke I stated before, I'm quite unfamiliar with working with anything but the latest and greatest released versions of anything, so I'm afraid I'm going to need a little more help to get this done. Thank you.


As for your suggestion of a unique id for the association, and provided I understand you correctly, I have considered that, but the benefit of having region_id as an explicit part of the composite key is significant as it allows any region encountering a reference to such a piece of data to determine which region to fetch it from. If I used a UUID the region would need to send each request for data not known locally off to a central directory (which would nullify distributed data benefits) to determine the region or each region would need a pre-emptive local copy of at least a record that maps the UUID to a region. Since the data items are small, that would mean having all data everywhere just like it would be if the database was doing to replication.

greg-rychlewski:
You can still have foreign keys while maintaining a unique ID for the association itself. 

For your deps issue it looks like it's asking you to put `override: true` so it overrides those nested deps.

MarthinL:
> You can still have foreign keys while maintaining a unique ID for the association itself.
> 
I suspect we're talking cross-purposes with regards to my use case. If it is of particular interest to you and/or would be valuable to others, I can lay out the use case involging a mixture of database replication based gloval data and regional data stored also in database though not replicated by the database engine but "manually" in the distributed server which requests copies directly from the region where the data originates. The bottom line is that I have a number of relationships, including a self-referencing many-to-many association where the association entity has additional fields in my project. Given that the entity involved in the many-to-many is also the one with the composite primary key, it should be quite a stress test for @soundmonster's code.

greg-rychlewski:
I'm sure your use case would benefit from this a lot and would be a good test. I'd just hate to see your app be limited if this is not merged anytime soon. I don't believe there is any situation that can't be solved by introducing a unique field in addition to the primary keys you currently have. But I'll leave it up to you what you think is the best way forward.

MarthinL:
> For your deps issue it looks like it's asking you to put `override: true` so it overrides those nested deps.

I solved it another way:

I cloned phoenix_ecto from the main branch, ecto_sql and ecto from soundmonster's pr and then used the path: "/local/dir/to/code" deps syntax to draw in the code I needed which meant that I could update the mix.exs in the now local phoenix_ecto directory to refer to the right version of ecto using the technique involving the environment variable ECTO_PATH I found in ecto_sql/mix.exs. I don't know how well this would play if I tried building a release with it, but it gets the job done. 

For reference: 
In my ~/.zshrc I have
export ECTO_PATH=/Users/marthin/projects/pr-workdir/ecto

In my (test project called Bennie)'s mix.exs i have 
defp deps do 
  [ 
      {:phoenix, "~> 1.6.15"},
      {:phoenix_ecto, path: "/Users/marthin/projects/pr-workdir/phoenix_ecto"},
      {:ecto_sql, path: "/Users/marthin/projects/pr-workdir/ecto_sql"},
...

and in .../phoenix_ecto/mix.exs I have:
 defp deps do
    [
      {:phoenix_html, "~> 2.14.2 or ~> 3.0", optional: true},
      ecto_dep(),
      # {:ecto, "~> 3.3"},  
      ....
    ]
  end
  defp ecto_dep do
    if path = System.get_env("ECTO_PATH") do
      {:ecto, path: path}
    else
      {:ecto, "~> 3.8.0-dev", github: "elixir-ecto/ecto"}
    end
  end

which has been copied verbatim from .../ecto_sql/mix.exs

mix deps.get and mix deps.compile runs without issues and in the one test so far I've cooked up the PR seems to work as expected. I'd be nice to chat with Leo B, Jose Valim or anyone involved to know more definitively what how the PR is intended to get activated. I'll write another comment asking more directly to clarify my understanding.

MarthinL:
Hi @soundmonster and @josevalim,

So far so good, I have set up this PR's code on my machine and it seems to be running as expected. I'm not sure I fully understand under what conditions it comes into play though. Is there a specific syntax or set of options I need to specify somewhere in the migration and/or the schema where I define the primary keys and the associations? If so, what are those?

My current understanding is as follows:
a) In the migration file I need to ensure that all the fields comprising the composite primary key is present.  For my case, where i have an items table with the usual autonumbering id field as well as a region_id field which is in fact a foreign key to another table, I'd have this in the migration
    create table(:items) do
      add :region_id, :integer, [primary_key: true, references: "regions"]
      add :name, :string

      timestamps()
    end

b) In the schema file, something similar is happening: Again the usual id field is kept as is and the region_id is added to that as in:
  schema "items" do
    field :region_id, :integer, [primary_key: true]
    field :name, :string

    belongs_to :region, Region, [define_field: false]
    timestamps()
  end

c) With the above in place, Item now has a composite key. When I now introduce a third schema Note (table notes) I need the migration to contain a region_id as well as an item_id field
  def change do
    create table(:notes) do
      add :line, :string
      add :region_id, :integer
      add :item_id, references("items", with: [region_id: :region_id])
     timestamps()
    end

and I need the schema to read:
  schema "notes" do
    field :line, :string
    belongs_to :region, Region, [primary_key: true]
    belongs_to :item, Item
    timestamps()
  end

It works as expected, but is gets a little fuzzy there. Do I need the belongs_to :region, Region, [primary_key: true] in schema "notes" or is it optional? 

It is my current understanding that the belongs_to :item, Item line in Note has the effect of adding the fields item_id (derived from the association name) as well as region_id because region_id is part of Item's primary key. Is that an accurate understanding, is the region_id field present as a result of the belongs_to :region, Region line or do I have the whole thing backwards?

Once I've have the belongs_to association sorted, I want to move onto the other associations and build my way up towards the self-referencing many-to-many with added fields for the associative entity (i.e. where the many-to-many association is read-only and one uses two has-many/belongs_to pairs to work with the associative entity's content) . 

Perhaps my little journey of discovery can help the efforts to write documentation for this feature. I suspect it is going to get a lot more hairy with options to set foreign keys and source fields as the use case ramps up.

MarthinL:
> I'm sure your use case would benefit from this a lot and would be a good test. I'd just hate to see your app be limited if this is not merged anytime soon. I don't believe there is any situation that can't be solved by introducing a unique field in addition to the primary keys you currently have. But I'll leave it up to you what you think is the best way forward.

Yes, adding a unique field in addition to the existing primary key is exactly what I am doing. (I thought you wanted to replace the existing id plus the unique field with a single globally unique field, which I don't thing would work for the reasons I've explained). The only thing is that I make the unique additional field a little more meaningful by having it as a foreign key to a regions table where I can store pertinetn information about the region such as the ip addresses and/or dns names where I the region's backend servers can be reached. The region table would form part of what the database replicates to all the distributed nodes.

MarthinL:
> I'm sure your use case would benefit from this a lot and would be a good test. I'd just hate to see your app be limited if this is not merged anytime soon. I don't believe there is any situation that can't be solved by introducing a unique field in addition to the primary keys you currently have. But I'll leave it up to you what you think is the best way forward.

On that: It is my understanding that it isn't a new feature to have additional fields as part of primary keys. The purpose and essence of @soundmonster's PR is to have associations honour those composite primary keys more or less transparently. Is that a fair assessment?

jadengis:
> On that: It is my understanding that it isn't a new feature to have additional fields as part of primary keys. The purpose and essence of @soundmonster's PR is to have associations honour those composite primary keys more or less transparently. Is that a fair assessment?

I'm using composite primary keys in an app currently for a multi-tenatnt app and the whole `ecto-sql` portion of it already works. You can prepare a database with composite primary keys and composite foreign keys today, and it works as you would expect. I believe this PR is just for teaching the association functionality built into `ecto` to understand those composite keys so it can join and preload associations using the correct primary key (instead of just one of the fields).

MarthinL:
### I'm stuck on syntax and semantics and need help.

My little exploratory project came along just fine as long as I stick to simple `belongs_to :owner, Owner` and `has_many :children, Child` relationships, i.e. where the field names derived from the association name and the (composite) primary key fields defined on the Child schema.

The relevant parts of my test code is as follows:

from the migration files
```
    create table(:items) do
      add :region_id, :integer, [primary_key: true, references: :regions]
      add :name, :string
    end
    create index(:items, [:region_id])

    create table(:items) do
      add :region_id, :integer, [primary_key: true, references: :regions]
      add :name, :string
    end
    create index(:items, [:region_id])

    create table(:notes) do
      add :line, :string
      add :region_id, :integer
      add :item_id, references("items", with: [region_id: :region_id])
    end
```

And from the schema files:
```
Region:
  schema "regions" do
    field :label, :string

    has_many :items, Item
  end

Item:
  schema "items" do
    field :region_id, :integer, primary_key: true
    field :name, :string

    belongs_to :region, Region, [define_field: false]
    has_many :notes, Note
  end

Note:
  schema "notes" do
    field :line, :string

    belongs_to :region, Region, [primary_key: true]
    belongs_to :item, Item
  end

```

I've put all of that through its paces and it all seems to work well.

But I get stuck when I need to start using options of the associations such as `foreign_key:` and `references:` to override the default field names. It might be required in other situations as well, but in my specific use case the situation where I have to refer to different field names as foreign keys is when I try to implement a self-referencing many-to-many relationship such as I have in my read system. Now, because my associative entity has attributes of its own, the many_to_many association I use is a :through association (and thus read-only) and I hardly ever actually refer to it, so if it needs to go it's no big deal.

What I use instead is to create two has_many relationships in the schema I aim to have the self-referencing many-to-many behaviour for with two (now sets of) foreign key fields and in the associative entity I define two belongs_to associations referring to those same fields. So far in my real project that has left me with very functional relationships that responds well to all the preload and build_assoc calls I throw at it. 

Having now introduced the composite primary key into the mix, I cannot seem to figure out how to modify the foreign_key: options I've used before so they would refer to both primary key fields and two differently named fields in the associative entity.

Let's call the associative entity Relship. Before declaring Item's region_id as part of its primary key, Relship's code looked like this:
```
    create table(:relships) do
      add :label, :string
      add :parent_item_id, references(:items)
      add :child_item_id, references(:items)
    end

    create index(:relships, [:parent_item_id])
    create index(:relships, [:child_item_id])

```

and the schema like this:
```
  schema "relships" do
    field :label, :string
    field :parent_item_id, :id
    field :child_item_id, :id

    belongs_to :parent_item,
               Item,
               foreign_key: :parent_item_id,
               define_field: false

    belongs_to :child_item,
               Item,
               foreign_key: :child_item_id,
               define_field: false
  end
```
which (i.e. the above code for Relship) is obviously broken as it would only work for the one default region because the relships themselves have id's that doesn't include region_id and the foreign keys in Relship doesn't store each Item's region_id. That's the problem I set out to try solve before I go live to avoid the hassle of having to migrate the database and code before I can introduce a second region.

So, adding region support in Relship, I get this far:

In migration:
```
    create table(:relships) do
      add :region_id, :integer, [references: :regions, primary_key: true]
      add :label, :string
      add :parent_item_region_id, :integer, [references: :regions]
      add :parent_item_id, references(:items, with: [parent_item_region_id: :region_id])
      add :child_item_region_id, :integer, [references: :regions]
      add :child_item_id, references(:items, with: [child_item_region_id: :region_id])
    end

    create index(:relships, [:parent_item_region_id, :parent_item_id])
    create index(:relships, [:child_item_region_id, :child_item_id])

```

and the schema like this:
```
Relship:
  schema "relships" do
    field :region_id, :integer, [primary_key: true]
    field :label, :string
    field :parent_item_region_id, :integer
    field :parent_item_id, :id
    field :child_item_region_id, :integer
    field :child_item_id, :id

    belongs_to :parent_item,
               Item,
               foreign_key: [parent_item_region_id: :region_id, parent_item_id: :id],
               define_field: false

    belongs_to :child_item,
               Item,
               foreign_key: [child_item_region_id: :region_id, child_item_id: :id],
               define_field: false
  end

Item: (added to schema)

    has_many :child_rels,
              Relship,
              foreign_key: [parent_item_region_id: :region_id, parent_item_id: :id]

    has_many :parent_rels,
              Relship,
              foreign_key: [child_item_region_id: :region_id, child_item_id: :id]

```
which is *obviously* invalid syntax and the compiler tells me so, loudly. But I haven't been able to figure out what the approriate syntax would be, or if it's more a matter of semantics than syntax, how I would get ecto to refer to the correct sets of primary and foreign key fields.

**Please, anyone, what is the intended way to specify foreign keys in associations when using composite primary keys?**

MarthinL:
> Hi @soundmonster and @josevalim,
> 
> So far so good,
...
**except, not really.**

When I couldn't seem to get the more complex variants of associations working I went back to trace through the simpler ones I reported on earlier. I know I'm running the PR code - my IO.inspect statements I insert in the code in my local copy of the code cloned from soundmonster's repository tells me so.

I have a schema called Notes, which has been set up with an additional field as its primary key called region_id. Each Note belongs_to a Region, and it works correct in that I can have the Items with the same id in different regions in my data and it associates 100% correctly. What threw me off at first was that when I generate those items on the same machine using the same database sequence to autogenerate the id field for Item, I naturally got unique id values. That mean that the association with the next schema, Notes, defined in the Item schema as has_many :notes, Notes, would relate the notes of each item based on the item id which was unique through the system. 

But as soon as I changed the ids of the Items so the id parts were duplicates and the primary keys only made unique by the region_id, the association code broke and returned the same notes for all Items with the same id regardless of the region_id. 

My assumption is that I am still doing something wrong here, but I cannot figure out what that would be. There is also a chance that I do have the correct version of the code and correct association definitions, but that the code simply doesn't take the whole primary key into account when resolving an association. 

Can we please figure this out? I'm very keen on using associations with composite key schemas in my project. 

Just a note about my use-case which differs significatly from a multi-tenancy use-case in that with multi-tenancy one would expect a single tenant-id to identify all the data of a tenant with no chance for cross-tenant references. My use case is closer to data partitioning than multi-tenancy, and I strongly rely on references which cross partition boundaries. Each partition, which I call a region, has its own database with own sequence from which id values are generated using nextval (PostgreSQL) and a "home" region configured in the runtime.exs for that region. Only a portion, mostly administrative data, of the database is replicated from a master region to the others at the database level. Beyond that, each region operates independently and requests data from other regions when such reference is encoutered. this is merely context, and not part of the problem I am facing.

MarthinL:
> > Hi @soundmonster and @josevalim,
> > So far so good,
> > ...
> > **except, not really.**
> 
> When I couldn't seem to get the more complex variants of associations working I went back to trace through the simpler ones I reported on earlier. I know I'm running the PR code - my IO.inspect statements I insert in the code in my local copy of the code cloned from soundmonster's repository tells me so.
> 
> I have a schema called Items, which has been set up with an additional field as its primary key called region_id. Each Item belongs_to a Region, and it works correct in that I can have the Items with the same id in different regions in my data and it associates 100% correctly. What threw me off at first was that when I generate those items on the same machine using the same database sequence to autogenerate the id field for Item, I naturally got unique id values. That mean that the association with the next schema, Notes, defined in the Item schema as` has_many :notes, Notes`, would relate the notes of each item based on the item id which was unique through the system.
> 
> But as soon as I changed the ids of the Items so the id parts were duplicates and the primary keys only made unique by the region_id, the association code broke and returned the same notes for all Items with the same id regardless of the region_id.
> 
> My assumption is that I am still doing something wrong here, but I cannot figure out what that would be. There is also a chance that I do have the correct version of the code and correct association definitions, but that the code simply doesn't take the whole primary key into account when resolving an association.
> 
> Can we please figure this out? I'm very keen on using associations with composite key schemas in my project.
> 
> Just a note about my use-case which differs significatly from a multi-tenancy use-case in that with multi-tenancy one would expect a single tenant-id to identify all the data of a tenant with no chance for cross-tenant references. My use case is closer to data partitioning than multi-tenancy, and I strongly rely on references which cross partition boundaries. Each partition, which I call a region, has its own database with own sequence from which id values are generated using nextval (PostgreSQL) and a "home" region configured in the runtime.exs for that region. Only a portion, mostly administrative data, of the database is replicated from a master region to the others at the database level. Beyond that, each region operates independently and requests data from other regions when such reference is encoutered. this is merely context, and not part of the problem I am facing.

**False Alarm** the code on my local machine was from soundmonster's repository, but from its original master branch instead of the composite_foreign_keys branch. I fixed that by running
`gh pr checkout 3638` in my local ecto directory
`gh pr checkout 324` in in my local ecto_sql directory

As a result, both `:foreign_keys` and `:references` opts for `has_many` and `belongs_to` now accepts a list of atoms where before it only accepted a single atom, meaning I the following lines of code now compiles and executes:

in Region schema:
```
    has_many :items, Item, foreign_key: :region_id, references: :id
```
in iIem schema:
```
    has_many :notes, Note, foreign_key: [:item_region_id, :item_id], references: [:region_id, :id]
```

in Note schema:
```
    belongs_to :item, Item, foreign_key: [:item_region_id, :item_id], references: [:region_id, :id], type: [:integer, :id]
```

I should now be able to tackle the self-referencing many_to_many :through associateion with the join_schema.

MarthinL:
Got tripped up by `Ecto.build_assoc` failing for regular as well as composite keys because the related_key and owner_key fields were arriving as lists but getting matched as atoms.

The trouble seems to be in association.ex in the Has module's build function around line 863. I made the change indicated below and it seems to have done the trick. The additional List.wrap calls are there becauswe I can't tell if the two parameters would reliably arrive as lists or not, but I gather that having wrap and rewrap and unwrap the keyss is part of why the PR hasn't been accepted yet. Anyhow, I leave it to the experts to tell if my change is acceptable or not.

association.ex from
 ```
   def build(%{owner_key: owner_key, related_key: related_key} = refl, owner, attributes) do
      data = refl |> build(owner) |> struct(attributes)
      %{data | related_key => Map.get(owner, owner_key)}
   end
```
to
```
  def build(%{owner_key: owner_key, related_key: related_key} = refl, owner, attributes) do
    data = refl |> build(owner) |> struct(attributes)
    Enum.zip(List.wrap(owner_key), List.wrap(related_key))
    |> Enum.reduce(data, fn {owner_key, related_key}, acc -> 
      %{acc | related_key => Map.get(owner, owner_key)} 
    end)
  end
```

Like I said before, this deep into the technical details I'm quite far outside my comfort zone. I'm more than willing to push this change into the branch if but someone will have to guide me as I don't want to mess things up or make enemies.

**Comment by soundmonster** [comment]
I can confirm that `Ecto.build_assoc` misbehaves now. Another thing that I ran into when trying out the branch on an existing project is that ExMachina depends on `related_key` being an atom, not a list. It has been mentioned that Absinthe breaks when related key is a list, too. `owner_key` and `related_key` are documented fields that should be considered part of the public Ecto API, so IMHO to support composite keys we would either have to:
* only use lists when there is more than one key, only breaking other libraries when the new feature is used. Existing codebases would not be affected but internal handling would be more complex
* always use lists (as this PR currently does) and mark this as a major release.

@josevalim you have previously gravitated towards using lists everywhere, but given the above - I would highly appreciate to hear your opinion on this.

**Comment by josevalim** [comment]
> only use lists when there is more than one key, only breaking other libraries when the new feature is used. Existing codebases would not be affected but internal handling would be more complex

:+1:

**Comment by soundmonster** [comment]
Current status:

* updated the branch to only use lists for owner/related keys if they're more than one, preserving backwards compatibility to packages like Absinthe and ExMachina.
* all tests are green in both `ecto` and `ecto-sql` 🥳 
* updated docs for `schema.ex` to include examples of composite foreign keys in associations


Trying out the branch on an in-house project surfaced a problem in an `after_compile_validation` hook; I've checked if I can reproduce the bug in tests but it appears that we currently have no way to test post-compile hooks at all, due to the way the compiler works in tests. Is this correct? (I'll fix the bug in the hook in the next commit, of course)

**Comment by michallepicki** [comment]
Hi @soundmonster ! To make the status clear, are you planning to add more changes or is this PR (and https://github.com/elixir-ecto/ecto_sql/pull/324 ) waiting for review now?

**Comment by soundmonster** [comment]
Thank you for the ping! Yes, I'm done with both and am looking for a review!

**Comment by MarthinL** [comment]
Hey @soundmonster,

A few weeks ago, just before your latest changes, I tried to compile my code that uses composite key associations with your active branch of that time but my schema code would not compile. This is probably because the syntax I use for specifying the association isn't what your code expects. The changes you've published since does not seem to impact this, so can I please trouble you to tell me the syntax your code is designed to work with. As an example, I've been using this syntax (but had to make changes to your code for it to work):

has_many :child_ties, Tie, foreign_key: [:parent_region_id, :parent_id], references: [:region_id, :id]
has_many: parent_ties, Tie, foreign_key: [:child_region_id, :child_id], references: [:region_id, :id]
 
but it failed to compile and with no clear indication of where the problem is.

Because I use multiple associations with the same schema I cannot use default field names but have to specify the foreign_key field names and what they refer to.

**Comment by soundmonster** [comment]
@MarthinL could you please post the error message you're getting when compilation fails? Also, it would be very helpful to see the full schema definition. Several things a can think of:
* are one or more of the fields `:parent_region_id`, `:parent_id`, `:child_region_id` and `:child_id` already defined on the schema? If yes, please try the `define_field: false` flag
* the colon on the second line looks off, it should be `has_many :parent_ties` instead of `has_many: parent_ties`
* a `has_many` (1:n) typically corresponds to a `belongs_to` (n:1) on the other end of the relation; it's possible to have a self-referencing hierarchy defined as in [this example](https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3-examples) from the Ecto docs:
```elixir
defmodule Comment do
  use Ecto.Schema
  schema "comments" do
    field :content, :string
    field :parent_id, :integer
    belongs_to :parent, Comment, foreign_key: :parent_id, references: :id, define_field: false
    has_many :children, Comment, foreign_key: :parent_id, references: :id
  end
end
```
Your example is different because it looks like a tie can have more than one parent, which would likely require a `many_to_many`.

MarthinL:
> @MarthinL could you please post the error message you're getting when compilation fails? 
It's not as easy for me to replicate the error message right now because I'm having dependency issues when I try to create and run a new project with referencing your PR in mix.exs. (If I don't set override: true then it mix complains and if I do, then an unknown function Phoenix.Ecto.CheckRepoStatus.init/1 is called). I'm more than willing to privide the code for a simplified test case if you can assist me with getting a project set up on my computer that uses the current Phoenix environment as well as your PR code for Ecto.

^ Also, it would be very helpful to see the full schema definition. 
I'll give you a all the schemas involved in a representative test case, but the real schemas are overly complex and would detract from the simple issue at hand. 

Several things a can think of:
> 
> * are one or more of the fields `:parent_region_id`, `:parent_id`, `:child_region_id` and `:child_id` already defined on the schema? If yes, please try the `define_field: false` flag
I'm fully aware of that issue and how to manage it, so no, that is not what goes wrong.

> * the colon on the second line looks off, it should be `has_many :parent_ties` instead of `has_many: parent_ties`
My bad, I retyped instead of copying and pasting and made a typo. 

> * a `has_many` (1:n) typically corresponds to a `belongs_to` (n:1) on the other end of the relation; it's possible to have a self-referencing hierarchy defined as in [this example](https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3-examples) from the Ecto docs:
I'm very familiar with the standard examples and syntax and have been using all types of associations extensively in my project, but I have no examples or reference material that shows or defines the syntax when referenced table has a composite key.  That's what I've asked for. 

> 
> ```elixir
> defmodule Comment do
>   use Ecto.Schema
>   schema "comments" do
>     field :content, :string
>     field :parent_id, :integer
>     belongs_to :parent, Comment, foreign_key: :parent_id, references: :id, define_field: false
>     has_many :children, Comment, foreign_key: :parent_id, references: :id
>   end
> end
> ```
> 
> Your example is different because it looks like a tie can have more than one parent, which would likely require a `many_to_many`.
At its heart, the associations in my project have a lot in common with many-to-many relationships, but Ecto's many-to-many implementation is clumsy when it comes to storing and accessing additional data fields in associative entities. Consequently I specify, create and access data using sets of has-many and belongs-to associations instead and have removed the many-to-many :through associations I used to keep form the schemas as they added no value and in fact started getting in the way. 

Thanks for responding to my message. I'm very grateful that you've tackled this issue and have nothing but respect for your efforts. My project would suffer if I was forced to postpone implementation support for multiple regions till afterthe initial  launch. Multiple regions is my sole use case for composite keys. It would be best if this feature becomes part of the main branch before I launch, but even if that doesn't happen, I'd want to avoid using the code I've adapted myself to support composite keys and use your branch directly. I'm willing to spend time and energy to get as close to a standard solution as I can, but I don't deal with the domain of dependencies and git branches regularly so it becomes very hard to participate in the manner you might be expecting. Please forgive me, but if you can help me create or recreate the environment in which your code works, I will provide code that represents my use case for you to correct the syntax (and semantics) for.

soundmonster:
> If I don't set override: true then it mix complains and if I do, then an unknown function Phoenix.Ecto.CheckRepoStatus.init/1 is called

`Phoenix.Ecto.CheckRepoStatus` is [a module](https://github.com/phoenixframework/phoenix_ecto/blob/v4.4.1/lib/phoenix_ecto/check_repo_status.ex#L1) from `:phoenix_ecto`. Something happened that removed that package from your build. In one of your previous comments you've been running a patched local copy of `phoenix_ecto`, that might be the problem. All you actually have to do to try out this build is to:
* upgrade to the absolute latest _regular_ releases of `ecto` and `ecto_sql`
* switch to this branch by adding the following two lines into your `mix.exs`:
```elixir
      {:ecto, github: "soundmonster/ecto", branch: "composite_foreign_keys", override: true},
      {:ecto_sql, github: "soundmonster/ecto_sql", branch: "composite_foreign_keys", override: true},
```

If you have an umbrella app, add it to the "root" mixfile. If you already have mentions of `ecto` and `ecto_sql` in your main mixfile, comment it out for now. There is no need to check out anything locally or to patch other packages like `phoenix_ecto`.

> > Your example is different because it looks like a tie can have more than one parent, which would likely require a `many_to_many`.
> 
> At its heart, the associations in my project have a lot in common with many-to-many relationships, but Ecto's many-to-many implementation is clumsy when it comes to storing and accessing additional data fields in associative entities. Consequently I specify, create and access data using sets of has-many and belongs-to associations instead and have removed the many-to-many :through associations I used to keep form the schemas as they added no value and in fact started getting in the way.

I'm confused now 😅 I don't see a way to have a many-to-many without an association entity in a relational database; a Tie can only point to _one_ parent or child, so one end of the relation will remain unary. The syntax (apart from the colon) seems fine for a `has_many`, but I don't get the semantics TBH.

soundmonster:
Given what you've just outlined, the example looks legit to me. I think this should work with the latest fixes, please try out the branch using a dependency override.

MarthinL:
> Given what you've just outlined, the example looks legit to me. I think this should work with the latest fixes, please try out the branch using a dependency override.

The deps section in mix.exs being generated at the moment (and this had been part of my issue from the start) is this: On a freshly generated (mix phx.new test) project, this is the deps sections from mix.exs.

```
  defp deps do
    [
      {:phoenix, "~> 1.7.2"},
      {:phoenix_ecto, "~> 4.4"},
      {:ecto_sql, "~> 3.6"},
      {:postgrex, ">= 0.0.0"},
      {:phoenix_html, "~> 3.3"},
      {:phoenix_live_reload, "~> 1.2", only: :dev},
      {:phoenix_live_view, "~> 0.18.16"},
      {:floki, ">= 0.30.0", only: :test},
      {:phoenix_live_dashboard, "~> 0.7.2"},
      {:esbuild, "~> 0.7", runtime: Mix.env() == :dev},
      {:tailwind, "~> 0.2.0", runtime: Mix.env() == :dev},
      {:swoosh, "~> 1.3"},
      {:finch, "~> 0.13"},
      {:telemetry_metrics, "~> 0.6"},
      {:telemetry_poller, "~> 1.0"},
      {:gettext, "~> 0.20"},
      {:jason, "~> 1.2"},
      {:plug_cowboy, "~> 2.5"}
    ]

```

Your instructions are to add these two lines in there:
```
 {:ecto, github: "soundmonster/ecto", branch: "composite_foreign_keys", override: true},
 {:ecto_sql, github: "soundmonster/ecto_sql", branch: "composite_foreign_keys", override: true},
```

Now, if I simply add them, mix complains - 

````
warning: the dependency :ecto_sql is duplicated at the top level, please remove one of them
Dependencies have diverged:
````

If I go and comment out the ecto_sql line that was in the file, mix complains - 
````
warning: Ecto.Migrator.migrations/3 defined in application :ecto_sql is used by the current application but the current application does not depend on :ecto_sql. To fix this, you must do one of:

  1. If :ecto_sql is part of Erlang/Elixir, you must include it under :extra_applications inside "def application" in your mix.exs

  2. If :ecto_sql is a dependency, make sure it is listed under "def deps" in your mix.exs

  3. In case you don't want to add a requirement to :ecto_sql, you may optionally skip this warning by adding [xref: [exclude: [Ecto.Migrator]]] to your "def project" in mix.exs

  lib/phoenix_ecto/check_repo_status.ex:99: Phoenix.Ecto.CheckRepoStatus.migrations/3
````

 I'm guessing there was a time, however long ago, that the generated mix.exs file contained the two dependencies you've been working on, but it isn't the case anymore or something else happened along the way. Either way, it's confusing to me.

**What is the correct, complete deps section to use given the generated version above?**

**Comment by MarthinL** [comment]
Through trial and error I found a deps function that allows the a new project to compile and run with your PR code to be:

```
defp deps do
    [
      {:ecto, github: "soundmonster/ecto", branch: "composite_foreign_keys", override: true},
      {:ecto_sql, github: "soundmonster/ecto_sql", branch: "composite_foreign_keys", override: true},
      {:phoenix, "~> 1.7.2"},
      {:phoenix_ecto, "~> 4.4"},
      # {:ecto_sql, "~> 3.6"},
      {:postgrex, ">= 0.0.0"},
      {:phoenix_html, "~> 3.3"},
      ...
   ]
```

i.e. the ecto and ecto_sql with the overrides must be stated before phoenix itself because the phoenix (these days at least) bring in the dependencies on ecto and ecto_sql and that causes the conflict. I believe that generally the order of dependencies don't matter because of how mix resolves circular dependencies, but when override is specified, order does appear to change the semantics. The key here is that by the time the phoenix dependencies are evaluated, the already loaded ecto / ecto_sql packages are marked as being reused instead of imported. Having your packages with override: true after the phoenix dependency results in a dependency chain that mix does not know how to resolve reliably, so it throws a warning and a some alternative actions by which to resolve it. To the uninformed dealing with somebody else's changes, those options does not make much sense.



I'm happy to report that the schemas syntax as I wrote them now compiles without error or warnings. Now I need to test whether they function as expected, which is a whole other matter.

I have a question though. You stated in your instructions that to use/test your code I should 

> * upgrade to the absolute latest regular releases of ecto and ecto_sql

which (I presume) is the case when I create a new project with a newly updates version of phoenix. But I don't know how to do that for an older project. Can you give me some pointers? Should I change the version strings in mix.exs? Should I run some command to upgrade?

**Comment by soundmonster** [comment]
> Should I change the version strings in mix.exs? Should I run some command to upgrade?

First, check if you have to upgrade anything with `mix hex.outdated`. If the dependency you want to update is marked as "Update possible" then you can upgrade it with `mix deps.update <package_name_1> [... <package_name_2> ...]`, e.g. `mix deps.update ecto ecto_sql`. If the dependency is marked as "Update not possible" then its version defined in `mix.exs` is probably too restrictive and you'll have to make changes there.

You can skip the upgrade if you have been running on a reasonably recent version of Ecto.

**Comment by MarthinL** [comment]
I've extracted minimal schema code from my main project by which to test composite primary keys and came across something that's either a misbehaviour or something I have misunderstood about schemas. It's still a little complicated so please bear with me as I try to explain.

For the sake of the example, I have three schemas, Region, Coxel and Tie. As explained before, logically speaking I have only Region and Coxel, where Coxels have a self-referencing many-to-many relationship, like such:
```mermaid
erDiagram
    Region ||--o{ Coxel : has
    Coxel }o--o{ Coxel : "tied to"
```

But because in my use case each of the associations have a schema with additional attributes, the "hidden" associative table (coxelcoxels) is insufficient and I need to introduce a named schema which I call Tie (because it ties together two Coxels). So now we can represent the self-referencing many-to-many relationship in the manner it would actually be implemented in an RDBMS, like such"

```mermaid
erDiagram
    Region ||--o{ Coxel : has
    Region ||--o{ Tie : has
    Tie }o--|| Coxel : child
    Tie |o--|| Coxel : parent
```

If we then expand the diagram to show and track the actual database fields involved it becomes possible to describe the problem I am experiencing. Note that I've excluded many fields and associations used in my actual project since they don't impact on the problem but would obfuscate the issue.

```mermaid
erDiagram
    Region {
        id id PK
        string name
    } 
    Coxel {
        id region_id PK, FK
        id id PK
        text value
    }
    Tie {
        id region_id PK, FK "Tie region"
        id id PK
        string label
        id child_region_id FK "Child Coxel Region"
        id child_id FK
        id parent_region_id FK "Parent Coxel Region"
        id parent_id FK
    }
    Region ||--o{ Coxel : has
    Region ||--o{ Tie : has
    Tie }o--|| Coxel : child
    Tie |o--|| Coxel : parent

```

MarthinL:
> If we then expand the diagram to show and track the actual database fields involved it becomes possible to describe the problem I am experiencing. Note that I've excluded many fields and associations used in my actual project since they don't impact on the problem but would obfuscate the issue.

In code the three schemas look like this:
```
  # Region
  schema "regions" do
    field :name, :string

    timestamps()
  end


  # Coxel
  schema "coxels" do
    field :region_id, :id, primary_key: true
    field :value, :string

    belongs_to :region, Region,
               foreign_key: :region_id,
               references: :id,
               define_field: false

    has_many  :child_ties, Tie,
              foreign_key: [:parent_region_id, :parent_id],
              references: [:region_id, :id]

    has_many  :parent_ties, Tie,
              foreign_key: [:child_region_id, :child_id],
              references: [:region_id, :id]

    timestamps()
  end


  # Tie
  schema "ties" do
    field :label, :string
    field :region_id, :id, primary_key: true

    belongs_to :region, Region,
               foreign_key: :region_id,
               references: :id,
               define_field: false

    belongs_to :child, Coxel,
               foreign_key: [:child_region_id, :child_id],
               references: [:region_id, :id]

    belongs_to :child_region, Region,
               foreign_key: :child_region_id,
               references: :id,
               define_field: false

    belongs_to :parent, Coxel,
               foreign_key: [:parent_region_id, :parent_id],
               references: [:region_id, :id]

    belongs_to :parent_region, Region,
               foreign_key: :parent_region_id,
               references: :id,
               define_field: false

    timestamps()
  end

```

This all compiles and I am able to create regions and coxels that correctly references the regions. 

If I manualy construct a Tie I can insert it and when I query Coxels again with preload([child_ties: :child]) it returns the correct data. 

But I cannot get build_assoc to work. Granted, the test code I use is not directly the code from my main project that used to work, but a manually recreated version of it. 

Given that where I run the test, cm2 and ct3 are two valid Coxel variables, I'm expecting
 
```
Ecto.build_assoc(cm2, :child_ties, %{label: "Main2Test3"}) 
```
or 
```
Ecto.build_assoc(cm1, :child_ties, %{label: "Main2Test3", child_region_id: ct3.region_id, child_id: ct3.id})
```
or at the very least
```
Ecto.build_assoc(cm2, :child_ties, %{label: "Main2Test3", child_region_id: ct3.region_id, child_id: ct3.id, parent_region_id: cm1.region_id, parent_id: cm1.id}) 
```
to return a duely constructed Tie I can insert using Repo.insert(). But in all cases I get the error
```
** (KeyError) key [:child_region_id, :child_id] not found
    (ecto 3.10.1) lib/ecto/association.ex:873: Ecto.Association.Has.build/3
```

I constructed a separate test case without using composite primary keys which seems to work as expected but I lack the proficiency with Phoenix to be certain that the two cases are otherwise equivalent.

MarthinL:
I've made some progresss isolating the error when doing build_assoc but can't tell what's really going wrong. Hopefully my feedback/findings would get you closer to an answer.

** (KeyError) key .... not found
    (ecto 3.10.1) lib/ecto/association.ex:873: Ecto.Association.Has.build/3

I've set up four minimalistic projects to help isolate the error. 

* Project **one** uses the standard libraries and (obviously) no composite keys. The build_assoc statement succeeds.
* Project **two** uses the modified libraries but no composite keys. The build_assoc statement succeeds.
* Project **three** uses the modified libraries and the notation for composite keys in the association specification but with single item lists, essentially mimicking Project two above but with the modified syntax. The build_assoc statement fails with the error message 
** (KeyError) key [:parent_id] not found
    (ecto 3.10.1) lib/ecto/association.ex:873: Ecto.Association.Has.build/3
* Project **four** uses the modified libraries and region_id added to the composite keys. The build_assoc statement fails with the error message
** (KeyError) key [:parent_region_id, :parent_id] not found
    (ecto 3.10.1) lib/ecto/association.ex:873: Ecto.Association.Has.build/3

Although I don't have a minimal replication of that exact project to test with, I do have evidence that the build_assoc statement had been working as intended in the old version of the libraries modified for composite keys which I have modified in the manner documented in the thread above. I don't know enough about how association.ex and the build function is constructed to understand how the version that is in the code today achieves the same result my modification achieved but it clearly has a different outcome.

How can I be of more help? I'm committed to using composite keys and I wawnt to get my project into production so I don't lack motivation to help get the problem sorted. I just lack the context of how the ecto code goes about its business and I lack exposure and experience in open source project culture. Tell me what I can do to help and I will get it done.

MarthinL:
> Although I don't have a minimal replication of that exact project to test with, I do have evidence that the build_assoc statement had been working as intended in the old version of the libraries modified for composite keys which I have modified in the manner documented in the thread above. I don't know enough about how association.ex and the build function is constructed to understand how the version that is in the code today achieves the same result my modification achieved but it clearly has a different outcome.

Update: I now have a project referencing the old version of the PR with my modified build function. I can confirm that the  build_assoc call which fails with the current version works as expected with the old byt modified code. It seems to be a logical conclusion that the reason it now fails has something to do with the alternative way you elected to address the issue whereby the build function did not cater for composite keys. The original issue I addressed with the change I made to the code was because the association specification in the schema didn't respond to the list syntax for own and foreign keys. It's baffling that the schema syntax now works but functions like build_assoc which I presume uses the constructs built from the schema definition now fails to match the relevant keys when it's a list.

cspeper:
Hi @soundmonster @MarthinL. I appreciate your efforts so far to get this over the line. I've become interested in this PR as I have a production application in which we'd love to start making use of PG table partitioning and composite FKs seems like a must have to make that work. Is this PR still as close as it seems to getting merged or should we be exploring other medium term solutions?

soundmonster:
@cspeper Thank you for reaching out, and I appreciate your interest. 

TBH, I would love to use it in prod for my projects as well, but please do not expect this to be on the main branch this year.

From my point of view, this PR needs:
* a rebase, and
* a proof-of-concept app that would showcase it, and validate that it actually works (some things are hard to test here)
* probably another iteration driven by the feedback on the PoC. 

I think I'll have some first results in October, but realistically I don't think the feature will be on hex.pm in 2023.

fendent:
If there is anything I can do to help get this over the line, happy to help!

hickscorp:
@soundmonster I am really interested in using this work... I'm dealing with a schema where records in most tables are identified  with primary keys such as `[id: UUID, version: String]`.
It's all fine in migrations:

```
add :version_here, :string
add :foo_id, references(:foos, type: :uuid, with: [version_here: :version_there]
```

But then `has_one`, `has_many`, `belongs_to`... Not fine at all.

Are there any plans to bring this to either a usable  fork, or have it merged?

MarthinL:
> Alternativelly, did you find a reliable way to approach this problem (composite foreign keys) in production Ecto projects?

I ultimately made the call not to use this in production. I had to use my own modification compensate for an omission from @soundmonster's code but the ecto and composite key changes was just too foreign for me to be comfortable with either putting my change into production or take on the task of rebasing the entire thing. EActo and ecto_sql had changed quite a bit since the fork so it was never going to be a trivial task.

For my production app I switched to a completely different approach with regards to the solution I used to need composite keys in associations for. You might want to consider doing something similar.

Let me explain. 

My composite key used to consist of a region identifier and a normal bigint identifier within the region, meaning I maintained a region table and the database could enfore referential integrity on the region_id field of my regionalised tables. I had it working with the PR code but it's a royal PITA to modified especially generated code with the additional field that needs to be set, passed, defaulted and generally taken care of everywhere where Phoenix assumes a single id. 

So what I did was to use an opaque primary key that presented as one binary value but consisted internally of several components. In my case I also discovered that I couldn't use all the bits of a bigint anyway so I was able to split a bigint into bits for the region and bits for the id. You're already looking at a UUID key for reasons of your own. UUID is in effect already such an opaque type comprising several internal components gerring generated differently.

My recommendation would be to look at implementing your own type similar to UUID (I'm sure I've seen examples and tutorials lurking around the web) which include additional bits for the version id you want to add. You could shave some bits off one oof the components if you wish to keep it the same size as UUID or you could simply make it bigger by as much as you need. It's already merely a binary at the database level. You'd then use custom functions in your code to compare these keys with or without regard for the version, to extract and update the version portion as required and whatever else you need done. 

In hindsight I'm not really surprised that this PR never went mainstream. Once patched it does what it says on the box, but the rest of the Phoenix and LiveView eco-system where ecto normally get used in really is not geared towards composite keys. It's one primary key field or none at all (which legitimately happens in a few places). That's the idiom and ethos of Phoenix. When you venture outside of those boundaries especially for new applications and code, you're in for a world of hurt. The stated and proven reason why there's any support for composite keys in migrations etc. is to allow Ecto to be set up to read from legacy databases that were created with different design constraints. The assumption being that those tables would not really get actively involved in associations but would be donor tables from which the new schemas import legacy data into structures with proper associations. In the end the people who saw opportunity in having composite keys involved in active associations probably all came to the same painful conclusion and lost interest in the initiative.

MarthinL:
They crux is that you (still) have composite keys that can and does cause you trouble. I designed mine right out of existence, but that is not a universal approach. Perhaps the best advice I can offer you is that if I hadn't, and still had composite keys in my schema, I'd probably be handling it via the Ash framework instead.

PR #3641, opened 2021-05-21
improvement: support a `skip_type_check` option on fields

zachdaniel:
Hello! Due to an implementation detail in Ash (which I understand is not necessarily a concern of the ecto maintainers), I have types that can't necessarily be compiled when the ecto schema itself is compiled. This check supports a toggle called `skip_type_check` which allows this to be configured on a field-by-field basis. This could potentially be useful to anyone who is building ecto schemas and types behind macros as well.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -670,6 +670,9 @@ defmodule Ecto.Schema do
       when inspected in changes inside a `Ecto.Changeset` and be excluded
       from inspect on the schema. Defaults to `false`.
 
+    * :skip_type_check - By default, types are checked at compile time to ensure
+    that they are defined. If using parameterized types, this will result in your 
+    parameterized types *not* using the `init` callback.
   """
   defmacro field(name, type \\ :string, opts \\ []) do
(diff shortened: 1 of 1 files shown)


zachdaniel:
I'd be happy to add a test for this, but I wanted to run it by before-hand.

josevalim:
Sorry but I am worried about adding such mechanisms. It seems it should be tackled on the caller and we can’t guarantee we won’t add APIs in the future that require to be invoked at compile time. For example, one of our sketches for type specs support for fields was compile time based.

zachdaniel:
I thought that might be the case. Ultimately in that case then recursive types would need to be implemented not as embeds but ecto types that traverse/cast to/from maps. If I wanted to support them in Ash, I'd likely need to do it that way. In my research I'm also finding that, since graphql doesn't support recursive queries, recursively specified types really aren't all that useful for most contexts I care about. I'd ultimately have to be sending it back in JSON form over the graphql.

zachdaniel:
Thanks for the feedback :)

Issue #3619, opened 2021-04-21
Can't find named binding in CTE subquery with parent_as/1


joaopaulobdac:
### Environment

* Elixir version: 1.11.3
* Database and version: PostgreSQL 13
* Ecto version: 3.6.1
* Database adapter and version: postgrex 0.15.8
* Operating system: Arch Linux Kernel 5.11.15

### Current behavior

I'm trying to convert the following query from SQL to elixir:

```sql
SELECT g."name", breadcrumbs."names" AS "query_test" FROM "groups" AS g
LEFT JOIN LATERAL (
  -- breadcrumbs_query
  WITH RECURSIVE "breadcrumbs" AS (
    -- breadcrumbs_initial_query
    SELECT sg0."id", sg0."name", sg0."supergroup_id" FROM "groups" AS sg0
    WHERE sg0."id" = g."id"
    UNION ALL (
      -- breadcrumbs_recursion_query
      SELECT sg1."id", sg1."name", sg1."supergroup_id" FROM "groups" AS sg1
      INNER JOIN "breadcrumbs" AS b ON b."supergroup_id" = sg1."id"
    )
  )
  SELECT array_agg("name") AS "names" FROM "breadcrumbs"
) AS breadcrumbs ON TRUE;
```

Sample result:

```
+-----------------------+------------------------------------------------------------------------+
| name                  | query_test                                                             |
|-----------------------+------------------------------------------------------------------------|
| Supergroup            | ['Supergroup']                                                         |
| Subgroup 1            | ['Subgroup 1', 'Supergroup']                                           |
| Subgroup 1.1          | ['Subgroup 1.1', 'Subgroup 1', 'Supergroup']                           |
| Subgroup 1.1.1        | ['Subgroup 1.1.1', 'Subgroup 1.1', 'Subgroup 1', 'Supergroup']         |
| Subgroup 1.1.2        | ['Subgroup 1.1.2', 'Subgroup 1.1', 'Subgroup 1', 'Supergroup']         |
+-----------------------+------------------------------------------------------------------------+
```

And here's the query in elixir:

```elixir
def list_breadcrumbs do
  breadcrumbs_initial_query =
    Group
    |> where([g], g.id == parent_as(:group).id)

  breadcrumbs_recursion_query =
    Group
    |> join(:inner, [g], b in "breadcrumbs", on: b.supergroup_id == g.id)

  breadcrumbs_cte_query =
    breadcrumbs_initial_query
[...]

PR #3642, opened 2021-05-24
Fix CTE subqueries not finding parent bindings (#3619)

joaopaulobdac:
This is a work in progress since there are still changes to be made in ecto_sql because If you run the query ([I provided a repo for testing here](https://github.com/joaopaulobdac/parenting)) with this PR, you'll get an error coming from ecto_sql caused by the binding source not being found.
I'll also ping @v0idpwn here because maybe he can shred some light on this since he provided the PR that closed the issue (#3624), although CTEs still remain a problem. Thank you!

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1632,7 +1632,7 @@ defmodule Ecto.Query.Planner do
         if kind == :select and not (ix < tuple_size(sources)) do
           error!(query, "the parent_as in a subquery select used as a join can only access the `from` binding")
         else
-          {ix, {:parent_as, [], [{:&, meta, [ix]}]}, query}
+          {ix, {:parent_as, [], [as]}, query}
         end
 
       %{} = parent ->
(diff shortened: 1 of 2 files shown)


v0idpwn:
Hey, @joaopaulobdac, I'd be happy to help with that.  I'll try to reproduce the issue with the repo you provided.

v0idpwn:
Hey, @joaopaulobdac, sorry for taking so long. 

While debugging this issue, I found a bug in my PR too. While it generates a valid query, it may have incorrect bindings  (hopefully not a big deal as it wasn't working before anyways, so there are no repos affected). 

I think the way bindings are being managed now is not really consistent (#3627 related) and maybe it would be good to rethink it a bit. I'll try to fix the issue introduced in #3624, and maybe after doing that I can find some insight that helps in here. Other than that I don't have much to offer atm.

joaopaulobdac:
@v0idpwn don't worry, you don't owe anything :pray:. I don't know if the issue you found is the same as the one I'm having here, that is, while in ecto we can already recursively get the binding from a parent above 1 level, in ecto_sql we still only get it using a index from the direct parent of the query, which can generate incorrect bindings.

josevalim:
Closing in favor of PR.

josevalim:

PR #3647, opened 2021-05-28
Add mix dialyzer to +lint step

azizk:
Hi!
I hope you don't mind if I add dialyzer to let the CI check the code automatically from now on. :+1:

I haven't found anything serious, but I removed some dead code, added a missing function clause and refactored a function to make dialyzer happy.


lib/ecto/type.ex
```diff
@@ -821,8 +821,6 @@ defmodule Ecto.Type do
   defp cast_decimal(term) when is_binary(term) do
     case Decimal.parse(term) do
       {:ok, decimal} -> check_decimal(decimal, false)
-      {decimal, ""} -> check_decimal(decimal, false)
-      {_, remainder} when is_binary(remainder) and byte_size(remainder) > 0 -> :error
```
josevalim, line 825: This is intentional because we support multiple Decimal versions.

lib/ecto/uuid.ex
```diff
@@ -227,4 +228,5 @@ defmodule Ecto.UUID do
   defp e(13), do: ?d
   defp e(14), do: ?e
   defp e(15), do: ?f
+  defp e(_),  do: throw(:error)
```
josevalim, line 231: It is impossible to have any other value, so I am not a big of adding this to appease dialyzer. :)

josevalim:
Hi @azizk! We have no plans to run dialyzer in this project, but if there is something that can be improved, we will accept PRs for it. Thanks!

azizk:
Ah, didn't know that. Never mind then. Thanks for the feedback. :)

azizk:
@josevalim Quick questions: Can I salvage some parts of this PR? Do you know some major Elixir projects where a PR like this would be useful?

lib/ecto/uuid.ex
azizk: Ok, but dialyzer or not, the catch clause doesn't do anything and it seems to be a remnant and it's there because otherwise the compiler would issue a warning. Shouldn't it be refactored into:
```elixir
@spec encode(raw) :: {:ok, t}
defp encode(...) do
  uuid =
    ...

  {:ok, uuid}
end
```

josevalim: Ah yes, we should be able to remove the try/catch. Salvaging sounds good to me!

PR #3648, opened 2021-05-28
Fix a few specs and some code in Planner, Schema and UUID.

azizk:
The salvaged parts from the previous PR. I polished `Ecto.UUID` a bit more. Added a comment regarding `Decimal.parse` in `cast_decimal()`. Not sure if you will like the way I refactored `check_field_type!`, but one of the cond clauses was missing the call to `Code.ensure_compiled` and I think the most common case will be a base type.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -195,7 +195,7 @@ defmodule Ecto.Query.Planner do
   This function is called by the backend before invoking
   any cache mechanism.
   """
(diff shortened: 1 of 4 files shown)


azizk:
Please don't merge yet. There's one more thing I'd like to improve in Ecto.UUID. :)

azizk:
@josevalim K, ready now. :+1:

josevalim:

PR #3650, opened 2021-05-28
Earthfile: use ARG variables to define default database images.

azizk:
Makes it possible to override the db image version on the command line.


josevalim:

PR #3651, opened 2021-05-29
Fix example

princemaple:


wisq:
Related — Placeholder behaviour only works as intended if you're using a schema, not if you're using a table name:

```elixir
placeholders = %{blob: "a bunch of text"}

entries = [
  %{title: "v1", body: {:placeholder, :blob}},
  %{title: "v2", body: {:placeholder, :blob}}
]

# So this works ...
Repo.insert_all(Post, entries, placeholders: placeholders)

# But this doesn't.
Repo.insert_all("posts", entries, placeholders: placeholders)
```

In the latter case, it expects placeholders to be in the format `{:placeholder, n}` where `n` is an integer, and maps that field to `$n` in the query.  The `placeholders: ...` option appears to be completely ignored, so there doesn't seem to be an easy way to actually ensure that `$n` points to the value you want.

Is this intended behaviour?  If so, we should probably mention this in the docs.

josevalim:

josevalim:
@wisq good call, it has been fixed in master!

PR #3652, opened 2021-05-29
prepare_changes of a many assoc

elielhaouzi:
returns the changeset with its associations.

Before the fix, the changeset returned was the one after the assoc was extracted (pop_assocs function). 

The problem is when the schema has a many assoc, and one of the many has an error in the prepare_changes function (when it is not in the prepare_changes, the changeset is directly return in do_insert function with the valid?: false match), and since the parent changeset is set to invalid the changeset (without the assocs) is returned and then the changeset does not contain the assoc..

PS: I added a test only for insert, should I add one also for update ?

### DIFF
--- a/test/ecto/repo_test.exs
+++ b/test/ecto/repo_test.exs
@@ -61,7 +61,7 @@ defmodule Ecto.RepoTest do
       field :w, :string, virtual: true
       field :array, {:array, :string}
       field :map, {:map, :string}
-      has_many(:children, MySchemaChild)
+      has_many :children, MySchemaChild
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/ecto/repo_test.exs
```diff
@@ -49,6 +61,7 @@ defmodule Ecto.RepoTest do
       field :w, :string, virtual: true
       field :array, {:array, :string}
       field :map, {:map, :string}
+      has_many(:children, MySchemaChild)
```
josevalim, line 64:
```suggestion
      has_many :children, MySchemaChild
```

josevalim:
Great find! We indeed need similar to change and tests to `update` and we should be good to go, thanks!

elielhaouzi:
I also add a regression test for the `delete` function. if it is not needed I will remove it.

josevalim:

PR #3654, opened 2021-06-02
Add guard for binary prefix in Ecto.Query.put_query_prefix/2

tfwright:
Currently passing a query and a non-binary prefix causes an infinite loop


v0idpwn (approved):

wojtekmach:
Thank you!

PR #3658, opened 2021-06-09
Add Ecto.Changeset.traverse_validations/2

benhaney:
This allows for changeset validations to be traversed in the same way that errors are (reusing nearly all of the implementation of `traverse_errors`)

A potential use (and the reason I personally want it) is to easily expose the validations of a heavily-nested changeset to the web form that populates it, so the validations defined on the backend can be checked live on the client side as well.

Proposed in https://groups.google.com/g/elixir-ecto/c/yY9WP26ZJzc

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2936,34 +2936,39 @@ defmodule Ecto.Changeset do
       when is_function(msg_func, 1) or is_function(msg_func, 3) do
     errors
     |> Enum.reverse()
-    |> merge_error_keys(msg_func, changeset)
-    |> merge_related_keys(changes, types, msg_func)
+    |> merge_keyword_keys(msg_func, changeset)
+    |> merge_related_error_keys(changes, types, msg_func)
   end
 
-  defp merge_error_keys(errors, msg_func, _) when is_function(msg_func, 1)  do
+  defp merge_keyword_keys(errors, msg_func, _) when is_function(msg_func, 1)  do
     Enum.reduce(errors, %{}, fn({key, val}, acc) ->
       val = msg_func.(val)
       Map.update(acc, key, [val], &[val|&1])
     end)
   end
 
-  defp merge_error_keys(errors, msg_func, changeset) when is_function(msg_func, 3)  do
+  defp merge_keyword_keys(errors, msg_func, changeset) when is_function(msg_func, 3)  do
     Enum.reduce(errors, %{}, fn({key, val}, acc) ->
       val = msg_func.(changeset, key, val)
       Map.update(acc, key, [val], &[val|&1])
     end)
   end
 
-  defp merge_related_keys(_, _, nil, _) do
+  defp merge_related_error_keys(map, changes, types, msg_function) do
+    merge_related_keys(map, changes, types, msg_function, &traverse_errors/2)
+  end
+
+  defp merge_related_keys(_, _, nil, _, _) do
     raise ArgumentError, "changeset does not have types information"
   end
-  defp merge_related_keys(map, changes, types, msg_func) do
+
+  defp merge_related_keys(map, changes, types, msg_func, traverse_function) do
     Enum.reduce types, map, fn
       {field, {tag, %{cardinality: :many}}}, acc when tag in @relations ->
         if changesets = Map.get(changes, field) do
           {errors, all_empty?} =
             Enum.map_reduce(changesets, true, fn changeset, all_empty? ->
-              errors = traverse_errors(changeset, msg_func)
+              errors = traverse_function.(changeset, msg_func)
               {errors, all_empty? and errors == %{}}
             end)
 
@@ -2976,7 +2981,7 @@ defmodule Ecto.Changeset do
         end
       {field, {tag, %{cardinality: :one}}}, acc when tag in @relations ->
         if changeset = Map.get(changes, field) do
-          case traverse_errors(changeset, msg_func) do
+          case traverse_function.(changeset, msg_func) do
             errors when errors == %{} -> acc
             errors -> Map.put(acc, field, errors)
           end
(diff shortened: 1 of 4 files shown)


lib/ecto/changeset.ex
```diff
@@ -2976,7 +2981,7 @@ defmodule Ecto.Changeset do
         end
       {field, {tag, %{cardinality: :one}}}, acc when tag in @relations ->
         if changeset = Map.get(changes, field) do
-          case traverse_errors(changeset, msg_func) do
+          case traverse_function.(changeset, msg_func) do
             errors when errors == %{} -> acc
             errors -> Map.put(acc, field, errors)
```
v0idpwn, line 2986:
```suggestion
            items when items == %{} -> acc
            items -> Map.put(acc, field, items)
```
Doesn't refer to errors only anymore, so guess these should be renamed. Likely theres a better option than `items` tho.

lib/ecto/changeset.ex
```diff
@@ -2936,34 +2936,39 @@ defmodule Ecto.Changeset do
...
+  defp merge_related_keys(map, changes, types, msg_func, traverse_function) do
     Enum.reduce types, map, fn
       {field, {tag, %{cardinality: :many}}}, acc when tag in @relations ->
         if changesets = Map.get(changes, field) do
           {errors, all_empty?} =
             Enum.map_reduce(changesets, true, fn changeset, all_empty? ->
-              errors = traverse_errors(changeset, msg_func)
+              errors = traverse_function.(changeset, msg_func)
               {errors, all_empty? and errors == %{}}
             end)
```
v0idpwn, line 2973: Same as above

lib/ecto/changeset.ex
```diff
@@ -2936,34 +2936,39 @@ defmodule Ecto.Changeset do
       when is_function(msg_func, 1) or is_function(msg_func, 3) do
     errors
     |> Enum.reverse()
-    |> merge_error_keys(msg_func, changeset)
-    |> merge_related_keys(changes, types, msg_func)
+    |> merge_keyword_keys(msg_func, changeset)
+    |> merge_related_error_keys(changes, types, msg_func)
```
v0idpwn, line 2940:
```suggestion
    |> merge_related_keys(changes, types, msg_func, &traverse_errors/2)
```
No need to hide this in a private function

lib/ecto/changeset.ex
```diff
@@ -3000,6 +3005,37 @@ defmodule Ecto.Changeset do
...
+      ...> end)
+      %{title: [format: "/pattern/", length: "1-20"]}
+  """
+  @spec traverse_validations(t, (error -> String.t) | (Changeset.t, atom, error -> String.t)) :: %{atom => [String.t | map]}
+  def traverse_validations(%Changeset{validations: validations, changes: changes, types: types} = changeset, msg_func)
+      when is_function(msg_func, 1) or is_function(msg_func, 3) do
+    validations
+    |> Enum.reverse()
+    |> merge_keyword_keys(msg_func, changeset)
+    |> merge_related_validation_keys(changes, types, msg_func)
```
v0idpwn, line 3033:
```suggestion
    |> merge_related_validation_keys(changes, types, msg_func, &traverse_validations/2)
```

lib/ecto/changeset.ex
```diff
@@ -2936,34 +2936,39 @@ defmodule Ecto.Changeset do
...
     Enum.reduce(errors, %{}, fn({key, val}, acc) ->
       val = msg_func.(val)
       Map.update(acc, key, [val], &[val|&1])
     end)
   end
 
-  defp merge_error_keys(errors, msg_func, changeset) when is_function(msg_func, 3)  do
+  defp merge_keyword_keys(errors, msg_func, changeset) when is_function(msg_func, 3)  do
     Enum.reduce(errors, %{}, fn({key, val}, acc) ->
       val = msg_func.(changeset, key, val)
```
v0idpwn, line 2952: Same thing regarding the `errors` variable for these two functions.

v0idpwn:
Looks good, I commented some small code improvements.

wojtekmach:
Should this function work on constraints too (then the naming is perhaps misleading) or we’d need a separate traverse_constraints/2? 

Maybe we allow this to build up things at the same time:

```
traverse(changeset, [:validations, :constraints, :errors], fn
  :validation, … ->
  :constraint, … ->
  …
end)
```

or that’d be too complicated?

wojtekmach:
oh, while traversing both validations and constraints may make sense, traversing both validations and errors probably doesn’t.

wojtekmach:
> A potential use (and the reason I personally want it) is to easily expose the validations of a heavily-nested changeset to the web form that populates it, so the validations defined on the backend can be checked live on the client side as well.

Right, so for this particular use case, traversing constraints is not going to be useful.

I wonder if we could come up with other cases for this feature? I'm starting to think that leaving this as is and then possibly having `traverse_constraints/2` (when we have a good case for _that_) is fine.

benhaney:
@wojtekmach
Since we're going to have `traverse_errors/2` for the foreseeable future, I think it probably makes sense to continue with a family of `traverse_*/2` functions as needed instead of one `traverse_errors/2` and one `traverse/2` that handles everything (or everything else).
And I agree on waiting for a use case for `traverse_constraints/2`. Ecto has managed for a long time without `traverse_validations/2`, it can probably go quite a bit longer or even indefinitely without `traverse_constraints/2`

benhaney:
@v0idpwn 
Replaced those instances of variables named `errors` you caught with the name `child` where they were being merged into a parent map, and `keyword_list` where they were being used in `merge_keyword_keys`. Thanks for catching that, and please feel free to unresolve those suggestions if you think different names would be more suitable.

lib/ecto/changeset.ex
```diff
@@ -3000,6 +3001,33 @@ defmodule Ecto.Changeset do
...
+      %{title: [format: ~r/pattern/, length: [min: 1, max: 20]]}
+
+      iex> traverse_validations(changeset, fn
+      ...>   {:length, opts} -> {:length, "#{Keyword.get(opts, :min, 0)}-#{Keyword.get(opts, :max, 32)}"}
+      ...>   {:format, %Regex{source: source}} -> {:format, "/#{source}/"}
+      ...>   {other, opts} -> {other, inspect(opts)}
+      ...> end)
+      %{title: [format: "/pattern/", length: "1-20"]}
+  """
+  @spec traverse_validations(t, (error -> String.t) | (Changeset.t, atom, error -> String.t)) :: %{atom => [String.t | map]}
```
josevalim, line 3023:
```suggestion
  @spec traverse_validations(t, (error -> String.t) | (Changeset.t, atom, error -> String.t)) :: %{atom => term}
```

josevalim: I think we need to do the same change for traverse_errors... theoretically it can return anything, no?

josevalim:
This looks great to me, thanks @benhaney! I dropped just one comment about typespecs. You happy with this @wojtekmach and @v0idpwn?

v0idpwn (approved):

lib/ecto/changeset.ex
benhaney: Don't the values still have to be a list? ie `%{atom => [term]}`

josevalim: Yes, sorry!

benhaney: 👍 updated

josevalim:

PR #3662, opened 2021-06-10
Document that `rollback` causes the code in the transaction to stop e…

mathieuprog:
…xecuting


josevalim:

PR #3663, opened 2021-06-10
Document that `validations/1` doesn't return the required fields

mathieuprog:


lib/ecto/changeset.ex
```diff
@@ -1614,6 +1614,9 @@ defmodule Ecto.Changeset do
     * `{:inclusion, Enum.t}`
     * `{:subset, Enum.t}`
 
+  Note that calling `validate_required/3` does not store the validation under the
+  changeset.validations key (and so won't be included in the result of this
+  function). The required fields are stored under the changeset.required key.
```
josevalim, line 1619:
```suggestion
  Note that calling `validate_required/3` does not store the validation under the
  `changeset.validations` key (and so won't be included in the result of this
  function). The required fields are stored under the `changeset.required` key.
```

josevalim:

PR #3667, opened 2021-06-11
Improve `validations/1` documentation

mathieuprog:


lib/ecto/changeset.ex
```diff
@@ -1622,6 +1621,8 @@ defmodule Ecto.Changeset do
   Note that calling `validate_required/3` does not store the validation under the
   `changeset.validations` key (and so won't be included in the result of this
   function). The required fields are stored under the `changeset.required` key.
+
+  The primary motive for `validations` is to generate HTML5 validations.
```
josevalim, line 1625: I don't think that's the primary motive. They are there for reflection purposes. HTML5 validation is one example. Maybe others exist in other domains. But it doesn't have to be tied to web. :)

mathieuprog: I briefly checked the codebase and it makes no use of `:validations`. Then I went on looking for other major repos that depend on ecto, such as phoenix_ecto, and found the HTML5 validations generator.

I know ecto is not tied to html forms or the web specifically, but it's nice to know the *initial* reason for the existence of this function. There's no reflection provided without reasons/use cases.

If it is used for reflection in some other major repo that depend on ecto, then indeed it is a little wrong. It just all made sense for me when I found phoenix_ecto's function.

Anyway no offense if discarded 🤣

mathieuprog: And I can send a new PR if you wanna keep that other minor text change.

josevalim: I can discard this small change, no worries. :) Thanks!

josevalim, line 1625:
```suggestion
```

josevalim:

PR #3670, opened 2021-06-13
Clearer error message when an embedded struct is expected.

azizk:
I tripped over this issue a while ago and only found the solution in the Elixir forum. Just want to make sure others aren't puzzled by the error message. :)

I have two questions though:
* In the first test the returned row contains a plain map and not the embedded struct. Is this correct?
* It doesn't make sense to implicitly cast a map value to the embedded struct, or would it?

### DIFF
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -57,24 +57,24 @@ defmodule Ecto.Embedded do
   def load(nil, _fun, %{cardinality: :one}), do: {:ok, nil}
 
   def load(value, fun, %{cardinality: :one, related: schema, field: field}) when is_map(value) do
-    {:ok, load(field, schema, value, fun)}
+    {:ok, load_field(field, schema, value, fun)}
   end
 
   def load(nil, _fun, %{cardinality: :many}), do: {:ok, []}
 
   def load(value, fun, %{cardinality: :many, related: schema, field: field}) when is_list(value) do
-    {:ok, Enum.map(value, &load(field, schema, &1, fun))}
+    {:ok, Enum.map(value, &load_field(field, schema, &1, fun))}
   end
 
   def load(_value, _fun, _embed) do
     :error
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
It looks correct but embeds_many expects a list of maps or a list of structs, so you need to check the cardinality for the error message.

azizk:
The cardinality is already being checked in `Ecto.Embedded.dump/3`, isn't it? Btw, I'd suggest to rename `dump/5` and `load/4` to `dump_field/5` and `load_field/4` respectively. They confusingly look very much like the behaviour functions from `Ecto.ParameterizedType`.

Also when you pass a single embed instead of a list, like so `TestRepo.insert_all(MySchemaWithEmbed, [%{embeds: %MyEmbed{x: "x"}}])`, you get a pretty long error message:

`** (Ecto.ChangeError) value `%Ecto.RepoTest.MyEmbed{id: nil, x: "x"}` for `Ecto.RepoTest.MySchemaWithEmbed.embeds` in `insert_all` does not match type {:parameterized, Ecto.Embedded, %Ecto.Embedded{cardinality: :many, field: :embeds, on_cast: nil, on_replace: :raise, ordered: true, owner: Ecto.RepoTest.MySchemaWithEmbed, related: Ecto.RepoTest.MyEmbed, unique: true}}`

This is coming from `Ecto.Repo.Schema.dump_field!/6`. Maybe it's worth improving this as well.

josevalim:
Yes, that is a separate discussion though. For this PR, what is important is for the error message needs to consider the cardinality. :)

azizk:
Alright, so I added another test schema which uses an `embeds_one`. The error message is dependent on the cardinality now. But it doesn't really sound good if you ask me. For example:

`cannot dump embed ``embeds``, expected a list of Ecto.RepoTest.MyEmbed struct values but got: %{x: "x"}`

The "but got: ..." part still lists a single value so it doesn't really fit the previous statement of expecting a list of struct values...

josevalim:

PR #3675, opened 2021-06-17
Fix changeset inspection not redacting when embedded

### LINKED ISSUE #3674: Embedded schemas do not support redact option
### Environment

* Elixir version (elixir -v): Elixir 1.10.3 (compiled with Erlang/OTP 22)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgres 10
* Ecto version (mix deps): 3.5.6
* Database adapter and version (mix deps): 0.15.8
* Operating system: OS X 11.4

### Current behavior

Embedded schema:
```elixir
defmodule MySchema do
  use Ecto.Schema

  embedded_schema do
    field :redacted_field, :string, redact: true
  end
end
```

```elixir
iex(1)> c = Ecto.Changeset.cast(%MySchema{}, %{redacted_field: "secret"}, [:redacted_field])
#Ecto.Changeset<
  action: nil,
  changes: %{redacted_field: "secret"},
  errors: [],
  data: #MySchema<>,
  valid?: true
>
iex(2)> inspect c
"#Ecto.Changeset<action: nil, changes: %{redacted_field: \"secret\"}, errors: [], data: #MySchema<>, valid?: true>"
```

### Expected behavior

When inspecting the changeset it would be expected for the `redacted_field` value to be redacted.

```elixir
iex(2)> inspect c
"#Ecto.Changeset<action: nil, changes: %{redacted_field: \"**redacted**\"}, errors: [], data: #MySchema<>, valid?: true>"
```

v0idpwn:
The protocol implementation was matching the `__meta__` key in the
struct. This key isn't available on embedded structs.

Closes #3674

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -3046,7 +3046,12 @@ defimpl Inspect, for: Ecto.Changeset do
     end
 
     redacted_fields = case data do
-      %type{__meta__: _} -> type.__schema__(:redact_fields)
+      %type{} -> 
+        if function_exported?(type, :__schema__, 1) do
+          type.__schema__(:redact_fields)
+        else
+          []
+        end
       _ -> []
     end
 
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1905,6 +1905,17 @@ defmodule Ecto.ChangesetTest do
     end
   end
 
+  defmodule RedactedEmbeddedSchema do
+    use Ecto.Schema
+
+    embedded_schema do
+      field :password, :string, redact: true
+      field :username, :string
+      field :display_name, :string, redact: false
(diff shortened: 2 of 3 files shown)


lib/ecto/changeset.ex
```diff
@@ -3046,7 +3046,12 @@ defimpl Inspect, for: Ecto.Changeset do
     end
 
     redacted_fields = case data do
-      %type{__meta__: _} -> type.__schema__(:redact_fields)
+      %type{} -> 
+        if function_exported?(type, :__schema__, 1) do
+          type.__schema__(:redact_fields)
+        else
+          []
```
v0idpwn, line 3053: If there's a more clever/idiomatic way to check if a struct is either an schema or embedded schema, please tell me :)

josevalim:

PR #3676, opened 2021-06-18
Add Ecto.Enum.mappings/2

v0idpwn:
The `mappings/2` function returns the enum values and their respective
mappings in a keyword list.

### DIFF
--- a/test/ecto/enum_test.exs
+++ b/test/ecto/enum_test.exs
@@ -25,7 +25,7 @@ defmodule Ecto.EnumTest do
                 %{
                   on_load: %{"bar" => :bar, "baz" => :baz, "foo" => :foo},
                   on_dump: %{bar: "bar", baz: "baz", foo: "foo"},
-                  values: [:foo, :bar, :baz],
+                  mappings: [foo: "foo", bar: "bar", baz: "baz"],
                   type: :string
                 }}
 
@@ -36,7 +36,7 @@ defmodule Ecto.EnumTest do
                   %{
                     on_dump: %{bar: "bar", baz: "baz", foo: "foo"},
                     on_load: %{"bar" => :bar, "baz" => :baz, "foo" => :foo},
-                    values: [:foo, :bar, :baz],
+                    mappings: [foo: "foo", bar: "bar", baz: "baz"],
                     type: :string
                   }}
                }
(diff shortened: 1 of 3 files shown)


v0idpwn:
I think originally Ecto.Enum was intended to work with exact matches only, so this didn't made a lot of sense. Now that it supports `:atom => arbitrary_strings` and `:atom => :arbitrary_integer`, I think it is a good idea having this function.

josevalim:
Yes, this makes sense, but puts us in a weird spot. Ecto.Enum.values do not return the values of the mappings... but the keys. They are loaded values though. I wonder if we should deprecate `values` and tell people to use `mappings` instead.

test/ecto/enum_test.exs
```diff
@@ -78,7 +79,7 @@ defmodule Ecto.EnumTest do
                   %{
                     on_dump: %{bar: "baar", baz: "baaz", foo: "fooo"},
                     on_load: %{"baar" => :bar, "baaz" => :baz, "fooo" => :foo},
-                    values: [:foo, :bar, :baz],
+                    mappings: [foo: "fooo", bar: "baar", baz: "baaz"],
```
AndrewDryga, line 82: Would map be a more reasonable data structure here? Because keyword would let you do something like this: `[foo: "foo", foo: "fooo"]` and the first one would win.

wojtekmach: I think we can detect and raise on duplicate keys. I can see how using list and thus preserving order can be useful for reflection purposes but no strong opinion either way.

v0idpwn: My reason to chose a keyword list over a map is the fact that the user inputs a keyword list when creating a Ecto.Enum field, so it seemed consistent returning the exact same thing the user inputted. Yet, as you pointed out, this didn't reflect the exact mappings that are done on `on_dump` and `on_load` (these are maps, and the last one would take precedence). I opened another PR that prevents the user from inputting keyword lists with duplicate keys.

v0idpwn:
I don't have an answer for that, @josevalim. My first thought was having both but indeed it feels a bit weird.

josevalim:
Let’s add mapping and deprecate values. :)

v0idpwn:
Added a commit with the deprecation.
I'm not really a fan of it it because I find `Ecto.Enum.values/2` to be really ergonomic.

josevalim:
Another option is to have values, dump_values, and mappings. So we extend the existing API with two types of values for clarity?

v0idpwn:
Sounds good to me. While `dump_values` doesn't seem like the most useful function, I think having it and values combined gives us enough consistency without sacrificing ergonomics. Will address that

josevalim:

PR #3677, opened 2021-06-19
Raise on duplicate values in Ecto.Enum

v0idpwn:
Duplicate values make so that only the last of the duplicates is taken
into account, due to the fact that the keyword list is cast into a map
for the `on_load` and `on_dump` values. With this commit we can avoid
this scenario in compile time.

Pointed out by @wojtekmach on #3676


josevalim:

PR #3678, opened 2021-06-19
Clearer error message when a map is expected for an embedded struct.

azizk:
This one is similar to the last PR I made. Last one was about `dump_field`, but I missed doing the same for `load_field`. I added an integration test but I have no idea if it's going to pass. Let's see. :)

I threw in an auto-format commit but you can ignore it if it's not important.

### DIFF
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -347,17 +347,13 @@ defmodule Ecto.Integration.TypeTest do
 
   @tag :map_type
   test "error when a map for an embedded struct is needed" do
-    item = %Item{price: 123, valid_at: ~D[2014-01-16]}
-
     assert_raise ArgumentError, ~r"expected a map value", fn ->
-      order = TestRepo.insert!(%Order{item: item})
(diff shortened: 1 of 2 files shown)


azizk:
Looks like it's raising a different error:
```
Expected exception ArgumentError but got Ecto.Query.CastError (/src/ecto/lib/ecto/repo/queryable.ex:171: value `123` in `update` cannot be cast to type {:parameterized, Ecto.Embedded, %Ecto.Embedded{cardinality: :one, field: :item, on_cast: nil, on_replace: :raise, ordered: true, owner: Ecto.Integration.Order, related: Ecto.Integration.Item, unique: true}} in query
```

josevalim:
To test the load callback you will need to manually insert bad data in the database, perhaps by using an insert_all with a string table name.

azizk:
Ah right, thanks! I tried it but I'm still getting an exception from `Ecto.Repo.Queryable.load!/5` first. I'll continue checking...

azizk:
I'm stuck and didn't have much time to check further. Gonna close this for now.

PR #3685, opened 2021-07-03
Add support for dynamic as(^as) and parent_as(^as)

arnodirlam:
This PR allows using `as(^as)` and `parent_as(^as)` for dynamically using aliases passed at runtime.

See [discussion](https://groups.google.com/g/elixir-ecto/c/RaoK6oom_FM) for rationale.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -570,16 +570,16 @@ defmodule Ecto.Query.Builder do
   defp escape_field!({var, _, context}, field, vars)
        when is_atom(var) and is_atom(context) do
     var   = escape_var!(var, vars)
-    field = get_quoted!(field)
+    field = quoted_atom!(field)
     dot   = {:{}, [], [:., [], [var, field]]}
     {:{}, [], [dot, [], []]}
   end
 
   defp escape_field!({kind, _, [atom]}, field, _vars)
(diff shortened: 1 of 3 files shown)


lib/ecto/query/builder.ex
```diff
@@ -872,20 +873,29 @@ defmodule Ecto.Query.Builder do
   Checks if the field is an atom at compilation time or
   delegate the check to runtime for interpolation.
   """
-  def quoted_field!({:^, _, [expr]}),
-    do: quote(do: Ecto.Query.Builder.field!(unquote(expr)))
-  def quoted_field!(atom) when is_atom(atom),
+  def quoted_atom!(_ast, used_fun \\ "field", used_arity \\ 2)
```
josevalim, line 876: Instead of two arguments, have a single argument called used_ref, and you can pass "field/2". Also, remove the default argument and instead make sure that all callers pass the proper reference. :)

arnodirlam: Addressed in [a85b3fa](https://github.com/elixir-ecto/ecto/pull/3685/commits/a85b3fad4efe11f148988fe8880fdd4da511b3f5).

josevalim:

Rutaba:
where can I find the documentation for these changes?

PR #3687, opened 2021-07-05
Fix Ecto.Repo.aggregate specs

qcam:
I think the function won't be ever returning `nil`?


wojtekmach:
aggregate/2 always returns an integer given it uses the :count aggregate, an empty set of rows would return a 0. So +1 on that change. aggregate/3 for other aggregates can be nil on empty rows set though.

qcam:
Good catch. I will revert the change for `aggregate/3` then.

PR #3688, opened 2021-07-06
Add support for @schema_context to set context metadata on schema definition

fcevado:
Currently, it isn't possible to set context metadata on schema definition. The only way to set that is by using `Ecto.put_meta/2`. This PR adds support for defining context on schema definition with `@schema_context`, a `_schema__(:context)` definition is added too to expose the context set to a schema.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -127,6 +127,10 @@ defmodule Ecto.Schema do
       is called "SCHEMA" (typically set via Postgres' `search_path`).
       In MySQL the prefix points to databases.
 
+    * `@schema_context` - configures the schema context. Defaults to `nil`,
+      which generates structs and queries without context. Context are not used
+      by the built-in adapters of `Ecto.Adapters.Postgres` and `Ecto.Adapters.MySQL`.
+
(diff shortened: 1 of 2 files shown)


fcevado:
I just saw that are a lot of unnecessary formatting changes, should I remove that?

josevalim:
Can you please send a PR without formatting the code? It will make it better to review and for history to.

Also, we shouldn’t export the reflection function, as the data is already in the meta. Thanks!

fcevado:
updated with no formating and removing the reflection function.

lib/ecto/schema.ex
```diff
@@ -127,6 +127,10 @@ defmodule Ecto.Schema do
       is called "SCHEMA" (typically set via Postgres' `search_path`).
       In MySQL the prefix points to databases.
 
+    * `@schema_context` - configures the schema context. Defaults to `nil`,
+      which generates structs and queries without context. Context are not used
+      by the built-in adapters of `Ecto.Adapters.Postgres` and `Ecto.Adapters.MySQL`.
```
josevalim, line 132:
```suggestion
      by the built-in SQL adapters.
```

josevalim:

PR #3692, opened 2021-07-10
Accept binary representation of atom on Ecto.Enum.cast/3

### LINKED ISSUE #3691: Enum with custom integer values doesn't cast string to atom
It's possible this is by design (I hope not!) but even so, improved documentation would be nice!

Given this field:
`field :status, Ecto.Enum, values: [foo: 1, bar: 2, baz: 5]`

If I try `%Struct{} |> Ecto.Changeset.cast(%{status: "baz"}, [:baz])` it will return as invalid with an error.  This same field would accept either `:baz` or `5` as the value, but will not cast the string into an atom.  Obviously, form data will be coming in as strings and not atoms, so my thought is that it would be desirable for the field to cast the string into an atom if its value is in the allowed list.  If this is not desired for some reason, it would be helpful for the documentation to spell this out more clearly, because the documentation reading "_Values will be cast to atoms safely and only if the atom exists in the list (otherwise an error will be raised). Attempting to load any string/integer not represented by an atom in the list will be invalid._" makes it seem like it should.

Elixir 1.12.2, Ecto 3.6.2.

Thanks!

v0idpwn:
We currently rely on the database representation for casting. This means
that, if :foo translates to 1 in the database, we only accept 1  and
:foo as input when casting. This PR changes it so "foo" is also
accepted.

Closes #3691

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -93,8 +93,10 @@ defmodule Ecto.Enum do
       end
 
     on_load = Map.new(mappings, fn {key, val} -> {val, key} end)
-    on_dump = Enum.into(mappings, %{})
-    %{on_load: on_load, on_dump: on_dump, mappings: mappings, type: type}
+    on_dump = Map.new(mappings)
+    on_cast = Map.new(mappings, fn {key, _} -> {Atom.to_string(key), key} end)
+
+    %{on_load: on_load, on_dump: on_dump, on_cast: on_cast, mappings: mappings, type: type}
   end
 
   defp validate_unique!(values) do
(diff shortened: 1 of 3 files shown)


v0idpwn:
Just a disclaimer: if you think that parameterized type params growing big is an issue, we may calculate everything from the `mappings` param. But it implies a small performance overhead, I think

v0idpwn:
Also, behaviour: 

If you have: [zero: "one", one: "two"]

casting "one" will return :zero

Otherwise it would be a compatibility break

josevalim:

Issue #3695, opened 2021-07-22
Ecto.Changeset.validate_number/3 Does not verify that the value is actually a number 


tomciopp:
### Environment

* Elixir version (elixir -v): 1.12.2
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgresql 13.3
* Ecto version (mix deps): 3.6.2
* Database adapter and version (mix deps): postgrex 0.15.9
* Operating system: OSX 11.5 (Big Sur)

### Current behavior

Include code samples, errors and stacktraces if appropriate.

```elixir
defmodule MyApp.Product do 
 ...
  def changeset(%Product{} = record, params) do
    record
    |> cast(params, [:value])
    |> validate_required([:value])
    |> validate_number(changeset, :value, greater_than_or_equal_to: 0)
  end
end

changeset = Product.changeset(%Product{}, %{"value" => "Oops"})
changeset.valid? => true
```

### Expected behavior

The changeset should not be valid, there should be an error returned that the value passed is not a number

### Why this should change

I'm sure the first reaction to this issue, is that casting should catch this error. However there are certain edge cases that cause this validation to fail. Specifically if you are encrypting the field before it is stored in the database, the underlying type for the field will be binary. Therefore casting will not catch the error and it will try to do a comparison for mismatched types, which will work in elixir. 

### The fix

We should add a guard clause `when is_number(value)` to the following line.
https://github.com/elixir-ecto/ecto/blob/e9dead98b1fe2f20358faee48b7fe00f17a71bc2/lib/ecto/changeset.ex#L2208

We should also add an additional function to handle values that are not numbers and return a format error. Something like.

```elixir
defp validate_number(field, _value, message, _spec_key, _spec_function, _target_value) do
  {field, {"has invalid format", [validation: :format]}}
end
```

josevalim:
If you give it something that is not a number, then in my opinion it should raise, as we won't automatically cast, and cast should handle the field properly. Thoughts?

tomciopp:
@josevalim I'm fine with that behavior as well.

josevalim:
@tomciopp can you please send a PR? :)

tomciopp:
@josevalim Yeah sure, give me a few minutes to put everything together and I'll send it along

dvic:
[...]

PR #3696, opened 2021-07-22
bugfix(Ecto.Changeset#validate_number/3): Raise an exception when we …

tomciopp:
…try to validate against a value that is not a number.

Handles issue #3695


josevalim:

PR #3697, opened 2021-07-22
Overridable get_dynamic_repo/0

cjbottaro:
Dynamic repos don't work well with libraries that take a repo and make queries on your behalf in async tasks, which is the case with Absinthe and Dataloader.

I think the easiest workaround is to make `get_dynamic_repo/0` overridable and let the user figure out how to track the current repo across child processes. Here's an example implementation that works with `put_dynamic_repo/1`:

```elixir
def get_dynamic_repo do
  [self() | Process.get(:"$callers", [])]
  |> Enum.find_value(fn pid ->
    {:dictionary, dictionary} = Process.info(pid, :dictionary)
    Enum.find_value(dictionary, fn
      {{__MODULE__, :dynamic_repo}, repo} -> repo
      _ -> nil
    end)
  end)
  |> case do
    nil -> super()
    repo -> repo
  end
end
```

But one could also just completely hand off dynamic repo management to an external multi tenancy library (which is my team's use case):
```elixir
def get_dynamic_repo do
  MyMultiTenancyLibrary.get_current_tenant()
  |> tenant_to_dynamic_repo()
end
```

Then you don't have to worry about the implementation of `put_dynamic_repo/1`.

Thanks!

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -247,6 +247,7 @@ defmodule Ecto.Repo do
       def get_dynamic_repo() do
         Process.get({__MODULE__, :dynamic_repo}, @default_dynamic_repo)
       end
+      defoverridable get_dynamic_repo: 0
 
       def put_dynamic_repo(dynamic) when is_atom(dynamic) or is_pid(dynamic) do
         Process.put({__MODULE__, :dynamic_repo}, dynamic) || @default_dynamic_repo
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks for the example! I understand the use case but ideally we want to push a callback upstream to those libs, so you control the spawning, meanwhile you can mark it as overridable yourself. :)

cjbottaro:
> meanwhile you can mark it as overridable yourself.

Can you explain how to do that? I didn't realize that was a feature.

josevalim:
I believe you can call defoverridable yourself or, in worst case scenario, do so inside a @before_compile hook.

cjbottaro:
Omg, it never occurred to me this would just work...
```elixir
defmodule Repo do
  use Ecto.Repo

  defoverridable get_dynamic_repo: 0

  def get_dynamic_repo do
    # ...
  end
end
```
Well, I hit my embarrassment quota for the month... :D

PR #3699, opened 2021-07-31
Update doc: Repo.transaction/2 may raise Ecto.StaleEntryError

mathieuprog:


josevalim:
This is not really related to transactions. They will reraise anything and this can also happen if you load the record, and then try to write to it outside a transaction. If you are loading inside the transaction, then it depends on the serialization level.

mathieuprog:
But does that matter? The user works with the public API, and `Repo.transaction/2` may raise `Ecto.StaleEntryError`. This is not documented.

If I look at the stack:

> stacktrace:
>   (ecto 3.6.2) lib/ecto/repo/schema.ex:742: Ecto.Repo.Schema.apply/4
>   (ecto 3.6.2) lib/ecto/repo/schema.ex:434: anonymous fn/15 in Ecto.Repo.Schema.do_update/4
>   (ecto 3.6.2) lib/ecto/multi.ex:716: Ecto.Multi.apply_operation/5
>   (elixir 1.11.3) lib/enum.ex:2193: Enum."-reduce/3-lists^foldl/2-0-"/3
>   (ecto 3.6.2) lib/ecto/multi.ex:690: anonymous fn/5 in Ecto.Multi.apply_operations/5
>   (ecto_sql 3.6.2) lib/ecto/adapters/sql.ex:1017: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
>   (db_connection 2.4.0) lib/db_connection.ex:1512: DBConnection.run_transaction/4
>   (ecto 3.6.2) lib/ecto/repo/transaction.ex:20: Ecto.Repo.Transaction.transaction/4

It seems that in terms of the public API, that it is `Repo.transaction/2` that raises, all the other calls are private/implementation details. It can happen because of the operation added by `Multi.update/4` for example.

Shouldn't we document the possible error that is raised for a function? I'm trying to understand the philosophy:)

mathieuprog:
Btw that's why I added that piece of doc under the Multi section. For Repo calls inside the transaction, I understand your point. But if you call `Multi.update/4` and you got a stale error, it should be documented that an error is raised.

PR #3700, opened 2021-07-31
Add stale struct in Ecto.StaleEntryError fields

mathieuprog:
If a stale entry error is raised, it is not possible to know which struct/schema is stale if you're working with associations.

Use case: say I want to update an appointment with its guests, if I get a stale entry, I would like to display an error message such as "The appointment no longer exists." or "A guest cannot be found" according to which struct is stale, as an example.

Note that I'm a little surprised about how stale entries are handled by default (https://groups.google.com/g/elixir-ecto/c/hdBMC0g8WgE), i.e. throwing errors while stale entries in a concurrent environment are mostly not errors, and maybe I am doing something wrong, so review carefully this PR. However it seems like a basic need for me to know which struct is stale in order to handle the error correctly for the user (appropriate error message).

### DIFF
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -258,13 +258,12 @@ defmodule Ecto.StaleEntryError do
 
   def exception(opts) do
     action = Keyword.fetch!(opts, :action)
-    struct = Keyword.fetch!(opts, :struct)
     changeset = Keyword.fetch!(opts, :changeset)
 
     msg = """
     attempted to #{action} a stale struct:
 
-    #{inspect struct}
+    #{inspect changeset.data}
     """
 
     %__MODULE__{message: msg, changeset: changeset}
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -254,11 +254,12 @@ defmodule Ecto.MigrationError do
 end
 
 defmodule Ecto.StaleEntryError do
-  defexception [:message, :struct]
+  defexception [:message, :changeset]
 
   def exception(opts) do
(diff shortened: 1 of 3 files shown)


josevalim:
Shouldn’t we be populating the struct field when we raise it then? :) that sounds ok to me. We can also raise the whole changeset instead of the struct, perhaps even better as it by definition has more information.

mathieuprog:
I think the struct *is* populated?
`raise Ecto.StaleEntryError, struct: user_changeset.data, action: action`
Or just no idea what you said :'D

Indeed we could raise the changeset, because that would include the changes, might be useful if we wanna have some more precise error messages (taking the example mentioned above "Guest Jose is not found" by a `fetch_field` call -> check changes, then data).

josevalim:
Sorry, I missed we were already passing it’s yeah, let’s migrate to changesets. :)

mathieuprog:
Should I kinda revert this commit (which command do you use?) or should I just correct it by adding a new commit? Ty

josevalim:
You can add on top, I will squash before merging!

mathieuprog:
Ok done

josevalim:
Let's make sure the field is set in one of the unit tests and this is ready to go!

mathieuprog:
Done

josevalim:

mathieuprog:
Just for the log, I mentioned the case of working with associations, but this is also useful when working with transaction, where it might not be clear which struct is stale (multiple update/delete operations).

Btw how do releases frequency work? When can I expect that change to be included in a new release?

josevalim:
The change will be included in the next release, which will be v3.7, likely happening in a month or so. :)

PR #3701, opened 2021-08-03
initial stab at creating dynamic aliases

zachdaniel:
I did some exploration on what would be required for this, and this is what I have so far. When I try to apply this in practice, I'm getting an error in ecto_sql on the following query (not a real query I used, just a sample to test the functionality):

```elixir
from post in "posts", 
  as: [:post], 
  select: post.title, 
  inner_lateral_join: comment in subquery(
    from comment in "comments", 
    where: comment.post_id == parent_as([:post]).id) 

** (RuntimeError) {:&, [], [0]} not found in parent sources: %{[:post] => 0}
```

I added the error message you see there as a fallback clause to this function:
https://github.com/elixir-ecto/ecto_sql/blob/e5663b8ce28717a68ff073bd645570a57075a314/lib/ecto/adapters/myxql/connection.ex#L1005

I don't fully understand the reason that its getting translated to `{:&, [], [0]}` here: https://github.com/elixir-ecto/ecto/commit/ce90e4eb1#diff-c4841538fead8249c07e55c917f52f1dd7bc64229e86ea281a81a22234c1313fR1615, but that seems to have something to do with it? I've confirmed that I don't get that error message using the master branches of ecto and ecto_sql, so it must have something to do with the contents of this PR as well.

For some more context, see the initial discussion: https://groups.google.com/g/elixir-ecto/c/yJrs2LHCFfA/m/gppCqkWkBQAJ?utm_medium=email&utm_source=footer

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -594,10 +594,15 @@ defmodule Ecto.Query.Builder do
     {:{}, [], [dot, [], []]}
   end
 
-  defp escape_field!({kind, _, [atom]}, field, _vars)
+  defp escape_field!({kind, _, [value]}, field, _vars)
        when kind in [:as, :parent_as] do
-    atom  = quoted_atom!(atom, "#{kind}/1")
-    as    = {:{}, [], [kind, [], [atom]]}
+    value =
+      case value do
+        {:^, _, [value]} ->
+          value
+        other -> other
+      end
+    as    = {:{}, [], [kind, [], [value]]}
     field = quoted_atom!(field, "field/2")
     dot   = {:{}, [], [:., [], [as, field]]}
     {:{}, [], [dot, [], []]}
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -600,7 +600,9 @@ defmodule Ecto.Query.Builder do
       case value do
         {:^, _, [value]} ->
           value
-        other -> other
+
+        other ->
+          quoted_atom!(other, "#{kind}/1")
       end
     as    = {:{}, [], [kind, [], [value]]}
     field = quoted_atom!(field, "field/2")
--- a/test/ecto/query/builder/join_test.exs
+++ b/test/ecto/query/builder/join_test.exs
@@ -137,13 +137,6 @@ defmodule Ecto.Query.Builder.JoinTest do
     end
   end
 
-  test "allows non-atom as" do
-    escape(quote do
-      as = "comments"
-      join("posts", :left, [p], c in "comments", on: true, as: ^as)
-    end, [], __ENV__)
-  end
-
   test "raises on non-string prefix" do
     assert_raise Ecto.Query.CompileError, ~r/`prefix` must be a compile time string/, fn ->
       escape(quote do
--- a/test/ecto/query/builder/join_test.exs
+++ b/test/ecto/query/builder/join_test.exs
@@ -137,18 +137,11 @@ defmodule Ecto.Query.Builder.JoinTest do
     end
   end
 
-  test "raises on non-atom as" do
-    assert_raise Ecto.Query.CompileError, ~r/`as` must be a compile time atom/, fn ->
-      escape(quote do
-        join("posts", :left, [p], c in "comments", on: true, as: "string")
-      end, [], __ENV__)
-    end
-
-    assert_raise Ecto.Query.CompileError, ~r/`as` must be a compile time atom/, fn ->
-      escape(quote do
-        join("posts", :left, [p], c in "comments", on: true, as: atom)
-      end, [], __ENV__)
-    end
+  test "allows non-atom as" do
+    escape(quote do
+      as = "comments"
+      join("posts", :left, [p], c in "comments", on: true, as: ^as)
+    end, [], __ENV__)
   end
 
   test "raises on non-string prefix" do
--- a/test/ecto/query/planner_test.exs
+++ b/test/ecto/query/planner_test.exs
@@ -851,19 +851,19 @@ defmodule Ecto.Query.PlannerTest do
   end
 
   test "normalize: creating dynamic bindings with as" do
-    as = [:posts]
+    as = {:posts}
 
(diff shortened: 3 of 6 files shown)


lib/ecto/query/builder.ex
```diff
@@ -584,10 +584,15 @@ defmodule Ecto.Query.Builder do
     {:{}, [], [dot, [], []]}
   end
 
-  defp escape_field!({kind, _, [atom]}, field, _vars)
+  defp escape_field!({kind, _, [value]}, field, _vars)
        when kind in [:as, :parent_as] do
-    atom  = quoted_atom!(atom, "#{kind}/1")
-    as    = {:{}, [], [kind, [], [atom]]}
+    value =
```
zachdaniel, line 589: This is just a temporary measure, unsure how it ought to be handled.

I can see why we might want to keep people "on the rails", e.g the only way that you can add a binding that isn't an atom would be if you used `^value` or something like that. Makes no difference to me. We could also just remove the necessity to pin anything in `as/1` and `parent_as/1`, but that would break things on upgrade.

test/ecto/query/builder/join_test.exs
```diff
@@ -137,20 +137,6 @@ defmodule Ecto.Query.Builder.JoinTest do
     end
   end
 
-  test "raises on non-atom as" do
```
zachdaniel, line 140: I tried just changing this test to use a string binding, but got an error about a malformed join. But it works in my manual tests so I'm not sure whats up. Didn't want to invest too much time into it w/o knowing if I'm moving in the generally correct direction.

josevalim:
It looks good to me. However, note the builder needs to either match on an atom, on an interpolation or raise. :)

zachdaniel:
Will get back to this within the next week or two, apologies for leaving it sit, just been super busy lately.

zachdaniel:
@josevalim so I pushed the change you mentioned, but I'm still getting the error described in the initial PR description `** (RuntimeError) {:&, [], [0]} not found in parent sources: %{[:post] => 0}` even though the ecto tests are passing. I'll investigate, but let me know if you have any ideas of where I should look 🙇🏻

kernel-io:
What's the status of this PR, this would be very useful for my work also, are there any pointers regarding what work remains to be done @josevalim?

josevalim:
I am not sure. I won't have time to look into this to provide guidance, unfortunately, as I will focus the time I have for Ecto on bug fixes and handle already completed PRs.

kernel-io:
I'll try set aside some time to have a look myself, thankyou 🙏🏾

zachdaniel:
I'm still investigating the test that was failing, and it appears we are getting into this clause in `escape`
```elixir
  def escape(join, vars, env) do
    case Macro.expand(join, env) do
      ^join ->
        Builder.error! "malformed join `#{Macro.to_string(join)}` in query expression"
      join ->
        escape(join, vars, env)
    end
  end
```
This is the test in question, which is a copied version of the test I removed from before:
```elixir
  test "does not raise on non-compile-time-atom as" do
    escape(quote do
      join("posts", :left, [p], c in "comments", on: true, as: "string")
    end, [], __ENV__)
  end
```

I'm still examining the structures to see why `Macro.expand(join, env) !== join`, but I'm also curious as to why exactly that is something that must be true.

zachdaniel:
Actually looks like I misread it. It is saying that `Macro.expand(join, env)` must *not* equal `join`. Still digging to try and understand why that needs to be the case.

zachdaniel:
Okay, I think that is a red herring. That `Macro.escape` clause is a last ditch case it appears. The new, non-atom aliases are just not being handled in a prior `escape` clause, or something along those lines. I should be able to figure this out.

zachdaniel:
Okay, so the original test fails even when I use atoms. I believe that this is unimportant/not relevant to the change I'm making. Specifically, the old test was just asserting that a certain error was raised, and now that it is not, some later error is being raised. There are other tests that validate the behavior I've added, and no tests aside from the tests that assert that a (now non-existent) error was raised have been removed. I think this is ready for review.

zachdaniel:
@josevalim just pinging you to let you know that I don't think that this PR needs anything aside from review (the issue I originally discussed appears to be a non-issue).

test/ecto/query/planner_test.exs
```diff
@@ -848,11 +848,23 @@ defmodule Ecto.Query.PlannerTest do
     assert_raise Ecto.QueryError, ~r/could not find named binding `as\(:posts\)`/, fn ->
       from(Post, where: as(^as).code == ^123) |> normalize()
     end
+  end
+
+  test "normalize: creating dynamic bindings with as" do
+    as = [:posts]
```
josevalim, line 854: Use a `{:posts}` in the test because it is invalid AST and it may help find bugs. :)

test/ecto/query/planner_test.exs
```diff
@@ -848,11 +848,23 @@ defmodule Ecto.Query.PlannerTest do
...
+
+    query = from(Post, as: as, where: as(^as).code == ^123) |> normalize()
+    assert Macro.to_string(hd(query.wheres).expr) == "&0.code() == ^0"
 
-    assert_raise Ecto.Query.CompileError, ~r/expected atom in as\/1/, fn ->
-      as = "posts"
+    query = from(Post, as: as, where: field(as(^as), :code) == ^123) |> normalize()
```
josevalim, line 859: I am thinking we should also use `as: ^as`. Seeing the "naked variable" is not very clear on intent because naked variables in queries usually refer to a table? WDYT?

zachdaniel: Yeah, that makes a lot of sense, Ill make that change.

test/ecto/query/planner_test.exs
zachdaniel: Will do 👌

zachdaniel:
@josevalim Thanks for the feedback, I've applied it to the PR. Agreed that using `^` is clearer as well.

zachdaniel:
Ah, looks like some issues were found in the integration test suite

josevalim:

PR #3707, opened 2021-08-13
Use semantic comparison in validations which work with collections

### LINKED ISSUE #3706: validate_inclusion should use Ecto.Type.equal?
### Environment

* Elixir version (elixir -v): 1.12.2
* Ecto version (mix deps):  3.6.2

### Current behavior
Complex data structures like Decimal or DateTime are not compared using appropriate interfaces.
https://github.com/elixir-ecto/ecto/blob/9e2d6e05aad145ef37a57a213c11c7be8257ca03/lib/ecto/changeset.ex#L2003

Here is an example with Decimal:
```elixir
> {%{}, %{value: :decimal}} |> Ecto.Changeset.cast(%{value: 0}, [:value]) |> validate_inclusion(:value, Enum.map([0.0, 0.1, 0.2], &Decimal.from_float/1))    
#Ecto.Changeset<
  action: nil,
  changes: %{value: #Decimal<0>},
  errors: [
    value: {"is invalid",
     [
       validation: :inclusion,
       enum: [#Decimal<0.0>, #Decimal<0.1>, #Decimal<0.2>]
     ]}
  ],
  data: %{},
  valid?: false
>
```

### Expected behavior
Values in Ecto.Changeset.validate_inclusion function must be compared using Ecto.Type.equal?

fuelen:
closes #3706

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2028,7 +2028,9 @@ defmodule Ecto.Changeset do
   @spec validate_subset(t, atom, Enum.t, Keyword.t) :: t
   def validate_subset(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:subset, data}, fn _, value ->
-      case Enum.any?(value, fn(x) -> not(x in data) end) do
+      {:array, element_type} = Map.fetch!(changeset.types, field)
+
+      case Enum.any?(value, fn element -> not Ecto.Type.include?(element_type, element, data) end) do
         true -> [{field, {message(opts, "has an invalid entry"), [validation: :subset, enum: data]}}]
         false -> []
       end
@@ -2050,7 +2052,9 @@ defmodule Ecto.Changeset do
   @spec validate_exclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_exclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:exclusion, data}, fn _, value ->
-      if value in data, do:
+      type = Map.fetch!(changeset.types, field)
+
+      if Ecto.Type.include?(type, value, data), do:
         [{field, {message(opts, "is reserved"), [validation: :exclusion, enum: data]}}], else: []
     end
   end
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset.ex
```diff
@@ -2000,7 +2000,9 @@ defmodule Ecto.Changeset do
   @spec validate_inclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_inclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:inclusion, data}, fn _, value ->
-      if value in data,
+      type = Map.get(changeset.types, field)
+
+      if Enum.any?(data, &Ecto.Type.equal?(type, value, &1)),
```
v0idpwn, line 2005: This loses the optimization for ranges, but if you're thinking about ranges you probably should be using validate_number anyway?

v0idpwn (approved):

lib/ecto/changeset.ex
fuelen: I was thinking about this too.
If you need numeric ranges, then validate_number provides better errors.

josevalim: Let's check the type. If the type is not an integer, float, or string, we use .equal?, otherwise `in` is fine. WDYT?

josevalim: Actually, we can move this to `Ecto.Type` and do something like this:

```elixir
  def include?(type, term, collection) do
    if fun = equal_fun(type) do
      Enum.any?(collection, &fun.(term, &1))
    else
      term in collection
    end
  end
```

It will remain efficient for base types and handle all equal types accordingly.

fuelen: Pushed an updated version

lib/ecto/changeset.ex
```diff
@@ -2000,7 +2000,9 @@ defmodule Ecto.Changeset do
   @spec validate_inclusion(t, atom, Enum.t, Keyword.t) :: t
   def validate_inclusion(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:inclusion, data}, fn _, value ->
-      if value in data,
+      type = Map.get(changeset.types, field)
```
josevalim, line 2003:
```suggestion
      type = Map.fetch!(changeset.types, field)
```

josevalim:
This looks great to me, thanks! I will merge it soon but please note we have similar issue in validate_exclusion and validate_subset. Do you want to update the PR to handle those too? :)

fuelen:
@josevalim updated a PR :)

josevalim:

PR #3710, opened 2021-08-17
Improve error message when nil is passed to `Ecto.assoc/2`

fuelen:
`nil` could be passed with association `post has_one author`, so the code
```elixir
post
|> Repo.preload(:author)
|> Map.fetch!(:author)
|> Ecto.assoc(:posts)
```
raises an error
```
** (ArgumentError) cannot retrieve association :posts for empty list
```
but `:author` is nil


josevalim:

PR #3711, opened 2021-08-17
fix: Consistent definition of `schema_or_source`

isaacsanders:
`Ecto.Multi` was using an inconsistent definition of `schema_or_source` from `c:Ecto.Repo.insert_all/3`.

https://github.com/elixir-ecto/ecto/blob/cc75abde90dbeea29e51611f530e2de768d48c69/lib/ecto/repo.ex#L1307

This PR would fix that inconsistency.


josevalim:

PR #3717, opened 2021-08-20
Add source to ecto schema metadata as in the code

kayuapi:
https://github.com/elixir-ecto/ecto/blob/11a46ffad9dfe42babe3a65fa9b80a760f948654/lib/ecto/schema/metadata.ex#L55


josevalim:

PR #3718, opened 2021-08-20
Fix typespecs in Ecto.Enum

timothyvanderaerden:
It seems that the specs for the functions `values`, `dump_values` and `mappings` in the Enum module are not correct. This causes Dialyzer to report `no_return` errors on these functions. Since these functions call `schema.__changeset__()` the typespec for the schema argument should be `module` instead of `Ecto.Schema.t`.


josevalim:

PR #3725, opened 2021-08-22
Make `Ecto.Embedded` public and describe struct fields

### LINKED ISSUE #3719: Documentation for `Ecto.Embedded`
I found, that `Ecto.Embedded` is not documented https://github.com/elixir-ecto/ecto/blob/144ea32a6da43f2f2c30af29b642a244e74d5767/lib/ecto/embedded.ex#L2, while documentation for `Ecto.Association.*` is present.
Documentation was removed explicitly in this commit https://github.com/elixir-ecto/ecto/commit/ac98aa555edecf81c6897d2a231e8c9ef57d524e, when the feature was experimental.

Is this still relevant? Could we add documentation back to the module and make it public?

fuelen:
closes #3719

`@type t` was removed for consistency with `Ecto.Association.*`.
`unique` field isn't documented, as it is absent in docs for `Ecto.Association.*` too.

There is an open question from my side: to which group in docs should this module be moved?

### DIFF
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -3,6 +3,7 @@ defmodule Ecto.Embedded do
   The embedding struct for `embeds_one` and `embeds_many`.
 
   Its fields are:
+
     * `cardinality` - The association cardinality
     * `field` - The name of the association field on the schema
     * `owner` - The schema where the association was defined
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -1,18 +1,21 @@
 defmodule Ecto.Embedded do
-  @moduledoc false
+  @moduledoc """
+  The embedding struct for `embeds_one` and `embeds_many`.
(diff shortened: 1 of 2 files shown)


lib/ecto/embedded.ex
```diff
@@ -1,19 +1,22 @@
 defmodule Ecto.Embedded do
-  @moduledoc false
+  @moduledoc """
+  The embedding struct for `embeds_one` and `embeds_many`.
+
+  Its fields are:
```
josevalim, line 5:
```suggestion
  Its fields are:

```

josevalim:
@fuelen looks great! Can you also please change the mix.exs? We should list the Ecto.Embedded under "Associations structs". Perhaps rename it to "Relation structs" but keeping the current name is also fine.

fuelen:
@josevalim done

josevalim:

PR #3726, opened 2021-08-23
Update docs for Ecto.Enum module

ihorkatkov:
## What?
Updates docs for `Ecto.Enum` module

## Why?
In the recent release, `mappings/2` and `dump_values/2` were introduced as well as the return type of `values/2` was changed from `Keyword.t` to `[atom()]` (**which is breaking change**) . The PR corrects an example of mappings function inside docs of the module.

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -37,7 +37,7 @@ defmodule Ecto.Enum do
   by an atom in the list will be invalid.
 
   The helper function `mappings/2` returns the mappings for a given schema and
-  field, which can be used in places like form drop-downs. For example, given 
+  field, which can be used in places like form drop-downs. For example, given
   the following schema:
 
       defmodule EnumSchema do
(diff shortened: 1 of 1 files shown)


lib/ecto/enum.ex
```diff
@@ -185,7 +185,8 @@ defmodule Ecto.Enum do
     try do
       schema.__changeset__()
     rescue
-      _ in UndefinedFunctionError -> raise ArgumentError, "#{inspect schema} is not an Ecto schema"
+      _ in UndefinedFunctionError ->
```
ihorkatkov, line 188: Formatter fix

josevalim:
> were introduced as well as the return type of values/2 was changed from Keyword.t to [atom()]

Just one correction, it was always a list of atoms:

https://github.com/elixir-ecto/ecto/commit/9b4c7632677ebbd18cc8852be7a56550c091ca71

https://github.com/elixir-ecto/ecto/blob/v3.6.2/test/ecto/enum_test.exs#L333-L339

But the docs were indeed wrong. Thanks!

josevalim:

PR #3731, opened 2021-08-27
helpers for updating nested changesets

woylie:
This is a proposal for the addition of some helper functions for manipulating nested changesets.

One common use case for changesets is as `FormData` for forms in Phoenix. With LiveView, we can easily build complex, dynamic, nested forms, but Ecto doesn't expose functionality for easily manipulating nested associations at the moment. I'm proposing the addition of four functions to `Ecto.Changeset`.

- `append_at/3`: appends a new item to an array field or `has_many` association at the given path
- `insert_at/3`: inserts a new item at a path/position
- `update_at/3`: updates a change or field at the given path
- `delete_at/2`: deletes an item at a given path/position

The path argument in all these functions would be a list of atoms (to reference fields) and integers (to reference the index in a list). For example, `append_at(changeset, [:posts, 4, :comments], %Comment{})` would append a comment at `root.posts[4].comments`. `delete_at(changeset, [:posts, 3])` would delete the post with index 3.

I sketched out a rough implementation for `append_at/3` in this PR. Before I flesh this out to handle all cases and errors, I'd like to hear someone's thoughts on this effort.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1468,11 +1468,14 @@ defmodule Ecto.Changeset do
     end
   end
 
-  @spec append_at(t, [atom | non_neg_integer], any) :: t
-  def append_at(changeset, field, value) when is_atom(field),
-    do: append_at(changeset, [field], value)
+  @spec append_at(t, [atom | non_neg_integer] | atom, any) :: t
+  def append_at(changeset, path, value),
+    do: nested_update(changeset, path, value, :append)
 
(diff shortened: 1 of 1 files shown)


woylie:
Hi @josevalim, just pinging you to let you know that a first version has been released: https://github.com/woylie/ecto_nested_changeset

There is an example Phoenix application in the `/example` folder.

josevalim:
Sweet, thanks for sharing!

PR #3742, opened 2021-09-08
ensure type is fetched for field in validate inclusion, subset, and exclusion

mpinkston:
This PR fixes an error when custom field types are used when calling validations for inclusion, subset, and exclusion.

Previously, a mismatch error would be raised (at least for validate_subset) as the module implementing the type behaviour would be returned rather than the type it defines.


josevalim:

dhedlund:
Is there an estimate on when this change will make it into a release?  :heart:

PR #3746, opened 2021-09-11
Mention automatic transaction wrapping in Repo docs

v0idpwn:
`Repo.insert/1`, `Repo.update/2` and `Repo.delete/1` may do many
database operations. When they need to do so, these are wrapped in
a transaction.


v0idpwn:
I've seen some confusion related to this, so I think it would be good having this information in the documentation

josevalim:

Issue #3747, opened 2021-09-15
Ecto.Schema.validate_default!/2 happens at compile time


tomciopp:
### Environment

* Elixir version (elixir -v): 1.12.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): 13.4
* Ecto version (mix deps): 3.7.0
* Database adapter and version (mix deps): Postgrex 0.15.10
* Operating system: OSX 11.6

### Current behavior

Include code samples, errors and stacktraces if appropriate.

This issue was brought up in #3483 and the code was merged in #3487 but there is a corner case which I'm running into that causes my application to fail to compile when getting ready for production. We use a library that encrypts values before they are dumped to the DB. This library relies on a secret being available at runtime in order to process the custom type's `dump` function. Since the application does not have this value available at compile time, the code fails to compile and we cannot release new versions.

### Expected behavior

I'm not 100% sure that this validation is necessary, but I'm sure @v0idpwn can give a legitimate reason as to why this fence should not be removed. When thinking about workarounds, I think the easiest option would be to create an escape valve within the validation if you have a legitimate reason to not check the default type. Something like `skip_default_validation: true` option should suffice. This way the previous behavior can be emulated and it is on the developer if they make a mistake. Let me know if you think of an alternative method to handle this issue or whether the added option field works. If that is good I can put together a PR and do some testing on my end to make sure everything is in working order.

josevalim:
Hi @tomciopp! Quick question: what is the default value in your case? Is it nil by any chance?

If not, maybe one option is to rescue if we can't validate the type and then warn instead of failing to compile. Would you be happy with this approach?

tomciopp:
@josevalim It's usually an integer, sometimes it's a string. I could live in that world, although I'm not particularly in love with the idea of warnings that I know I will never be able to address. When I see a warning, my general response is to fix the code that has been deprecated. Ideally I want to live warning free, since a warning is a hint that I'm doing something wrong. In this case I want to tell ecto, "nah man it's cool don't worry about it, I got this"

josevalim:
So I think skip_default_validation sounds good. Can you please submit a PR?

tomciopp:
@josevalim Yeah sure you should expect something in the next 24 hrs.

v0idpwn:
Hello! I'll be the devils advocate here (just because you asked in advance :))

[...]

PR #3749, opened 2021-09-16
feat(Ecto.Schema): Add option to skip validations for default values

tomciopp:
@josevalim Addresses Issue #3747. Let me know if you need any additional changes or want to add documentation for this feature.

### DIFF
--- a/test/ecto/schema_test.exs
+++ b/test/ecto/schema_test.exs
@@ -418,6 +418,17 @@ defmodule Ecto.SchemaTest do
     end
   end
 
+  test "skipping validations on invalid types" do 
+    assert(
+      defmodule SchemaSkipValidationsDefault do
+        use Ecto.Schema
+
+        schema "invalid_default" do
+          field :count, :integer, default: "1", skip_default_validation: true
+        end
+      end, ~s/value "1" can be set for integer due to overridden validation/)
+  end
+
   test "invalid field type" do
     assert_raise ArgumentError, "invalid or unknown type {:apa} for field :name", fn ->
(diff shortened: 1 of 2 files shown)


v0idpwn:
Could you mention the new option around here? https://github.com/elixir-ecto/ecto/blob/0e8954d91d4a4520f96cecc28cc7786037ce163b/lib/ecto/schema.ex#L647

It would also be good if `:default` documentation specified that it validates the value in compile time.

test/ecto/schema_test.exs
```diff
@@ -418,6 +418,17 @@ defmodule Ecto.SchemaTest do
...
 
+  test "skipping validations on invalid types" do 
+    assert(
+      defmodule SchemaSkipValidationsDefault do
+        use Ecto.Schema
+
+        schema "invalid_default" do
+          field :count, :integer, default: "1", skip_default_validation: true
+        end
+      end, ~s/value "1" can be set for integer due to overridden validation/)
```
josevalim, line 429:
```suggestion
    defmodule SchemaSkipValidationsDefault do
      use Ecto.Schema

      schema "invalid_default" do
        # Without skip_default_validation this would fail to compile
        field :count, :integer, default: "1", skip_default_validation: true
      end
    end
```

josevalim:
LGTM, thanks! We should definitely document it. Perhaps list it right after `:default`, so it addresses @v0idpwn's concerns?

tomciopp:
@josevalim I put it at the bottom since it is likely going to be the least used option for a field.

josevalim:

Issue #3748, opened 2021-09-16
Embedded Schemas not validating decimal scale


ah0y:
### Environment

* Elixir version (elixir -v): Elixir 1.12.0 (compiled with Erlang/OTP 22)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): N/A
* Ecto version (mix deps): 3.7
* Database adapter and version (mix deps): N/A
* Operating system: Mac OS

### Current behavior

I have an embedded schema which looks something like this:

```elixir
  embedded_schema do
    field(:id, :id)
    field(:date, :date)
    field(:tx_type, :string)
    field(:price, :decimal, scale: 2)
    field(:quantity, :decimal, scale: 8)
  end
 ```
 
 and when passing in either `"10.123"` for price or `"10.123456789"` for quantity (as either strings or actual Decimals) to my schemas changeset, nothing gets raised as invalid.

### Expected behavior

Decimals that go over a fields defined scale should throw some sort of changeset error. Currently I'm working around this by defining my own parametrized type but it would be cool if ecto changesets could do this validation natively.

v0idpwn:
I think that, as types don't support options (only parameterized types do), it is not possible to have this behaviour with the decimal type.

ah0y:
> I think that, as types don't support options (only parameterized types do), it is not possible to have this behaviour with the decimal type.

oh you are totally right. I just noticed that I can pass random options to string fields and it not raise an error. is that expected as well?

josevalim:
@ah0y that's likely the root cause. I think we were use to validate fields but we removed the validation once we add parameterized types. Perhaps we should keep the validation and skip it only if a parameterized type is given.

v0idpwn:
Shall we close this in favor of #3750?
I think it makes intuitive that you can't define type-specific options for non-parameterized types.

PR #3750, opened 2021-09-16
Validate options given to schema fields

v0idpwn:
Related to #3748.

I'd like to validate options specifically for field type (relation, normal field, etc), will try to improve it but opening for reference

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -486,6 +486,23 @@ defmodule Ecto.Schema do
     end
   end
 
+  @field_opts [
+    :default,
+    :source,
+    :autogenerate,
+    :read_after_writes, 
+    :virtual, 
+    :primary_key,
+    :load_in_query,
+    :redact,
+    :foreign_key,
+    :on_replace, 
+    :defaults,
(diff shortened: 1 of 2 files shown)


lib/ecto/schema.ex
```diff
@@ -1874,17 +1891,40 @@ defmodule Ecto.Schema do
...
     define_field(mod, name, type, opts)
   end
 
+  defp validate_field_opts!(name, type, opts) do
+    validate_opts = fn opts ->
+      Enum.each(opts, fn {k, _v} -> 
+          unless k in @field_opts, do:
+            raise ArgumentError, "invalid option #{inspect(k)} given for field #{inspect(name)}"
+        end)
+     end
```
josevalim, line 1908:
```suggestion
      end
```

josevalim (approved):

v0idpwn:
I'm changing this to draft because, while this works, I'm finding a better solution.

v0idpwn:
I refactored so I could use `check_options!` which is used by relations already

josevalim (approved):

PR #3753, opened 2021-09-22
fix validation entry for unsafe_validate_unique/4 (fixes #3752)

### LINKED ISSUE #3752: unsafe_validate_unique/4 adds an incorrect validation entry
My code fails as I assumed that keys returned by `validations/1` are field names, as mentioned here:
https://hexdocs.pm/ecto/Ecto.Changeset.html#validations/1
> The keys in the list are the names of fields, and the values are a validation associated with the field. A field may occur multiple times in the list.

However this is not always the case, for example:
```elixir
[
  unsafe_unique: [:email],
  email: {:length, [max: 160]},
  email: {:format, ~r/^[^\s]+@[^\s]+$/}
]
```

`unsafe_unique` is not a field name.

Should it rather be `email: :unsafe_unique` ?

mathieuprog:
See discussion at https://github.com/elixir-ecto/ecto/issues/3752

Please can you run the test because I'm full of "(column 24, code point U+000D)t:" errors 😥

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1868,7 +1868,7 @@ defmodule Ecto.Changeset do
         %Ecto.Changeset{} ->
           raise ArgumentError, "unsafe_validate_unique/4 does not work with schemaless changesets"
       end
-    changeset = %{changeset | validations: [{:unsafe_unique, fields} | validations]}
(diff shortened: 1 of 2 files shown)


v0idpwn (approved):

josevalim:

PR #3754, opened 2021-09-29
Suggest alternatives to inexistent fields in queries

v0idpwn:
Looks for similar fields and tries to suggest them when raising about
invalid query.

### DIFF
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -28,7 +28,7 @@ defmodule Ecto.QueryError do
   def exception(opts) do
     message = Keyword.fetch!(opts, :message)
     query   = Keyword.fetch!(opts, :query)
-    hint    = Keyword.fetch(opts, :hint)
+    hint    = Keyword.get(opts, :hint)
 
     message = """
     #{message} in query:
@@ -48,9 +48,10 @@ defmodule Ecto.QueryError do
       end
 
     message =
-      case hint do
-        {:ok, text} -> message <> "\n" <> text <> "\n"
-        _ -> message
+      if hint do
+        message <> "\n" <> hint <> "\n"
+      else
+        message
       end
 
     %__MODULE__{message: message}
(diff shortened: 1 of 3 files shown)


lib/ecto/query/planner.ex
```diff
@@ -1737,7 +1737,31 @@ defmodule Ecto.Query.Planner do
...
+    schema.__schema__(:fields)
+    |> Enum.map(fn field ->
+      {field, String.jaro_distance(input_string, Atom.to_string(field))}
+    end)
+    |> Enum.max(fn {_f1, s1}, {_f2, s2} -> s1 >= s2 end)
+    |> case do
+      {field, score} when score >= 0.8 ->
+        field
+      _ ->
+        nil
```
josevalim, line 1764: What we do in Elixir is to show multiple suggestions (up to 5) above a threshold:

```elixir
      for({key, val} <- candidates,
          dist = String.jaro_distance(base, Atom.to_string(key)),
          dist >= @function_threshold,
          do: {dist, key, val})
      |> Enum.sort(&(elem(&1, 0) >= elem(&2, 0)))
      |> Enum.take(@max_suggestions)
      |> Enum.sort(&(elem(&1, 1) <= elem(&2, 1)))
```

Should we do the same here? Your call. :)

josevalim (approved):
Great idea!

lib/ecto/query/planner.ex
v0idpwn: This sound better! I'll try :)

v0idpwn:
Thanks for the suggestion, I think its a lot better now. Let me know if there's anything else to address :)

lib/ecto/exceptions.ex
```diff
@@ -28,7 +28,7 @@ defmodule Ecto.QueryError do
   def exception(opts) do
     message = Keyword.fetch!(opts, :message)
     query   = Keyword.fetch!(opts, :query)
-    hint    = Keyword.fetch(opts, :hint)
+    hint    = Keyword.get(opts, :hint)
```
v0idpwn, line 31: I changed it here because I may pass `nil`, and if I do so, I wish it to be ignored. This won't happen with fetch + case, but will with get + if

josevalim (approved):

PR #3757, opened 2021-10-05
Add Ecto.ConstraintError to Repo.delete/2 docs

tfwright:
Just ran into this and was a bit confused because I thought the docs implied an error tuple would be returned in the relevant case. I also tend to think referencing how to use a changeset with `no_assoc_constraint` here as well to avoid raising the error might be a good idea. If others agree I would be happy to add. But for now listing this error along with the others seems like the minimal positive change here.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1622,9 +1622,10 @@ defmodule Ecto.Repo do
 
   If the struct has no primary key, `Ecto.NoPrimaryKeyFieldError`
   will be raised. If the struct has been removed from db prior to
-  call, `Ecto.StaleEntryError` will be raised. If more than one 
-  database operation is required, they're automatically wrapped 
-  in a transaction.
+  call, `Ecto.StaleEntryError` will be raised. If removing the struct 
+  from the db will violate an existing constraint, `Ecto.ConstraintError` 
+  will be raised. If more than one database operation is required, 
+  they're automatically wrapped in a transaction.
 
   It returns `{:ok, struct}` if the struct has been successfully
   deleted or `{:error, changeset}` if there was a validation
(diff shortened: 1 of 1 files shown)


josevalim:
I think we can mention that you can use a changeset with `foreign_key_constraint` and `no_assoc_constraint` to convert the exceptions into changeset errors. It is your call. :)

tfwright:
Added! Rather than listing all the possibilities, I just included a reference to the Changeset module.

lib/ecto/repo.ex
```diff
@@ -1621,14 +1621,16 @@ defmodule Ecto.Repo do
   Deletes a struct using its primary key.
 
   If the struct has no primary key, `Ecto.NoPrimaryKeyFieldError`
-  will be raised. If the struct has been removed from db prior to
-  call, `Ecto.StaleEntryError` will be raised. If more than one 
-  database operation is required, they're automatically wrapped 
-  in a transaction.
+  will be raised. If the struct has been removed prior to call,
```
josevalim, line 1624:
```suggestion
  will be raised. If the struct has been removed prior to the call,
```

lib/ecto/repo.ex
```diff
@@ -1621,14 +1621,16 @@ defmodule Ecto.Repo do
...
+  `Ecto.StaleEntryError` will be raised. If more than one database
+  operation is required, they're automatically wrapped in a transaction.
 
   It returns `{:ok, struct}` if the struct has been successfully
   deleted or `{:error, changeset}` if there was a validation
-  or a known constraint error.
+  or a known constraint error. By default, constraint errors will
+  raise the `Ecto.ConstraintError` exception, unless a changeset is
+  given as the first argument with the relevant constraints declared
+  in it (see `Ecto.Changeset`).
```
josevalim, line 1633: I moved all constraint discussion to this part so there is less back and forth. :)

josevalim:

PR #3760, opened 2021-10-09
Replace http with https for URLs

kianmeng:


mix.exs
```diff
@@ -54,7 +54,7 @@ defmodule Ecto.MixProject do
     [
       main: "Ecto",
       source_ref: "v#{@version}",
-      canonical: "http://hexdocs.pm/ecto",
+      canonical: "https://hexdocs.pm/ecto",
```
josevalim, line 57:
```suggestion
```

mix.exs
```diff
@@ -54,7 +54,6 @@ defmodule Ecto.MixProject do
     [
       main: "Ecto",
       source_ref: "v#{@version}",
-      canonical: "http://hexdocs.pm/ecto",
```
josevalim, line 57: I removed it because Hex sets it by default. :)

PR #3764, opened 2021-10-21
Fix regression in validating inclusion of custom types

v0idpwn:
Reported here: https://github.com/elixir-ecto/ecto/pull/3742#issuecomment-948320014

This PR:
- reverts #3742
- adds a regression test for the introduced bug
- adds a note in `validate_subset` documentation about its usage

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2023,6 +2023,8 @@ defmodule Ecto.Changeset do
   If you need to validate if a single value is inside the given enumerable,
   you should use `validate_inclusion/4` instead.
 
+  Type of the field must be array.
+
   ## Options
 
     * `:message` - the message on failure, defaults to "has an invalid entry"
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Beautiful, just one tiny comment!

patrickbiermann:
Hi there, I'm pretty sure this set of PRs broke `validate_subset/3` with custom Ecto types with `type = {:array, _}`.

To reproduce:
```
defmodule StringListType do
  @moduledoc """
  An Ecto type that wraps {:array, :string}.
  """

  use Ecto.Type

  def type, do: {:array, :string}

  def cast(params) when is_list(params) do
    {:ok, Enum.reject(params, fn item -> item == "" end)}
  end

  def cast(_), do: :error

  def dump(params) when is_list(params) do
    {:ok, params}
  end

  def dump(_), do: :error

  def load(data) when is_list(data) do
    {:ok, data}
  end
end
```

And validate a field of this type by `validate_subset :string_list_field, ["foo"]`. Previous behavior was working as intended, now it crashes with `no match of right hand side value: StringListType` around https://github.com/elixir-ecto/ecto/pull/3764/files#diff-edb6c9aaeb40387eb81c6b238954c0b4d813876d18805c6ae00d7ccc4d78e3f1L2042

Is there are a reason forcing us to drop the `Ecto.Type.type()` step there? Otherwise I'm guessing this would be the root cause here.

PR #3769, opened 2021-11-07
Allow using type/2 with virtual fields

### LINKED ISSUE #3767: Cast Fragment to virtual field type
### Precheck

* Informed by Jose to post on here

### Environment

* Elixir version (elixir -v): 1.12.1
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgres 14
* Ecto version (mix deps): 3.6.1
* Operating system: Mac OS

### Current behavior

'm stuck at an issue. I have a virtual field that I setup as an Ecto.Enum, and I'm trying to make that value into an atom. I have a "friend_status" field, and if there's a friend request that's not been accepted, then it's pending, if it's accepted, then they are friends, and if no field, then they are basically strangers.

I have my case statement working great in a fragment, but it's returning the strings instead of loading the atom. Here's a sample:
```elixir
from(
...
select: %{ u | friend_status: fragment("CASE WHEN ? THEN 'stranger' ELSE 'pending' END", is_nil(f.id))}
)
```

I tried using Ecto.Query.API.type/2 function, but I'm not sure what to put it as, since when I tried just the column name, it doesn't work, it returns an error saying it's a virtual field.

Is there a way to make this work?

### Expected behavior

Somehow, I should be able to cast the string into an atom because it's using Ecto.Enum as the type.

v0idpwn:
This would close #3767, but I think some discussion must be done:

1. I'm adding `__schema__(:virtual_type, type)` to avoid breaking compatibility with existing places that may rely on `__schema__(:type, type)` not returning virtual fields types. Ecto's own codebase does that, and I think elsewhere it may happen, so its probably a bad idea to break this behaviour.
2. All information returned by `__schema__(:virtual_type, type)` and `__schema__(:virtual_fields)` is possible to get through `__changeset__`, yet its probably not fine to rely on this function from Planner.

If we agree on something I can create new tests and make improvements.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1181,7 +1181,7 @@ defmodule Ecto.Query.Planner do
 
   defp prewalk({:type, _, [arg, type]}, kind, query, expr, acc, adapter) do
     {arg, acc} = prewalk(arg, kind, query, expr, acc, adapter)
-    type = field_type!(kind, query, expr, type)
+    type = field_type!(kind, query, expr, type, true)
     {%Ecto.Query.Tagged{value: arg, tag: type, type: Ecto.Type.type(type)}, acc}
   end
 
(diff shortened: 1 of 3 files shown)


josevalim (approved):
Beautiful! Just remember to document the two new clauses under the `Ecto.Schema` moduledoc, in the Reflection section, and we are good to go!

Issue #2500, opened 2018-04-07
using `type/2` with `coalesce/2` is not a valid query expression


zachdaniel:
### Environment

* Elixir version (elixir -v): 1.6.3
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 9.6
* Ecto version (mix deps): 
* Database adapter and version (mix deps): master
* Operating system: MacOS

### Current behavior

```elixir
from row in "posts", select: type(coalesce(row.title, "title!"), :string)
```

```elixir

** (Ecto.Query.CompileError) `type(coalesce(row.title(), "title!"), :string)` is not a valid query expression.

* If you intended to call a database function, please check the documentation
  for Ecto.Query to see the supported database expressions

* If you intended to call an Elixir function or introduce a value,
  you need to explicitly interpolate it with ^

    (ecto) expanding macro: Ecto.Query.select/3
    iex:13: (file)
    (ecto) expanding macro: Ecto.Query.from/2
    iex:13: (file)
```

### Expected behavior

I'll look into it an open a PR w/ a fix shortly if possible, just posting the issue in case anyone knows off of the top of their head why it doesn't work :)

zachdaniel:
It looks like `type` is factored out of the query using `prewalk`?

josevalim:
@zachdaniel we support type only on a handful of expressions. Is there any reason why we want to support it on coalesce? I would rather not support it.

josevalim:
Closing in favor of the PR.

coop:
I hate to dig up the past but the PR wasn't linked and I've had trouble finding it. I've got a situation where I would like to do:

```
from m in query,
  join: a in assoc(m, :account),
  select_merge: %{current_offset_amount: type(coalesce(a.balance, 0), Money.Ecto.Amount.Type)}
```

which doesn't work. Is there a different way to write the query? /cc @zachdaniel

josevalim:
Did you try latest Ecto? Otherwise you will have to write that as a fragment:

     type(fragment("...", a.balance), Money.Ecto.Amount.Type)}

coop:
It does not work with the latest Ecto (3.5.7):

```
== Compilation error in file lib/plutus/properties/mortgage.ex ==
** (Ecto.Query.CompileError) the first argument of type/2 must be one of:

  * interpolations, such as ^value
  * fields, such as p.foo or field(p)
  * fragments, such fragment("foo(?)", value)
[...]

PR #3782, opened 2021-11-23
Allow coalesce in type/2

LostKobrakai:
There don't seem to be (unit) tests around functions/expressions within type/2. I've confirmed it working on a local project.

For context see https://github.com/elixir-ecto/ecto/issues/2500

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -95,6 +95,10 @@ defmodule Ecto.Query.Builder do
     escape_with_type(expr, type, params_acc, vars, env)
   end
 
+  def escape({:type, _, [{:coalesce, _, [_ | _]} = expr, type]}, _type, params_acc, vars, env) do
+    escape_with_type(expr, type, params_acc, vars, env)
+  end
+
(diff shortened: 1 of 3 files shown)


josevalim:
I believe we have integration tests for this: integration_test/cases/type.exs.

lib/ecto/query/builder.ex
```diff
@@ -124,10 +128,11 @@ defmodule Ecto.Query.Builder do
         the first argument of type/2 must be one of:
 
           * interpolations, such as ^value
-          * fields, such as p.foo or field(p)
+          * fields, such as p.foo or field(p, :foo)
           * fragments, such fragment("foo(?)", value)
           * an arithmetic expression (+, -, *, /)
           * an aggregation or window expression (avg, count, min, max, sum, over, filter)
+          * an conditional expression (coalesce)
```
v0idpwn, line 135:
```suggestion
          * a conditional expression (coalesce)
```

v0idpwn (approved):

v0idpwn:
Looks great! I think an example could be added on Ecto.Query.API entry for type/2

josevalim:

Issue #3783, opened 2021-11-23
Interpolation of `dynamic` expression in `order_by`


fuelen:
### Environment

* Elixir version (elixir -v): 1.12
* Ecto version (mix deps): 3.7.1

### Current behavior
usage of `dynamic` as a part of`order_by` causes the following compilation error
```
"users" |> order_by(asc: ^dynamic([_], true))  
** (ArgumentError) expected a field as an atom in `order_by`, got: `dynamic([_], true)`
    (ecto 3.7.1) lib/ecto/query/builder/order_by.ex:118: Ecto.Query.Builder.OrderBy.field!/2
```

It is not obvious that the query should be written as follows
```
"users" |> order_by(^[asc: dynamic([_], true)])  
```

### Expected behavior
Support this type of interpolation or provide a helpful message how to deal with this

PR #3785, opened 2021-11-26
Better error message when interpolating dynamic exprs on order by

v0idpwn:
Closes https://github.com/elixir-ecto/ecto/issues/3783

### DIFF
--- a/lib/ecto/query/builder/order_by.ex
+++ b/lib/ecto/query/builder/order_by.ex
@@ -119,7 +119,7 @@ defmodule Ecto.Query.Builder.OrderBy do
     raise ArgumentError,
       "expected a field as an atom in `#{kind}`, got: `#{inspect dynamic_expression}`. " <>
       "To use dynamic expressions, you need to interpolate at root level, as in: " <>
-      "`^[asc: #{inspect dynamic_expression}, desc: :id]`"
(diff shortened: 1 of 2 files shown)


josevalim (approved):

lib/ecto/query/builder/order_by.ex
```diff
@@ -114,6 +114,14 @@ defmodule Ecto.Query.Builder.OrderBy do
   def field!(_kind, field) when is_atom(field) do
     to_field(field)
   end
+
+  def field!(kind, %Ecto.Query.DynamicExpr{} = dynamic_expression) do
+    raise ArgumentError,
+      "expected a field as an atom in `#{kind}`, got: `#{inspect dynamic_expression}`. " <>
+      "To use dynamic expressions, you need to interpolate at root level, as in: " <>
+      "`^[asc: #{inspect dynamic_expression}, desc: :id]`"
```
josevalim, line 122: Maybe we should just say dynamic as a var instead?

```suggestion
      "`^[asc: dynamic, desc: :id]`"
```

Otherwise people may try to interpolate the inspected value?

v0idpwn: Thats fair :)

Issue #3784, opened 2021-11-24
ParameterizedType in `type/2`


fuelen:
### Environment

* Elixir version (elixir -v): 1.12
* Database and version: PostgreSQL 11.8
* Ecto version (mix deps): 3.8.0-dev
* Database adapter and version (mix deps): "0.15.13"

Initially I was trying to do this:

```elixir
"users"
|> select([u], fragment("CASE WHEN ? THEN ? ELSE ? END", is_nil(u.joined_at), :pending, :active))
|> limit(1)
|> Repo.all()
```
<details>
  <summary> Error</summary>

_The error is a bit cryptic, is it possible to improve it?_
```
** (FunctionClauseError) no function clause matching in Ecto.Adapters.Postgres.Connection.expr/3    
    
    The following arguments were given to Ecto.Adapters.Postgres.Connection.expr/3:
    
        # 1
        :pending
    
        # 2
        {{[34, "users", 34], [117 | "0"], nil}, []}
    
        # 3
        #Ecto.Query<from u0 in "users", limit: 1,
         select: fragment("CASE WHEN ? THEN ? ELSE ? END", is_nil(u0.joined_at), :pending, :active)>

    (ecto_sql 3.7.1) lib/ecto/adapters/postgres/connection.ex:590: Ecto.Adapters.Postgres.Connection.expr/3
    (elixir 1.12.2) lib/enum.ex:1582: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.12.2) lib/enum.ex:1582: Enum."-map/2-lists^map/1-0-"/2
    (ecto_sql 3.7.1) lib/ecto/adapters/postgres/connection.ex:648: Ecto.Adapters.Postgres.Connection.expr/3
    (ecto_sql 3.7.1) lib/ecto/adapters/postgres/connection.ex:1282: Ecto.Adapters.Postgres.Connection.intersperse_map/4
    (ecto_sql 3.7.1) lib/ecto/adapters/postgres/connection.ex:340: Ecto.Adapters.Postgres.Connection.select/3
    (ecto_sql 3.7.1) lib/ecto/adapters/postgres/connection.ex:113: Ecto.Adapters.Postgres.Connection.all/2
    (ecto_sql 3.7.1) lib/ecto/adapters/postgres.ex:102: Ecto.Adapters.Postgres.prepare/2
    (ecto 3.7.1) lib/ecto/query/planner.ex:180: Ecto.Query.Planner.query_without_cache/4
[...]

PR #3788, opened 2021-11-29
Support parameterized types in type/2

v0idpwn:
Related to discussion in #3784

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -645,8 +645,17 @@ defmodule Ecto.Query.API do
 
       from p in Post, select: type(coalesce(p.cost, 0), :integer)
 
+  For parameterized types, use `type/3`
   """
   def type(interpolated_value, type), do: doc! [interpolated_value, type]
+
+  @doc """
+  Same as type/2, but for parameterized types.
+
+  from p in Post,
+    type(fragment("array_agg(?)", p.tag), {:array, Ecto.Enum}, values: [:a, :b])
+  """
+  def type(interpolated_value, parameterized_type, type_params), do: doc! [interpolated_value, parameterized_type, type_params]
 
   @doc """
   Refer to a named atom binding.
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -645,17 +645,8 @@ defmodule Ecto.Query.API do
 
       from p in Post, select: type(coalesce(p.cost, 0), :integer)
 
-  For parameterized types, use `type/3`
   """
   def type(interpolated_value, type), do: doc! [interpolated_value, type]
-
-  @doc """
-  Same as type/2, but for parameterized types.
-
-      from p in Post,
-        type(fragment("array_agg(?)", p.tag), {:array, Ecto.Enum}, values: [:a, :b])
-  """
-  def type(interpolated_value, parameterized_type, type_params), do: doc! [interpolated_value, parameterized_type, type_params]
 
   @doc """
   Refer to a named atom binding.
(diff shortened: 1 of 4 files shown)


lib/ecto/query/builder.ex
```diff
@@ -83,6 +83,13 @@ defmodule Ecto.Query.Builder do
   end
 
   # tagged types
+  def escape({:type, x, [expr, parameterized_type, type_params]}, type, params_acc, vars, env) do
+    parameterized_type = Macro.expand(parameterized_type, get_env(env))
+    expanded_type = {:parameterized, parameterized_type, parameterized_type.init(type_params)}
+
+    escape({:type, x, [expr, expanded_type]}, type, params_acc, vars, env)
```
v0idpwn, line 90:
```suggestion
  def escape({:type, meta, [expr, parameterized_type, type_params]}, type, params_acc, vars, env) do
    parameterized_type = Macro.expand(parameterized_type, get_env(env))
    expanded_type = {:parameterized, parameterized_type, parameterized_type.init(type_params)}

    escape({:type, meta, [expr, expanded_type]}, type, params_acc, vars, env)
```

v0idpwn:
I don't understand why test is failing on CI, it runs fine on my machine... It has a compile-time dependency to test helpers, but I guess test helpers are available at the time the tests are compiled?

lib/ecto/query/api.ex
```diff
@@ -645,9 +645,18 @@ defmodule Ecto.Query.API do
...
 
+  For parameterized types, use `type/3`
   """
   def type(interpolated_value, type), do: doc! [interpolated_value, type]
 
+  @doc """
+  Same as type/2, but for parameterized types.
+
+  from p in Post,
+    type(fragment("array_agg(?)", p.tag), {:array, Ecto.Enum}, values: [:a, :b])
```
v0idpwn, line 656:
```suggestion
      from p in Post,
        type(fragment("array_agg(?)", p.tag), {:array, Ecto.Enum}, values: [:a, :b])
```

lib/ecto/query/builder.ex
```diff
@@ -682,6 +689,8 @@ defmodule Ecto.Query.Builder do
     do: type
   def validate_type!({:__aliases__, _, _} = type, _vars, env),
     do: Macro.expand(type, get_env(env))
+  def validate_type!({:parameterized, _, _} = type, _vars, _env),
+    do: Macro.escape(type)
```
v0idpwn, line 693: In a second thought, I should've escaped this **before** the validate_type!/3 call, right?

v0idpwn:
I get the reasoning and agree on removing the `type/3` expansion.

But to support the expanded parameterized type on `type/2` (as you suggested) we should merge part of this PR, as otherwise, it will raise a bit further in the pipeline.

josevalim:
@v0idpwn sounds good to me! In the test, let's build the type using Ecto.PT.init/2 too instead of hardcoding it.

v0idpwn:
I added a mention in type/2 docs, but didn't elaborate (and I'm not even sure if we should mention it), as it seems like this is not the first recommended approach

lib/ecto/query/api.ex
```diff
@@ -618,6 +618,10 @@ defmodule Ecto.Query.API do
 
       type(^title, p.title)
 
+  Or a parameterized type, which must be already initialized:
+
+      type(^title, ^@parameterized_type)
```
josevalim, line 623:
```suggestion
  Or a parameterized type, which must be previously initialized
  with `Ecto.ParameterizedType.init/2`:

      @my_enum Ecto.ParameterizedType.init(Ecto.Enum, values: [:foo, :bar, :baz])
      type(^title, ^@my_enum)
```

josevalim (approved):
Just one tiny comment and ship it. :)

PR #3789, opened 2021-11-30
Raise on atoms in planner

josevalim:


josevalim:
Closing this because I don't think it is worth it. :)

PR #3794, opened 2021-12-09
Improve Repo.insert_all/3 typespec

v0idpwn:
Makes it a bit more specific, and also easier to read

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1331,9 +1331,9 @@ defmodule Ecto.Repo do
   @doc group: "Schema API"
   @callback insert_all(
               schema_or_source :: binary | {binary, module} | module,
-              entries_or_query :: [map | [{atom, term | Ecto.Query.t}]] | Ecto.Query.t,
+              entries_or_query :: [%{atom => value} | Keyword.t(value)] | Ecto.Query.t,
               opts :: Keyword.t()
-            ) :: {integer, nil | [term]}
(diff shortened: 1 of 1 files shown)


v0idpwn:
Makes it a bit more specific, and also easier to look at. Had someone confused because they were trying to insert_all with %{String.t() => term}

lib/ecto/repo.ex
```diff
@@ -1331,9 +1331,9 @@ defmodule Ecto.Repo do
   @doc group: "Schema API"
   @callback insert_all(
               schema_or_source :: binary | {binary, module} | module,
-              entries_or_query :: [map | [{atom, term | Ecto.Query.t}]] | Ecto.Query.t,
+              entries_or_query :: [%{atom => value} | Keyword.t(value)] | Ecto.Query.t,
               opts :: Keyword.t()
-            ) :: {integer, nil | [term]}
+            ) :: {integer, nil | [term]} when value: term | Ecto.Query.t()
```
v0idpwn, line 1336: If you have a better name than `value`, tell me :P

josevalim (approved):

lib/ecto/repo.ex
fuelen: what about `integer` -> `non_neg_integer` for all *_all functions?

v0idpwn: I think that's fair, @fuelen. Pr welcome :)

Issue #3768, opened 2021-11-06
`validate_subset/3` broken on custom Ecto types with `type = {:array, _}`


patrickbiermann:
Hi there, I'm pretty sure https://github.com/elixir-ecto/ecto/pull/3764 and related PRs broke `validate_subset/3` with custom Ecto types with `type = {:array, _}`.

To reproduce:
```
defmodule StringListType do
  @moduledoc """
  An Ecto type that wraps {:array, :string}.
  """

  use Ecto.Type

  def type, do: {:array, :string}

  def cast(params) when is_list(params) do
    {:ok, Enum.reject(params, fn item -> item == "" end)}
  end

  def cast(_), do: :error

  def dump(params) when is_list(params) do
    {:ok, params}
  end

  def dump(_), do: :error

  def load(data) when is_list(data) do
    {:ok, data}
  end
end
```

And validate a field of this type by `validate_subset :string_list_field, ["foo"]`. Previous behavior was working as intended, now it crashes with `no match of right hand side value: StringListType` around https://github.com/elixir-ecto/ecto/pull/3764/files#diff-edb6c9aaeb40387eb81c6b238954c0b4d813876d18805c6ae00d7ccc4d78e3f1L2042

Is there are a reason forcing us to drop the `Ecto.Type.type()` step there? Otherwise I'm guessing this would be the root cause here.

_Originally posted by @patrickbiermann in https://github.com/elixir-ecto/ecto/issues/3764#issuecomment-962427506_

josevalim:
I think what we can do in subset is: we get the type, if it is not an array, then we call Ecto.Type.type and see if at least that is an array? And if so, we proceed with the first type. What do you think @v0idpwn?

v0idpwn:
This PR wasn't released yet, otherwise you'd get a nicer error. The version you're running is from even before the PR that added this feature :). I believe I should've added more info about why I reverted it totally.

I think that this is expected behaviour, and believe assuming otherwise is what led to https://github.com/elixir-ecto/ecto/pull/3742#issuecomment-948320014 as well.

If you define a custom type, **we shouldn't make strong assumptions** about how it works (as assuming that inclusion operation is the same of the underlying type), because not all types are shallow. For example:

```elixir
defmodule IntegerArray do
  use Ecto.Type
  def type, do: {:array, :integer}
  def cast(value) when is_list(value), do: {:ok, :array.from_list(value)}
  def cast(_), do: :error
[...]

PR #3797, opened 2021-12-13
Backwards compatibility on Changeset.validate_subset/3

v0idpwn:
If the underlying type of a custom type is `{:array, internal_type}`, assume it is Enumerable of internal_type, doing a semantic comparison for internal_type. Closes issue #3768.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2043,13 +2043,14 @@ defmodule Ecto.Changeset do
   def validate_subset(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:subset, data}, fn _, value ->
       element_type =
-        case Map.fetch!(changeset.types, field) do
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -2043,13 +2043,14 @@ defmodule Ecto.Changeset do
   def validate_subset(changeset, field, data, opts \\ []) do
     validate_change changeset, field, {:subset, data}, fn _, value ->
       element_type =
-        case Map.fetch!(changeset.types, field) do
+        case Map.fetch!(changeset.types, field)do
```
v0idpwn, line 2046:
```suggestion
        case Map.fetch!(changeset.types, field) do
```

josevalim (approved):

patrickbiermann:
Thanks @v0idpwn for the fix, much appreciated! @josevalim any idea on when the next Hex release will be cut?

PR #3798, opened 2021-12-13
Optionally log last known call, publish stacktrace in telemetry

v0idpwn:
This PR is a response to https://github.com/elixir-ecto/ecto_sql/pull/363 and https://github.com/elixir-ecto/ecto_sql/pull/364.

Ecto SQL PR: https://github.com/elixir-ecto/ecto_sql/pull/366

This is how logs will be displayed in Ecto SQL:
```
17:41:01.272 [debug] QUERY OK source="my_schema" db=2.7ms queue=0.1ms idle=985.4ms
DELETE FROM "my_schema" AS m0 []
↳ EctoBug.reproduce/1, at: [file: 'lib/ecto_bug.ex', line: 6]
 
17:41:01.275 [debug] QUERY OK db=3.2ms idle=965.0ms
INSERT INTO "my_schema" ("title","view_count") VALUES ($1,$2),($3,$4),($5,$6),($7,$8),($9,$10),($11,$12) ["sample", 1, "sample", 2, "sample", 3, "sample", 4, "sample", 5, "sample", 6]
↳ EctoBug.reproduce/1, at: [file: 'lib/ecto_bug.ex', line: 7]
 
17:41:01.277 [debug] QUERY OK source="my_schema" db=2.1ms idle=963.9ms
SELECT m0."id", m0."title", m0."view_count" FROM "public"."my_schema" AS m0 []
↳ EctoBug.reproduce/1, at: [file: 'lib/ecto_bug.ex', line: 16]
 
# Note this one doesn't show the exact repo call but the function that calls the function that does the repo call. This is
# due to underlying BEAM optimizations in tail calls
17:41:01.280 [debug] QUERY OK source="my_schema" db=2.5ms idle=955.0ms
SELECT m0."id", m0."title", m0."view_count" FROM "public"."my_schema" AS m0 WHERE (m0."view_count" > $1) AND (m0."view_count" >= $2) [3, 2]
↳ TaskRunner.handle_info/2, at: [file: 'lib/ecto_bug/application.ex', line: 10]
```

To enable this feature, one would need to put in the repo options:
```elixir
config :my_app, MyApp.Repo, get_stacktrace?: true
```

If enabled, the stacktrace also gets published in its entirety in respective telemetry events. This should probably only be enabled in dev environment due to performance impact.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -200,6 +200,7 @@ defmodule Ecto.Repo do
       @adapter adapter
       @default_dynamic_repo opts[:default_dynamic_repo] || __MODULE__
       @read_only opts[:read_only] || false
+      @get_stacktrace? opts[:get_stacktrace?] || false
       @before_compile adapter
       @aggregates [:count, :avg, :max, :min, :sum]
 
@@ -242,7 +243,7 @@ defmodule Ecto.Repo do
         adapter.checked_out?(meta)
       end
 
-      @compile {:inline, get_dynamic_repo: 0, with_default_options: 2, maybe_put_stacktrace: 1}
+      @compile {:inline, get_dynamic_repo: 0, prepare_options: 2, maybe_put_stacktrace: 1}
 
       def get_dynamic_repo() do
         Process.get({__MODULE__, :dynamic_repo}, @default_dynamic_repo)
@@ -253,21 +254,18 @@ defmodule Ecto.Repo do
       end
 
       # temporary: need to set this in repo options and get it here somehow (?)
-      def default_options(_operation), do: [get_stacktrace?: true]
+      def default_options(_operation), do: []
       defoverridable default_options: 1
 
-      defp with_default_options(operation_name, opts) do
-        opts = 
-          operation_name
-          |> default_options()
-          |> Keyword.merge(opts)
-          |> maybe_put_stacktrace()
-
-        opts
+      defp prepare_options(operation_name, opts) do
+        operation_name
+        |> default_options()
+        |> Keyword.merge(opts)
+        |> maybe_put_stacktrace()
       end
 
       defp maybe_put_stacktrace(opts) do
-        if opts[:get_stacktrace?] do
+        if @get_stacktrace? do
           stacktrace = 
             self() 
             |> Process.info(:current_stacktrace)
@@ -284,7 +282,7 @@ defmodule Ecto.Repo do
 
       if Ecto.Adapter.Transaction in behaviours do
         def transaction(fun_or_multi, opts \\ []) do
-          Ecto.Repo.Transaction.transaction(__MODULE__, get_dynamic_repo(), fun_or_multi, with_default_options(:transaction, opts))
+          Ecto.Repo.Transaction.transaction(__MODULE__, get_dynamic_repo(), fun_or_multi, prepare_options(:transaction, opts))
         end
 
         def in_transaction? do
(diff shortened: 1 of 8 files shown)


lib/ecto/repo.ex
```diff
@@ -252,18 +253,36 @@ defmodule Ecto.Repo do
...
+          stacktrace = 
+            self() 
+            |> Process.info(:current_stacktrace)
+            |> elem(1)
+            |> List.delete_at(0)
+
+          Keyword.put(opts, :stacktrace, stacktrace)
+        else
+          opts
+        end
```
v0idpwn, line 278: Is the compiler smart enough to infer this in compile time?

lib/ecto/repo.ex
```diff
@@ -252,18 +252,22 @@ defmodule Ecto.Repo do
         Process.put({__MODULE__, :dynamic_repo}, dynamic) || @default_dynamic_repo
       end
 
+      # temporary: need to set this in repo options and get it here somehow (?)
```
v0idpwn, line 255:
```suggestion
```

josevalim (approved):

lib/ecto/repo/supervisor.ex
```diff
@@ -154,6 +154,29 @@ defmodule Ecto.Repo.Supervisor do
...
+    {adapter, adapter_meta, maybe_put_stacktrace(opts, adapter_meta)}
+  end
+
+  defp maybe_put_stacktrace(opts, adapter_meta) do
+    if opts[:get_stacktrace?] || adapter_meta[:get_stacktrace?] do
+      stacktrace = 
+        self() 
+        |> Process.info(:current_stacktrace)
+        |> elem(1)
+        |> List.delete_at(0)
```
josevalim, line 172:
```suggestion
        |> tl()
```

lib/ecto/repo/supervisor.ex
```diff
@@ -154,6 +154,29 @@ defmodule Ecto.Repo.Supervisor do
...
+
+  @doc false
+  def triplet(name, opts) do
+    {adapter, adapter_meta} = Ecto.Repo.Registry.lookup(name)
+      
+    {adapter, adapter_meta, maybe_put_stacktrace(opts, adapter_meta)}
+  end
+
+  defp maybe_put_stacktrace(opts, adapter_meta) do
+    if opts[:get_stacktrace?] || adapter_meta[:get_stacktrace?] do
```
josevalim, line 167: Hrm, I am not sure this is the best name for the option. It makes sense from internally, but from the user perspective, they are not getting any stacktrace. I would go with `:stacktrace` or `:include_stacktrace_in_telemetry` or similar.

```suggestion
    if opts[:get_stacktrace?] || adapter_meta[:get_stacktrace?] do
```

josevalim: And then we need to document it. :)

v0idpwn: I'll go with stacktrace, because it may also be used in logs, so i think it is a better option

lib/ecto/repo.ex
```diff
@@ -56,6 +56,9 @@ defmodule Ecto.Repo do
       use the `:repo` property in the event metadata for distinguishing
       between repos.
 
+    * `:stacktrace`- publishes the stacktrace in telemetry events and 
+      allows more advanced logging.
+
```
v0idpwn, line 61: Its a bit of a gray area because, well, this is how ecto_sql uses it, but it may be used differently elsewhere :~

josevalim (approved):

lib/ecto/repo/supervisor.ex
```diff
@@ -154,6 +154,29 @@ defmodule Ecto.Repo.Supervisor do
     end
   end
 
+  @compile {:inline, triplet: 2, maybe_put_stacktrace: 2}
```
michalmuskala, line 157: Since the `triplet` function is not used in this module the `inline` annotation has no effect - it only potentially inlines the function inside of the module

josevalim, line 157:
```suggestion
  @compile {:inline, maybe_put_stacktrace: 2}
```

josevalim:

gmile:
Hi, @v0idpwn! Are you using this feature in development, or in production too? If you're using it in production, would you mind sharing how are you storing the stacktraces for later analysis?

v0idpwn:
Hello, @gmile. Our intention here was improving development logs (paired with https://github.com/elixir-ecto/ecto_sql/pull/366). For production environment, the existing approaches to trace and correlate queries with spans are usually more than enough in my experience.

gmile:
@v0idpwn well noted on improving development logs, thanks for making this possible!

> but with opentelemetry you will always be able to relate queries to calling code

I am actively studying OpenTelemetry right now, but am failing to see an out-of-the-box (e.g. provided by a package + configuration) solution 🤔 Would you happen to have an example of how to configure Ecto + OpenTelemetry to make Ecto include Ecto stacktraces in spans? I have briefly looked into `opentelemetry-ecto` (specifically, [this part](https://github.com/open-telemetry/opentelemetry-erlang-contrib/blob/596fcbc2cbc53d6b31d1126e11a13c1a9ee36c0b/instrumentation/opentelemetry_ecto/lib/opentelemetry_ecto.ex#L122-L137)), but it doesn't seem to be including into spans out of the box - it seems like some custom coding might be needed 🤔

v0idpwn:
@gmile it doesn't include the stacktrace, but it does include the "span-trace", which is generally better than a stack-trace for most cases, no?

gmile:
@v0idpwn in the grand scheme of things, a tree of spans (together constituting, what I understand, what's known as a "a trace") is enough for most cases, I agree. But what I am looking for is more resolution, so to say. For example, I want to select a given span coming from Ecto, and be able to find the file/line that produced that span - as quickly as possible.

Basically, I want to do what another person was asking in original issue here https://github.com/elixir-ecto/ecto_sql/pull/364, e.g. "I need the line number in the code", but available in production.

Issue #3800, opened 2021-12-14
Passing sigils to Ecto.Query.from's select option produces subtly different query and nonsense results


zimmydev:
### Environment

* Elixir version (elixir -v): Elixir 1.12.3 (compiled with Erlang/OTP 24)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): postgres (PostgreSQL) 14.1
* Ecto version (mix deps): ecto 3.7.1
* Database adapter and version (mix deps): postgrex 0.15.13
* Operating system: macOS Catalina 10.15.7

### Current behavior

Passing in a words-sigil with an atom modifier to the `select` option produces a different result than passing in an identical list with no syntactic sugar, and it is a very subtle and hard-to-find bug.

Specifically, the queries `from a in "artists", select: [:id, :name]` and `from a in "artists", select: ~w(id name)a` both produce similar-looking queries via the `Inspect` protocol (i.e. `#Ecto.Query<from a0 in "artists", select: [:id, :name]>`). However, when run on the repository, they produce completely different results. More specifically, running `Repo.all/1` on the former in my test database results in `[%{id: 1, name: "Miles Davis"}]` and the latter results in `[[:id, :name]]`.

### Expected behavior

`from a in "artists", select: [:id, :name]` and `from a in "artists", select: ~w(id name)a` should return identical queries since `[:id, :name] == ~w(id name)a`.

aegatlin:
I was able to reproduce this behavior (shown below). I think the root cause is you are passing a `~w` sigil macro into another macro, `Ecto.Query.from`. Macros receive the representation of code instead of the results of code (see near the end of the [intro section here](https://hexdocs.pm/elixir/1.12/Macro.html). `quote do: ~w(a b)a` shows a different representation to `quote do: [:a, :b]`.

If you assign the sigil to a variable and then pin that variable to the query, it will work as desired (see below).

The first query is standard with expected output. The second using the sigil with unexpected output. The third uses a pinned variable of the sigil with expected output.

```elixir
iex(61)> q = Ecto.Query.from Artist, select: [:id, :name] 
#Ecto.Query<from a0 in Friends.Artist, select: [:id, :name]>
iex(62)> q |> Repo.all                                    

13:57:16.130 [debug] QUERY OK source="artists" db=0.6ms idle=1586.3ms
SELECT a0."id", a0."name" FROM "artists" AS a0 []
[
  %Friends.Artist{
[...]

Issue #2958, opened 2019-04-01
Passing sigils to select in from fails to compile


jegaxd26:
### Environment

* Elixir version (elixir -v): Elixir 1.8.1 (compiled with Erlang/OTP 21)
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): none
* Ecto version (mix deps): 3.0.8
* Database adapter and version (mix deps): none
* Operating system: os x Mojave 10.14.3

### Current behavior

Passing list sigils to select (like ~w(id name url)a) fails to compile with error `:id is not a valid query expression` I have created an example repo [here](https://github.com/jegaxd26/ecto_sigils) . I've commented the functions that break compilation, if you uncomment them, mix fails to compile the project.

### Expected behavior

I would expect sigils to work just as lists

wojtekmach:
You need to escape all non-literals in queries, instead of:

```elixir
from(u in __MODULE__, select: @fields)
from(u in __MODULE__, select: ~w(name)a)
```

do:

```elixir
from(u in __MODULE__, select: ^@fields)
from(u in __MODULE__, select: ^~w(name)a)
```

PS thanks for the sample app, it was very helpful to reproduce it!

PR #3804, opened 2021-12-17
Handle atom list sigils in select

v0idpwn:
Related to #3800 and #2958.
Atom list sigils were being escaped as literals instead of going into select_expr.take.

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -34,7 +34,7 @@ defmodule Ecto.Query.Builder.Select do
   def escape(other, vars, env) do
     cond do
       take?(other) ->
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder/select.ex
```diff
@@ -34,7 +34,7 @@ defmodule Ecto.Query.Builder.Select do
   def escape(other, vars, env) do
     cond do
       take?(other) ->
-        {{:{}, [], [:&, [], [0]]}, {[], %{0 => {:any, other}}}}
+        {{:{}, [], [:&, [], [0]]}, {[], %{0 => {:any, Macro.expand(other, env)}}}}
```
v0idpwn, line 37: I think this macro expansion may be not necessary, but I like the idea of expanding the sigil here instead of passing it on. wdyt?

josevalim (approved):

PR #3807, opened 2021-12-29
Update docs for unsafe_fragment of conflict target option

nallwhy:
`{:unsafe_fragment, binary_fragment}` of `:conflict_target` option is necessary for upsert with partial index, but its docs don't have enough information.

I'm not good at English, so any suggestions for the comments are welcome.


josevalim:

Issue #3808, opened 2021-12-30
Validate `validate_number` target


LostKobrakai:
### Precheck

* Do not use the issues tracker for help or support requests (try Elixir Forum, Stack Overflow, IRC or mailing lists, etc).
* For proposing a new feature, please start a discussion on [elixir-ecto](https://groups.google.com/forum/#!forum/elixir-ecto).
* For bugs, do a quick search and make sure the bug has not yet been reported.
* Finally, be nice and have fun!

### Environment

* Elixir version (elixir -v): 1.13.0
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):
* Ecto version (mix deps): 3.7.1
* Database adapter and version (mix deps):
* Operating system:

### Current behavior

```
{%{num: 0}, %{num: :integer}}
|> Ecto.Changeset.cast(%{num: "10"}, [:num])
|> Ecto.Changeset.validate_number(:num, greater_than: 1..10)
```
yields
```
#Ecto.Changeset<
  action: nil,
  changes: %{num: 10},
  errors: [
    num: {"must be greater than %{number}",
     [validation: :number, kind: :greater_than, number: 1..10]}
  ],
  data: %{num: 0},
  valid?: false
>
```

### Expected behavior

Raised error for not having supplied a number to compare to.

josevalim:
PR is welcome!

PR #3809, opened 2021-12-31
validate target in validate_number

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/3808.

### DIFF
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1360,8 +1360,8 @@ defmodule Ecto.ChangesetTest do
 
   test "validate_number/3 with bad target" do
     # Number value
-    assert_raise ArgumentError, "expected target value to be of type Integer or Float, got: #Decimal<10>", fn ->
-      validate_number(changeset(%{"upvotes" => 11}), :upvotes, greater_than: Decimal.new(10))
+    assert_raise ArgumentError, "expected target value to be of type Decimal, Integer or Float, got: 0..10", fn ->
+      validate_number(changeset(%{"upvotes" => 11}), :upvotes, greater_than: 0..10)
     end
 
     # Decimal value
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1358,6 +1358,18 @@ defmodule Ecto.ChangesetTest do
     end
   end
 
+  test "validate_number/3 with bad target" do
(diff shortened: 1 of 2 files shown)


test/ecto/changeset_test.exs
```diff
@@ -1358,6 +1358,18 @@ defmodule Ecto.ChangesetTest do
     end
   end
 
+  test "validate_number/3 with bad target" do
+    # Number value
+    assert_raise ArgumentError, "expected target value to be of type Integer or Float, got: #Decimal<10>", fn ->
+      validate_number(changeset(%{"upvotes" => 11}), :upvotes, greater_than: Decimal.new(10))
+    end
```
v0idpwn, line 1365: I think this should be valid. If we allow decimal to int, why not int to decimal for consistency? Code would also be simpler.

greg-rychlewski: I think I understand what you mean. Could you let me know if this makes sense:

There are several ways the types of `value` and `target_value` can be combined. They can both be `number`, `Decimal`, or something else. So 9 validation scenarios.

a) Both are decimal (1 scenario) - should do a decimal compare
b) Both are number (1 scenario) - should do a `spec_function` compare
c) One is decimal and one is number (2 scenarios) - should do a decimal compare
d) One is decimal and one is not a number/Decimal (2 scenarios) - should raise
e) One is number and one is not a number/Decimal (2 scenarios) - should raise
f) Both are not number/Decimal (1 scenario) - should raise

josevalim: I think you can decouple the validation from the operation. The validation is that any can be a number or a decimal. Then the operation is, if any is a decimal, decimal op, otherwise Kernel op. Does this make sense?

greg-rychlewski: Makes sense! I will re-submit shortly. Thanks both.

greg-rychlewski: I made the changes but was wondering what you guys think of the style. I'm not if you prefer more pattern matches and simpler guards or complex guards and fewer pattern matches. The complex guards look a bit awkward to me but happy change.

josevalim: I think we might be able to clean it up but I am not sure. I will play with it a bit locally. :)

josevalim:

PR #3815, opened 2022-01-14
Support qualified macros in ecto queries

ericfreese:
Previously, if you tried to use a qualified macro via `require`, you would get a compile error that tells you to make sure that you're using `require` correctly. As far as I can tell, there was no way to actually get qualified macros to work because Ecto never actually tries to expand them.

Without the code changes to `lib/ecto/query/builder.ex` but with the modifications to the tests in `test/ecto/query/builder_test.exs`, the test fails with the following error despite the `macro_equal/2` being a macro and the `Macros` module being required:

    == Compilation error in file test/ecto/query_test.exs ==
    ** (Ecto.Query.CompileError) `Macros.macro_equal(q.title, ^test_data)` is not a valid query expression. If you want to invoke Macros.macro_equal/2 in a query, make sure that the module Macros is required and that macro_equal/2 is a macro

        (ecto 3.8.0-dev) expanding macro: Ecto.Query.where/3
        test/ecto/query_test.exs:39: Ecto.QueryTest."test query building allows macros"/1
        (elixir 1.12.2) expanding macro: Kernel.|>/2
        test/ecto/query_test.exs:39: Ecto.QueryTest."test query building allows macros"/1

The code change makes this test pass by trying expansion when encountering a qualified macro rather than raising an error. If the expansion fails (invalid macro), it still raises an error with the same messaging that you would get if you had used an invalid unqualified macro. This messaging still suggests that you make sure that the module is required and that the function is a macro, so the impact to developer experience in this situation should be minimal.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -459,16 +459,9 @@ defmodule Ecto.Query.Builder do
   end
 
   # Raise nice error message for remote calls
-  def escape({{:., _, [mod, fun]}, _, args} = other, _type, _params_acc, _vars, _env)
+  def escape({{:., _, [_, fun]}, _, _} = other, type, params_acc, vars, env)
       when is_atom(fun) do
-    fun_arity = "#{fun}/#{length(args)}"
-
-    error! """
-    `#{Macro.to_string(other)}` is not a valid query expression. \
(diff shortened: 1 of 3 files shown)


josevalim:

PR #3817, opened 2022-01-15
Add no_return to callbacks that might raise

periclesd:
Where the counterpart functions that do not raise, ie without a `!`, are usually described with `:: Ecto.Schema.t() | nil`

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -654,7 +654,7 @@ defmodule Ecto.Repo do
   """
   @doc group: "Query API"
   @callback get!(queryable :: Ecto.Queryable.t(), id :: term, opts :: Keyword.t()) ::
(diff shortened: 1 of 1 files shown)


josevalim:
`no_return()` is actually not required and sometimes may make Dialyzer analysis worse. `no_return()` is typically only used when it is the only option.

periclesd:
cool, I pondered opening this PR for a bit until I found https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/uuid.ex#L50 and https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/uuid.ex#L138 which compelled me to do it, anyway, thanks for the clarification, have a nice weekend!

josevalim:
We should actually remove those!

PR #3818, opened 2022-01-15
Remove no_return from function that might return other options as well

periclesd:
> no_return() is actually not required and sometimes may make Dialyzer analysis worse. no_return() is typically only used when it is the only option.

https://github.com/elixir-ecto/ecto/pull/3817#issuecomment-1013689802


josevalim:

PR #3823, opened 2022-01-24
Small dynamic binding fixes

zachdaniel:
1. escape dynamic named bindings that appear in the bindings list:
  This allows you to do `from [{^as, row}] in query, ....`, whereas before the dynamic bindings only worked with `as(^binding)`. This can be cleaner if you need to reference the same binding multiple times.

2. require literal atom or interpolated value in `join` `as` just like we do in the `from` `as`
This was simply an oversight in the initial implementation, which resulted in queries looking like:

```elixir
from row in query,
  as: ^as,
  join: foo in assoc(row, :foo),
  as: other_as
```
which is clearly problematic.


zachdaniel:
The test doesn't quite feel like it is in the right place though, now that I look at it. Should that test go in `test/ecto/query_test.exs`?

josevalim:

PR #3824, opened 2022-01-24
fix: set proper typespec for `Ecto.Query.Builder.From.apply`

zachdaniel:
The typespec was incorrect given the new values allowed for the `as`. I could have done something like `atom | Macro.t()` or `atom | {:^, term, term}` or something, but that seemed like overkill to me. Happy to make the type more specific if necessary.


josevalim:
I would go with Macro.t as that’s what we use in most of the builder specs. :)

zachdaniel:
done :)

josevalim:

PR #3825, opened 2022-01-25
Allow custom error key in unique_constraint

Malian:
`unsafe_validate_unique/3` allows a custom error key to which changeset error will be added when check fails.

This PR adds the same feature to `unique_constraint`.

It is useful when the composite key is `[:id, :slug]` for example, and the slug is duplicated. Saying to the user that the ID already exists does not make sense as most of the time the ID is auto generated and is not duplicated. With the PR we are able to set `:slug` as the error key.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2646,6 +2646,10 @@ defmodule Ecto.Changeset do
       `:suffix` matches any repo constraint which `ends_with?` `:name`
        to this changeset constraint.
 
+    * `:error_key` - the key to which changeset error will be added when
+      check fails, defaults to the first field name of the given list of
+      fields.
+
   ## Complex constraints
 
   Because the constraint logic is in the database, we can leverage
@@ -2727,7 +2731,8 @@ defmodule Ecto.Changeset do
     constraint = opts[:name] || unique_index_name(changeset, fields)
     message    = message(opts, "has already been taken")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :unique, to_string(constraint), match_type, first_field, message)
+    error_key  = Keyword.get(opts, :error_key, first_field)
+    add_constraint(changeset, :unique, to_string(constraint), match_type, error_key, message)
   end
 
   defp unique_index_name(changeset, fields) do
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -2723,11 +2727,12 @@ defmodule Ecto.Changeset do
     unique_constraint(changeset, [field], opts)
   end
 
-  def unique_constraint(changeset, [first_field | _] = fields, opts) do
+  def unique_constraint(changeset, fields, opts) when is_list(fields) do
```
josevalim, line 2730: I would actually revert this to what it was because we need to pattern match on the list not being empty anyway!

```suggestion
  def unique_constraint(changeset, [first_field | _] = fields, opts) do
```

lib/ecto/changeset.ex
```diff
@@ -2723,11 +2727,12 @@ defmodule Ecto.Changeset do
...
     constraint = opts[:name] || unique_index_name(changeset, fields)
     message    = message(opts, "has already been taken")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :unique, to_string(constraint), match_type, first_field, message)
+    error_key  = Keyword.get(opts, :error_key, hd(fields))
```
josevalim, line 2734:
```suggestion
    error_key  = Keyword.get(opts, :error_key, first_field)
```

test/ecto/changeset_test.exs
```diff
@@ -1825,7 +1825,13 @@ defmodule Ecto.ChangesetTest do
...
            [%{type: :unique, field: :permalink, constraint: "posts_url_color_index", match: :exact,
               error_message: "has already been taken", error_type: :unique}]
-  end
+
+    changeset = change(%Post{}) |> unique_constraint([:permalink, :color], error_key: :color)
+
+    assert constraints(changeset) ==
+           [%{type: :unique, field: :color, constraint: "posts_url_color_index", match: :exact,
+              error_message: "has already been taken", error_type: :unique}]
+    end
```
josevalim, line 1834:
```suggestion
  end
```

lib/ecto/changeset.ex
Malian: Good catch!

PR #3829, opened 2022-02-02
Clarify the benefit of a composite type

trodrigu:
When reading the docs about `Ecto.Enum` I wasn't exactly sure what the benefit of a composite type was so I wanted to add a little bit more to clarify.

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -30,7 +30,7 @@ defmodule Ecto.Enum do
   Composite types, such as `:array`, are also supported which allow selecting
   multiple values per record:
 
-      field :roles, {:array, Ecto.Enum}, values: [:Author, :Editor, :Admin]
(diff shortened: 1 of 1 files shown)


lib/ecto/enum.ex
```diff
@@ -27,7 +27,8 @@ defmodule Ecto.Enum do
   Some databases also support enum types, which you could use in combination
   with the above.
 
-  Composite types, such as `:array`, are also supported:
+  Composite types, such as `:array`, are also supported which allow selecting
+  multiple values per record:
 
       field :roles, {:array, Ecto.Enum}, values: [:Author, :Editor, :Admin]
```
josevalim, line 33:
```suggestion
      field :roles, {:array, Ecto.Enum}, values: [:author, :editor, :admin]
```

josevalim:

PR #3840, opened 2022-02-28
Detect `empty_values` inside lists when casting.

marcandre:
See https://elixirforum.com/t/phoenix-casting-multiple-checkboxes-to-array-ecto-enum/46252/6

This way, an HTML form with checkboxes having names ending with `[]` can have
a hidden input with empty value to ensure the presence of that parameter
if none of the checkboxes are checked.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -867,9 +867,7 @@ defmodule Ecto.Type do
   @spec filter_empty_values(t, any, [any]) :: {:ok, any} | :empty
   def filter_empty_values({:array, type}, value, empty_values) when is_list(value) do
     value = for elem <- value,
-      result = filter_empty_values(type, elem, empty_values),
-      :empty != result,
-      {:ok, elem} = result,
+      {:ok, elem} <- [filter_empty_values(type, elem, empty_values)],
       do: elem
 
     {:ok, value}
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -863,6 +863,26 @@ defmodule Ecto.Type do
   defp same_date(%Date{} = term), do: {:ok, term}
   defp same_date(_), do: :error
 
+  @doc false
+  @spec filter_empty_values(t, any, [any]) :: {:ok, any} | :empty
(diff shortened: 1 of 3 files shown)


lib/ecto/type.ex
```diff
@@ -863,6 +863,26 @@ defmodule Ecto.Type do
   defp same_date(%Date{} = term), do: {:ok, term}
   defp same_date(_), do: :error
 
+  @doc false
+  @spec filter_empty_values(t, any, [any]) :: {:ok, any} | :empty
+  def filter_empty_values({:array, type}, value, empty_values) when is_list(value) do
+    value = for elem <- value,
+      result = filter_empty_values(type, elem, empty_values),
+      :empty != result,
+      {:ok, elem} = result,
```
josevalim, line 872:
```suggestion
      {:ok, elem} <- [filter_empty_values(type, elem, empty_values)],
```

josevalim:

lib/ecto/type.ex
marcandre: Thanks, I was wondering if `pattern <- [...]` was idiomatic or not 👍

leandrocp:
Hi @marcandre and @josevalim I just found out this PR has caused a breaking change on our app because oddly enough we were relying on the old behavior, ie: we expect the changeset to _keep_ `""`  😆

It's an easy fix on our side and the change is welcome, I'm just wondering if it's worth adding a note on CHANGELOG.md that it's _potentially_ a breaking change.

marcandre:
Sounds like a good idea. I'm trying to double check if it was included in version 3.7.2, but that commit (960725b9e0719c79d8296c4852072780e8d5cc3e) "does not belong to any branch on this repository"...

fuelen:
Another breaking change that I've noticed is that this code now doesn't put the error "can't be blank" into changeset:
```elixir
{%{}, %{uris: {:array, :string}}}
|> Ecto.Changeset.cast(%{uris: []}, [:uris], empty_values: ["", []])
|> Ecto.Changeset.validate_required(:uris)
```
but it was there in Ecto. 3.7.2

PR #3846, opened 2022-03-13
 Mention `init/2` in ParameterizedType moduledoc 

v0idpwn:


josevalim:
I think this would be good to have in the ParameterizedType docs, but too specific here. WDYT?

v0idpwn:
I think the root issue is that it is not very clear how to use parameterized types outside of schemas, so I probably adding an example of usage of init/2 in the ParameterizedType moduledoc should be enough, yes.

josevalim:
Ping :D

josevalim (approved):

Issue #3851, opened 2022-03-25
Cannot remove entry from deeply nested embed


jbcden:
### Precheck

* Do not use the issues tracker for help or support requests (try Elixir Forum, Stack Overflow, IRC or mailing lists, etc).
* For proposing a new feature, please start a discussion on [elixir-ecto](https://groups.google.com/forum/#!forum/elixir-ecto).
* For bugs, do a quick search and make sure the bug has not yet been reported.
* Finally, be nice and have fun!

### Environment

* Elixir version (elixir -v): 1.13.0
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 13.6
* Ecto version (mix deps): 3.7.2
* Database adapter and version (mix deps): postgrex: 0.16.2
* Operating system: macOS, Manjaro Linux

### Current behavior

Given the following schema:

```elixir
defmodule Embeddy.Rules do
  @moduledoc false
  alias Embeddy.Rules.Rule
  alias Embeddy.Repo

  def create_rule(attrs \\ %{}) do
    %Rule{}
    |> Rule.changeset(attrs)
    |> Repo.insert()
  end

  def update_rule(%Rule{} = rule, attrs \\ %{}) do
    rule
    |> Rule.changeset(attrs)
    |> Repo.update()
  end
end

defmodule Embeddy.Rules.Condition do
  use Ecto.Schema

  import Ecto.Changeset

  alias Embeddy.Rules.Condition

  @moduledoc false

  embedded_schema do
    field :title
    field :bar
    embeds_many :conditions, Condition, on_replace: :delete
  end

  def changeset(%Condition{} = condition, attrs = %{}) do
    condition
    |> cast(attrs, [:title, :bar])
    |> cast_embed(:conditions)
  end
end
```

I am trying to remove an entry from an embedded schema:

```elixir
    {:ok, rule} =
      Rules.create_rule(%{
        name: "my cool rule",
        description: "Let's delete some conditions",
        conditions: %{
          title: "Top-level",
          bar: "baz",
          conditions: [
            %{title: "cond 1", bar: "> 123"},
            %{title: "cond 2", bar: "< 99"}
          ]
        }
      })

    condition_to_remove_id =
      get_in(rule.conditions, [
        Access.key(:conditions),
        Access.filter(fn i -> i.title == "cond 1" end),
        Access.key(:id)
      ])
      |> List.first()

    {_original_condition, new_rule} =
      pop_in(rule, [
[...]

PR #3852, opened 2022-03-27
allow kw and map updates from put_embed/4

greg-rychlewski:
While investigating https://github.com/elixir-ecto/ecto/issues/3851 I found that using keyword lists and maps to update existing embedded structs with `put_embed/4` had some unexpected behaviour.

The issue is that the action for the resulting changeset is always `:insert`:

```elixir
defp action_from_changeset(_) do
   :insert # We don't care if it is insert/update for embeds (no meta)
end
```
But when you are updating an existing embedded struct with a primary key, the allowed actions are `:update` and `:delete`:

```elixir
defp allowed_actions(pk_values) do
    if Enum.all?(pk_values, &is_nil/1) do
      [:insert, :update, :delete]
    else
      [:update, :delete]
    end
end
```

So I think there is a need to differentiate between `:insert` and `:update` actions in this scenario. Otherwise `check_action!/2` will always raise.

The way it happens when not using kw lists/maps is that a `nil` current value forces an `:insert` action and anything else forces an `:update` action. I just extended this logic over to the kw list/map scenario.

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -174,7 +174,7 @@ defmodule Ecto.Changeset.Relation do
     {:ok,
      changeset
      |> assert_changeset_struct!(relation)
-     |> put_new_action(action_from_changeset(changeset))}
+     |> put_new_action(action_from_changeset(changeset, nil))}
   end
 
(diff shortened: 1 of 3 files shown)


josevalim:

PR #3859, opened 2022-04-05
Add updated_at timestamp assertion to "changing embeds on update" test

richthedev:
We noticed some problems when updating from `ecto` version `2.2.12` to `3.7.2`.

Upon further investigation, we found out that the `updated_at` field is not updated with the current datetime when updating embeds on a schema.

I pulled the repo and added a quick assertion in the repo test and it seems to indicate that the field is indeed not updating correctly. It seems to me that the following test should succeed.

Thanks for the help

### DIFF
--- a/test/ecto/repo/embedded_test.exs
+++ b/test/ecto/repo/embedded_test.exs
@@ -240,7 +240,11 @@ defmodule Ecto.Repo.EmbeddedTest do
   end
 
   test "changing embeds on update" do
-    sample = %MyEmbed{x: "xyz", id: @uuid}
+    an_hour_ago =
+      DateTime.utc_now()
+      |> DateTime.add(-3600)
+      |> DateTime.to_naive()
+    sample = %MyEmbed{x: "xyz", id: @uuid, updated_at: an_hour_ago}
     sample_changeset = Ecto.Changeset.change(sample, x: "abc")
 
     changeset =
@@ -253,6 +257,7 @@ defmodule Ecto.Repo.EmbeddedTest do
     assert embed.x == "abc"
     refute embed.inserted_at
     assert embed.updated_at
+    assert NaiveDateTime.compare(embed.updated_at, an_hour_ago) == :gt
 
     kw_changeset = Ecto.Changeset.put_embed(changeset, :embed, id: @uuid, x: "def")
     schema = TestRepo.update!(kw_changeset)


richthedev:
After some more research, the test passes on `v3.4.5` but not on `v3.4.6`

greg-rychlewski:
I believe it's because you are already setting the `updated_at` manually as a change. The autogenerated fields will only set themselves [if they haven't already been set](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/embedded.ex#L260)

```elixir
defp autogenerate(changes, action, schema) do
  autogen_fields = action |> action_to_auto() |> schema.__schema__()

  Enum.reduce(autogen_fields, changes, fn {fields, {mod, fun, args}}, acc ->
    case Enum.reject(fields, &Map.has_key?(changes, &1)) do
      [] ->
        acc

      fields ->
        generated = apply(mod, fun, args)
        Enum.reduce(fields, acc, &Map.put(&2, &1, generated))
    end
  end)
end
```

Try changing the test to this and it will pass:

```elixir
test "embed update" do
  sample = %MyEmbed{x: "xyz", id: @uuid}
  sample_changeset = Ecto.Changeset.change(sample, x: "abc")

  changeset =
    %MySchema{id: 1, embed: sample}
    |> Ecto.Changeset.change
    |> Ecto.Changeset.put_embed(:embed, sample_changeset)

  schema = TestRepo.update!(changeset)
  datetime1 = schema.embed.updated_at
  schema = TestRepo.update!(changeset)
  datetime2 = schema.embed.updated_at

  assert NaiveDateTime.compare(datetime2, datetime1) == :gt
end
```

lucbernard:
Hi, I can see how your test can pass. In your case, you're generating the embed's changeset yourself when performing the test. That indeed works.

In our case, we are not generating the embed's changeset ourselves. We update the embeds by passing through the parent's changeset function and using `cast_embed`, like so:

```elixir
def changeset(%__MODULE__{} = parent, attrs) do
  parent
  |> cast(...)
  |> cast_embed(:embed)
end
```

In that case, ecto is itself generating the embed's changeset along with the parent's cs. That's the flow that's broken. The issue seems to be with `Ecto.Changeset.Relation.surface_changes`. More specifically here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset/relation.ex#L552. We go through this line for every field that's not been explicitly updated by the "user" and is not nil. The `updated_at` field falls in that category. 

So when exiting this function, we end up having an updated changes _manifest_ that suddenly and incorrectly includes the `updated_at` field. And then, like you said, the autogeneration of the updated_at value is skipped.

The following test illustrates the problem:

```elixir
```

greg-rychlewski:
Thanks that test is really helpful. You're right about what's happening. `Ecto.Changeset.Relation.surface_changes` is moving previously persisted autogenerated fields into the `changes` map and causing `updated_at` to be passed over for autogeneration.

If autogenerated/autoupdated fields are meant to mean "this field will be created/updated if not specified by the user" then maybe the best solution is to autogenerate them before calling `Ecto.Changeset.Relation.surface_changes`. @josevalim does that sound reasonable?

josevalim:
Yes, let's explore this route!

greg-rychlewski:
@richthedev  @lucbernard 

Thank you for your help uncovering this issue! It has been fixed here: https://github.com/elixir-ecto/ecto/pull/3868.

PR #3862, opened 2022-04-07
Warn when validate_required/3 called on has_many

### LINKED ISSUE #3861: Have `validate_required/3` raise when used on has_many/embed_many?
### Environment

* Elixir version (elixir -v): 1.13.3
* Ecto version (mix deps): `master` branch

### Current behavior

The following would work both with embedded schemas or regular `has_many` associations:

```elixir
defmodule Post do
  use Ecto.Schema
  import Ecto.Changeset

  embedded_schema do
    field :text, :string
  end

  def changeset(pets, attrs) do
    pets
    |> cast(attrs, [:text])
    |> validate_required([:text])
  end
end

defmodule Author do
  use Ecto.Schema
  import Ecto.Changeset

  embedded_schema do
    field :name, :string

    embeds_many :posts, Posts
  end

  def changeset(user, attrs) do
    user
    |> cast(attrs, [:name])
    |> cast_embed(:posts)  # should actually be `required: true`
    |> validate_required([:name, :posts])  # no effect
  end
end
```

The following will be considered valid:

```elixir
iex> Author.changeset(%Author{}, %{name: "John"})
#Ecto.Changeset<
  action: nil,
  changes: %{name: "John"},
  errors: [],
  data: #Author<>,
  valid?: true
>
```

The [doc for validate_required/3](https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_required/3) points out to the actual solution, which is to add `required: true` to `cast_embed` or `cast_assoc`.
But unlike we write a unit-test to check this scenario, it is really easy to have this mistake remain unnoticed and makes it a bit of a footgun.

### Suggested behavior

Since the use of validate_required on `has_many`/`embed_many` is most likely a mistake in every possible case, maybe we could raise with an helpful message to remove, e.g.

```
attempting to validate has_many association :posts with validate_required/3 which has no effect.
You can pass the :required option to Ecto.Changeset.cast_assoc/3 to achieve this.
```

What do you think? If this issue is accepted, I have a PR ready to go.

sabiwara:
Calling `validate_required/3` on a `has_many`/`embeds_many` has no effect, add a warning to make it obvious.

Close #3861

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1798,7 +1798,7 @@ defmodule Ecto.Changeset do
   data given to the `changeset` is not empty.
 
   Do not use this function to validate associations that are required,
-  instead pass the `:required` option to `cast_assoc/3`.
+  instead pass the `:required` option to `cast_assoc/3` or `cast_embed/3`.
 
   Opposite to other validations, calling this function does not store
   the validation under the `changeset.validations` key. Instead, it
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1824,6 +1824,7 @@ defmodule Ecto.Changeset do
 
     fields_with_errors =
       for field <- fields,
+          ensure_field_not_many!(changeset, field),
           missing?(changeset, field, trim),
           ensure_field_exists!(changeset, field),
           is_nil(errors[field]),
@@ -1982,6 +1983,23 @@ defmodule Ecto.Changeset do
       raise ArgumentError, "unknown field #{inspect(field)} in #{inspect(data)}"
     end
     true
+  end
+
+  defp ensure_field_not_many!(%Changeset{types: types}, field) do
+    case types do
+      %{^field => {:assoc, %Ecto.Association.Has{cardinality: :many}}} ->
+        IO.warn("attempting to validate has_many association #{inspect(field)} " <>
+                "with validate_required/3 which has no effect. You can pass the " <>
+                ":required option to Ecto.Changeset.cast_assoc/3 to achieve this.")
+
+      %{^field => {:embed, %Ecto.Embedded{cardinality: :many}}} ->
+        IO.warn("attempting to validate embed_many field #{inspect(field)} " <>
+                "with validate_required/3 which has no effect. You can pass the " <>
+                ":required option to Ecto.Changeset.cast_embed/3 to achieve this.")
+
+      _ ->
+        true
+    end
   end
 
   defp missing?(changeset, field, trim) when is_atom(field) do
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset.ex
```diff
@@ -1984,6 +1985,23 @@ defmodule Ecto.Changeset do
     true
   end
 
+  defp ensure_field_not_many!(%Changeset{types: types}, field) do
```
josevalim, line 1988: We can merge this function with `ensure_fields_exists!`, so we don't lookup `types` twice. :)

sabiwara: Actually `ensure_fields_exists!` doesn't get called, because we filter out by `missing?` first before calling `ensure_fields_exists!`.
`missing?` calls `get_field`, gets `[]`, and return `false`.
Will can still avoid looking it up twice and actually look it up before looping on `fields`.
Pushed https://github.com/elixir-ecto/ecto/pull/3862/commits/245b4e179f11acf386fdf947d19584993772b070

lib/ecto/changeset.ex
```diff
@@ -1978,14 +1978,14 @@ defmodule Ecto.Changeset do
     end
   end
 
-  defp ensure_field_exists!(%Changeset{types: types, data: data}, field) do
+  defp ensure_field_exists!(changeset = %Changeset{}, types, field) do
     unless Map.has_key?(types, field) do
-      raise ArgumentError, "unknown field #{inspect(field)} in #{inspect(data)}"
+      raise ArgumentError, "unknown field #{inspect(field)} in #{inspect(changeset.data)}"
```
sabiwara, line 1983: Stop fetching `data` in the happy path

lib/ecto/changeset.ex
```diff
@@ -1818,15 +1818,15 @@ defmodule Ecto.Changeset do
   """
   @spec validate_required(t, list | atom, Keyword.t) :: t
   def validate_required(%Changeset{} = changeset, fields, opts \\ []) when not is_nil(fields) do
-    %{required: required, errors: errors, changes: changes} = changeset
+    %{required: required, errors: errors, changes: changes, types: types} = changeset
```
sabiwara, line 1821: fetching `types` before the loop

test/ecto/changeset/embedded_test.exs
```diff
@@ -536,6 +536,17 @@ defmodule Ecto.Changeset.EmbeddedTest do
 
   ## Others
 
+  test "validate_required/3 with has_many raises" do
+    import ExUnit.CaptureIO
+
+    base_changeset = Changeset.change(%Author{})
+
+    assert capture_io(:stderr, fn ->
```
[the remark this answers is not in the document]
josevalim: >  Technically I would need to remove async: true to be able to do this safely.

In more recent Elixir versions it is fine to use capture_io with async :) so it is fine.

josevalim, line 544:
```suggestion
    base_changeset = Changeset.change(%Author{})

    assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
```

josevalim:

Issue #3855, opened 2022-04-03
Wrong result when referencing association on preload


pmargreff:
### Environment

- Elixir version (elixir -v): Elixir 1.12.3 (compiled with Erlang/OTP 22)
- Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): postgres:13.5-alpine  (docker image)
- Ecto version (mix deps): ecto 3.7.2
- Database adapter and version (mix deps): postgrex 0.16.2 
- Operating system: Ubuntu 20.04

### Current behavior

When doing a join and using a preload to reference the field I want to preload, `Repo.all` always brings only one result, even when I have multiple results in the table representing the schema `Foo`.

```elixir
query = (from f in Foo,
  left_join: b in ^(from Bar),
  on: f.a == b.a and f.b == b.b
  preload: [bars: b])

  Repo.all(query)
```

The interesting part is that `Repo.aggregate(query, :count)` gives me the counter result that I expect. 

Running the SQL that Ecto creates directly in the database also gives me the expected amount of results for the query. If I pass the preload as a query (`preload: [bars: query]`) it also works as expected. So I think that Ecto is failing to send the entire set of results back to the users.

### Expected behavior

When referencing the join value in the preload the result set has multiple entries if my query has multiple results. 

### Sample repository

A working sample of the issue can be found at: https://github.com/pmargreff/ecto_preload_bug/blob/main/test/preload_bugs_test.exs

dvic:
You could do a 

```elixir
{q, args} = Ecto.Adapters.SQL.to_sql(:all, Repo, query)
IO.puts(q)
IO.inspect(args)
```

to see what kind of query is being produced?

pmargreff:
@dvic I've added a sample repository with the issue to lower the friction. Testing that should be as easy as a `git clone` + `mix deps.get`.  Let me know if you need any other info about the problem.

greg-rychlewski:
If you define the primary key on the schema from the `has_many` side of the association it will work. i.e change this:

```elixir
@primary_key false
schema "pull_requests" do
  field :repo, :string
  field :number, :integer
  field :title, :string

  has_many :commits, PreloadBugs.Commit, references: :number
end
```

to this

```elixir
@primary_key {:number, :integer, autogenerate: false}
schema "pull_requests" do
  field :repo, :string
  field :title, :string

  has_many :commits, PreloadBugs.Commit, references: :number, foreign_key: :pr_number
end
```

[...]

PR #3867, opened 2022-04-09
raise if association doesn't have a primary key

greg-rychlewski:
Related to https://github.com/elixir-ecto/ecto/issues/3855

When an association doesn't have a primary key, the data structure used to track the relationships between associated structs doesn't work as expected because it relies on the primary key. This PR raises early if any of the associations are missing a primary key. 

The queries can still technically function correctly if the leaf association doesn't have a PK, but I'm assuming the desired behaviour is to raise for symmetry/consistency reasons. For instance, if the user switched the order of the query so that the leaf is the parent it would no longer work. But if I'm mistaken please let me know!

### DIFF
--- a/lib/ecto/repo/assoc.ex
+++ b/lib/ecto/repo/assoc.ex
@@ -111,7 +111,6 @@ defmodule Ecto.Repo.Assoc do
     end)
 
     {_source, schema, _prefix} = elem(sources, idx)
-    {schema.__schema__(:primary_key), %{}, initial_dict, acc}
 
     case schema.__schema__(:primary_key) do
       [] -> raise Ecto.NoPrimaryKeyFieldError, schema: schema
--- a/lib/ecto/repo/assoc.ex
+++ b/lib/ecto/repo/assoc.ex
@@ -112,5 +112,10 @@ defmodule Ecto.Repo.Assoc do
 
     {_source, schema, _prefix} = elem(sources, idx)
     {schema.__schema__(:primary_key), %{}, initial_dict, acc}
+
+    case schema.__schema__(:primary_key) do
+      [] -> raise Ecto.NoPrimaryKeyFieldError, schema: schema
+      pk -> {pk, %{}, initial_dict, acc}
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/ecto/repo/assoc.ex
```diff
@@ -112,5 +112,10 @@ defmodule Ecto.Repo.Assoc do
 
     {_source, schema, _prefix} = elem(sources, idx)
     {schema.__schema__(:primary_key), %{}, initial_dict, acc}
```
josevalim, line 114:
```suggestion
```

josevalim:
Hi Greg! Thank you for the amazing for pull requests. Would you like to join the Ecto team?

It doesn’t Require new commitments From your side. But All help is appreciated And you have been doing a fantastic job. It is your decision, no pressure! just let me know :)

Have a great weekend!

greg-rychlewski:
Hi @josevalim 

I would love to join the Ecto team! I feel extremely honoured that you asked. Thank you so much.

I put my email and location in my profile now, in case you or anyone else would like a more direct way to contact me. And so you know when I might be asleep.

Have a great weekend! Thanks again!

josevalim:
Invite sent, welcome to the team!

Feel free to merge your own PRs from now on after they get at least one approval by another team member. Also feel free to merge PR from external contributors. :)

josevalim (approved):

JustMikey:
Thanks for the PR but I am afraid it was a breaking change for me and I wonder if we should update the changelog file for the version `3.9.1`.

For example, in my case I have a schema `Profile` using a table `profiles` and a schema `ProfileStats` using a materialized view `profile_stats`, where the latter simply contains the analytics about the former and as such, it doesn't have a primary key, only column `profile_id` that is used as a foreign key, so the schema definition has `@primary_key false` and `belongs_to :profile, Profile` and it has stopped working after I updated to `3.9.1`. It used to work both with inner joins and preloads but now it seems to be broken. And I imagine other folks may end up in a similar situation after the upgrade.

greg-rychlewski:
Thanks for letting us know @JustMikey. That sounds like a good idea to me. I think this one went into 3.8.0. 

@josevalim could I submit a modified changelog and have it published to hexdoc or is it too late?

josevalim:
Let’s update the CHANGELOG and it will be fixed in the last release. :)

Issue #3860, opened 2022-04-05
updated_at timestamp not updated on embeds


richthedev:
### Environment

* Elixir version: Elixir 1.13.3 (compiled with Erlang/OTP 23)
* Ecto version (mix deps): 3.8.0-dev
* Operating system: macOS Monterey 12.3

### Current behavior

The `updated_at` field of embeds using the `timestamp()` macro is never updated.

### Expected behavior

The `updated_at` field of embeds is updated and tracks the embed changes.

### Notes

Please see #3859 for details

richthedev:
I did some more digging and found out the bug seems to have been introduced in `v3.4.6`.

PR #3868, opened 2022-04-10
don't surface changes on embed updates

greg-rychlewski:
Fixes https://github.com/elixir-ecto/ecto/pull/3859 and https://github.com/elixir-ecto/ecto/issues/3860

When preparing an embed for insert/update, it always calls `Relation.surface_changes`. When updating an embed, this causes the previously persisted `updated_at` value to be surfaced and auto-update doesn't generate a new one.

The proposal here is to only call `Relation.surface_changes` on `:insert`. I believe this is the intended use, because on `:insert` the fields of the underlying struct are considered to be new changes. But on `:update` I believe the underlying struct is taken to be the persisted data. This is also how it works for non-embedded schema.

I made a suggestion to create the autogenerated fields before calling `Relation.surface_changes` but I don't think this is correct anymore. On `:insert` it will cause any value the user supplied in the struct for an autogenerated field to be overwritten.

### DIFF
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -199,7 +199,7 @@ defmodule Ecto.Embedded do
 
   defp to_struct(%Changeset{data: data} = changeset, action, %{related: schema}, adapter) do
     %{data: struct, changes: changes} = changeset =
-      Relation.surface_changes(changeset, data, schema.__schema__(:fields))
+      maybe_surface_changes(changeset, data, schema, action)
 
     embeds = prepare(changeset, schema.__schema__(:embeds), adapter, action)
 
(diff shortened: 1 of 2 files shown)


josevalim (approved):

gcauchon:
@greg-rychlewski @josevalim I also investigated the issue and came up with an alternative solution :

Screen Shot 2022-04-10 at 08 40 06

Matching on `changes` with a value that DID changed allows to reduce on updated fields only, thus not injecting changes in the `changeset`.

Which leads to the following question: Is `Relation.surface_changes/3` intend only relevant on `insert` like it was described in the PR’s description?

greg-rychlewski:
What happens when you apply that change and run the full test suite? I think it will break some things, because you are losing the ability to pull values from the struct into the changes map on insert, unless they are also specified in the changes map.

PR #3869, opened 2022-04-12
log uuids

greg-rychlewski:
Resolves https://github.com/elixir-ecto/ecto_sql/issues/368

There is a companion PR in ecto_sql that would have to be merged after this one: https://github.com/elixir-ecto/ecto_sql/pull/391.

The idea is to have the adapter dumpers specify whether they want to keep the casted value for a parameter. This will wrap both the dumped and the casted values in a `%Ecto.Query.Tagged{}` struct. When the query is executed, the logger will display the casted value and the dumped value will be sent to the driver. 

I'm not too familiar with `Ecto.Query.Tagged` but it seems like I have to extract the value from it before sending it to the driver, otherwise I get this: 

`** (Protocol.UndefinedError) protocol Jason.Encoder not implemented for %Ecto.Query.Tagged{tag: nil, type: :binary_id, value: <<161, 252, 251, 221, 173, 96, 76, 203, 139, 249, 71, 170, 189, 102, 208, 115>>} of type Ecto.Query.Tagged (a struct), Jason.Encoder protocol must always be explicitly implemented.`

But if I extract the value then I'm not sure if other adapters are using the struct for other reasons and need it to be passed through. So I'm a bit unsure if I should stick with `Ecto.Query.Tagged` or create a new struct. 

In terms of logging, it seems like it is working well. I tested using all these different types of queries:

```elixir
test "uuid" do
  uuid1 = "a1fcfbdd-ad60-4ccb-8bf9-47aabd66d073"
  uuid2 = "52fcfbcc-32fe-ca21-ecd8-5849044fbca2"

  p = TestRepo.insert!(%Post{bid: uuid1})
  TestRepo.insert_all(Post, [%{bid: uuid1}, %{bid: uuid2}])
  TestRepo.update!(Ecto.Changeset.change(p, %{bid: uuid2})) 
  TestRepo.update_all(from(p in Post, where: p.bid == ^uuid1), set: [public: true])
  TestRepo.all(from(p in Post, where: p.bid == ^uuid1))
  TestRepo.delete_all(from(p in Post, where: p.bid == ^uuid1))

  t = TestRepo.insert!(%Tag{uuids: [uuid1]})
  TestRepo.insert_all(Tag, [%{uuids: [uuid1]}, %{uuids: [uuid2]}])
  TestRepo.update!(Ecto.Changeset.change(t, %{uuids: [uuid2]})) 
  TestRepo.update_all(from(t in Tag, where: t.uuids == ^[uuid2]), set: [uuids: [uuid1]])
  TestRepo.all(from(t in Tag, where: t.uuids == ^[uuid1]))
end
```

The PostgreSQL logs:
Screen Shot 2022-04-12 at 4 59 33 PM

The MySQL logs:
Screen Shot 2022-04-12 at 5 00 03 PM


greg-rychlewski:
I haven't completed this but I got to a point where I wasn't too crazy about the direction and stopped to try to think of a better way. I thought I'd put what I have so far in case other people are curious.

The current attempt (the one pushed now) tries to keep the cast and dump versions of all the parameters and push them all the way to the adapter, where the logging occurs.

The results so far:
1. queries that go through the planner are very simple accommodate 
2. queries not going through the planner (e.g. update, delete) are a bit more complicated due to the way the parameters are separated (example below)

An example of the complications with `update`:

The parameters for `update` queries are separated into two types, fields and filters. The adapter will treat each type separately, possibly removing nil values, and then combine them into a single parameter list. This is from the ecto_sql sql adapter module:

```elixir
def update(adapter_meta, %{source: source, prefix: prefix}, fields, params, returning, opts) do
  {fields, field_values} = :lists.unzip(fields)
  filter_values = params |> Keyword.values() |> Enum.reject(&is_nil(&1))
  sql = @conn.update(prefix, source, fields, params, returning)
  Ecto.Adapters.SQL.struct(adapter_meta, @conn, sql, :update, source, params, field_values ++ filter_values, :raise, returning, opts)
end
```

Where `field_values ++ filter_values` is the final parameter list.  

So to accommodate this behaviour, Ecto has to be able to separate the cast values it sends to this function, see here for a very naive/quick attempt: https://github.com/elixir-ecto/ecto/pull/3869/files#diff-91ed3e8c550063fcf5e2df36020d92405a303c5f5745c0bf643c7dc4c538befbR452. 

And see here for how it would be handled in ecto_sql: https://github.com/elixir-ecto/ecto_sql/pull/391/files#diff-9a458b905f34396e522038e878f83c957e4649c66a4a13f263bb8fcdcbbe744eR193.

The story for insert/delete is similar.

josevalim:
@greg-rychlewski this looks great to me! I wonder if we can simplify the update/delete by doing two things:

1. Remove all nil filters before-hand. When will the filter be `nil`? Can we remove it before we reach the adapter?

2. If we can do this, then we can already send the parameters as `field_values ++ filter_values` and not have to worry about building them.

greg-rychlewski:
Thanks @josevalim! 

It looks to me like the filters are only used for the pk and optimistic locking. [The pk filter will raise if it's `nil`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L964), but I think technically the optimistic locking filter can be `nil` if the user [doesn't set a default value for the field](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L2506). 

The `nil` filters are immediately removed by the adapter (see [here](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/sql.ex#L194) and [here](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/sql.ex#L202)) so I think removing the `nil` cast values in Ecto is ok for adapters that `use Ecto.Adapters.SQL`. For the other ones, would it suffice to have documentation to explain how they can take advantage of the logging, incase they are handling the `nil` values differently?

josevalim:
Yes, I think in this case the bug is in the changeset. We should not really add a `nil` filter. I would expect it either to filter by nil (a bad idea) or raise. So should we change that as well?

greg-rychlewski:
Yeah that's a good idea. I can add it to this PR, for sure. 

From what I can see,  I just need to raise in `Ecto.Changeset.optimistic_lock/3` because the pk filter is already raising and I don't see any other place filters can be added. Do you think raising there + adding a note in the optimistic_lock documentation would be enough?

josevalim:
It is probably a bit too risky to raise, it may be a breaking change, but we can warn if `nil` and not set the filter. :)

greg-rychlewski:
I finished everything except for `insert_all`. I was mostly done with it before I ran into two complications:

1. In Ecto the literal values and the query values are dumped separately instead of being interleaved. This makes it harder to aggregate the casted parameters in the same order as the adapter, because the adapter interleaves them.
2. The parameter counter has a bug when a query value has more than one parameter. The whole query is counted as one parameter because of this: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L99

I was wondering if you would mind if I tackle `insert_all` in a separate PR right after this. The refactoring is a bit more involved than the stuff in this PR and it would be easier for me to reason about it on its own.

Oh and btw, I pushed all the corresponding ecto_sql changes here: https://github.com/elixir-ecto/ecto_sql/pull/391. I added some integration tests to make sure i didn't completely screw up the logging.

josevalim:
Let’s definitely leave insert_all for another PR. Please assign me as a reviewer once this ready for a final review. :)

josevalim (approved):

fuelen:
@greg-rychlewski
I thought I can remove guessing UUIDs from https://github.com/fuelen/ecto_dev_logger after merging this PR, but I see that UUIDs are still binaries in telemetry events. Is this how it supposed to be?

greg-rychlewski:
@fuelen Thanks for pointing that out, it's an oversight. I will push a fix soon.

greg-rychlewski:
@fuelen The fix is now in ecto_sql master branch.

Issue #3870, opened 2022-04-13
`subquery`ing a query that `select`s a literal atom value is broken


jaminthorns:
### Environment

* Elixir version: `1.13.1`
* Database and version: PostgreSQL `13.4`
* Ecto version: Ecto `3.7.2`, Ecto SQL `3.7.2`
* Database adapter and version: Postgrex `0.16.2`
* Operating system: macOS `12.2.1 (21D62)`

### Current behavior

Looking back at the Ecto changelog from release `3.5.0`, it looks like the ability to have literal atom values within queries was added in Ecto 3.5:

> [Ecto.Query] Allow atoms in query without interpolation in order to support Ecto.Enum

I seem to have found a case where this doesn't work correctly. If I select from a table with a literal atom value:

```elixir
from tt in "test_table", select: %{test: tt.test_column, literal_value: :value}
```

The generated query looks like this:

```sql
SELECT t0."test_column" FROM "test_table" AS t0
```

It looks like `literal_value` select is excluded, which I assume is because it will later be merged back into the results, since running the query yields rows with atom values as expected. The issue comes when this query is passed to `subquery`:

```elixir
subquery(from tt in "test_table", select: %{test: tt.test_column, literal_value: :value})
```

The generated SQL attempts to select the literal atom value as a field from the subquery:

```sql
SELECT s0."test", s0."literal_value" FROM (SELECT st0."test_column" AS "test" FROM "test_table" AS st0) AS s0
```

Which breaks when trying to run the query:

```
** (Postgrex.Error) ERROR 42703 (undefined_column) column s0.literal_value does not exist
```

### Expected behavior

Either:

- Querying with a literal atom values should work in all circumstances, even within a subquery. (Preferable)
- Querying with a literal atom values should be banned in subqueries (like maps, lists, tuples and sources).

josevalim:
Can you please try master? I remember tackling some related issues recently.

jaminthorns:
I changed my `mix.exs` to use the `master` branch for both `ecto` and `ecto_sql`:

```elixir
# ...
{:ecto, git: "git@github.com:elixir-ecto/ecto.git", branch: "master", override: true},
{:ecto_sql, git: "git@github.com:elixir-ecto/ecto_sql.git", branch: "master", override: true},
# ...
```

[...]

PR #3874, opened 2022-04-16
Raise on atom literals in subqueries

greg-rychlewski:
Resolves https://github.com/elixir-ecto/ecto/issues/3870

This is my first attempt at changing the planner, so please approve with caution :).

The issue seems to be that the planner wasn't ignoring atom literals when collecting `query.select.fields` for a subquery, causing it to show up in the sql statement.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -398,7 +398,7 @@ defmodule Ecto.Query.Planner do
         if valid_subquery_value?(value) do
           {key, value}
         else
-          error!(query, "maps, lists, tuples and sources are not allowed as map values in subquery, got: `#{Macro.to_string(expr)}`")
+          error!(query, "atoms, maps, lists, tuples and sources are not allowed as map values in subquery, got: `#{Macro.to_string(expr)}`")
         end
     end)
   end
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @greg-rychlewski! The changes look great but I actually think we should raise in this case... because you wouldn't be able to select this field in the parent query.

If we give a step back, our options are to raise or discard silently. The benefit of raising is clearer behaviour but it makes queries harder to reuse as I need to build different selects for when using as subquery. Discard silently can lead to further confusion down the road though, when selecting a literal field or the whole subquery... which can lead to further bug reports. That's why I prefer raising. WDYT?

greg-rychlewski:
That's a really good point about not being able to select on the parent query, completely agreed!

I also prefer raising...it's better to be clear even if it introduces a minor inconvenience in some situations, imo.

lib/ecto/query/planner.ex
```diff
@@ -407,6 +407,8 @@ defmodule Ecto.Query.Planner do
   defp valid_subquery_value?(args) when is_list(args), do: false
   defp valid_subquery_value?({container, _, args})
        when container in [:{}, :%{}, :&] and is_list(args), do: false
+  defp valid_subquery_value?(nil), do: true
+  defp valid_subquery_value?(arg) when is_atom(arg), do: false
```
josevalim, line 411: We need to make sure we still allow booleans too:

```suggestion
  defp valid_subquery_value?(arg) when is_atom(arg) and not is_boolean(arg), do: false
```

josevalim (approved):
One tiny comment and ship it!

PR #3880, opened 2022-04-25
Fix put_query_prefix docs

krns:
See https://hexdocs.pm/ecto/Ecto.Query.html#put_query_prefix%2F2=


josevalim:

Issue #3882, opened 2022-04-27
In a `many_to_many` relationship, updates do not consider `join_where`


thiagopromano:
### Environment

* Elixir version (elixir -v): 1.13.0
* Ecto version (mix deps): 3.8.0
* Database adapter and version (mix deps): postgrex 0.16.2

### Current behavior

Given a Many to Many table with the following association fields:

```elixir
    many_to_many :public_tags, ManyToManyTest.Core.Tag,
      join_through: ManyToManyTest.Core.ProductTag,
      join_defaults: [public: true],
      join_where: [public: true],
      on_replace: :delete

    many_to_many :private_tags, ManyToManyTest.Core.Tag,
      join_through: ManyToManyTest.Core.ProductTag,
      join_defaults: [public: false],
      join_where: [public: false],
      on_replace: :delete
```

When updating tags, the following happens:

We start with a product with no tags:

`Product: [private_tags: [], public_tags[]]`

Then we add a private tag called tag1:

`Product: [private_tags: [tag1], public_tags[]]`

Then we add the same tag as public:

`Product: [private_tags: [tag1], public_tags[tag1]]`

Everything up to here works fine.

Finally, we ask to remove the private tag:

`Product: [private_tags: [], public_tags[tag1]]`

Ecto executes this query, ignoring the `join_where`:

DELETE FROM "products_tags" AS p0 WHERE ((p0."product_id" = $1) AND (p0."tag_id" = $2))

So the resulting product is:

`Product: [private_tags: [], public_tags[]]`

### Expected behavior

The Ecto follows the `join_where` and run the following query:

`DELETE FROM "products_tags" AS p0 WHERE ((p0."product_id" = $1) AND (p0."tag_id" = $2) AND (p0."private" = false))`

## Notes

I have created a simple repo that shows this bug here: https://github.com/thiagopromano/many_to_many_test

The test that shows this happening is here: https://github.com/thiagopromano/many_to_many_test/blob/main/test/many_to_many_test/core_test.exs

Note that this bug does not depend on having two different associations using the same intermediate table. For example the `join_where` could have been `available_for_pro_users: true`.

thiagopromano:
Ecto could also use the `primary_key` of the association table if it exists for deletions.

josevalim:
We should use the `join_where`. We may not have the primary key and we would need to do additional queries to fetch it.

PR #3883, opened 2022-04-28
Respect `join_where` in many to many on_replace deletes

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/3882

### DIFF
--- a/test/ecto/repo/many_to_many_test.exs
+++ b/test/ecto/repo/many_to_many_test.exs
@@ -53,6 +53,7 @@ defmodule Ecto.Repo.ManyToManyTest do
       field :x, :string
       field :y, :binary
       many_to_many :assocs, MyAssoc, join_through: "schemas_assocs", on_replace: :delete
+      many_to_many :where_assocs, MyAssoc, join_through: "schemas_assocs", join_where: [public: true], on_replace: :delete
       many_to_many :schema_assocs, MyAssoc, join_through: MySchemaAssoc, join_defaults: [public: true]
       many_to_many :schema_prefix_assocs, MyAssoc, join_through: MySchemaPrefixAssoc, join_defaults: [public: true]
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
Apologies, I think this is more complicated that I originally thought.

Is it that `join_where` can be a keyword list of arbitrary size and the values can be any of these:

```
nil - which specifies the field must be nil
{:not, nil} - which specifies the field must not be nil
{:in, list} - which specifies the field must be one of the values in a list
{:fragment, expr} - which specifies a fragment string as the filter (see [Ecto.Query.API.fragment/1](https://hexdocs.pm/ecto/Ecto.Query.API.html#fragment/1)) with the field's value given to it as the only argument
or any other value which the field is compared directly against
```

josevalim:
And I am pretty sure you already found the correct place :D

test/ecto/repo/many_to_many_test.exs
```diff
@@ -426,6 +427,28 @@ defmodule Ecto.Repo.ManyToManyTest do
...
+    sample = %MyAssoc{id: 10, x: "xyz"} |> Ecto.put_meta(state: :loaded)
+
+    # removing assoc with join_where
+    %MySchema{id: 3, assocs: [sample], where_assocs: [sample]}
+    |> Ecto.Changeset.change()
+    |> Ecto.Changeset.put_assoc(:where_assocs, [])
+    |> TestRepo.update!()
+
+    # assert 3 where conditions: parent join key, child join key and join_where
+    assert_received {:delete_all, %{wheres: [_, _, _]}}
```
josevalim, line 440: Maybe get the query and call `inspect(...)` on it and match using `=~` on part of the query representation?

greg-rychlewski:
Thank you! I believe this is correct now.

v0idpwn (approved):

PR #3884, opened 2022-04-28
Use Repo.exists?/2 instead of Repo.one/2 in unsafe_validate_unique/4

fuelen:
Just small refactoring


v0idpwn (approved):
LGTM. One could argue that this could cause a regression if some adapter doesn't implement `Ecto.Adapter.Queryable` but implements `one` (and doesn't implement `exists?`). Not sure if we should consider these cases.

josevalim:

PR #3891, opened 2022-05-03
Add integration test for jsonb comparison with boolean values

leandrocp:
Follow up of https://github.com/elixir-ecto/ecto_sql/pull/399

### DIFF
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -253,18 +253,6 @@ defmodule Ecto.Integration.TypeTest do
   end
 
   @tag :map_type
-  test "typed boolean map" do
-    TestRepo.insert!(%Post{meta: %{"enabled" => true}})
-    TestRepo.insert!(%Post{meta: %{"extra" => [%{"enabled" => false}]}})
-
-    assert TestRepo.all(from p in Post, where: p.meta["enabled"] == true, select: p.meta) ===
-      [%{"enabled" => true}]
-
-    assert TestRepo.all(from p in Post, where: p.meta["extra"][0]["enabled"] == false, select: p.meta) ===
-      [%{"extra" => [%{"enabled" => false}]}]
-  end
-
-  @tag :map_type
   test "map type on update" do
     post = TestRepo.insert!(%Post{meta: %{"world" => "hello"}})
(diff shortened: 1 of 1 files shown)


leandrocp:
@josevalim what's the approach to deal with

```
   +integration-test *failed* |   1) test typed boolean map (Ecto.Integration.TypeTest)
   +integration-test *failed* |      /src/ecto/integration_test/cases/type.exs:256
   +integration-test *failed* |      ** (Ecto.QueryError) Tds adapter does not support json_extract_path expression, use fragment with JSON_VALUE/JSON_QUERY in query:
   +integration-test *failed* |      
   +integration-test *failed* |      from p0 in Ecto.Integration.Post,
   +integration-test *failed* |        where: p0.meta["enabled"] == true,
   +integration-test *failed* |        select: p0.meta
   ```
   
   ?

File: integration_test/cases/type.exs [hunk 1]
```diff
@@ -252,6 +252,18 @@ defmodule Ecto.Integration.TypeTest do
            [%{"foo" => 1.0, "bar" => 416500.0}]
   end
 
+  @tag :map_type
```
wojtekmach, line 255: please add this tag so it's excluded on TDS

```suggestion
  @tag :map_type
  @tag :json_extract_path
```

integration_test/cases/type.exs
```diff
@@ -307,12 +318,16 @@ defmodule Ecto.Integration.TypeTest do
     assert TestRepo.one(from o in Order, select: o.meta["'single quoted'"]) == "bar"
     assert TestRepo.one(from o in Order, select: o.meta["';"]) == nil
     assert TestRepo.one(from o in Order, select: o.meta["\"double quoted\""]) == "baz"
+    assert TestRepo.one(from o in Order, select: o.meta["enabled"]) == true
```
leandrocp, line 321: `== true` is redundant but trying to make it explicit if that makes sense

integration_test/cases/type.exs
leandrocp: Thanks! Decided to reuse `test "json_extract_path with primitive values"` instead of adding a new test case.

josevalim:

PR #3900, opened 2022-05-08
Fix named typed selection query

### LINKED ISSUE #3899: Unable to cast a value when specifying its type is a name-bound column in a `select/3` expression.
### Environment

* Elixir version (elixir -v):

> Erlang/OTP 24 [erts-12.3.1] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit]
> 
> Elixir 1.13.4 (compiled with Erlang/OTP 24)

* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.):

> psql (PostgreSQL) 13.3 (Debian 13.3-1.pgdg100+1)

* Ecto version (mix deps):

> ecto 3.8.1 (Hex package) (mix)
> locked at 3.8.1 (ecto) f1b68f8d

> postgrex 0.15.5 (Hex package) (mix)
> locked at 0.15.5 (postgrex) ed90c81e

* Operating system: 

> Distributor ID: Ubuntu
> Description:    Ubuntu 20.04.4 LTS
> Release:        20.04
> Codename:       focal

### Current behavior

I was composing query expressions with named bindings and came across a bug when trying to cast the result to the underlying column's type, specifically:

```elixir
from(t in Task, as: :task)
|> join(:inner, [task: t], assoc(t, :current_state), as: :task_state)
|> select([task_state: ts], type(fragment("date_trunc('hour', ?)", min(ts.planned_start)), ts.planned_start))
|> Repo.one()
```

I managed to isolate the error into the following test, running on this repo's master branch `65b1c65`, which errors on the last assert for the `named_typed_column` query:
```elixir
  test "select cast to named column type" do
    named_plain_column = from(Schema, as: :s) |> select([s: s], s.x) |> plan()
    assert all(named_plain_column) == ~s{SELECT s0."x" FROM "schema" AS s0}

    named_typed_atom = from(Schema, as: :s) |> select([s: s], type(s.x, :integer)) |> plan()
    assert all(named_typed_atom) == ~s{SELECT s0."x"::bigint FROM "schema" AS s0}

    positional_typed_column = from(Schema, as: :s) |> select([s], type(s.x, s.x)) |> plan()
    assert all(positional_typed_column) == ~s{SELECT s0."x"::bigint FROM "schema" AS s0}

    named_typed_column = from(Schema, as: :s) |> select([s: s], type(s.x, s.x)) |> plan()
    assert all(named_typed_column) == ~s{SELECT s0."x"::bigint FROM "schema" AS s0}
  end
```

```

  1) test select cast to named column type (Ecto.Adapters.PostgresTest)
     test/ecto/adapters/postgres_test.exs:48
     ** (ArgumentError) unsupported type `{{{:., [], [{:__aliases__, [alias: false], [:Ecto, :Query, :Builder]}, :count_alias!]}, [], [{:query, [], Ecto.Query.Builder}, :s]}, :x}`. The type can either be an atom, a string or a tuple of the form `{:map, t}` or `{:array, t}` where `t` itself follows the same conditions.
     code: assert all(named_typed_column) == ~s{SELECT s0."x"::bigint FROM "schema" AS s0}
     stacktrace:
       (ecto_sql 3.8.1) lib/ecto/adapters/postgres/connection.ex:1377: Ecto.Adapters.Postgres.Connection.ecto_to_db/1
       (ecto_sql 3.8.1) lib/ecto/adapters/postgres/connection.ex:754: Ecto.Adapters.Postgres.Connection.expr/3
       (ecto_sql 3.8.1) lib/ecto/adapters/postgres/connection.ex:1320: Ecto.Adapters.Postgres.Connection.intersperse_map/4
       (ecto_sql 3.8.1) lib/ecto/adapters/postgres/connection.ex:362: Ecto.Adapters.Postgres.Connection.select/3
       (ecto_sql 3.8.1) lib/ecto/adapters/postgres/connection.ex:135: Ecto.Adapters.Postgres.Connection.all/2
       test/ecto/adapters/postgres_test.exs:64: Ecto.Adapters.PostgresTest.all/1
       test/ecto/adapters/postgres_test.exs:56: (test)

```

### Expected behavior

It should be able to execute the query when referencing a type from a named binding in a `select/3` expression.

03juan:
Calling type in a select expression with named bindings causes downstream errors as it seems the `Query.Builder.count_alias!/2` function is not being applied correctly.

Fixes #3899

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1707,7 +1707,13 @@ defmodule Ecto.Query.Planner do
     type!(kind, query, expr, ix, field, allow_virtuals?)
   end
 
-  defp field_type!(kind, query, expr, {{{:., _, [_, :count_alias!]}, _, [_, name]}, field}, allow_virtuals?) do
+  defp field_type!(
+         kind,
+         query,
+         expr,
+         {{{:., _, [Ecto.Query.Builder, :count_alias!]}, _, [_, name]}, field},
+         allow_virtuals?
+       ) do
     ix = Ecto.Query.Builder.count_alias!(query, name)
     type!(kind, query, expr, ix, field, allow_virtuals?)
   end
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1707,6 +1707,11 @@ defmodule Ecto.Query.Planner do
     type!(kind, query, expr, ix, field, allow_virtuals?)
   end
 
+  defp field_type!(kind, query, expr, {{{:., _, [_, :count_alias!]}, _, [_, name]}, field}, allow_virtuals?) do
+    ix = Ecto.Query.Builder.count_alias!(query, name)
+    type!(kind, query, expr, ix, field, allow_virtuals?)
+  end
+
   defp field_type!(_kind, _query, _expr, type, _) do
     type
   end
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -826,7 +826,7 @@ defmodule Ecto.Query.Builder do
       for {:named, key, name} <- vars do
         {key,
          quote do
-           Ecto.Query.Builder.count_alias!(query, unquote(name))
+           unquote(__MODULE__).count_alias!(query, unquote(name))
          end}
       end
 
--- a/test/ecto/query/inspect_test.exs
+++ b/test/ecto/query/inspect_test.exs
@@ -391,6 +391,15 @@ defmodule Ecto.Query.InspectTest do
            ~s{from p0 in Inspect.Post, select: [:foo]}
   end
 
+  test "select bound name type after planner" do
+    query =
+     from(x in Post, as: :p)
+      |> select([p: p], type(p.visits, p.visits))
+      |> plan()
+
+    assert i(query) == ~s{from p0 in Inspect.Post, as: :p, select: type(p0.visits, :integer)}
+  end
+
   test "params" do
     assert i(from(x in Post, where: ^123 > ^(1 * 3))) ==
            ~s{from p0 in Inspect.Post, where: ^123 > ^3}


03juan:
This fixes the issue at the `Ecto.Query.Planner.field_type!/5` level but I'm not sure if this is the correct place. Is there an underlying issue in the Builder's escape/apply chain?

Removing the fix causes this error in the new test:

```elixir
1) test select bound name type after planner (Ecto.Query.InspectTest)
   test/ecto/query/inspect_test.exs:394
   Assertion with == failed
   code:  assert i(query) == ~s"from p0 in Inspect.Post, as: :p, select: type(p0.visits, :integer)"
   left:  "from p0 in Inspect.Post, as: :p, select: type(p0.visits, {Ecto.Query.Builder.count_alias!(query, :p), :visits})"
   right: "from p0 in Inspect.Post, as: :p, select: type(p0.visits, :integer)"
```

03juan:
@greg-rychlewski do you mind taking a look at your convenience? 🙏

greg-rychlewski:
@03juan This is the exact same change I was thinking :). Without the fix it falls to this pattern:

```elixir
 defp field_type!(_kind, _query, _expr, type, _) do
    type
  end
```
And that's why you see the type as `{Ecto.Query.Builder.count_alias!(query, :p), :visits}`.

LGTM but just want to ping @josevalim to make sure I'm not missing anything. I'm a bit green in this area of the codebase.

03juan:
> And that's why you see the type as `{Ecto.Query.Builder.count_alias!(query, :p), :visits}`.

For sure, I'm just wondering whether we could be applying the quoted code, generated in [`calculate_named_binds/2`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder.ex#L819), properly somewhere else in the apply chain.


> I'm a bit green in this area of the codebase.
Totally with you on this, and I'm out of my depth in macros too!

Thanks for the comment

lib/ecto/query/planner.ex
```diff
@@ -1707,6 +1707,11 @@ defmodule Ecto.Query.Planner do
     type!(kind, query, expr, ix, field, allow_virtuals?)
   end
 
+  defp field_type!(kind, query, expr, {{{:., _, [_, :count_alias!]}, _, [_, name]}, field}, allow_virtuals?) do
```
josevalim, line 1710:
```suggestion
  defp field_type!(kind, query, expr, {{{:., _, [__MODULE__, :count_alias!]}, _, [_, name]}, field}, allow_virtuals?) do
```

?

josevalim:
It is a bit weird but it seems to be correct. I have just added a tiny comment. If it passes with my comment, we are good to go!

lib/ecto/query/planner.ex
josevalim: If it doesn't pass with my comment, then change `calculate_named_binds` to this:

    unquote(__MODULE__).count_alias!(query, unquote(name))

03juan: Wouldn't `__MODULE__` in `defp field_type!(... {{{:., _, [__MODULE__, :count_alias!]},  ...) do` refer to `Ecto.Query.Planner`?

I had to make both changes for it to pass but it still feels like a bandage over another issue...

The intention seems to be that some other part of the code would apply the `count_alias!` before this point, calculating the index and matching on the expression before this one.

However if you're happy with the changes so am I. I have some other tests that exercise the named select-type combination in a few ways if you're interested in them?

https://github.com/03juan/ecto_sql/blob/named-typed-column/test/ecto/adapters/postgres_test.exs#L112

03juan (commented):
Can add more tests if desired
https://github.com/03juan/ecto_sql/blob/named-typed-column/test/ecto/adapters/postgres_test.exs#L113

lib/ecto/query/planner.ex
josevalim: I agree that it is a bit awkward but I think doing the pre-processing after would be even more work. I will explore it a bit after merge!

josevalim:

josevalim:
I was able to clean it up a bit further in a later commit. The real issue was that we were escaping all types, which made it so the ast arrived to the planner, not its value!

03juan:
Thank you for the feedback! I'm happy there was an idiomatic fix, this one wasn't sitting right. 👏

PR #3903, opened 2022-05-13
Unnecessary updated_at update when changes are removed in prepare_change

juanperi:
This is an issue in the form of a failing tests PR

I find myself doing some housekeeping in the prepare changes stage of the lifecycle. In some particular scenarios, such housekeeping means removing some changes from the changeset.
Semantically, the changeset after and before the prepare_changes is the same, but I'm taking that opportunity to do some duplication cleanup

That's how I found the failing test case.
As far as I understand by reading the code, the prepare_changeset is supposed to be able to modify the changeset (adding or removing things, or even turning it into a changeset that did not pass validations.
But I found that the feature that controls if an empty changeset should update the autogenerated fields is not compatible with removing all the changes from the changeset. 
https://github.com/elixir-ecto/ecto/blob/2f0548c3532b6c0f69a42f5eff142d2e1663aa9c/lib/ecto/repo/schema.ex#L447 uses the "force?" parameter, which is set on the state of the changeset before it goes through the prepare changes call

So, the question is: Is this behavior a bug, or outside of the scope of when autogenerated fields should kick in?

My workaround for now has been going through all the cast_assoc functions for the related entities and set `force_update_on_change: false`, and add a new prepare_changes that deals with updating the updated_at field. But as this is per changeset and not at a schema level, it might be error prone

Thank you for your time reading this!!!


josevalim:
Thanks for the description.

I don't think we should fix this because `prepare_changes` is only invoked if the process of insert/updating/deleting has already started. To me, you should instead do the clean-up as you create/modify the changeset, before you invoke the repository operation.

Skipping the update after prepare changes can also lead to other issues. For example, what if a prepare_changes callback is issuing a query direct to the database as a side-effect? Are you sure you want to skip/revert that if later on the other change is removed?

That's always going to be the issue with callbacks, which is why we kept the callback API limited, and I think the best option here is indeed to rely less on callbacks.

juanperi:
Thanks for your thoughtful response.
I was lead to believe that this was far game, because of the structure of the code:
In https://github.com/elixir-ecto/ecto/blob/2f0548c3532b6c0f69a42f5eff142d2e1663aa9c/lib/ecto/repo/schema.ex#L435 the code is checking again if the changeset is valid. the only thing that can change this state is (as far as I understand) the prepare_changes. In that scenario, we are in the same situation as the one you described before. What would happen if an update was already issued against the repo, but then the changeset is converted into an invalid one? The transaction should be rolled back
And then in line https://github.com/elixir-ecto/ecto/blob/2f0548c3532b6c0f69a42f5eff142d2e1663aa9c/lib/ecto/repo/schema.ex#L447, the code checks again that there are no changes, leading me to believe that updating the changes was supported.

In any case, I understand your point.
Migrating these housekeeping to the changeset itself (before the prepare_update) is possible, but tricky taking into account user feedback when facing errors. But can probably be done splitting the logic into:
```
changeset = build_changeset(%SomeChangeset{}, params)
if changeset.valid?
  changeset = housekeeping(changeset)
  Repo.insert(changeset)
end
```

Have a nice day José!!

josevalim:
>  the code checks again that there are no changes, leading me to believe that updating the changes was supported.

Right, but it is also checking for the original changes on purpose.

I agree that we can already abort and revert today, the concern is that aborting and reverting for invalid changesets are expected, but less for empty changes.

> Have a nice day José!!

You too!

juanperi:
I'm going to close this, as I'm not going to purse the change in the functionality.
But just a clarification on something you just said:
> Right, but it is also checking for the original changes on purpose.

if you are refering to line 477, it is not really checking for the original (understanding original as the changeset before executing the prepare_changes callback), but the resulting changeset. Is that a bug then?

removing unrelated lines, this is the flow:
```
    # changeset here is the original one
    if changeset.changes != %{} or force? do
      wrap_in_transaction(adapter, adapter_meta, opts, changeset, assocs, embeds, prepare, fn ->
        user_changeset = run_prepare(changeset, prepare)
        {changeset, parents, children} = pop_assocs(user_changeset, assocs)
        # changeset here is the processed one
        changeset = process_parents(changeset, user_changeset, parents, adapter, assoc_opts)
        # `original` here does not reference the original changeset changes
        original = changeset.changes |> Map.merge(embeds) |> Map.take(fields)
        if original != %{} or (autogen != [] and force?),
```

juanperi:
and by the way, I'm not expecting the empty changes to do a rollback. I thought that empty changes would do a `:noop` as that is already supported

juanperi:
I think i'm this 👌  close to turn into annoying, but I think there is something off in the calculation of the "force?" flag.

Even If I followed the steps we discussed in the previous conversation, where instead of using prepare_chages I deal directly with the changeset 
(updated test case)
```
  test "does not update updated_at when associated record changes are removed from the changeset" do
    company =
      TestRepo.insert!(%Company{
        updated_at: ~N[2000-01-01 00:00:00],
        offices: [%Office{id: 1, name: "1"}]
      })

    changes = %{offices: [%{id: 1, name: "updated"}]}

    updated_company =
      company
      |> Ecto.Changeset.cast(changes, [])
      |> Ecto.Changeset.cast_assoc(:offices)
      |> Ecto.Changeset.delete_change(:offices)
      |> TestRepo.update!()

    assert company.updated_at == updated_company.updated_at
  end
```
the test still fails.
the problem is that the "force" opt is calculated when we do the cast_assoc with some changes. But if afterwards we remove the changes in the changeset itself (not in the callback) the force param is still there, and so the updated_at field is still updated
So, the only option here is to use the workaround I mentioned before, using `force_update_on_change: false` and dealing manually with the updated_at field

juanperi:
hey @josevalim , not sure if my last comment fell through the cracks. 
If you want I can open a new issue, as the effect is the same as the original issue, but in this case it is not tied to callbacks but pure changeset manipulation

josevalim:
Thanks for the ping, your follow up fell through the cracks indeed.

Luckily I think we can fix the current behaviour for your second issue. Instead of computing the `force` early on, we can move its computing to `Ecto.Repo.Schema` and set `force: true` only if there is a relation change. We will need to introduce an option to turn it off though (and swap the `force_update_on_change` to false in the changeset, possibly deprecating it).

Would you like to work on a PR for that?

juanperi:
I can work on a PR indeed, now that I know it's a valid use case.
I'll try to tackle it this weekend.

juanperi:
I've been thinking on this a bit. Do you really think that we need to provide a different option and deprecate the current one?
To me it looks more like a bug fix. It was calculating the force option too early in the lifecycle, and so we would be moving it further down the line accommodating manual changeset manipulations.
If it is considered a bugfix, then we don't need new options nor deprecations, and the option would still be doing what is advertised, which is to touch the parent if there are changes in the relation.
There is indeed a change of behavior, but i would think it's still conforming to the expectation

josevalim:
Unless you see another way, we need to add the option to a new API (Repo.update) and the current name doesn’t fit there.

josevalim:
The current name, force_update_on_change doesn’t imply the child/parent relationship on an API such as Repo.update.

juanperi:
Moving it to the repo update and deprecating it from the schema is the part that i'm not convinced about. 

I can think of two points for which i think the current API is right:
1. In the changeset we can control if we want the parent to be updated of each relation, passing the configuration `force_update_on_change` when we do the cast_assoc. This way, we have the the granular control over each one of the relations in a given changeset. 
2. The functionality is constrained and repeatable. No matter where we use the changeset, the fact that we want to update the parent of a relationship is always ensured.

Point 2 is opposed to what _I understood_ that you meant about moving the configuration to `Repo.update`
From what I understand, you meant something like 
`Repo.update(changeset, update_timestamp_if_any_related_model_is_updated_: true`
(there is probably a better name out there for this option 😹)
But if we go that route, it would be global to the changeset. How would it work with nested changesets?, or maybe two separate related entities, but where you want to touch the timestamp of the parent only if one of them changes but not the other?

If we don't take into account the problem being consistent with a changeset that can be manipulated, the current version of the code complies with both points.

So, what do you think the approach should be:

1. We keep the control of when to touch the parent at the changeset level. (if you have two related entities, you can say that one forces a touch but not the other)
   This will require a new `repo_opt`, and pass information of which relationships we are tracking (sort of, `touch_if_any_of_the_following_relationships_change: [:books, :authors]`)
2. We keep the control of when to touch the parent at the changeset level, but irrespective of which relation changed.
   This simplifies the `repo_opt` parameter, but I'm not sure we currently have an API to do so at the changeset level.
3. We remove the functionality from the changeset, and make it global (if you have two related entities, any of them would trigger the touch). The same changeset might have different behavior based on the parameters you pass to Repo.update

In any of these options, I might be missing something with nested changesets though. I think the version that we set the flag in the changeset would work seamlessly with nested changesets, but I'm not soure about the version where we move it directly to the repo_update API

josevalim:
You are correct on all accounts. I think option 1 works great and we can even call it force_update_on_change as well and document that it expects a list of assocs/embeds.

juanperi:
Hey, I took a first stab at the proposed changes. There are a couple of points where I have doubts though:
1. Is the way in which i'm looking if the children have changes correct? (function `defp force_update_on_children_change`)
2. With the current implementation, nested changesets tracking works if the new opt comes from repo_opts, as the changesets keep track of their own repo opts. But you have no control over it if calling directly from the `Repo.update`
eg: Repo.update cannot receive something like `force_update_on_children_change: [offices: :meeting_rooms]`
Is that a usecase we want to support? As it stands right now, you can only track first level childs when calling `Repo.update`
3. I think that most of the tests I added to the autogenerate file should go to the repo test file. But i'll wait for the answer to point 2.

lib/ecto/repo/schema.ex
```diff
@@ -470,6 +471,14 @@ defmodule Ecto.Repo.Schema do
     {:error, put_repo_and_action(changeset, :update, repo, tuplet)}
   end
 
+  defp force_update_on_children_change(nil, _children) do
+    false
+  end
+
+  defp force_update_on_children_change(forced_relations, children) do
+    !!Enum.find(children, fn {assoc, _} -> assoc.field in forced_relations end)
```
josevalim, line 479:
```suggestion
    Enum.any?(children, fn {assoc, _} -> assoc.field in forced_relations end)
```

josevalim:
@juanperi looks great! However, I think you also forgot to check for `parents`. For example, in a has_one/belongs_to relationship, a comment is always the child, but you may want to update the comment if the associated post changes. That's why I didn't use `children` in the name. We either need to call relation or not use the word at all.

josevalim:
@juanperi we should not support nested associations directly. If such cases, you should update the "whole tree". I.e. force update offices when meeting_rooms so the "parent" also changes.

juanperi:
> @juanperi looks great! However, I think you also forgot to check for `parents`. For example, in a has_one/belongs_to relationship, a comment is always the child, but you may want to update the comment if the associated post changes. That's why I didn't use `children` in the name. We either need to call relation or not use the word at all.

you are right, I didn't think on this use case at all. Will update the PR to support it. 
I added the word because the opt looked very weird without it. we would have `force` and `force_update_on_change` in the repo opts, which I think doesn't tell you much in the context of the repo. It is different though in the context of the changeset. But `force_update_on_relation_change` should work

greg-rychlewski:
Sorry hopefully I'm not being dense, but couldn't you leave out the new option from `Repo.update` and track the relations internally?

If `:force_update_on_change` is used when casting any of the relations, add it to the list in `repo_opts`. If `Ecto.Changeset.delete_change` is used on any of the relations being tracked, delete it from the list in `repo_opts`.  Then when deciding to force or not, check if the list is empty.

josevalim:
@greg-rychlewski that approach would work but I think having the repository check for an empty list or not is a weird contract? I think dealing with `delete_changes` is perhaps more work?

juanperi:
> ... couldn't you leave out the new option from Repo.update and track the relations internally?
If we use repo_opts, it needs to be part of the API for repo (unless we want to make it a non documented option passed to Repo.update.
It doesn't matter if we check for empty list or the values. If it's there, it needs to be an opt to `Repo.update`

> If `Ecto.Changeset.delete_change` is used on any of the relations being tracked, delete it from the list in repo_opts
I would honestly rather track the relations, and check the status in the `Repo.update` call, moving it to the last possible moment. That way, it doesn't matter if the change was done directly on the changeset, or trough `delete_change`
The way I see it, it is a configuration. When casting, we mark it as "hey, if any of the records of this relation changed, update the timestamp of the main entity". This configuration is independent of the current state of the changeset.
In fact, looking at it in this way, maybe there is an extra change to incorporate. Right now, the tracking in `repo_opts` is only added if there are changes at the moment of the casting in the relation. Shouldn't I remove that and always store the new configuration? So, if there are extra changes done to the relation through changeset manipulation, the option will still be there and it will work as expected when updating the DB.
I'm not fully convinced of this last one option, but I think it would make it consistent

lib/ecto/repo/schema.ex
```diff
@@ -444,7 +444,14 @@ defmodule Ecto.Repo.Schema do
 
           # If there are no changes or all the changes were autogenerated but not forced, we skip
           {action, autogen} =
-            if original != %{} or (autogen != [] and force?),
+            if original != %{} ||
```
juanperi, line 447: I refactored this part to make the evaluation short-circuiting.
This was because I'm doing some array traversal inside the `forced_update?` function and I'm trying only to run it when necessary

test/ecto/repo/autogenerate_test.exs
```diff
@@ -220,6 +242,93 @@ defmodule Ecto.Repo.AutogenerateTest do
...
+    updated_company =
+      company
+      |> Ecto.Changeset.cast(changes, [])
+      |> Ecto.Changeset.cast_assoc(:offices)
+      |> TestRepo.update!()
+
+    assert company.updated_at != updated_company.updated_at
+  end
+
+  test "updates updated_at when there are parent records with changes" do
```
juanperi, line 263: this test is passing, but because of the wrong reasons and I didn't figure out why.
In line https://github.com/elixir-ecto/ecto/pull/3903/files#diff-91ed3e8c550063fcf5e2df36020d92405a303c5f5745c0bf643c7dc4c538befbR447, there are always changes in the `original` variable: `%{company_id: 1}`
I'm not sure why there is always that value there, but it means that is not going through the path of the new code

test/ecto/repo/autogenerate_test.exs
```diff
@@ -13,18 +13,35 @@ defmodule Ecto.Repo.AutogenerateTest do
...
+    end
+  end
+
   defmodule Office do
     use Ecto.Schema
 
     schema "offices" do
       field :name, :string
-      belongs_to :company, Company
+      belongs_to :company, Ecto.Repo.AutogenerateTest.Company
```
juanperi, line 36: The full qualified name is necessary because _I think_ that the module company is not yet defined when compiling `Office` and so the insert call fails saying that the module `Company` is not defined

greg-rychlewski:
> If we use repo_opts, it needs to be part of the API for repo (unless we want to make it a non documented option passed to Repo.update.
> It doesn't matter if we check for empty list or the values. If it's there, it needs to be an opt to `Repo.update`

Oh I see why it was in both places now, thank you. Having the option in two places is a bit confusing to me...here is where I was having difficulty, in case you find any use for the feedback:

- With these changes `:force_update_on_change` in `Ecto.Changeset` is being used to track the relations, but only in one direction (addition, but not removal). I think the end goal for the user is similar to the `changes` field on the changeset. When the user adds/removes changes the update function checks if there is anything there: `if changeset.changes != %{} ... do`. 
-  It's a bit confusing having the option in both modules, because you could do all the work setting the option in the changeset but then overwrite it completely with the option in the update function. It could also be unclear to the user why both exist and how they interact. The interaction part could be cleared up with docs, but it feels a bit like it's too many choices to me.

I could be mistaken....but my $0.02:

It seems cleaner to have the option in only one of the modules. I would probably try to keep the option in the `Changeset` module, to minimize the impact on the users. I don't think it's a lot of work, but I admit I didn't look too closely. You should just have to keep another map around where the keys are the assocs. You already have the addition done and the deletion should just be a one liner, unless I'm way off base.

It's a bit of a weird contract like Jose said. But it seems like lesser of 2 evils to me personally.

greg-rychlewski:
>The autogenerated functionality lives in the Repo. So, even if we set it from the changeset, it needs to go through the repo, and the only way to do so is through a repo opt.

I probably didn't express myself clearly before. What I was trying to say is that we have a field in the changeset to track if there were any direct changes and the repo acts on it. Similarly we could have a field that checks if any relations were changed that force an update and the repo would act on it. 

This seems simpler/more intuitive to me than using `repo_opts`, for the reasons I specified before. But of course I could be in the minority.

juanperi:
Hey @greg-rychlewski, I don't have any strong opinion on opening up the functionality to be used directly as a parameter to the repo. I thought it was the only way but i'm open to do it differently.
What you mention of the specific field to track changes to the relation is interesting. Can you elaborate on which one is such a field? I only saw `changes`, but that one would not be the place to store the configuration

And one extra thought. I don't think we should store if a relation changed on any given time. I would rather store the configuration that we want to touch the parent if a given relation changes, but leave figuring out if it did change at the very end when executing the `Repo.update` code. If we don't do it this way, we open up ourselves to bugs when the changeset is manipulated outside of `delete_change`

So, in my mind instead of storing
`related_entities_that_changed: [:room]`, and deleting from that list if `delete_change(:room)` is called, I would store
`touch_parent_if_related_entity_has_changes: [:room]` and do nothing in this configuration if `delete_change(:room)` is called

This might be attainable with your approach. If you point me to the field you were considering for storing this, i'll take a look

greg-rychlewski:
Oh it would have to be a new field on the changeset. Nothing exists right now to do it. So you would just do what you're doing now but take it out of `repo_opts`.

>I would rather store the configuration that we want to touch the parent if a given relation changes, but leave figuring out if it did change at the very end when executing the Repo.update code. If we don't do it this way, we open up ourselves to bugs when the changeset is manipulated outside of delete_change

I don't think it's a bug if users are doing something they shouldn't. If anything this pushes people towards the correct behaviour.  

But besides that, I don't believe one-way tracking will give correct behaviour. If you add the same field twice, the first with `force_update_on_change = true` and the second with `force_update_on_change = false` then you're going against the user's wishes because it's the tracking list is not updated.

josevalim:
> But besides that, I don't believe one-way tracking will give correct behaviour. If you add the same field twice, the first with force_update_on_change = true and the second with force_update_on_change = false then you're going against the user's wishes because it's not removed from the tracking list.

This is a good point. I can cast once with force, then have no changes, then cast again without force (but changes) and the first one will in. I think the best option for now is to document that, if you cast changes and then you force an update, and then you delete that association, the force will remain. You can remove the force flag yourself if you are inclined.

juanperi:
>  I can cast once with force, then have no changes, then cast again without force (but changes) and the first one will in

That makes sense, but is easily solvable. When processing the cast, if `force` is configured, we add it to the list. If the force is not used, we delete it from the list. That way, the last cast is the one that mandates the behavior of the changeset on update. I've just pushed a commit in this direction to the PR

> I'm assuming you're talking about playing around with the internals of the changeset struct instead of using the functions in the changeset moduIe? I don't think it's a bug if users are doing something they shouldn't. If anything this pushes people towards the correct behaviour.

Well, I'm indeed going through the `changes` field and modifying it in some scenarios. In a subset of those scenarios, it just happens that all the changes are dropped. But as far as I understand, it is not doing something that I shouldn't. I am not aware of any other mechanism to modify changesets.

So, at this point I think we should decide if we want to go in the direction of the `repo_opts` or to craft a new field in the changeset.
I'm not sure that I like adding a new field to changeset just for this purpose, it seems like an overkill to me. But I wouldn't be against it either.
If we do add it, it would be for the sake of using it in the `Repo.update` call, without passing through the `opts` parameter. Do you think it is worth it?

josevalim:
> That makes sense, but is easily solvable. When processing the cast, if force is configured, we add it to the list. If the force is not used, we delete it from the list. That way, the last cast is the one that mandates the behavior of the changeset on update. I've just pushed a commit in this direction to the PR

Right, but that's up to the interpretation of the option. I could expect the last one to not win, given the first stuck.

It overall sounds too complex so I am thinking documenting the behaviour may be the best option after all. :(

greg-rychlewski:
>  I am not aware of any other mechanism to modify changesets.
I could be completely misunderstanding, but there is an `Ecto.Changeset` module with a lot of functions to modify the changeset. I assumed these are meant to be used instead touching the internals of the struct. 

>If we do add it, it would be for the sake of using it in the `Repo.update` call, without passing through the `opts` parameter. Do you think it is worth it?
Personally I do, because I feel like it's the only option that makes sense to me. But you and Jose made good points against it. 

Going back to the beginning of this issue, it was raised because changes were being deleted without undoing the force command. So it looks like the missing like is having a way to track which changes are causing the force.

juanperi:
> Going back to the beginning of this issue, it was raised because changes were being deleted without undoing the force command. So it looks like the missing like is having a way to track which changes are causing the force.

As I understand it, the problem with existing approach is that it calculates the `force` value too early in the lifecycle. It should calculate it when there are no more changes possible in the changeset. And that is inside the `Repo.update` call.
What you are mentioning I think is addressed in my PR, where I added the list of relations that should force to the repo_opts? that's the simplest solution, conforming with the current approach.
The part that we were not sure about is if it should be a new `Repo` opt or not.

I would say that adding a new field to the changeset is too much work for the functionality. So, if we don't want the current approach with the `force_update_on_relation_change`, we should drop this PR and document that if you modify the changeset after the cast, the force flag might be outdated.

I do think that there is merit on having the repo doing this calculation as it is the only place where we are sure that the changeset is not further manipulated, and so it needs the new `Repo` opt. But I would be happy to submit the PR to the docs clarifying this case if you think it's not worth it.

> It overall sounds too complex so I am thinking documenting the behaviour may be the best option after all. :(

do you mean dropping the whole PR, or documenting that as long as there is one `force_update_on_change` when casting a relation multiple times it will take precedence, no matter if it's the first or the last?

josevalim:
Closing this per the above. Sorry @juanperi that this went nowhere but I believe this to be the right call.

juanperi:
hey! no issues!!
thank you for your time discussing and reviewing it.

for posterity, this is what I had to do as a workaround in all the affected changesets

> My workaround for now has been going through all the cast_assoc functions for the related entities and set force_update_on_change: false, and add a new prepare_changes that deals with updating the updated_at field. But as this is per changeset and not at a schema level, it might be error prone

Issue #3904, opened 2022-05-18
Regression when fragment |> filter in Ecto.Query


kaaboaye:
### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 24 [erts-12.2.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]
Elixir 1.13.3 (compiled with Erlang/OTP 24)
```

* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): Postgres 12
* Ecto version (mix deps): 3.8.2, 3.8.3
* Database adapter and version (mix deps): ecto_sql: 3.8.1
* Operating system: M1 MacOs

### Code snippet

```ex
  defp customer_income_query(customer_id) do
    from(entity in SomeEntity,
      # some where, group_by, and having,
      select: %{
        notes_for_customer:
          fragment("array_agg(distinct ?)", entity.note_for_customer)
          |> filter(not is_nil(entity.note_for_customer))
          |> coalesce(fragment("'{}'"))
      }
    )
  end
```

### Current behavior

```
** (exit) an exception was raised:
    ** (Ecto.SubQueryError) the following exception happened when compiling a subquery.

    ** (Ecto.QueryError) filter(...) expects the first argument to be an aggregate expression, got: `filter(
      fragment({:raw, "array_agg(distinct "}, {:expr, &0.note_for_customer()}, {:raw, ")"}),
      not is_nil(&0.note_for_customer())
    )` in query:
    
    from b0 in SomeEntity,
      # where, group_by, and having,
      select: %{
        # some other fields
        notes_for_customer:
          coalesce(
            filter(
              fragment("array_agg(distinct ?)", b0.note_for_customer),
              not is_nil(b0.note_for_customer)
            ),
            fragment("'{}'")
          ),
        # some other fields
    }

```

### Expected behavior

Expect it to work, and it works at ecto 3.7.2 and below.

josevalim:
A PR is welcome! We likely are missing a clause somewhere that allows fragments in filters. You can likely find where in the source by checking where we raise the error message. :)

kaaboaye:
I've added failing test and it looks like it's caused by this https://github.com/kaaboaye/ecto/commit/ef442e2e517b17be46d40e207232252d06ffffbb commit

josevalim:
Nice! We should allow fragments there too!

kaaboaye:
[...]

PR #3905, opened 2022-05-18
add fragment filter tests

kaaboaye:
Adds failing test `filter from fragment compiles` for https://github.com/elixir-ecto/ecto/issues/3904

### DIFF
--- a/test/ecto/query/planner_test.exs
+++ b/test/ecto/query/planner_test.exs
@@ -1505,4 +1505,26 @@ defmodule Ecto.Query.PlannerTest do
       end
     end
   end
+
+  describe "filter tests" do
+    test "filter basic filter compiles" do
+      from(c in Comment,
+        group_by: c.post_id,
+        select: %{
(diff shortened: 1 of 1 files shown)


test/ecto/query/planner_test.exs
```diff
@@ -1505,4 +1505,26 @@ defmodule Ecto.Query.PlannerTest do
...
+      from(c in Comment,
+        group_by: c.post_id,
+        select: %{
+          not_aaaa_comments_count: count(c.text) |> filter(c.text != "aaaa")
+        }
+      )
+      |> normalize()
+    end
+
+    test "filter from fragment compiles" do
```
josevalim, line 1520:
```suggestion
  describe "filter" do
    test "with aggregate" do
      from(c in Comment,
        group_by: c.post_id,
        select: %{
          not_aaaa_comments_count: count(c.text) |> filter(c.text != "aaaa")
        }
      )
      |> normalize()
    end

    test "with fragment" do
```

PR #3906, opened 2022-05-20
Support arbitrary scalar subqueries as query expressions

ericfreese:
[PostgreSQL allows scalar subqueries (subqueries that return one row and one column) anywhere in a value expression][1].

Add support for subqueries outside of other expressions like `in/2`, `any/1`, etc. I see this as a natural extension of [this PR that adds support for scalar subqueries in `Repo.insert_all`][2].

This appears to work in postgres with no modifications to `ecto_sql`, but I'm not sure how/if other database backends would be affected.

Here is an example query that previously errored but now works:

```ex
avg_amount =
  from(
    m in "measurements",
    select: avg(m.amount)
  )

from(
  m in "measurements",
  where: m.amount < subquery(avg_amount),
  select: m.amount
)
|> Repo.all()
```

This code previously failed to compile with the following error:

```
== Compilation error in file lib/ecto_app.ex ==
** (Ecto.Query.CompileError) `subquery(avg_amount)` is not a valid query expression.

  for Ecto.Query.API to see the supported database expressions

* If you intended to extend Ecto's query DSL, make sure that you have required
  the module or imported the relevant function. Note that you need macros to
  extend Ecto's querying capabilities
```

[1]: https://www.postgresql.org/docs/14/sql-expressions.html#SQL-SYNTAX-SCALAR-SUBQUERIES
[2]: https://github.com/elixir-ecto/ecto_sql/pull/39

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -169,6 +169,14 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:fragment, [], merge_fragments(pieces, frags)]}, params_acc}
   end
 
+  # subqueries
+  def escape({:subquery, _, [expr]}, _, {params, subqueries}, _vars, _env) do
+    subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
+    index = length(subqueries)
+    expr = {:subquery, index} # used both in ast and in parameters, as a placeholder.
+    {expr, {[expr | params], [subquery | subqueries]}}
+  end
+
   # interval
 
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you! Looks great! We just need some tests. We can see the places where we added support for subqueries in `any`/`in` and add a similar test for `<` or similar.

ericfreese:
Ok @josevalim I've added a couple tests. What do you think?

josevalim:

PR #3908, opened 2022-05-24
Fragment escape

kipcole9:
Adds support for `escape!/1` as a fragment expression.  This expression will:

1. Quote the query string using an adapter specific quote_name/1 function (this requires the companion pull request to echo_sql
2. Interpolate the result of the expression directly into the query string
2. Raise is the interpolated value includes the quoting character

This PR evolved from the [conversation](https://elixirforum.com/t/proposal-unsafe-interpolation-in-ecto/47922) on Elixir Forum.

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -12,7 +12,7 @@ defmodule Ecto.Query.API do
     * Aggregates: `count/0`, `count/1`, `avg/1`, `sum/1`, `min/1`, `max/1`
     * Date/time intervals: `datetime_add/3`, `date_add/3`, `from_now/2`, `ago/2`
     * Inside select: `struct/2`, `map/2`, `merge/2` and literals (map, tuples, lists, etc)
-    * General: `fragment/1`, `field/2`, `type/2`, `as/1`, `parent_as/1`
+    * General: `fragment/1`, `field/2`, `type/2`, escape!/1, `as/1`, `parent_as/1`
 
   Note the functions in this module exist for documentation
   purposes and one should never need to invoke them directly.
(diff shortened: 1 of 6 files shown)


lib/ecto/query/api.ex
```diff
@@ -428,6 +428,32 @@ defmodule Ecto.Query.API do
   The only downside is that it will show up as a fragment when
   inspecting the Elixir query.  Other than that, it should be
   equivalent to a built-in Ecto query function.
+
+  ## Mutatinng the query string with escape!/1
```
LostKobrakai, line 432:
```suggestion
  ## Mutating the query string with escape!/1
```

kipcole9: I wanted to convey a strong term that gives pause to anyone thinking of using escape!/1 because of the implications.  Its definitely not completely accurate as you say.  I'd like to hear some additional views if there are any about the best term to use here.

josevalim:
Thank you @kipcole9!

Feature-wise this looks good, but implementation wise I don't think we need to add `quote_name`. Instead, send the fragment as is to the adapter and add a new clause here for converting `escape!`:

https://github.com/elixir-ecto/ecto_sql/blob/520494878635949924a6ae585f7a4d0b84d6a221/lib/ecto/adapters/postgres/connection.ex#L676

This means the changes should be minimal:

1. You should not need to change the planner, as long as you change `Builder.escape`, `escape!` will find its way to the adapter query
2. For each adapter, you need a new `expr` clause matching on `{:escape!, _, [arg]}`

Please let me know if you have any questions!

kipcole9:
José, thanks for the guidance. The reason I went with updating the planner is that the planner has better context of both the expressions and the parameters.

I have made an implementation following your recommendation which does indeed dramatically simplify the expression expansion. However .... I can't find a way to interpolate the parameters and remove the interpolated parameters from the parameter list.

This leads me to believe that the work has to be done after `Ecto.Query.Builder` but before `prepare/2` (which is called in the planner).  

### ✅ In Ecto.Query.Builder
```elixir
  def escape({:escape!, _, [expr]}, type, params_acc, vars, env) do
    {escaped, params_acc} = escape(expr, type, params_acc, vars, env)
    {{:{}, [], [:escape!, [], [escaped]]}, params_acc}
  end
```

### ✅ In Ecto.Adapters.Postgres.Connection
Works fine, but is too late in the lifecycle of the query. There is no `params` context available.
```elixir
    defp expr({:escape!, [], [arg]}, _sources, _query) when is_binary(arg) do
      quote_name(arg)
    end

    defp expr({:escape!, [], [{:^, [], [index]}]}, _sources, _query) do
      [:escape!, index]
    end
```

### ❌ How to resolve the params?
In  `Ecto.Adapters.SQL` we have
```elixir
      def prepare(:all, query) do
        {:cache, {System.unique_integer([:positive]), IO.iodata_to_binary(@conn.all(query))}}
      end
```
Where `all/1` doesn't have the parameters, and neither does `prepare/2`. And anyway, `all/1` received a binary which is totally reasonable but too late for parameter interpolation..  This is what lead me to implement in the planner. 

Doing this in `execute/5` is also too late: the query cache key is incorrect, the query is a string so parameter interpolation would be string interpolation which isn't very safe.

Anything you can see that I am missing?

josevalim:
@kipcole9 the tricky is the builder. The builder knows that the argument to `escape!` will never be a parameter, so the `Builder.escape` function should enforce that its argument is either a literal string or a `{:^, _, [ast]}`, which won't become a parameter but rather an AST argument. We do something very similar for the `field` macro in `Builder.escape`.

kipcole9:
Closing in favour of alternative implementation (which is not done yet and my head hurts a lot).

josevalim:
Pushed it as part of 6e7bc3f4d757b7c09ced10135e0c5c4ce1f4ea2f. Just the ecto_sql part missing!

josevalim:
Support for ecto_sql was also added. Please give master/main a try and if it all works I will ship a new version!

PR #3915, opened 2022-05-31
Extending the Doc about executing the Multi.put(...)

AlexR2D2:
Sometimes when we use Multi.put operation along with others Multi operations like insert, update, delete (etc) what will be executed in transaction we may automatically decide what Multi.put will be executed in transaction too. So, may be to add additional note to the Doc to catch additional Dev attention what Multi.put will be execute outside of transaction?

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -615,14 +615,26 @@ defmodule Ecto.Multi do
   @doc """
   Adds a value to the changes so far under the given name.
 
-  The value put is out of the transaction.
+  If you would like to run arbitrary functions as part of your transaction, see `run/3` or `run/5`.
 
   ## Example
 
+  Adding a pre-received company as value to the changes
+
+      company = Repo.insert(Company)
+
       Ecto.Multi.new()
-      |> Ecto.Multi.put(:params, params)
-      |> Ecto.Multi.insert(:user, fn changes -> User.changeset(changes.params) end)
-      |> Ecto.Multi.insert(:person, fn changes -> Person.changeset(changes.user, changes.params) end)
+      |> Ecto.Multi.put(:company, company)
+      |> Ecto.Multi.insert(:user, fn changes -> User.changeset(changes.company) end)
+      |> Ecto.Multi.insert(:person, fn changes -> Person.changeset(changes.user, changes.company) end)
+      |> MyApp.Repo.transaction()
+
+  the same but in more shorter form
+
+      Ecto.Multi.new()
+      |> Ecto.Multi.put(:company, Repo.insert(Company))
+      |> Ecto.Multi.insert(:user, fn changes -> User.changeset(changes.company) end)
+      |> Ecto.Multi.insert(:person, fn changes -> Person.changeset(changes.user, changes.company) end)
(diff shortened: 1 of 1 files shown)


v0idpwn:
Hi, thanks for the PR!

`Multi.put/3` doesn't allow for arbitrary execution (so it doesn't **run outside (or inside) of the transaction**) . This is the goal of `Multi.run/3`.

Doing:
```elixir
Mullti.new()
|> Multi.insert(:a, ...)
|> Multi.put(:b, Repo.insert(...))
|> Repo.transaction()
```
Is not much different from:
```elixir
b = Repo.insert(...)

Mullti.new()
|> Multi.insert(:a, ...)
|> Multi.put(:b, b)
|> Repo.transaction()
```

Maybe a better approach would be rephrasing to suggest the usage of `Multi.run/3` when one wants to execute something inside the transaction?

AlexR2D2:
Hi, thanks for help! I updated the PR. You are welcome to rephrase this directly any way you want since you are native speaker but i am not :)

lib/ecto/multi.ex
```diff
@@ -615,12 +615,26 @@ defmodule Ecto.Multi do
   @doc """
   Adds a value to the changes so far under the given name.
 
+  If you would like to run arbitrary functions as part of your transaction, see `run/3` or `run/5`.
```
josevalim, line 618:
```suggestion
  The given `value` is added to the multi before the transaction starts.
  If you would like to run arbitrary functions as part of your transaction,
  see `run/3` or `run/5`.
```

lib/ecto/multi.ex
```diff
@@ -615,12 +615,26 @@ defmodule Ecto.Multi do
...
+      |> Ecto.Multi.insert(:person, fn changes -> Person.changeset(changes.user, changes.company) end)
+      |> MyApp.Repo.transaction()
+
+  the same but in more shorter form
+
       Ecto.Multi.new()
-      |> Ecto.Multi.put(:params, params)
-      |> Ecto.Multi.insert(:user, fn changes -> User.changeset(changes.params) end)
-      |> Ecto.Multi.insert(:person, fn changes -> Person.changeset(changes.user, changes.params) end)
+      |> Ecto.Multi.put(:company, Repo.insert(Company))
```
josevalim, line 635:
```suggestion
      |> Ecto.Multi.put(:company, company)
```

lib/ecto/multi.ex
```diff
@@ -615,12 +615,26 @@ defmodule Ecto.Multi do
...
+
+      company = Repo.insert(Company)
+
+      Ecto.Multi.new()
+      |> Ecto.Multi.put(:company, company)
+      |> Ecto.Multi.insert(:user, fn changes -> User.changeset(changes.company) end)
+      |> Ecto.Multi.insert(:person, fn changes -> Person.changeset(changes.user, changes.company) end)
+      |> MyApp.Repo.transaction()
+
+  the same but in more shorter form
```
josevalim, line 632:
```suggestion
  Imagine there is an existing company schema that you retrieved from
  the database. You can insert it as a change in the multi using `put/3`:
```

josevalim:

PR #3917, opened 2022-06-01
add one/2 and all/2 functions to Ecto.Multi

lohayon:
Hi,
This PR aims to add query operations to the multi. It is useful to lock an entity inside the transaction.

PS: Thank you for your wonderful work. (ecto and all elixir ecosystem)

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -398,10 +398,12 @@ defmodule Ecto.Multi do
   end
 
   @doc """
-  Adds a one operation to the multi.
-  It is usefull to lock an entity inside the transaction.
-  Accepts the same arguments and options as `c:Ecto.Repo.one/2` does.
+  Runs a query expecting one result and stores it in the multi.
+
+  Accepts the same arguments and options as `c:Ecto.Repo.one/2`.
+
   ## Example
+
       Ecto.Multi.new()
       |> Ecto.Multi.one(:post, Post)
       |> MyApp.Repo.transaction()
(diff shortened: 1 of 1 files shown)


lib/ecto/multi.ex
```diff
@@ -397,6 +397,79 @@ defmodule Ecto.Multi do
     run(multi, name, operation_fun({:delete, fun}, opts))
   end
 
+  @doc """
+  Adds a one operation to the multi.
+  It is usefull to lock an entity inside the transaction.
+  Accepts the same arguments and options as `c:Ecto.Repo.one/2` does.
+  ## Example
```
josevalim, line 404:
```suggestion
  Runs a query expecting one result in the multi.

  Accepts the same arguments and options as `c:Ecto.Repo.one/2`.

  ## Example

```

lib/ecto/multi.ex
```diff
@@ -397,6 +397,79 @@ defmodule Ecto.Multi do
...
+  def one!(multi, name, queryable, opts) do
+    run(multi, name, operation_fun({:one!, fn _ -> queryable end}, opts))
+  end
+
+  @doc """
+  Adds an all operation to the multi.
+  It is usefull to lock entities inside the transaction.
+  Accepts the same arguments and options as `c:Ecto.Repo.all/2` does.
+  ## Example
+      Ecto.Multi.new()
```
josevalim, line 453:
```suggestion
  Runs a query in the multi and stores all entries.

  Accepts the same arguments and options as `c:Ecto.Repo.all/2` does.

  ## Example

      Ecto.Multi.new()
```

josevalim:
I pushed some small fixes to the docs and I will merge it once CI passes. Thank you!

josevalim:

lohayon:
Hi, I saw `one!` has been removed.
So the function at line https://github.com/elixir-ecto/ecto/blob/v3.8.4/lib/ecto/multi.ex#L860, has no more use.

josevalim:
Great catch, thank you, fixed too!

PR #3918, opened 2022-06-01
add on_valid_changeset/2 to changeset

lohayon:
Sometimes we want to do an action on a changes only if it is valid.
This function comes to reduce loc like this:
```elixir
 defp foo(%Ecto.Changeset{valid?: false} = changeset),
    do: changeset

  defp foo(%Ecto.Changeset{valid?: true} = changeset), do:
   changeset |> put_change(:subject, "hello")

```


josevalim:
Thank you but I am personally not a big fan of such APIs, as I prefer to relying on pattern matching or even conditionals if both branches are available. Of course you can always have similar helpers in your app though!

lohayon:
I understand your position.
Thanks for your attention.

PR #3920, opened 2022-06-01
Fix description of  again

spencerdcarlson:
Ooops

Here is some real data:
**idle_time**: 1890963
**queue_time**: 28
**query_time**: 37635616
**decode_time**: 1016
**total_time**: 37636661

So (28 + 37635616 + 1016) = 37636660 which looks like it is a rounding error off from the reported `total_time` of 37636661


josevalim:

Issue #3921, opened 2022-06-03
datetime_utc virtual fields returned as NaiveDateTime instead of DateTime


aschiavon91:
### Elixir version

1.13.4

### Database and Version

PostgreSQL 13.4

### Ecto Versions

3.8.2

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.16.3

### Current behavior

I'm getting into this bug that when I'm trying to do a select_merge, inside a subquery, and this select_merge, selects a datetime_utc_usec from an association, the value is returned as NaiveDateTime instead of DateTime.

POC: https://github.com/aschiavon91/ecto_bug_poc

### Expected behavior

return a DateTime instead of NaiveDateTime.

greg-rychlewski:
Fixed in https://github.com/elixir-ecto/ecto/pull/3922.

PR #3922, opened 2022-06-04
keep type for virtual fields in subquery

greg-rychlewski:
This is addressing https://github.com/elixir-ecto/ecto/issues/3921.

The issue is that the virtual fields are not in the type list for the source/schema so their type is set to `:any`. Since the Postgres adapter is not using `timestamptz`, these fields are returned as NaiveDateTime.

I think there's another solution that would work by taking the virtual fields into account here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1583. But that would be a bigger change and would have to alter how `schema.__schema__(:dump)` is created. That's why I went this route first, but please let me know what you think!

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -398,7 +398,7 @@ defmodule Ecto.Query.Planner do
         if valid_subquery_value?(value) do
           {key, value}
         else
-          error!(query, "atoms, maps, lists, tuples and sources are not allowed as map values in subquery, got: `#{Macro.to_string(expr)}`")
+          error!(query, "atoms, structs, maps, lists, tuples and sources are not allowed as map values in subquery, got: `#{Macro.to_string(expr)}`")
         end
     end)
   end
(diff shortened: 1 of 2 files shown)


v0idpwn (dismissed):

lib/ecto/query/planner.ex
```diff
@@ -307,7 +307,12 @@ defmodule Ecto.Query.Planner do
   defp subquery_source(nil, types), do: {:map, types}
   defp subquery_source(name, types) when is_atom(name), do: {:struct, name, types}
   defp subquery_source({:source, schema, prefix, types}, only) do
-    types = Enum.map(only, fn {field, _} -> {field, Keyword.get(types, field, :any)} end)
+    types =
+      Enum.map(only, fn
+        {field, {:value, type}} -> {field, Keyword.get(types, field, type)}
+        {field, _} -> {field, Keyword.get(types, field, :any)}
```
v0idpwn, line 313:
```suggestion
        {field, {:value, type}} -> {field, Keyword.get(types, field, type)}
```
All tests seem pass without this clause. If it is really required maybe it would be good to have a test that needs it

greg-rychlewski: Good call! I thought maybe literals wouldn't match this pattern but they do. It seems like the `collect_fields` function is taking care of this.

greg-rychlewski: Oh I guess I found one case, when you select a struct:

```elixir
query = from p in Post, join: c in assoc(p, :comments), select_merge: %{text: %Post{}}
```

Now I'm a bit unsure if structs should be disallowed, the same as maps, or if I should add the pattern back.

v0idpwn: +1 for disallowing structs

greg-rychlewski: Done :).  I believe this is a safe way to match the AST for structs, but never did it before so not 100% certain.

v0idpwn (approved):

PR #3926, opened 2022-06-18
Make docs for schemaless inserts more specific

tfwright:
Since this doc is devoted specifically to performing inserts without a schema, it seems more pertinent for the example code to use a source rather than a schema.


josevalim:

PR #3927, opened 2022-06-19
Select: Support dynamic map values

arnodirlam:
This PR extends the `select` and `select_merge` functionality with interpolation of maps with dynamic values.

```elixir
metric_ref = dynamic([p], p.view_count)

from(p in "posts",
  select: ^%{
    metric: metric_ref
  }
)
```

### Previous work

- #3928 
- #3931 
- #3935 
- #3952 

### Outdated parts

The original PR also contained a proposal to allow nested interpolations with dynamics. This was redacted in favor of the now implemented proposal (see above) that allows for almost the same use cases by combining `select` and `select_merge`, but with drastically lower complexity of implementation

```elixir
metric_ref = dynamic([p], p.view_count)

from(p in "posts",
  select: %{
    title: p.title
    metric: ^metric_ref
  }
)
```

As discussed in the [original thread](https://groups.google.com/g/elixir-ecto/c/4mdQAo_jTCQ) leading to this PR, this would have been the first time Ecto would not enforce interpolation at the root level, like it currently is across the board.

This is currently only implemented for map literals (`%{}`). I can add support for nested interpolations in lists (for :struct and :map) but they could only return field names, if I understand this correctly.

I've tested it extensively with another project using the Postgres adapter in addition to the tests added.

What I wasn't sure of is at what point exactly to traverse the select expression and expand any dynamic interpolations. Right now, it happens as the first thing at runtime, on initializing the `Ecto.Query.SelectExpr` struct. This is a change of behaviour though, because interpolated select was handled as a param before (see changed `inspect_test.exs` for an example).

Appreciate any feedback and hope to learn even more! 😄

### Special thanks

My open-source work is sponsored and kindly supported by [Team Engine](https://www.teamengine.co.uk/) as part of evolving the [Dx library](https://github.com/dx-beam/dx).


josevalim:
Hi @arnodirlam! Can you please break this PR in two? One for dynamic and another for subquery? The dynamic has implications at runtime if we try to expand it, so it needs to be carefully assessed. Thank you! :)

arnodirlam:
Thanks, @josevalim, that made a lot of sense!

Extracted the subquery part into #3928.

lib/ecto/query/builder/dynamic.ex
```diff
@@ -40,14 +40,14 @@ defmodule Ecto.Query.Builder.Dynamic do
   list is not reversed. This is useful when the dynamic expression
   is given in the middle of an expression.
   """
-  def partially_expand(kind, query, %{binding: binding} = dynamic, params, count) do
-    {expr, {_binding, params, subqueries, count}} = expand(query, dynamic, {binding, params, [], count})
+  def partially_expand(kind, query, %{binding: binding} = dynamic, params, subqueries, count) do
+    {expr, {_binding, params, subqueries, count}} = expand(query, dynamic, {binding, params, subqueries, count})
 
-    if subqueries != [] do
+    if subqueries != [] and kind != :select do
```
josevalim, line 46: Maybe we should two partially_expand. One that accepts subqueries (and therefore returns subqueries) and another one without. I think I prefer this one.

Thoughts?

arnodirlam: That was my initial implementation. I was on the fence, but decided to expose the `subqueries` to all callers, in case we want to support subqueries for them in the future.

Looking at it again, I think they are actually structurally different - ordering, grouping or updating by a subquery doesn't really seem to make sense. So having two `partially_expand` seems to make more sense.

I've implemented the change in [583795f](https://github.com/elixir-ecto/ecto/pull/3927/commits/583795ff2c3b46847d41a225d2ebb25f40bef427). Not sure about the args list - I removed `kind` as it's not needed anymore, so it matches only on the position of the `%{binding: binding}`, which is not ideal but works. Open for suggestions!

arnodirlam:
> The dynamic has implications at runtime if we try to expand it, so it needs to be carefully assessed.

How about flagging whether a `SelectExpr` has any interpolations at compile-time, so the `expand_dynamic` at runtime can be skipped if there are none? Would that make sense?

josevalim:
That's typically what it does, although no tagging necessary. We simply emit different code if it has interpolation or not.

josevalim:
Hi @arnodirlam, awesome work as usual!

Unfortunately this still makes it so every select with interpolation need to traverse its arguments at runtime, even if most of them are not using this feature (most likely). It is not an issue with your implementation (which is flawless) but rather with the approach.

This is the reason why we only allow interpolating dynamic at the root, for all constructs, except for dynamic itself (i.e. you can nested dynamic in dynamics, but that's it).

I think the path forward is actually to allow something like `select: dynamic([p], %{... => ...})`, but I don't know how complex this would be in practice as `select` has very peculiar traversal rules.

Another option is to support `dynamic(^maybe_dynamic)` in queries, and this would allow us to interpolate them anywhere, but this may defeat the purpose of dynamic in the first place.

arnodirlam:
Thank you for your kind words, @josevalim! 🙏 

> Another option is to support `dynamic(^maybe_dynamic)` in queries, and this would allow us to interpolate them anywhere, but this may defeat the purpose of dynamic in the first place.

I actually find this pretty intuitive, because it's in line with using subqueries. For that reason, I'd also suggest omitting the `^` for `dynamic` in queries:

```elixir
# subquery
subquery = from(c in "comments", where: c.post_id == parent_as(:post).id, select: count())
from(p in "posts", as: :post, where: subquery(subquery) >= 5)

# dynamic
metric_field = dynamic([p], p.view_count)
from(p in "posts", select: %{metric: dynamic(metric_field)})
```

Why would this defeat the purpose of dynamic?

josevalim:
Alright, let's give this approach a try then!

arnodirlam:
I've made an attempt. Is this going in the right direction?

Ideally, I'd like to expand each dynamic where it is, instead of traversing the entire select expression, but in order to expand them, I need access to the query. Is there some way to achieve it?

lib/ecto/query/builder.ex
```diff
@@ -188,6 +188,12 @@ defmodule Ecto.Query.Builder do
     {expr, {[expr | params], acc}}
   end
 
+  # dynamic
+  def escape({:dynamic, _, [expr]}, _, {params, acc}, _vars, _env) do
+    acc = %{acc | expand_dynamic?: true}
```
josevalim, line 193: Should we check if `acc` has a `expand_dynamic?` flag and, if not, we raise saying the `type` does not support `dynamic`?

josevalim:
Yes, this is exactly it. :)

But now I am wondering what do we do for the other cases... will we also support a dynamic deeply nested anywhere in a select? For example, should this work:

    select: %{foo: p.bar and dynamic(...)}

What about this?

    select: {p.foo, %{key: dynamic(...)}}

Also, will we support the same for select, where, etc?

If we want to say yes, then supporting this in an ad-hoc fashion across all of Ecto is complicated. Perhaps we should move the expansion of dynamics to inside the Planner... but I don't think that's any easier... :(

The other option is to limit the approach in this PR and only allow `dynamic(...)` in select as map values, inside lists, and tuples at the select root... but then we start asking why `select` requires `dynamic` when none of `order_by`, `where`, etc and friends need it.

So I think this is right but I am not quite sure if we should really go down this path. :)

arnodirlam:
Thanks for your feedback, and for making your thought process transparent. This is incredibly valuable, and I do see the intricacies. I'll also think about it for a bit first.

arnodirlam:
I've thought about it, and I think that requiring `dynamic` inside the query everywhere a dynamic is (or could be) used would be the ideal outcome. That being said, I'd like to understand the concerns you mentioned better.

> will we also support a dynamic deeply nested anywhere in a select? [...] If we want to say yes, then supporting this in an ad-hoc fashion across all of Ecto is complicated.

As of now, `dynamic` inside the query will already set the `expand_dynamic?` flag during the `escape` phase (at compile time). From my (limited) understanding, the next step would be to make sure that after escaping, for each query part (probably most modules under `Ecto.Query.Builder`) for which `expand_dynamic?` is `true`, the macro expression tree is traversed recursively and all `Dynamic` structs are expanded. The hard part, I guess, is to catch all valid syntax expressions to do the deep recursion for each query part (what I did in [expand_nested_dynamic/2](https://github.com/elixir-ecto/ecto/blob/aa2f38eabbb270f9c400ddbb0cf310c7fc02ccb9/lib/ecto/query/builder/select.ex#L213-L255) for `select`).

Do you think this is feasible at all? What else am I missing?

Another idea I had is to deal with `dynamics` like we do with `subqueries`: extract them during the `escape` phase, only leaving (numbered) references, and then expand and put them back in during in the Planner. Not sure how much we'd gain from that though 🤔

josevalim:
Your assessment is correct. Doing all of those things is exactly what is complex. :) It is certainly doable, but it is complexity and extra code to add to an already complex codebase. :(

arnodirlam:
Alright, let's go! 🙌🚀

josevalim:
Sorry, to be clear, I am not convinced we should add this feature yet. I am more inclined towards a no than a yes. It is fine if you still explore it, I just don't want to mislead you.

arnodirlam:
Okay, thanks for the clarification. To be honest, I did understand it like this, but have some ideas that I want to explore. I should have made that clear. I think it _could_ actually lead to less complexity and better performance, in comparison to not adding it. Will describe in detail tomorrow.

arnodirlam:
I've taken a step back and re-read all dynamic-related docs, and your comments here, @josevalim.

First, my idea I mentioned above:

Assuming that all dynamic parts are inserted using `dynamic()` inside the query, and can appear deeply nested inside a query, my idea was to "bubble up" the `dynamic()` and internally (during escape, at compile time) add `dynamic()` also to all parent nodes ("ancestors") in the tree. The recursion for expanding dynamic references could then, at each level, only look at the nested nodes that are marked as `dynamic()`, and skip traversing the others.

The reduction of complexity I meant would be through unifying the code for expanding dynamics across the different `Builder` modules, so that code like [this](https://github.com/elixir-ecto/ecto/blob/25319b4db27306bed967732d8ce92f95309e5f8e/lib/ecto/query/builder/filter.ex#L104-L113) would not be needed any more. But I'm not sure about this part.

However, I do realize that

- complexity not only means lines of code (if that can even be reduced) but also the mental overhead of abstractions and layers of processing, which would definitely be increased once we start allowing nested dynamics.
- introducing `dynamic()` inside queries in a consistent way would be a considerable change in the API, while the use case of this feature is relatively narrow. I guess, most people either use relatively static queries, or fully dynamic ones (e.g. libraries generating queries).

So I understand your hesitancy much better now, and I agree. It'd be like making 3 steps at once, while one carefully considered step would probably be enough.

### Suggestion

How about only allowing dynamics as map values inside a map that's interpolated at root level:

```elixir
dynamic = dynamic(...)
from(p in "posts", select: ^%{a: dynamic}, select_merge: ^%{b: dynamic})
```

Much like `order_by`:

```elixir
dynamic = dynamic(...)
order_by("posts", ^[asc: dynamic, desc: :bar])
```

In both cases you have a required structure (map/keyword list) and have to interpolate at root level, if at least one value is dynamic.

With this approach,

- expanding dynamics would only take place with one specific syntax, a map interpolated at root level
- this syntax is currently not valid, so there can be no performance degradations in existing code
- the internal complexity is limited to `select`
- the API is consistent with existing syntax patterns (`order_by`); no new syntax patterns are introduced
- it's pretty powerful, because dynamic map values could also be merged into structs (I assume)
- we could even allow nested maps with dynamic values, as long as there's a `^` at the root level (I assume)

Would this make more sense?

josevalim:
This is a great wrap up @arnodirlam, thank you so much, it is very appreciated!

> How about only allowing dynamics as map values inside a map that's interpolated at root level:

I assume today it is not possible to interpolate a map or a tuple in the select root, so we could add such a feature. Can you please check what are the valid values for interpolation at the root of a select? I assume an atom or list of atoms? So maybe we should support `%{key: field_as_atom | dynamic}`, exactly as order_by?

arnodirlam:
Today, root level interpolation only supports a list of atoms:

```elixir
fields = [:foo, :bar, :baz]
select("q", ^fields)
```

So anything that's not a list of atoms _could_ support dynamic, if interpolated at root level. For example, tuples or keyword lists.

Other interpolations possible today (not root level):

```elixir
fields = [:foo, :bar, :baz]
select("q", [q], map(q, ^fields))
select("q", [q], struct(q, ^fields))
select("q", [q], %{body: ^"body"})
select("q", [^:foo, ^:bar, ^:baz])

binding = :post
field = :author
select_merge("q", [{^binding, bound}], %{^field => field(bound, ^field)})
```

> So maybe we should support `%{key: field_as_atom | dynamic}`, exactly as order_by?

I'm not sure whether the similarity goes that far, because here it would effectively be a mapping of field names, whereas in `order_by`, `:asc` and `:desc` are part of the syntax rather than identifiers 🤔 

Also, atoms currently are valid values in maps. Would these change behaviour as well?

```elixir
select("q", %{foo: :bar})
```

Lastly, if we also supported nested maps, what would you expect the following to return?

```elixir
from(p in Post,
  join: c in assoc(p, :comments),
  select: ^%{title: :title, comments: %{inserted_at: :inserted_at}}
)
```

josevalim:
Perfect, so I think we can support map interpolation for select and lookup for dynamics as part of the values and all other values are kept as is?

arnodirlam:
The implementation is ready for review now 👍 Happy to make any changes you'd prefer, also small ones. Will update the docs next.

arnodirlam:
Just realized the expanding part can be simplified even further. Will do that in a bit.

arnodirlam:
Should we support or error on subqueries nested in interpolated maps?

```elixir
subquery = from(...)
query = from(..., select: ^%{metric: subquery(subquery)})
```

In [0d6152c](https://github.com/elixir-ecto/ecto/pull/3927/commits/0d6152c8b4bb142145bb7cbd2ffada3cff2899a0) I made it so that it's supported, for now.

arnodirlam:
The same question extends to fixed values. In a `where`, for example, we replace fixed values in interpolated keyword lists with param references.

```elixir
      %{wheres: [where]} = where(from(p in "posts"), [p], ^[foo: 1, bar: "baz"])
      assert Macro.to_string(where.expr) ==
             "&0.foo() == ^0 and &0.bar() == ^1"
      assert where.params ==
             [{1, {0, :foo}}, {"baz", {0, :bar}}]
```

I'm working on a commit to do the same for interpolated maps in `select`, so we can see how it would look like.

lib/ecto/query.ex
```diff
@@ -529,6 +529,34 @@ defmodule Ecto.Query do
...
+  But this will:
+
+      from query, select: ^%{field: dynamic(...)}
+
+  Maps with dynamics can also be merged into existing `select` structures,
+  enabling a variety of possibilities for partially dynamic selects:
+
+      metric = dynamic([p] p.distance)
+
+      from query, select: {:map, [:id, :period]}, select_merge: ^metric
```
josevalim, line 558:
```suggestion
      from query, select: [:id, :period], select_merge: ^metric
```

?

josevalim:
It looks great to me. My only question is if we want to raise or do something special about lists. Imagine this:

    select: ^%{post: [:id, :title, :body], other: dynamic}

I think we should either raise if we find a list there or support it. I think raising is better. WDYT?

lib/ecto/query.ex
arnodirlam: Updated in [cb6a223](https://github.com/elixir-ecto/ecto/pull/3927/commits/cb6a223d0e6734d5d192c21465dd1acea5e83a8d).

josevalim:

PR #3928, opened 2022-06-19
Select: Support subquery values when using plain map

arnodirlam:
This PR extends the `select` functionality to support subqueries.

```elixir
subquery =
  from(c in "comments",
    where: c.post_id == parent_as(:post).id,
    select: count()
  )

from(p in "posts,
  as: :post,
  select: %{
    comment_count: subquery(subquery)
  }
)
```

This is extracted from the larger PR #3927.

All feedback welcome! 🙏

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -168,6 +168,10 @@ defmodule Ecto.Query.Builder do
   end
 
   # subqueries
+  def escape({:subquery, _, _} = expr, type, {params, {_take, subqueries}}, vars, env) do
+    escape(expr, type, {params, subqueries}, vars, env)
+  end
+
   def escape({:subquery, _, [expr]}, _, {params, subqueries}, _vars, _env) do
     subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
     index = length(subqueries)
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -168,15 +168,13 @@ defmodule Ecto.Query.Builder do
   end
 
   # subqueries
-  def escape({:subquery, _, _} = expr, type, {params, {_take, subqueries}}, vars, env) do
-    escape(expr, type, {params, subqueries}, vars, env)
+  def escape({:subquery, _, _} = expr, _, {params, {take, subqueries}}, _vars, _env) do
+    {expr, {params, subqueries}} = escape_subquery(expr, {params, subqueries})
+    {expr, {params, {take, subqueries}}}
   end
 
-  def escape({:subquery, _, [expr]}, _, {params, subqueries}, _vars, _env) do
-    subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
-    index = length(subqueries)
-    expr = {:subquery, index} # used both in ast and in parameters, as a placeholder.
-    {expr, {[expr | params], [subquery | subqueries]}}
+  def escape({:subquery, _, _} = expr, _, params_acc, _vars, _env) do
+    escape_subquery(expr, params_acc)
   end
 
   # interval
@@ -534,6 +532,14 @@ defmodule Ecto.Query.Builder do
     frag
     |> split_fragment("")
     |> merge_fragments(args)
+  end
+
+  defp escape_subquery(expr, {params, subqueries}) do
+    subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
+    index = length(subqueries)
+    # used both in ast and in parameters, as a placeholder.
+    expr = {:subquery, index}
+    {expr, {[expr | params], [subquery | subqueries]}}
   end
 
   defp escape_window_description([], params_acc, _vars, _env),
--- a/lib/ecto/query/builder/dynamic.ex
+++ b/lib/ecto/query/builder/dynamic.ex
@@ -40,14 +40,14 @@ defmodule Ecto.Query.Builder.Dynamic do
   list is not reversed. This is useful when the dynamic expression
   is given in the middle of an expression.
   """
-  def partially_expand(kind, query, %{binding: binding} = dynamic, params, count) do
-    {expr, {_binding, params, subqueries, count}} = expand(query, dynamic, {binding, params, [], count})
+  def partially_expand(kind, query, %{binding: binding} = dynamic, params, subqueries, count) do
+    {expr, {_binding, params, subqueries, count}} = expand(query, dynamic, {binding, params, subqueries, count})
 
-    if subqueries != [] do
+    if subqueries != [] and kind != :select do
       raise ArgumentError, "subqueries are not allowed in `#{kind}` expressions"
     end
 
-    {expr, params, count}
+    {expr, params, subqueries, count}
   end
 
   defp expand(query, %{fun: fun}, {binding, params, subqueries, count}) do
--- a/lib/ecto/query/builder/dynamic.ex
+++ b/lib/ecto/query/builder/dynamic.ex
@@ -40,14 +40,14 @@ defmodule Ecto.Query.Builder.Dynamic do
   list is not reversed. This is useful when the dynamic expression
   is given in the middle of an expression.
   """
-  def partially_expand(kind, query, %{binding: binding} = dynamic, params, subqueries, count) do
-    {expr, {_binding, params, subqueries, count}} = expand(query, dynamic, {binding, params, subqueries, count})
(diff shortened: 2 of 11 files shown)


lib/ecto/query/builder/dynamic.ex
```diff
@@ -40,14 +40,14 @@ defmodule Ecto.Query.Builder.Dynamic do
   list is not reversed. This is useful when the dynamic expression
   is given in the middle of an expression.
   """
-  def partially_expand(kind, query, %{binding: binding} = dynamic, params, count) do
-    {expr, {_binding, params, subqueries, count}} = expand(query, dynamic, {binding, params, [], count})
+  def partially_expand(kind, query, %{binding: binding} = dynamic, params, subqueries, count) do
+    {expr, {_binding, params, subqueries, count}} = expand(query, dynamic, {binding, params, subqueries, count})
 
-    if subqueries != [] do
+    if subqueries != [] and kind != :select do
```
josevalim, line 46: I think now, instead of raising here, perhaps we should raise on the callers to `partially_expand`?

josevalim: The reason I mention this is because it feels backwards to have this function know specifically about `select`... although moving this out will lead to more duplication.

Another option is to have two `partially_expand`. One that accepts subqueries (and therefore returns subqueries) and another one without. I think I prefer this one.

Thoughts?

arnodirlam: As I was replying, I noticed this is actually not needed for this PR yet. I reverted the changes around `partially_expand`. Sorry, my bad 😅 Would you mind copying your comment to the other PR?

josevalim: I see. But this allows subquery inside dynamic... and perhaps we should support that on both `where` and `select`. So maybe this should be another PR to be considered before #3927?

lib/ecto/query/builder/select.ex
```diff
@@ -98,20 +99,24 @@ defmodule Ecto.Query.Builder.Select do
...
     take = add_take(take, Builder.find_var!(var, vars), {tag, taken})
-    {expr, {params, take}}
+    {expr, {params, take, subqueries}}
   end
 
-  defp escape(expr, params_take, vars, env) do
-    Builder.escape(expr, :any, params_take, vars, {env, &escape_expansion/5})
+  defp escape(expr, {params, take, subqueries}, vars, env) do
+    {expr, {params, subqueries}} =
+      Builder.escape(expr, :any, {params, subqueries}, vars, {env, &escape_expansion/5})
```
josevalim, line 112: There is a bug hiding here and apparently we have no tests, but your change makes it so we ignore any "take" operation that happens inside expansion.

The correct approach should be to have `{params, take_subqueries}`, which are in the form of `{[], {%{}, []}}` instead of a triplet.

lib/ecto/query/builder/dynamic.ex
arnodirlam: Alright, will extract it into another PR tomorrow.

lib/ecto/query/builder/select.ex
arnodirlam: I was already looking at this earlier. It _seems_ that `Builder.escape` never modifies the `acc` part of `{params, acc}`, except for adding subqueries. Any pointer as to how to produce the bug?

josevalim: It never modifies it but it passes it to expand_expansion, which may modify it. You can reproduce it by having a macro `my_macro(p)` that expands to `map(p, [:some, :field])`. The fields inside the expansion won’t be collected.

arnodirlam: I managed to produce the bug, add a test, and fix it in [4119e46](https://github.com/elixir-ecto/ecto/pull/3928/commits/4119e4602a2b036f053f39dac4329c6e8cbf2676). Can't say I fully understand it quite yet, which probably shows in the way I implemented it. What could I improve?

lib/ecto/query/builder/dynamic.ex
arnodirlam: As far as I can see, subquery inside dynamic is already supported on `where`:

```elixir
subquery =
  from(t in ListTemplate,
    where: parent_as(^:list).from_template_id == t.id,
    select: %{title: t.title}
  )

where = dynamic([l], subquery(subquery) == l.title)
query = from(l in List, as: :list, where: ^where)

assert [%{title: "FANCY TEMPLATE"}] = Repo.all(query)
```

What am I missing?

josevalim:
Your changes are great and correct. The change in builder is not ideal but it is because we were lazy in the past, so definitely not your fault. I will sleep on it and likely merge tomorrow. Thank you!

lib/ecto/query/builder/dynamic.ex
josevalim: Yes, it works for where! The idea is to provide the same for select!

arnodirlam: But there currently is no dynamic for select, that's what #3927 would introduce.

Do you mean creating a separate PR for introducing dynamic for select at root level only, extracted from #3927? 🤔

josevalim:
@arnodirlam the cleanest approach would be to explicitly pass subqueries as a parameter to escape, but this would be a hundreds LOC changes and I am thinking it is probably just better to keep the "hack" and drop a comment to why it exists.

lib/ecto/query/builder/select.ex
```diff
@@ -98,16 +99,19 @@ defmodule Ecto.Query.Builder.Select do
...
-    {expr, {params, take}}
+    {expr, {params, {take, subqueries}}}
   end
 
-  defp escape(expr, params_take, vars, env) do
-    Builder.escape(expr, :any, params_take, vars, {env, &escape_expansion/5})
+  defp escape(expr, {_params, {take, _subqueries}} = params_take, vars, env) do
+    case Builder.escape(expr, :any, params_take, vars, {env, &escape_expansion/5}) do
+      {expr, {params, {take, subqueries}}} -> {expr, {params, {take, subqueries}}}
+      {expr, {params, subqueries}} -> {expr, {params, {take, subqueries}}}
```
josevalim, line 113: If this is missing the `take`, it means a bug somewhere. This clause should not exist.

josevalim: Can you please rename `params_take` in this file to `params_acc`, as it is more than just take?

lib/ecto/query/builder.ex
```diff
@@ -168,6 +168,10 @@ defmodule Ecto.Query.Builder do
   end
 
   # subqueries
+  def escape({:subquery, _, _} = expr, type, {params, {_take, subqueries}}, vars, env) do
+    escape(expr, type, {params, subqueries}, vars, env)
```
josevalim, line 172: This is the code that is making the `take` disappear. It is probably best to introduce a function called `escape_subquery` which can be used by this function and the one below, so you do something like this:

```elixir
  def escape({:subquery, _, _} = expr, type, {params, {take, subqueries}}, _vars, _env) do
    {expr, params, subqueries} = escape_subquery(expr, params, subqueries)
    {expr, {params, {take, subqueries}}}
```

josevalim (commented):
Sorry for the delay! I found one last change and we are good to go! I actually have a plan on how to clean up the code, we can discuss it after merging. :)

arnodirlam:
All good, better to find long-term solutions than to rush things 🙂 That's great, looking forward to your plan!

josevalim:

arnodirlam:
Addressed your recent comments, @josevalim. Thanks for the great guidance! 🙏

josevalim:
@arnodirlam I think the correct fix would be to change `params_acc` to, instead of a `{list, term}`, it should be a tuple of shape `{list, map}`. The map can have the optional keys `:take` or `:subqueries`. This should be a more doable change because we don't need to change the shape of `params_acc` altogether, only of the second element, and the map shape makes it easier to know when take and subqueries are supported.

Would you like to send a PR to tidy this up? It would be very welcome! You can start by changing the code modified in this PR to pass a map instead of `{take, subquery}` and, if it works, we generalize for other operations. WDYT?

arnodirlam:
Yes, happy to implement this! 👍 will have a first look later today.

I opened #3931 for the renaming of params_take that was still missing here. I can extend that PR with the refactoring, if you prefer fewer, bigger PRs, or you can merge it and I open a new one.

PR #3929, opened 2022-06-21
Support values lists

narrowtux:
To support values lists in select from, joins, update from and delete from, I added a `values(list_of_maps, schema)` function that can be used in `in` expressions like 

```elixir
query = from u in User, 
  join: v in values([%{id: 1, name: "Moritz"}, ...]), on: u.id == v.id,
  update: [set: [name: v.name]]

Repo.update_all(query, [])
```

This is to support constant tables.

I am opening this PR as a draft early to gather feedback if I should change the syntax before proceeding with the more complicated details.

There's also an ecto_sql PR with support for the postgres adapter already: https://github.com/elixir-ecto/ecto_sql/pull/416

Postgres docs: https://www.postgresql.org/docs/current/queries-values.html
MySQL docs: https://dev.mysql.com/doc/refman/8.0/en/values.html
MSSQL docs: https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver16

## TODO List

 - [x] Documentation for `values/1`
 - [ ] Implement MySQL Adapter
 - [ ] Implement MSSQL Adapter

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -670,6 +670,83 @@ defmodule Ecto.Query do
     end
   end
 
+  @doc """
+  Passes a constant table/values list to the query as a source in from or joins.
+
+  ## Usage
+
+  ### `values(list_of_maps)`
+
+  Pass a list of maps and let ecto infer the type automatically:
+
+      list = [
+        %{id: 1, name: "Peter"},
+        %{id: 2, name: "Lisa"}
+      ]
+      %Ecto.ValuesList{schema: [id: :integer, name: :string]} = values(list)
+
+  ### `values(list_of_structs)`
+
+  When passing a list of Ecto.Schema structs, `values/1` will copy the types for
+  the fields from the schema definition.
+
+      defmodule Schema do
+        use Ecto.Schema
+        @primary_key false
+        embedded_schema do
+          field :id, :integer
+          field :stat, :map
+          field :count, :integer
+        end
+      end
+      list = [
+        %Schema{id: 1, stat: %{foo: 3.6, bar: 7.3}, count: 14},
+        # ...
+      ]
+      %Ecto.ValuesList{schema: [id: :integer, stat: :map, count: :integer]} = values(list)
+
+  ### `values(type(list_of_maps, types))`
+
+  To be specific about the types without defining a schema module, `type/2` can
+  be called with a map or keyword list of fields and their ecto types:
+
+      types = %{id: :integer, name: :string, updated_at: :utc_datetime}
+      %Ecto.ValuesList{schema: [id: :integer, name: :string, updated_at: :utc_datetime]} =
+        values(type(list_of_maps, types))
+
+  ### Usage in a query
+
+  So far, we have seen what arguments can be passed to the values macro and how
+  it will use any given information to extract a schema for the values list.
+
+  To actually use it in a query, pass it as the right argument of `x in source`
+  in the from clause or in a join:
+
+      from v in values([%{x: 1}, %{x: 10}])
+
+      from u in User,
+        join: v in values([%{id: 1, os: "macOS"}, %{id: 2, os: "Windows"}]),
+        on: v.id == u.id,
+        select: {u.name, v.os}
+
+  ### Use with `update_all`
+
+  Using a values list in a join is very useful to update multiple rows with
+  different values:
+
+      list = [
+        %{id: 1, name: "Alice"},
+        %{id: 2, name: "Bob"}
+      ]
+      query = from u in User,
+        join: v in values(list), on: v.id == u.id,
+        update: [set: [name: v.name]]
+
+      Repo.update_all(query, [])
+
+  After executing the above block, the user with id 1 will have the name "Alice",
+  while the user with id 2 is now called "Bob".
+  """
   defmacro values({:type, _, [values, fields]}) do
     quote do
       # store the fields as keyword list so the order of the fields is stable
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -12,7 +12,7 @@ end
 
 defmodule Ecto.ValuesList do
   @moduledoc false
-  defstruct [:values, :schema, :params]
+  defstruct [:values, :schema, :params, :param_offset]
 end
 
 defmodule Ecto.Query do
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -12,7 +12,7 @@ end
 
 defmodule Ecto.ValuesList do
   @moduledoc false
-  defstruct [:values, :schema]
+  defstruct [:values, :schema, :params]
 end
 
 defmodule Ecto.Query do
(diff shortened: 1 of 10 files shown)


lib/ecto/query/planner.ex
```diff
@@ -913,7 +921,7 @@ defmodule Ecto.Query.Planner do
   rescue
     e ->
       # Reraise errors so we ignore the planner inner stacktrace
-      filter_and_reraise e, __STACKTRACE__
+      reraise e, __STACKTRACE__
```
josevalim, line 924:
```suggestion
      filter_and_reraise e, __STACKTRACE__
```

josevalim:
Very clean and elegant, great work. I believe we only need docs and this is good to go. However on the Ecto sql side, you may need you may have trouble handling MySQL. Also please add tests there to make sure that Fields are escaped and that  passing two Value lists do not clobber each other.

josevalim:
Oh, additional questions. Do we consider the whole values list as part of the cache key? If so, We don’t need to encode the values as parameters when building the sql and rather put them in the queries.

Alternatively we encode only the field names and the size is cache key, and then we encode parameters.

I guess another option is to encode the whole thing is JSON and then convert JSON to values. This will actually be the best approach if we want to reuse the cache And it likely requires a much smaller PR.

narrowtux:
> Do we consider the whole values list as part of the cache key?

No, that was just a quick way to get it through the planner. 

> Alternatively we encode only the field names and the size is cache key, and then we encode parameters.

I think that should work.

> I guess another option is to encode the whole thing is JSON and then convert JSON to values. This will actually be the best approach if we want to reuse the cache And it likely requires a much smaller PR. It may be worth benchmarking this.

I've thought about this too, but it feels like extra work on behalf of the user making sure all the types can be dumped to and loaded from json. But I will definitely benchmark this.

I also still need to implement escaping the values into params so the generated query from ecto_sql actually works.

narrowtux:
I could use a pointer on how to plumb the values into params with the correct types and all.

I tried doing it in `Ecto.Query.Planner.source_cache/2`, where I return a list of tuples with the actual value and its type (`[{1, :integer}, {"Bob", :string}, ...]`, but that seems to be at the wrong place since that results in postgrex complaining about having to dump values such as `{1, :integer}` where it should just be `1`.

narrowtux:
Nevermind, I think I solved it by explicitely assigning the type to each parameter in the generated SQL.

lib/ecto/query.ex
```diff
@@ -9,6 +9,12 @@ defmodule Ecto.SubQuery do
   @type t :: %__MODULE__{}
 end
 
+
+defmodule Ecto.ValuesList do
+  @moduledoc false
+  defstruct [:values, :schema]
+end
```
v0idpwn, line 16: Do we really need a schema? If the maps all have the same keys, can't we just infer the type? This would make the usage significantly easier.

v0idpwn (commented):
Love this so far, just one question :)

lib/ecto/query.ex
narrowtux: how would you handle more complex types like UUID? I don't think that can be inferred since they're just strings

narrowtux: Maybe a schema can be given optionally. I also thought about allowing schemaless key to type maps, like in `Ecto.Changeset`.

v0idpwn: I'd imagine something with `type/2` + type maps (i.e.: `type(^my_list, {:array, %{a: :string}})`, but it probably would be a bigger change. Guess taking schemas in is okay :]

narrowtux: I pushed some WIP changes where the schema is actually being used (most of it in ecto_sql, actually)

josevalim: That's pushing concerns to the adapter and we try to avoid that. The best is to normalize everything in the planner.

In any case, the need for a schema was not something I realized we needed in my first review and I am a bit more worried about the feature now but I can't quite put my finger on it yet.

narrowtux: I agree it is a bit clunky to pass a schema. I could also explore the suggestion with the optional type.

narrowtux: I had to upgrade `def values` in `Ecto.Query` to a macro, but I'm now able to write something like this:

```elixir
list_a = [%{id: 1, name: "Foo"}, %{id: 2, name: "Peter"}]
list_b = [%{id: 1, name: "Bar"}, %{id: 2, name: "Parker"}]
schema = %{id: :integer, name: :string}
from v1 in values(type(list_a, schema)), join: v2 in values(type(list_b, schema)), on: v1.id == v2.id
```

I can also omit the `type/2` call which will generate a map like `%{id: :any, name: :any}` automatically:

```elixir
list_a = [%{id: 1, name: "Foo"}, %{id: 2, name: "Peter"}]
list_b = [%{id: 1, name: "Bar"}, %{id: 2, name: "Parker"}]

from v1 in values(list_a), join: v2 in values(list_b), on: v1.id == v2.id
```

Now the only problem I still need to solve is the offset for the params (`$1`, etc).

narrowtux: I'm also open to enforce wrapping the schema map with `{:array, %{...}}` if it helps with consistency.

fuelen: Can we derive types if a list of structs is formed by embedded schema?

narrowtux:
I have pushed another set of commits that fix the necessary offset in the parameters.

While testing with the inferred type `:any` I noticed that postgrex (or postgres) always assumes that we want to give strings in all of the untyped (`:any`) params, so it's not very useful unless we actually infer some types directly in ecto by mapping the elixir primitive types to ecto types.

lib/ecto/query.ex
v0idpwn: @fuelen I could see it being done, you'd just need to assume that they all use the same schema (maybe raise otherwise). But I'm not sure how this would be regarding separation of concerns on ecto/adapters.

narrowtux: @fuelen I implemented that, also added the function `Ecto.Type.infer_type_from_value/1` which guesses the ecto type from a given value, which works with most primitive types.

lib/ecto/query.ex
```diff
@@ -664,6 +670,125 @@ defmodule Ecto.Query do
...
+      query = from u in User,
+        join: v in values(list), on: v.id == u.id,
+        update: [set: [name: v.name]]
+
+      Repo.update_all(query, [])
+
+  After executing the above block, the user with id 1 will have the name "Alice",
+  while the user with id 2 is now called "Bob".
+  """
+  defmacro values({:type, _, [values, fields]}) do
```
josevalim, line 750: Suggestion for the initial implementation: force it to be `values(values, types)` (no `type(..., ...)`) and we won't support inference for now (we might add it later).

greg-rychlewski:
Closed in favor of https://github.com/elixir-ecto/ecto/pull/4270

PR #3934, opened 2022-06-26
Drop fields overriden with :%{}

sabiwara:
Avoid selecting fields overridden with the `%{p | my_field: nil})` syntax.

The following works as expected and returns posts with `nil` as a title:

```elixir
iex> Post |> select([p], %{p | text: nil}) |> Repo.all()

23:06:01.855 [debug] QUERY OK source="blog_posts" db=8.5ms decode=0.8ms queue=2.9ms idle=1124.6ms
SELECT b0."id", b0."title", b0."text", b0."inserted_at", b0."updated_at" FROM "blog_posts" AS b0 LIMIT 2 []
```

but the underlying SQL still selects `p.text`, which could be avoided. This PR should improve this.

Comments welcome 🙂


josevalim:

PR #3935, opened 2022-06-26
Builder: Unify passing argument {params, acc} to escape functions

arnodirlam:
As suggested by @josevalim in https://github.com/elixir-ecto/ecto/pull/3928#issuecomment-1166486873:
> I think the correct fix would be to change `params_acc` to, instead of a `{list, term}`, it should be a tuple of shape `{list, map}`. The map can have the optional keys `:take` or `:subqueries`. This should be a more doable change because we don't need to change the shape of `params_acc` altogether, only of the second element, and the map shape makes it easier to know when take and subqueries are supported.
> 
> Would you like to send a PR to tidy this up? It would be very welcome! You can start by changing the code modified in this PR to pass a map instead of `{take, subquery}` and, if it works, we generalize for other operations. WDYT?

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -57,7 +57,11 @@ defmodule Ecto.Query.Builder do
 
   If the subqueries field is available, subquery escaping must take place.
   """
-  @type acc :: %{optional(:subqueries) => list(Macro.t()), optional(any) => any}
+  @type acc :: %{
+          optional(:subqueries) => list(Macro.t()),
+          optional(:take) => %{non_neg_integer => Macro.t()},
+          optional(any) => any
+        }
 
   @doc """
   Smart escapes a query expression and extracts interpolated values in
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -175,13 +175,13 @@ defmodule Ecto.Query.Builder do
   end
 
   # subqueries
-  def escape({:subquery, _, _} = expr, _, {params, acc}, _vars, _env) do
+  def escape({:subquery, _, _} = expr, _, {params, %{subqueries: subqueries} = acc}, _vars, _env) do
     subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
-    subqueries = Map.get(acc, :subqueries, [])
     index = length(subqueries)
     # used both in ast and in parameters, as a placeholder.
     expr = {:subquery, index}
-    {expr, {[expr | params], Map.put(acc, :subqueries, [subquery | subqueries])}}
+    acc = %{acc | subqueries: [subquery | subqueries]}
+    {expr, {[expr | params], acc}}
   end
 
   # interval
@@ -725,7 +725,7 @@ defmodule Ecto.Query.Builder do
     do: {find_var!(var, vars), field}
 
   def validate_type!(type, _vars, _env) do
-    error! "type/2 expects an alias, atom, initialized parameterized type or " <> 
+    error! "type/2 expects an alias, atom, initialized parameterized type or " <>
            "source.field as second argument, got: `#{Macro.to_string(type)}`"
   end
 
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -54,10 +54,10 @@ defmodule Ecto.Query.Builder do
 
   @typedoc """
   The accumulator during escape.
-  
-  If the subqueries field is available, it means subquery esscaping must take place.
+
+  If the subqueries field is available, subquery escaping must take place.
   """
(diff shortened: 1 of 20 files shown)


lib/ecto/query/builder.ex
```diff
@@ -168,13 +168,13 @@ defmodule Ecto.Query.Builder do
...
-  def escape({:subquery, _, _} = expr, _, params_acc, _vars, _env) do
-    escape_subquery(expr, params_acc)
+  def escape({:subquery, _, _} = expr, _, {params, acc}, _vars, _env) do
+    subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
+    subqueries = Map.get(acc, :subqueries, [])
+    index = length(subqueries)
+    # used both in ast and in parameters, as a placeholder.
+    expr = {:subquery, index}
+    {expr, {[expr | params], Map.put(acc, :subqueries, [subquery | subqueries])}}
   end
```
josevalim, line 178: Instead of assuming it might have subqueriess, it is better to do this only if it has subqueries:

```suggestion
  def escape({:subquery, _, _} = expr, _, {params, %{subqueries: subqueries} = acc}, _vars, _env) do
    subquery = quote(do: Ecto.Query.subquery(unquote(expr)))
    index = length(subqueries)
    # used both in ast and in parameters, as a placeholder.
    expr = {:subquery, index}
    acc = %{acc | subqueries: [subquery | subqueries]}
    {expr, {[expr | params], acc}}
  end
```

josevalim: Now, if an expression supports subqueries, you should explicitly initialize it as `subqueries: []`. :)

josevalim (commented):
Looks great! I just dropped one tiny comment about us explicitly setting `subqueries: []` when it is supported. :)

arnodirlam:
Yea, that makes sense. Very nice! 🎉

josevalim:
Can you please replace the Map.get/Map.fetch on the subqueries field by `acc.subqueries`? Then we are good to go!

lib/ecto/query/builder/select.ex
```diff
@@ -210,17 +214,17 @@ defmodule Ecto.Query.Builder.Select do
 
   def build(kind, query, binding, expr, env) do
     {query, binding} = Builder.escape_binding(query, binding, env)
-    {expr, {params, {take, subqueries}}} = escape(expr, binding, env)
+    {expr, {params, acc}} = escape(expr, binding, env)
     params = Builder.escape_params(params)
-    take   = {:%{}, [], Map.to_list(take)}
+    take = {:%{}, [], Map.to_list(Map.get(acc, :take, %{}))}
```
josevalim, line 219: Let's to the same with `take` and make sure it is always set to an empty map. Then we can do `acc.take` here and slightly simplify `add_take`. :)

josevalim (commented):
Sorry, I found another place for improvement and then we should be good for real. :)

josevalim:
Actually, I lied. There is yet another small change: find the other places we call `Builder.escape` and make sure the second element of the params tuple is a map. In many places it is the atom `:acc` or an empty list. :)

Sorry and thank you! :)

arnodirlam:
All good, no worries! Keep up the suggestions as long as they come to you 😄 Will implement the next two tomorrow or Tuesday.

lib/ecto/query/builder.ex
```diff
@@ -52,6 +52,13 @@ defmodule Ecto.Query.Builder do
   """
   @type quoted_type :: Ecto.Type.primitive | {non_neg_integer, atom | Macro.t}
 
+  @typedoc """
+  The accumulator during escape.
+  
+  If the subqueries field is available, it means subquery esscaping must take place.
```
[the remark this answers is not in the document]
arnodirlam: Thank you! Addressed in [2350657](https://github.com/elixir-ecto/ecto/pull/3935/commits/235065774d451813b0bcf89ef07c1b821bab9284).

lib/ecto/query/builder.ex
```diff
@@ -61,7 +68,7 @@ defmodule Ecto.Query.Builder do
   with `^index` in the query where index is a number indexing into the
   map.
   """
-  @spec escape(Macro.t, quoted_type | {:in, quoted_type} | {:out, quoted_type}, {list, term},
+  @spec escape(Macro.t, quoted_type | {:in, quoted_type} | {:out, quoted_type}, {list, acc},
                Keyword.t, Macro.Env.t | {Macro.Env.t, fun}) :: {Macro.t, {list, term}}
```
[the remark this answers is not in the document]
arnodirlam: Thank you! Addressed in [2350657](https://github.com/elixir-ecto/ecto/pull/3935/commits/235065774d451813b0bcf89ef07c1b821bab9284).

PR #3939, opened 2022-06-29
Add term to typespec for selects

MitchellJeppson:
When you use the `select` option in queries you can get back a `term` but some of the Repo query functions only show `Ecto.Schema.t()` or `nil`. This branch just adds term to the typespec for those functions. 

```elixir
query = from(u in User, select: u.email)
Repo.get(query)
>> "example@example.com"
```


josevalim:

PR #3940, opened 2022-06-29
fix insert_all param count when using placeholders

greg-rychlewski:
There were a couple of subtle bugs with the parameter counting in `insert_all` that coincidentally weren't caught in the existing tests because of unlucky test values. I will create comments below to highlight them.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1043,15 +1043,17 @@ defmodule Ecto.Integration.RepoTest do
       bar_ph = {:placeholder, :bar}
       foo_ph = {:placeholder, :foo}
 
+      intensity_query = from p in Post, where: p.id == ^0, select: p.intensity
+
       entries = [
-        %{intensity: 1.0, title: bar_ph, posted: ~D[2020-12-21], visits: foo_ph},
+        %{intensity: intensity_query, title: bar_ph, posted: ~D[2020-12-21], visits: foo_ph},
         %{intensity: 2.0, title: bar_ph, posted: ~D[2000-12-21], visits: foo_ph}
       ] |> Enum.map(&Map.put(&1, :uuid, Ecto.UUID.generate))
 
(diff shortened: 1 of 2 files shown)


integration_test/cases/repo.exs
```diff
@@ -1043,15 +1043,17 @@ defmodule Ecto.Integration.RepoTest do
       bar_ph = {:placeholder, :bar}
       foo_ph = {:placeholder, :foo}
 
+      intensity_query = from p in Post, where: p.id == ^0, select: p.intensity
+
       entries = [
-        %{intensity: 1.0, title: bar_ph, posted: ~D[2020-12-21], visits: foo_ph},
+        %{intensity: intensity_query, title: bar_ph, posted: ~D[2020-12-21], visits: foo_ph},
```
greg-rychlewski, line 1049: This altered test would fail with the previous code because `intensity` would be the first value acted on in `plan_query_in_rows`. Since the counter in that function used to start at 0, the parameters in this query would not be offset by the number of placeholders.

integration_test/cases/repo.exs
```diff
@@ -1075,25 +1075,25 @@ defmodule Ecto.Integration.RepoTest do
...
-      placeholders = %{posted: Date.utc_today(), title: "title"}
+      placeholders = %{visits: 1, title: "title"}
 
       post1 = [
-        visits: visits_value,
+        visits: {:placeholder, :visits},
         title: {:placeholder, :title},
         uuid: Ecto.UUID.generate(),
-        posted: {:placeholder, :posted}
+        posted: posted_value
```
greg-rychlewski, line 1090: This test should have caught the bug, but previously the placeholder values were being counted before the `visits` query in `plan_query_in_rows`. So the effect was the same as if the placeholder count was passed into this function.

When I make it so that the query comes before the placeholder, the test fails.

josevalim:
Sorry, I pressed the  merge button ionstead of approve. Habit! :D

Issue #3933, opened 2022-06-26
Send casted parameters to adapters for `insert_all`


greg-rychlewski:
After https://github.com/elixir-ecto/ecto/commit/f1e7afd7e3d850cc5ced50cb90f21ceb768696eb all the Ecto operations that produce queries will send the casted version of the parameters to the adapters, except for `insert_all`. 

`insert_all` requires extra care because the literal/query value parameters produced in Ecto are interleaved in the adapters. This will need refactoring in Ecto to handle correctly.

PR #3941, opened 2022-06-30
Send insert_all cast params to adapters

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/3933.

This completes the work done in https://github.com/elixir-ecto/ecto/pull/3869 to send the cast version of the query parameters to the adapters.

Corresponding ecto_sql PR: https://github.com/elixir-ecto/ecto_sql/pull/424

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -100,8 +100,8 @@ defmodule Ecto.Repo.Schema do
       |> Enum.map(&{elem(&1, 1), elem(&1, 2)})
       |> Enum.unzip
 
-    {rows, cast_params, counter} = plan_query_in_rows(rows, header, adapter, placeholder_size)
-    {rows, header, cast_params, placeholder_cast_params, placeholder_dump_params, fn -> counter end}
+    {rows, row_cast_params, counter} = plan_query_in_rows(rows, header, adapter, placeholder_size)
+    {rows, header, row_cast_params, placeholder_cast_params, placeholder_dump_params, fn -> counter end}
   end
 
   defp extract_header_and_fields(repo, %Ecto.Query{} = query, _schema, _dumper, _autogen_id, _placeholder_map, adapter, opts) do
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -44,7 +44,7 @@ defmodule Ecto.Repo.Schema do
       |> returning(opts)
       |> fields_to_sources(dumper)
 
-    {rows_or_query, header, placeholder_values, counter} =
+    {rows_or_query, header, row_cast_params, placeholder_cast_params, placeholder_dump_params, counter} =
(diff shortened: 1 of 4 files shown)


lib/ecto/repo/schema.ex
```diff
@@ -81,44 +82,33 @@ defmodule Ecto.Repo.Schema do
...
-        Enum.reduce(
-          rows,
-          placeholder_size,
-          &(Enum.count(&1, fn {_, val} -> not match?({:placeholder, _}, val) end) + &2)
-        )
-      end
-
-      {rows, header, placeholder_vals_list, counter}
-    end
+    {rows, cast_params, counter} = plan_query_in_rows(rows, header, adapter, placeholder_size)
```
greg-rychlewski, line 103: This is the biggest compromise that had to be made to accommodate this change. When none of the rows contains a query, we still have to traverse and re-order the fields in the rows here.

There is another option that can be done to avoid this. We can send the casted parameters to the adapter without re-ordering them and add the logic for re-ordering them in the adapter. The downside of this is that we are putting more of the cast parameter logic into the adapters for `insert_all` when none of the other query methods are doing so.

josevalim: I assume that before it was responsibility of the adapters to sort the fields? If we are sorting them here now, does it mean it less work to do on the adapters too?

greg-rychlewski: Before this PR the fields would be sorted once in Ecto and once in the adapters if any of the rows had a query for a value. And only sorted in the adapters if there weren't any query value.

Ecto needed to sort them in order to have the correct query parameter count when planning the row queries. If there were no queries then it wouldn't be sorted in Ecto. 

The adapters needed to sort them as well because rows could be missing fields present in other rows that need to have their [values converted to `nil`](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/sql.ex#L838) so that they can have their. value [replaced by the keyword `DEFAULT`](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L241). 

I originally wanted to take the sorting out of the adapter but it would mean changing the value of the rows going into the adapter. Which would be ok for anything that uses `Ecto.Adapters.SQL` but not custom ones that aren't using it.

The other option I see is to pass the unsorted cast parameters into the adapter and then once inside the adapter zip them with the rows and sort them here: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/sql.ex#L826.

josevalim: Hrm... so what you are saying is that we always keep the params and cast_params in the same order? So if the adapter needs to re-order one, then it reorders both?

Yeah, I actually like that more then. Let's go with it. :)

greg-rychlewski: Done :).  The code is a bit more complex because of the need to keep track of which row the cast params are in, but it gets rid of the double sorting when there is no row query.  ecto_sql changes: https://github.com/elixir-ecto/ecto_sql/pull/424

PR #3943, opened 2022-07-11
Add support for settings idle_interval in connection URL

AndrewDryga:
Related change in Postgrex: https://github.com/elixir-ecto/postgrex/commit/031018f8b6953ab2ef1efb22da7ebaeb40572755

### DIFF
--- a/lib/ecto/repo/supervisor.ex
+++ b/lib/ecto/repo/supervisor.ex
@@ -35,7 +35,7 @@ defmodule Ecto.Repo.Supervisor do
   defp telemetry_prefix(repo) do
     repo
     |> Module.split()
-    |> Enum.map(&(&1 |> Macro.underscore() |> String.to_atom()))
+    |> Enum.map(& &1 |> Macro.underscore() |> String.to_atom())
   end
 
   defp repo_init(type, repo, config) do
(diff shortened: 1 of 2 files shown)


lib/ecto/repo/supervisor.ex
```diff
@@ -58,8 +58,9 @@ defmodule Ecto.Repo.Supervisor do
     end
 
     if Code.ensure_compiled(adapter) != {:module, adapter} do
-      raise ArgumentError, "adapter #{inspect adapter} was not compiled, " <>
-                           "ensure it is correct and it is included as a project dependency"
+      raise ArgumentError,
+            "adapter #{inspect(adapter)} was not compiled, " <>
+              "ensure it is correct and it is included as a project dependency"
```
AndrewDryga, line 63: Should I roll back or keep formatter changes? They don't look so bad and I guess we want to apply formatter on codebase as we change it.

josevalim: No formatter changes, please. It makes it harder to review and messes up blame. :)

AndrewDryga: @josevalim true, I tell that to people all the time too but still did it for some reason 🤣. It's fixed!

josevalim:

PR #3945, opened 2022-07-15
cast_embed/3 shouldn't register changes if value is unchanged

tcrossland:


josevalim:
Hi @tcrossland, as mentioned in the mailing list, the changeset can keep important information such as which validations were attempted and other metadata. You can perform this logic in your app but we don't plan to do it by default on Ecto. Thank you!

tcrossland:
Thanks Jose, happy to continue the discussion on the mailing list, if you have a moment, please take a brief look at the test case which demonstrates what I believe is an issue.

josevalim:
I don't believe it is an issue. For nested associations/embeds, we will always keep the inner changeset in changes, and they need to be visited for changes. Perhaps this is leading to a bug elsewhere in Repo, but the code as is is correct.

josevalim:
To further add to this, imagine that in the child changeset you add a `prepare_changeset` callback, which _will_ add a change when given to the repository. If we discard it upfront, we are changing the behaviour.

tcrossland:
Understood, perhaps the insert+replace behaviour makes sense for an association, but it seems strange to me for an embedded schema (e.g. an unchanged jsonb document). Anyway, we'll control this in our app as you suggested. Thanks for your feedback.

tcrossland:
Also worth noting that if the embedded schema has a primary key no change is registered.

tcrossland:
One final try before I give up... discarding unchanged values from a valid changeset's `changes` could be controlled by an option (disabled by default), using `delete_change/2` instead of returning the original changeset:

https://github.com/elixir-ecto/ecto/compare/master...tcrossland:cast_embed_issue?diff=split

I hope I'm not being "pesado", but I think detecting changes is quite an important use case for Changeset.

josevalim:
If the goal is to detect changes, I would rather have a changed?(…) function in Ecto.Changeset. As mentioned, discarding the underlying changeset is not something we should do.

tcrossland:
> If the goal is to detect changes, I would rather have a changed?(…) function in Ecto.Changeset. As mentioned, discarding the underlying changeset is not something we should do.

I thought that was the whole purpose of the `changes` field ("The changes from parameters that were approved in casting"). The trouble in this case is that an **unchanged** value ends up in the Changeset's `changes`. So a `changed?/1` function could do something like this...

```elixir
def changed?(%Changeset{data: data} = changeset, name) when is_map(data) do
  original = Map.get(data, name)

  case fetch_field(changeset, name) do
    {:changes, changes} -> changes != original
    _ -> false
  end
end
```

... but it would be a lot nicer if the unchanged value didn't end up in `changes` in the first place (hence the idea of discarding - it should be a no-op if no change exists).

An alternative approach would be to support `embeds_many` with `on_replace: :update` to ignore unchanged nested embeds in the same way as `embeds_one` does. But that looks like a lot of work for something that's trivial to detect a posteriori.

Anyway, I don't want to waste your time, and if I haven't managed to convince you yet that it's an issue it's not for lack of trying :cry:. And I'll still :heart: Ecto.

josevalim:
I meant `changed?(changeset)` which traverses all changes recursively guaranteeing they are all empty.

tcrossland:
> I meant `changed?(changeset)` which traverses all changes recursively guaranteeing they are all empty.

I'm not sure that would be useful in this case (`embeds_many on_replace: :delete` without primary key) as each unchanged item has an `:insert` changeset despite the fact that the params match the existing data.

PR #3946, opened 2022-07-16
Improve docs about aborted transactions

thiagomajesk:
I made some changes to the initial docs by @josevalim trying to make language a little more accessible and easy to read by someone unfamiliar with some concepts.

- The first change reverts the "Elixir exception" part to avoid redundancy. This avoids introducing a perceived new concept since we talk about exceptions in other places always meaning elixir exceptions.
-  Updates the example so it mirrors the failure scenario making it clear that we want to execute some code after the first operation. The idea is to make sure users  understand that there's nothing special about the `{:error, changeset}` part.
- Changes the order of the explanation to prioritize what users might search after having this transaction error thrown at them.  I don't think we have to explain what happens with the success scenario, only the unexpected one (changeset valid + error from constraint). It's less of an effort since we are explaining just one thing.
- I tried to simplify some explanations to be more direct: "if x happens, y is the result", instead of: "when x happens, under y, z is the result".
- Add other options that might be useful to deal with this kind of problem besides just handling outside of the transaction (most of this was dug up from the forums in old posts and I think most people didn't even remember those are options).

PS.: I tried to pack some more information and make some stylistic changes to expand the concepts and streamline the language focusing on someone not well versed as the maintainers. More so, from the optics of someone just recently challenged by the docs and the overall technical language 🙈.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1760,7 +1760,7 @@ defmodule Ecto.Repo do
         repo.insert!(%Post{})
       end)
 
-  If an Elixir exception occurs the transaction will be rolled back
+  If an exception occurs the transaction will be rolled back
   and the exception will bubble up from the transaction function.
   If no exception occurs, the transaction is committed when the
   function returns. A transaction can be explicitly rolled back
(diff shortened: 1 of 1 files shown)


lib/ecto/repo.ex
```diff
@@ -1760,7 +1760,7 @@ defmodule Ecto.Repo do
         repo.insert!(%Post{})
       end)
 
-  If an Elixir exception occurs the transaction will be rolled back
+  If an exception occurs the transaction will be rolled back
```
josevalim, line 1763: Databases also have exceptions. So since you said errors were confusing, it doesn't hurt to be double explicit. :)

```suggestion
  If an Elixir exception occurs the transaction will be rolled back
```

lib/ecto/repo.ex
```diff
@@ -1828,43 +1828,51 @@ defmodule Ecto.Repo do
...
+  If the changeset is invalid before it reaches the database due to a validation error, 
+  no statement is sent to the database, an `:error` tuple is returned, and `repo.insert(%Failure{})` 
+  operation will execute as usual. 
+
+  We have a few options to deal with such scenarios:
+  
+  The most obvious one would be to prevent the the database from raising using 
+  something like `repo.insert(post, on_conflict: :nothing)`. In this case, the 
+  operation will still fail at the database level, but since we are supressing the error, 
+  we will receive back a `:ok` tuple with an unsaved post: `%Post{id: nil}`.
```
josevalim, line 1864: This is actually much more complex because `on_conflict: :nothing` is database specific and you can match on `id: nil` only if you are using auto generated primary keys and it also means the error is no longer added to the changeset. I am thinking we should not recommend this path:

```suggestion
  We have two options to deal with such scenarios:
```

lib/ecto/repo.ex
```diff
@@ -1828,43 +1828,46 @@ defmodule Ecto.Repo do
...
+  Take the following transaction as an example:
 
       Repo.transaction(fn repo ->
         case repo.insert(changeset) do
           {:ok, post} ->
-            post
+            repo.insert(%Success{})
 
           {:error, changeset} ->
             repo.insert(%Failure{})
```
josevalim, line 1844:
```suggestion
            repo.insert(%Status{value: "success"})

          {:error, changeset} ->
            repo.insert(%Status{value: "failure"})
```

josevalim:

lib/ecto/repo.ex
thiagomajesk: Don’t you think that we should at least present this as an option with some remarks about it? I’d rather making clear it’s not recommended than denying the information (I’m surprised how this didn’t come up more often in the forums btw).

I’m coming from the perspective of giving sharp tools to users and let them decide. 

I trust your judgment though if you think this absolutely won’t improve anything. :)

lib/ecto/repo.ex
thiagomajesk: Isn’t the transaction rolled back in both cases here? So shouldn’t we understand “exceptions” as both application exceptions and database exceptions?

The original text was about unhandled erros and I think that this gave a more generic understanding that it could be from application or database. WDYT?

josevalim: The rest of the sentence does not apply to DB transactions.

PR #3947, opened 2022-07-16
Improve aborted transaction docs

thiagomajesk:
I overlooked the savepoints feature in https://github.com/elixir-ecto/ecto/pull/3946 and found that it's actually an SQL standard. It seems it's currently supported by all our default adapters, so we shouldn't make the case that it's specific to Postgres.

BTW, Postgrex has this option documented here https://hexdocs.pm/postgrex/Postgrex.html#transaction/3, but I had to look up the source code to find that MyXQL and TDS also support the savepoints feature. Should we document this as part of Ecto `Repo.transaction/2` as well?

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1858,10 +1858,10 @@ defmodule Ecto.Repo do
 
   We have two options to deal with such scenarios:
   
-  If you are using Postgres and don't want to change the semantics of your code, 
-  you can also use the savepoints feature by passing the `:mode` option like this: 
-  `repo.insert(changeset, mode: :savepoint)`. In case of an exception, this will make 
-  the transaction rollback to the savepoint and prevent the transaction from failing.
+  If don't want to change the semantics of your code,  you can also use the savepoints 
+  feature by passing the `:mode` option like this: `repo.insert(changeset, mode: :savepoint)`. 
+  In case of an exception, this will make the transaction rollback to the savepoint and prevent 
+  the transaction from failing.
 
   Another alternative is to handle this operation outside of the transaction. 
   For example, you can choose to perform an explicit `repo.rollback` call in the 
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1860,7 +1860,7 @@ defmodule Ecto.Repo do
   
   If don't want to change the semantics of your code,  you can also use the savepoints 
   feature by passing the `:mode` option like this: `repo.insert(changeset, mode: :savepoint)`. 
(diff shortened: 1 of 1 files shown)


lib/ecto/repo.ex
```diff
@@ -1858,10 +1858,10 @@ defmodule Ecto.Repo do
 
   We have two options to deal with such scenarios:
   
-  If you are using Postgres and don't want to change the semantics of your code, 
-  you can also use the savepoints feature by passing the `:mode` option like this: 
-  `repo.insert(changeset, mode: :savepoint)`. In case of an exception, this will make 
-  the transaction rollback to the savepoint and prevent the transaction from failing.
+  If don't want to change the semantics of your code,  you can also use the savepoints 
+  feature by passing the `:mode` option like this: `repo.insert(changeset, mode: :savepoint)`. 
+  In case of an exception, this will make the transaction rollback to the savepoint and prevent
```
josevalim, line 1863:
```suggestion
  In case of an exception, the transaction will rollback to the savepoint and prevent 
```

josevalim:

josevalim:
Adding more information to transaction is welcome, as well as PRs to MyXQL and TDS. :)

thiagomajesk:
Hi @josevalim! Thanks for your feedback. One question I have and I'm not sure where this discussion should go...

The documentation for transactions is currently scattered across [Ecto](https://hexdocs.pm/ecto/Ecto.Repo.html#c:transaction/2), [Postgrex](https://hexdocs.pm/postgrex/Postgrex.html#transaction/3), [MyXQL](https://hexdocs.pm/myxql/MyXQL.html#transaction/3), [Tds](https://hexdocs.pm/tds/Tds.html#transaction/3) and [DBConnection](https://hexdocs.pm/db_connection/2.4.1/DBConnection.html#transaction/3) and I think we need to think about a strategy to document it that avoids repeating the same information over and over.

- Ecto links to its own shared options
- Postgrex docs lists some options
- MyXQL links to the options present in DBConnection
- Tds doesn't have documentation about it

I think we should find some what's common in all these layers and document only what's different between them. But in your experiences, what would be the best way to do it?

PS.: I wonder if this isn't the case for other options and this is actually a much deeper work to tidy up the docs 🙃.

josevalim:
We should generally document all useful options in Ecto, except for the places we explicitly link to the adapter. But some options don't make sense to discuss in Ecto, because DBConnection is more general purpose than Ecto and some options are not useful for the Ecto case. I think `:mode` should be documented in Ecto though.

Issue #3948, opened 2022-07-17
 ** (Postgrex.Error) ERROR 42P01 (undefined_table) missing FROM-clause entry for table "s0"


ns-blee:
### Elixir version

Elixir 1.13.0 

### Database and Version

14.4 (Ubuntu 14.4-1.pgdg20.04+1), server 12.11 (Ubuntu 12.11-1.pgdg20.04+1)

### Ecto Versions

3.8.4

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex

### Current behavior

Below I have provided the working SQL equivalent of what Im building in Ecto however Im given  `** (Postgrex.Error) ERROR 42P01 (undefined_table) missing FROM-clause entry for table "s0"`

Thank you for your time and help!

```
def get_top(_, table_list, from, to) do
    order_by = [
    desc_nulls_last: dynamic([d], fragment("(?->>'async')::float", d.data)),
  ]

 subQ = Enum.reduce(table_list, %Ecto.Query{}, fn table, acc ->
  query =
    from(d in table[:table_name],
      distinct: d.data["data_id"],
      where: d.message_timestamp >= ^from,
      where: d.message_timestamp <= ^to,
      order_by: ^order_by,
      select: %{
        id: d.id,
        message_timestamp: d.message_timestamp,
        receipt_timestamp: d.receipt_timestamp,
        async: fragment("(data->>'async')::float"),
        outer: fragment("(data->>'outer')::float"),
        data_id: d.data["data_id"],
        table: ^table[:table_name]
      }
    )
    cond do
      acc == %Ecto.Query{} ->
        query

      true ->
        query |> union(^acc)
    end
  end)

  TimeseriesRepo.all(
    from(d in subquery(subQ),
      order_by: [
        desc_nulls_last: d.async,
      ],
      limit: 20,
      select: %{
        id: d.id,
        message_timestamp: d.message_timestamp,
        receipt_timestamp: d.receipt_timestamp,
        async: d.async,
        outer: d.outer,
        data_id: d.tenant_id,
        table_name: d.table
      }
    )
  )
end
```

### Expected behavior

SELECT * FROM (
[...]

PR #3950, opened 2022-07-18
combination order by documentation

greg-rychlewski:
Wanted to supplement the combination query docs because I saw this kind of question surface a few times in the past few months https://github.com/elixir-ecto/ecto/issues/3948

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1526,17 +1526,34 @@ defmodule Ecto.Query do
   consider using `union_all/2`.
 
   Note that the operations `order_by`, `limit` and `offset` of the
-  current `query` apply to the result of the union.
+  current `query` apply to the result of the union. `order_by` must
+  be specified in one of the following ways, since the union of two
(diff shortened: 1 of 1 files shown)


lib/ecto/query.ex
```diff
@@ -1526,18 +1526,35 @@ defmodule Ecto.Query do
   consider using `union_all/2`.
 
   Note that the operations `order_by`, `limit` and `offset` of the
-  current `query` apply to the result of the union.
+  current `query` apply to the result of the union. `order_by` must
+  be specified in one of the following ways, since the union of two
+  or more queries is not automatically aliased:
 
-  ## Keywords example
+    - Use `Ecto.Query.API.fragment/1` to pass an unbound `order_by` statement in a query expression.
```
josevalim, line 1533:
```suggestion
    - Use `Ecto.Query.API.fragment/1` to pass an `order_by` statement that directly access the union fields.
```

?

josevalim (approved):

PR #3951, opened 2022-07-19
Delete ISSUE_TEMPLATE.md

greg-rychlewski:
I believe it's okay to delete this since the repo is using the new templates under `.github` but please let me know if I'm being presumptuous.


wojtekmach (approved):
Nice catch!

PR #3952, opened 2022-07-20
Builder: Fix escaping subqueries

arnodirlam:
Fixes escaping of subqueries.

This broke in #3928 and was not properly reverted as part of the refactor in #3935 (see [comparison](https://github.com/elixir-ecto/ecto/compare/f1e7afd7e3d850cc5ced50cb90f21ceb768696eb..412e5f51f61f62b51a7506be1c29f2b55c9811be#diff-28c09a2701e1bdc727200dd4b3515822eb08c2d502ba20e7b132ae5926ce1e33L171-R188)).

`master` has been affected since #3928, but no releases.

I tested it using ecto `master` in another project. Not sure how to test it within `ecto`. Appreciate any leads! 🙏

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -179,7 +179,7 @@ defmodule Ecto.Query.Builder do
   end
 
   # subqueries
-  def escape({:subquery, _, _} = expr, _, {params, %{subqueries: subqueries} = acc}, _vars, _env) do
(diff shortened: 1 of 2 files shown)


josevalim:
You should be able to write a failing test in `filter_test.exs` as long as your tests unimports the subquery functions from `Ecto.Query`. Something like this:

```
import Ecto.Query, except: [subquery: 1]
from ..., where: p.foo in subquery(...)
```

arnodirlam:
Great, that worked! 🎉

josevalim:

PR #3953, opened 2022-07-21
Fix many-to-many on_delete options

tfwright:
Seeing the following error when using `delete_all` as documented:

```
** (ArgumentError) invalid `:on_replace` option for :my_assoc. The only valid options are: `:raise`, `:mark_as_invalid`, `:delete`
```

Not sure if this change is the best path forward but it seemed like the easiest way to report the issue.


josevalim:
Thank you! The error message says `on_replace`, but you changed the docs for `on_delete`, which one is it? :)

tfwright:
🤦 my mistake, delete_all is correct for on_delete, sorry!

PR #3954, opened 2022-07-21
Update example for cast/2

hl:
The example was missing an `{:ok, ...}` tuple as the result of `cast/2`


josevalim:

PR #3955, opened 2022-07-23
Allow `unsafe_validate_unique/4` on Schemaless Changesets with Query Option

aj-foster:
Hello there 👋🏼 This PR is for discussion and ideas; the code is secondary. 🙂 I also intend to send a message to the mailing list. (Edit: it's [here](https://groups.google.com/g/elixir-ecto/c/bn9aItlVH0Q))

[In 2017](https://groups.google.com/g/elixir-ecto/c/BQr-mUZ2zH8/m/6A-JpDfOAwAJ), the mailing list discussed adding a `schema` option to `Ecto.Changeset.unsafe_validate_unique/4` in order to support validation of schemaless changesets. Since then, there is a new option `query` for overriding the base query. Although its goal is unrelated to schemaless changesets, it provides an opportunity to support this validation without the additional option.

This PR suggests one way to accomplish this.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1878,8 +1878,8 @@ defmodule Ecto.Changeset do
       case changeset do
         %Ecto.Changeset{validations: validations, data: %schema{}} ->
           {validations, schema}
-        %Ecto.Changeset{} ->
-          raise ArgumentError, "unsafe_validate_unique/4 does not work with schemaless changesets"
+        %Ecto.Changeset{validations: validations} ->
+          {validations, nil}
       end
     changeset = %{changeset | validations: [{hd(fields), {:unsafe_unique, fields: fields}} | validations]}
 
@@ -1899,8 +1899,12 @@ defmodule Ecto.Changeset do
     if unrelated_changes? || any_nil_values_for_fields? || any_prior_errors_for_fields? do
       changeset
     else
+      base_query =
+        Keyword.get(opts, :query, schema) ||
+          raise "unsafe_validate_unique/4 requires specifying the `query` option with schemaless changesets"
+
       query =
-        Keyword.get(opts, :query, schema)
+        base_query
         |> maybe_exclude_itself(schema, changeset)
         |> Ecto.Query.where(^where_clause)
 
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1922,6 +1926,8 @@ defmodule Ecto.Changeset do
     end
   end
 
+  defp maybe_exclude_itself(base_query, nil, _changeset), do: base_query
+
```
v0idpwn, line 1930: Could we virtually support this, maybe through an extra option?

v0idpwn (commented):
Thanks for the PR! The code looks great to me. While I agree on the implementation, I'm a bit inclined to disagree in the fundamental idea: is it a unique validation if it takes an **arbitrary** query? The reason we support a query there is not for arbitrary use but for cases when just checking if a field is set isn't enough (such as when there are partial unique indexes)

greg-rychlewski:
> [In 2017](https://groups.google.com/g/elixir-ecto/c/BQr-mUZ2zH8/m/6A-JpDfOAwAJ), the mailing list discussed adding a schema option to Ecto.Changeset.unsafe_validate_unique/4 in order to support validation of schemaless changesets

When I read that thread it seems the OP's intention is a bit different. It sounds like they want to validate embedded schema fields by specifying the parent schema they belong to.

aj-foster:
> The reason we support a query there is not for arbitrary use but for the case of partial unique indexes.

That's totally fair. This is definitely an abuse of the `query` option. It's the kind of shortcut that is helpful for my particular use-case, but perhaps not appropriate for the broad audience of the library.

> It sounds like they want to validate embedded schema fields by specifying the parent schema they belong to.

That's true, I'm conflating embedded schemas and schemaless changesets here. Although the effect is similar: the `query` option can act as a bridge between the changeset data and the uniqueness query.

greg-rychlewski:
It looks like the intention behind the function is centered around primary keys though

> Validates that no existing record with a different primary key
  has the same values for these fields

So allowing any arbitrary query confuses the scope of the function a bit. 

The only thing I can think of, based on [this comment by @v0idpwn](https://github.com/elixir-ecto/ecto/pull/3955#discussion_r928086815), is to allow a `primary_keys` option and use that for schemaless changesets. But it's kind of getting into the territory Jose was worried about:

>That folks would need more and more complex options when running the validation and the consensus was that you should just roll your own validation logic in those cases

aj-foster:
That all sounds fair. As with the original discussion, it's easy enough to "roll your own" validation logic for this sort of thing. Thanks for all of your input!

PR #3956, opened 2022-07-23
Raise when schemaless struct is used in unsafe_validate_unique/4

greg-rychlewski:
I noticed this when looking over this PR: https://github.com/elixir-ecto/ecto/pull/3955.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1873,14 +1873,12 @@ defmodule Ecto.Changeset do
   """
   @spec unsafe_validate_unique(t, atom | [atom, ...], Ecto.Repo.t, Keyword.t) :: t
   def unsafe_validate_unique(%Changeset{} = changeset, fields, repo, opts \\ []) when is_list(opts) do
-    %{validations: validations, data: data} = changeset
+    {validations, schema} =
+      case changeset do
+        %{validations: validations, data: %schema{__meta__: %Metadata{}}} ->
+          {validations, schema}
 
-    schema =
-      case data do
-        %schema{__meta__: %Metadata{}} ->
-          schema
-
-        _ ->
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset.ex
```diff
@@ -1877,6 +1877,10 @@ defmodule Ecto.Changeset do
     {validations, schema} =
       case changeset do
         %Ecto.Changeset{validations: validations, data: %schema{}} ->
+          unless function_exported?(schema, :__schema__, 1) do
+            raise ArgumentError, "unsafe_validate_unique/4 does not work with schemaless changesets"
+          end
+
           {validations, schema}
         %Ecto.Changeset{} ->
```
josevalim, line 1885: Perhaps it is no longer worth having this clause? Check if it is a struct and if function_exported? is defined instead in a single conditional, without the case?

greg-rychlewski: Oh yeah good point, I will make that change. Btw this function won't work for embeds either right? Would it also be ok to add a check for the `__meta__` field and raise in its absence?

josevalim: I think pattern matching on __meta__ is enough, yeah!

greg-rychlewski: Updated the code. I think it has to be split like this otherwise someone can randomly put `__meta__` in a non-schema struct? But let me know if I'm being a newb.

josevalim:
You can pattern match if the `__meta__` field has the Ecto.Schema.Metadata struct. That should be enough and a single `case` should do it?

greg-rychlewski:
Ah thank you. Sorry I had it in my head that metadata was a regular map.

josevalim (approved):

Issue #3957, opened 2022-07-24
Test fails: test fully_expand/2 with subquery and dynamic interpolation


Sleepful:
### Elixir version

Elixir 1.13.4 (compiled with Erlang/OTP 25)

### Database and Version

none

### Ecto Versions

master branch

### Database Adapter and Versions (postgrex, myxql, etc)

none

### Current behavior

`mix test` fails on 5 test cases, submitting this issue only for 1 test case first in case I am doing something wrong. 
This is the first case it fails on:
```
1) test fully_expand/2 with subquery and dynamic interpolation (Ecto.Query.Builder.DynamicTest)
     test/ecto/query/builder/dynamic_test.exs:51
     Assertion with == failed
     code:  assert Macro.to_string(expr) ==
              "&0.foo() == ^0 and (&0.bar1() == ^1 or &0.sq() in {:subquery, 0} or &0.bar3() == ^3) and &0.baz() == ^4"
     left:  "&0.foo() == ^0 and (&0.bar1() == ^1 or &0.sq() in {:subquery, 0} or &0.bar3() == ^3) and\n  &0.baz() == ^4"
     right: "&0.foo() == ^0 and (&0.bar1() == ^1 or &0.sq() in {:subquery, 0} or &0.bar3() == ^3) and &0.baz() == ^4"
     stacktrace:
       test/ecto/query/builder/dynamic_test.exs:57: (test)
```

### Expected behavior

Does not fail

josevalim:
Yes! That's because `Macro.to_string/2` is adding newlines on Elixir v1.13. We should add a if/else branch to those particular tests, similar to other places in the codebase (search for `if Version.match?(System.version(), ">= 1.13.0-dev") do`).

josevalim:
PRs welcome!

PR #3958, opened 2022-07-26
Fix tests failing due to Elixir 1.13 string representation of macros

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/3957.

### DIFF
--- a/test/ecto/query/inspect_test.exs
+++ b/test/ecto/query/inspect_test.exs
@@ -268,15 +268,29 @@ defmodule Ecto.Query.InspectTest do
            ~s{from p0 in Inspect.Post, lock: "FOOBAR"}
   end
 
-  test "preload" do
-    assert i(from(x in Post, preload: :comments)) ==
-           ~s"from p0 in Inspect.Post, preload: [:comments]"
+  # TODO: AST is represented as string differently on versions pre 1.13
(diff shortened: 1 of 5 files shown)


test/ecto/query/inspect_test.exs
```diff
@@ -268,15 +268,29 @@ defmodule Ecto.Query.InspectTest do
            ~s{from p0 in Inspect.Post, lock: "FOOBAR"}
   end
 
-  test "preload" do
-    assert i(from(x in Post, preload: :comments)) ==
-           ~s"from p0 in Inspect.Post, preload: [:comments]"
+  # TODO: AST is represented as string differently on versions pre 1.13
+  if Version.match?(System.version(), ">= 1.13.0-dev") do
```
greg-rychlewski, line 272: I'm not 100% sure if this one is the same as the others. It is removing the brackets around the keyword list

josevalim (approved):

greg-rychlewski:
@josevalim Is there a particular strategy Ecto uses for choosing Elixir versions to run the CI against? It's a couple versions behind now so I could work on updating the CI, if it's something that makes sense.

josevalim:
Typically three:

1.  lowest Elixir version with lowest OTP version
2. highest Elixir version with lowest OTP version
3. highest Elixir version with highest OTP version

PR #3962, opened 2022-07-28
Document omitting fields in select

sabiwara:
In https://github.com/elixir-ecto/ecto/pull/3934, the `%{schema | my_field: nil}` syntax stopped retrieving the `my_field` from the DB, making it a viable way of achieving omitting fields (a feature discussed [here](https://groups.google.com/g/elixir-ecto/c/9lCaPISWebg)).

Should this be part of the documented public API?

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1267,6 +1267,11 @@ defmodule Ecto.Query do
 
   If you want a map with only the selected fields to be returned.
 
+  To select a struct but omit only given fields, you can
+  override them with `nil` or another default value:
+
+      from(city in City, select: %{city | geojson: nil})
+
   For more information, read the docs for `Ecto.Query.API.struct/2`
   and `Ecto.Query.API.map/2`.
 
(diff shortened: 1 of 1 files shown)


josevalim:
Do we apply to only `nil` values or any literal value? What about `%{p | text: "skipped"}`? In any case, documenting it sounds good. :)

sabiwara:
> Do we apply to only nil values or any literal value? What about %{p | text: "skipped"}? In any case, documenting it sounds good. :)

Any literal works 👍. It would probably be good to add a `%{c | text: "<redacted>"}` to the examples as well to illustrate this, but also keep the `nil` example  since it is probably the most common thing to do when omitting data. Will push.

Additional question: do you think this should be mentioned in the changelog?

josevalim:
:+1: for changelog too!

josevalim:

sabiwara:
nvm it was already [here](https://github.com/elixir-ecto/ecto/commit/cf445edc0deb05b9a6fe7f729912bde8f7cc6034#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR16) 🙂

PR #3964, opened 2022-07-28
Improve embedded schemas guide

inaniyants:
Examples in guide were missing `schema` argument for `embeds_one`. 
Also I added correct references to `Ecto.Schema` functions and more details for inline and explicit use cases

### DIFF
--- a/guides/introduction/Embedded Schemas.md
+++ b/guides/introduction/Embedded Schemas.md
@@ -63,7 +63,6 @@ end
 ```
 
 Defining embedded schema in such a way is called inline-defined embedded schemas which are:
-- the simpliest way to define embedded schema
 - generating module in parent scope with the appropriate struct (for example above, module will be `User.Profile`)
 - persited within parent schema
 - requiring to provide `with` option to `Ecto.Changeset.cast_embed/3`
--- a/guides/introduction/Embedded Schemas.md
+++ b/guides/introduction/Embedded Schemas.md
@@ -62,7 +62,8 @@ defmodule User do
 end
 ```
 
-Defining embedded schema in such a way is called inline-defined embedded schemas which are:
+Embedded schemas defined in such way are said to be defined inline, which:
+
 - generating module in parent scope with the appropriate struct (for example above, module will be `User.Profile`)
 - persited within parent schema
 - requiring to provide `with` option to `Ecto.Changeset.cast_embed/3`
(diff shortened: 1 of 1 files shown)


guides/introduction/Embedded Schemas.md
```diff
@@ -62,9 +62,15 @@ defmodule User do
 end
 ```
 
+Defining embedded schema in such a way is called inline-defined embedded schemas which are:
+- the simpliest way to define embedded schema
```
greg-rychlewski, line 66:
```suggestion
```
I would personally take this out, given the other use case that is highlighted further down:

> Like was said before, they are persistence agnostic. This makes embedded schemas ideal for scenarios where you want to manage structured data without necessarily persisting it.

So it's only the simplest if you aren't doing the above.

inaniyants: That's reasonable point of view. Besides this I have a typo in this line, so I'll totally remove it, like you suggested.
Thank you.

guides/introduction/Embedded Schemas.md
```diff
@@ -95,7 +100,12 @@ defmodule UserProfile do
...
+Defining embedded schema in such a way is called explicit-defined embedded schemas, which:
+- are dedicated modules having own scope, changeset functions, props, documentation, etc...
+- could be embedded by multiple parent schemas
+- are persistence agnostic, which means that `embedded_schema` doesn't require to be persisted
+
+It is important to remember that `embedded_schema` has many use cases independent of `embeds_one` and `embeds_many`. Like was said before, they are persistence agnostic. This makes embedded schemas ideal for scenarios where you want to manage structured data without necessarily persisting it. For example, if you want to build a contact form, you still want to parse and validate the data, but the data is likely not persisted anywhere. Instead, it is used to send an email. Embedded schemas would be a good fit for such a use case.
```
josevalim, line 108:
```suggestion
It is important to remember that `embedded_schema` has many use cases independent of `embeds_one` and `embeds_many`. As they are persistent agnostic, they are ideal for scenarios where you want to manage structured data without necessarily persisting it. For example, if you want to build a contact form, you still want to parse and validate the data, but the data is likely not persisted anywhere. Instead, it is used to send an email. Embedded schemas would be a good fit for such a use case.
```

guides/introduction/Embedded Schemas.md
```diff
@@ -95,7 +100,12 @@ defmodule UserProfile do
 end
 ```
 
-It is important to remember that `embedded_schema` has many use cases independent of `embeds_one` and `embeds_many`. You can think of embedded schemas as persistence agnostic `schema`s. This makes embedded schemas ideal for scenarios where you want to manage structured data without necessarily persisting it. For example, if you want to build a contact form, you still want to parse and validate the data, but the data is likely not persisted anywhere. Instead, it is used to send an email. Embedded schemas would be a good fit for such a use case.
+Defining embedded schema in such a way is called explicit-defined embedded schemas, which:
```
josevalim, line 103:
```suggestion
Embedded schemas defined in such way are said to be explicit-defined, which:

```

guides/introduction/Embedded Schemas.md
```diff
@@ -62,9 +62,14 @@ defmodule User do
 end
 ```
 
+Defining embedded schema in such a way is called inline-defined embedded schemas which are:
```
josevalim, line 65:
```suggestion
Embedded schemas defined in such way are said to be defined inline, which:

```

josevalim:

PR #3966, opened 2022-07-29
Allow alias for selected value

greg-rychlewski:
This can come in handy when there is a complicated select value that doesn't require `fragment` and is referenced in `order_by` and/or `group_by`.

Corresponding ecto_sql PR: https://github.com/elixir-ecto/ecto_sql/pull/432

### DIFF
--- a/test/ecto/query/builder/select_test.exs
+++ b/test/ecto/query/builder/select_test.exs
@@ -88,43 +88,43 @@ defmodule Ecto.Query.Builder.SelectTest do
       end
     end
 
-    test "supports aliasing a selected value with alias/2" do
-      escaped_alias = {:alias, [], [{{:., [], [{:&, [], [0]}, :id]}, [], []}, :ident]}
+    test "supports aliasing a selected value with selected_as/2" do
+      escaped_alias = {:selected_as, [], [{{:., [], [{:&, [], [0]}, :id]}, [], []}, :ident]}
 
       # single field
-      query = from p in "posts", select: alias(p.id, :ident)
+      query = from p in "posts", select: selected_as(p.id, :ident)
       assert escaped_alias == query.select.expr
 
-      query = select("posts", [p], alias(p.id, :ident))
+      query = select("posts", [p], selected_as(p.id, :ident))
       assert escaped_alias == query.select.expr
 
       # maps
-      query = from p in "posts", select: %{id: alias(p.id, :ident)}
+      query = from p in "posts", select: %{id: selected_as(p.id, :ident)}
       assert {:%{}, [], [id: escaped_alias]} == query.select.expr
 
-      query = select("posts", [p], %{id: alias(p.id, :ident)})
+      query = select("posts", [p], %{id: selected_as(p.id, :ident)})
       assert {:%{}, [], [id: escaped_alias]} == query.select.expr
 
       # structs
-      query = from p in "posts", select: %{p | id: alias(p.id, :ident)}
+      query = from p in "posts", select: %{p | id: selected_as(p.id, :ident)}
       assert {:%{}, [], [{:|, [], [{:&, [], [0]}, [id: escaped_alias]]}]} == query.select.expr
 
-      query = select("posts", [p], %{p | id: alias(p.id, :ident)})
+      query = select("posts", [p], %{p | id: selected_as(p.id, :ident)})
       assert {:%{}, [], [{:|, [], [{:&, [], [0]}, [id: escaped_alias]]}]} == query.select.expr
 
       # keyword lists
-      query = from p in "posts", select: [id: alias(p.id, :ident)]
+      query = from p in "posts", select: [id: selected_as(p.id, :ident)]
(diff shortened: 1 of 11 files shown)


josevalim:
Looks great! Although I wonder if we should introduce “alias/1” to explicitly match the aliased field (and perhaps enforce the aliases to be atoms). This would even allow us to do validations in the future (i.e. the alias exist).

this is also similar to the “as/parent_as” feature, but for selected fields. Maybe we could call it “select_as/selected_as”.

josevalim:
Looks great! Although I wonder if we should introduce “alias/1” to explicitly match the aliased field (and perhaps enforce the aliases to be atoms). This would even allow us to do validations in the future (i.e. the alias exist).

this is also similar to the “as/parent_as” feature, but for selected fields. Maybe we could call it “select_as/selected_as”.

v0idpwn (commented):
Love this! +1 for Jose's suggestion above

greg-rychlewski:
> Although I wonder if we should introduce “alias/1” to explicitly match the aliased field
This is a really good idea, thank you. I've updated to code to implement it.

Are guys ok with calling it `alias/1`? It's no issue for me to change it to `select(ed)_as/1` but I had a couple reasons for preferring `alias/1`:

- I'm a bit worried that anything with `select` in the name will tempt people to assume that structs/maps/kw lists in the select statement are automatically aliased and try to use the keys. There should be less confusion if both functions have the same name.
- There's some kind of symmetry there with `:as` and `as/1` where the "get" and "put" operations both have the same name.

josevalim:
Biggest concern with alias is that there is an `alias` special form in Elixir, so there is potential for confusion there too.

greg-rychlewski:
Good point I didn't think about that. Ok so how about I change it to `selected_as/1` and `selected_as/2`?

josevalim:
> Ok so how about I change it to selected_as/1 and selected_as/2?

I am on the fence. @v0idpwn, any preferences?

greg-rychlewski:
Another point against `alias` is that it can be confused with other things in the Ecto code base. The named bindings for sources are also called aliases internally.

v0idpwn:
Of the presented alternatives, I like the best `selected_as` for both functions. Reasons are:
- `*_as` is closer to the SQL `as` than `alias`. When going through Ecto Query API, I usually expect some symmetry with SQL APIs
- `selected_as` sounds more declarative and `select_as` sounds imperative, I'm inclined towards a more declarative syntax here

> I'm a bit worried that anything with select in the name will tempt people to assume that structs/maps/kw lists in the select statement are automatically aliased and try to use the keys. There should be less confusion if both functions have the same name.
Could you please elaborate with an example? Not sure if I follow.

josevalim:
`selected_as` it is then!

greg-rychlewski:
Beautiful, I can leave the code as is then :) . It's ready for review whenever you guys have time.

@v0idpwn just so I don't leave your question hanging (not trying to make a case for `alias`, I'm convinced now it's the worse option):

> Could you please elaborate with an example? Not sure if I follow.

I just meant that someone might assume `selected_as/`` can work like this:

```elixir
from p in Post, select: %{sum_visits: sum(p.visits)}, order_by: selected_as(:sum_visits)
```

If they assume map (or struct or kw list) keys automatically registers an alias.

test/ecto/query/builder/select_test.exs
```diff
@@ -86,6 +87,46 @@ defmodule Ecto.Query.Builder.SelectTest do
         escape(quote do [:foo, ^:bar] end, [], __ENV__)
       end
     end
+
+    test "supports aliasing a selected value with selected_as/2" do
```
AndrewDryga, line 91: Sorry for a question out of nowhere, but what happens when `selected_as/2` is used two times in the same query and for the same name?

greg-rychlewski: Oh it will raise: https://github.com/elixir-ecto/ecto/pull/3966/files#diff-a4782321bba4a264f78ae4d81bc9be152d20af799129133dd078d59f2c1fb4d0R464.

Actually thank you for asking. I wanted to add a test for this but the error ^ happens before the test starts. I meant to ask for advice testing it.

josevalim (approved):

greg-rychlewski:
@josevalim  / @v0idpwn 

I noticed that the planner emits subquery fields as keyword lists so they can be aliased by the adapters and referenced from the parent: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1007.

I thought I could refactor the `selected_as` planning to be consistent with this. Do you think it's a good idea or would you rather I stick with the way it is? I thought it might be a good idea to not have different ways to do the same thing. These would be the changes to ecto and ecto_sql:

ecto: https://github.com/elixir-ecto/ecto/compare/master...greg-rychlewski:ecto:selected_as_planner
ecto_sql: https://github.com/elixir-ecto/ecto_sql/compare/master...greg-rychlewski:ecto_sql:selected_as_refactor

josevalim:
@greg-rychlewski I think that's a great idea! :+1:

PR #3965, opened 2022-07-29
 Support preloading associations in embedded schemas from the parent schema

Sleepful:
Implements enhancement https://github.com/elixir-ecto/ecto/issues/3890

Sorry for the huge diff, I do not know if my formatter is working properly or not, I tried:

```
mix format --dot-formatter ./.formatter.exs ./lib/ecto/repo/preloader.ex
```

...to no avail.

There is this one bit of code hiding a stacktrace, I left a comment there, confused me a bit.

I am an Elixir newbie so I am mostly unaware of what makes "good" code.

Issue #3890, opened 2022-05-01
Support preloading associations in embedded schemas from the parent schema


nallwhy:
### Current behavior

```elixir
defmodule Child do
  use Ecto.Schema

  schema "children" do
  end
end

defmodule Parent do
  use Ecto.Schema

  schema "parents" do
    embeds_many(:details, Detail, primary_key: false, on_replace: :delete) do
      belongs_to(:child, Child)
    end
  end
end

Parent
|> preload(details: :child)
|> Repo.all()
```

```
** (ArgumentError) schema Parent does not have association :details
...
```

By #3368, `Ecto.Repo.c:preload/3` works from the embedded schema, but not from the parent schema yet.

josevalim:
Good find. We should probably "allow" embeds in preload as long as they are loading an association inside.

josevalim:
A pull request to address this is welcome!

Sleepful:
Working on a pull request for this

wakeless:
Thanks for doing this @Sleepful so nice.

I've been working on this as well and hadn't seen you'd been working along on it. My approach is a little more naive but thought I'd link it here in case there's something useful. https://github.com/elixir-ecto/ecto/compare/master...wakeless:ecto:allow-preloading-within-nested-schemas

greg-rychlewski:
Closed by https://github.com/elixir-ecto/ecto/pull/3967

PR #3967, opened 2022-07-29
Support preloading associations in embedded schemas from the parent schema

Sleepful:
Implements enhancement https://github.com/elixir-ecto/ecto/issues/3890

Re-submission of https://github.com/elixir-ecto/ecto/pull/3965

Might add more tests later

### DIFF
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -739,6 +739,11 @@ defmodule Ecto.Integration.PreloadTest do
       end)
     end)
 
+    # There is no need to preload an embedded schema alone
+    assert_raise ArgumentError,
+      ~r/Cannot preload embedded field/,
+      fn -> TestRepo.preload(order1, :item) end
+
     # Now we preload it
     [order1, order2, order3] = TestRepo.preload(
       # Can preload multiple schemas
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -10,6 +10,7 @@ defmodule Ecto.Integration.PreloadTest do
   alias Ecto.Integration.Permalink
   alias Ecto.Integration.User
   alias Ecto.Integration.Custom
+  alias Ecto.Integration.Order
 
   test "preload with parameter from select_merge" do
     p1 = TestRepo.insert!(%Post{title: "p1"})
@@ -706,6 +707,37 @@ defmodule Ecto.Integration.PreloadTest do
     # Now we preload it
     item = TestRepo.preload(item, :user)
     assert %User{id: ^uid1} = item.user
+  end
+
+  test "preload belongs_to in embedded_schema from container schema" do
+    # Order embeds_many items
+    # Items belong to Users
+    # Preload Users from an Order schema
+
+    %User{id: uid1} = TestRepo.insert!(%User{name: "1"})
+    %User{id: uid2} = TestRepo.insert!(%User{name: "2"})
+    %User{id: uid3} = TestRepo.insert!(%User{name: "3"})
+    item1 = %Item{id: 1, user_id: uid1}
+    item2 = %Item{id: 2, user_id: uid2}
+    item3 = %Item{id: 3, user_id: uid3}
+    order = %Order{items: [item1, item3, item2]}
+    %Order{id: oid1} = TestRepo.insert!(order)
+
+    Enum.map(order.items, fn item ->
+      assert %Ecto.Association.NotLoaded{} = item.user
+    end)
+
+    # Now we preload it
+    order = TestRepo.preload(order, items: :user)
+
+    names =
+      Enum.map(order.items, fn item ->
+        item.user.name
+      end)
+
+    # Different asserts for variety
+    assert %User{id: ^uid1} = hd(order.items).user
+    assert ["1", "3", "2"] = names
   end
 
   defp sort_by_id(values) do
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -753,9 +753,9 @@ defmodule Ecto.Integration.PreloadTest do
       %{order1: order1, order2: order2, order3: order3, order4: order4, order5: order5,
         uid1: uid1}
       = context
-      [order1, order2, order3, order4, order5] = TestRepo.preload(
+      [order1, nil, order2, order3, order4, order5] = TestRepo.preload(
       # Can preload multiple schemas
-      [order1, order2, order3, order4, order5],
+      [order1, nil, order2, order3, order4, order5],
       # Works with multiple preloads
       # Works with embeds_many and embeds_one
       [[items: :user],
@@ -765,13 +765,16 @@ defmodule Ecto.Integration.PreloadTest do
       assert %User{id: ^uid1} = hd(order1.items).user
 
       [names1, names2, names3, names4, names5] =
-        Enum.map([order1, order2, order3, order4, order5], fn order ->
-          case order.items do
-            nil -> nil
-            _   ->
-              Enum.map(order.items, fn item -> item.user.name end)
-          end
-        end)
+        Enum.map([order1, order2, order3, order4, order5],
+          fn
+            %{ :items => nil } -> nil
+            %{ :items => items } ->
+              Enum.map(items,
+                fn
+                  nil -> nil
+                  item -> item.user.name
+                end)
+          end)
       assert ["1", "3", "2"] = names1
       assert ["1", "2"] = names2
       assert [] = names3
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -720,24 +720,52 @@ defmodule Ecto.Integration.PreloadTest do
     item1 = %Item{id: 1, user_id: uid1}
     item2 = %Item{id: 2, user_id: uid2}
     item3 = %Item{id: 3, user_id: uid3}
-    order = %Order{items: [item1, item3, item2]}
-    %Order{id: oid1} = TestRepo.insert!(order)
+    order1 = %Order{items: [item1, item3, item2],
+                    item: item1}
+    order2 = %Order{items: [item1, item2, item3],
+                    item: item2}
+    order3 = %Order{items: [item3, item2, item1],
+                    item: item3}
+    TestRepo.insert!(order1)
+    TestRepo.insert!(order2)
+    TestRepo.insert!(order3)
 
-    Enum.map(order.items, fn item ->
-      assert %Ecto.Association.NotLoaded{} = item.user
+    Enum.map([order1, order2, order3], fn order ->
+      assert %Ecto.Association.NotLoaded{}
+        = order.item.user
+      Enum.map(order1.items, fn item ->
+        assert %Ecto.Association.NotLoaded{}
+          = item.user
+      end)
     end)
 
     # Now we preload it
-    order = TestRepo.preload(order, items: :user)
-
-    names =
-      Enum.map(order.items, fn item ->
-        item.user.name
-      end)
+    [order1, order2, order3] = TestRepo.preload(
+      # Can preload multiple schemas
+      [order1, order2, order3],
+      # Works with multiple preloads
+      # Works with embeds_many and embeds_one
+      [[items: :user],
+      [item: :user]])
 
     # Different asserts for variety
-    assert %User{id: ^uid1} = hd(order.items).user
-    assert ["1", "3", "2"] = names
+    assert %User{id: ^uid1} = hd(order1.items).user
+
+    [names1, names2, names3] =
+      Enum.map([order1, order2, order3], fn order ->
+        Enum.map(order.items, fn item ->
+          item.user.name
+        end)
+    end)
+    assert ["1", "3", "2"] = names1
+    assert ["1", "2", "3"] = names2
+    assert ["3", "2", "1"] = names3
+
+    embeds_one_names =
+      Enum.map([order1, order2, order3], fn order ->
+        order.item.user.name
+    end)
+    assert ["1", "2", "3"] = embeds_one_names
   end
 
   defp sort_by_id(values) do
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -722,9 +722,11 @@ defmodule Ecto.Integration.PreloadTest do
       item3 = %Item{id: 3, user_id: uid3}
       order1 = %Order{items: [item1, item3, item2],
                       item: item1}
-      order2 = %Order{items: [item1, item2, item3],
+      order2 = %Order{items: [item1, item2],
                       item: item2}
       order3 = %Order{items: [item3, item2, item1],
+                      item: item3}
+      order4 = %Order{items: [],
                       item: item3}
       TestRepo.insert!(order1)
       TestRepo.insert!(order2)
@@ -738,8 +740,7 @@ defmodule Ecto.Integration.PreloadTest do
           = item.user
         end)
       end)
-      %{order1: order1, order2: order2, order3: order3,
-        item1: item1, item2: item2, item3: item3,
+      %{order1: order1, order2: order2, order3: order3, order4: order4,
         uid1: uid1}
     end
 
@@ -750,13 +751,12 @@ defmodule Ecto.Integration.PreloadTest do
     end
 
     test "Preloads multiple schemas, with multiple preloads, with embeds_one and embeds_many", context do
-      %{order1: order1, order2: order2, order3: order3,
-        item1: item1, item2: item2, item3: item3,
+      %{order1: order1, order2: order2, order3: order3, order4: order4,
         uid1: uid1}
       = context
-      [order1, order2, order3] = TestRepo.preload(
+      [order1, order2, order3, order4] = TestRepo.preload(
       # Can preload multiple schemas
-      [order1, order2, order3],
+      [order1, order2, order3, order4],
       # Works with multiple preloads
       # Works with embeds_many and embeds_one
       [[items: :user],
@@ -765,21 +765,22 @@ defmodule Ecto.Integration.PreloadTest do
       # Different asserts for variety
       assert %User{id: ^uid1} = hd(order1.items).user
 
-      [names1, names2, names3] =
-        Enum.map([order1, order2, order3], fn order ->
+      [names1, names2, names3, names4] =
+        Enum.map([order1, order2, order3, order4], fn order ->
           Enum.map(order.items, fn item ->
             item.user.name
           end)
         end)
       assert ["1", "3", "2"] = names1
-      assert ["1", "2", "3"] = names2
+      assert ["1", "2"] = names2
       assert ["3", "2", "1"] = names3
+      assert [] = names4
 
       embeds_one_names =
-        Enum.map([order1, order2, order3], fn order ->
+        Enum.map([order1, order2, order3, order4], fn order ->
           order.item.user.name
         end)
-      assert ["1", "2", "3"] = embeds_one_names
+      assert ["1", "2", "3", "3"] = embeds_one_names
     end
 
   end
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -745,7 +745,7 @@ defmodule Ecto.Integration.PreloadTest do
 
     test "There is no need to preload an embedded schema alone", %{order1: order1} do
       assert_raise ArgumentError,
-      ~r/Cannot preload embedded field/,
+      ~r/cannot preload embedded field/,
       fn -> TestRepo.preload(order1, :item) end
     end
 
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -726,10 +726,12 @@ defmodule Ecto.Integration.PreloadTest do
                       item: item2}
       order3 = %Order{items: [],
                       item: nil}
-      order4 = %Order{items: [item3, item2, item1],
+      order4 = %Order{items: nil,
+                      item: nil}
+      order5 = %Order{items: [item3, item2, item1],
                       item: item3}
 
-      Enum.map([order1, order2, order4], fn order ->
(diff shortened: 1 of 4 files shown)


lib/ecto/repo/preloader.ex
```diff
@@ -522,12 +546,25 @@ defmodule Ecto.Repo.Preloader do
...
+
+        case !!embed_result do
+          true ->
+            embed_result
+
+          false ->
+            raise ArgumentError,
+                  "schema #{inspect(schema)} does not have association #{inspect(assoc_result)}#{maybe_module(assoc_result)}"
+        end
+    end
```
josevalim, line 567:
```suggestion
    schema.__schema__(:association, preload) || schema.__schema__(:embed, preload) ||
      raise ArgumentError, "schema #{inspect schema} does not have association #{inspect assoc}#{maybe_module(assoc)}"
```

lib/ecto/embedded.ex
```diff
@@ -283,4 +283,8 @@ defmodule Ecto.Embedded do
   def build(%Embedded{related: related}, _owner) do
     related.__struct__
   end
+
+  def preload_info(_embed) do
+    {:embed}
```
josevalim, line 288:
```suggestion
    :embed
```

lib/ecto/repo/preloader.ex
```diff
@@ -154,6 +155,21 @@ defmodule Ecto.Repo.Preloader do
 
   defp preload_assocs([], [], _repo_name, _tuplet), do: []
 
+  defp preload_embeds(structs, embeds, repo_name, tuplet) do
+    structs =
+      Enum.reduce(structs, [], fn struct, acc ->
```
josevalim, line 160: I think this can be a Enum.map?

lib/ecto/repo/preloader.ex
```diff
@@ -154,6 +155,21 @@ defmodule Ecto.Repo.Preloader do
 
   defp preload_assocs([], [], _repo_name, _tuplet), do: []
 
+  defp preload_embeds(structs, embeds, repo_name, tuplet) do
```
josevalim, line 158:
```suggestion
  defp preload_embeds(structs, [], _repo_name, _tuplet), do: structs

  defp preload_embeds(structs, [_ | _] = embeds, repo_name, tuplet) do
```

integration_test/cases/preload.exs
```diff
@@ -708,6 +709,37 @@ defmodule Ecto.Integration.PreloadTest do
...
+    # Preload Users from an Order schema
+
+    %User{id: uid1} = TestRepo.insert!(%User{name: "1"})
+    %User{id: uid2} = TestRepo.insert!(%User{name: "2"})
+    %User{id: uid3} = TestRepo.insert!(%User{name: "3"})
+    item1 = %Item{id: 1, user_id: uid1}
+    item2 = %Item{id: 2, user_id: uid2}
+    item3 = %Item{id: 3, user_id: uid3}
+    order = %Order{items: [item1, item3, item2]}
+    %Order{id: oid1} = TestRepo.insert!(order)
```
greg-rychlewski, line 724:
```suggestion
    %Order{} = TestRepo.insert!(order)
```
`oid1` is causing a warning because it's not being used anywhere:

>variable "oid1" is unused (if the variable is not meant to be used, prefix it with an underscore)

lib/ecto/repo/preloader.ex
```diff
@@ -493,22 +509,30 @@ defmodule Ecto.Repo.Preloader do
...
             |> Enum.reverse()
             |> Enum.reduce({fields, query, sub_preloads}, &{nil, nil, [{&1, &2}]})
             |> elem(2)
-          expand(schema, through, {assocs, Map.put(throughs, preload, info)})
+
+          expand(schema, through, {assocs, Map.put(throughs, preload, info), embeds})
+
+        {:embed} ->
+          embeds = [{assoc_or_embed, sub_preloads} | embeds]
+          {assocs, throughs, embeds}
```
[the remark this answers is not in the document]
Sleepful: Awesome!

Sleepful:
Added all requested changes.
The test function is getting a little long but it is all closely related so I'm leaving it tightly coupled.

Offtopic: I am listening to the Thinking Elixir Podcast and learning some interesting tidbits from the Elixir Review conversations @josevalim 😁

greg-rychlewski:
I have a suspicion the parallel preloads aren't working as expected, but need more time to look at the code to be sure.

It looks to me like anytime there is an embed it will have its assocs preloaded in isolation from all the other embeds and assocs. And if there are nested embeds then only a single "level" of assocs will be preloaded at once.

I could be wrong though. But that's what it looks like at a glance.

lib/ecto/repo/preloader.ex
```diff
@@ -154,6 +155,28 @@ defmodule Ecto.Repo.Preloader do
 
   defp preload_assocs([], [], _repo_name, _tuplet), do: []
 
+  defp preload_embeds(structs, [], _repo_name, _tuplet), do: structs
+
+  defp preload_embeds(structs, embeds, repo_name, tuplet) do
+      Enum.map(structs, fn struct ->
+          Enum.reduce(embeds, struct, fn {embed, sub_preloads}, struct_acc ->
```
greg-rychlewski, line 162: I think you want to collect all the structs together instead of preloading each struct one by one. That's what it looks like is happening for regular assocs.

integration_test/cases/preload.exs
```diff
@@ -708,6 +709,70 @@ defmodule Ecto.Integration.PreloadTest do
     assert %User{id: ^uid1} = item.user
   end
 
+  test "preload belongs_to in embedded_schema from container schema" do
```
greg-rychlewski, line 712: Since the test is long and also not covering all cases you might want to consider the following:

1. Create a describe block for preloading embeds. Use the setup block to create all the data for the tests.
2. Create separate tests for the following cases:
    i.    parent -> embeds_one -> belongs_to
    ii.   parent -> embeds_many -> has_many
    iii.  parent -> belongs_to -> embeds_one -> belongs_to
    iv.   parent -> has_many -> embeds_many -> has_many
    v.    raising when embed is preloaded without an assoc

These will cover whether assoc can preload embed properly and vice versa.

Sleepful: Regard #2:

a. I don't think it matters whether the preloaded association is `belongs_to` or `has_many` because this PR does not touch on that part of the code. This is only dealing with `embeds_one` and `embeds_many`, the actual association is loaded through the code paths that are already tested.
b. Can we keep the scope of this PR to the case where the parent contains the embed schema and then the association is within the embed? I agree that `parent -> assoc -> embed -> assoc` would be neat, but maybe that can be a separate issue?

lib/ecto/repo/preloader.ex
Sleepful: I applied `Task.async_stream` like `preload_assocs` do, is this what you meant?

integration_test/cases/preload.exs
Sleepful: Took my shot at using the describe block!

Sleepful: Reagrd `#2` again:

I am also a bit new so I don't know what a preload with a `parent -> assoc -> embed -> assoc` would look like, maybe you can help me with the function call for that?

lib/ecto/repo/preloader.ex
```diff
@@ -154,6 +155,37 @@ defmodule Ecto.Repo.Preloader do
...
   defp preload_assocs([], [], _repo_name, _tuplet), do: []
 
+  defp preload_embeds(structs, [], _repo_name, _tuplet), do: structs
+
+  defp preload_embeds(structs, embeds, repo_name, tuplet) do
+    structs
+    |> Task.async_stream(fn struct ->
+      { adapter_meta, opts } = tuplet
+      opts = Keyword.put_new(opts, :caller, self())
+      tuplet = {adapter_meta, opts}
```
Sleepful, line 165: I am cargo-culting here btw, following the same pattern as:

Sleepful: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/preloader.ex#L132

lib/ecto/repo/preloader.ex
Sleepful: btw is there a way to test for this type of detail? It seems difficult to build a test that makes sure something is happening concurrently.

greg-rychlewski: Oh sorry what I meant is that `preload_assocs` calls `preload_each(structs,...)` with the list of relevant structs. But `preload_embeds` is calling `preload_each` one time for each relevant struct.

greg-rychlewski: Not sure if I'm making sense but I'll try to explain the downsides

Both ways you've presented so far (Enum.map and Task.async_stream) will perform a different query for each individual struct instead of batching them all into one query. So there is an inefficiency there in terms of # of queries.

There are some other downsides too. The first way (Enum.map) will perform the queries synchronously, hurting performance. The second way (Task.async_stream) will tie up more DB connections than expected. Ecto advertises that it's one extra connection per assoc but now you have multiple connections for a single assoc.

integration_test/cases/preload.exs
greg-rychlewski: Sorry you made good points. Some of the tests I suggested are pointless. 

1. We don't care about `has_many` because we know it works already when `embed` is the parent. So only using `belongs_to` is ok.
2. `parent -> assoc -> embed -> assoc` isn't necessary because we're calling `preload_each` recursively. So it's no different than testing `parent -> embed`

So I guess just making sure `embeds_one` and `embeds_many` works as expected is enough.

lib/ecto/repo/preloader.ex
Sleepful: @greg-rychlewski oh that is a great explanation, thank you, let me what I can do 👍 
🐈 
⌨️

Sleepful: nice catch btw! 🕵️

josevalim:
@greg-rychlewski is this good to you? :)

greg-rychlewski:
oh not yet sorry. they are still working on batching the `preload_embed` function: https://github.com/elixir-ecto/ecto/pull/3967#discussion_r938151423

Sleepful:
had to leave this on hold for a while, taking more stabs at it every now and then though

Sleepful:
yep this is proving difficult with the given requirements so it will take awhile longer but I think I'm on the right path, the thing is I need to wire the `embeds` into the `prepare_queries` function and it's not as simple as bundling the `embeds` with the other `assocs` because the `embeds` require different arguments `structs, module` when calling `prepare_queries`.

Sorry if that doesn't make sense, it is WIP

greg-rychlewski:
How come `prepare_queries` needs to be changed? I think it should be working already because preloading embeds at the root works.

From what I can tell the only missing part is mimicking this part of `preload_assocs`:

```elixir
all = preload_each(Enum.reverse(loaded_structs, fetch_structs), repo_name, preloads, tuplet)
entry = {:assoc, assoc, assoc_map(assoc.cardinality, Enum.reverse(loaded_ids, fetch_ids), all)}
[entry | preload_assocs(assocs, queries, repo_name, tuplet)]
```

- The first line you will pass the embed structs and it will preload all their assocs.
- The second line you will reattach your embed with the preloads onto the original structs.
- The third line will recursively preload the next embed.

greg-rychlewski:
Something like this:

```elixir
defp preload_embeds(structs, [], _repo_name, _tuplet), do: structs

defp preload_embeds(structs, [embed | embeds], repo_name, tuplet) do
  {%{field: field}, sub_preloads} = embed
  embeds = Enum.map(structs, &Map.get(&1, field)) |> preload_each(repo_name, sub_preloads, tuplet)
  structs = Enum.zip(structs, embeds) |> Enum.map(fn {struct, embed} -> Map.put(struct, field, embed) end)
  preload_embeds(structs, embeds, repo_name, tuplet)
end
```

Sorry not optimized or tested but this is the gist of what I was thinking.

Sleepful:
this line in `fetch_ids` (inside `prepare_queries`) is the one that will not accept an assoc which comes from an embed with the parent (meaning, the struct that contains the embed) struct as `structs`:
```
        %{^owner_key => id, ^field => value} = struct
```

greg-rychlewski:
I might not have explained myself properly. What I meant was that you extract the embeds from the parent structs and then pass just the list of embeds into `preload_each`. Wouldn't that be the same as passing the embeds themselves into `preload_each` even if there was no parent?

Sleepful:
> I might not have explained myself properly.

Don't worry, it's probably on me.

> What I meant was that you extract the embeds from the parent structs and then pass just the list of embeds into preload_each.

Sort of, current implementation is more similar to this:

```
For each relevant struct: Extract embed, call `preload_each` for this embed
```

> Wouldn't that be the same as passing the embeds themselves into preload_each even if there was no parent?

It is exactly that, passing the embeds themselves into `preload_each`, this `preload_each` has no information about the parameters being "embeds".

So, if you don't mind, can we come back to the previous comment again?

You mentioned that current implementation ties up more DB connections than needed:

> Ecto advertises that it's one extra connection per assoc but now you have multiple connections for a single assoc

How does this happen?

Lets say we have `[RelevantStructs]` which embed `EmbeddedStruct` and these embeds contain `EmbeddedAssoc`. Current implementation will call `preload_each` once for each `EmbeddedStructs`, which may be multiple calls because that would be once for each `RelevantStruct`.

Should it instead call `preload_each` once with all of the `[EmbeddedStructs]`?

greg-rychlewski:
> Should it instead call `preload_each` once with all of the `[EmbeddedStructs]`?

Yeah this is exactly what I was trying to propose here: https://github.com/elixir-ecto/ecto/pull/3967#issuecomment-1237619953.

So you call `preload_each` once with all the embeds and then reattach them to the parents with all the assocs preloaded.

Sleepful:
Ah, perfect! Thanks 😁 

I misunderstood at first and thought all embeds had to be loaded on the same connection as the other assocs. Oops!

Sleepful:
Hi! I think I got it (crossing fingers), one thing I did not have to do is deal with any `nil` values which you mentioned here https://github.com/elixir-ecto/ecto/pull/3967#issuecomment-1237619953, so please let me know if I am missing a `nil` edge case.

Other than that I believe this is now calling `prealod_each` *once* per unique `Ecto.Embedded` in `structs`.

Sleepful:
Do you have any tips for debugging code? I have been print-debugging but it feels so clunky, I know iex has a lot of facilities but using iex seems so difficult, mostly because typing in the terminal is really tough (and iex takes away some of my terminal facilities) compared to typing in the editor. I have enjoyed using nREPLs in other environments like clojure, not sure if elixir can do anything similar.

integration_test/cases/preload.exs
```diff
@@ -708,6 +709,82 @@ defmodule Ecto.Integration.PreloadTest do
...
+                      item: item2}
+      order3 = %Order{items: [item3, item2, item1],
+                      item: item3}
+      order4 = %Order{items: [],
+                      item: item3}
+      TestRepo.insert!(order1)
+      TestRepo.insert!(order2)
+      TestRepo.insert!(order3)
+
+      Enum.map([order1, order2, order3], fn order ->
```
greg-rychlewski, line 735:
```suggestion
      order4 = %Order{items: [],
                      item: nil}
```
We don't need to insert the orders because the preload will just take the user ids and put them into these structs, even if the orders are not in the db.

Also we should have a test case where `embed_one` is `nil`.

integration_test/cases/preload.exs
```diff
@@ -708,6 +709,82 @@ defmodule Ecto.Integration.PreloadTest do
...
+      assert ["1", "3", "2"] = names1
+      assert ["1", "2"] = names2
+      assert ["3", "2", "1"] = names3
+      assert [] = names4
+
+      embeds_one_names =
+        Enum.map([order1, order2, order3, order4], fn order ->
+          order.item.user.name
+        end)
+      assert ["1", "2", "3", "3"] = embeds_one_names
```
greg-rychlewski, line 783:
```suggestion
      embeds_one_names =
        Enum.map([order1, order2, order3], fn order ->
          order.item.user.name
        end)
      assert ["1", "2", "3"] = embeds_one_names
      assert is_nil(order4.item)
```

Just need to change this so we test that `nil` is treated properly.

greg-rychlewski:
Thanks @Sleepful this is looking great. I just had 2 smalls suggestions for the unit test and one suggestion to refactor `preload_embeds`. But these should be quick to implement then we can merge :).

>Do you have any tips for debugging elixir? I have been print-debugging every time I run the integration tests but it feels so clunky, I know iex has a lot of facilities (mostly unknown to me) but using iex seems so difficult, mostly because typing in the terminal is really tough compared to typing in the editor; iex even takes away some of my terminal's typing facilities, rlwrap didn't even help me enable arrow keys in iex to navigate my input . I have enjoyed using nREPLs in other environments like clojure, not sure if elixir can do anything similar.
Sorry I'm pretty newb/lazy about tooling and don't really do anything special. But @josevalim might have some more information about this.

lib/ecto/repo/preloader.ex
```diff
@@ -154,6 +155,67 @@ defmodule Ecto.Repo.Preloader do
...
+            # list with quantity of items in embeds :many per owner field
+            track_positions = Enum.map(list_of_lists, &Enum.count(&1))
+            # unwrap the lists
+            flat_schemas = List.flatten(list_of_lists)
+            [{flat_schemas, embed, sub_preloads, track_positions } | embeds_acc ]
+          :one ->
+            list_of_schemas = all_schemas
+            [{list_of_schemas, embed, sub_preloads, :one } | embeds_acc ]
+        end
+    end)
```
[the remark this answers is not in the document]
Sleepful: Oh look at you, rewriting the entire thing in one go and without mistakes. Awesome @greg-rychlewski , How do you do that? This is like 50% less code! 

I will feel a little guilty putting my name on the coming commit.

I like the `foo([head | rest]) do: .... foo(rest)` recursion, much concise. Also the usage of `flat_map_reduce`, very tidy.

integration_test/cases/preload.exs
Sleepful: I tried to search the docs for the possibility of `embed_one` being `nil` but did not find much, I also did not search too much. Adding it to the tests 👍

lib/ecto/repo/preloader.ex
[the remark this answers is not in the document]
greg-rychlewski: > I like the foo([head | rest]) do: .... foo(rest) recursion, much concise. Also the usage of flat_map_reduce, very tidy.

Oh yes those are nice tricks you get used to after some time using this language =).

Btw I updated the suggestion just a bit because you made me realize that `nil` can be passed in the list of structs. I didn't think it was possible before so this new suggestion is more fool proof.

lib/ecto/repo/preloader.ex
```diff
@@ -152,8 +153,40 @@ defmodule Ecto.Repo.Preloader do
...
+    {%{field: field, cardinality: card}, sub_preloads} = embed
+    {embed_structs, counts} =
+      Enum.flat_map_reduce(structs, [], fn struct, counts ->
+        case {Map.get(struct, field), card} do
+          {nil, _card} ->{[], [0 | counts]}
+          {embed_struct, :one} -> {[embed_struct], [1 | counts]}
+          {embed_structs, :many} -> {embed_structs, [length(embed_structs) | counts]}
+        end
+      end)
+    embed_structs = preload_each(embed_structs, repo_name, sub_preloads, tuplet)
```
greg-rychlewski, line 171:
```suggestion
    {%{field: field, cardinality: card}, sub_preloads} = embed
    
    {embed_structs, counts} = 
      Enum.flat_map_reduce(structs, [], fn 
        %{^field => embeds}, counts when is_list(embeds) -> {embeds, [length(embeds) | counts]}
        %{^field => embed}, counts -> {[embed], [1 | counts]}
        nil, counts -> {[], [0 | counts]}
        struct, _counts -> raise ArgumentError, "expected #{inspect(struct)} to contain embed `#{field}`"
      end)
      
    embed_structs = preload_each(embed_structs, repo_name, sub_preloads, tuplet)
```

@Sleepful sorry last change then we can merge :). You discovered that `structs` can have `nil` inside of it so we need to consider it. Also just some minor formatting (adding new lines between single line statements and multiline statements).

lib/ecto/repo/preloader.ex
```diff
@@ -152,8 +153,40 @@ defmodule Ecto.Repo.Preloader do
     [entry | preload_assocs(assocs, queries, repo_name, tuplet)]
   end
 
+
```
greg-rychlewski, line 156:
```suggestion
```
line added here by mistake

lib/ecto/repo/preloader.ex
Sleepful: > You discovered that structs can have nil inside of it so we need to consider it

Oh no, I am not sure about this. Current code already handles the case when `embeds_one` is `nil` and I just added tests where `embeds_many` is `nil` instead of `[]`, which I am not sure if it is possible but it might be so why not handle that case too.

But is this last change necessary?

Sleepful: If you can think of a test case that fails without this suggestion but passes with the suggestion, that would be great for me to understand.

greg-rychlewski: oh sure so in your integration test instead of massing `[order1, order2, order3, order4]` into preload pass `[order1, nil, order2, order3, order4]`.

It will try to do `Map.get(nil, field)` and raise an error

Sleepful: Ditto! Review the updated test case

greg-rychlewski:
Thanks @Sleepful, great work!.

josevalim:
Awesome job @Sleepful with the code and @greg-rychlewski with the reviews! :heart:

PR #3969, opened 2022-07-31
Added a mode to ignore empty strings

qtamaki:
We propose to add an option to avoid errors caused by forcibly rewriting to nil when trying to insert or update an empty string ("") for an attribute that is not null in the database.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1799,12 +1799,13 @@ defmodule Ecto.Changeset do
   def validate_required(%Changeset{} = changeset, fields, opts \\ []) when not is_nil(fields) do
     %{required: required, errors: errors, changes: changes, types: types} = changeset
     trim = Keyword.get(opts, :trim, true)
+    ignoreBlank = Keyword.get(opts, :ignore_blank, false)
     fields = List.wrap(fields)
 
     fields_with_errors =
       for field <- fields,
           ensure_field_not_many!(types, field),
-          missing?(changeset, field, trim),
+          missing?(changeset, field, trim, ignoreBlank),
           ensure_field_exists!(changeset, types, field),
           is_nil(errors[field]),
           do: field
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you for the PR! We already have an empty_values configuration that we use to consider `""` as empty. Is that not enough?

qtamaki:
As for this proposal, if there is a schema in the repository like this, and remark_type and remark_text are not null, the desired behavior is that nil should be caught by validate_required, but blanks should be registered.

```elixir
defmodule Model.Remark do
  use Ecto.Schema
  import Ecto.Changeset

  schema "remarks" do
    field :remark_type, :string
    field :remark_text, :string
  end

  @doc false
  def changeset(model, attrs) do
    model
    |> cast(attrs, [:remark_type, :remark_text], empty_values: [])
    |> validate_required([:remark_type, :remark_text], ignore_blank: false)
  end
end

```

However, in reality, if a space is inserted, it is not distinguished from nil and an error occurs as shown below. (Note that empty_values is applied as a cast/3 argument.)

```
iex(11)> Hello.Repo.get!(Model.Remark, 1) |> Remark.changeset(%{remark_text: "", remark_type: "buy"}) |> Repo.update()
[debug] QUERY OK source="remarks" db=0.2ms queue=0.3ms idle=1826.6ms
SELECT r0.`id`, r0.`remark_type`, r0.`remark_text` FROM `remarks` AS r0 WHERE (r0.`id` = ?) [1]
↳ :erl_eval.do_apply/6, at: erl_eval.erl:685
{:error,
 #Ecto.Changeset<
   action: :update,
   changes: %{},
   errors: [remark_text: {"can't be blank", [validation: :required]}],
   data: #Model.Remark<>,
   valid?: false
 >}
 ```

By passing "ignore_blank: true", blank can be regarded as a normal string and this problem can be avoided.

```elixir
    |> validate_required([:remark_type, :remark_text], ignore_blank: true)
```

```
iex(14)> Hello.Repo.get!(Model.Remark, 1) |> Remark.changeset(%{remark_text: "", remark_type: "buy"}) |> Repo.update()
[debug] QUERY OK source="remarks" db=0.3ms idle=1449.2ms
SELECT r0.`id`, r0.`remark_type`, r0.`remark_text` FROM `remarks` AS r0 WHERE (r0.`id` = ?) [1]
↳ :erl_eval.do_apply/6, at: erl_eval.erl:685
{:ok,
 %Model.Remark{
   __meta__: #Ecto.Schema.Metadata<:loaded, "remarks">,
   id: 1,
   remark_text: "",
   remark_type: "buy"
 }}
```

josevalim:
Thank you, I see.

In this case, the issue is that invalid data has entered the schema and it should not be the responsibility of `validate_required` to validate or ignore data that is already in the schema. You can do a pass in your schema and convert empty to nulls in those cases before casting, but it is not a pattern we should necessarily encourage in Ecto. :)

qtamaki:
Thank you, I see.

I am concerned as a programmer that you cannot distinguish between null and blank ("").

josevalim:
You should be able to distinguish it as a parameter by changing `:empty_values`. In the struct, you can always look at the struct itself. :)

PR #3974, opened 2022-08-05
Add with_named_binding/3

### LINKED ISSUE #3970: Introduce with_named_binding
### Elixir version

1.14

### Database and Version

All

### Ecto Versions

All

All

### Current behavior

Today we often have to write this code when a named binding may be set conditionally:

```elixir
  if has_named_binding?(query, :comments) do
    query
  else
    join(query, :left, c in assoc(p, :comments), as: :comments)
  end
```

Should we provide a small function that encapsulates this pattern?

### Expected behavior

One option is to introduce a `with_named_binding`:

```elixir
def with_named_binding(query, name, fun) do
  if has_named_binding?(query, name) do
    query
  else
    fun.(query)
  end
end
```

Then we can rewrite it as:

```elixir
  with_named_binding(query, :comments, fn ->
    join(query, :left, c in assoc(p, :comments), as: :comments)
  end)
```

Another option is to make it specific to joins and introduce a `with_named_join`:

```elixir
with_named_join(query, ..., as: ...)
```

We would only add the join to the query if the `:as` does not already exist. This is potentially less typing but also more magic.

Thoughts?

AndrewDryga:
Closes #3970

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2300,10 +2300,15 @@ defmodule Ecto.Query do
   def with_named_binding(queryable, key, fun) do
     queryable
     |> Ecto.Queryable.to_query()
-    |> apply_binding_callback(fun, key)
-    |> raise_on_invalid_callback_return(key)
+    |> with_named_binding(key, fun)
   end
-
+  
+  defp apply_binding_callback(query, fun, _key) when is_function(fun, 1), do: query |> fun.() 
+  defp apply_binding_callback(query, fun, key) when is_function(fun, 2), do: query |> fun.(key)
+  defp apply_binding_callback(_query, fun, _key) do
+    raise ArgumentError, "callback function for with_named_binding/3 should accept one or two arguments, got: #{inspect(fun)}"
+  end
+  
   defp raise_on_invalid_callback_return(%Ecto.Query{} = query, key) do
     if has_named_binding?(query, key) do
       query
@@ -2314,12 +2319,6 @@ defmodule Ecto.Query do
 
   defp raise_on_invalid_callback_return(other, _key) do
     raise RuntimeError, "callback function for with_named_binding/3 should return an Ecto.Query struct, got: #{inspect(other)}"
-  end
-  
-  defp apply_binding_callback(query, fun, _key) when is_function(fun, 1), do: fun.(query)
-  defp apply_binding_callback(query, fun, key) when is_function(fun, 2), do: fun.(query, key)
-  defp apply_binding_callback(_query, fun, _key) do
-    raise ArgumentError, "callback function for with_named_binding/3 should accept one or two arguments, got: #{inspect(fun)}"
   end
 
   @doc """
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2263,26 +2263,36 @@ defmodule Ecto.Query do
   end
 
   @doc """
-  Applies the callback function to the query if it does not have a named binding,
-  otherwise returns the query without changes.
+  Applies a callback function to a query if it doesn't contain the given named binding. 
+  Otherwise, returns the original query.
 
-  The function accepts an `Ecto.Query` struct and must return the query with an added 
-  named binding.
+  The callback function must accept a queryable and return an `Ecto.Query` struct 
+  that contains the provided named binding, otherwise an error is raised. It can also 
+  accept second argument which is the atom representing the name of a binding.
+  
+  For example, one might use this function as a convenience to conditionally add a new 
+  named join to a query:
 
-  ## Examples
-
-      with_named_binding(query, :comments, fn query ->
+      if has_named_binding?(query, :comments) do
+        query
+      else
         join(query, :left, c in assoc(p, :comments), as: :comments)
-      end)
+      end
 
-  For more information on named bindings see "Named bindings" in this module doc.
+  With this function it can be simplified to:
+
+      with_named_binding(query, :comments, fn  query, binding ->
+        join(query, :left, a in assoc(p, ^binding), as: ^binding)
+      end) 
+
+  For more information on named bindings see "Named bindings" in this module doc or `has_named_binding/2`. 
   """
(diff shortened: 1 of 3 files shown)


lib/ecto/query.ex
```diff
@@ -2262,6 +2262,49 @@ defmodule Ecto.Query do
     has_named_binding?(Ecto.Queryable.to_query(queryable), key)
   end
 
+  @doc """
+  Applies the callback function to the query if it does not have a named binding,
+  otherwise returns the query without changes.
+
+  The function accepts an `Ecto.Query` struct and must return the query with an added 
+  named binding.
```
AndrewDryga, line 2270: I would appreciate it if you have ideas for better wording here and in raised errors.

lib/ecto/query.ex
```diff
@@ -2262,6 +2262,49 @@ defmodule Ecto.Query do
...
+
+  For more information on named bindings see "Named bindings" in this module doc.
+  """
+  def with_named_binding(%Ecto.Query{} = query, key, fun) when is_function(fun, 1) do
+    if has_named_binding?(query, key) do
+      query
+    else
+      query
+      |> fun.()
+      |> raise_on_invalid_callback_return(key)
```
AndrewDryga, line 2286: I made it defensive to protect people from shooting the foot (eg. debugging for hours if you have a typo in the name of a binding and call the function multiple times).

lib/ecto/query.ex
```diff
@@ -2262,6 +2262,49 @@ defmodule Ecto.Query do
...
+    end
+  end
+
+  def with_named_binding(queryable, key, fun) do
```
greg-rychlewski, line 2290: Please keep me honest, but if it's not a query already then there shouldn't be any uncertainty whether the binding exists and this function wouldn't be needed?

AndrewDryga: There is one edge case why I left it as is: you can implement `Ecto.Queryable` protocol for any struct, protocol implementation can do pretty much anything it wants, including creating aliases. So it's a not safe to assume that `to_query/1` call will always return an `Ecto.Query` without aliases. (However, I agree it's highly unlikely as I have never seen such use cases.)

Plus the code in `has_named_binding?/3` can change and I feel generally safer to rely on it rather than duplicating the logic.

lib/ecto/query.ex
greg-rychlewski: FWIW I like what you did. It gives some more focus to the function. It didn't really feel right to me that the anon function could do whatever it wanted.

lib/ecto/query.ex
greg-rychlewski: How about this:

```
Applies a callback function to a query if it doesn't contain the given named binding. Otherwise, returns the original query.

The callback function must return an `Ecto.Query` struct that contains the provided named binding, otherwise an error is raised.

This is a convenience function that can be used to simplify conditional modifications of a query. For example, one might write the following code to conditionally add a new named join to a query:

    if has_named_binding?(query, :comments) do
      query
    else
      join(query, :left, c in assoc(p, :comments), as: :comments)
    end

With this function it can be simplified to:

    with_named_binding(query, :comments, fn ->
      join(query, :left, c in assoc(p, :comments), as: :comments)
    end) 

For more information on named bindings see "Named bindings" in this module doc or `has_named_binding/2`. 
``

File: lib/ecto/query.ex [hunk 4]
```diff
@@ -2262,6 +2262,49 @@ defmodule Ecto.Query do
...
+      end)
+
+  For more information on named bindings see "Named bindings" in this module doc.
+  """
+  def with_named_binding(%Ecto.Query{} = query, key, fun) when is_function(fun, 1) do
+    if has_named_binding?(query, key) do
+      query
+    else
+      query
+      |> fun.()
```
greg-rychlewski, line 2285:
```suggestion
  def with_named_binding(%Ecto.Query{} = query, key, fun) when is_function(fun, 2) do
    if has_named_binding?(query, key) do
      query
    else
      fun.(query, key)
```

This was one of @v0idpwn's suggestion in the proposal discussion. It's a really good idea because then you can create reusable callbacks.

greg-rychlewski:
@AndrewDryga It looks good to me. Just had 2 small comments.

lib/ecto/query.ex
josevalim: We can also check the arity and support both.

josevalim:
@greg-rychlewski @v0idpwn given the discussion, do you think it is still worth having this feature?

greg-rychlewski:
It seems worth it to me since a few people spoke up about using the same pattern. 

I think even if something [like this](https://github.com/elixir-ecto/ecto/issues/3970#issuecomment-1208032723) is implemented it can co-exist with the function in this PR, since this one allows you to arbitrarily group any number of actions under the same condition.

v0idpwn:
As Greg, I've been convinced that it is useful based on the comments. I think named bindings don't get enough attention and providing a smarter APIs around them is a good way to get the users to use them more (and better) where adequate.

lib/ecto/query.ex
v0idpwn: I didn't comment in the issue but I got the inspiration from `validate_change`, whose typespec is:
```elixir
@spec validate_change(
  t(), 
  atom(), 
  (atom(), term() -> [{atom(), String.t()} | {atom(), {String.t(), Keyword.t()}}])
) :: t()
```

lib/ecto/query.ex
AndrewDryga: @greg-rychlewski I like your examples, thank you. I'll add some tests to make sure such bindings are supported (did not find a test for that for `as:` and that they can be inspected).

AndrewDryga:
Thank you guys ❤️. I applied all suggestions and went with a path of supporting both functions with arity 1 or 2 depending if you need the name of a binding or not.

josevalim (approved):
Someone else from the team please merge if you also approve. :)

greg-rychlewski:
thanks @AndrewDryga, merging now

PR #3833, opened 2022-02-08
Supporting parent_as in type/2 calls.

wilkinson4:
See the following [conversation](https://groups.google.com/g/elixir-ecto/c/VYV5D-1gHBk).

PR #3976, opened 2022-08-09
Allow parent_as/1 in type/2

greg-rychlewski:
Motivation: https://github.com/elixir-ecto/ecto/pull/3833

### DIFF
--- a/integration_test/cases/type.exs
+++ b/integration_test/cases/type.exs
@@ -1,7 +1,7 @@
 defmodule Ecto.Integration.TypeTest do
   use Ecto.Integration.Case, async: Application.compile_env(:ecto, :async_integration_tests, true)
 
-  alias Ecto.Integration.{Custom, Item, ItemColor, Order, Post, User, Tag, Usec}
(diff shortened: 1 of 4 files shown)


integration_test/cases/type.exs
```diff
@@ -141,6 +142,11 @@ defmodule Ecto.Integration.TypeTest do
     # Comparison expression
     assert [12] = TestRepo.all(from p in Post, select: type(coalesce(p.visits, 0), :integer))
     assert [1.0] = TestRepo.all(from p in Post, select: type(coalesce(p.intensity, 1.0), :float))
+
+    # parent_as/1
+    child = from c in Comment, where: type(parent_as(:posts).id, :string) == c.text, select: c.post_id
+    query = from p in Post, as: :posts, where: p.id in subquery(child), select: p.id
+    assert [post_id] == TestRepo.all(query)
```
josevalim, line 149: I don't think we need the integration test in this case as the restriction is fully on Ecto.Query side, but since you wrote it already, we can go ahead and commit it. :)

josevalim (approved):

PR #3978, opened 2022-08-10
Refactor selected_as planning

greg-rychlewski:
This will turn the `selected_as` field into a 2-tuple so that it can be aliased the same way subquery fields are aliased. Subquery reference: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1007

Will update this with the corresponding ecto_sql PR

ecto_sql PR: https://github.com/elixir-ecto/ecto_sql/pull/434

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -404,7 +404,7 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:over, [], [aggregate, window]]}, params_acc}
   end
 
-  def escape({:selected_as, _, [_expr, name]}, _type, _params_acc, _vars, _env) do
+  def escape({:selected_as, _, [_expr, _name]}, _type, _params_acc, _vars, _env) do
     error! """
     selected_as/2 cannot be used unless it is at the root of a select statement. \
     If you are trying to use it inside of an expression, consider putting the \
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -406,7 +406,7 @@ defmodule Ecto.Query.Builder do
 
   def escape({:selected_as, _, [_expr, _name]}, _type, _params_acc, _vars, _env) do
     error! """
-    selected_as/2 cannot be used unless it is at the root of a select statement. \
+    selected_as/2 can only be used at the root of a select statement. \
     If you are trying to use it inside of an expression, consider putting the \
     expression inside of `selected_as/2` instead. For instance, instead of:
 
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -404,6 +404,20 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:over, [], [aggregate, window]]}, params_acc}
   end
 
+  def escape({:selected_as, _, [_expr, name]}, _type, _params_acc, _vars, _env) do
+    error! """
+    selected_as/2 cannot be used unless it is at the root of a select statement. \
+    If you are trying to use it inside of an expression, consider putting the \
+    expression inside of `selected_as/2` instead. For instance, instead of:
+
+        from p in Post, select: coalesce(selected_as(p.visits, :v), 0)
+
+    use:
+
+        from p in Post, select: selected_as(coalesce(p.visits, 0), :v)
+    """
+  end
+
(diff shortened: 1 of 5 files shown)


josevalim:
Perhaps we should now raise in Ecto  if selected_as is not at the root? This would require to change the implementation slightly (i.e. check for selected_as before we call collect_fields).

greg-rychlewski:
I'd like to make that change too. Actually now I realize there are 2 situations:

1. `selected_as` is nested in another function i.e. `SELECT coalesce(x as alias, 0) FROM table`
2. `selected_as` is at the root of a subquery i.e. `SELECT * from (SELECT x as y FROM table)`

The first one is a problem because it's illegal. The second is a problem because Ecto is already automatically aliasing subquery fields. So you'll get a double `AS` and it will be illegal (at least sometimes, postgres doesn't like it for example).

For 1, I'm not entirely sure what's the idiomatic way to handle it. My inclination is to have a boolean `root` variable being passed to `collect_fields` that is true on the first call, false for the nested calls. But is there a better way?

For 2, it seems like `:never` is passed into `collect_fields` for the `from` argument. I could check for that and raise if `selected_as` is being used alongside it.

josevalim:
> For 1, I'm not entirely sure what's the idiomatic way to handle it. My inclination is to have a boolean root variable being passed to collect_fields that is true on the first call, false for the nested calls. But is there a better way?

There are already too many vars on `collect_fields`. Perhaps the best option is to:

a. do a pass on the collected fields (after collect_fields), converting them to tuples
b. keep the raise on ecto_sql for strays selected_as for better UX

Then solving 2 is easy, you never perform step a. above.

greg-rychlewski:
Thanks that helped clean it up a lot. I'm going to leave some comments where I'm not sure if it's clear why I did something.

lib/ecto/query/builder.ex
```diff
@@ -404,6 +404,20 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:over, [], [aggregate, window]]}, params_acc}
   end
 
+  def escape({:selected_as, _, [_expr, _name]}, _type, _params_acc, _vars, _env) do
```
greg-rychlewski, line 407: I forgot that escaping `selected_as/2` inside the select builder enforces it being at the root, because the only other escapes in there are for maps/tuples/structs, etc... everything else goes into this generic builder. So I put a nicer error message here to catch the non-root selects.

But I also formatted selected_as in the planner like you said (after collect_fields) to pick off stragglers (and also help with subqueries).

lib/ecto/query/planner.ex
```diff
@@ -1287,11 +1287,11 @@ defmodule Ecto.Query.Planner do
     end
   end
 
-  defp normalize_select(%{select: nil} = query, _keep_literals?) do
+  defp normalize_select(%{select: nil} = query, _keep_literals?, _allow_alias?) do
     {query, nil}
   end
 
-  defp normalize_select(query, keep_literals?) do
+  defp normalize_select(query, keep_literals?, allow_alias?) do
```
greg-rychlewski, line 1294: I put the third parameter because it seemed cleaner to me than doing it outside `normalize_select`. But let me know what you think.

test/ecto/query/builder/select_test.exs
```diff
@@ -120,6 +120,13 @@ defmodule Ecto.Query.Builder.SelectTest do
       assert [{:{}, [], [:id, escaped_alias]}] == query.select.expr
     end
 
+    test "supports aliasing a selected value in select_merge with selected_as/2" do
```
greg-rychlewski, line 123: I realized I didn't test select_merge before and added it in instead of opening another PR

lib/ecto/query/builder.ex
```diff
@@ -404,6 +404,20 @@ defmodule Ecto.Query.Builder do
     {{:{}, [], [:over, [], [aggregate, window]]}, params_acc}
   end
 
+  def escape({:selected_as, _, [_expr, name]}, _type, _params_acc, _vars, _env) do
+    error! """
+    selected_as/2 cannot be used unless it is at the root of a select statement. \
```
josevalim, line 409:
```suggestion
    selected_as/2 can only be used at the root of a select statement. \
```

josevalim (approved):

PR #3979, opened 2022-08-15
Use `local_now` by default

youxingz:
In different locations, the macro `timestamps` does not work very well because of the timezone, and it causes a lot of problems for beginners. So we use the new api `local_now` by default.


josevalim:
Ecto is meant for the times to be stored in UTC in the database. Then, if you need to convert them, for example to format them in the UI, you do so in the presentation layer (in your views, templates, etc), while keeping the database consistent.

Issue #3981, opened 2022-08-17
Inconsistent behavior with Ecto.Enum and jsonb


maxsalven:
### Elixir version

1.13.4

### Database and Version

Postgres 14.4

### Ecto Versions

ecto 3.8.4

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.16.4

### Current behavior

Migration:
```elixir
  def change do
    create table(:users) do
      add(:status, :string)
      add(:preferences, :jsonb)
    end
  end
```

Schema:
```elixir
defmodule MyApp.Users.User do
  use MyApp.ModelBase
  require Ecto.Changeset

  schema "users" do
    field :status, Ecto.Enum,
      values: [active: "ACTIVE", inactive: "INACTIVE"],
      default: :active

    embeds_one :preferences, Preferences, primary_key: false, on_replace: :update do
      field :theme, Ecto.Enum, values: [light: "LIGHT", dark: "DARK"]
    end
  end

  def changeset(user, attrs) do
    user
    |> cast(attrs, [:status])
    |> cast_embed(:preferences, with: &preferences_changeset/2)
  end

  defp preferences_changeset(schema, params) do
    schema
    |> cast(params, [:theme])
  end
end
```

`iex -S mix`
```elixir
MyApp.Repo.insert(MyApp.Users.User.changeset(%MyApp.Users.User{}, %{status: :active, preferences: %{theme: :light}}))
```

Record created in DB:
`1 | ACTIVE | {"theme": "light"}`

### Expected behavior

I expect the JSONB value to be capitalized:
`1 | ACTIVE | {"theme": "LIGHT"}`

maxsalven:
I've also just noticed that if you use `values: [light: 1, dark: 2]`, then it still stores `"light"` in the JSONB object

josevalim:
We have an option for custom types that controls if the type is stored as the loaded or dumped value in JSONB. We should probably make this option customizable for enum types.

josevalim:
You would like to send a PR that makes the return of the `embed_as` function in `Ecto.Enum` customizable? I think changing that is what you need. :)

PR #3982, opened 2022-08-19
Support dumping when embed in Ecto.Enum

v0idpwn:
Adds a new option to the Ecto.Enum parameterized type.
Closes https://github.com/elixir-ecto/ecto/issues/3981

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -49,13 +49,16 @@ defmodule Ecto.Enum do
         end
       end
 
-  you can call `mappings/2` like this:
+  You can call `mappings/2` like this:
 
       Ecto.Enum.mappings(EnumSchema, :my_enum)
       #=> [foo: "foo", bar: "bar", baz: "baz"]
 
   If you want the values only, you can use `Ecto.Enum.values/2`, and if you want
   the dump values only, you can use `Ecto.Enum.dump_values/2`.
+
+  When using Ecto.Enum within embeds, it is possible to chose whether to dump them
+  or not with the `:dump_embed?` option, which defaults to `false`.
   """
 
   use Ecto.ParameterizedType
@@ -96,8 +99,16 @@ defmodule Ecto.Enum do
     on_load = Map.new(mappings, fn {key, val} -> {val, key} end)
     on_dump = Map.new(mappings)
     on_cast = Map.new(mappings, fn {key, _} -> {Atom.to_string(key), key} end)
+    embed_as = if opts[:dump_embed?], do: :dump, else: :self
 
-    %{on_load: on_load, on_dump: on_dump, on_cast: on_cast, mappings: mappings, type: type}
+    %{
+      on_load: on_load,
+      on_dump: on_dump,
+      on_cast: on_cast,
+      mappings: mappings,
+      embed_as: embed_as,
+      type: type
+    }
   end
 
   defp validate_unique!(values) do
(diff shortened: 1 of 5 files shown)


test/ecto/type_test.exs
```diff
@@ -230,19 +230,18 @@ defmodule Ecto.TypeTest do
...
-    @uuid_binary <<191, 224, 136, 140, 92, 89, 75, 179, 173, 253, 113, 240, 184, 93, 61, 183>>
 
     test "one" do
       embed = %Ecto.Embedded{field: :embed, cardinality: :one,
                              owner: __MODULE__, related: Schema}
       type  = {:parameterized, Ecto.Embedded, embed}
 
       assert {:ok, %Schema{id: @uuid_string, a: 1, c: 0}} =
-             adapter_load(Ecto.TestAdapter, type, %{"id" => @uuid_binary, "abc" => 1})
+             adapter_load(Ecto.TestAdapter, type, %{"id" => @uuid_string, "abc" => 1})
```
v0idpwn, line 240: These tests were faulty: they didn't respect Ecto.UUID `embed_as` (self) because `TestRepo` didn't respect `embed_as`

greg-rychlewski:
Would it be a disaster to switch `embed_as` to `:dump` for `Ecto.Enum` and consider it a bug fix?

lib/ecto/enum.ex
```diff
@@ -96,8 +99,16 @@ defmodule Ecto.Enum do
     on_load = Map.new(mappings, fn {key, val} -> {val, key} end)
     on_dump = Map.new(mappings)
     on_cast = Map.new(mappings, fn {key, _} -> {Atom.to_string(key), key} end)
-
-    %{on_load: on_load, on_dump: on_dump, on_cast: on_cast, mappings: mappings, type: type}
+    embed_as = if opts[:dump_embed?], do: :dump, else: :self
```
josevalim, line 102: What if we name the option `embed_as: :enum | :values`?

josevalim:
> Would it be a disaster to switch embed_as to :dump for Ecto.Enum and consider it a bug fix?

I think we would break all enums currently stored in embeds because we would not be able to load them :(

lib/ecto/enum.ex
v0idpwn: I agree its better to have names than the boolean interface, but I would suggest other names that make what happens more obvious (specially since `Ecto.Enum` receive `:values` which are the "enumerated options")

Perhaps `embed_as: :values | :dumped_values`?

josevalim: Oh, that's what I meant to say but I got confused. 😅  Maybe `:values | :dumped`? Or maybe just stick with `:self | :dump`?

v0idpwn: I like `:values | :dumped`, lets go with it :)

v0idpwn: Pushed a commit changing this

v0idpwn:
>I think we would break all enums currently stored in embeds because we would not be able to load them :(
I did some tests for that and apparently loading works. I'll try to figure out what is happening there later. But even under this assumption, it could be considered a breaking change since people may rely on how data is represented within the database, specially on jsonb fields and such.

josevalim:
We can detect if we are in an embedded schema and if so, we ask users to explicitly set the `:embed_as` option. We will tell in the message that, if the schema already exists, they must set it to `embed_as: :values` for backwards compatibility, otherwise it must be `:dumped`?

v0idpwn:
I will come back to this soon. I'll try to bring more data/tests so that we have specified behaviour for all the cases.

josevalim:
@v0idpwn I think we are ready for v3.9 but we probably want to get this in. WDYT? :)

v0idpwn:
@josevalim I think we can put it in. I think I can finish it in the weekend or are we in a hurry?

josevalim:
Not in a hurry, take your time!

josevalim:

josevalim:
I will merge and do the final touch ups locally. :)

PR #3996, opened 2022-09-01
Add minimal example for `Multi.one/2` and `Multi.all/2` using function

RodolfoSilva:
I think this kind of example should help some people.

When using `one/2` and `all/2` I was wondering why this example was missing from the method documentation as it was supported.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -406,6 +406,9 @@ defmodule Ecto.Multi do
 
       Ecto.Multi.new()
       |> Ecto.Multi.one(:post, Post)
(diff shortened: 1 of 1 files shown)


lib/ecto/multi.ex
```diff
@@ -406,6 +406,9 @@ defmodule Ecto.Multi do
 
       Ecto.Multi.new()
       |> Ecto.Multi.one(:post, Post)
+      |> Ecto.Multi.one(:author, fn %{ post: post } ->
```
josevalim, line 409:
```suggestion
      |> Ecto.Multi.one(:author, fn %{post: post} -> 
```

josevalim:

PR #3998, opened 2022-09-08
fragments as source

greg-rychlewski:
I saw a [thread on elixir forum](https://elixirforum.com/t/ecto-fragment-as-a-query-source/33030) where some people wanted to use a fragment as a query source. I thought it wouldn't be too terrible to implement that so created this PR.

Companion PR in ecto_sql: https://github.com/elixir-ecto/ecto_sql/pull/438

edit:
Actually now I'm seeing a problem. It only works nicely for queries using `from` but for the other ones it's a bit messy. It would require escaping the fragment  in all the other macros like `select`, `where`, etc. So I guess there are 3 options:

1. Only have it for `from` queries. This has the potential to create confusion when people notice it works sometimes and not others.
2. Handle fragments specially for all the other macros. This is potentially not worth the effort.
3. Don't do this change.

I was previously leaning towards 1 or 3 but 2 is not looking as terrible as I thought. I'll wait to hear other opinions:).


lib/ecto/query/planner.ex
```diff
@@ -1956,10 +1957,10 @@ defmodule Ecto.Query.Planner do
   end
 
   defp error!(query, expr, message) do
-    raise Ecto.QueryError, message: message, query: query, file: expr.file, line: expr.line
+    raise Ecto.QueryError, message: message, query: query, file: Map.get(expr, :file), line: Map.get(expr, :line)
   end
 
   defp error!(query, expr, message, hint) do
-    raise Ecto.QueryError, message: message, query: query, file: expr.file, line: expr.line, hint: hint
+    raise Ecto.QueryError, message: message, query: query, file: Map.get(expr, :file), line: Map.get(expr, :line), hint: hint
```
josevalim, line 1964: Should we add `file` and `line` fields to `From`?

josevalim:
This looks great! Very nicely done. :)

> it only works nicely for queries using from but for the other ones it's a bit messy. It would require escaping the fragment in all the other macros like select, where, etc.

Apologies but I don't get the issue. :) Can you please explain it further?

lib/ecto/query/planner.ex
greg-rychlewski: Good idea, that would be a lot nicer than this. Will work on it.

greg-rychlewski:
> Apologies but I don't get the issue. :) Can you please explain it further?

Oh yeah sorry. So if you do stuff like this it will complain that `fragment/1` is not a function because I only escaped it in `from`:

```elixir
fragment("select 1 as x") |> select([f], f.x)
fragment("select 1 as x") |> where([f], f.x == 1)
```

So it's a bit different behaviour from the other queryables where you can do this:

```elixir
Comment |> select([c], c.text)
Comment |> wheret([c], c.text == "abc")
```
At first I thought maybe it's annoying to remember to treat fragments differently than the other queryables in all the macros. But now I'm looking at it and the change would be pretty small. I would just have to create a reusable function that would convert `{:fragment, meta, args}` to `{{:{}, [], [:fragment, meta, args]}, params}` and then leave the other queryables as they are.

lib/ecto/query/planner.ex
greg-rychlewski: I made this change but right now the file/line are only being set in the `from` macro. All the other macros that call `to_query` (`select`, `filter`, etc...) won't set the file/line. I was wondering if you mind me doing that in a separate PR because it's a bit involved and would be nice to tackle it separately.

Then actually there are 2 other things I want to tackle in another PR too:

1. A nice error message when trying to preload from a fragment source.
2. A nice error message when trying to use a fragment source in a subquery. This fails because subqueries have to know which fields are in the source.

greg-rychlewski:
> At first I thought maybe it's annoying to treat fragments differently than the other queryables in all the macros. But now I'm looking at it and the change would be pretty small. 

Adding this change, it wasn't as bad as I originally thought. Ready for re-review when you have time :)

josevalim:
> Added this change, it wasn't as bad as I originally thought. Ready for re-review when you have time :)

Maybe `Ecto.Query.fragment` could be a macro that already returns an Ecto.Query struct? This way you don't need stuff like `escape_queryable` and you don't need to change the protocol implementation. Plus we get file+line.

greg-rychlewski:
> > Added this change, it wasn't as bad as I originally thought. Ready for re-review when you have time :)
> 
> Maybe `Ecto.Query.fragment` could be a macro that already returns an Ecto.Query struct? This way you don't need stuff like `escape_queryable` and you don't need to change the protocol implementation. Plus we get file+line.

@josevalim Thanks that was much better returning an Ecto.Query struct from fragment. I'm not sure fragment can be turned into a macro because it has variable arity. I think an extra function is still needed to escape it/turn it into Ecto.Query.  Could you let me know if I'm overlooking something?

josevalim:
@greg-rychlewski gah, the variable arity. So I think we should rather say you can only use fragment when specifying `from(p in fragment(...))`. I.e. it only exists as part of the query syntax?

greg-rychlewski:
Good call it's much cleaner this way. Ready for review when you have time :).

One thing still missing is setting file/line in the `FromExpr` struct when the `from` macro isn't used. i.e. `select(City, [c], c)` will generate a query where the `FromExpr` has `nil` file/line. Would it be alright if I follow up with that in the next PR?

josevalim (approved):

PR #4000, opened 2022-09-13
Compile time CTE fix.

greg-rychlewski:
When building a CTE  the `apply` function adds at a `%Ecto.Query.WithExpr{}` struct for the first CTE. This causes an issue for compile time queries because that struct is not escaped afterwards.

Since `apply` doesn't know whether it's called at runtime or compile time I think the right way to fix this is to escape any added structs when re-escaping the compile time query. But please let me know if I'm missing a better way to do it.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -1292,7 +1292,15 @@ defmodule Ecto.Query.Builder do
 
   # Escapes an `Ecto.Query` and associated structs.
   @spec escape_query(Query.t) :: Macro.t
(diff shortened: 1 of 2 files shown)


josevalim:
@greg-rychlewski if we do so, now we need to unescape the structs as well, and that may make it more expensive.

I think this is a design issue in `WithExpr`: we can only assume that the _query_ is known at compile-time, nothing more. So we need to check if we are doing the work at compile or runtime and act accordingly. I will push a fix. Good find!

Issue #3839, opened 2022-02-24
Select fields do not match when using a union in a preload


MarcusHoile:
### Environment

* Elixir version: 1.13.0
* Database and version: PostgreSQL 12.3
* Ecto version: 3.7.1
* Database adapter and version: postgrex 0.16.2
* Operating system: MacOS 10.15.7

### Current behavior

```elixir
defmodule MyApp do
  def query do
    from(u in User, preload: [posts: ^user_posts()])
    |> Repo.all()
  end

  def user_posts do
    posts_with_comments()
    |> union(^posts_with_tags())
  end

  def posts_with_comments do
    from(p in Post, join: c in assoc(p, :comments))
  end

  def posts_with_tags do
    from(p in Post, join: t in assoc(p, :tags))
  end
end
```

`MyApp.query()` will raise an error as the select fields do not match. The left query, `post_with_comments`, has posts.user_id selected twice and the right query, `post_with_tags` has the select fields you would expect.

```
** (Postgrex.Error) ERROR 42601 (syntax_error) each UNION query must have the same number of columns

         query: SELECT p0."id", p0."title", p0."user_id", p0."inserted_at", p0."updated_at", p0."user_id" FROM "posts" AS p0 INNER JOIN "comments" AS c1 ON c1."post_id" = p0."id" WHERE (p0."user_id" = ANY($1)) UNION (SELECT p0."id", p0."title", p0."user_id", p0."inserted_at", p0."updated_at" FROM "posts" AS p0 INNER JOIN "tags" AS t1 ON t1."post_id" = p0."id") ORDER BY p0."user_id"
```

I have an example app with failing test that shows the issue https://github.com/MarcusHoile/union_query_select_fields

### Expected behavior

I'd expect the select fields to match by default

josevalim:
This is expected. It is your responsibility to make sure both sides of the union select the same fields, have the same preloads, and so on. :)

MarcusHoile:
no worries. So my workaround to the example above is to match the dupe field selected which doesn't sound so crazy an idea now

```
def posts_with_tags do
  fields = Post.__schema__(:fields__) ++ [:user_id]
  from(p in Post, join: t in assoc(p, :tags), select: map(p, ^fields))

josevalim:
I think you should use preloads on both sides, especially because we can't really preload only one side of the union.

MarcusHoile:
I might be missing something, im not preloading on either side of the union. The union query is used as the query to preload posts

[...]

PR #4001, opened 2022-09-15
Preload with union query

pedro-gutierrez:
I am raising this PR to illustrate a scenario that involves preloading and union or intersect queries that I can't quite figure out. 

The CI will fail with the error below.
 
In this PR, I've got two integration tests.  Each test builds the same query on `comments` which is the union of two other queries:

```elixir
q1 = from c in Comment, where: c.text == "a"
q2 = from c in Comment, where: c.text == "b"
query = union(q1, ^q2)
```

When I use this query to fetch all comments that match, using `Repo.all(query)`, I get an empty list, as expected. This means the query was properly accepted by the database and did execute successfully. 

However, if I use the same query to preload a specific post with some comments, using `Repo.preload([p1], [{:comments, query}], [])`, then I get the error:

```
** (Postgrex.Error) ERROR 42601 (syntax_error) each UNION query must have the same number of columns
``` 

and if I take a look at the generated sql: 

```
query: SELECT c0."id", c0."text", c0."lock_version", c0."post_id", c0."author_id", c0."post_id" FROM "comments" AS c0 WHERE (c0."text" = 'a') AND (c0."post_id" = $1) UNION (SELECT c0."id", c0."text", c0."lock_version", c0."post_id", c0."author_id" FROM "comments" AS c0 WHERE (c0."text" = 'b')) ORDER BY c0."post_id"
```

I can see indeed there is a difference in the number of columns selected by the two queries. The first query is selecting `c0.post_id` twice (not expected?), while the second is selecting it just once (expected).

The same happens with intersect queries. It is highly possible that I am abusing the preloader, or maybe this is expected, but I was curious to know why this behaviour?  

And most importantly, is there a way I could workaround it ? :)

Thanks in advance!

### DIFF
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -778,7 +778,7 @@ defmodule Ecto.Integration.PreloadTest do
       q2 = from c in Comment, select: map(c, ^fields), where: c.text == "b"
       query = union(q1, ^q2)
 
-      p1 = TestRepo.preload([p1], [{:comments, query}], [])
+      [p1] = TestRepo.preload([p1], [{:comments, query}], [])
 
       assert [] == p1.comments
     end
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -768,6 +768,32 @@ defmodule Ecto.Integration.PreloadTest do
     end
   end
 
+  describe "union queries" do
+    test "should preload comments into posts" do
+      p1 = TestRepo.insert!(%Post{title: "p1"})
+      TestRepo.insert!(%Comment{text: "c1", post: p1})
+
+      q1 = from c in Comment, where: c.text == "a"
+      q2 = from c in Comment, where: c.text == "b"
+      query = union(q1, ^q2)
+
+      p1 = TestRepo.preload([p1], [{:comments, query}], [])
+
+      assert [] == p1.comments
+    end
+
+    test "should also fetch comments" do
+      p1 = TestRepo.insert!(%Post{title: "p1"})
+      TestRepo.insert!(%Comment{text: "c1", post: p1})
+
+      q1 = from c in Comment, where: c.text == "a"
+      q2 = from c in Comment, where: c.text == "b"
+
+      query = union(q1, ^q2)
+      assert [] = TestRepo.all(query)
+    end
+  end
+
   defp sort_by_id(values) do
     Enum.sort_by(values, &(&1.id))
   end
--- a/integration_test/cases/preload.exs
+++ b/integration_test/cases/preload.exs
@@ -774,7 +774,8 @@ defmodule Ecto.Integration.PreloadTest do
       TestRepo.insert!(%Comment{text: "c1", post: p1})
 
       q1 = from c in Comment, where: c.text == "a"
-      q2 = from c in Comment, where: c.text == "b"
+      fields = Comment.__schema__(:fields) ++ [:post_id]
+      q2 = from c in Comment, select: map(c, ^fields), where: c.text == "b"
       query = union(q1, ^q2)
 
       p1 = TestRepo.preload([p1], [{:comments, query}], [])


greg-rychlewski:
There are for sure a couple issues when using combinations to preload, but I'm not sure it makes sense to use them in the first place. Here is my thinking:

A preload is done against a single table. There isn't a need to use intersections on a single table because it can be done using filters. Unions correspond to `OR`, intersections to `AND`, set differences `AND NOT`.

If ^ makes sense, maybe raising when a preload query is a combination is the way to go.

pedro-gutierrez:
Hey thank you for your response!

Yeah, rewriting my query also crossed my mind, but I am not sure this would be doable in 100% of the cases.

Also came across https://github.com/elixir-ecto/ecto/issues/3839 which hits the same problem. The workaround described there works well:

```elixir
q1 = from c in Comment, where: c.text == "a"
fields = Comment.__schema__(:fields) ++ [:post_id] 
q2 = from c in Comment, select: map(c, fields), where: c.text == "b"
query = union(q1, ^q2)
```

greg-rychlewski:
Oh I'm surprised that works. Preloading a `has_many` association will also try to order the preload query, which I thought would break with a union. When you run this test it passes?

```elixir
test "should preload comments into posts" do
      p1 = TestRepo.insert!(%Post{title: "p1"})
      TestRepo.insert!(%Comment{text: "c1", post: p1})

      q1 = from c in Comment, where: c.text == "a"
      fields = Comment.__schema__(:fields) ++ [:post_id] 
      q2 = from c in Comment, select: map(c, ^fields), where: c.text == "b"
      query = union(q1, ^q2)

      p1 = TestRepo.preload([p1], [{:comments, query}], [])

      assert [] == p1.comments
    end
```

pedro-gutierrez:
Actually, you are right.

The new error I get is: 

```
** (Postgrex.Error) ERROR 42P01 (undefined_table) missing FROM-clause entry for table "c0"

 query: SELECT c0."id", c0."text", c0."lock_version", c0."post_id", c0."author_id", c0."post_id" FROM "comments" AS c0 WHERE (c0."text" = 'a') AND (c0."post_id" = $1) UNION (SELECT c0."id", c0."text", c0."lock_version", c0."post_id", c0."author_id", c0."post_id" FROM "comments" AS c0 WHERE (c0."text" = 'b')) ORDER BY c0."post_id"
```

I had tested this with a `belongs_to` relationship, and I assumed that would also work with `has_many`. 

Good catch!

greg-rychlewski:
I just remembered you can wrap it a subquery to make both issues go away.

```elixir
q1 = from c in Comment, where: c.text == "a"
q2 = from c in Comment, where: c.text == "b"
query = from s in subquery(union(q1, ^q2))

[p1] = TestRepo.preload([p1], [{:comments, query}], [])
```

But it's still a bit dicey because you have to make sure the foreign key is in all the union queries otherwise the subquery won't be able to select it. So the logic that adds the field to ensure it's there is kind of neutered. All this is also pretty unintuitive unless you've studied the Ecto code or have had it explained to you.

I think there are 2 options:

1. Ecto handles combinations with special logic.
2. Ecto detects combinations in preloads and raises in error suggesting to wrap it in a subquery

I'm still pretty skeptical that combinations are needed for preloads so leaning towards 2. @josevalim Do you have a preference?

josevalim:
I would go with 2 too!

pedro-gutierrez:
Thank you guys for the insight !

greg-rychlewski:
Thanks for the report @pedro-gutierrez! Are you interested in submitting the fix? If not I can do it so no worries.

pedro-gutierrez:
> Thanks for the report @pedro-gutierrez! Are you interested in submitting the fix? If not I can do it so no worries.

Hey I'd love to contribute, but I also feel you're gonna spend a lot less time than me in this, also cos you really know the library inside out, and I don't.

We can't close this PR too.

PR #4004, opened 2022-09-18
Raise if combination query is used in a `many` preload

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/pull/4001.

I highlighted 2 points where I could see someone disagreeing with what I decided to do

### DIFF
--- a/lib/ecto/repo/preloader.ex
+++ b/lib/ecto/repo/preloader.ex
@@ -255,9 +255,9 @@ defmodule Ecto.Repo.Preloader do
       case {card, query.combinations} do
         {:many, [{kind, _} | []]} ->
           raise ArgumentError,
-                "#{kind} queries must be wrapped inside of a subquery " <>
+                "`#{kind}` queries must be wrapped inside of a subquery " <>
                   "when preloading a `has_many` or `many_to_many` association. " <>
-                  "You must also ensure that all members of the #{kind} query " <>
+                  "You must also ensure that all members of the `#{kind}` query " <>
                   "select the parent's foreign key"
 
         {:many, _} ->
--- a/lib/ecto/repo/preloader.ex
+++ b/lib/ecto/repo/preloader.ex
@@ -252,13 +252,20 @@ defmodule Ecto.Repo.Preloader do
 
     # If we are returning many results, we must sort by the key too
     query =
-      case card do
-        :many ->
+      case {card, query.combinations} do
+        {:many, [{kind, _} | []]} ->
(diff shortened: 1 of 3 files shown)


lib/ecto/repo/preloader.ex
```diff
@@ -252,13 +252,20 @@ defmodule Ecto.Repo.Preloader do
...
     query =
-      case card do
-        :many ->
+      case {card, query.combinations} do
+        {:many, [{kind, _} | []]} ->
+          raise ArgumentError,
+                "#{kind} queries must be wrapped inside of a subquery " <>
+                  "when preloading a `has_many` or `many_to_many` association. " <>
+                  "You must also ensure that all members of the #{kind} query " <>
+                  "select the parent's foreign key"
```
greg-rychlewski, line 261: I was going to put an example in this error message because I could see someone being a bit confused. But I couldn't think of any example where I felt a combination preload was the best option and I felt weird advertising a query that I think shouldn't be used.

lib/ecto/repo/preloader.ex
```diff
@@ -252,13 +252,20 @@ defmodule Ecto.Repo.Preloader do
 
     # If we are returning many results, we must sort by the key too
     query =
-      case card do
-        :many ->
+      case {card, query.combinations} do
+        {:many, [{kind, _} | []]} ->
```
greg-rychlewski, line 256: I'm only raising for `:many` because you can get by without a subquery for `:one` queries doing something like this: https://github.com/elixir-ecto/ecto/issues/3839#issuecomment-1050581380.

This means someone using combinations for a `:one` query won't get a nice error message if they don't use the trick above. But I'm thinking that's better than raising for valid queries or breaking people's code who are using the trick above.

lib/ecto/repo/preloader.ex
```diff
@@ -252,13 +252,20 @@ defmodule Ecto.Repo.Preloader do
...
+          raise ArgumentError,
+                "#{kind} queries must be wrapped inside of a subquery " <>
+                  "when preloading a `has_many` or `many_to_many` association. " <>
+                  "You must also ensure that all members of the #{kind} query " <>
```
josevalim, line 260:
```suggestion
                "`#{kind}` queries must be wrapped inside of a subquery " <>
                  "when preloading a `has_many` or `many_to_many` association. " <>
                  "You must also ensure that all members of the `#{kind}` query " <>
```

josevalim (approved):

PR #4006, opened 2022-09-21
Add :force_changes option to cast/4

lmarlow:
Add a `:force_changes` option to `Ecto.Changeset.cast/4`.
When set to true all changes given will be put in the changeset's `:changes` event if they
match the current value or default value.

This is useful for doing an upsert without having to do a read first. Currently to "clear" a value,
which is often the default value for a field, you either need to read the original row first or call `Ecto.Changeset.force_change/3` for each relevant field.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -451,8 +451,7 @@ defmodule Ecto.Changeset do
       Empty values are always replaced by the default value of the respective field.
       If the field is an array type, any empty value inside of the array will be removed.
       Defaults to `[""]`
-    * `:force_changes` - a boolean indicating whether to include default values
-      in `:changes`. Defaults to `false`
+    * `:force_changes` - a boolean indicating whether to include values that don't alter the current data in `:changes`. Defaults to `false`
 
   ## Examples
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -362,6 +362,12 @@ defmodule Ecto.Changeset do
 
   See `cast/4` if you'd prefer to cast and validate external parameters.
 
+  ## Options
+
+    * `:force_changes` - a boolean indicating whether to include values
+      in `:changes` even if they match the current value or the field's default
+      value. Defaults to `false`
+
   ## Examples
 
       iex> changeset = change(%Post{})
@@ -378,6 +384,10 @@ defmodule Ecto.Changeset do
       iex> changeset = change(%Post{title: "title"}, title: "title")
       iex> changeset.changes
       %{}
+
+      iex> changeset = change(%Post{title: "title"}, %{title: "title"}, force_changes: true)
+      iex> changeset.changes
+      %{title: "title"}
 
       iex> changeset = change(changeset, %{title: "new title", body: "body"})
       iex> changeset.changes.title
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -387,35 +387,38 @@ defmodule Ecto.Changeset do
 
   """
   @spec change(Ecto.Schema.t | t | {data, types}, %{atom => term} | Keyword.t) :: t
-  def change(data, changes \\ %{})
+  def change(data, changes \\ %{}, opts \\ [])
 
-  def change({data, types}, changes) when is_map(data) do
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
I was really confused how this helps with upserts until I saw this: https://github.com/elixir-ecto/ecto/pull/2837. Is your issue the same thing where you are trying to do this:

```elixir
params = %{title: "foo", published: true}
changeset = cast(%Post{}, params, [:title, :published])
set_values = Keyword.new(changeset.changes)

options = [
  conflict_target: [:id],
  on_conflict: Post |> update(set: ^set_values) |> where([p], p.foo_condition...)
]
```

If so I'm not sure it makes sense to add this to `change`. The value of doing it with `cast` is that it can modify the value. But with `change` it would be no different than setting the update values from `params`.

lmarlow:
@greg-rychlewski, it is a similar issue to #2837. Nice find!

We get a fixed set of fields every time we want to upsert and we want all those values to override whatever is in the DB. We use `cast` most of the time. There are some cases where we build a `params` map ourselves with our own manipulations and then want a changeset. This is where we use `change`.

Without know what's in the database, currently `change` gives us:

```elixir
iex> changeset = Ecto.Changeset.change({%{}, %{title: :string}}, %{title: nil})
iex> changeset.changes
%{}
```

with `force_changes: true` we get:

```elixir
iex> changeset = Ecto.Changeset.change({%{}, %{title: :string}}, %{title: nil}, force_changes: true)
iex> changeset.changes
%{title: nil}
```

What we have to do now is something like this:

```elixir
  def changeset(params) do
    params =
      Map.new(params, fn
        # set values to nil that do not cast from empty strings like `:date`
        {k, v} when is_blank?(v) -> {k, nil}
        {k, v} -> {k, v}
      end)

    changeset = cast(%__MODULE__{}, params, __MODULE__.__schema__(:fields), empty_values: [])
    
    for {k, nil} <- params, reduce: changeset do
      changeset -> Ecto.Changeset.force_change(changeset, k, nil)
    end
  end
```

greg-rychlewski:
Oh yes I can see how this would help with `cast`. 

But for `change` it looks a bit funny to me because there is no extra transformation as with casting. And `force_change` already exists for this purpose. Maybe it would make more sense to extend `force_change` to also work with a map of changes as a slight convenience over iterating over the params and calling it one by one.

greg-rychlewski:
I think `change` gets a bit hairy too with associations/embeds. You don't want to allow those to be forced. So then you have a situation where people might be using the force option and expecting it to work with them. With `cast` and `force_change` it raises a nice error.

josevalim:
Agreed. Let's go with `cast` only for now. :)

lmarlow:
I removed it from `change` and updated the PR title and description.

Thank you for the feedback!

lib/ecto/changeset.ex
```diff
@@ -436,7 +436,9 @@ defmodule Ecto.Changeset do
   During casting, all `permitted` parameters whose values match the specified
   type information will have their key name converted to an atom and stored
   together with the value as a change in the `:changes` field of the changeset.
-  All parameters that are not explicitly permitted are ignored.
+  If the cast value matches the default value for the schema, it will not be
+  included in the `:changes` unless the `:force_changes: true` option is
+  provided. All parameters that are not explicitly permitted are ignored.
```
greg-rychlewski, line 441:
```suggestion
  If the cast value matches the current value for the field, it will not be
  included in the `:changes` unless the `:force_changes: true` option is
  provided. All parameters that are not explicitly permitted are ignored.
```

Sorry one last nitpick. The change won't be registered if it doesn't alter the current data, which in the case of a new struct is the default value.

lib/ecto/changeset.ex
```diff
@@ -449,6 +451,8 @@ defmodule Ecto.Changeset do
       Empty values are always replaced by the default value of the respective field.
       If the field is an array type, any empty value inside of the array will be removed.
       Defaults to `[""]`
+    * `:force_changes` - a boolean indicating whether to include default values
+      in `:changes`. Defaults to `false`
```
greg-rychlewski, line 455:
```suggestion
    * `:force_changes` - a boolean indicating whether to include values that don't alter the current data in `:changes`. Defaults to `false`
```

greg-rychlewski:
Thanks!

PR #4007, opened 2022-09-22
Clarify :returning option for inserts

nathanl:
I was confused when I had `on_conflict: :nothing` and I got back a struct with `id: nil`.

If that behavior is PostgreSQL-specific, is there some other way we can alert people to this?

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1524,7 +1524,9 @@ defmodule Ecto.Repo do
 
     * Specify `read_after_writes: true` in your schema for choosing
       fields that are read from the database after every operation.
-      Or pass `returning: true` to `insert` to read all fields back:
+      Or pass `returning: true` to `insert` to read all fields back.
+      (Note that it will only read from the database if at least one
+      field is updated).
 
           MyRepo.insert(%Post{title: "this is unique"}, returning: true,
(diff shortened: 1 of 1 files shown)


josevalim:
It is actually a bit more nuanced and we do talk about it a couple paragraphs above!

josevalim:
Closing this because I believe we cover this information. Please let me know if that's not the case, thank you @nathanl :heart:

nathanl:
@josevalim I forgot about this PR and am re-reading the docs, but I think the existing documentation is actually misleading without this change. It says:

>   If you need to guarantee the data in the returned struct mirrors the
  database, you have three options:

one of which is

> Specify `read_after_writes: true` in your schema for choosing
>      fields that are read from the database after every operation.
>      Or pass `returning: true` to `insert` to read all fields back:

But passing `returning: true` *does not* "guarantee the data in the returned struct mirrors the database", as my comment is intended to clarify.

greg-rychlewski:
To play devil's advocate, if nothing was inserted/updated then isn't it correct that your data mirrors the database?

josevalim:
@greg-rychlewski I guess it only partially reflects the data in the database.

greg-rychlewski (approved):

PR #4020, opened 2022-10-03
Allow selected_as at the root of dynamic/2

greg-rychlewski:
This can probably be extended to subqueries as well but will look into that later.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -557,6 +557,18 @@ defmodule Ecto.Query do
 
       from query, select: [:period, :metric], select_merge: ^%{metric: metric}
 
+  Aliasing fields with `selected_as/2` and referencing them with `selected_as/1`
+  is also allowed:
+
+      fields = %{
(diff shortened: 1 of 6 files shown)


josevalim (approved):

greg-rychlewski:
@ak-14 You can use selected_as inside dynamics now. Please see the doc changes to see how to use it.

ak-14:
@greg-rychlewski Thank you very much for adding support for this case so quickly! I tested it and it works as shown in the provided examples.
At first I tried selected_as with a pipe (similar to the example in the [docs](https://hexdocs.pm/ecto/Ecto.Query.API.html#selected_as/2)) but that raised an error. Then I tried wrapping the entire statement in selected_as and that worked fine (```dynamic([p],  selected_as(p.visits |> coalesce(0) |> sum(), :sum_visits))```).

greg-rychlewski:
ah, thank you for pointing that out. there is a flaw in the way `selected_as` is compiled. i'll work on a fix.

PR #4023, opened 2022-10-07
Refactor functions to use repo variable

kelcecil:
I was reading the `Repo` source and noticed that the `insert/2` function had been updated to use the `repo` variable, but none of the functions had been updated. This MR updates the remaining functions to use the existing `repo` variable instead of calling `get_dynamic_repo/0` twice.


josevalim:

PR #4026, opened 2022-10-10
add dynamic `from` hints to typespec

ruslandoga:
👋

This PR continues on https://github.com/elixir-ecto/ecto/pull/3592 and adds dynamic hints to typespecs.

### DIFF
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -50,6 +50,9 @@ defmodule Ecto.Query.Builder.From do
   end
 
   defp escape_source(query, _env), do: query
+  
+  @typep dynamic_hint :: {atom, term}
+  @typep hints :: String.t() | dynamic_hint | [String.t() | dynamic_hint]
 
   @doc """
   Builds a quoted expression.
@@ -58,14 +61,14 @@ defmodule Ecto.Query.Builder.From do
   If possible, it does all calculations at compile time to avoid
   runtime work.
   """
-  @spec build(Macro.t(), Macro.Env.t(), atom, String.t | nil, nil | {:ok, String.t | nil} | [String.t]) ::
+  @spec build(Macro.t(), Macro.Env.t(), atom, String.t | nil, hints | nil) ::
           {Macro.t(), Keyword.t(), non_neg_integer | nil}
   def build(query, env, as, prefix, maybe_hints) do
     hints = List.wrap(maybe_hints)
 
     unless Enum.all?(hints, &is_valid_hint/1) do
       Builder.error!(
-        "`hints` must be a compile time string, list of strings, or a tuple " <>
+        "`hints` must be a compile time string, list of strings, a tuple, or a list of tuples " <>
           "got: `#{Macro.to_string(maybe_hints)}`"
       )
     end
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder/from.ex
```diff
@@ -58,14 +61,14 @@ defmodule Ecto.Query.Builder.From do
   If possible, it does all calculations at compile time to avoid
   runtime work.
   """
-  @spec build(Macro.t(), Macro.Env.t(), atom, String.t | nil, nil | {:ok, String.t | nil} | [String.t]) ::
```
ruslandoga, line 61: Why was `{:ok, String.t | nil}` allowed as `hints`?

ukutaht:
Context: `mix dialyzer` fails on https://github.com/plausible/analytics because we use the dynamic hints feature.

lib/ecto/query/builder/from.ex
```diff
@@ -50,6 +50,9 @@ defmodule Ecto.Query.Builder.From do
   end
 
   defp escape_source(query, _env), do: query
+  
+  @typep dynamic_hint :: {atom, term}
+  @typep hints :: String.t() | dynamic_hint | [String.t() | dynamic_hint]
```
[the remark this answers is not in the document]
josevalim: Ignore me, it does not.

lib/ecto/query/builder/from.ex
```diff
@@ -58,14 +61,14 @@ defmodule Ecto.Query.Builder.From do
   If possible, it does all calculations at compile time to avoid
   runtime work.
   """
-  @spec build(Macro.t(), Macro.Env.t(), atom, String.t | nil, nil | {:ok, String.t | nil} | [String.t]) ::
+  @spec build(Macro.t(), Macro.Env.t(), atom, String.t | nil, hints | nil) ::
```
josevalim, line 64:
```suggestion
  @spec build(Macro.t(), Macro.Env.t(), atom, {:ok, String.t | nil} | nil, hints | nil) ::
```

lib/ecto/query/builder/from.ex
josevalim, line 55: Hints is always a list or nil, so we want this:

```suggestion
  @typep hints :: [String.t() | {atom, term}]
```

josevalim, line 55: I think we should consider calling `List.wrap` in `Ecto.Query.from` so we always receive a list. This way, we can do this:

```suggestion
  @typep hints :: [String.t() | dynamic_hint]
```

This simplifying both the code and the specs here.

lib/ecto/query/builder/from.ex
```diff
@@ -58,15 +60,13 @@ defmodule Ecto.Query.Builder.From do
   If possible, it does all calculations at compile time to avoid
   runtime work.
   """
-  @spec build(Macro.t(), Macro.Env.t(), atom, String.t | nil, nil | {:ok, String.t | nil} | [String.t]) ::
+  @spec build(Macro.t(), Macro.Env.t(), atom, {:ok, String.t | nil} | nil, hints | nil) ::
```
josevalim, line 63:
```suggestion
  @spec build(Macro.t(), Macro.Env.t(), atom, {:ok, String.t | nil} | nil, hints) ::
```

lib/ecto/query/builder/from.ex
```diff
@@ -58,15 +60,13 @@ defmodule Ecto.Query.Builder.From do
...
           {Macro.t(), Keyword.t(), non_neg_integer | nil}
-  def build(query, env, as, prefix, maybe_hints) do
-    hints = List.wrap(maybe_hints)
-
+  def build(query, env, as, prefix, hints) do
     unless Enum.all?(hints, &is_valid_hint/1) do
       Builder.error!(
-        "`hints` must be a compile time string, list of strings, or a tuple " <>
-          "got: `#{Macro.to_string(maybe_hints)}`"
+        "`hints` must be a list of compile time strings or tuples " <>
```
josevalim, line 68: The previous warning was correct. We changed the implementation but the user API is the same!

ruslandoga: 🙈

I got carried away.

ruslandoga: Fixed.

josevalim:

PR #4028, opened 2022-10-12
Clarify how the equal?/2 callback is used

davydog187:
I'm working on a custom ecto type for postgres `range` types, and I wanted to better understand how `equal?/2` would be used. Knowing this sent me down the right path :) 

It was also helpful to read https://github.com/elixir-ecto/ecto/pull/2651/files


josevalim:

PR #4031, opened 2022-10-14
Fix example with_named_binding/3

maartenvanvliet:
Current example gave an error:
```
** (ArgumentError) invalid binding passed to Ecto.Query.join/5, should be list of variables, got: c in assoc(p, :comments)
```


josevalim:

PR #4034, opened 2022-10-21
fix the example of `Ecto.Enum`'s `embed_as` option

shutangyu:
it should be `:dumped` instead of `:dump`


josevalim:

PR #4039, opened 2022-10-27
Expand documentation for logger options

greg-rychlewski:
Since there was a bit of confusion in a recent PR, wanted to expand on this option a bit.


greg-rychlewski:
`true` is technically allowed from the `Repo` functions but it doesn't do anything different than leaving the option unspecified, so I thought it was simpler to leave it out.

josevalim (approved):

Issue #4033, opened 2022-10-18
Using `selected_as` in a scope of subquery fails


andrhua:
### Elixir version

Elixir 1.13.4 (compiled with Erlang/OTP 24)

### Database and Version

Postgres 13.5-1

### Ecto Versions

ecto 3.9.1, ecto_sql 3.9.0

### Database Adapter and Versions (postgrex, myxql, etc)

0.16.5

### Current behavior

I have following query:
```elixir
def foo_query(...) do
  Ecto.Query.from(m in Model,
    ...,
    select: %{
      ...,
      bar: column |> complex_fragment() |> selected_as(:foo)
    },
    group_by: [selected_as(:foo), ...]
  )
end
```
which is then used as a subquery:
```elixir
Ecto.Query.from(os in subquery(other_query()),
  join: s in subquery(foo_query(...)),
  on:  os.bar == s.bar
  ...
)
```
This returns a runtime error
```
`selected_as/2` can only be used in the outer most `select` expression. If you are attempting to alias a field from a subquery or cte, it is not allowed because the fields are automatically aliased by the corresponding map/struct key. The following field aliases were specified: [:foo].
```
as if I was joining on `selected_as(:foo)` in outer query, and that's fair, because you could just use key of a struct

### Expected behavior

I think there's should no problems for properly scoped usage of `selected_as`, which doesn't "leak" between nested queries

greg-rychlewski:
Hi @andrhua, 

If you use `selected_as` inside of a subquery then the SQL that is eventually generated would produce a double alias and the query would be invalid. Something like this:

```elixir
postgres=# select 1 as foo as bar;
ERROR:  syntax error at or near "as"
LINE 1: select 1 as foo as bar;
```

Could you elaborate, which valid SQL would you expect to be produced from your example above?

andrhua:
```sql
select complex_transform(column) as foo
...
group by foo, ...
```
in other words I don't wanna to copy `complex_transform(column)` in GROUP BY clause

greg-rychlewski:
Sorry not sure I understand. When you put it in a map with the key `bar` it will create the alias bar automatically. So what I mean is when you are already creating that alias and then on top of it use `selected_as` what behaviour are you expecting with 2 aliases?

andrhua:
> When you put it in a map with the key bar inside of a subquery it will create the alias bar automatically
right, but there's no way to use it in `group_by: ...` if I'm not mistaken
[...]

PR #4040, opened 2022-10-28
Allow selected_as to be used in subquery

greg-rychlewski:
This addresses the concern raised in https://github.com/elixir-ecto/ecto/issues/4033. 

I'm not sure how desirable this change is, but wanted to put it out there for feedback. I think the issue reporter has a valid point about re-using the same query inside and outside of a subquery. But it's not the cleanest change to make it work.

The idea is that the subquery tracks which of its fields are aliased and then `collect_fields` and `source_take!` will swap them. This is for better UX so that the user can always use the map/keyword list key to reference the subquery fields from the outer query.

Since the change is kind of dirty I won't fuss if it's rejected :).

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -727,17 +727,18 @@ defmodule Ecto.Query.API do
 
   ## Subqueries
 
-  When used inside of a subquery, the alias given to `selected_as/2` must be referenced
-  by the outer query. For example, the outer query below selects `s.num_visits` instead of
-  `s.visits`:
+  Subqueries automatically alias the selected fields, for example, one can write:
 
-      s = from p in Post, select: %{visits: p.visits |> coalesce(0) |> selected_as(:num_visits)}
+      s = from p in Post, select: %{visits: coalesce(p.visits, 0)}
+      from(s in subquery(s), select: s.visits)
+
+  However, one can also use `selected_as` to override the default naming:
+
+      s = from p in Post, select: %{visits: coalesce(p.visits, 0) |> selected_as(:num_visits)}
       from(s in subquery(s), select: s.num_visits)
 
-  This is because queries must reference the fields of the query source and not the
-  Elixir data structure. In some circumstances, such as a subquery without `selected_as/2`,
-  these two are the same. This does not hold true when the user modifies the names of the fields
-  using `selected_as/2`.
+  The name given to `selected_as/2` can also be referenced in `selected_as/1`,
+  as in regular queries.
   """
   def selected_as(selected_value, name), do: doc! [selected_value, name]
 
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -697,8 +697,6 @@ defmodule Ecto.Query.API do
   If an error is raised mentioning an unknown column, most likely the alias is being
   referenced somewhere that is not allowed. Consult the documentation for the database
   to ensure the alias is being referenced correctly.
-
-
   """
   def selected_as(name), do: doc! [name]
 
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -697,6 +697,8 @@ defmodule Ecto.Query.API do
   If an error is raised mentioning an unknown column, most likely the alias is being
   referenced somewhere that is not allowed. Consult the documentation for the database
   to ensure the alias is being referenced correctly.
+
+
   """
   def selected_as(name), do: doc! [name]
 
@@ -724,6 +726,20 @@ defmodule Ecto.Query.API do
 
   Using this in conjunction with `selected_as/1` is recommended to ensure only defined aliases
   are referenced.
+
+  ## Subqueries
+
+  When used inside of a subquery, the alias given to `selected_as/2` must be referenced
+  by the outer query. For example, the outer query below selects `s.num_visits` instead of
+  `s.visits`:
+
+      s = from p in Post, select: %{visits: p.visits |> coalesce(0) |> selected_as(:num_visits)}
+      from(s in subquery(s), select: s.num_visits)
+
+  This is because queries must reference the fields of the query source and not the
+  Elixir data structure. In some circumstances, such as a subquery without `selected_as/2`,
+  these two are the same. This does not hold true when the user modifies the names of the fields
+  using `selected_as/2`.
   """
   def selected_as(selected_value, name), do: doc! [selected_value, name]
 
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -329,8 +329,8 @@ defmodule Ecto.Query.Planner do
   defp normalize_subquery_types([{source_alias, type_value} | types], [field | fields], select_aliases, acc) do
     if Map.has_key?(select_aliases, source_alias) do
       raise ArgumentError, """
-      the alias, #{inspect(source_alias)}, provided to `selected_as/2`
-      conflicts with the subquery's automatically aliasing.
+      the alias, #{inspect(source_alias)}, provided to `selected_as/2` conflicts 
+      with the subquery's automatically aliasing.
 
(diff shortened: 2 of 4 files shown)


josevalim:
I agree. Ideally we would provide this but the implementation would have to be somehow simpler. I also think some of our APIs rely on subqueries, such as `Repo.exists?`, which means a selected_as given to exists would break.

Perhaps one way to implement this is to say that the map fields in the subquery are always automatically aliased as their map keys? Perhaps we even do this for regular queries. This way all subquery fields are automatically `selected_as`, then it is only a debate if you can also explicitly alias them with `selected_as` in subqueries too or only the implicit alias is allowed.

greg-rychlewski:
>Perhaps one way to implement this is to say that the map fields in the subquery are always automatically aliased as their map keys? 

Thanks, that's a lot nicer than this. I'll work on your idea instead.

>then it is only a debate if you can also explicitly alias them with selected_as in subqueries too or only the implicit alias is allowed.

It might be confusing to allow both because then you won't be able to select the map key from the outer query. Maybe starting with only the implicit one and waiting for feedback to extend it to custom selected_as?

josevalim:
> It might be confusing to allow both because then you won't be able to select the map key from the outer query. Maybe starting with only the implicit one and waiting for feedback to extend it to custom selected_as?

That works for me! But then we should also allow implicit for map keys on all queries. Is that doable?

josevalim:
Not only that, we should discuss about ambiguities: what happens if a map key matches a selected_as.

greg-rychlewski:
I thought about this more carefully and have a new proposal. Please let me know what you think. Sorry for the mouthful, tried to be as concise as possible while still touching on all the intricacies

- Don't add implicit aliases to outer queries

    **Reason**

    It will break existing queries like this:

    ```elixir
    select: %{field: fragment("some_fragment as alias)}, order_by: fragment("alias")
    ```

    I expect this kind of construct to not be uncommon since most built-in/custom SQL functions require a fragment. I think it could also be really annoying for users to find these types of queries in all their apps and adjust them. 

- Allow the implicit aliases in subqueries/CTEs to be referenced in `selected_as/1`

    **Reason**

    This can be considered a new helpful feature. It's a bit suboptimal to have to explicitly create the alias when they are already being created.
 
- Allow explicit aliases with `selected_as/2` in subqueries/CTEs. Explicit takes precedence over implicit. The outer query will have to reference the explicit alias, i.e. 

    ```elixir
    s = from q in query, select: %{field: selected_as(q.field, :alias)}
    from s in subquery(s), select: s.alias (instead of s.field)
    ```

    **Reason**

    Without this the user won't be able to re-use the query inside and outside of a subquery, if we accept outer queries shouldn't have implicit aliases.

    The reason for selecting the alias in the outer query is because otherwise we need to do something like my original PR, which is a big mess. I think semantically this makes more sense too, because you should be referencing the "fields" from the query and not the elixir construct. I expect there to be some confusion about this because things in Ecto work so seamlessly that it's hard to forget you are really working with SQL instead of Elixir data structures. My gut feeling is this is ok because using `selected_as` inside of a subquery shouldn't be too common of a thing. And the confusion will eventually lead to thinking in the "right" way, which is good.

- If conflicts between implicit/explicit aliases arise, raise a nice error message.

    **Reason**

    This is the only way to handle it IMO because you either can't have the same alias twice or if you can, it will be ambiguous what you are referencing. So i'm thinking we re-use the existing validation (i.e. check the `aliases` map to see if it was already used and raise if so). The error message can be extending to explain that subqueries/CTEs have implicit aliases that can cause conflicts.

Also I can extend the existing subquery documentation to explain the existing implicit aliasing and how it will interact with `selected_as`.

josevalim:
Thank you @greg-rychlewski! Given what you said, I think we only need this:

> Allow explicit aliases with selected_as/2 in subqueries/CTEs. Explicit takes precedence over implicit. The outer query will have to reference the explicit alias

Especially because the main goal is to support regular queries running as subqueries.

However, we can also consider doing "Allow the implicit aliases in subqueries/CTEs to be referenced in selected_as/1" if that would simplify the implementation.

Thoughts?

greg-rychlewski:
>However, we can also consider doing "Allow the implicit aliases in subqueries/CTEs to be referenced in selected_as/1" if that would simplify the implementation.

Oh that would add complexity, so I agree just adding the explicit aliasing seems the best.

greg-rychlewski:
Thanks for talking this out with me. It's ready for review now.

I hope you don't mind, but I would like to handle CTEs a bit later. I'm gonna be short on free time for a while so it would be nice to get this one working and then start on CTE after.

lib/ecto/query/planner.ex
```diff
@@ -307,10 +307,52 @@ defmodule Ecto.Query.Planner do
...
+
+      For instance, the following query is not allowed because the alias `:x2` used in `selected_as/2`
+      is the same as another key in the map being selected:
+
+        s = from(s in Schema, select: %{x: selected_as(s.x, :x2), x2: s.x})
+        from s in subquery(s)
+
+      If the alias provided to `selected_as/2` is changed to any atom other than `:x2`, the above query will work.
+      The alias `:x` would also be allowed because it is the key corresponding to the `selected_as/2` value.
+      """
```
josevalim, line 344: Should we just make it so `selected_as` wins, as you originally proposed?

lib/ecto/query/api.ex
```diff
@@ -724,6 +724,20 @@ defmodule Ecto.Query.API do
...
+  by the outer query. For example, the outer query below selects `s.num_visits` instead of
+  `s.visits`:
+
+      s = from p in Post, select: %{visits: p.visits |> coalesce(0) |> selected_as(:num_visits)}
+      from(s in subquery(s), select: s.num_visits)
+
+  This is because queries must reference the fields of the query source and not the
+  Elixir data structure. In some circumstances, such as a subquery without `selected_as/2`,
+  these two are the same. This does not hold true when the user modifies the names of the fields
+  using `selected_as/2`.
```
josevalim, line 740: Assuming we will make it so `selected_as` always wins, I would go with this:

```suggestion
  Subqueries automatically alias the selected fields, for example, one can write:

      s = from p in Post, select: %{visits: coalesce(p.visits, 0)}
      from(s in subquery(s), select: s.visits)

  However, one can also use `selected_as` to override the default naming:

      s = from p in Post, select: %{visits: coalesce(p.visits, 0) |> selected_as(:num_visits)}
      from(s in subquery(s), select: s.num_visits)

  The name given to `selected_as/2` can also be referenced in `selected_as/1`,
  as in regular queries.
```

lib/ecto/query/planner.ex
greg-rychlewski: Oh what I was trying to say is this:

If you have `select: %{x: selected_as(field, :alias)}` then `:alias` wins over `:x`.

If you have `select: %{x: selected_as(field, :alias), alias: field2}` then `selected_as(field, :alias)` isn't allowed anymore because it conflicts with the other field's alias.

My wording might be unclear. I updated it.

greg-rychlewski: Now that I look at this again it isn't perfect. i.e. this shouldn't raise: `select: %{x: selected_as(field, :alias), alias: selected_as(field2, :not_alias)}`.

I think I can change this to do a pass over the over the keys and make them `selected_as` as long as their value isn't `selected_as` something different.

josevalim: Oh, I see.

I think raising on this case is enough:

> If you have select: %{x: selected_as(field, :alias), alias: field2} then selected_as(field, :alias) isn't allowed anymore because it conflicts with the other field's alias.

I think we should also raise on this:

> Now that I look at this again it isn't perfect. i.e. this shouldn't raise: select: %{x: selected_as(field, :alias), alias: selected_as(field2, :not_alias)}.

Because that would be just too confusing anyway.

josevalim (approved):

Issue #3405, opened 2020-09-10
:idle_time and :queue_time telemetry only invoked for SELECT queries with Postgres adapter


sezaru:
### Environment

* Elixir version (elixir -v): 1.10.4
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): PostgreSQL 12
* Ecto version (mix deps): 3.4.5
* Database adapter and version (mix deps): postgrex 0.15.5
* Operating system: Ubuntu 20.04

### Current behavior

Ecto.Repo documentation says that some events are expected to be invoked during query execution, these events are `:idle_time`, `:queue_time`, `:query_time`, `:decode_time` and `:total_time`.

The issue I'm having is that I'm only seeing events `:query_time`, `:decode_time` and `:total_time` being invoked when doing INSERT and UPDATE queries, `:idle_time` and `:queue_time` only shows up in SELECT queries.

For completeness, I use these two functions to filter my events:
```elixir
  def query_metadata(%{result: {_, %{command: command}}}), do: %{command: command}

  def keep?(%{result: {_, %{command: _}}}), do: true
  def keep?(_), do: false
```

And my metrics are configured as follows:
```elixir
[
  distribution("data_db.exchanges.binance.repo.query.idle_time",
    unit: {:native, :millisecond},
    reporter_options: [
      buckets: [1, 3, 6, 10, 25, 50, 100, 500, 1000]
    ],
    keep: &keep?/1,
    tag_values: &query_metadata/1,
    tags: [:command]
  ),
  distribution("data_db.exchanges.binance.repo.query.queue_time",
    unit: {:native, :millisecond},
    reporter_options: [
      buckets: [1, 3, 6, 10, 25, 50, 100, 500, 1000]
    ],
    keep: &keep?/1,
    tag_values: &query_metadata/1,
    tags: [:command]
  ),
  distribution("data_db.exchanges.binance.repo.query.query_time",
    unit: {:native, :millisecond},
    reporter_options: [
      buckets: [1, 3, 6, 10, 25, 50, 100, 500, 1000]
    ],
    keep: &keep?/1,
    tag_values: &query_metadata/1,
    tags: [:command]
  ),
  distribution("data_db.exchanges.binance.repo.query.decode_time",
    unit: {:native, :millisecond},
    reporter_options: [
      buckets: [1, 3, 6, 10, 25, 50, 100, 500, 1000]
    ],
    keep: &keep?/1,
    tag_values: &query_metadata/1,
[...]

PR #4042, opened 2022-11-03
docs(Ecto.Repo): Inform that measurements may, but don't have to, be present

jtomaszewski:
As described in https://github.com/elixir-ecto/ecto/issues/3405 , measurements like `:idle_time` may be nil. Documentation currently doesn't mention it. It gives impression that it those measurements are always present (that they are never `nil`).


josevalim:

PR #4047, opened 2022-11-12
Update docs to use Config instead of Mix.Config

adamzapasnik:
Ecto requires `elixir: "~> 1.10",` and `Config` was introduced back in `1.9`, think it's a good idea to update the docs 🤔


josevalim:

PR #4050, opened 2022-11-17
Simplify the handling of empty values

josevalim:
Prior to this patch, the logic for empty values
was scatted both on cast and on validate_required/2 (which would check for empty trimmed strings).

This PR adds support for functions in empty_values, which allows us to move the prunning from
validate_required to cast.

On the release notes, we should advise those passing `empty_values` to build on top of
`Ecto.Changeset.empty_values/1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## v3.10.0-dev
 
+### Enhancements
+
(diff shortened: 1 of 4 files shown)


greg-rychlewski (approved):
Maybe release notes can also emphasize that strings like `"   "` will now be caught by default `empty_values`

josevalim:

PR #4052, opened 2022-11-20
keep url scheme

ruslandoga:
Clickhouse Ecto adapter uses HTTP to connect to the database where the scheme (http or https) is important to know. Right now [small hacks](https://github.com/plausible/analytics/pull/2427) are required to work around the missing scheme information when configuring the adapter with `:url`.


josevalim:

PR #4054, opened 2022-11-23
Add support for cross lateral joins.

wilkinson4:
Ran into a case where we needed to use a Postgres Cross Join Lateral. Figured I would make a PR to allow Ecto to support it. The reason we needed it is because we have had multiple situations that fit [this stackoverflow post's answer](https://dba.stackexchange.com/a/126974)

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -329,7 +329,7 @@ defmodule Ecto.Query.Planner do
   defp normalize_subquery_types([{source_alias, type_value} | types], [field | fields], select_aliases, acc) do
     if Map.has_key?(select_aliases, source_alias) do
       raise ArgumentError, """
-      the alias, #{inspect(source_alias)}, provided to `selected_as/2` conflicts 
+      the alias, #{inspect(source_alias)}, provided to `selected_as/2` conflicts
       with the subquery's automatic aliasing.
 
       For example, the following query is not allowed because the alias `:y`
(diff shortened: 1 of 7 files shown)


greg-rychlewski:
Thank you. This would also require a change in ecto_sql otherwise the postgres adapter won't understand the join qualifier: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L515

wilkinson4:
> Thank you. This would also require a change in ecto_sql otherwise the postgres adapter won't understand the join qualifier: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L515

Thank you! I was just finishing the [ecto_sql PR](https://github.com/elixir-ecto/ecto_sql/pull/458).

josevalim (approved):

lib/ecto/query/planner.ex
```diff
@@ -891,11 +891,11 @@ defmodule Ecto.Query.Planner do
       end
 
       case find_source_expr(query, child_ix) do
-        %JoinExpr{qual: qual} when qual in [:inner, :left, :inner_lateral, :left_lateral] ->
+        %JoinExpr{qual: qual} when qual in [:inner, :left, :inner_lateral, :left_lateral, :cross_lateral] ->
```
greg-rychlewski, line 894:
```suggestion
        %JoinExpr{qual: qual} when qual in [:inner, :left, :inner_lateral, :left_lateral] ->
```
I don't know if this change is right. I think the point of this clause is that association joins need to be on the appropriate key. If `:cross` is not allowed here I don't believe `:cross_lateral` should be

greg-rychlewski:
Thank you! Please use the commit from master in your ecto_sql PR then we should be good to go :)

wilkinson4:
> Thank you! Please use the commit from master in your ecto_sql PR then we should be good to go :)
You want me to add this: `{:ecto, git: "https://github.com/elixir-ecto/ecto"}` and commit the mix.lock file?

PR #4057, opened 2022-11-25
Don't wrap changes to embeds in a transaction

derekkraan:
Hi all,

I came across this line while attempting to discover why Ecto is wrapping some inserts in a transaction.

As far as I can tell, inserting or updating an embed should always be a single row insert / update, and therefore does not require a transaction.

Or am I missing something?


derekkraan:
Ahh then I understand. I might come back with a new PR taking into account this additional information :)

PR #4058, opened 2022-11-25
improve embed docs and remove dead code

pmargreff:
- While playing with embed, by a hunch confirmed by [the code](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L2080) I discovered it accepts the primary key as an option for inline embedded schemas. After reading docs for nth time I noticed it was documented, but not with the options. So I'm taking the call to group `primary_key` with other options accepted by `embeds_one`.
 - While documenting the option I also found a `strategy` option, after not finding any relevant ref to that I ~blamed~ tracked down to an old embed struct altered [long ago](https://github.com/elixir-ecto/ecto/commit/7b0a2475223c70d87fb65493130a3d6ce99813b4) to remove the `:strategy` as a feature for both `embed_one` and `embed_many` structs, looks like the current `:strategy` was left behind of that removal.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -2049,7 +2049,7 @@ defmodule Ecto.Schema do
     Module.put_attribute(mod, :ecto_changeset_fields, {name, {:assoc, struct}})
   end
 
-  @valid_embeds_one_options [:strategy, :on_replace, :source]
+  @valid_embeds_one_options [:on_replace, :source]
 
   @doc false
   def __embeds_one__(mod, name, schema, opts) when is_atom(schema) do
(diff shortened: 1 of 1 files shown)


lib/ecto/schema.ex
```diff
@@ -1621,16 +1626,12 @@ defmodule Ecto.Schema do
...
         field :name, :string
         field :age,  :integer
       end
 
-  Primary keys are automatically set up for  embedded  schemas as well,
-  defaulting  to  `{:id,  :binary_id, autogenerate:   true}`. You can
-  customize it by passing a `:primary_key` option with the same arguments
-  as `@primary_key` (see the [Schema attributes](https://hexdocs.pm/ecto/Ecto.Schema.html#module-schema-attributes)
-  section for more info).
+
```
josevalim, line 1634:
```suggestion
```

josevalim:

PR #4060, opened 2022-12-06
Add `:serial` type to Ecto Schema docs

FreedomBen:
Add an entry to the type table for `:serial` type.  This is an autoincrementing integer.


josevalim:
Thanks for the PR but I believe there is no such Ecto type. It exists for migrations but they are not part of the schema. Does it make sense?

FreedomBen:
Ah yes, that makes sense.  Is that also the case with the `:id` type or is that used in schemas as well as migrations?

PR #4062, opened 2022-12-07
Add support for dynamics in query preloads

zachallaun:
First-pass implementation of this proposal from the mailing list: [Support runtime dynamic preload bindings in Ecto.Query](https://groups.google.com/g/elixir-ecto/c/Z_tFUi6ONWU/m/e-YPRjIzCgAJ).

In the proposal, I pointed to `Ecto.Query.Builder.Update` as a potential "template", but ended up using the strategy used by the `OrderBy`, `GroupBy`, `Filter`, etc. modules instead. This strategy uses a separate `build/4` function clause that delegates to `Ecto.Query.Builder.Preload.preload!/2` at runtime if the outer-most expression is interpolated.

This means that any existing code that interpolates preloads at the root, regardless of whether they use dynamics, will go down this new code-path. Previously, anything that was interpolated at the root was just added to `query.preloads` via this clause:

```elixir
defp escape({:^, _, [inner]}, _mode, preloads, assocs, _vars) do
  {[inner|preloads], assocs}
end
```

Now, the preload structure will be walked at runtime and converted to a `{preloads, assocs}` tuple by `Ecto.Query.Builder.Preload.expand/2`. I've added additional tests to try and account for this, but more may be warranted.

Would love to get feedback at this point!

### DIFF
--- a/lib/ecto/query/builder/preload.ex
+++ b/lib/ecto/query/builder/preload.ex
@@ -88,6 +88,7 @@ defmodule Ecto.Query.Builder.Preload do
     idx = Builder.find_var!(var, vars)
     {inner_preloads, inner_assocs} = escape(list, :assoc, [], [], vars)
     assocs = [{key, {idx, Enum.reverse(inner_assocs)}}|assocs]
+
     case inner_preloads do
       [] -> {preloads, assocs}
       _  -> {[{key, Enum.reverse(inner_preloads)}|preloads], assocs}
@@ -116,12 +117,6 @@ defmodule Ecto.Query.Builder.Preload do
     Builder.error! "malformed key in preload `#{Macro.to_string(other)}` in query expression"
   end
 
-  defp assert_assoc!(mode, _atom, _var) when mode in [:both, :assoc], do: :ok
-  defp assert_assoc!(_mode, atom, var) do
-    Builder.error! "cannot preload join association `#{Macro.to_string atom}` with binding `#{var}` " <>
-                   "because parent preload is not a join association"
-  end
-
   @doc """
   Called at runtime to check dynamic preload keys.
   """
@@ -140,6 +135,12 @@ defmodule Ecto.Query.Builder.Preload do
   runtime work.
   """
   @spec build(Macro.t, [Macro.t], Macro.t, Macro.Env.t) :: Macro.t
+  def build(query, _binding, {:^, _, [expr]}, _env) do
+    quote do
+      Ecto.Query.Builder.Preload.preload!(unquote(query), unquote(expr))
+    end
+  end
+
   def build(query, binding, expr, env) do
     {query, binding} = Builder.escape_binding(query, binding, env)
     {preloads, assocs} = escape(expr, binding)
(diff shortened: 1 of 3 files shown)


lib/ecto/query/builder/preload.ex
```diff
@@ -156,4 +157,126 @@ defmodule Ecto.Query.Builder.Preload do
...
+
+  defp expand(list, query, mode, preloads, assocs) when is_list(list) do
+    Enum.reduce(list, {preloads, assocs}, fn item, acc ->
+      expand_each(item, query, mode, acc)
+    end)
+  end
+
+  defp expand(other, _query, _mode, _preloads, _assocs) do
+    raise ArgumentError, "`#{inspect(other)}` is not a valid preload expression."
+  end
```
josevalim, line 188:
```suggestion
  def expand(preloads, query) when is_atom(preloads) or is_list(preloads) do
    expand_each(preloads, query, :both, [], [])
  end

  def expand(other, _query) do
    raise ArgumentError, "`#{inspect(other)}` is not a valid preload expression."
  end
```

josevalim (approved):
This looks great to me. One minor comment and we are good!

Waiting for @greg-rychlewski's feedback/merging. :)

lib/ecto/query/builder/preload.ex
```diff
@@ -156,4 +157,126 @@ defmodule Ecto.Query.Builder.Preload do
...
+
+  defp expand(list, query, mode, preloads, assocs) when is_list(list) do
+    Enum.reduce(list, {preloads, assocs}, fn item, acc ->
+      expand_each(item, query, mode, acc)
+    end)
+  end
+
+  defp expand(other, _query, _mode, _preloads, _assocs) do
+    raise ArgumentError, "`#{inspect(other)}` is not a valid preload expression."
```
josevalim, line 187:
```suggestion
    raise ArgumentError, "`#{inspect(other)}` is not a valid preload expression, expected an atom or a list"
```

lib/ecto/query/builder/preload.ex
zachallaun: Just want to make sure I understand this suggestion correctly.

Right now, `expand_each` doesn't have a case that operates on a list -- it expects either an atom or a 2-element tuple. When `expand_each` needs to process a list, it calls back into `expand`. So basically, we need the `Enum.reduce(...)` clause currently handled by `defp expand(list, query, mode, preloads, assocs) when is_list(list)` somewhere.

My goal here was to emulate as much as possible the strategy that `escape` uses to process the AST. Moving the reduce into `expand_each` would, I think, mean that something like this would start working:

```elixir
# expressions interpolated at root are handled by `expand`
preload(^[:foo, [:bar, [:baz]])
```

whereas this wouldn't work:

```elixir
# expressions not interpolated at root are handled by `escape`
preload([:foo, [:bar, [:baz]])
```

josevalim: You are right, please ignore this suggestion then except for the change the error message. :)

greg-rychlewski:
Sorry if I missed it, but how would you do something like this dynamically:

```elixir
from p in "posts",
  join: c in "comments",
  join: l in "likes",
  preload: [likes: l, comments: c]
```

(without the named bindings)

greg-rychlewski:
Thank you. Maybe I'm being blind again, but if there isn't already an example of using positional bindings could you add one? When I read the docs I got the idea it only works with named bindings.

zachallaun:
Nope, you didn't miss anything! I'll add an example momentarily.

greg-rychlewski:
Thanks. It looks good to me. It's a very nice feature.

I'll just do one last scan in a couple hours when I'm off diaper duty and then merge.

zachallaun:
Added example usage of both and also chose to split a test into two in order to explicitly test the use of both named and positional bindings.

zachallaun:
Went over everything myself again and found a situation where the compile-time escape would raise an `ArgumentError` but the runtime expand would fail with a `FunctionClauseError`.

In seeing how `escape` handles this situation, it made me realize that it also _did_ handle arbitrary nesting, e.g.

```elixir
preload([:foo, [:bar, [:baz]]])
# is equivalent to
preload([:foo, :bar, :baz])
```

My latest commit fixes the `FunctionClauseError` and adds some additional test cases to ensure that the compile and runtime preload option is being handled the same.

However, this also caused me to realize another weird piece of existing behavior:

```elixir
preload("posts", [:foo, :bar, :baz]).preloads == [:baz, :bar, :foo]
```

This is because [`escape`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder/preload.ex#L48) reverses the recursively-accumulated preloads and assocs, and then [`build`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder/preload.ex#L143) reverses them back. It's not clear to me that this was intended.

The new runtime expansion maintains the same order that was passed in, i.e.

```elixir
preload("posts", ^[:foo, :bar, :baz]).preloads == [:foo, :bar, :baz]
```

My personal feeling is that behavior between compile and runtime should be the same. I'm also hesitant to change behavior that has seemingly lived happily in Ecto for at least 5 years. 😄

josevalim:
I think those changes are good. In the nested listed case, perhaps we can warn for now, since it is compilation behaviour, and flip to an error later. This way we get people to fix their code without breaking it?

zachallaun:
For reference, here's the existing behavior without this PR:

```elixir
preload("posts", [:foo, :bar, :baz]).preloads == [:baz, :bar, :foo]
preload("posts", ^[:foo, :bar, :baz]).preloads == [[:foo, :bar, :baz]]

preload("posts", [:foo, [:bar, [:baz]]]).preloads == [:baz, :bar, :foo]
preload("posts", ^[:foo, [:bar, [:baz]]]).preloads == [[:foo, [:bar, [:baz]]]]
```

And with the PR after my last commit:

```elixir
preload("posts", [:foo, :bar, :baz]).preloads == [:baz, :bar, :foo]
preload("posts", ^[:foo, :bar, :baz]).preloads == [:foo, :bar, :baz]

preload("posts", [:foo, [:bar, [:baz]]]).preloads == [:baz, :bar, :foo]
preload("posts", ^[:foo, [:bar, [:baz]]]).preloads == [:foo, :bar, :baz]
```

josevalim (approved):
LGTM!

The nested list behaviour can be unexpected but at the same time it is not ambiguous and it facilitates merging preloads. For example, I can write this code:

    preload: [:foo, ^additional]

And additional can be an atom or a list. Without the nesting, you would need to wrap and concatenate lists. So this is good to go IMO!

greg-rychlewski (approved):

zachallaun:
> The nested list behaviour can be unexpected but at the same time it is not ambiguous and it facilitates merging preloads.

Yep, agreed. And I'm not overly worried about the final order of elements being different between compile and runtime since they're different already on master. My main concern is just making sure that I'm not responsible for breaking existing code. 😄

josevalim:

zachallaun:
Awesome! Thank you both very much for being such diligent stewards.

zachallaun:
Hey friends! Zero pressure on this, but I have a library using this feature, but don’t want to cut a release of it while I’m depending on master. Any rough timeline on when a version might be cut that includes this PR? No worries if y’all don’t have anything planned yet — just thought I’d ask!

PR #4063, opened 2022-12-13
Add Ecto.Multi.exists?/4

greg-rychlewski:
This would help me use Multi.run in one less place =).

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -459,7 +459,7 @@ defmodule Ecto.Multi do
   end
 
   @doc """
-  Checks if there exists an entry matching the given query and stores it in the multi.
+  Checks if there exists an entry matching the given query and stores a boolean in the multi.
 
   Accepts the same arguments and options as `c:Ecto.Repo.exists?/2`.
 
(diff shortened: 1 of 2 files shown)


lib/ecto/multi.ex
```diff
@@ -458,6 +458,37 @@ defmodule Ecto.Multi do
...
+      |> MyApp.Repo.transaction()
+
+      Ecto.Multi.new()
+      |> Ecto.Multi.exists?(:post, fn _changes -> Post end)
+      |> MyApp.Repo.transaction()
+  """
+  @spec exists?(
+          t,
+          name,
+          queryable :: Ecto.Queryable.t | (any -> Ecto.Queryable.t),
```
v0idpwn, line 479: The function should accept the `changes` type. There is a type declared for this in Ecto.Multi: `fun(result_type)`

josevalim (approved):
LGTM except for @v0idpwn's comment. :)

lib/ecto/multi.ex
greg-rychlewski: Thanks I didn't notice that. I also fixed it for `one` and `all`

PR #4065, opened 2022-12-18
Create 3.9.3

greg-rychlewski:


josevalim:
@greg-rychlewski I wonder if it makes sense to backport several of those to v3.9 and do a new release there? Especially because I think the one with cast_params should go on the v3.9 branch. WDYT?

greg-rychlewski:
I think that makes sense. For this repo probably everything except for the empty value stuff can go into 3.9?

I'll also modify the ecto_sql PR to be 3.9

josevalim:
Yeah, I think everything mentioned in this PR is safe for v3.9 :)

greg-rychlewski:
Sorry not sure if I did this right. I cherry-picked all the commits to be back-ported and changed the target branch to v3.9

greg-rychlewski:
crap sorry, I thought I was pushing to here but I pushed directly to v3.9. I can revert them or you'd prefer to leave as is and close this PR?

josevalim:
Perfect. :heart:

PR #4066, opened 2022-12-19
Add support for materialized CTEs

maartenvanvliet:
Adds support for MATERIALIZED CTE's. Before Postgres 12 CTE's were materialized by default but from then on it was opt-in  with the MATERIALIZED keyword. In some cases this can be beneficial because then the CTE can act as a optimisation fence, which can result in better query plans sometimes.

This PR still needs some work, so I'm opening it to see if there's interest in this. If so, have I taken the right approach here?

Corresponding echo_sql PR https://github.com/elixir-ecto/ecto_sql/pull/466

See also https://www.postgresql.org/docs/current/queries-with.html#id-1.5.6.12.7 and https://paquier.xyz/postgresql-2/postgres-12-with-materialize/

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1216,6 +1216,8 @@ defmodule Ecto.Query do
   ## Options
 
     * `:as` - the CTE query itself or a fragment
+    * `:materialized` - (Postgres only) whether to materialize
+    the CTE (defaults to `false`)
 
   ## Recursive CTEs
 
@@ -1288,8 +1290,9 @@ defmodule Ecto.Query do
   invalid queries. If you are running into such scenarios, your best
   option is to use a fragment as your CTE.
   '''
-  defmacro with_cte(query, name, as: with_query) do
-    Builder.CTE.build(query, name, with_query, __CALLER__)
+  defmacro with_cte(query, name, [{:as, with_query} | opts]) do
+    materialized = Keyword.get(opts, :materialized, false)
+    Builder.CTE.build(query, name, with_query, materialized, __CALLER__)
   end
 
   @doc """
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1218,7 +1218,7 @@ defmodule Ecto.Query do
     * `:as` - the CTE query itself or a fragment
     * `:materialized` - a boolean indicating whether the CTE should
     be materialized. If blank, the database's default behaviour
-    will be used
+    will be used (only supported by Postgrex, for the built-in adapters)
 
   ## Recursive CTEs
 
@@ -1295,7 +1295,7 @@ defmodule Ecto.Query do
     with_query = opts[:as]
 
     if !with_query do
-      Builder.error! "`as` option must specified"
+      Builder.error! "`as` option must be specified"
     end
 
     Builder.CTE.build(query, name, with_query, opts[:materialized], __CALLER__)
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1216,7 +1216,7 @@ defmodule Ecto.Query do
   ## Options
 
     * `:as` - the CTE query itself or a fragment
-    * `:materialized - a boolean indicating whether the CTE should
+    * `:materialized` - a boolean indicating whether the CTE should
     be materialized. if blank, the database's default behaviour
     will be used
 
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1216,7 +1216,7 @@ defmodule Ecto.Query do
   ## Options
 
     * `:as` - the CTE query itself or a fragment
-    * `:materialize - a boolean indicating whether the CTE should
+    * `:materialized - a boolean indicating whether the CTE should
     be materialized. if blank, the database's default behaviour
     will be used
 
@@ -1295,7 +1295,7 @@ defmodule Ecto.Query do
     with_query = opts[:as]
 
     if !with_query do
-      Builder.error! "`as` keyword must specified"
+      Builder.error! "`as` option must specified"
     end
 
     Builder.CTE.build(query, name, with_query, opts[:materialized], __CALLER__)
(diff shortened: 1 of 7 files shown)


lib/ecto/query.ex
```diff
@@ -1234,8 +1236,9 @@ defmodule Ecto.Query do
   invalid queries. If you are running into such scenarios, your best
   option is to use a fragment as your CTE.
   '''
-  defmacro with_cte(query, name, as: with_query) do
-    Builder.CTE.build(query, name, with_query, __CALLER__)
+  defmacro with_cte(query, name, [{:as, with_query} | opts]) do
```
maartenvanvliet, line 1239: The `as` keyword now needs to be the first list item. Is that desired?

v0idpwn: Based in the docs from https://www.postgresql.org/docs/15/queries-with.html (section 7.8) the behaviour is a bit more complex than it may seem at first:
- Not specifying either `MATERIALIZED` or `NOT MATERIALIZED` will leave up to postgres whether to materialize
- Specifying overrides the postgres selection, which can bring performance improvements.
So we would need to support both options.

maartenvanvliet: Shall we code it with `:not_materialized | :materialized | nil` internally? `nil` would leave it up to postgres whether to materialize.

v0idpwn: I'm not sure, cc: @josevalim, @greg-rychlewski

greg-rychlewski: Can't it be like this

```elixir
with_cte(name, query, opts)

## options
  `:as` - the CTE query itself or a fragment
  `:materialize - a boolean indicating whether the CTE should be materialized. if blank, the database's default behaviour will be used
```  

Also I don't think there's a need to force the keyword list order. You can just make sure they specify `:as` and raise if not.

josevalim: :+1: for @greg-rychlewski's proposal!

lib/ecto/query/builder/cte.ex
```diff
@@ -35,9 +35,9 @@ defmodule Ecto.Query.Builder.CTE do
   If possible, it does all calculations at compile time to avoid
   runtime work.
   """
-  @spec build(Macro.t, Macro.t, Macro.t, Macro.Env.t) :: Macro.t
-  def build(query, name, cte, env) do
-    Builder.apply_query(query, __MODULE__, [escape(name, env), build_cte(name, cte, env)], env)
+  @spec build(Macro.t, Macro.t, Macro.t, boolean(), Macro.Env.t) :: Macro.t
+  def build(query, name, cte, materialized, env) do
+    Builder.apply_query(query, __MODULE__, [escape(name, env), build_cte(name, cte, env), materialized], env)
```
greg-rychlewski, line 40:
```suggestion
  def build(query, name, cte, materialized?, env) do
    Builder.apply_query(query, __MODULE__, [escape(name, env), build_cte(name, cte, env), materialized?], env)
```

I think the convention is to add a trailing question mark  but not 100% sure. I'm thinking of things like `changeset.valid?`. cc @josevalim

greg-rychlewski (commented):
Something is a bit off, some of the tests that should be there are not in your branch. Like this one in the planner test:

```elixir
   test "with CTEs" do
      cte_query = from(s in "schema", select: %{x1: selected_as(s.x, :integer), x2: s.x})

      %{with_ctes: %{queries: [{"schema_cte", inner_query}]}} =
        Comment
        |> with_cte("schema_cte", as: ^cte_query)
        |> normalize()

      assert [{:integer, _}, {:x2, _}] = inner_query.select.fields
    end
```
By any chance is your fork out of date? would you be able to rebase it?

test/ecto/query/builder/cte_test.exs
```diff
@@ -12,16 +12,24 @@ defmodule Ecto.Query.Builder.CTETest do
   test "fragment CTE" do
     query = "products" |> with_cte("categories", as: fragment("SELECT * FROM categories"))
 
-    assert [{"categories", %Ecto.Query.QueryExpr{expr: expr}}] = query.with_ctes.queries
+    assert [{"categories", false, %Ecto.Query.QueryExpr{expr: expr}}] = query.with_ctes.queries
     assert {:fragment, [], [raw: "SELECT * FROM categories"]} = expr
     refute query.with_ctes.recursive
   end
 
+  test "materialized CTE" do
```
greg-rychlewski, line 20: we should add a test for materialized = false as well once that change has been made

greg-rychlewski (approved):
Looks good to me thank you. I'll just let @v0idpwn or @josevalim approve as well. Then once this is merged we can point your ecto_sql PR back to it.

lib/ecto/query.ex
```diff
@@ -1216,6 +1216,9 @@ defmodule Ecto.Query do
   ## Options
 
     * `:as` - the CTE query itself or a fragment
+    * `:materialized` - a boolean indicating whether the CTE should
+    be materialized. If blank, the database's default behaviour
+    will be used
```
v0idpwn, line 1221: Not sure, but could be worth mentioning which adapters support this feature.

lib/ecto/query.ex
```diff
@@ -1288,8 +1291,14 @@ defmodule Ecto.Query do
   invalid queries. If you are running into such scenarios, your best
   option is to use a fragment as your CTE.
   '''
-  defmacro with_cte(query, name, as: with_query) do
-    Builder.CTE.build(query, name, with_query, __CALLER__)
+  defmacro with_cte(query, name, opts) do
+    with_query = opts[:as]
+
+    if !with_query do
+      Builder.error! "`as` option must specified"
```
v0idpwn, line 1298:
```suggestion
      Builder.error! "`:as` option must be specified"
```

v0idpwn (commented):
LGTM.

lib/ecto/query.ex
v0idpwn: With this fixed its approved from my side :)

maartenvanvliet:
Thx everyone! Happy holidays! 🎅

Issue #4067, opened 2022-12-20
Change with_cte representation to {map, query}


josevalim:
### Elixir version

Any

### Database and Version

Any

### Ecto Versions

master

### Database Adapter and Versions (postgrex, myxql, etc)

Postgres

### Current behavior

We changed the internal representation of with_cte to be:

    {string, materialized, query}

And that's a breaking change for adapters, so it should be removed from v3.9.

### Expected behavior

In order to avoid future breaking changes, I propose we change the format to be `{map, query}`, and we store both `:as` (and :materialized if any is given) in the map.

Thoughts?

greg-rychlewski:
+1

josevalim:
@maartenvanvliet, would you like to send a PR to amend this too? Thank you!

maartenvanvliet:
Sure!

PR #4068, opened 2022-12-20
Change with_cte representation to {map, query}

maartenvanvliet:
The first element of the tuple now is a map storing the name of the cte and whether it's materialized. 
  * I've named the map `opts` in places, not sure if there's a better name.
  * Not really happy with the `merge_queries` function, it's certainly less elegant than the List.keystore. Open to suggestions

I'll follow up with a PR in ecto_sql

Fixes https://github.com/elixir-ecto/ecto/issues/4067#issuecomment-1360056485

### DIFF
--- a/lib/ecto/query/builder/cte.ex
+++ b/lib/ecto/query/builder/cte.ex
@@ -94,24 +94,19 @@ defmodule Ecto.Query.Builder.CTE do
   end
 
   @doc false
-  def apply_cte(nil, name, with_query, materialized) do
-    %Ecto.Query.WithExpr{queries: [{%{name: name, materialized: materialized}, with_query}]}
+  def apply_cte(nil, name, with_query, materialized) when is_boolean(materialized) do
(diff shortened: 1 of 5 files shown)


lib/ecto/query/builder/cte.ex
```diff
@@ -95,10 +95,23 @@ defmodule Ecto.Query.Builder.CTE do
 
   @doc false
   def apply_cte(nil, name, with_query, materialized) do
-    %Ecto.Query.WithExpr{queries: [{name, materialized, with_query}]}
+    %Ecto.Query.WithExpr{queries: [{%{name: name, materialized: materialized}, with_query}]}
```
josevalim, line 98: Ah, I did not know the name is used for merging. My suggestion then is to keep it as:

```
{name, map, with_query}
```

WDYT?

And then we only include materialize in the map if it is a boolean?

josevalim:

PR #4071, opened 2022-12-21
Fix runtime nested preload expansion (fixes #4069)

### LINKED ISSUE #4069: `preload/2` ends up with error if nested bindings
### Elixir version

1.14.1

### Database and Version

PostgreSQL 14.5

### Ecto Versions

v3.9.3

postgrex 0.16.5

### Current behavior

When i do preload with `^` operator and nested bindings Ecto checking `field2` association on `Schema2` which doesn't have one, looks like in case with `^` we have wrong preload expanding
```elixir
Schema1
|> preload(^[:field1, :field2, schema2: [:field1]])
|> Repo.all()

** (ArgumentError) schema App.Schemas.Schema2 does not have association or embed :field2
    (elixir 1.14.1) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto 3.9.3) lib/ecto/repo/queryable.ex:235: Ecto.Repo.Queryable.execute/4
    (ecto 3.9.3) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
    iex:27: (file)
```

Actually expand has a bug

```elixir
iex(11)> Ecto.Query.Builder.Preload.expand([:field1, :field2, schema2: [:field1]], [])        
{[{:schema2, [:field1, :field2, :field1]}, :field2, :field1], []}
```

### Expected behavior

Should work like without `^`

zachallaun:
The call to `expand/5` on line 269 incorrectly passed the currently accumulated preloads and associations instead of resetting the accumulators, causing already accumulated values to be applied to inner preloads.

* Fix bug reported in #4069 
* Add doctests for `expand/2`, similar to those present for `escape/2`
* Reverse accumulated lists in `expand/5` instead of `preload!/2`; this simplifies some code elsewhere and is more consistent with `escape`

### DIFF
--- a/lib/ecto/query/builder/preload.ex
+++ b/lib/ecto/query/builder/preload.ex
@@ -163,11 +163,42 @@ defmodule Ecto.Query.Builder.Preload do
   """
   def preload!(query, preload) do
     {preloads, assocs} = expand(preload, query)
-    apply(query, Enum.reverse(preloads), Enum.reverse(assocs))
+    apply(query, preloads, assocs)
   end
 
   @doc """
   Expands preloads at runtime.
+
+  ## Examples
+
+      iex> expand(:foo, [])
+      {[:foo], []}
+
+      iex> expand([foo: :bar], [])
+      {[foo: :bar], []}
+
+      iex> expand([:foo, :bar], [])
+      {[:foo, :bar], []}
+
+      iex> expand([foo: [:bar, bar: :bat]], [])
+      {[foo: [:bar, bar: :bat]], []}
+
+      iex> expand([:a, :b, c: [:d]], [])
+      {[:a, :b, c: [:d]], []}
+
+      iex> expand([foo: ["external"]], [])
+      ** (ArgumentError) `"external"` is not a valid preload expression, expected an atom or a list.
+
+      iex> require Ecto.Query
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder/preload.ex
```diff
@@ -226,14 +260,14 @@ defmodule Ecto.Query.Builder.Preload do
...
-    {[{key, {query_or_fun, Enum.reverse(inner_preloads)}}|preloads], assocs}
+    {[{key, {query_or_fun, inner_preloads}}|preloads], assocs}
   end
 
   defp expand_each({key, list}, query, _mode, {preloads, assocs}) when is_list(list) do
     assert_key!(key)
 
-    {inner_preloads, []} = expand(list, query, :preload, preloads, assocs)
-    {[{key, Enum.reverse(inner_preloads)}|preloads], assocs}
+    {inner_preloads, []} = expand(list, query, :preload, [], [])
```
zachallaun, line 269: This is the "important part".

greg-rychlewski:
Let's add a test to prevent regression :)

zachallaun:
@greg-rychlewski The doctest on line 186 tests for this case (I wrote the tests prior to the fix and had it failing with the reported behavior). I'm happy to add a test to `preload_test.exs` as well!

zachallaun:
Amended to include a test case using `preload`.

josevalim:

PR #4074, opened 2023-01-07
Warn on missing :on in joins

sabiwara:
Add a warning when `:on` is missing and defaults to `true`, as proposed [here](https://groups.google.com/g/elixir-ecto/c/QgcepU4rSPU/m/pV3yXtySCAAJ):

> I think emitting a warning if the `on` option is missing and there is no assoc is a good compromise.

### DIFF
--- a/lib/ecto/query/builder/join.ex
+++ b/lib/ecto/query/builder/join.ex
@@ -185,7 +185,7 @@ defmodule Ecto.Query.Builder.Join do
       hints: hints
     ]
 
-    on = ensure_on(on, join, env)
+    on = ensure_on(on, join_assoc, join_qual, join_source, env)
     query = build_on(on, join, as, query, binding, count_bind, env)
     {query, binding, next_bind}
   end
@@ -231,23 +231,23 @@ defmodule Ecto.Query.Builder.Join do
     end
   end
 
-  defp ensure_on(nil, join, env) do
-    unless join[:assoc] do
-      maybe_source =
-        with {source, alias} <- join[:source],
-          source when source != nil <- source || alias do
-          " on #{inspect(source)}"
-        else
-          _ -> ""
-        end
+  defp ensure_on(on, _assoc, _qual, _source, _env) when on != nil, do: on
 
-      IO.warn("Missing `:on` in join#{maybe_source}, defaulting to `on: true`.", env)
-    end
+  defp ensure_on(nil, _assoc = nil, qual, source, env) when qual not in [:cross, :cross_lateral] do
+    maybe_source =
+      with {source, alias} <- source,
+        source when source != nil <- source || alias do
+        " on #{inspect(source)}"
+      else
+        _ -> ""
+      end
+
+    IO.warn("Missing `:on` in join#{maybe_source}, defaulting to `on: true`.", env)
 
     true
   end
 
-  defp ensure_on(on, _join, _env), do: on
+  defp ensure_on(nil, _assoc, _qual, _source, _env), do: true
 
   @doc """
   Applies the join expression to the query.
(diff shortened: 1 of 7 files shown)


lib/ecto/query/builder/join.ex
```diff
@@ -229,6 +231,26 @@ defmodule Ecto.Query.Builder.Join do
...
+        with {source, alias} <- join[:source],
+          source when source != nil <- source || alias do
+          " on #{inspect(source)}"
+        else
+          _ -> ""
+        end
+
+      Logger.warning(
+        "#{join[:file]}:#{join[:line]} Missing `:on` in join#{maybe_source}, defaulting to `on: true`."
+      )
```
sabiwara, line 246: I struggled a bit to come up with a useful warning with the proper line number, else it would be quite un-helpful:

Screenshot 2023-01-07 at 9 31 58

With `IO.warn` (stacktrace isn't helpful):
Screenshot 2023-01-07 at 9 55 50

greg-rychlewski:
I think documenting that it defaults to `on: true` would also be helpful. 

Also not sure I understand why `on: true` had to be added to all those tests.

sabiwara:
> I think documenting that it defaults to `on: true` would also be helpful.

Got it, will add it 👍

> Also not sure I understand why `on: true` had to be added to all those tests.

This was to avoid having a wall of warnings when running tests, but I can revert test files if you prefer?
I should have asked before fixing all of them, this was most of the work to be honest 😂

lib/ecto/query/builder/join.ex
```diff
@@ -229,6 +231,26 @@ defmodule Ecto.Query.Builder.Join do
...
+        with {source, alias} <- join[:source],
+          source when source != nil <- source || alias do
+          " on #{inspect(source)}"
+        else
+          _ -> ""
+        end
+
+      Logger.warn(
+        "#{join[:file]}:#{join[:line]} Missing `:on` in join#{maybe_source}, defaulting to `on: true`."
+      )
```
josevalim, line 246: Please do `IO.warn` instead with passing the caller (aka __CALLER__) to retrieve the stacktrace. Using IO.warn means compilation will halt if using warnings as errors. :)

greg-rychlewski:
>This was to avoid having a wall of warnings when running tests, but I can revert test files if you prefer?

Ah I thought the tests were run on `:error` level. Sorry about that.

lib/ecto/query/builder/join.ex
```diff
@@ -229,6 +231,26 @@ defmodule Ecto.Query.Builder.Join do
     end
   end
 
+  defp ensure_on(nil, join) do
```
greg-rychlewski, line 234: We don't have to warn on cross joins so can do something like this the below. Sorry being a bit lazy so not completely sure if this is the right thing to warn on.

`defp ensure_on(nil, %{qual: qual}) when qual in [:cross_join, :cross_lateral_join], do: true`

sabiwara: Great point I completely missed that 🙇
Fixed in https://github.com/elixir-ecto/ecto/pull/4074/commits/de747ba0a890d6f4f64bb9a8d67484366cf5c2d8

lib/ecto/query/builder/join.ex
sabiwara: Thanks, this is perfect, exactly what I was looking for 💜
Fixed in https://github.com/elixir-ecto/ecto/pull/4074/commits/fe18c3e28e3668de79918a6777e40e8bdd077ba0, but I needed to get the stacktrace manually to support elixir <1.14 https://github.com/elixir-ecto/ecto/pull/4074/commits/e52d35840665a1dc3aa4507ee3a83d97371bcb01

josevalim (approved):

greg-rychlewski (approved):

greg-rychlewski:
Thank you!

PR #4075, opened 2023-01-07
Fix wrong typespec

sabiwara:
I noticed that this typespec was out of date, since the acc is now a map, not just a list.


josevalim:

PR #4077, opened 2023-01-12
Update `Ecto.Query.exclude/2` to be able to exclude `:update`

vereis:
Per [my proposal](https://groups.google.com/g/elixir-ecto/c/z59zF3I_r68/m/oWguRKoAAgAJ), I extended the ability of `Ecto.Query.exclude/2` to be able to exclude the `select` keyword in the form:

```elixir
query = from x in MySchema, update: [set: [name: "test"]]
Ecto.Query.exclude(query, :update)
```


greg-rychlewski:
Thank you.

PR #4079, opened 2023-01-18
Add `get_assoc`/`get_embed`

LostKobrakai:
I've recently written a [blog post](https://kobrakai.de/kolumne/one-to-many-liveview-form) about a common one-to-many form setup using LV. There's a lot of footguns to consider, but one common pain point mentioned, which I had also talked about with a few people on slack doing similar thing, is is modifying assocs/embeds (usually the *_many variant) on an existing changeset. This is possible with existing APIs (as seen in the blogpost), but it's more complex than it needs to be.

- There's no function, which returns a list of changesets of the relation and transparently falls back to the unchanged field data if there are no changes (yet)
- `get_field` returns a list of structs with changes applied, but using `put_assoc/embed` "forgets" those changes
- The same happens when editing the structs returned by `get_field` and putting them back using `put_assoc/embed`

The code in the PR is a proposal to add a `update_assoc/embed` API, which basically consolidates gathering the "current state of the relation" as changesets, passes it to a user prodived function to allow for updates, and then puts the result back into the changeset. Passing data normalized to changesets should make it clear that adding additional changes is supposed to happen via the `Ecto.Changeset` APIs and not by applying existing changes and editing the returned struct.

I'm aware that LV is meant to have its form integration updated, but this feels useful even beyond the immediate LV forms usecase.

Edit:
One such usecase could be "assign positional values" to order assocs by, when the input didn't include such and relied on the order of the list passed into the changeset.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1161,9 +1161,10 @@ defmodule Ecto.Changeset do
   then falls back on the data, finally returning `default` if
   no value is available.
 
-  For relations consider using `get_assoc/3`/`get_embed/3`. This
-  function will return the changeset data with all changes applied,
-  which might not be what you're looking for.
+  For associations and embeds, this function returns the changeset
+  data with all changes applied. Use  `get_assoc/3`/`get_embed/3`
+  if you want to retrieve the relations as changesets or if you want
+  more fine-grained control.
 
       iex> post = %Post{title: "A title", body: "My body is a cage"}
       iex> changeset = change(post, %{title: "A new title"})
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1570,6 +1570,61 @@ defmodule Ecto.Changeset do
   end
 
   @doc """
+  Update the existing association entry or entries as a change in the changeset.
+
+  For association with cardinality one the passed function will receive a changeset.
+  For association with many entries the passed function will receive a list of changesets.
+
+  Returned values or lists of values need to match what is described in
+  `put_assoc/4`. If the given value is not any of values listed there,
+  it will raise. Commonly the returned value(s) are expected to be changesets.
+
+  Lists of entries can be reordered, e.g. for forms rendered based on the parent
+  changeset. But persisting associations does not retain order unless maintained
+  separately – e.g. using `prepare_changes/2`.
+  """
+  def update_assoc(%Changeset{} = changeset, name, default, fun) do
+    update_relation(:assoc, changeset, name, default, fun)
+  end
+
+  @doc """
+  Update the existing embed entry or entries as a change in the changeset.
+
+  For embeds with cardinality one the passed function will receive a changeset.
+  For embeds with many entries the passed function will receive a list of changesets.
+
+  Returned values or lists of values need to match what is described in
+  `put_assoc/4`. If the given value is not any of values listed there,
+  it will raise. Commonly the returned value(s) are expected to be changesets.
+
+  Lists of entries can be reordered.
+  """
+  def update_embed(%Changeset{} = changeset, name, default, fun) do
+    update_relation(:embed, changeset, name, default, fun)
+  end
+
+  defp update_relation(_tag, %{types: nil}, _name, _default, _fun) do
+    raise ArgumentError, "changeset does not have types information"
+  end
+
+  defp update_relation(tag, changeset, name, default, fun) when is_function(fun, 1) do
+    %{data: data, types: types, changes: changes, errors: errors, valid?: valid?} = changeset
+    relation = relation!(:put, tag, name, Map.get(types, name))
+    existing = get_change(changeset, name) || get_field(changeset, name, default)
+
+    value =
+      case existing do
+        list when is_list(list) -> list |> Enum.map(&change/1) |> fun.()
+        item ->  item |> change() |> fun.()
+      end
+
+    {changes, errors, valid?} =
+      put_change(data, changes, errors, valid?, name, value, {tag, relation})
+
+    %{changeset | changes: changes, errors: errors, valid?: valid?}
+  end
+
+  @doc """
   Forces a change on the given `key` with `value`.
 
   If the change is already present, it is overridden with
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1242,8 +1242,8 @@ defmodule Ecto.Changeset do
   @doc """
   Gets the embedded entry or entries from changes or from the data.
 
-  Returned data is either normalized to be changesets or structs with any changes
-  applied to them. The `:struct` option returns the same value as `get_field/2`.
+  Returned data is normalized to changesets by default. Pass the `:struct`
+  flag to retrieve the data as structs with changes applied, similar to `get_field/2`.
 
   ## Examples
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1281,9 +1281,14 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "changeset does not have types information"
   end
 
-  defp get_relation(tag, %{types: types} = changeset, name) do
+  defp get_relation(tag, %{changes: changes, data: data, types: types}, name) do
     _ = relation!(:get, tag, name, Map.get(types, name))
-    existing = get_change(changeset, name) || get_field(changeset, name)
+
+    existing =
+      case Map.fetch(changes, name) do
+        {:ok, value} -> value
+        :error -> Relation.load!(data, Map.fetch!(data, name))
+      end
 
     case existing do
       nil -> nil
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1285,9 +1285,9 @@ defmodule Ecto.Changeset do
     _ = relation!(:get, tag, name, Map.get(types, name))
 
     existing =
-      case Map.fetch(changes, name) do
-        {:ok, value} -> value
-        :error -> Relation.load!(data, Map.fetch!(data, name))
+      case changes do
+        %{^name => value} -> value
+        %{} -> Relation.load!(data, Map.fetch!(data, name))
       end
 
     case existing do
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1208,8 +1208,8 @@ defmodule Ecto.Changeset do
   @doc """
   Gets the association entry or entries from changes or from the data.
 
-  Returned data is either normalized to be changesets or structs with any changes
-  applied to them. The `:struct` option returns the same value as `get_field/2`.
+  Returned data is normalized to changesets by default. Pass the `:struct`
+  flag to retrieve the data as structs with changes applied, similar to `get_field/2`.
 
   ## Examples
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1215,7 +1215,7 @@ defmodule Ecto.Changeset do
 
       iex> %Author{posts: [%Post{id: 1, title: "hello"}]}
       ...> |> change()
-      ...> |> get_assoc(:posts, :changeset)
+      ...> |> get_assoc(:posts)
       [%Ecto.Changeset{data: %Post{id: 1, title: "hello"}, changes: %{}}]
 
       iex> %Author{posts: [%Post{id: 1, title: "hello"}]}
@@ -1231,6 +1231,8 @@ defmodule Ecto.Changeset do
       [%Post{id: 1, title: "world"}]
 
   """
+  def get_assoc(changeset, name, as \\ :changeset)
+
   def get_assoc(%Changeset{} = changeset, name, :struct) do
     get_field(changeset, name)
   end
@@ -1249,7 +1251,7 @@ defmodule Ecto.Changeset do
 
       iex> %Post{comments: [%Comment{id: 1, body: "hello"}]}
       ...> |> change()
-      ...> |> get_embed(:comments, :changeset)
+      ...> |> get_embed(:comments)
       [%Ecto.Changeset{data: %Comment{id: 1, body: "hello"}, changes: %{}}]
 
       iex> %Post{comments: [%Comment{id: 1, body: "hello"}]}
@@ -1265,6 +1267,8 @@ defmodule Ecto.Changeset do
       [%Comment{id: 1, body: "world"}]
 
   """
+  def get_embed(changeset, name, as \\ :changeset)
+
   def get_embed(%Changeset{} = changeset, name, :struct) do
     get_field(changeset, name)
   end
@@ -3103,7 +3107,7 @@ defmodule Ecto.Changeset do
   @spec assoc_constraint(t, atom, Keyword.t) :: t
   def assoc_constraint(changeset, assoc, opts \\ []) do
     constraint = opts[:name] ||
-      case get_assoc(changeset, assoc) do
+      case get_assoc_type(changeset, assoc) do
         %Ecto.Association.BelongsTo{owner_key: owner_key} ->
           "#{get_source(changeset)}_#{owner_key}_fkey"
         other ->
@@ -3159,7 +3163,7 @@ defmodule Ecto.Changeset do
   @spec no_assoc_constraint(t, atom, Keyword.t) :: t
   def no_assoc_constraint(changeset, assoc, opts \\ []) do
     {constraint, message} =
-      case get_assoc(changeset, assoc) do
+      case get_assoc_type(changeset, assoc) do
         %Ecto.Association.Has{cardinality: cardinality,
                               related_key: related_key, related: related} ->
           {opts[:name] || "#{related.__schema__(:source)}_#{related_key}_fkey",
@@ -3238,12 +3242,10 @@ defmodule Ecto.Changeset do
     raise ArgumentError, "cannot add constraint because a changeset was not supplied, got: #{inspect item}"
   end
 
-  defp get_assoc(%{types: types}, assoc) do
+  defp get_assoc_type(%{types: types}, assoc) do
     case Map.fetch(types, assoc) do
-      {:ok, {:assoc, association}} ->
-        association
-      _ ->
-        raise_invalid_assoc(types, assoc)
+      {:ok, {:assoc, association}} -> association
+      _ -> raise_invalid_assoc(types, assoc)
     end
   end
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1161,8 +1161,9 @@ defmodule Ecto.Changeset do
   then falls back on the data, finally returning `default` if
   no value is available.
 
-  For relations, these functions will return the changeset data
-  with changes applied. To retrieve raw changesets, please use `get_change/3`.
+  For relations consider using `get_assoc/3`/`get_embed/3`. This
+  function will return the changeset data with all changes applied,
+  which might not be what you're looking for.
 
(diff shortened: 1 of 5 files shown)


josevalim:
:+1: from me.

greg-rychlewski:
Still trying to fully comprehend the use case. But if it's meant to update existing entries should it raise or return an error instead of creating new data from the default?

lib/ecto/changeset.ex
```diff
@@ -1569,6 +1569,61 @@ defmodule Ecto.Changeset do
...
+  end
+
+  defp update_relation(_tag, %{types: nil}, _name, _default, _fun) do
+    raise ArgumentError, "changeset does not have types information"
+  end
+
+  defp update_relation(tag, changeset, name, default, fun) when is_function(fun, 1) do
+    %{data: data, types: types, changes: changes, errors: errors, valid?: valid?} = changeset
+    relation = relation!(:put, tag, name, Map.get(types, name))
+    existing = get_change(changeset, name) || get_field(changeset, name, default)
```
v0idpwn, line 1613: `Map.update/4` and `Keyword.update/4` don't pass the `default` into the update function. I see the use-case for this here, but is it adequate to have different semantics under the same name?

greg-rychlewski:
After reading how `get_field/3` works, I shifted my questions/concerns a bit:

1. `get_field/3` only uses `default` if the key is not in the changes or the data. If it's in the data, even `nil` or `%NotLoaded{}` the default is ignored. I think there is no such thing as a schemaless embed or association, so `default` will never be used?
2. Assuming (1) is correct, then I think a few other questions arise:
   - What to do with associations that are not loaded? If they are not loaded I think you want to stop right there because you are acting on incomplete information. If you use `get_field/3` on it then it will be converted to `nil` or `[]` and you have the potential to wipe your data without knowing.
   - What to do with `nil` and `[]`?  My own personal preference is to stop right there and say there is no data to update. But I'm far removed from the use case described in the issue. If there is a big need to create new data when it doesn't exist, probably it should be opt-in?  I don't think there is a need to force people to create new data when doing updates if they don't want to.

LostKobrakai:
`Relation.load!` prevents any issues with non-preloaded associations (it raises if things are not preloaded) and it's called in the fetching as well as in the applying changes codepaths. So accidentally wiping data shouldn't happen. Also nothing in there is "new code" besides allowing for that callback. The rest is just calling functionality already existing on the module.

> What to do with nil and []? My own personal preference is to stop right there and say there is no data to update. But I'm far removed from the use case described in the issue.

The issue here is that "update" doesn't mean update existing records, but update the existing set of assocs/embeds. E.g. someone might just want to add a new record to the end of the list of assocs. As long as the field is an association/embed field there is data to update.

Essentially it maps to this:
```elixir
data = %{items: []}
Map.update!(data, :items, fn cur -> [:new | cur] end)
```

As for the question around the default: I'll have to look at the possible values for embeds/assocs and if a default makes sense. I know that embeds can at least become `nil` atm (see https://groups.google.com/g/elixir-ecto/c/21He5yAPTRA), but maybe that should just be passed into the callback as is.

lib/ecto/changeset.ex
josevalim: Perhaps it makes sense to remove default and the user can either check for nil or an empty list?

LostKobrakai: Yeah, that might make the most sense.

greg-rychlewski:
>Relation.load! prevents any issues with non-preloaded associations 

Do you mean it's not possible for someone to shoot themselves in the foot hitting this condition:

```elixir
def load!(%{__meta__: %{state: :built}}, %NotLoaded{__cardinality__: cardinality}) do
   cardinality_to_empty(cardinality)
 end
```

Honestly not sure. Was just thinking a "you shouldn't have done that" moment probably shouldn't erase data.

LostKobrakai:
> In this case, I'm not sure why we would expect or want anything but a no-op or error if the association isn't preloaded.

`get_field` will raise if associations are not preloaded. If you don't get to that callpath then there's already changes for the association, which to my knowledge are also only possible if data was preloaded (a.k.a. using `cast_assoc`/`put_assoc`). Yes having the `default` option might have been the wrong call, but if you find a way to delete data unknowingly then this would affect any existing attempt to manipulate associations as well. 

> If you are building a brand new struct and you know it's not in the database, cast_assoc/put_assoc exist for this purpose.

The issue this is new function is trying to solve is making the editing of associations(/embeds) generic over all possible cases. When the usecase is "add an additional association record" one shouldn't need to care much how many rows are there currently (granted the assocs are preloaded). Also one shouldn't need to care if there are changes on those records or not – any tracked changes should be kept around if they're not manipulated in the callback. Both of those are really tough to get right without studing the ecto docs for minor details.

lib/ecto/changeset.ex
josevalim: Shouldn't `get_field` be enough as it looks on changes first and then falls back to the struct?

```suggestion
    existing = get_field(changeset, name)
```

josevalim:
The current iteration of this function is a wrapper around `put_change(fun.(get_field))` while also ensuring we always get changesets back. So I think it does abstract boilerplate but, if we are worried about its updating bits, we could also introduce a `get_assoc` function, which does the getting+ensuring, but without doing any writes. This way, someone can do:

    one_or_many = get_assoc(...)
    put_assoc(transform(one_or_many))

The benefit of this approach is that we are not introducing a new verb for modifying associations.

However, we will have to explain what is the difference between `get_field`, `get_change`, and `get_assoc`. AFAIK, `get_field` returns structs or changesets, `get_change` returns changesets (if any), and `get_assoc` would return changesets. That said, I am not sure if `get_assocs` is the best name.

lib/ecto/changeset.ex
LostKobrakai: It isn't and that's exactly one of the footguns I want to get rid of by adding this. `get_field` applies changes for relations and doesn't return changesets if there are changes for those related records. 

Consider this:
```elixir
    base_changeset =
      Changeset.change(%Author{posts: [%Post{id: 1, title: "hello"}]})
      |> Changeset.cast(%{posts: [%{id: 1, title: "world"}]}, [])
      |> Changeset.cast_assoc(:posts)

    existing_posts = Changeset.get_field(base_changeset, :posts)
    changeset = Changeset.put_assoc(base_changeset, :posts, existing_posts)

    assert Map.has_key?(changeset.changes, :posts)
```

This test `fails` for actually two intertwined reasons. `get_field` returns a list of structs (so no way to know what is a change or the base data anymore) and `put_assoc` also treats `structs` as input as a special case of replacing existing data, not updating it as well. This breaks even before attempting to modify the `existing_posts` in between.

LostKobrakai:
A `get_assoc` would be fine with me as well. Really the goal for this is a good way to get to a clean and normalized list of existing associations, so modifying it is neither full of footguns nor full of normalization or conditionals around multiple data representations.

josevalim:
So I would prefer `get_assoc`, we just need to find a better name?

The idea is to add a note to `get_field` to say that it normalizes associations to structs. To get associations as changesets, call `get_assoc`. My only concern is that `get_assoc` does not reflect the "as changesets" bits and `get_assoc_as_changes` is too long. Another option is `get_assoc(changeset, name, as \\ :changesets | :structs)`.

LostKobrakai:
`get_field` already has a small note that it applies changes on relations, but yeah it could be more obvious and pointing to an alternative would be very useful.

I like the idea of providing the expected return format, but for `:struct` there's imo a bit of ambiguity of "which struct". Is it the base data or is it the base data with changes applied? How about making `get_change` accept an option around applying changes on relations. Default to the current behaviour and add an option to have it return changes as a list of changesets?

josevalim:
I think it is fine to say the :struct is with the changes applied because if you want the raw one you go through the data, no changeset needed.

options are tricky for get_field, because it already has a default as optional argument. So I am would officially propose `get_assoc(changeset, name, as \\ :changesets | :structs)`.

thoughts?

LostKobrakai:
> thoughts?

I figured this would be an issue, so I had moved to implementing `get_assoc/3` already.

lib/ecto/changeset.ex
```diff
@@ -1161,8 +1161,9 @@ defmodule Ecto.Changeset do
   then falls back on the data, finally returning `default` if
   no value is available.
 
-  For relations, these functions will return the changeset data
-  with changes applied. To retrieve raw changesets, please use `get_change/3`.
+  For relations consider using `get_assoc/3`/`get_embed/3`. This
+  function will return the changeset data with all changes applied,
+  which might not be what you're looking for.
```
josevalim, line 1167:
```suggestion
  For associations and embeds, this function returns the changeset
  data with all changes applied. Use  `get_assoc/3`/`get_embed/3`
  if you want to retrieve the relations as changesets or if you want
  more fine-grained control.

```

lib/ecto/changeset.ex
```diff
@@ -1203,6 +1204,89 @@ defmodule Ecto.Changeset do
     end
   end
 
+  @doc """
+  Gets the association entry or entries from changes or from the data.
+
+  Returned data is either normalized to be changesets or structs with any changes
+  applied to them. The `:struct` option returns the same value as `get_field/2`.
```
josevalim, line 1211:
```suggestion
  Returned data is normalized to changesets by default. Pass the `:struct`
  flag to retrieve the data as structs with changes applied, similar to `get_field/2`.
```

lib/ecto/changeset.ex
```diff
@@ -1203,6 +1204,89 @@ defmodule Ecto.Changeset do
...
+
+  def get_assoc(%Changeset{} = changeset, name, :changeset) do
+    get_relation(:assoc, changeset, name)
+  end
+
+  @doc """
+  Gets the embedded entry or entries from changes or from the data.
+
+  Returned data is either normalized to be changesets or structs with any changes
+  applied to them. The `:struct` option returns the same value as `get_field/2`.
```
josevalim, line 1245:
```suggestion
  Returned data is normalized to changesets by default. Pass the `:struct`
  flag to retrieve the data as structs with changes applied, similar to `get_field/2`.
```

josevalim (approved):

greg-rychlewski (approved):

greg-rychlewski:
@LostKobrakai I feel like I repeated this a few times but maybe didn't say it properly this is only true if the struct is loaded

>get_field will raise if associations are not preloaded.

If it's built it will populate empty assocs. Which can happen if you have the parent struct id and don't want to load the entire record.

greg-rychlewski:
So I still think there is a bit of an issue with this new function but don't want to stop the merge if this works for everyone else. This is the flow I think can cause a problem:

1. You receive the parent struct id and don't care about anything but this association. 
2. You create a new struct with just that id and preload the assoc
3. You call get_assoc

If step 2 is missed then step 3 will tell you the assoc is `nil` or `[]` when that might not be the case.

josevalim:
@greg-rychlewski isn't your comment true though basically for all changesets opereations, including accessing the struct fields themselves?

greg-rychlewski:
> including accessing the struct fields themselves

I might misunderstand, but let's say there is a `User` schema with association `:comments`. Then `%User{}.comments` will return a `NotLoaded` struct. If you meant accessing the field through `get_field`, then yes it will transform to an empty association.

This stood out to me because the original intention of the PR was to immediately perform modifications based on the existing associations. In this case I wouldn't want anything normalized to empty. I would want to know if it wasn't preloaded. So personally to me it feels safer to access the field directly instead of using `get_field`.

josevalim:
Ah, good point with `struct` actually being correct. But `get_field`/`get_change`/... would all initialize it to empty, right? So I think regardless of that, `get_assoc` would be consistent. We could perhaps introduce `get_assoc!` to assume it is loaded?

greg-rychlewski:
>get_field/get_change/... would all initialize it to empty, right? 

Yeah exactly. This part I was actually surprised about. But I'm certain there is a reason for it. I never use `get_field` on assocs personally.

>We could perhaps introduce get_assoc! to assume it is loaded?

This feels like a really good compromise to me. And maybe a note in the non bang function about this normalization. Unless it is already well understood by most.

greg-rychlewski:
Actually sorry I spoke to soon. `get_change` is just taking the value

```elixir
def get_change(%Changeset{changes: changes} = _changeset, key, default \\ nil)
      when is_atom(key) do
    Map.get(changes, key, default)
  end
```

lib/ecto/changeset.ex
```diff
@@ -1161,8 +1161,10 @@ defmodule Ecto.Changeset do
   then falls back on the data, finally returning `default` if
   no value is available.
 
-  For relations, these functions will return the changeset data
-  with changes applied. To retrieve raw changesets, please use `get_change/3`.
+  For associations and embeds, this function returns the changeset
+  data with all changes applied. Use  `get_assoc/3`/`get_embed/3`
+  if you want to retrieve the relations as changesets or if you want
+  more fine-grained control.
```
josevalim, line 1167:
```suggestion
  data with all changes applied. If the association has not been loaded
  yet, we assume it is empty. Use `get_assoc/3`/`get_embed/3`
  if you want to retrieve the relations as changesets or if you want
  more fine-grained control.
```

lib/ecto/changeset.ex
```diff
@@ -1203,6 +1205,93 @@ defmodule Ecto.Changeset do
     end
   end
 
+  @doc """
+  Gets the association entry or entries from changes or from the data.
+
+  Returned data is normalized to changesets by default. Pass the `:struct`
+  flag to retrieve the data as structs with changes applied, similar to `get_field/2`.
```
josevalim, line 1212:
```suggestion
  Returned data is normalized to changesets by default. If the association has not
  been loaded yet, it assumes it is empty. Pass the `:struct` flag to retrieve the
  data as structs with changes applied, similar to `get_field/2`.
```

lib/ecto/changeset.ex
```diff
@@ -1203,6 +1205,93 @@ defmodule Ecto.Changeset do
...
+
+  def get_assoc(%Changeset{} = changeset, name, :changeset) do
+    get_relation(:assoc, changeset, name)
+  end
+
+  @doc """
+  Gets the embedded entry or entries from changes or from the data.
+
+  Returned data is normalized to changesets by default. Pass the `:struct`
+  flag to retrieve the data as structs with changes applied, similar to `get_field/2`.
```
josevalim, line 1248:
```suggestion
  Returned data is normalized to changesets by default. If the embed has not
  been loaded yet, it assumes it is empty. Pass the `:struct` flag to retrieve the
  data as structs with changes applied, similar to `get_field/2`.
```

josevalim:
I sent a PR to update the docs and make the behaviour clear. We can add `get_assoc!` either in this PR or another one, your call @LostKobrakai.

lib/ecto/changeset.ex
```diff
@@ -1203,6 +1205,93 @@ defmodule Ecto.Changeset do
...
+    get_relation(:embed, changeset, name)
+  end
+
+  defp get_relation(_tag, %{types: nil}, _name) do
+    raise ArgumentError, "changeset does not have types information"
+  end
+
+  defp get_relation(tag, %{types: types} = changeset, name) do
+    _ = relation!(:get, tag, name, Map.get(types, name))
+    existing = get_change(changeset, name) || get_field(changeset, name)
```
greg-rychlewski, line 1286:
```suggestion
    existing = get_field(changeset, name)
```

I feel like I must be missing something something because the part I crossed off is really confusing me: 

- `get_field/3` will already check the changes map
-  if `%{assoc: nil}` is in the change map it will effectively be ignored until `get_field`

josevalim: The issue is that `get_field` will apply the changes, and we don't want that here. But perhaps it is worth just inlining the implementation `get_field` here. It is basically two nested cases anyway.

greg-rychlewski: oh I get it. it would be less confusing me to inline but won't complain either way

lib/ecto/changeset.ex
greg-rychlewski:
```suggestion
  If the changeset's data has not been loaded, we assume any association that has not been loaded
  yet is empty
```

It's slightly different. If the parent data is loaded it will raise if the association is not loaded. If the parent data is built it will assume a not loaded association is empty. And then the reason the latter might be an issue is because you can preload onto built structs as long as they have the primary key.

LostKobrakai: `get_assoc` here only highlights what is already working the same with `get_field` or `put_assoc`. `get_assoc` by itself doesn't really "populate assocs", only a later `put_assoc` will. And `put_assoc` has always allowed you to put associations on a `:built` schema struct without an explicit preload needed. So to me it sounds like what you're describing here as "not ok" is actually the intended functionality. Both `get_field` and `put_assoc` don't care if associations are preloaded if the struct hasn't been persisted yet. Nothing about that changed with `get_assoc`.

josevalim: > %User{id: id} |> get_assoc(...) = not ok because will populate empty assoc

Ah, it was use to behave like this but it was a major pain because you had to do this:

```
%User{}
|> preload(...)
|> put_assoc(...)
```

for something you knew that was empty. And the only way to make those decisions reliably is based if the struct is loaded or not.

So I understand your point of consistency, but we would be enforce consistency for a weird behaviour (manually setting the primary key) at the cost of making it more annoying for everyone for a very common behaviour.

Another way to put it, if you are manually setting the `id` for a struct for updates, you are already starting from a point where the whole data in the struct does not match the database. Assocs/embeds are only part of the same issue. If you want to load directly, then I suggest to use Repo.load or similar.

josevalim: No docs changed needed, i agree.

greg-rychlewski: Oh I thought Jose already changed the docs last night based on the misunderstand we had. If he didn't then no change.

greg-rychlewski (approved):

lib/ecto/changeset.ex
```diff
@@ -1203,6 +1205,98 @@ defmodule Ecto.Changeset do
...
+  end
+
+  defp get_relation(tag, %{changes: changes, data: data, types: types}, name) do
+    _ = relation!(:get, tag, name, Map.get(types, name))
+
+    existing =
+      case Map.fetch(changes, name) do
+        {:ok, value} -> value
+        :error -> Relation.load!(data, Map.fetch!(data, name))
+      end
```
josevalim, line 1291:
```suggestion
      case changes do
        %{^name => value} -> value
        %{} -> Relation.load!(data, Map.fetch!(data, name))
      end
```

josevalim:

josevalim:
Sorry for bringing this back but I am thinking `get_assoc` should default to `:struct` for consistency with `get_field`. Thoughts?

josevalim:
PR here #4141.

LostKobrakai:
@josevalim While I don't think it'll make much of a difference for anymore aware of all the options I'm not in favor of changing the defaults for anyone else. I've seen quite a few people in the past few month actively debug their code and still not notice that doing a `fetch_assocs_somehow |> modify |> put_assoc` with structs instead of changesets to silently swallow existing changes on those assocs.

This is not a concern for `get_field`/`get_change` as `put_change` will always just work.

If the default is changed I think there should be a warning in the documentation that hints to `put_assoc` likely not doing what is expected when it's passed a struct (or a changeset it build from those structs).

`get_field` includes:
> For relations, these functions will return the changeset data with changes applied. To retrieve raw changesets, please use get_change/3.

On `get_assoc` I think that would need to be highlighted even stronger including what this means to a later`put_assoc`.

Therefore I'd prefer people opting into getting raw structs rather than opting into getting changesets.

PR #4081, opened 2023-01-22
Improve the docs of schemaless changesets

c4710n:
`Ecto.Enum` is a parameterized type which can be used in schemaless changesets. But, the usage is not mentioned in the related docs.

This PR tries to:
+ explain the mechanism of parameterized types in schemaless changesets
+ describe the usage of `Ecto.Enum` in schemaless changesets

> Feel free to edit it. Any improvement is welcome.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -236,7 +236,28 @@ defmodule Ecto.Changeset do
         |> Ecto.Changeset.validate_required(...)
         |> Ecto.Changeset.validate_length(...)
 
-  Such functionality makes Ecto extremely useful to cast, validate and prune data even
+  Besides the basic types which are mentioned above, such as `:boolean` and `:string`, 
+  parameterized types can also be used in schemaless changesets. They implement
+  the `Ecto.ParameterizedType` behaviour and we can create the necessary type info by 
+  calling the `init/2` function.
+
+  For example, to use `Ecto.Enum` in a schemaless changeset:
+
+      types = %{
+        name: :string, 
+        role: Ecto.ParameterizedType.init(Ecto.Enum, values: [:reader, :editor, :admin])
+      }
+
+      data  = %{}
+      params = %{name: "Callum", role: "reader"}
+
+      changeset =
+        {data, types}
+        |> Ecto.Changeset.cast(params, Map.keys(types))
+        |> Ecto.Changeset.validate_required(...)
+        |> Ecto.Changeset.validate_length(...)
+
+  Schemaless changesets make Ecto extremely useful to cast, validate and prune data even
   if it is not meant to be persisted to the database.
 
   ### Changeset actions


lib/ecto/changeset.ex
```diff
@@ -236,6 +236,34 @@ defmodule Ecto.Changeset do
...
+
+  The first element `:parameterized` indicates you want to use a parameterized type.
+  The second element indicates which module to use for handling data. The third element 
+  is the options for this type. 
+
+  Generally, parameterized types will implement `Ecto.ParameterizedType` behaviour. Because
+  of that, in actual usage, we call `Ecto.ParameterizedType.init/2` directly, which will 
+  generate the three-element tuple we need.
+
+  Take `Ecto.Enum`, which implements `Ecto.ParameterizedType`, as an example:
```
josevalim, line 253:
```suggestion
  parameterized types can also be used in schemaless changesets. They implement
  the `Ecto.ParameterizedType` behaviour and therefore provide an `init/2` function.

  For example, to use `Ecto.Enum` in a schemaless changeset:
```

lib/ecto/changeset.ex
```diff
@@ -236,6 +236,34 @@ defmodule Ecto.Changeset do
...
+      types = %{
+        name: :string, 
+        sex: Ecto.ParameterizedType.init(Ecto.Enum, values: [:male, :female, :secret])
+      }
+      params = %{name: "Callum", sex: "male"}
+      changeset =
+        {data, types}
+        |> Ecto.Changeset.cast(params, Map.keys(types))
+        |> Ecto.Changeset.validate_required(...)
+        |> Ecto.Changeset.validate_length(...)
```
josevalim, line 265:
```suggestion
      types = %{
        name: :string, 
        role: Ecto.Enum.init(values: [:reader, :editor, :admin])
      }

      data  = %{}
      params = %{name: "Callum", role: "reader"}

      changeset =
        {data, types}
        |> Ecto.Changeset.cast(params, Map.keys(types))
        |> Ecto.Changeset.validate_required(...)
        |> Ecto.Changeset.validate_length(...)
```

josevalim:
Hi @c4710n! I could not edit, so I sent some suggestions. Thank you.

c4710n:
Hi @josevalim, the suggestions are nice.

Done.

lib/ecto/changeset.ex
c4710n: I found I have to write something like this:
```diff
- role: Ecto.Enum.init(values: [:reader, :editor, :admin])
+ role: Ecto.ParameterizedType.init(Ecto.Enum, values: [:reader, :editor, :admin])
```

lib/ecto/changeset.ex
```diff
@@ -236,6 +236,26 @@ defmodule Ecto.Changeset do
         |> Ecto.Changeset.validate_required(...)
         |> Ecto.Changeset.validate_length(...)
 
+  Besides the basic types which are mentioned above, such as `:boolean` and `:string`, 
+  parameterized types can also be used in schemaless changesets. They implement
+  the `Ecto.ParameterizedType` behaviour and therefore provide an `init/2` function.
+
+  For example, to use `Ecto.Enum` in a schemaless changeset:
```
v0idpwn, line 243: Not sure, but perhaps we should just show how to init the Ecto.Enum type, instead of including the whole flow?

lib/ecto/changeset.ex
```diff
@@ -236,6 +236,26 @@ defmodule Ecto.Changeset do
...
+      params = %{name: "Callum", role: "reader"}
+
+      changeset =
+        {data, types}
+        |> Ecto.Changeset.cast(params, Map.keys(types))
+        |> Ecto.Changeset.validate_required(...)
+        |> Ecto.Changeset.validate_length(...)
+
   Such functionality makes Ecto extremely useful to cast, validate and prune data even
   if it is not meant to be persisted to the database.
```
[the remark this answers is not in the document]
c4710n: done.

lib/ecto/changeset.ex
c4710n: > They implement the `Ecto.ParameterizedType` behaviour and therefore provide an `init/2` function.

The parameterized types that implement the `Ecto.ParameterizedType` behaviour, provides `init/1` instead of `init/2`. 

I changed suggested changes a little bit.

lib/ecto/changeset.ex
c4710n: I think that's a good idea.

There are two types can be used in schema changesets or schemaless changesets:
+ basic types - `Ecto.Type`
+ parameterized types - `Ecto.ParameterizedType`

We didn't mention `Ecto.Type` when describing the basic types, then maybe it's not necessary to mention `Ecto.ParameterizedType` when describing the parameterized types.

---
But, from another view, we are calling `Ecto.ParameterizedType.init/2`. If we don't mention the relationship between `Ecto.Enum` and `Ecto.ParameterizedType`, then it might be a bit abrupt, which makes the reader wonder "why do I need to call `Ecto.ParameterizedType.init/2`?"

josevalim:

PR #4085, opened 2023-01-26
Log warning when neither on nor assoc is not specified in join macro

dkuku:
Hi. As per discussion on the mailing list I added a warning when there are no on or assoc specified when joining tables.
https://groups.google.com/g/elixir-ecto/c/QgcepU4rSPU/m/q5k4St-QCAAJ
I updated most of the tests that emitted the warning, just 3 are left.
2 are emitted in a macro and It shows me line number as 0 and I can't find it :)
and the third one is with binding-less query
`from(p in "posts") |> join(:inner, "comments")` I'm not sure how to make this to not emit the error and if it's possible


dkuku:
I see it's already upstream :) I did not update my fork

PR #4089, opened 2023-01-31
Optionally update_change only on valid changesets

ding-an-sich:
This PR introduces an opt to `Changeset.update_change/4` that allows us to only invoke the update fn if the changeset is valid.

This can be useful in cases of dangerous or expensive updates. Examples:

```elixir
changeset
|> validate_inclusion(:reason, ["account_ownership_changed"])
|> update_change(:reason, &String.to_existing_atom/1, only_valid: true)
```

```elixir
changeset
|> update_change(:value, &call_slow_external_service_to_update_value/1, only_valid: true)
```

It keeps the default behavior of always invoking the updating fn.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -236,15 +236,15 @@ defmodule Ecto.Changeset do
         |> Ecto.Changeset.validate_required(...)
         |> Ecto.Changeset.validate_length(...)
 
-  Besides the basic types which are mentioned above, such as `:boolean` and `:string`, 
+  Besides the basic types which are mentioned above, such as `:boolean` and `:string`,
   parameterized types can also be used in schemaless changesets. They implement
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
Would there be a need for this in other functions? I'm wondering if instead of having it be an option on this function there should be something like this (ignore the lazy function name):

```elixir
def call_function_if_valid(cs, fun) do
   if cs.valid? do
      fun.(cs)
   else
     cs
  end
end
```

LostKobrakai:
Wouldn't the example be a good usecase for `Ecto.Enum`? There's also `Ecto.Changeset.prepare_changes` for doing things on validated changesets.

josevalim:
Since we are unsure about this, we will stick on the safer side and ask folks to still wrap it in a conditional for now. Thank you!

PR #4090, opened 2023-02-01
Add `with_ties/3` to `Ecto.Query`

greg-rychlewski:
Companion ecto_sql PR: https://github.com/elixir-ecto/ecto_sql/pull/471. The integration tests fail here but pass there because I was hesitant to update the earthfile.

**Background**

The SQL standard version of the `LIMIT` command is `FETCH`. The commands are almost identical except `FETCH` comes with an option `WITH TIES` which says that the result should include any records that are tied with the last record. For example `LIMIT 3` might return `1, 2, 3` while `FETCH 3 WITH TIES` would return `1, 2, 3, 3` because 2 records have the number 3.

**Proposal**

Add a `:with_ties` option to the current `limit` macro. There are a few reasons I'm proposing reusing the existing macro instead of creating a new `fetch` macro:

1. There is already [normalization happening in the SQL Server adapter](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/tds/connection.ex#L592), which has `FETCH` and `TOP` but not `LIMIT`.
2. If there are 2 macros then we need to decide whether to normalize both to each other in situations where they are the same and an adapter supports one but not the other. Normalizing both ways increases the code complexity and number of decisions we have to make. And if we decide not to normalize then we need to justify why we're already doing it for SQL Server.

This proposal has a couple difficulties:

1. `limit` was using the generic `%QueryExpr{}` struct so doesn't have a place to store this new option. If we create a new struct for limits then any adapter pattern matching on `%QueryExpr{}` instead of `%{}` will have to be updated. All the built-in adapters had to be updated and I noticed SQLite will have to be as well.
2. Because the signature will be `limit(query, bindings \\ [], expr, opts \\ [])` it's a little awkward specifying the option outside of keyword queries if you don't care about bindings. But this would be the case for a new macro as well.

**Reference**

The postgres docs talk about it here: https://www.postgresql.org/docs/current/sql-select.html. It's the only built-in adapter that supports `FETCH WITH TIES`. SQL Server supports `FETCH` but not `WITH TIES` option. And MySQL supports neither.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -435,6 +435,11 @@ defmodule Ecto.Query do
     defstruct [recursive: false, queries: []]
   end
 
+  defmodule LimitExpr do
+    @moduledoc false
+    defstruct [:expr, :file, :line, :with_ties, params: []]
+  end
+
   defmodule Tagged do
     @moduledoc false
     # * value is the tagged value
@@ -944,7 +949,37 @@ defmodule Ecto.Query do
   @from_join_opts [:as, :prefix, :hints]
   @no_binds [:union, :union_all, :except, :except_all, :intersect, :intersect_all]
   @binds [:lock, :where, :or_where, :select, :distinct, :order_by, :group_by, :windows] ++
-           [:having, :or_having, :limit, :offset, :preload, :update, :select_merge, :with_ctes]
+           [:having, :or_having, :offset, :preload, :update, :select_merge, :with_ctes]
+
+  defp from([{:limit, expr}|t], env, count_bind, quoted, binds) do
+    # If all bindings are integer indexes keep AST Macro expandable to %Query{},
+    # otherwise ensure that quoted code is evaluated before macro call
+    quoted =
+      if Enum.all?(binds, fn {_, value} -> is_integer(value) end) do
+        quote do
+          Ecto.Query.limit(unquote(quoted), unquote(binds), unquote(expr))
+        end
+      else
+        quote do
+          query = unquote(quoted)
+          Ecto.Query.limit(query, unquote(binds), unquote(expr))
+        end
+      end
+
+    {t, with_ties} = collect_with_ties(t, nil)
+
+    quoted =
+      if with_ties != nil do
+        quote do
+          query = unquote(quoted)
+          Ecto.Query.with_ties(query, unquote(binds), unquote(with_ties))
+        end
+      else
+        quoted
+      end
+
+    from(t, env, count_bind, quoted, binds)
+  end
 
   defp from([{type, expr}|t], env, count_bind, quoted, binds) when type in @binds do
     # If all bindings are integer indexes keep AST Macro expandable to %Query{},
@@ -983,6 +1018,10 @@ defmodule Ecto.Query do
     from(t, env, count_bind, quoted, to_query_binds(binds))
   end
 
+  defp from([{:with_ties, _value}|_], _env, _count_bind, _quoted, _binds) do
+    Builder.error! "`with_ties` keyword must immediately follow a limit"
+  end
+
   defp from([{:on, _value}|_], _env, _count_bind, _quoted, _binds) do
     Builder.error! "`on` keyword must immediately follow a join"
   end
@@ -1012,6 +1051,13 @@ defmodule Ecto.Query do
   defp join_qual(:cross_lateral_join), do: :cross_lateral
   defp join_qual(:left_lateral_join), do: :left_lateral
   defp join_qual(:inner_lateral_join), do: :inner_lateral
+
+  defp collect_with_ties([{:with_ties, with_ties} | t], nil),
+    do: collect_with_ties(t, with_ties)
+  defp collect_with_ties([{:with_ties, _} | _], _),
+    do: Builder.error! "`with_ties` keyword was given more than once to the same limit"
+  defp collect_with_ties(t, with_ties),
+    do: {t, with_ties}
 
   defp collect_on([{key, _} | _] = t, on, as, prefix, hints) when key in @from_join_opts do
     {t, as, prefix, hints} = collect_as_and_prefix_and_hints(t, as, prefix, hints)
@@ -1955,6 +2001,27 @@ defmodule Ecto.Query do
     Builder.LimitOffset.build(:limit, query, binding, expr, __CALLER__)
   end
 
+   @doc """
+  A limit query expression.
+
+  Limits the number of rows returned from the result. Can be any expression but
+  has to evaluate to an integer value and it can't include any field.
+
+  If `limit` is given twice, it overrides the previous value.
+
+  ## Keywords example
+
+      from(u in User, where: u.id == ^current_user, limit: 1)
+
+  ## Expressions example
+
+      User |> where([u], u.id == ^current_user) |> limit(1)
+
+  """
+  defmacro with_ties(query, binding \\ [], expr) do
+    Builder.LimitOffset.build(:with_ties, query, binding, expr, __CALLER__)
+  end
+
   @doc """
   An offset query expression.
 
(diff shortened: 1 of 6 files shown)


v0idpwn:
Not trying to impose, just asking: have you considered making a `fetch` macro instead of adding an option to `limit`?

lib/ecto/query.ex
```diff
@@ -989,6 +990,44 @@ defmodule Ecto.Query do
...
+
+  defp quoted_with_binds([{type, expr} | t], quoted, binds, true) do
+    quoted =
+      quote do
+        query = unquote(quoted)
+        Ecto.Query.unquote(type)(query, unquote(binds), unquote(expr))
+      end
+    
+    {quoted, t}
+  end
```
josevalim, line 1029: Can we implement this like we implement `on` for consistency? Basically, as soon as you see a `:limit`, you call `{with_ties, t} = collect_limit(t)`.

lib/ecto/query.ex
```diff
@@ -1928,17 +1967,41 @@ defmodule Ecto.Query do
...
       User |> where([u], u.id == ^current_user) |> limit(1)
 
+      User |> where([u], u.id == ^current_user) |> limit([], 1, with_ties: true)
+
   """
-  defmacro limit(query, binding \\ [], expr) do
-    Builder.LimitOffset.build(:limit, query, binding, expr, __CALLER__)
+  defmacro limit(query, binding \\ [], expr, opts \\ [])
+  defmacro limit(query, binding, expr, opts) when is_list(binding) and is_list(opts) do
+    with_ties = opts[:with_ties] || false
```
josevalim, line 1993: This is assuming `opts` is a keyword at compile-time. This means this would break:

```elixir
opts = [with_ties: true]
limit(query, [], 10, opts)
```

You need to handle the default value inside `Builder.Limit` inside a quote expression. :)

lib/ecto/query.ex
```diff
@@ -1928,17 +1967,41 @@ defmodule Ecto.Query do
...
   ## Expressions example
 
       User |> where([u], u.id == ^current_user) |> limit(1)
 
+      User |> where([u], u.id == ^current_user) |> limit([], 1, with_ties: true)
+
   """
-  defmacro limit(query, binding \\ [], expr) do
-    Builder.LimitOffset.build(:limit, query, binding, expr, __CALLER__)
+  defmacro limit(query, binding \\ [], expr, opts \\ [])
```
josevalim, line 1991: There is another approach here, which is to introduce a `with_ties` macro and it can only be called if a `limit` is set. It is similar to the `recursive_ctes` approach. :) I think it may be cleaner?

lib/ecto/query/builder/limit.ex
```diff
@@ -0,0 +1,58 @@
...
+    {expr, {params, _acc}} = Builder.escape(expr, :integer, {[], %{}}, binding, env)
+    params = Builder.escape_params(params)
+
+    if contains_variable?(expr) do
+      Builder.error! "query variables are not allowed in limit expressions"
+    end
+
+    limit = quote do: %Ecto.Query.LimitExpr{
+                        expr: unquote(expr),
+                        with_ties: unquote(with_ties),
```
josevalim, line 27: Based on the previous discussion, this would be something like:

```suggestion
                        with_ties: Keyword.get(unquote(opts), :with_ties, false),
```

But if we go with the separate function, then I think we don't need to change this at all. :)

lib/ecto/query.ex
greg-rychlewski: That would be a lot cleaner. Was tough trying to fit this into the existing macros.

The downside I see is using it in kw queries, you couldn't do this: `limit: 5, with_ties: true`. That seemed like the most natural way to use it to me. But yeah it doesn't seem there's a perfect solution either way.

josevalim:
> Not trying to impose, just asking: have you considered making a fetch macro instead of adding an option to limit?

at the moment, `fetch` means `limit`, so I think a separate `fetch` macro is not necessary. We can reevaluate in the future if more features are added. :)

> I'm open to this as well but didn't consider it before. I can't remember if that can be used in kw queries or not.

The KW queries can translate to anything we want. :D So I would keep basically the same code that you have in this PR, you would simply emit `limit`  |> `with_ties`.

lib/ecto/query.ex
greg-rychlewski: Ahh, I get it. Sorry for being dense. Then basically keep limit as it was and add a `:with_ties` field to the query struct?

josevalim: I would still define LimitQuery. But you may be able to keep using the `LimitOffset` builder if you pass the struct you want it to build as argument.

lib/ecto/query/builder/limit_offset.ex
```diff
@@ -5,39 +5,91 @@ defmodule Ecto.Query.Builder.LimitOffset do
...
   def apply(query, kind, expr) do
     apply(Ecto.Queryable.to_query(query), kind, expr)
   end
+
+  defp apply_limit(nil, _with_ties) do
+    Builder.error! "`with_ties` can only be applied to queries containing a `limit`"
+  end
+
+  defp apply_limit(limit, with_ties) do
+    %{limit | with_ties: with_ties}
```
greg-rychlewski, line 93: @josevalim Sorry to bother. I was trying to make `with_ties` a macro to keep the query compile time as long as possible.

I keep running into an issue here though because during compile time the limit struct is escaped when it gets here most of the time. And sometimes the module name is aliased, sometimes not.

Do you know if I should give up trying to make it a macro and make it a function instead? Or do you think I should be able to get this to work.

josevalim: You can make it a macro but then you need to use the whole builder infrastructure. It has the logic necessary for escaping and unescaping the query to keep it at compile-time. I am fine with whatever decision. :) We can start simpler as well and make it a macro later.

greg-rychlewski: Thanks. I think I got it working now. Requested re-review whenever you have the time

lib/ecto/query.ex
```diff
@@ -961,6 +966,8 @@ defmodule Ecto.Query do
         end
       end
 
+    {t, quoted} = maybe_with_ties(kw, quoted, binds)
```
josevalim, line 969:
```suggestion
    {t, quoted} = maybe_with_ties(type, t, quoted, binds)
```

So we don't match on limit again. :)

lib/ecto/query/builder/limit_offset.ex
```diff
@@ -5,39 +5,105 @@ defmodule Ecto.Query.Builder.LimitOffset do
...
+
   def apply(%Ecto.Query{} = query, :offset, expr) do
     %{query | offset: expr}
   end
+
   def apply(query, kind, expr) do
     apply(Ecto.Queryable.to_query(query), kind, expr)
   end
+
+  def apply_limit(nil, _with_ties) do
```
josevalim, line 94: Should this function be `defp`?

josevalim (approved):
Two minor nits and ship it!

lib/ecto/query/builder/limit_offset.ex
greg-rychlewski: Oh I made it public because the builder is unescaping the query struct but not the nested structs. So had to [do this](https://github.com/greg-rychlewski/ecto/blob/with_ties/lib/ecto/query/builder/limit_offset.ex#L104). There is something [similar here with ctes](https://github.com/greg-rychlewski/ecto/blob/with_ties/lib/ecto/query/builder/cte.ex#L82).

Does it seem ok? This part tripped me up the most not 100% sure I did it the best way.

josevalim: Doh, I saw just now that it has been called from the quote too. So this is good, just add a small `@doc` to align with the rest of the module?

greg-rychlewski: done!

josevalim (approved):

PR #4092, opened 2023-02-03
Fix docs on :prefix option for Repo Queryable callbacks

maltoe:
Hi :wave: 

We noticed that the docs for `c:Ecto.Repo.update_all/3` and `c:Ecto.Repo.delete_all/2` read

> This overrides the prefix set in the query and any
> `@schema_prefix` set in the schema.

This is not true, `update_all` and `delete_all` follow the query prefix precendence as the other functions in `Queryable`.

We also added a fourth precendence rule to point out that an already set query prefix (e.g. applied with `put_query_prefix/2`) takes precendence over the `:prefix` option of the `Repo` function.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -380,9 +380,10 @@ defmodule Ecto.Query do
     1. The `:prefix` option given to `from`/`join` has the highest precedence
     2. Then it falls back to the `@schema_prefix` attribute declared in the schema
       given to `from`/`join`
-    3. Then it falls back to the query prefix as applied with `put_query_prefix/2`
-    4. Last it falls back to the prefix option given to the respective
-       function on the `Repo` module
+    3. Then it falls back to the query prefix. The query prefix may be
+       set either on the query with `put_query_prefix/2` or by passing
+       the `:prefix` option when calling the `Repo` module (where the
+       former wins if both methods are used)
 
   The prefixes set in the query will be preserved when loading data.
   """
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -380,7 +380,9 @@ defmodule Ecto.Query do
     1. The `:prefix` option given to `from`/`join` has the highest precedence
     2. Then it falls back to the `@schema_prefix` attribute declared in the schema
       given to `from`/`join`
-    3. Then it falls back to the query prefix
+    3. Then it falls back to the query prefix as applied with `put_query_prefix/2`
+    4. Last it falls back to the prefix option given to the respective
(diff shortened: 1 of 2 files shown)


lib/ecto/query.ex
```diff
@@ -380,7 +380,9 @@ defmodule Ecto.Query do
     1. The `:prefix` option given to `from`/`join` has the highest precedence
     2. Then it falls back to the `@schema_prefix` attribute declared in the schema
       given to `from`/`join`
-    3. Then it falls back to the query prefix
+    3. Then it falls back to the query prefix as applied with `put_query_prefix/2`
+    4. Last it falls back to the prefix option given to the respective
+       function on the `Repo` module
```
josevalim, line 385:
```suggestion
    3. Then it falls back to the query prefix. The query prefix may be
       set either on the query with `put_query_prefix/2` or by passing
       the `:prefix` option when calling the `Repo` module
```

josevalim: Only one tiny change:  conceptually they  are the same thing, set in two different ways, so I wanted to keep it a single bullet but with more context. :)

maltoe: Reasoning behind my change here was that it took me by surprise once that passing the `prefix` in the options has no effect when the query already has a prefix set via `put_query_prefix/2` ([here](https://github.com/elixir-ecto/ecto/blob/d07d6eade6aa7145970214d878622722cafa56e3/lib/ecto/query/planner.ex#L1774), right?), which doesn't quite come across in your version now. But admittedly it's a bit of an edge case to use both at the same time, so maybe the added noise isn't worth it.

```suggestion
    3. Then it falls back to the query prefix
```

^^ apply to revert the change entirely

josevalim: I cannot apply suggestions because there maintainers cannot edit this PR. I think your clarifications are great, but I don't want to necessarily add a lookup as they all come from the same place (the query itself). Does that make sense? :) So my initial suggestion helps but you can improve it as well (without adding a fourth bullet item). WDYT?

maltoe: > I cannot apply suggestions

Right :see_no_evil: 

Sure, we could make it something like

```suggestion
    3. Then it falls back to the query prefix. The query prefix may be
       set either on the query with `put_query_prefix/2` or by passing
       the `:prefix` option when calling the `Repo` module (where the
       former wins if both methods are used)
```

josevalim: ship it! :)

josevalim:

Issue #4095, opened 2023-02-06
Preload behavior with `field: false`


Sleepful:
### Elixir version

Erlang/OTP 25 [erts-13.0.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]  Elixir 1.14.2 (compiled with Erlang/OTP 23)

### Database and Version

postgresql

### Ecto Versions

* ecto 3.9.1 (Hex package) (mix)

### Database Adapter and Versions (postgrex, myxql, etc)

* postgrex 0.16.5 (Hex package) (mix)

### Current behavior

Given a schema like
```elixir
defmodule Comment do
      schema "comment" do
        belongs_to :post,
          Post,
          [foreign_key: :post_id,
		  define_field: false]  # <--- this makes Preload fail

        field :related_post, :integer,
          [source: :post_id]
      end

      def changeset(schema, attrs) do
        schema
        |> cast(attrs, [:post_id])
        |> cast_assoc(:post)
      end
end
```
the following fails:
```elixir
Comment
  |> Repo.get!(id)
  |> Repo.preload(:post)
```
with error like:
`(MatchError) no match of right hand side value`

### Expected behavior

I expected `define_field: false` to work fine with Preload.

Perhaps I am misunderstanding its purpose?

removing `define_field: false` works fine, so maybe not-a-bug and I am just holding-it-wrong

greg-rychlewski:
If you provided a full stacktrace it would be easier for us to say for sure. But without knowing more preload will look for the value of the foreign key. So if it's not defined there will be a problem.

Sleepful:
The stacktrace wasn't valuable at all, it was rather confusing, there were no line-numbers referencing any Ecto files, just some native Enum functions.

The `define_field` docs say: `When false, does not automatically define a :foreign_key field, implying the user is defining the field manually elsewhere`

So this is what I did in the schema, with `field :related_post, :integer, source: :post_id` which matches the `:foreign_key` in `belongs_to`.

Removing the `define_field: false` fixed the issue, but now I have two definitions of a foreign key, I thought this was going to cause issues but it works fine. So what is the use case for `define_field`? 

I suppose the meaning of `does not automatically define a :foreign_key field` is a little lost on me, what does it mean to `define an FK field` in this case?

Let me know if my questions are too specific and I should answer them by reading the source myself 😅

greg-rychlewski:
[...]

PR #4096, opened 2023-02-06
Augment documentation for the `:source` schema field option

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4095

I think this should help with the confusion...hopefully.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -691,9 +691,11 @@ defmodule Ecto.Schema do
       contains the same value defined as default, validations will not be performed
       since there are no changes after all.
 
-    * `:source` - Defines the name that is to be used in database for this field.
+    * `:source` - Defines the name that is to be used in the database for this field.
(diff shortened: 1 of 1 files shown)


lib/ecto/schema.ex
```diff
@@ -691,9 +691,11 @@ defmodule Ecto.Schema do
       contains the same value defined as default, validations will not be performed
       since there are no changes after all.
 
-    * `:source` - Defines the name that is to be used in database for this field.
+    * `:source` - Defines the name that is to be used in the database for this field.
       This is useful when attaching to an existing database. The value should be
-      an atom.
+      an atom. This is a last minute translation before the query goes to the database.
+      All references within your Elixir code must still be to the field name,
+      i.e. in association foreign keys.
```
josevalim, line 698:
```suggestion
      All references within your Elixir code must still be to the field name,
      such as in association foreign keys.
```

PR #4097, opened 2023-02-07
Don't raise if query variables are used in limit/offset

greg-rychlewski:
Please let me know if I'm off base. But I think the intention was to not allow stuff like this: 

`SELECT a, b FROM table LIMIT a + 1`. 

If this is correct then the old check won't work because the query variable was being escaped before it got there. 

i.e. it is `{:{}, outer_meta, [:&, inner_meta, [var_num]]}` instead of `{:&, inner_meta, [var_num]}`.

### DIFF
--- a/test/ecto/query/builder/limit_offset_test.exs
+++ b/test/ecto/query/builder/limit_offset_test.exs
@@ -1,7 +1,10 @@
+Code.require_file "../../../support/eval_helpers.exs", __DIR__
+
 defmodule Ecto.Query.Builder.LimitOffsetTest do
   use ExUnit.Case, async: true
 
   import Ecto.Query
+  import Support.EvalHelpers
 
   test "overrides on duplicated limit and offset" do
     query = "posts" |> limit([], 1) |> limit([], 2)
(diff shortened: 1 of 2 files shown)


test/ecto/query/builder/limit_offset_test.exs
```diff
@@ -10,4 +10,16 @@ defmodule Ecto.Query.Builder.LimitOffsetTest do
     query = "posts" |> offset([], 1) |> offset([], 2) |> select([], 3)
     assert query.offset.expr == 2
   end
+
+  test "does not allow query variables in limit and offset" do
+    assert_raise Ecto.Query.CompileError, "query variables are not allowed in limit expression", fn ->
+      quoted = quote do: from p in "posts", limit: p.x + 1
+      Code.eval_quoted(quoted, [], __ENV__)
```
greg-rychlewski, line 17: I did it like this because the helper here didn't quite work for this case: https://github.com/elixir-ecto/ecto/blob/master/test/support/eval_helpers.exs.

It seems like the query compiled before going into the macro.

greg-rychlewski: actually i didn't import the macro...gg

josevalim (approved):
Looks good but i am woindering if we should just remove the check. Given it didn't work, it doesn't seem like it was a source of confusion plus I assume the database will raise something here anyway?

greg-rychlewski:
I agree. Updated the code to remove the check instead.

josevalim (approved):

PR #4098, opened 2023-02-07
Support interpolated values for from/join prefixes

maltoe:
Hi :wave: 

This patch adds support for interpolated prefix options on the from/join keywords of `Ecto.Query.from/2`.

## Problem

We have a little library called [Carbonite](https://github.com/bitcrowd/carbonite) which maintains a bunch of tables in a Postgres schema, defaulting to `carbonite_default`. We want to support "normal" Ecto usage of the corresponding schema modules, e.g. to allow the user to write `Repo.all(Carbonite.Change)`. Hence we have set the `@schema_prefix` option on our schemas. 

However, at the same time we need to support other (dynamic) schema prefixes and thus need to be able to override the `@schema_prefix` in various functions. Some of these functions are part of `Ecto.Repo`'s "Query API", so we figured ideally we'd like to be able to pass a variable to the `:prefix` option of `Ecto.Query.from/2`.

## Example

    prefix = "hello"
    from(c in "comments", prefix: ^prefix)

Hope this is a welcome addition. If there's good reason not to support this which we didn't see, sorry for the noise.

cheers,
malte

### DIFF
--- a/lib/ecto/query/builder/from.ex
+++ b/lib/ecto/query/builder/from.ex
@@ -80,12 +80,13 @@ defmodule Ecto.Query.Builder.From do
       )
     end
 
-    case prefix do
-      nil -> :ok
-      {:ok, prefix} when is_binary(prefix) or is_nil(prefix) -> :ok
-      _ -> Builder.error!("`prefix` must be a compile time string, got: `#{Macro.to_string(prefix)}`")
+    prefix = case prefix do
+      nil -> nil
+      {:ok, prefix} when is_binary(prefix) or is_nil(prefix) -> {:ok, prefix}
+      {:ok, {:^, _, [prefix]}} -> {:ok, quote(do: Ecto.Query.Builder.From.prefix!(unquote(prefix)))}
+      {:ok, prefix} -> Builder.error!("`prefix` must be a compile time string or an interpolated value using ^, got: #{Macro.to_string(prefix)}")
     end
-    
+
     as = case as do
       {:^, _, [as]} -> as
       as when is_atom(as) -> as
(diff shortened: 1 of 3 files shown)


lib/ecto/query/builder/from.ex
```diff
@@ -80,12 +80,13 @@ defmodule Ecto.Query.Builder.From do
...
     end
 
-    case prefix do
-      nil -> :ok
-      {:ok, prefix} when is_binary(prefix) or is_nil(prefix) -> :ok
-      _ -> Builder.error!("`prefix` must be a compile time string, got: `#{Macro.to_string(prefix)}`")
+    prefix = case prefix do
+      nil -> nil
+      {:ok, prefix} when is_binary(prefix) or is_nil(prefix) -> {:ok, prefix}
+      {:ok, {:^, _, [prefix]}} -> {:ok, prefix}
```
josevalim, line 86:
```suggestion
      {:ok, {:^, _, [prefix]}} -> {:ok, quote(do: Ecto.Query.Builder.From.prefix!(unquote(prefix))}
```

The idea is that we will wrap it in a code that will validate it at runtime. :) The same needs to be done for join!

josevalim (changes requested):
I have dropped a comment and once it is addressed I think this is good to go!

lib/ecto/query/builder/from.ex
maltoe: You mean like this? https://github.com/elixir-ecto/ecto/pull/4098/commits/6e546ac962c34726afb23d743d090a3add3b25ac

maltoe: We were unsure whether to use `Builder.error!` at this point as it's not really an `Ecto.Query.CompileError`?

maltoe:
CI failure seems unrelated, likely flaky.

josevalim:

Issue #3736, opened 2021-08-30
Missing documentation for some schema reflection APIs


joeapearson:
### Environment

* Ecto version (mix deps): latest (master branch)

### Current behavior

Some of the `Ecto.Schema` reflection APIs are not publicly documented.

Here's the existing docs:  https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L399
And here is the implementation to which I'm referring: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L602

Examples of documented APIs:

* `__schema__(:fields)`
* `__schema__(:source)`
* …etc

Examples of undocumented APIs:

* `__schema__(:autogenerate)`
* `__schema__(:query)`

### Expected behavior

There should be clear documentation for all APIs which are intended to be publicly consumed.

I'm happy to provide  PR but before I do so, I wonder whether some of the APIs remain intentionally undocumented, perhaps because they're intended to be private?

josevalim:
Yes, they are intentionally private. We can always make them public if there are use cases. :)

joeapearson:
Fair enough!

One use case we have for `__schema__(:autogenerate)` is to be able to use the referenced MFA generator to automatically generate values during `insert_all`.

josevalim:
With the new placeholders feature in Ecto's insert_all, it probably makes more sense to rely on that instead. Not sure if the autogenerate helps much then?

joeapearson:
Thanks @josevalim - we can use the placeholders feature I think yes, but for auto-generated values we would still like to be able to refer to the original generating function so that we can generate new ones.  (And then subsequently if someone needs to change or extend the schema, new auto-generated values should "just work").

josevalim:
@joeapearson in this case, perhaps a better API would be `__schema__(:autogenerate, field)`, and it autogenerates a value for you, so we are not tied to the particular storage detail of it. WDYT? Do you want to send a PR? :)

PR #4099, opened 2023-02-07
Add `:autogenerate_fields` to the schema reflection API

zachallaun:
Prior discussion: #3736

This PR adds `:autogenerate_fields` to the `__schema__` reflection API, which returns the field names of those fields that are auto generated on insert, _except for the primary key_, which is accessible through the existing `__schema__(:autogenerate_id)`.

There is currently no (public) API for accessing the fields that will be auto generated on insert. While this information is accessible through the private/undocumented `__schema__(:autogenerate)`, that would tie Ecto to the current storage of auto generated fields, as @josevalim mentioned in #3736. I've chosen instead to return a flat list of auto generated field names.

### Use-case

My specific use-case for this is to use it to skip auto generated fields in a library that constructs match ASTs from runtime values that can then be matched on later. To give a concrete example, the test:

```elixir
test "create_user/1 creates a user with valid attrs" do
  auto_assert Accounts.create_user(%{email: "valid@example.org"})
end
```

would be automatically updated the first time it is run to:

```elixir
test "create_user/1 creates a user with valid attrs" do
  auto_assert {:ok, %User{email: "valid@example.org"}} <- Accounts.create_user(%{email: "valid@example.org"})
end
```

The generated match pattern is minimized by comparing runtime field values to the default values returned by `User.__struct__()`, but without knowing which fields will be auto generated, it would _actually_ be written as:

```elixir
test "create_user/1 creates a user with valid attrs" do
  auto_assert {:ok,
               %User{
                 email: "valid@example.org",
                 inserted_at: ~N[2023-02-07 ...],
                 updated_at: ~N[2023-02-07 ...]
               }} <- Accounts.create_user(%{email: "valid@example.org"})
end
```

The next time this test is run, it would fail. `__schema__(:autogenerate_fields)` would be used to exclude by default fields that are auto generated if an Ecto schema is detected.

### DIFF
--- a/test/ecto/schema_test.exs
+++ b/test/ecto/schema_test.exs
@@ -21,14 +21,14 @@ defmodule Ecto.SchemaTest do
   end
 
   test "schema metadata" do
-    assert Schema.__schema__(:source)              == "my schema"
-    assert Schema.__schema__(:prefix)              == nil
-    assert Schema.__schema__(:fields)              == [:id, :name, :email, :password, :count, :array, :uuid, :query_excluded_field, :comment_id]
-    assert Schema.__schema__(:virtual_fields)      == [:temp]
-    assert Schema.__schema__(:query_fields)        == [:id, :name, :email, :password, :count, :array, :uuid, :comment_id]
-    assert Schema.__schema__(:read_after_writes)   == [:email, :count]
-    assert Schema.__schema__(:primary_key)         == [:id]
-    assert Schema.__schema__(:autogenerate_id)     == {:id, :id, :id}
+    assert Schema.__schema__(:source) == "my schema"
+    assert Schema.__schema__(:prefix) == nil
+    assert Schema.__schema__(:fields) == [:id, :name, :email, :password, :count, :array, :uuid, :query_excluded_field, :comment_id]
+    assert Schema.__schema__(:virtual_fields) == [:temp]
+    assert Schema.__schema__(:query_fields) == [:id, :name, :email, :password, :count, :array, :uuid, :comment_id]
+    assert Schema.__schema__(:read_after_writes) == [:email, :count]
+    assert Schema.__schema__(:primary_key) == [:id]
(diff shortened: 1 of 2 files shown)


zachallaun:
Regarding the decision to not include the primary key in the list of returned fields: I flip-flopped on it a bit, but decided not to include it because it's already covered by an existing API and because it simplified the implementation, since the `:autogenerate_id` may be `nil`.

test/ecto/schema_test.exs
```diff
@@ -21,13 +21,15 @@ defmodule Ecto.SchemaTest do
...
-    assert Schema.__schema__(:autogenerate_id)    == {:id, :id, :id}
+    assert Schema.__schema__(:source)              == "my schema"
+    assert Schema.__schema__(:prefix)              == nil
+    assert Schema.__schema__(:fields)              == [:id, :name, :email, :password, :count, :array, :uuid, :query_excluded_field, :comment_id]
+    assert Schema.__schema__(:virtual_fields)      == [:temp]
+    assert Schema.__schema__(:query_fields)        == [:id, :name, :email, :password, :count, :array, :uuid, :comment_id]
+    assert Schema.__schema__(:read_after_writes)   == [:email, :count]
+    assert Schema.__schema__(:primary_key)         == [:id]
+    assert Schema.__schema__(:autogenerate_id)     == {:id, :id, :id}
+    assert Schema.__schema__(:autogenerate_fields) == [:name, :uuid]
```
josevalim, line 32: Since we are rewriting it all, can we stop aligning the ==, like the formatted would? Thanks! :)

zachallaun: Sounds good to me! Would you prefer me to format the rest of the assertions in that test like the formatter would, or revert them to how they were aligned previously to minimize the diff?

zachallaun: Went with the first for now, but happy to rebase if you meant the second.

josevalim:

PR #4100, opened 2023-02-09
Support dynamic field inside json_extract_path

gabrielpra1:
This PR adds support for dynamic fields inside `json_extract_path`, useful when not only the JSON path is dynamic, but also which JSON field we want to access.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -237,6 +237,11 @@ defmodule Ecto.Query.Builder do
         {field, params_acc} = escape(field, type, params_acc, vars, env)
         {{:{}, [], [:json_extract_path, [], [field, path]]}, params_acc}
 
+      {:field, _, _} ->
+        path = escape_json_path(path)
+        {field, params_acc} = escape(field, type, params_acc, vars, env)
+        {{:{}, [], [:json_extract_path, [], [field, path]]}, params_acc}
+
       _ ->
         error!("`#{Macro.to_string(expr)}` is not a valid query expression")
     end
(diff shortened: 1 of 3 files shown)


lib/ecto/query/builder.ex
```diff
@@ -237,6 +237,11 @@ defmodule Ecto.Query.Builder do
         {field, params_acc} = escape(field, type, params_acc, vars, env)
         {{:{}, [], [:json_extract_path, [], [field, path]]}, params_acc}
 
+      {:field, _, _} ->
+        path = escape_json_path(path)
+        {field, params_acc} = escape(field, type, params_acc, vars, env)
+        {{:{}, [], [:json_extract_path, [], [field, path]]}, params_acc}
+
```
greg-rychlewski, line 244: Since the logic is duplicated maybe something like this?

```elixir
def escape({:json_extract_path, _, [field, path]} = expr, type, params_acc, vars, env) do
  field = json_field!(field)
  path = escape_json_path(path)
  {field, params_acc} = escape(field, type, params_acc, vars, env)
  {{:{}, [], [:json_extract_path, [], [field, path]]}, params_acc}
end

defp json_field({{:., _, _}, _, _} = field), do: field
defp json_field({:field, _, _} = field), do: field
defp json_field(field), do: error!("`#{Macro.to_string(field)}` is not a valid json field")
```

greg-rychlewski (commented):
Thank you for the PR. Would you also add a quick sentence and an example to the function doc so users know it's possible?

lib/ecto/query/builder.ex
gabrielpra1: I was reluctant to add it to the documentation because in my mind it was already expected for this to work, but I guess documentation is never too much :)

And nice refactoring suggestion, thanks!

greg-rychlewski:
Thanks!

PR #4101, opened 2023-02-09
Fix validate_json_path for nested map inside embed

gabrielpra1:
Currently `validate_json_path!` assumes that once it finds one embedded, everything down the path will also be embeds, which breaks when we have for example a `:map` inside an embedded schema.


josevalim:
@gabrielpra1 can you please take a look at CI? Thank you!

gabrielpra1:
@josevalim Done, just needed a rebase 🙂

josevalim:

PR #4103, opened 2023-02-10
Clarify documentation for `Repo.aggregate\4`

noahduncan:
Aggregating a by count on a numerical field is confusing, it is counting rows with a non-null value rather than counting the number of visits.  Summing the number of visits is more intuitive to understand and clarifies the usage a bit.


josevalim:

PR #4104, opened 2023-02-10
Add optional format/1 callback to ParameterizedType

skbolton:
The purpose of this callback is to format the output of a type when it
needs to be injected into errors and other logs. Previously, types were
injected into messaging by using `inspect(type)`. For ParameterizedTypes
this resulted in very noisy output. An example of this was `Ecto.Enum`.
Given this schema:

```elixir
defmodule User do
  use Ecto.Schema

  schema "users" do
    field :email, :string
    field :role, Ecto.Enum, values: [:admin, :editor], default: :editor

    timestamps()
  end
end
```

Doing the following action:

```elixir
Repo.insert!(%User{email: "alice@example.com", role: "admin"})
```

Would give you an error message similar to this

```
** (Ecto.ChangeError) value `"admin"` for `User.role` in `insert` does not match type
{:parameterized, Ecto.Enum, %{embed_as: :self, mappings: [admin: "admin", editor: "editor"],
on_cast: %{"admin" => :admin, "editor" => :editor}, on_dump: %{admin: "admin", editor:
"editor"}, on_load: %{"admin" => :admin, "editor" => :editor}, type: :string}}
```

The first implementer of this callback is `Ecto.Enum`. Given the
previous schema and incorrect type usage will result in the new error:

```
** (Ecto.ChangeError) value `"admin"` for `User.role` in `insert` does
not match type #Ecto.Enum<values: [:admin, :editor]>
```

This also includes parameterized types used in composite types. Given a
schema of:

```elixir
defmodule UserPermissions do
  use Ecto.Schema

  schema "user_permissions" do
    field :actions, {:array, Ecto.Enum}, values: [:read, :create, :update]

    timestamps()
  end
end
```

The following error could be produced:

```
** (Ecto.ChangeError) value `["write"]` for `UserPermissions.action` in `insert` does
not match type {:array, #Ecto.Enum<values: [:read, :create, :update]>}
```

For types that do not implement `format/1` the old behaviour of
`inspect(type)` will be the fallback.

RESOLVES: elixir-ecto#4093

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -1409,7 +1409,7 @@ defmodule Ecto.Association.ManyToMany do
       :error ->
         raise Ecto.ChangeError,
           "value `#{inspect value}` for `#{inspect struct.__struct__}.#{field}` " <>
-            "in `#{action}` does not match type #{inspect type}"
+            "in `#{action}` does not match type #{Ecto.Type.format(type)}"
     end
   end
 
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -208,6 +208,11 @@ defmodule Ecto.Enum do
   @impl true
   def embed_as(_, %{embed_as: embed_as}), do: embed_as
 
+  @impl true
+  def format(%{mappings: mappings}) do
+    "#Ecto.Enum<values: #{inspect Keyword.keys(mappings)}>"
+  end
+
   @doc "Returns the possible values for a given schema and field"
   @spec values(module, atom) :: [atom()]
   def values(schema, field) do
--- a/lib/ecto/parameterized_type.ex
+++ b/lib/ecto/parameterized_type.ex
@@ -169,7 +169,17 @@ defmodule Ecto.ParameterizedType do
   """
   @callback autogenerate(params()) :: term()
 
-  @optional_callbacks autogenerate: 1
+  @doc """
+  Formats output when a ParameterizedType is printed in exceptions and
+  other logs.
+
+  Note this callback is not used when constructing `Ecto.Changeset` validation
+  errors. See the `:message` option of most `Ecto.Changeset` validation
+  functions for how to customize error messaging on a per `Ecto.Changeset` basis.
+  """
+  @callback format(params()) :: String.t()
+
+  @optional_callbacks autogenerate: 1, format: 1
 
   @doc """
   Inits a parameterized type given by `type` with `opts`.
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1913,7 +1913,7 @@ defmodule Ecto.Query.Planner do
       {:ok, v} ->
         {:ok, v}
       _ ->
-        {:error, "value `#{inspect v}` in `#{kind}` cannot be cast to type #{inspect type}"}
+        {:error, "value `#{inspect v}` in `#{kind}` cannot be cast to type #{Ecto.Type.format(type)}"}
     end
   end
 
@@ -1922,7 +1922,7 @@ defmodule Ecto.Query.Planner do
       {:ok, v} ->
         {:ok, v}
       :error ->
-        {:error, "value `#{inspect v}` cannot be dumped to type #{inspect type}"}
+        {:error, "value `#{inspect v}` cannot be dumped to type #{Ecto.Type.format(type)}"}
     end
   end
 
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -1209,6 +1209,23 @@ defmodule Ecto.Type do
     false
   end
 
+  @doc """
+  Format type for error messaging and logs.
+  """
+  def format({:array, type}) do
+    "{:array, #{format(type)}}"
+  end
+
+  def format({:parameterized, type, params}) do
+    if function_exported?(type, :format, 1) do
+      apply(type, :format, [params])
+    else
+      "##{inspect(type)}<#{inspect(params)}>"
+    end
+  end
+
+  def format(type), do: inspect(type)
+
   ## Helpers
 
   # Checks if a value is of the given primitive type.
--- a/lib/ecto/exceptions.ex
+++ b/lib/ecto/exceptions.ex
@@ -145,7 +145,7 @@ defmodule Ecto.CastError do
   def exception(opts) do
     type  = Keyword.fetch!(opts, :type)
     value = Keyword.fetch!(opts, :value)
-    msg   = opts[:message] || "cannot cast #{inspect value} to #{inspect type}"
+    msg   = opts[:message] || "cannot cast #{inspect value} to #{Ecto.Type.format(type)}"
     %__MODULE__{message: msg, type: type, value: value}
   end
 end
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -417,7 +417,7 @@ defmodule Ecto.Repo.Queryable do
         struct = struct && " in #{inspect(struct)}"
 
         raise ArgumentError,
-              "cannot load `#{inspect(value)}` as type #{inspect(type)}#{field}#{struct}"
+              "cannot load `#{inspect(value)}` as type #{Ecto.Type.format(type)}#{field}#{struct}"
     end
   end
 
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -836,7 +836,7 @@ defmodule Ecto.Repo.Schema do
       {:ok, value} ->
         load_each(%{struct | key => value}, kv, types, adapter)
       :error ->
-        raise ArgumentError, "cannot load `#{inspect value}` as type #{inspect type} " <>
+        raise ArgumentError, "cannot load `#{inspect value}` as type #{Ecto.Type.format(type)} " <>
                              "for field `#{key}` in schema #{inspect struct.__struct__}"
     end
   end
@@ -1009,7 +1009,7 @@ defmodule Ecto.Repo.Schema do
       :error ->
         raise Ecto.ChangeError,
               "value `#{inspect(value)}` for `#{inspect(schema)}.#{field}` " <>
-              "in `#{action}` does not match type #{inspect type}"
+              "in `#{action}` does not match type #{Ecto.Type.format(type)}"
     end
   end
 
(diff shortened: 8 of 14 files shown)


skbolton:
Going to keep at this and thinking about this issue. Some things did pop up that need more exploring. Putting in draft so any one passing by can add their input that I can use for the next time I dive into this.

lib/ecto/association.ex
```diff
@@ -12,16 +12,16 @@ defmodule Ecto.Association.NotLoaded do
   """
 
   @type t :: %__MODULE__{
-    __field__: atom(),
```
skbolton, line 15: formatting kicked in. I'll need to explore what I need to do to get this to not be the case.

lib/ecto/association.ex
```diff
@@ -1396,20 +1655,31 @@ defmodule Ecto.Association.ManyToMany do
...
 
     case Ecto.Type.adapter_dump(adapter, type, value) do
       {:ok, value} ->
         value
+
       :error ->
-        raise Ecto.ChangeError,
-          "value `#{inspect value}` for `#{inspect struct.__struct__}.#{field}` " <>
-            "in `#{action}` does not match type #{inspect type}"
+        message =
```
skbolton, line 1671: This is the only meaningful change in this file other than formatting

skbolton:
@wojtekmach @josevalim @greg-rychlewski I'll pick this back up in a few days and dive more into things. If you have any thoughts in the meantime on approach this PR will be a better place to talk about things than the original issue.

lib/ecto/enum.ex
```diff
@@ -81,6 +81,7 @@ defmodule Ecto.Enum do
   """
 
   use Ecto.ParameterizedType
+  import Kernel, except: [inspect: 1]
```
wojtekmach, line 84: yeah let's definitely call the callback something else. How about `format_error` until we come up with something better?

wojtekmach (commented):
This is a great start, thank you for picking this up. Please revert formatter changes and add some tests!

lib/ecto/parameterized_type.ex
```diff
@@ -191,7 +201,12 @@ defmodule Ecto.ParameterizedType do
       @doc false
       def equal?(term1, term2, _params), do: term1 == term2
 
-      defoverridable embed_as: 2, equal?: 3
+      # @doc false
+      def inspect(parameterized_type) do
+        inspect(parameterized_type)
```
wojtekmach, line 206: 😨

lib/ecto/association.ex
josevalim: Yes, please disable the formatter in your editor when working on Ecto. :)

lib/ecto/association.ex
josevalim: Instead of adding the `case `here, add a function called `Ecto.Type.inspect(type, opts \\ [])` that will be responsible for checking which type it is and delegate accordingly.

lib/ecto/enum.ex
josevalim: I think `inspect` is fine. It is exact the purpose we expect it to have.

lib/ecto/parameterized_type.ex
```diff
@@ -169,6 +169,16 @@ defmodule Ecto.ParameterizedType do
...
 
+  @doc """
+  Customizes output when a ParameterizedType is printed in exceptions and
+  other logs.
+
+  Note this callback is not used when constructing `Ecto.Changeset` validation
+  errors. See the `:message` option of most `Ecto.Changeset` validation
+  functions for how so customize error messaging on a per `Ecto.Changeset` basis.
+  """
+  @callback inspect({:parameterized, type :: atom(), params()}) :: String.t()
```
josevalim, line 180: The callback should receive only the params.

lib/ecto/parameterized_type.ex
josevalim, line 206:
```suggestion
        Kernel.inspect(parameterized_type)
```

And, as per the above, it should only receive the params. :)

lib/ecto/enum.ex
josevalim: Nah, I see that it will be confusing with each one we are calling. Let's rename it to `format`.

josevalim:
Thank you @skbolton for the PR! It seems that my suggestion to call it `inspect/1` was not the best idea and your PR helped show that. Your PR also help show other flaws in our proposal. So here is a new proposal:

The flow will be like this:

1. Add a `format/1` as _optional_ callback for parameterized types
2. Introduce a function called `Ecto.Type.format/1`
3. If `Ecto.Type.format/1` receives a parameterized type:
  a. Check if `format/1` in the type module is exported, if it is, call it
  b. if not defined, format it as `##{inspect(param_module)}<#{inspect(params)}>`
4. For all other types, call `inspect(type)`

Now formatting it anywhere should be a matter of replacing `inspect(type)` by `Ecto.Type.format(type)` :)

lib/ecto/enum.ex
```diff
@@ -208,6 +208,22 @@ defmodule Ecto.Enum do
...
+    # `:value1, :value2, or :value3`
+    [last_option | first_options] = Enum.reverse(valid_options)
+
+    comma_options =
+      first_options
+      |> Enum.reverse()
+      |> Enum.map(fn option -> ":" <> Atom.to_string(option) end)
+      |> Enum.join(", ")
+
+    "Ecto.Enum with values #{comma_options} or :#{last_option}"
```
greg-rychlewski, line 224:
```suggestion
    "Ecto.Enum with values #{inspect Keyword.keys(mappings)}"
```
Not sure if this will be popular, but since it's an error message for developers maybe this is alright?

josevalim:
```suggestion
    [last_option | first_options] = Enum.reverse(valid_options)

    comma_options =
      first_options
      |> Enum.reverse()
      |> Enum.map(fn option -> ":" <> Atom.to_string(option) end)
      |> Enum.join(", ")

    "#Ecto.Enum<values:  #{inspect Keyword.keys(mappings)}>"
```

lib/ecto/query/planner.ex
```diff
@@ -1913,7 +1913,7 @@ defmodule Ecto.Query.Planner do
       {:ok, v} ->
         {:ok, v}
       _ ->
-        {:error, "value `#{inspect v}` in `#{kind}` cannot be cast to type #{inspect type}"}
+        {:error, "value `#{inspect v}` in `#{kind}` cannot be cast to type #{Ecto.Type.format(type)}"}
```
skbolton, line 1916: I went to everywhere in the code that some form of `inspect type` was called and changed almost all of them to this new signature. After this PR there will be some places where `inspect(type)` is still called because it felt like an overreach to push them down this path.

skbolton: Let me know @josevalim @wojtekmach if you would prefer that any instance of `inspect(type)` goes down this path

lib/ecto/type.ex
```diff
@@ -1209,6 +1209,19 @@ defmodule Ecto.Type do
...
+  """
+  def format({:parameterized, type, params}) do
+    if function_exported?(type, :format, 1) do
+      apply(type, :format, [params])
+    else
+      "##{inspect(type)}<#{inspect(params)}>"
+    end
+  end
+
+  def format(type), do: inspect(type)
```
v0idpwn, line 1223: We should handle composite types. For example, `{:array, {:parameterized, Ecto.Enum, params}` should be formatted as `{:array, ##Ecto.Enum<...>}`

v0idpwn (commented):
Looks good to me, just some improvements

lib/ecto/query/planner.ex
skbolton: @v0idpwn fixed two of those see 2295 explanation in my original comment

skbolton: Feeling good that I have all desired `inspect type` spots that are related to this
refactor covered.

skbolton:
@v0idpwn I resolved [comment](https://github.com/elixir-ecto/ecto/pull/4104/files#r1103881772) because it was really clogging up the review imo. Take a look and open back up if you feel I missed something.

greg-rychlewski:
I think we have to do it for map composites as well: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L184

tbh I never heard of that type before but it looks like it means the values in the map are the parameterized type

greg-rychlewski:
And maybe these ones but not sure...I have to look closer: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L184

skbolton:
> Maybe not these ones since they are private..but not totally sure...I have to look closer: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L186
> 
> edit: yeah sorry I was overthinking it. I don't believe these matter for logging to users

Maybe I should... I could soften the pattern match and ask if it's a composite type and then wrap in logging

greg-rychlewski:
> Maybe I should... I could soften the pattern match and ask if it's a composite type and then wrap in logging

We can see what Jose/v0idpwn think. If they are private types then I don't think they'll be logged to the user. Otherwise the user no way of knowing what it means.

skbolton:
The [polyembed](https://github.com/mathieuprog/polymorphic_embed/blob/v3.0.5/lib/polymorphic_embed.ex#L1) library embeds parameterized types as maps. This callback might be something they would be interested in. It's not a big stretch to include that change so I'll leave that as the last todo for me.

> > Maybe I should... I could soften the pattern match and ask if it's a composite type and then wrap in logging
> 
> We can see what Jose/v0idpwn think. If they are private types then I don't think they'll be logged to the user. Otherwise the user no way of knowing what it means.

The [polyembed](https://github.com/mathieuprog/polymorphic_embed/blob/v3.0.5/lib/polymorphic_embed.ex#L1) embeds parameterized types as maps come to think of it. They would probably be interested in this callback. It's not a big stretch to try and cover this case so I'll leave it as the last todo for me to cover.

greg-rychlewski:
> The [polyembed](https://github.com/mathieuprog/polymorphic_embed/blob/v3.0.5/lib/polymorphic_embed.ex#L1) library embeds parameterized types as maps come to think of it. They would probably be interested in this callback. It's not a big stretch to try and cover this case so I'll leave it as the last todo for me to cover.

Oh sorry I was talking about the private composite types: ` @typep private_composite :: {:maybe, t} | {:in, t} | {:param, :any_datetime}`

The map composite we should add for sure because it's public.

josevalim:

PR #4107, opened 2023-02-17
Add option to Multi.one to consider nil value as error

lucas-cegatti:
This proposed change adds the option `nil_as_error` to `Ecto.Multi.one/4` to return `{:error, nil}` instead of `{:ok, nil}` which could be useful on some cases where we need to break the pipeline when no value is returned.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -402,11 +402,15 @@ defmodule Ecto.Multi do
 
   Accepts the same arguments and options as `c:Ecto.Repo.one/2`.
 
+  ## Options
+
+    * `:nil_as_error` - Returns {:error, nil} instead of {:ok, nil}.
+
   ## Example
 
       Ecto.Multi.new()
       |> Ecto.Multi.one(:post, Post)
-      |> Ecto.Multi.one(:author, fn %{post: post} -> 
+      |> Ecto.Multi.one(:author, fn %{post: post} ->
         from(a in Author, where: a.id == ^post.author_id)
       end)
       |> MyApp.Repo.transaction()
@@ -725,7 +729,7 @@ defmodule Ecto.Multi do
   In the example above there isn't a large benefit in putting the
   `company` in the multi, because you could also access the
   `company` variable directly inside the anonymous function.
-  
+
   However, the benefit of `put/3` is when composing `Ecto.Multi`s.
   If the insert operations above were defined in another module,
   you could use `put(:company, company)` to inject changes that
@@ -890,8 +894,17 @@ defmodule Ecto.Multi do
   end
 
   defp operation_fun({:one, fun}, opts) do
+    nil_as_error = Keyword.get(opts, :nil_as_error, false)
+
     fn repo, changes ->
-      {:ok, repo.one(fun.(changes), opts)}
+      repo.one(fun.(changes), opts)
+      |> case do
+        nil when nil_as_error ->
+          {:error, nil}
+
+        result ->
+          {:ok, result}
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you for the PR but I would prefer to not overload the meaning of `one` here with a separate meaning than the one in `Ecto.Repo`. It would likely be clearer to have a separate function but it is unclear what should be its name.

Another option is to have an explicit step to abort on nil: `Ecto.Multi.one |> Ecto.Multi.error_on_nil(& &1.foo)`

lucas-cegatti:
> Thank you for the PR but I would prefer to not overload the meaning of `one` here with a separate meaning than the one in `Ecto.Repo`. It would likely be clearer to have a separate function but it is unclear what should be its name.
> 
> Another option is to have an explicit step to abort on nil: `Ecto.Multi.one |> Ecto.Multi.error_on_nil(& &1.foo)`

Thank you for the feedback, @josevalim 

To be honest I'm not sure what would be the best approach here, right now I'm adding validations to a `Ecto.Multi.run`, I'll send another commit with the suggested function `Ecto.Multi.error_on_nil(& &1.foo)`. 

Would a function named fetch makes sense in the Multi context? e.g. `Ecto.Multi.fetch(:foo, from(f in Foo, where: f.id == 1)) ` it would run a repo.one but returning :error on nil.

Thanks again.

greg-rychlewski:
Maybe `Multi.halt?/1`,  similar to plug? It accepts a function where the argument is the changes and if it evaluates to true then returns error?

lucas-cegatti:
> Maybe `Multi.halt?/2`,  similar to plug? It accepts a function where the argument is the changes and if it evaluates to true then returns error? Or `/3` with an option to configure the error?
> 
> This would also let you stop the pipeline with conditions based on multiple steps. 

That's great @greg-rychlewski .

I'll implement it later and submit it to review.

Thanks!!

greg-rychlewski:
> I'll implement it later and submit it to review.

Oh maybe let's wait for Jose's opinion. Actually I thought about what I said some more and I have a counter argument to it:

It's basically the same amount of code as 'run' and `run` is letting you control the error message more naturally.  Also not sure if the meaning of `halt` is the same as this. Reading the docs again, `halt` seems more like "stop and return what you have", not "stop and create an error". 

Sorry for the noise. Probably using `run` is better than what I said. Still trying to think of possible new function but not sure yet.

One thing though, I'm not too sure about returning an error straight from a Repo-like function. I'm not completely sure we should ever say it's an error for a query to have no results. That seems more like post-processing to me.

greg-rychlewski:
The best thing I can think of for a new function is something like this:

`Multi.error(multi, name, fn changes -> true/false end, reason)`

Conditionally returns an error with `{:error, reason}`. Otherwise the pipeline continues. 

I was thinking about `Multi.fetch` as well but multi aborts on `{:error, _}` and not `:error` so we wouldn't be using fetch completely inline with other parts of Elixir.

greg-rychlewski:
Let's say we agree there is no good way to add a Repo-like function to return `{:error, nil}`. Then I think the issue boils down to this:

Either this is good enough or it's too inconvenient:

```elixir
Multi.new()
|> Multi.one(:one, Post)
|> Multi.merge(fn 
  %{post: nil} -> Multi.new() |> Multi.error(:no_post, "no post")
  _ -> Multi.new()
end)
```

It is maybe a little verbose if you just want to conditionally add something. Assuming I didn't overlook anything and we agree this can be improved, maybe adding something to conditionally execute a step? Merge seems better suite to dynamically creating a step  you know you want to exist.

josevalim:
And you can compose in other ways: `run(multi, :foo, fetch_one(query))` where fetch_one does:

```elixir
def fetch_one(query)
fn repo ->
  if val = repo.one(query) do
    {:ok, val}
  else
    {:error, ...}
  end
end
end
```

So we probably don't need to extend the API. THanks for the PR in any case @lucas-cegatti !

PR #4108, opened 2023-02-20
Add Ecto.Type.format/1 clause for maps with inner type

v0idpwn:


lib/ecto/type.ex
```diff
@@ -1212,8 +1212,8 @@ defmodule Ecto.Type do
   @doc """
   Format type for error messaging and logs.
   """
-  def format({:array, type}) do
-    "{:array, #{format(type)}}"
+  def format({composite, type}) when composite in [:array, :map] do
```
v0idpwn, line 1215: I can make it two clauses if you think its more aesthetically pleasing :)

josevalim (approved):

lib/ecto/type.ex
skbolton: There is also `Ecto.Type.composite?/1` that can be called to see if its a composite
type. Might be better to latch to that rather than having the composite keys in
2 spots.

Issue #4109, opened 2023-02-21
Discussion: validate_required_or(_xor)


greg-rychlewski:
### Elixir version

All

### Database and Version

All

### Ecto Versions

3.10-dev

### Database Adapter and Versions (postgrex, myxql, etc)

All

### Current behavior

Functions don't exist.

### Expected behavior

I have a feeling this came up before but I couldn't find any previous discussion. Apologies if I missed it.

The idea would be to have 2 new functions in `Ecto.Changeset`:

- `validate_required_or/3`
- `validate_required_xor/3`

They would both take a list of fields. The former would validate that at least one of the fields is present. The latter that exactly one of the fields is present. 

I came across this when validating API query parameters and didn't see anything in the current API to do something like it. It's not terrible to roll my own solution either so not a big deal either way.

v0idpwn:
validate_required/3 accepts options. It could be an alternative to add options for such behaviours. Though maybe its not a good idea because it changes the semantics of the function.

wojtekmach:
Yeah I've needed something like this myself a few times. Having it out of the box would be a nice addition.

I'm not sure about implementing it though. We keep required validations as `changeset.required`, not `changeset.validations` like all the other ones,

```elixir
iex> Ecto.Changeset.cast({%{}, %{name: :string}}, %{}, ~w(name)a) |> Ecto.Changeset.validate_required(:name) |> Map.take([:required, :validations])
%{required: [:name], validations: []}
```

and so not sure how to represent at-least-one-required and at-most-one-required using the current scheme. I guess if neither at-least-one nor at-most-one is given, we keep the validation in `changeset.errors`, otherwise we add `{:required, at_least_one|at_most_one: true}`?

josevalim:
@greg-rychlewski I agree with both decisions and their rationale.

wojtekmach:
+1 from me too. It will be slightly awkward that regular required validations will be represented differently than new required validations but we cannot reconcile the two and shouldn’t make breaking change and I don’t think it’s a huge deal anyway.

PR #4111, opened 2023-02-23
Add `field_missing?/2` to Ecto.Changeset.

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4109.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2096,8 +2096,7 @@ defmodule Ecto.Changeset do
 
   """
   @spec validate_any_required(t, list | atom, Keyword.t) :: t
-  def validate_any_required(%Changeset{} = changeset, fields, opts \\ [])
-      when not is_nil(fields) do
+  def validate_any_required(%Changeset{} = changeset, fields, opts \\ []) when not is_nil(fields) do
     %{validations: validations, errors: errors, types: types} = changeset
     fields = List.wrap(fields)
 
@@ -2149,8 +2148,7 @@ defmodule Ecto.Changeset do
 
   """
   @spec validate_one_of_required(t, list | atom, Keyword.t) :: t
-  def validate_one_of_required(%Changeset{} = changeset, fields, opts \\ [])
-      when not is_nil(fields) do
+  def validate_one_of_required(%Changeset{} = changeset, fields, opts \\ []) when not is_nil(fields) do
     %{validations: validations, errors: errors, types: types} = changeset
     fields = List.wrap(fields)
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2077,6 +2077,23 @@ defmodule Ecto.Changeset do
 
     * `:message` - the message on failure, defaults to "at least one field must be present"
 
+  ## Examples
+
+      iex> changeset = cast(%Post{}, %{title: "Title"}, [:title])
+      iex> changeset = validate_any_required(changeset, [:title, :body])
+      iex> changeset.valid?
+      true
+
+      iex> changeset = cast(%Post{}, %{title: "Title", body: "Body"}, [:title, :body])
+      iex> changeset = validate_any_required(changeset, [:title, :body])
+      iex> changeset.valid?
+      true
+
+      iex> changeset = cast(%Post{}, %{color: "Red"}, [:color])
+      iex> changeset = validate_any_required(changeset, [:title, :body])
+      iex> changeset.valid?
+      false
+
   """
   @spec validate_any_required(t, list | atom, Keyword.t) :: t
   def validate_any_required(%Changeset{} = changeset, fields, opts \\ [])
@@ -2112,6 +2129,23 @@ defmodule Ecto.Changeset do
   ## Options
 
     * `:message` - the message on failure, defaults to "exactly one field must be present"
+
+  ## Examples
+
+      iex> changeset = cast(%Post{}, %{title: "Title"}, [:title])
+      iex> changeset = validate_one_of_required(changeset, [:title, :body])
+      iex> changeset.valid?
+      true
+
+      iex> changeset = cast(%Post{}, %{title: "Title", body: "Body"}, [:title, :body])
+      iex> changeset = validate_one_of_required(changeset, [:title, :body])
+      iex> changeset.valid?
+      false
+
+      iex> changeset = cast(%Post{}, %{color: "Red"}, [:color])
+      iex> changeset = validate_one_of_required(changeset, [:title, :body])
+      iex> changeset.valid?
+      false
 
   """
   @spec validate_one_of_required(t, list | atom, Keyword.t) :: t
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2049,9 +2049,7 @@ defmodule Ecto.Changeset do
 
     fields_with_errors =
       for field <- fields,
-          ensure_field_not_many!(types, field),
-          missing?(changeset, field),
-          ensure_field_exists!(changeset, types, field),
+          invalid_required_field!(changeset, types, field),
           is_nil(errors[field]),
           do: field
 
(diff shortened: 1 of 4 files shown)


lib/ecto/changeset.ex
```diff
@@ -2067,6 +2065,124 @@ defmodule Ecto.Changeset do
...
+      iex> changeset.valid?
+      false
+
+      iex> changeset = cast(%Post{}, %{color: "Red"}, [:color])
+      iex> changeset = validate_one_of_required(changeset, [:title, :body])
+      iex> changeset.valid?
+      false
+
+  """
+  @spec validate_one_of_required(t, list | atom, Keyword.t) :: t
```
josevalim, line 2150:
```suggestion
  @spec validate_exactly_one_required(t, list | atom, Keyword.t) :: t
```

?

lib/ecto/changeset.ex
```diff
@@ -2067,6 +2065,124 @@ defmodule Ecto.Changeset do
...
+        end
+      end)
+
+    case valid_fields do
+      [valid_field] ->
+        new_validation = {valid_field, {:one_of_required, fields: fields}}
+        %{changeset | validations: [new_validation | validations]}
+
+      _ ->
+        error_field = if valid_fields == [], do: hd(fields), else: hd(valid_fields)
```
josevalim, line 2174: I think it would likely be easier to manage by callers if the field we add the error is always the same (i.e. the first field).

josevalim:
I have been thinking more about this and I was wondering if instead we can add a function called `filter_required`, that receives a list of fields and returns all fields that are required.

```elixir
case filter_required(changeset, fields) do
  [] -> add_error(changeset, :foo, ....)
  [_ | _] -> changeset
end
```

We may prefer this route because:

1. Fewer LOCs
2. We force the user to write better error messages. The error messages we generate in the functions above are not good but they are the best we can do. Ideally we want the user to write: "only one of x or y must be set".
3. Users can do more complex validations if required

Sorry about coming up with this idea just now but the insight only came after reviewing the code!

greg-rychlewski:
@josevalim that's a really good idea. how's this?

greg-rychlewski:
Actually I think I got it backwards sorry. I will reverse the logic.

greg-rychlewski:
Fixed.

lib/ecto/changeset.ex
```diff
@@ -2067,6 +2065,42 @@ defmodule Ecto.Changeset do
     end
   end
 
+  @doc """
+  Returns a list of fields that are present.
+
+  Each field passed into this function will have its prescence evaluated
```
wojtekmach, line 2071:
```suggestion
  Each field passed into this function will have its presence evaluated
```

lib/ecto/changeset.ex
```diff
@@ -2067,6 +2065,42 @@ defmodule Ecto.Changeset do
...
+      iex> changeset = cast(%Post{}, %{title: "Title"}, [:title])
+      iex> filter_required(changeset, [:title, :body])
+      [:title]
+
+      iex> changeset = cast(%Post{}, %{color: "Red"}, [:color])
+      iex> filter_required(changeset, [:title, :body])
+      []
+
+  """
+  @spec filter_required(t, list | atom) :: list
```
wojtekmach, line 2091:
```suggestion
  @spec filter_required(t(), [atom()] | atom()) :: [atom()]
```

lib/ecto/changeset.ex
```diff
@@ -2067,6 +2065,42 @@ defmodule Ecto.Changeset do
...
+  def filter_required(%Changeset{} = changeset, fields) when not is_nil(fields) do
+    %{types: types} = changeset
+    fields = List.wrap(fields)
+
+    Enum.filter(fields, &(not invalid_required_field!(changeset, types, &1)))
```
wojtekmach, line 2096:
```suggestion
    fields = List.wrap(fields)
    Enum.filter(fields, &(not invalid_required_field!(changeset, changeset.types, &1)))
```

test/ecto/changeset_test.exs
```diff
@@ -1061,6 +1061,26 @@ defmodule Ecto.ChangesetTest do
     assert changeset.errors == [topics: {"can't be blank", [validation: :required]}]
   end
 
+  test "filter_required/2" do
+    required_fields =
+      changeset(%{"body" => "something"})
+      |> filter_required([:title, :body])
+
+    assert [:body] == required_fields
```
wojtekmach, line 1069: nitpick: either `assert actual == expected` or `assert expected = actual` but not `assert expected == actual`, wdyt?

```suggestion
    assert required == [:body]
```

test/ecto/changeset_test.exs
```diff
@@ -1061,6 +1061,26 @@ defmodule Ecto.ChangesetTest do
...
+      |> filter_required([:title, :body])
+
+    assert [:title, :body] == required_fields
+
+    required_fields =
+      changeset(%{})
+      |> filter_required([:title, :body])
+
+    assert [] == required_fields
+  end
```
wojtekmach, line 1082: consider adding a test that raises when someone passed a field that doesn't exist.

lib/ecto/changeset.ex
```diff
@@ -2067,6 +2065,42 @@ defmodule Ecto.Changeset do
...
+  end
+
+  defp invalid_required_field!(changeset, types, field) do
+    ensure_field_not_many!(types, field) && missing?(changeset, field) &&
```
wojtekmach, line 2100: `ensure_field_not_many!` warns with a message mentioning calling `validate_required` which would not be correct when the user called filter_required. Could you look into that?

greg-rychlewski: Thanks good catch. How does the change I made look?

lib/ecto/changeset.ex
```diff
@@ -2067,6 +2065,41 @@ defmodule Ecto.Changeset do
     end
   end
 
+  @doc """
+  Returns a list of fields that are present.
```
josevalim, line 2069:
```suggestion
  Returns which fields in a list would pass required validations (the fields that are present).
```

I am worried that the name is not obvious, given your original commit. The fact we don’t mention “required” in the description can also be considered a flag. Is this description clearer? We could call it `filter_present` but I don’t want to introduce this terminology to the API. Thoughts?

greg-rychlewski: Your wording is definitely better.

I wouldn't want to introduce "present" into the function names either. I was struggling to think of a clearer function name though. It wasn't the most intuitive to me when I first read it.

The only other ones I could think of were `keep_required` or `take_required`. Maybe they make the intent clearer but I don't know if `keep` is a verb used in Elixir and `take` is associated with maps iirc.

josevalim (approved):
LGTM! Although we should probably collect a thumbs up from @wojtekmach and @v0idpwn before merging to see if we are all ok with the name. <3

test/ecto/changeset_test.exs
```diff
@@ -1061,6 +1061,21 @@ defmodule Ecto.ChangesetTest do
...
+
+    assert required == [:body]
+
+    # When unknown field
+    assert_raise ArgumentError, ~r/unknown field :bad in/, fn  ->
+      changeset(%{"title" => "hello", "body" => "something"})
+      |> filter_required(:bad)
+    end
+  end
+
```
wojtekmach, line 1078: could you add a test like this: `changeset(%Foo{x: 1}, %{y: 2}); …; filter_required(changeset, [:x, :y])`, in other words, we look at both changeset.data and changeset.changes, right?

wojtekmach (approved):
Naming sounds good to me!

test/ecto/changeset_test.exs
greg-rychlewski: thanks, done!

v0idpwn (approved):
Looks good to me.

On a note: `filter_required/2` makes me think its related to the `changeset.required` value. Maybe `filter_set_fields/2` would reflect the intent better? Of course, depending on what you think :)

greg-rychlewski:
>On a note: filter_required/2 makes me think its related to the changeset.required value

That's a really good point.  I could see people thinking it should filter the ones inside of that field.

The only thing about `filter_set_fields/2` is you could technically set it as `nil` or empty string and it would fail the validation. Now this is making me think `filter_present` is actually the most clear. 

@josevalim do you think it's worth it to introduce that term given @v0idpwn's observation? I thought about `filter_non_empty`  but I don't know if this will get confused with `empty_values`.

josevalim:
I like filter_non_empty. I think the docs can quickly clarify that it is matching of the values precisely against empty_values.

greg-rychlewski:
>I like filter_non_empty. I think the docs can quickly clarify that it is matching of the values precisely against empty_values.

Sorry just wanna make sure I don't misinterpret. Did you mean:

- changing this behaviour of the function to check for `nil` and inclusion in `Ecto.Changeset.empty_values/0` (right now this is a bit different than `validate_required/3` because the latter is not trimming white space anymore)
- callling it `filter_non_empty` but pointing out the the docs it's not exactly the same thing as `empty_values`

greg-rychlewski:
I’m not sure if empty_values is lining up completely with this function since it’s not lining up with validate_required perfectly. 

Maybe filter_required is the best by virtue of it having the smallest chance of confusion/misuse? 

I’m inclined to keep what we have but let me know if anyone thinks strongly otherwise.

josevalim:
I meant to keep the behavior but note on the name. Everything that is empty will not be added to the changes anyway. From them on, we just check as nil, right?

josevalim:
Alternatively we simplify it even further and add `field_missing?` which receives a changeset and see if the field is not present as change and not nil. It is similar to this function but we push the Enum.filter to the caller.

josevalim (approved):

josevalim:
I am happy with this, I think we finally landed somewhere!

greg-rychlewski:
Yeah this is pretty perfect. Thank you for suggesting!

PR #4115, opened 2023-02-27
add dependency docs for telemetry

rands0n:
Not sure if this is right. But reading the Repo section about telemetry. The `telemetry` library is not linked accordingly.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -47,7 +47,7 @@ defmodule Ecto.Repo do
       Defaults to `10`
 
     * `:telemetry_prefix` - we recommend adapters to publish events
-      using the `Telemetry` library. By default, the telemetry prefix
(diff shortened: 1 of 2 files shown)


josevalim:
That won't work, unfortunately. The way to go about it is to write `` [`:telemetry`](`:telemetry`) `` instead of only `` `:telemetry` ``. Can you please update or send a new PR?

wojtekmach:
Also, please do test docs changes using `mix docs` locally :)

rands0n:
@wojtekmach for some reason I don't know the `mix docs` is not working locally.

image

Will see later what's happening

wojtekmach:
in Ecto, ExDoc is only for :docs env. Try: `MIX_ENV=docs mix docs`.

rands0n:
Ok, that worked. Updated! 🙏🏼

josevalim:

PR #4116, opened 2023-03-01
Fix `Repo.exists?/2` prefix/schema example

smaximov:
The example mentioned an invalid option `:schema` instead of `:prefix`.


wojtekmach:
Thanks!

PR #4118, opened 2023-03-01
Add Changeset.changed?/2,3

### LINKED ISSUE #4117: Add Ecto.Changeset.changed?(changeset, field, filters \\ [])
### Elixir version

Any

### Database and Version

Any

### Ecto Versions

All

All

### Current behavior

No function exists.

### Expected behavior

I would like to add `Ecto.Changeset.changed?(changeset, field)` that returns a boolean if the field has changed. You can pass additional filters that will further restrict the verification. For example:

    Ecto.Changeset.changed?(changeset, :role, from: :admin)

will return true only if the field was previously an admin.

The following returns true only if the field was nilified:

    Ecto.Changeset.changed?(changeset, :role, to: nil)

Or only if the role changed from admin to editor:

    Ecto.Changeset.changed?(changeset, :role, from: :admin, to: :editor)

I feel this will make understanding how a field changed much more accessible. It doesn't cover some cases (the field was something BUT admin), but it should be generally useful.

v0idpwn:
Closes #4117

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -478,17 +478,41 @@ defmodule Ecto.Changeset do
   """
   @spec changed?(t, atom, Keyword.t) :: boolean
   def changed?(%Changeset{} = changeset, field, opts \\ []) when is_atom(field) do
-    case fetch_change(changeset, field) do
-      {:ok, new_value} ->
-        Enum.all?(opts, fn
-          {:from, from} -> Map.get(changeset.data, field) == from
-          {:to, to} -> new_value == to
-          other -> raise ArgumentError, "unknown option #{inspect(other)}"
-        end)
+    case Map.fetch(changeset.types, field) do
+      {:ok, type} ->
+        case fetch_change(changeset, field) do
+          {:ok, new_value} ->
+            case type do
+              {tag, _relation} when tag in @relations ->
+                if opts != [], do: raise ArgumentError, "invalid options for #{tag} field"
+
+                if is_list(new_value) do
+                  Enum.any?(new_value, association_changed?(new_value))
+                else
+                  association_changed?(new_value)
+                end
+              _ ->
+                Enum.all?(opts, fn
+                  {:from, from} ->
+                    Ecto.Type.equal?(type, Map.get(changeset.data, field), from)
+                  {:to, to} ->
+                    Ecto.Type.equal?(type, new_value, to)
+                  other ->
+                    raise ArgumentError, "unknown option #{inspect(other)}"
+                end)
+            end
+
+          :error ->
+            false
+        end
 
       :error ->
-        false
+        raise ArgumentError, "field #{inspect(field)} doesn't exist"
     end
+  end
+
+  defp association_changed?(changeset) do
+    changeset.action != :update or changeset.changes != %{}
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -454,6 +454,59 @@ defmodule Ecto.Changeset do
     end)
   end
 
+  @doc """
+  Returns true if a field was changed in a changeset
```
wojtekmach, line 458:
```suggestion
  Returns true if a field was changed in a changeset.
```

lib/ecto/changeset.ex
```diff
@@ -454,6 +454,59 @@ defmodule Ecto.Changeset do
...
+      true
+
+      iex> changed?(changeset, :title, to: "NEW TITLE")
+      false
+  """
+  @spec changed?(t, atom, Keyword.t) :: boolean
+  def changed?(%Changeset{} = changeset, field, opts \\ []) when is_atom(field) do
+    case fetch_change(changeset, field) do
+      {:ok, new_value} ->
+        changed_from?(changeset, field, opts) and changed_to?(new_value, opts)
```
josevalim, line 483:
```suggestion
        Enum.all?(opts, fn
          {:from, from} -> Map.get(changeset.data, field) == from
          {:to, to} -> new_value == to
          other -> raise ArgumentError, "unknown option #{inspect(other)}"
        end)
```

greg-rychlewski:
Sorry for the bad memory. Is it possible that the change associated with an association is a changeset? If so should we have special behaviour for associations or maybe even just raise?

lib/ecto/changeset.ex
v0idpwn: That's clever. I was thinking that there should be a better way to do it :)

josevalim:
@greg-rychlewski Good point about associations/embeds!

1. For relation == :one, we check if `action != :update or changes != %{}` 
2. For relation == :many, we check if at least one `action != :update or changes != %{}` 

However, we should raise if `:from` or `:to` are given to an association/embed.

lib/ecto/changeset.ex
```diff
@@ -454,6 +454,43 @@ defmodule Ecto.Changeset do
...
+      iex> changed?(changeset, :title, to: "NEW TITLE")
+      false
+  """
+  @spec changed?(t, atom, Keyword.t) :: boolean
+  def changed?(%Changeset{} = changeset, field, opts \\ []) when is_atom(field) do
+    case fetch_change(changeset, field) do
+      {:ok, new_value} ->
+        Enum.all?(opts, fn
+          {:from, from} -> Map.get(changeset.data, field) == from
+          {:to, to} -> new_value == to
```
v0idpwn, line 485: Another note: should do semantic comparison for `:from` and `:to`

josevalim: Because of Decimal and stuff, right? Good call.

josevalim: @v0idpwn ping just in case you forgot. if you are busy, no worries, i would still wait 2-4 weeks for a new release.

v0idpwn:
I pushed a new version that (I believe) handles all cases, but it hurts my aesthetic sense

josevalim (approved):
Beautiful!

Do we have any tests for Decimal or anything that uses Ecto.Type.equal? If so, ship it!

PR #4125, opened 2023-03-08
Constraint match regex

nirev:
## Why

I'm currently faced with a situation where we have a partitioned table in postgres where indexes are being generated with increased numbers in the middle and at the end, making both prefix and suffix match insufficient. We cannot change how this is defined at DB level.

Eg: `partition_table_XXX_index_name_idx_YY`

## How

This PR adds a `:regex` matcher for constraints, when doing so the constraint `name` is used to compile a regex and then check if the raised constraint matches it

Let me know if this is an acceptable solution.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -346,7 +346,7 @@ defmodule Ecto.Changeset do
   @type error :: {String.t, Keyword.t}
   @type action :: nil | :insert | :update | :delete | :replace | :ignore | atom
   @type constraint :: %{type: :check | :exclusion | :foreign_key | :unique,
-                        constraint: String.t, match: :exact | :suffix | :prefix | :regex,
+                        constraint: String.t | Regex.t, match: :exact | :suffix | :prefix,
                         field: atom, error_message: String.t, error_type: atom}
   @type data :: map()
   @type types :: map()
@@ -361,7 +361,7 @@ defmodule Ecto.Changeset do
   }
 
   @relations [:embed, :assoc]
-  @match_types [:exact, :suffix, :prefix, :regex]
+  @match_types [:exact, :suffix, :prefix]
 
   @doc """
   Wraps the given data in a changeset or adds changes to a changeset.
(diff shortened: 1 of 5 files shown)


josevalim:
I am :+1: for this feature but I think that we should simply allow a regex to be given and raise if the `:match` option is given alongside a regex (since the regex already specifies it is a partial match).

nirev:
@josevalim you mean something like this?

```elixir
changeset = change(%Post{}) |> unique_constraint(:title, name: ~r/whatever\d+/", message: "is taken")
    assert constraints(changeset) ==
          [%{type: :unique, field: :title, constraint: "whatever\\d+", match: :regex, error_message: "is taken", error_type: :unique}]

    assert_raise ArgumentError, ~r/match type cannot be set if name is a regex/, fn ->
      change(%Post{}) |> unique_constraint(:title, name: ~r/whatever\d+/, match: :regex, message: "is taken")
    end
```

or also changing constraints to be able to store constraints as String or Regex ?:
```elixir
@type constraint :: %{type: :check | :exclusion | :foreign_key | :unique,
                        constraint: String.t | Regex.t, match: :exact | :suffix | :prefix | :regex,
                        field: atom, error_message: String.t, error_type: atom}
```

josevalim:
Yes! You may need to update adapters in ecto_sql accordingly.

lib/ecto/exceptions.ex
```diff
@@ -286,7 +286,17 @@ defmodule Ecto.ConstraintError do
...
+              name =
+                case c.constraint do
+                  name when is_binary(name) ->
+                    name
+
+                  %Regex{} = r ->
+                    inspect(r)
+                end
+              "    * #{name} (#{c.type}_constraint)"
+            end)
```
nirev, line 299: Regex doesn't implement `String.Chars`, so when storing the constraint as a String | Regex, we need to do this
I was torn between using `inspect(r)` or `Regex.source(r)`

nirev:
@josevalim Modified it so a Changeset constraint can now be a String or a Regex.

I checked `ecto_sql` and it didn't find any references to using `Changeset.constraints` directly

Let me know if there's any other changes required!

lib/ecto/exceptions.ex
josevalim, line 299:
```suggestion
            Enum.map_join(constraints, "\n", &"    * #{inspect(&1.constraint)} (#{&1.type}_constraint)")
```

lib/ecto/changeset.ex
```diff
@@ -3272,10 +3311,15 @@ defmodule Ecto.Changeset do
...
   def exclusion_constraint(changeset, field, opts \\ []) do
-    constraint = opts[:name] || "#{get_source(changeset)}_#{get_field_source(changeset, field)}_exclusion"
+    name = opts[:name] || "#{get_source(changeset)}_#{get_field_source(changeset, field)}_exclusion"
     message    = message(opts, "violates an exclusion constraint")
     match_type = Keyword.get(opts, :match, :exact)
-    add_constraint(changeset, :exclusion, to_string(constraint), match_type, field, message, :exclusion)
+    constraint = case name do
+      %Regex{} -> name
+      _ -> to_string(name)
+    end
```
josevalim, line 3320: Let's encapsulate this case into a tiny function called `normalize_constraint/1`.

josevalim:
> I checked ecto_sql and it didn't find any references to using Changeset.constraints directly

You are right! We were smart enough to encapsulate this :) I pushed some minor comments and we can ship it!

lib/ecto/changeset.ex
nirev: Done!

josevalim:

Issue #4124, opened 2023-03-08
Aliased modules names and nested embedded schemas


josevalim:
### Elixir version

1.14

### Database and Version

Any

### Ecto Versions

All?

### Database Adapter and Versions (postgrex, myxql, etc)

Any

### Current behavior

From: https://groups.google.com/g/elixir-ecto/c/rxZT_kzPp2o

Today we got an issue that was very hard to understand the root cause, and maybe a compilation warning or failure would help when defining Ecto schema.

Let's say we had a module like this:

```elixir
defmodule Blog do
  use Ecto.Schema
 
  embeds_many :comments, Comment do
    field :text, :string
  end
...
end
```

It will generate two modules Blog, and Blog.Comment.

Later, an unware developer adds an alias:

```
 defmodule Blog do
   use Ecto.Schema
+  alias Utils.Comment
 
   embeds_many :comments, Comment do
     field :text, :string
   end
 ...
  end
```

The Blog.Comment will not exist anymore, and the developer will have a failure telling that Blog.Comment doesn't exist anymore in another file.

### Expected behavior

In Elixir, defining a module always overrides the previous alias. Perhaps Ecto should preserve the same behaviour as Elixir.

greg-rychlewski:
I was trying to look into this. 

From what I can tell the difference between `defmodule` and `Module.create` in this situation is that eventually `defmodule` passes a tuple into `:elixir_module.compile` of the form `{full, old, new}` which is capturing the alias. But `Module.create` is just passing the module name.

Does it seem like I'm on the right track? Right now `Module.create` cannot accept a tuple and not sure if there is a workaround.

josevalim:
Both :elixir_module.compile and Module.create already expect the expanded name, so I think we need to make sure we are properly expanding the name in Ecto. The logic is here: https://github.com/elixir-lang/elixir/blob/50741091baa10170805730ed91943c481b7ab270/lib/elixir/lib/kernel.ex#L4971-L4993 WDYT?

josevalim:
PS: we can use `Module.concat` instead of the private `:elixir_aliases`

PR #4127, opened 2023-03-09
Embedded schemas overwrite conflicting aliases

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4124

Properly expands the nested embed module name so that it won't be overwritten by conflicting aliases.

I think we only need to care about expanding nested module aliases so I didn't consider the other cases from: https://github.com/elixir-lang/elixir/blob/50741091baa10170805730ed91943c481b7ab270/lib/elixir/lib/kernel.ex#L4971-L4993.

But this is not my strong suit so please correct me if I'm mistaken.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -2387,7 +2387,7 @@ defmodule Ecto.Schema do
     end
   end
 
-  defp expand_nested_module_alias({:__aliases__, _, [h | t]} = m, env) when is_atom(h) do
+  defp expand_nested_module_alias({:__aliases__, _, [h | t]}, env) when is_atom(h) do
     module = Module.concat([env.module, h])
 
     case t do
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -2387,11 +2387,11 @@ defmodule Ecto.Schema do
     end
   end
 
-  defp expand_nested_module_alias(expanded, _env) when is_atom(expanded), do: expanded
-
   defp expand_nested_module_alias({:__aliases__, _, [Elixir, _ | _] = alias}, _env),
     do: Module.concat(alias)
 
   defp expand_nested_module_alias({:__aliases__, _, [h | t]}, env) when is_atom(h),
     do: Module.concat([env.module, h | t])
+
+  defp expand_nested_module_alias(other, _env), do: other
 end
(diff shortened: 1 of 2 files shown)


lib/ecto/schema.ex
```diff
@@ -2385,4 +2386,13 @@ defmodule Ecto.Schema do
...
     end
   end
+
+  defp expand_nested_module_alias({:__aliases__, _, [h | t]}, env) when is_atom(h) do
+    module = Module.concat([env.module, h])
+
+    case t do
+      [] -> module
+      _ -> String.to_atom(Enum.join([module | t], "."))
+    end
```
josevalim, line 2396: I think you can write it as `Module.concat([env.module, h | t])`?

You also need to handle two cases:

1. `embeds_many :foo, Elixir.MyModule` - In this case, just return the alias is, as Elixir means defining a new root module

2. `embeds_many :foo, :"Elixir.FooBar"` - the module is hardcoded as an atom. may come from a macro that has already expanded the module

josevalim:
Nice start! I dropped some comments for additional cases to watch out for!

lib/ecto/schema.ex
greg-rychlewski: Thanks I've updated the PR with these changes. Just for my understanding, would you be able to tell me if I am "getting it":

Is prefixing a module name with `Elixir.` is basically an accepted convention that it will create a root module? For instance if I didn't treat it specially, `Module.create` would have no issue creating nested module `Root.Elixir.MyModule`?

josevalim: It is more than a convention. It is a compiler "rule". For example, if you do:

    alias List, as: MyList

`MyList` will resolve to `List` but `Elixir.MyList` always resolves to `MyList`.

> For instance if I didn't treat it specially, Module.create would have no issue creating nested module Root.Elixir.MyModule?

Yes, you can still have Elixir in the middle of the module name (although it is probably a confusing idea to do so!).

josevalim (approved):

lib/ecto/schema.ex
greg-rychlewski: Thank you this has been really interesting/insightful.

josevalim: And thank you for the PR <3

PR #4134, opened 2023-03-26
Allow user to provide custom error messages to `Ecto.Changeset.cast/4`

greg-rychlewski:
I was reading Elixir Forum today and came across this: https://elixirforum.com/t/how-do-you-provide-custom-error-message-for-ectos-cast-errors/54837.

```
If you provide a string to an :integer field you get
errors: [count: {"is invalid", [type: :integer, validation: :cast]}]
in a changeset, which is not helpful for a user.
```

It seems like a valid request IMO so I created this PR for feedback.

The implementation I chose is to have an option to define messages for whichever fields you want. Some other methods I thought of and discarded:

- Providing a function to produce the error message. It's a bit more error-prone for the user because they have to define their own default instead of relying on Ecto to provide the default. Also you can define your own function and use it to create the option in this implementation.
- Passing ` :field | {:field, error_msg}` into the cast function instead of using an option. It seems a bit messier/weird compared to existing APIs.


josevalim:
Thanks for bringing it up @greg-rychlewski!

Honestly, I am not sure. It feels like we have enough options to tackle this:

1. You can do post processing
2. You can provide custom message with custom types

Thoughts?

josevalim:
If we want to provide this, then perhaps we allow an error generation function to be given? We give it the metadata and it returns the message? But then, I am not sure how well it would work with errors from custom types.

greg-rychlewski:
>But then, I am not sure how well it would work with errors from custom types.

I was thinking similarly...not sure how easy it is for the user to catch all the possible cases.

> You can do post processing

Yeah tbh I was seeing this as a convenience over the post-processing route. Similar to the other functions but more awkward because this function takes several fields.

I'm pretty much on the fence...I can see its use but also can see it's an unusual/awkward option.

lib/ecto/changeset.ex
```diff
@@ -704,11 +724,15 @@ defmodule Ecto.Changeset do
...
-          |> Keyword.pop(:message, "is invalid")
+          {default_message, custom_errors} = Keyword.pop(custom_errors, :message, "is invalid")
+          message = messages[key] || default_message
+
+          new_errors =
+            custom_errors
+            |> Keyword.put_new(:validation, :cast)
+            |> Keyword.put(:type, type)
+
+          {message, new_errors}
```
v0idpwn, line 735:
```suggestion
        {default_message, custom_errors} = Keyword.pop(custom_errors, :message, "is invalid")
        message = messages[key] || default_message

        new_errors =
          custom_errors
          |> Keyword.put_new(:validation, :cast)
          |> Keyword.put(:type, type)
```

v0idpwn:
In one side, I think that is weird that we return an error **string** and don't provide a way to customize it. Feels a bit inconsistent with validations. In the other hand, `Ecto.Changeset.traverse_errors/2` already exists and gives you more flexibility.

josevalim:
Another downside of this approach is that if you want to change all messages, you would need to list all fields you are passing in. So I think a more general approach would be better but then a general approach is not much better than traversing it after the fact, right? So I am really unsure.

greg-rychlewski:
What about a function that takes in just the field name and then in the docs put a warning that if their function fails for any of the casted fields it will error? Then you could do both...

```elixir
defp cast_errors(_), do: "is not castable"

@cast_errors [
  field1: "field 1 error",
  field2: "field 2 error"
]

defp cast_errors(key) do
   @cast_errors[key]
end
```

josevalim:
Maybe the metadata as well? Is it also called for cast errors from custom types?

greg-rychlewski:
I think it makes sense to let them overwrite custom types as well if they don't like the message.

By metadata do you mean the type information?

josevalim:
metadata is the additional keyword list with the error message!

greg-rychlewski:
Ah I understand now. What do you think about this?

greg-rychlewski:
They probably don't need the `:validation` key in the metadata but I thought adding the `:type` helps in case they want to generalize...i.e. all `:string` have this message. And then the old message as like a fallback.

greg-rychlewski:
And then I guess the debate is: is this worth it or just do something similar in `traverse_errors`? It could be nice to have this alongside your schema so it just works™. Maybe you are calling `traverse_errors` differently in a few places and need to remember to format these errors properly everywhere.

lib/ecto/changeset.ex
```diff
@@ -604,6 +610,15 @@ defmodule Ecto.Changeset do
       iex> changeset.params
       %{}
 
+  You can define a custom error message function.
+
+      iex> params = %{title: 1, body: ""}
+      iex> custom_errors = [title: "must be a string"]
+      iex> error_msg_func = fn field, meta -> custom_errors[field] || meta[:message]
```
josevalim, line 617:
```suggestion
      iex> error_msg_func = fn field, meta -> custom_errors[field] || meta[:message] end
```

josevalim: Although instead of passing the message, I would suggest to allow the function to return nil and, if nil is returned, it uses the default one?

greg-rychlewski: thanks that's much nicer. updated.

PR #4135, opened 2023-03-26
Reduce macro generated code in Ecto.Schema

v0idpwn:
This behaviour is already present in other `Ecto.Schema` macros.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -757,30 +757,7 @@ defmodule Ecto.Schema do
   """
   defmacro timestamps(opts \\ []) do
     quote bind_quoted: binding() do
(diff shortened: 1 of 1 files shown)


lib/ecto/schema.ex
```diff
@@ -2002,6 +1979,32 @@ defmodule Ecto.Schema do
     end
   end
 
+  @doc false
+  def __define_timestamps__(mod, timestamps) do
```
v0idpwn, line 1983: Couldn't name it `__timestamps__` (like `__field__` or `__has_one__`) because we already have a `__timestamps__` function in this module, used for timestamp generation.

josevalim (approved):

PR #4138, opened 2023-03-27
[DRAFT] Clarify required preload on updating assoc

feynmanliang:
Not sure if my understanding is correct here, so please confirm before merging.

AFAICT the `preload` is mandatory and omitting it leads to a `RuntimeError`] raised at https://github.com/elixir-ecto/ecto/blob/a67675a6437fd8d16d6c4686e080ccff0e6bb560/lib/ecto/changeset/relation.ex#L74


feynmanliang:
NVM, don't know what I was thinking...

PR #4139, opened 2023-03-28
Add `map/1` to `Ecto.Query.API`

greg-rychlewski:
**Motivation** 

It seems it helps with LiveBook: https://groups.google.com/g/elixir-ecto/c/8E8MGOma-XM. 

**Implementation**

There were 2 options I thought were possible:

1. Change the select `expr` to signify it is a map.
2. Change the `take` value to signify we want all the fields

This PR was done with option 2. I thought the blast radius would be a lot larger for option 1.

Basically I am setting take to be `nil` if `map/1` is used. I believe this is safe because otherwise it's guaranteed to be a list. Also `nil` is safer over other atoms because sometimes `Access.fetch` is called on `take` to find association fields.

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -362,8 +362,8 @@ defmodule Ecto.Query.Builder.Select do
   defp merge(query, select, old_expr, old_params, old_subqueries, old_take, old_aliases, new_select) do
     %{expr: new_expr, params: new_params, subqueries: new_subqueries, take: new_take, aliases: new_aliases} = new_select
 
-    {old_expr, old_take} = expand_map(query, old_expr, old_take)
-    {new_expr, new_take} = expand_map(query, new_expr, new_take)
+    {old_expr, old_take} = expand_map!(query, old_expr, old_take)
+    {new_expr, new_take} = expand_map!(query, new_expr, new_take)
 
     new_expr =
       new_expr
@@ -436,7 +436,7 @@ defmodule Ecto.Query.Builder.Select do
     %{query | select: select}
   end
 
-  defp expand_map(query, {:map, _, [{:&, _, [ix]} = amp]}, take) do
+  defp expand_map!(query, {:map, _, [{:&, _, [ix]} = amp]}, take) do
     take =
       case get_source(query, ix) do
         {_, schema} when schema != nil ->
@@ -450,7 +450,7 @@ defmodule Ecto.Query.Builder.Select do
     {amp, take}
   end
 
-  defp expand_map(_, expr, take), do: {expr, take}
+  defp expand_map!(_, expr, take), do: {expr, take}
 
   defp get_source(query, ix) do
     Enum.at([query.from | query.joins], ix).source
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -444,7 +444,7 @@ defmodule Ecto.Query.Builder.Select do
           add_take(take, ix, {:map, fields})
 
         _ ->
-          take
+          raise "map/1 requires a source with a schema"
       end
 
     {amp, take}
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -107,8 +107,15 @@ defmodule Ecto.Query.Builder.Select do
        when tag in [:map, :struct] and is_atom(var) and is_atom(context) do
     taken = escape_fields(fields, tag, env)
     expr = Builder.escape_var!(var, vars)
-    acc = add_take(acc, Builder.find_var!(var, vars), {tag, taken})
-    {expr, {params, acc}}
+    take = add_take(acc.take, Builder.find_var!(var, vars), {tag, taken})
+    {expr, {params, %{acc | take: take}}}
+  end
+
+  # map(var)
+  defp escape({:map, _, [{var, _, context}]}, {params, acc}, vars, _env)
+       when is_atom(var) and is_atom(context) do
+    expr = Builder.escape_var!(var, vars)
+    {{:{}, [], [:map, [], [expr]]}, {params, acc}}
   end
 
   # aliased values
@@ -163,6 +170,13 @@ defmodule Ecto.Query.Builder.Select do
             "(3) map with dynamic values"
         )
     end
+  end
+
+  @doc """
+  Validates and adds fields to the take map.
+  """
+  def add_take(take, key, value) do
+    Map.update(take, key, value, &merge_take_kind_and_fields(key, &1, value))
   end
 
   @doc """
@@ -348,6 +362,9 @@ defmodule Ecto.Query.Builder.Select do
   defp merge(query, select, old_expr, old_params, old_subqueries, old_take, old_aliases, new_select) do
     %{expr: new_expr, params: new_params, subqueries: new_subqueries, take: new_take, aliases: new_aliases} = new_select
 
+    {old_expr, old_take} = expand_map(query, old_expr, old_take)
+    {new_expr, new_take} = expand_map(query, new_expr, new_take)
+
     new_expr =
       new_expr
       |> Ecto.Query.Builder.bump_interpolations(old_params)
@@ -419,6 +436,26 @@ defmodule Ecto.Query.Builder.Select do
     %{query | select: select}
   end
 
+  defp expand_map(query, {:map, _, [{:&, _, [ix]} = amp]}, take) do
+    take =
+      case get_source(query, ix) do
+        {_, schema} when schema != nil ->
+          fields = schema.__schema__(:query_fields)
+          add_take(take, ix, {:map, fields})
+
+        _ ->
+          take
+      end
+
+    {amp, take}
+  end
+
+  defp expand_map(_, expr, take), do: {expr, take}
+
+  defp get_source(query, ix) do
+    Enum.at([query.from | query.joins], ix).source
+  end
+
   defp classify_merge({:&, meta, [ix]}, take) when is_integer(ix) do
     case take do
       %{^ix => {:map, _}} -> {:map, meta, :runtime}
@@ -456,11 +493,6 @@ defmodule Ecto.Query.Builder.Select do
     Macro.to_string(other)
   end
 
-  defp add_take(acc, key, value) do
-    take = Map.update(acc.take, key, value, &merge_take_kind_and_fields(key, &1, value))
-    %{acc | take: take}
-  end
-
   defp bump_subquery_params(new_params, old_subqueries) do
     len = length(old_subqueries)
 
@@ -482,7 +514,7 @@ defmodule Ecto.Query.Builder.Select do
           # If merging with a schemaless source, do nothing so the planner can take all the fields.
           case old_expr do
             {:&, _, [^binding]} ->
-              source = Enum.at([query.from | query.joins], binding).source
+              source = get_source(query, binding)
 
               case source do
                 {_, schema} when schema != nil ->
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -3,6 +3,7 @@ defmodule Ecto.Query.Planner do
   @moduledoc false
 
   alias Ecto.Query.{BooleanExpr, DynamicExpr, FromExpr, JoinExpr, QueryExpr, SelectExpr, LimitExpr}
+  alias Ecto.Query.Builder.Select
 
   if map_size(%Ecto.Query{}) != 21 do
     raise "Ecto.Query match out of date in builder"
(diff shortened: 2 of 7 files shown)


josevalim:
I wonder if implementation wise we could get all fields from the schema and make it equivalent to `map(p, all_fields_from_p)`?

greg-rychlewski:
I can take a look. I think when you are building the query it's not that simple to get the schema without some decent sized changes. But I will try more earnestly to make that work.

josevalim:
I was thinking we would expand like that in the planner. This way we don’t need to add the new `nil` cases. :)

greg-rychlewski:
Ah ok, so something like this?

1. Expand query.from.source before going into the select expression `escape/3` recursion?
2. Pass the expanded source into `escape/3`
3. When hitting `{:map, _, [var]}` turn it into `map/2` with all the source fields?

greg-rychlewski:
Sorry you said planner. Deleted my other comment. I get it now.

greg-rychlewski:
I think I got it working without too much damage. The biggest change is that we need to accumulate the `take` when prewalking through the query in the planner. Because `map/2` is essentially expanded to `{:&, _, [ix]}` plus adding the fields to the take.

lib/ecto/query/builder/select.ex
```diff
@@ -348,6 +362,9 @@ defmodule Ecto.Query.Builder.Select do
   defp merge(query, select, old_expr, old_params, old_subqueries, old_take, old_aliases, new_select) do
     %{expr: new_expr, params: new_params, subqueries: new_subqueries, take: new_take, aliases: new_aliases} = new_select
 
+    {old_expr, old_take} = expand_map!(query, old_expr, old_take)
+    {new_expr, new_take} = expand_map!(query, new_expr, new_take)
```
josevalim, line 366: Instead of expand it here, I think you could say in classify_merge that it has type `{:map, meta, :runtime}`.

lib/ecto/query/planner.ex
```diff
@@ -403,6 +404,13 @@ defmodule Ecto.Query.Planner do
     fields = subquery_source_fields(source)
     {keep_source_or_struct(source), subquery_fields(fields, ix)}
   end
+  defp subquery_select({:map, _, [{:&, _, [ix]}]}, take, query) do
+    fields = map_fields!(:select, query, ix)
+    take = Select.add_take(take, ix, {:map, fields})
+    {source, _} = source_take!(:select, query, take, ix, ix, %{})
+    fields = subquery_source_fields(source)
+    {keep_source_or_struct(source), subquery_fields(fields, ix)}
+  end
```
josevalim, line 413: I think for subqueries it doesn't matter. This can be treated exactly the same as `{:&, _, [ix]}` because we never return a struct or a map. Can you try something like:

```
  defp subquery_select({:map, _, [{:&, _, [_]} = entry]}, take, query) do
    subquery_select(entry, take, query)
  end
```

?

lib/ecto/query/planner.ex
```diff
@@ -1264,40 +1273,69 @@ defmodule Ecto.Query.Planner do
...
+    {ix, ix_expr, ix_query} = get_ix!(left, kind, query)
+    extra = if kind == :select, do: [type: type!(kind, ix_query, expr, ix, field)], else: []
+    field = field_source(get_source!(kind, ix_query, ix), field)
+    {{{:., extra ++ dot_meta, [ix_expr, field]}, meta, []}, expr, counter}
   end
 
-  defp prewalk(list, kind, query, expr, acc, adapter) when is_list(list) do
-    Enum.map_reduce(list, acc, &prewalk(&1, kind, query, expr, &2, adapter))
+  defp prewalk(other, _kind, _query, expr, counter, _adapter) do
+    {other, expr, counter}
```
josevalim, line 1328: Why did we need those changes to prewalker? In any case, the `acc` is opaque, so if you need a counter, you should include the counter in the accumulator. :)

lib/ecto/query/builder/select.ex
greg-rychlewski: Oh yeah I tried this but then it makes it less powerful than `map/2`. For instance you can do this with map/2:

`from c in Comment, select: map(c, [:id]), select_merge: [:dislikes]`

But if I don't expand `map/1` here it will complain that a source can't be merged into a map when doing this

`from c in Comment, select: map(c), select_merge: [:dislikes]`

lib/ecto/query/planner.ex
josevalim: I think I understand. `counter` is the previous `acc` but now you need to return the `take`. Let me think about it.

josevalim: Here is any idea: what if instead of changing `prewalk` to expand `map/1`, you change `collect_fields` to consider `map/1`? You can test it out this way:

1. Remove the `{:map, _, [_]}` clause from prewalk
2. Add the clause to `collect_fields`
3. If it works, you can revert all changes to prewalk (probably easier to revert the file :D)

greg-rychlewski: oh actually I tried the collect_fields route first and then I noticed something broke here https://github.com/elixir-ecto/ecto/blob/8c232ccc7169f1c18bd74966d9993e95f67b7e63/lib/ecto/query/planner.ex#L1387 particularly with preloads. i'll reproduce to get your opinion on it

josevalim: Hrm, it seems I should be the one giving it a try then. :)

lib/ecto/query/builder/select.ex
greg-rychlewski: I could make it something similar `{:map, meta, something_about_map/1}` to catch this case?

lib/ecto/query/planner.ex
greg-rychlewski: Sorry I couldn't remember the exact issue. I think it `was that when you do something like this

`from p in Post, join: c in Comment, select: map(p), preload: [:comments]`

the `from` take doesn't know it's a map so it errors trying to find `__struct__` somewhere in the post processor. But for `map/2` it knows it's a map because it's tagged that way in the `take` struct.

I could reproduce later tonight (my time) to make it easier for you. During the day I got a baby trying to press my computer key though =) so can't do anything too complicated atm.

greg-rychlewski: I think there was also an issue with missing fields because the `from` take didn't know about them.

josevalim: @greg-rychlewski perfect, no rush. if you can push a failing test case (which should be working right now), then it would help me explore this. :)

lib/ecto/query/planner.ex
greg-rychlewski: I remember why I did this now. If you try this integration test:

```elixir
@tag :map_1
  test "map/1" do
    p = TestRepo.insert!(%Post{})
    c = TestRepo.insert!(%Comment{post_id: p.id})
    q = from p in Post, select: map(p)
    IO.inspect TestRepo.all(subquery(q))
  end
```

It will return make a struct because the `take` is not tagging the fields as `:map`.

lib/ecto/query/planner.ex
greg-rychlewski: I created a new branch in my fork with the following:

1. Don't expand map/1 while calling the select merge function. Wait until the planner.
2. Don't expand map/1 while prewalking but handle it during `collect_fields`
3. Added a couple integration tests that pass in this branch but not with these changes

Please see here: https://github.com/greg-rychlewski/ecto/commit/0b40a751bcef7a143294059c77afd428a301746f

I trie to make it as clear as possible with my comments but please let me know if I can explain better. Thank you for looking at it.

josevalim: Thank you so much. We should definitely include those tests in the final merge!

I think the root cause of all or our problems is here:

https://github.com/elixir-ecto/ecto/pull/4139/files#diff-a4782321bba4a264f78ae4d81bc9be152d20af799129133dd078d59f2c1fb4d0R117-R118

Note that the handling of `map/struct` fully happens on the escape function. However, because we don't have access to fields here, we keep `map` as is and that's what's pushing complexity everywhere.

Maybe we can try this implementation?

```
  defp escape({tag, _, [{var, _, context}]}, {params, acc}, vars, env)
       when tag in [:map, :struct] and is_atom(var) and is_atom(context) do
    expr = Builder.escape_var!(var, vars)
    take = add_take(acc.take, Builder.find_var!(var, vars), {tag, :all})
    {expr, {params, %{acc | take: take}}}
  end
```

The idea is that we will have this `:all` annotation, which we will have to handle downstream, but perhaps without needing to change `prewalk`. We will no longer need `expand_map` but we will need to be able to merge `:all` (merging `anything` with all returns `anything`).

WDYT?

This is one of the most complex parts of Ecto, which makes doing a small change really hard, but it also one of the reasons why making a small change is important!

josevalim: I am also thinking we should remove the `take` system altogether. Instead of doing:

    %{:&, [], [ix]}

We could keep it in the AST as:

    {:map, _, [%{:&, [], [ix]}, [:foo, :bar]]}

In the planner, every time we see `%{:&, [], [ix]}` we immediately fetch its taken value anyway, and now we will no longer need to because it is immediately available. In `Select.merge`, we already have to consider both the expression and its take anyway, and we could then do it in a single pass.

Perhaps the smarter thing to do is to find a new representation altogether. We have four possible values:

  * sources
  * structs
  * maps
  * merge

And maybe we could represent them as (at the AST level)

  * source(ix, :map | :struct | :any, fields | :all)
  * struct(name, fields)
  * map(fields)
  * merge(any_of_the_four, map)

This should reduce the amount of back and forth between the merger, planner, and the postprocessor and the number of distinct representations.

josevalim: If we don't want to rock the boat too much we can keep the representation as is today except that:

* `&ix` - represents :any (of map or struct)
* `map(&ix, fields)` - represents map
* `struct(&ix, fields)` - represents struct
* `%{...}` - map
* `%name{...}` - struct
* `merge(left, right)`

It is the same as above except more flattened. If we move to this structure, then we can say `map/1` is equivalent to `map(&ix, :all)`.

greg-rychlewski: Sorry need to parse the other messages but I could reply to this one now :)

> The idea is that we will have this :all annotation, which we will have to handle downstream, but perhaps without needing to change prewalk. We will no longer need expand_map but we will need to be able to merge :all (merging anything with all returns anything).
> WDYT?

Oh yes actually this is similar to what I had first except I used `nil` instead of `:all` because sometimes `Access.fetch` is called on it. So that makes it work in those places without an extra case.

If you're ok with this being a possibility I could for sure bring that version back to look at. It's been a while but over here I think you weren't too thrilled about it https://github.com/elixir-ecto/ecto/pull/4139#issuecomment-1487571422

But if it is looking better now I could bring it back for consideration.

greg-rychlewski: >This should reduce the amount of back and forth between the merger, planner, and the postprocessor and the number of distinct representations.

josevalim: Awesome. So let's give a try at the refactoring. I was planning to give it a go but, if you could do it, even better, cause then we reduce the bus number. According to classify_merge, we have three cases:

  * &ix
  * %{...} - map
  * %name{...} - struct
  * merge(left, right)

BUT the first case actually breaks in three:

  * &ix - no map or struct specified in take
  * &ix - but map specified in take
  * &ix - but struct specified in take

My suggestion is to, when we see map/2 or struct/2 in escape (the last two cases above), we will keep keep the `map/2` and `struct/2` in the AST. Once we start doing this, you will need to change both merge!, subquery, and collect_fields, but once you do so, then we should no longer need `take`. There are some cleanups I see we can do before hand, so I will give them a try first.

josevalim: @greg-rychlewski I pushed two small fixes and some code comments to collect fields. I made `merge` itself recursive and removed an outdated `:never` check. This means that supporting `map/2` and `struct/2` on collect fields is easier, as we only need to add two clauses (one for map and another for struct) instead of several closes that consider how `merge` and other factors interplay.

greg-rychlewski: beautiful changes! I'l give the refactoring a try and push them here

josevalim: Let’s do the refactor in a separate PR and then add this functionality later so we don’t mix them. :)

greg-rychlewski: Btw sorry I forgot to ask yesterday. Do you mean completely remove the take system or just for map/structs? For instance for this:

```elixir
from p in Post, select: [:title, :id]
```

Would we want to represent it as something like this `{:source, _, [{:&, _, [ix]}, [:title, :id]]}`? I like the idea of completely getting rid of take but not sure if you saw a reason to keep this one.

josevalim: We will keep the system but instead of storing it as a separate field we keep all info in the AST. We can rewrite the case you posted as `map_or_struct(&ix, […])` (if we have :all we can wrap all sources in that but maybe once we do it we will see the benefits of keeping it separate).

josevalim:
@greg-rychlewski shall we close this one? :)

greg-rychlewski:
Yeah i am not sure when I will have time to return to it. If I do get time then I can submit a completed PR based on our discussion

PR #4142, opened 2023-03-30
Add support to sort_param and drop_param

josevalim:
In examples we typically pass -many style association as a list:

    %{"name" => "john doe", "addresses" => [
      %{"street" => "somewhere", "country" => "brazil", "id" => 1},
      %{"street" => "elsewhere", "country" => "poland"},
    ]}

However, it is also common to pass the addresses as a map, where each
key is an integer representing its position:

    %{"name" => "john doe", "addresses" => %{
      0 => %{"street" => "somewhere", "country" => "brazil", "id" => 1},
      1 => %{"street" => "elsewhere", "country" => "poland"}
    }}

Using indexes becomes specially useful with two supporting options:
`:sort_param` (embeds only) and `:drop_param`. These options tell
the indexes should be reordered or deleted from the data. For example,
if you did:

    cast_embed(changeset, :addresses,
      sort_param: :addresses_sort,
      drop_param: :addresses_drop)

You can now submit this:

    %{"name" => "john doe", "addresses" => %{...}, "addresses_drop" => [0]}

And now the entry with index 0 will be dropped from the params before casting.
Note this requires setting the relevant `:on_replace` option on your
associations/embeds definition.

Similar, for sorting, you could do:

    %{"name" => "john doe", "addresses" => %{...}, "addresses_sort" => [1, 0]}

And that will internally sort the elements so 1 comes before 0. Note that
any index not present in "addresses_sort" will come _before_ any of the
sorted indexes. If an index is not found, an empty entry is added in its
place.

These parameters can be powerful in certain UIs as it allows you to decouple
the sorting and replacement of the data from its representation.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1015,9 +1015,8 @@ defmodule Ecto.Changeset do
       }}
 
   Using indexes becomes specially useful with two supporting options:
-  `:sort_param` (embeds only) and `:drop_param`. These options tell
-  the indexes should be reordered or deleted from the data. For example,
-  if you did:
+  `:sort_param` and `:drop_param`. These options tell the indexes should
+  be reordered or deleted from the data. For example, if you did:
 
       cast_embed(changeset, :addresses,
         sort_param: :addresses_sort,
@@ -1039,6 +1038,27 @@ defmodule Ecto.Changeset do
   any index not present in "addresses_sort" will come _before_ any of the
   sorted indexes. If an index is not found, an empty entry is added in its
   place.
+
+  For embeds, this guarantees the embeds will be rewritten in the given order.
+  However, for associations, this is not enough. You will have to add a
+  `field :position, :integer` to the schema and then do a post-processing
+  of the association, something like this:
+
+      changeset
+      |> cast_assoc(:children, sort_param: ...)
+      |> copy_children_positions()
+
+      defp copy_children_positions(changeset) do
+        if children = Ecto.Changeset.get_change(changeset, :children) do
+          children
+          |> Enum.with_index(fn child, index ->
+            Ecto.Changeset.put_change(child, :position, index)
+          end)
+          |> then(&Ecto.Changeset.put_change(changeset, :children, &1))
+        else
+          changeset
+        end
+      end
 
   These parameters can be powerful in certain UIs as it allows you to decouple
   the sorting and replacement of the data from its representation.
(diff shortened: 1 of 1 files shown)


greg-rychlewski:
Would it be a good idea to have an option to raise on missing sort indexes? Just in case someone is making a mistake instead of intentionally trying to add blanks.

josevalim:
@greg-rychlewski good question. We are currently using missing entries as a mechanism to create new entries. We could add an option to either raise, create, or discard them, but there is the question of which one should be the default.

Discard them is certainly bad. There is an argument for raising at the cost of making it more bureaucratic. Creating a new one will be empty and most likely lead to validation errors, but can be surprising.

Another option is to raise but have a special value to indicate you want to create new entries, such as "new". However, because the indexes can be anything, there is a chance of conflict at least in theory, so we would need to make this opt-in (which perhaps is fine), something like: `sort_param: "addresses_sort", add_empty_entries_on: "new"`. We likely need a better name though.

Thoughts?

greg-rychlewski:
I like the one option that is either raise/discard/create, like you said. The special value to me seems like a bit too much work for not much gain. And introduces some new ways to make mistakes.

I would personally default to `raise` but if a large % of the people who would use this would mostly want to `create` I can see the case for that. I am not too much into UI work so I'm mostly concerned about introducing mistakes. It might be that no one using this for UI wants to raise. For sure I wouldn't discard it by default, 100% agree.

LostKobrakai:
I like this idea of "declarative" editing for one-to-many relationships. I'm not sure I like the asymmetry between using `*_drop` for dropping elements, but using `_sort` for creating elements.

Symmetrical would be using just `_sort` and any missing indexes are dropped and any additional indexes have an record created. Not sure if that's a sound idea.

The other symmetrical approach would be `_drop` and `_add`, but that's less flexible in terms of where new records should be sorted into the existing ones.

josevalim:

josevalim:
@LostKobrakai the rationale for being distinct is that you may want to expose different capabilities. Some UIs may only allow sorting. Others only deletions. And so forth.

LostKobrakai:
@josevalim That's a totally fine approach hence my two alternate suggestions. My concern was more about how sorting and creation are bundled into one capability, while deletion is separate. I don't know how to do improve on that front though without the downsides mentioned before.

josevalim:
@LostKobrakai perhaps it is a matter of documentation but adding is not bundled in sorting. You can "append" by simply submitting something like `parent[child][new][_]`. What addition on sort_param allow is arbitrary additions between any children but you _need_ sorting to be able to that anyway. :D

PR #4147, opened 2023-04-07
Allow changesets or schemas in Ecto.Enum values / mappings functions

noahduncan:
The current implementation of `Ecto.Enum.values/2`, `Ecto.Enum.dump_values/2`, and `Ecto.Enum.mappings/2` will only work for a module that uses `Ecto.Schema`.  However, Ecto supports schemaless changesets and so `Ecto.Enum` should too.

This PR is an idea that would allow `Ecto.Enum` to be used with a schemaless changeset by defining a type map as follows:

```elixir
types = %{
  permissions: {:array, Ecto.ParameterizedType.init(Ecto.Enum, values: [:read, :write])}
}
Ecto.Enum.values(types)
```

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -213,35 +213,41 @@ defmodule Ecto.Enum do
     "#Ecto.Enum<values: #{inspect Keyword.keys(mappings)}>"
   end
 
-  @doc "Returns the possible values for a given schema and field"
-  @spec values(module, atom) :: [atom()]
-  def values(schema, field) do
-    schema
+  @doc "Returns the possible values for a given schema or types map and field"
+  @spec values(map | module, atom) :: [atom()]
+  def values(schema_or_types, field) do
+    schema_or_types
     |> mappings(field)
     |> Keyword.keys()
   end
 
-  @doc "Returns the possible dump values for a given schema and field"
-  @spec dump_values(module, atom) :: [String.t()] | [integer()]
-  def dump_values(schema, field) do
-    schema
+  @doc "Returns the possible dump values for a given schema or types map and field"
+  @spec dump_values(map | module, atom) :: [String.t()] | [integer()]
+  def dump_values(schema_or_types, field) do
+    schema_or_types
     |> mappings(field)
     |> Keyword.values()
   end
 
-  @doc "Returns the mappings for a given schema and field"
+  @spec mappings(map, atom) :: Keyword.t()
+  def mappings(types, field) when is_map(types) do
+    case types do
+      %{^field => {:parameterized, Ecto.Enum, %{mappings: mappings}}} -> mappings
+      %{^field => {_, {:parameterized, Ecto.Enum, %{mappings: mappings}}}} -> mappings
+      %{^field => _} -> raise ArgumentError, "#{field} is not an Ecto.Enum field"
+      %{} -> raise ArgumentError, "#{field} does not exist"
+    end
+  end
+
   @spec mappings(module, atom) :: Keyword.t()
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
Thanks for the PR. 

I'm not sure the changeset is the schemaless equivalent in this case. I believe it should be the 2-tuple of the struct and the type information. Maybe we should accept that instead?

Some other comments:

- It looks like quite a few CI tests failed. Could you please have a look?
- Could you please add tests for the new functionality?

wojtekmach:
Thank you for the PR!

> ```elixir
> {:array, {:parameterized, Ecto.Enum, Ecto.Enum.init(values: [:read, :write])}}
> ```

Personally I'd like to see something more ergonomic:

```elixir
{:array, {Ecto.Enum, values: [:read, :write]}}
```

As an aside, I think I proposed it to be "the" syntax for defining enums on schemas too:

```elixir
field :mode, {:array, {Ecto.Enum, values: [:read, :write]}}
```

so we'd have nice symmetry there. But I can't find the discussion if there was one.

josevalim:
This is what to use for init: https://github.com/elixir-ecto/ecto/blob/v3.9.5/lib/ecto/parameterized_type.ex#L179

For schemaless change sets, we should pass the types as argument.

noahduncan:
Thanks for the quick responses, I'll work on implementing those next week.

> It looks like quite a few CI tests failed. Could you please have a look?

I'll see what I can figure out.

> Could you please add tests for the new functionality?

For sure.

lib/ecto/enum.ex
```diff
@@ -213,35 +213,47 @@ defmodule Ecto.Enum do
...
+  def mappings(%Ecto.Changeset{types: types}, field) do
+    case types do
       %{^field => {:parameterized, Ecto.Enum, %{mappings: mappings}}} -> mappings
       %{^field => {_, {:parameterized, Ecto.Enum, %{mappings: mappings}}}} -> mappings
       %{^field => _} -> raise ArgumentError, "#{field} is not an Ecto.Enum field"
       %{} -> raise ArgumentError, "#{field} does not exist"
     end
   end
+
+  @spec mappings(Ecto.Schema.t, atom) :: Keyword.t()
```
v0idpwn, line 248: This typespec isn't correct. `Ecto.Schema.t()` refers to a struct from a schema. This function receives a module. The same applies to the other typespecs.

v0idpwn (commented):
+1 for using {types, data} or types instead of a changeset, plus a note regarding typespecs.

lib/ecto/enum.ex
```diff
@@ -213,35 +213,39 @@ defmodule Ecto.Enum do
...
+      %{^field => _} -> raise ArgumentError, "#{field} is not an Ecto.Enum field"
+      %{} -> raise ArgumentError, "#{field} does not exist"
+    end
+  end
+
   @spec mappings(module, atom) :: Keyword.t()
   def mappings(schema, field) do
     try do
-      schema.__changeset__()
+      mappings(schema.__changeset__(), field)
```
josevalim, line 245: Instead of wrapping the mappings call as well, please try this:

```
try do
  schema.__changeset__()
rescue
  _ ->
    raise ArgumentError, "#{inspect(schema)} is not an Ecto schema or a types map"
else
  %{} = types -> mappings(types, field)
end
```

noahduncan: Done. I kept the original rescue clause, but let me know if you'd prefer the clause from your comment (without the match on `UndefinedFunctionError`) and I'll switch it up.

josevalim:

PR #4148, opened 2023-04-07
Use correct fields when select merging into a source

greg-rychlewski:
When select merging into a source there is a bit of a hiccup where you don't immediately know all of the source's fields. Last year there was a fix made for this: https://github.com/elixir-ecto/ecto/pull/3989.

The issue is that it always assumed the source you are merging into is the `from` source. I was playing around with this and saw there are situations where might have a `join` source (see the new test).

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -412,7 +412,7 @@ defmodule Ecto.Query.Builder.Select do
       select | expr: expr,
                params: old_params ++ bump_subquery_params(new_params, old_subqueries),
                subqueries: old_subqueries ++ new_subqueries,
-               take: merge_take(query.from.source, old_expr, old_take, new_take),
+               take: merge_take(query, old_expr, old_take, new_take),
                aliases: merge_aliases(old_aliases, new_aliases)
     }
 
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder/select.ex
```diff
@@ -480,12 +480,17 @@ defmodule Ecto.Query.Builder.Select do
           # If merging with a schema, add the schema's query fields. This comes in handy if the user
           # is merging fields with load_in_query = false.
           # If merging with a schemaless source, do nothing so the planner can take all the fields.
-          case {old_expr, source} do
-            {{:&, _, [^binding]}, {_source, schema}} when not is_nil(schema) ->
-              Map.put(acc, binding, {new_kind, Enum.uniq(new_fields ++ schema.__schema__(:query_fields))})
+          case old_expr do
+            {:&, _, [^binding]} ->
+              source = Enum.at([query.from | query.joins], binding).source
```
greg-rychlewski, line 485: I spent a long time trying to figure out if this is legitimate. From what I can tell it is alright.

It seems like `from` is always indexed at `0` and the joins are iterated through in this way. Also I couldn't find anywhere in the planner that was adjusting the `take` indexes.

greg-rychlewski:
The integration test failures look related to the ecto sql constraint changes and not here. I can look closer a bit later.

josevalim (approved):

PR #4154, opened 2023-04-12
support map(q, [:list, :of, :fields]) in source query of insert_all 

narrowtux:
It allows insert_all source queries to be written like this:

```elixir
fields = [:a, :foo, :baz]
query = from d in Data, select: map(d, ^fields)
Repo.insert_all(Stuff, query)
```

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -114,6 +114,11 @@ defmodule Ecto.Repo.Schema do
       %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}} ->
         Enum.map(args, &elem(&1, 0))
 
+      %Ecto.Query.SelectExpr{take: take} when is_map(take) ->
+        # we want to support any table index, so we just get the first entry
+        {:map, fields} = take |> Enum.at(0) |> elem(1)
+        fields
+
       _ ->
         raise ArgumentError, """
         cannot generate a fields list for insert_all from the given source query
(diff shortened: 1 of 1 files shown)


narrowtux:
I'll add support for `struct/1` (fields list from `__schema__(:fields)` in that case) and `struct/2` and add some tests.

narrowtux:
Seems like ecto checks for `map(q, [])` and `struct(q, [])` in another place already:
```
Expected exception ArgumentError but got Ecto.QueryError (at least one field must be selected for binding `0`, got an empty list in query:
     
     from m0 in Ecto.RepoTest.MySchema,
       select: map(m0, [])
     )
```

narrowtux:
I refactored the match once again, so we don't need to write that error message twice

lib/ecto/repo/schema.ex
```diff
@@ -110,10 +110,18 @@ defmodule Ecto.Repo.Schema do
...
+      %Ecto.Query.SelectExpr{expr: {:&, _, [ix]}} -> ix
+      _ -> nil
+    end
+
     header = case query.select do
       %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}} ->
         Enum.map(args, &elem(&1, 0))
 
+      %Ecto.Query.SelectExpr{take: %{^ix => {fun, fields}}} when not is_nil(ix) and fun in [:struct, :map] ->
+        fields
```
josevalim, line 123:
```suggestion
      %Ecto.Query.SelectExpr{take: %{^ix => {_fun, fields}}} ->
        fields
```

josevalim: This would also support the default case where fun is any.

josevalim:

PR #4159, opened 2023-04-17
Guard against invalid kw lists

jakub-gonet:
> a small check if the first element of the list is a tuple when non-empry

😄

### DIFF
--- a/test/ecto/repo_test.exs
+++ b/test/ecto/repo_test.exs
@@ -318,7 +318,7 @@ defmodule Ecto.RepoTest do
       assert query.prefix == "fallback_schema"
     end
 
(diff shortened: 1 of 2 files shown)


test/ecto/repo_test.exs
```diff
@@ -317,6 +317,13 @@ defmodule Ecto.RepoTest do
       assert query.from.source == {"my_parent", Ecto.RepoTest.MyParent}
       assert query.prefix == "fallback_schema"
     end
+
+    test "raises if options aren't kw list" do
```
josevalim, line 321:
```suggestion
    test "raises if options aren't a keyword list" do
```

josevalim:

PR #4160, opened 2023-04-17
Fix `prepare_opts` test

greg-rychlewski:
This was causing the CI to fail. If you have any time, was also wondering if you could let me know why it's emitting a `CaseClauseError`. Thanks!

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -248,7 +248,7 @@ defmodule Ecto.Repo do
         adapter.checked_out?(meta)
       end
 
-      @compile {:inline, get_dynamic_repo: 0, prepare_opts: 2}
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
Does this turn function heads into cases?

`@compile {:inline, get_dynamic_repo: 0, prepare_opts: 2}`

josevalim:
> Does this turn function heads into cases?

Yes!

josevalim (approved):

josevalim:
Actually, we should remove the inline. The CaseClauseError won't be as readable in error messages.

greg-rychlewski:
@josevalim How is this? Sorry there are some end of line whitespace removals.

josevalim (approved):

PR #4161, opened 2023-04-18
Update query.ex

TwistingTwists:
typo in preload queries for join `on`. 
instead of `and` it should use `where` for conditional expressions.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2306,7 +2306,8 @@ defmodule Ecto.Query do
       comments_query =
         from c in Comment,
         join: r in subquery(ranking_query),
-        on: c.id == r.id and r.row_number <= 5
+        on: c.id == r.id, 
+        where: r.row_number <= 5
 
       Repo.all from p in Post, preload: [comments: ^comments_query]
 


josevalim:
I believe both approaches are fine. Or is there a reason to prefer one over the other?

Defman21:
It all boils down to readability and some specific behavior when using left/right/outer joins. I think it's better to keep filter conditions in `where`, and as for left/right/outer:

For example, if we have:

```
TABLE posts(id serial, title varchar, body varchar)
TABLE comments(id serial, post_id fk(posts.id), body varchar, score int)
```

Doing something like this:

```sql
SELECT posts.id, comments.body 
FROM posts 
LEFT JOIN comments ON (posts.id = comments.post_id AND comments.score > 5)
```

Could produce something like this:


| posts.id | comments.body |
| --- | --- |
| 1 | hello |
| 1 | good comment |
| 2 | NULL |
| 3 | NULL |

While this query:

```sql
SELECT posts.id, comments.body 
FROM posts LEFT JOIN comments ON posts.id = comments.post_id 
WHERE comments.score > 5
```

Will return this data:

| posts.id | comments.body |
| --- | --- |
| 1 | hello |
| 1 | good comment |

greg-rychlewski:
>It all boils down to readability 

I have heard a lot of people with similar opinions so don't have a strong feeling either way. But to me the readability actually goes the opposite way. Putting it in the `on` clause is actually closer to what is happening. The planner pushing `where` to `on` is an optimization not reflected in the query text.

josevalim:
Closing this for now as there is no consensus and it seems to be a matter of preference. Thank you!

PR #4165, opened 2023-04-21
Enhance behaviour of alternative (or) in select

cichacz:
Hello!

I have interesting (I hope) contribution to the select functionality in Ecto.
In my project I was trying to explore ways to solve single/class table inheritance. 
It was looking fine but was missing one thing: Ecto couldn't guess what type it would be since it expects type to be known upfront.

So this pseudocode would return always one type (Details):
```
module Animal: 
    schema "animal": 
        field: number_of_legs,
        has_one: species_details, Details

module Dog: 
    schema "animal": 
        field: number_of_legs,
        has_one: species_details, DogDetails

module Cat: 
    schema "animal": 
        field: number_of_legs,
        has_one: species_details, CatDetails

Animal |> preload(species_details) |> Repo.all()
```

What I wanted was to have concrete type for `species_detail`
After long thinking I decided to try simplest solution that doesn't break many concepts (I hope 😅).

So I made this possible:
```
module Animal: 
    schema "animal": 
        field: number_of_legs,
        virtual field: species_details
        has_one: dog_details, DogDetails
        has_one: cat_details, CatDetails

Animal 
|> join(assoc(dog_details))
|> join(assoc(cat_details))
|> select_merge(species_details: cat_details or dog_details)
|> Repo.all()
```
This will return list of animals with `species_details` filled with appropriate struct responsible for specific species.

You can find some "dummy" examples in tests.

I'm aware that this **changes** current functionality of `or` in the `select`. Until now this would be sent to the DB and was evaluated there, with this change it still works for things like `a.active or b.active` but will be evaluated in code.
There are two reasons I think it's fine to change this behaviour:
1. One can easily do `fragment("? or ?")` to make it work as previously
2. I think (but may be wrong) that it's not very common to use `fieldA or fieldB` in select clause to get some value. One would rather use `coalesce`

Let me know your thoughts! 😄

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -111,6 +111,31 @@ defmodule Ecto.Query.Builder.Select do
     {expr, {params, acc}}
   end
 
+  # alternative (a or b)
+  defp escape({:or, _, [left, right]}, params_acc, vars, env) do
+    Macro.prewalk(left, fn
+      # if binding isn't direct then we're fine, no need to go that branch
+      {:., _, _} ->
+        nil
+
+      {binding, _, _} = ast when is_atom(binding) ->
+        if Keyword.get(vars, binding) == 0 do
+          Builder.error!(
+            "binding `#{Macro.to_string(ast)}` can be put only at the end of an alternative"
+          )
+        end
+
+        ast
+
+      other ->
+        other
+    end)
+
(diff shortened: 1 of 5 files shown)


cichacz:
Ok, that was fast :D Doing a second pass on the data adds some overhead, especially when it's a nested preload, so I thought that doing this at one pass would be better.

PR #4167, opened 2023-04-21
Add abstract schema

v0idpwn:
This PR attempts to improve the API for schemas without source (because they want a dynamic source). 

Before, you had to use a regular schema with a fake source and override it (described in the documentation, I changed it, check the diff).

Would need more tests, and I'm willing to put the work if you think the idea is decent. The only issue I see with this is that the `Ecto.Meta.source()` can now be `nil`, which means that code relying on it being `String.t()` can break.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -13,6 +13,7 @@ defmodule Ecto.Integration.RepoTest do
   alias Ecto.Integration.Barebone
   alias Ecto.Integration.CompositePk
   alias Ecto.Integration.PostUserCompositePk
+  alias Ecto.Integration.Abstract
 
   test "returns already started for started repos" do
     assert {:error, {:already_started, _}} = TestRepo.start_link()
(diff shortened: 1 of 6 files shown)


v0idpwn:
(Integration tests fail because they depend on a migration on ecto_sql, no actual code changes)

josevalim:
Thanks @v0idpwn but I think I am a -1. I don't think it is worth introducing a new concept, since embedded schemas achieve similar?

v0idpwn:
@josevalim We could adapt embed schemas to support this use-case, but at the moment they don't

josevalim:
Oh, I see what you mean. Yeah, in this case I would just define a bogus name. I think having something such as cannot find table "abstract: posts" is clearer than having to handle all of the complexity of the source being nil, no? Or is the benefit that we can force in the planner for a source to be defined and raise if not?

v0idpwn:
Yes! I had the impression the planner would already error in this situation, but it did not. Pushed a commit that addresses this.

v0idpwn:
I'm also not sure that adding a new schema type is a good trade-off here. I just thought that using the bogus name felt wrong and figured out it would be a small change in Ecto to provide a more correct way to do this.

v0idpwn:
@josevalim do you have any update on this? Maybe we should try some alternative approach to provide a better API for this use-case? Maybe improve/merge this? Maybe just throw it away?)

josevalim:
Yeah, honestly I don’t see the benefits to justify the code and API changes :)

PR #4170, opened 2023-04-25
Discriminated Union Type Embeds - Proof of Concept

bamorim:
For a long time I've been re-implementing some sort of this idea in multiple different ways, although most of the time I'd implement as an Ecto.Type that serializes, but I've been thinking that maybe this could be part of Ecto's embeds. Let me elaborate.

The idea is to allow embeds_one and embeds_many to contain different embedded schemas based on a discriminator (a "tag").

This is useful to allow people to model their domain (I like it because it makes invalid states harder and because it allows a more flexible schema). One of the use cases I've used a lot in the past was having a `EventLog` schema that would embed the event, which could be one of multiple different embedded schemas representing the event (like `UserSignedUp`, `OrderPlaced`, etc).

This is just a proof of concept to start sketching out the idea. My approach is to make the Embed relationship `:related` field also accept a `{:one_of, variants}`. Although maybe we could represent with another struct and some helper functions, like `Ecto.Embedded.OneOf`, but for now I just wanted to implement the basic of the idea since some of the Ecto internals are still pretty new to me.

The current solution also expects the dumped format and the "to-be-casted" format to have a type and data values, but maybe we should allow this to be customized (I just wanted to avoid that complexity for now).

I'm opening this PR to start a discussion around:
- Do we think this idea has value?
- If yes, how can I improve and change it to the point it is good enough?

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1254,6 +1254,21 @@ defmodule Ecto.Changeset do
 
   defp key_as_int(key_val), do: key_val
 
+  defp on_cast_default(_relation_type, {:one_of, variants}) do
+    fn struct, params ->
+      with {_, type} <- Enum.find(params, &to_string(elem(&1, 0)) == "type"),
+           {_, data} <- Enum.find(params, &to_string(elem(&1, 0)) == "data"),
+           {_type, mod} <- Enum.find(variants, fn {key, _val} -> to_string(key) == to_string(type) end) do
+        mod.changeset(struct, data)
+      else
+        _ ->
+          raise ArgumentError, """
+            Expected a type and a data field
+          """
+      end
+    end
+  end
+
   defp on_cast_default(type, module) do
     fn struct, params ->
       try do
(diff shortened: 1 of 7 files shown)


josevalim:
Thanks @bamorim for the PR.

In such cases I would go ahead with an Ecto Type (perhaps to be encapsulated as an Ecto Parameterized Type). That's because embeds come with several features, some that would be hard to implement or maintain for unions and others that do not make much sense at all (such as compile-time checking of embed fields in queries).

The Ecto codebase has two parts that are complex: relationship management and query planning and this would add complexity to both, especially because they are performance sensitive and adding complexity often makes things harder to optimize.

For cases like AuditLog, which is read-only, an Ecto Type is perfect, cause you read the map and convert to the relevant structs. And parameterized types could provide a possible route for generalization. :)

Malian:
@bamorim I may be wrong, but doesn't this package solve your issue? https://github.com/mathieuprog/polymorphic_embed

PR #4178, opened 2023-05-05
Support receiving index in relation cast

### LINKED ISSUE #4175: cast_assoc won't persist updated sort order as specified in the docs
### Elixir version

1.14.4

### Database and Version

PostgreSQL 15.2 with PostGIS 3.3.2

### Ecto Versions

3.10.0

postgrex

### Current behavior

1. call Changeset.cast_asoc with `sort_param` and `drop_param`.
2. call a custom function to update the sort, ending with a Changeset.put_change

```elixir

has_many(:addresses, Address,
  on_replace: :delete
)

changeset
|> cast_assoc(changeset, :addresses,
  sort_param: :addresses_sort,
  drop_param: :addresses_drop)
|> copy_children_positions()

defp copy_children_positions(changeset) do
  if children = Ecto.Changeset.get_change(changeset, :children) do
    children
    |> Enum.with_index(fn child, index ->
      Ecto.Changeset.put_change(child, :position, index)
    end)
    |> then(&Ecto.Changeset.put_change(changeset, :children, &1))
  else
    changeset
  end
end
```

### Expected behavior

The changeset should successfully change with the updated addresses. However it fails with:
```
** (RuntimeError) cannot replace related %Address{__meta__: #Ecto.Schema.Metadata<:loaded, "address">,
...
This typically happens when you are calling put_assoc/put_embed with the results of a previous 
put_assoc/put_embed/cast_assoc/cast_embed operation, which is not supported. You must call 
such operations only once per embed/assoc, in order for Ecto to track changes efficiently
    (ecto 3.10.0) lib/ecto/changeset/relation.ex:430: Ecto.Changeset.Relation.check_action!/2
    (ecto 3.10.0) lib/ecto/changeset/relation.ex:180: Ecto.Changeset.Relation.do_change/4
    (ecto 3.10.0) lib/ecto/changeset/relation.ex:347: Ecto.Changeset.Relation.map_changes/9
    (ecto 3.10.0) lib/ecto/changeset/relation.ex:156: Ecto.Changeset.Relation.change/3
    (ecto 3.10.0) lib/ecto/changeset.ex:1729: Ecto.Changeset.put_change/7
    (ecto 3.10.0) lib/ecto/changeset.ex:1720: Ecto.Changeset.put_change/3
```

v0idpwn:
Closes #4175.

This is a WIP, need to amend docs and add more tests. I'd like to assert that the approach is the correct one.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1081,9 +1081,11 @@ defmodule Ecto.Changeset do
       repository if there is a change, defaults to `true`
 
     * `:with` - the function to build the changeset from params. Defaults to the
-      `changeset/2` function of the associated module. It must be an anonymous
+      `changeset/2` function of the associated module. It can be an anonymous
       function that expects two arguments, a changeset to be modified and its
-      parameters
+      parameters. For associations of `:many` cardinality, functions with arity
+      3 are accepted, and the third argument will be the position of the associated
+      element in the list.
 
     * `:drop_param` - the parameter name which keeps a list of indexes to drop
       from the relation parameters
@@ -1126,7 +1128,9 @@ defmodule Ecto.Changeset do
     * `:with` - the function to build the changeset from params. Defaults to the
       `changeset/2` function of the associated module. It must be an anonymous
       function that expects two arguments, a changeset to be modified and its
-      parameters
+      parameters. For associations of `:many` cardinality, functions with arity
+      3 are accepted, and the third argument will be the position of the associated
+      element in the list.
 
     * `:drop_param` - the parameter name which keeps a list of indexes to drop
       from the relation parameters
@@ -1211,12 +1215,14 @@ defmodule Ecto.Changeset do
         {[], value}
       end
 
+    (
     sorted ++
       (pending
        |> Map.drop(drop)
        |> Enum.map(&key_as_int/1)
        |> Enum.sort()
        |> Enum.map(&elem(&1, 1)))
+    )
   end
 
   defp cast_params(%{cardinality: :one}, value, sort, drop) do
@@ -1265,7 +1271,7 @@ defmodule Ecto.Changeset do
 
                 1. implement the #{type}.changeset/2 function
                 2. pass the :with option to cast_#{type}/3 with an anonymous
-                   function.
+                   function
 
               When using an inline embed, the :with option must be given
               """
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset/relation.ex
```diff
@@ -295,7 +310,10 @@ defmodule Ecto.Changeset.Relation do
     {current_pks, current_map} = process_current(current, current_pks_fun, relation)
     %{unique: unique, ordered: ordered} = relation
     ordered = if ordered, do: current_pks, else: []
-    map_changes(value, new_pks_fun, fun, current_map, [], true, true, unique && %{}, ordered)
+
+    value
+    |> Enum.with_index()
+    |> map_changes(new_pks_fun, fun, current_map, [], true, true, unique && %{}, ordered)
```
v0idpwn, line 316: I'm computing `Enum.with_index` for all associations, even the ones that aren't using this feature. Probably this could be avoided, but would imply deeper refactor.

josevalim: Given `map_changes` is recursive, you could pass `idx` there and increment it yourself on each iteration?

lib/ecto/changeset.ex
```diff
@@ -1081,9 +1076,11 @@ defmodule Ecto.Changeset do
       repository if there is a change, defaults to `true`
 
     * `:with` - the function to build the changeset from params. Defaults to the
-      `changeset/2` function of the associated module. It must be an anonymous
+      `changeset/2` function of the associated module. It can be an anonymous
       function that expects two arguments, a changeset to be modified and its
-      parameters
+      parameters. For associations of `:many` cardinality, functions with arity
+      3 are accepted, and the third argument will be the position of the associated
+      element in the list.
```
josevalim, line 1083: Or nil if the changeset is being deleted, right?

josevalim (approved):
LGTM so far!

lib/ecto/changeset/relation.ex
v0idpwn: great call, i missed it! fixed :)

v0idpwn:
I think I'm done with this. Tell me if you want any changes.

lib/ecto/changeset.ex
```diff
@@ -1126,7 +1123,9 @@ defmodule Ecto.Changeset do
     * `:with` - the function to build the changeset from params. Defaults to the
       `changeset/2` function of the associated module. It must be an anonymous
       function that expects two arguments, a changeset to be modified and its
-      parameters
+      parameters. For associations with cardinality `:many`, functions with arity
+      3 are accepted, and the third argument will be the position of the associated
+      element in the list, or `nil`, if the changeset is being deleted.
```
josevalim, line 1128:
```suggestion
      element in the list, or `nil`, if the changeset is being replaced.
```

josevalim (approved):

PR #4181, opened 2023-05-05
Add warning about using schema based tenancy in PostgreSQL

BobbyMcWho:
Unraveling PostgreSQL schema-based tenancy is a nightmare. The amount of resources needed to scale this approach to hundreds of tenants is obscene. We should put a disclaimer here to warn folks about the potential dangers of using it. 

Articles linked: 
https://medium.com/infinite-monkeys/our-multi-tenancy-journey-with-postgres-schemas-and-apartment-6ecda151a21f
https://www.citusdata.com/blog/2016/12/18/schema-sharding-lessons/

### DIFF
--- a/guides/howtos/Multi tenancy with query prefixes.md
+++ b/guides/howtos/Multi tenancy with query prefixes.md
@@ -2,6 +2,9 @@
 
 With Ecto we can run queries in different prefixes using a single pool of database connections. For databases engines such as Postgres, Ecto's prefix [maps to Postgres' DDL schemas](https://www.postgresql.org/docs/current/static/ddl-schemas.html). For MySQL, each prefix is a different database on its own.
 
+> ⚠️ **Caution**  
+> When using PostgreSQL, schema-based tenancy may not be the best approach for your application if you plan to scale to tens to hundreds of tenants. For more information, see [this article](https://www.citusdata.com/blog/2016/12/18/schema-sharding-lessons/) from Citus Data, and [this article](https://medium.com/infinite-monkeys/our-multi-tenancy-journey-with-postgres-schemas-and-apartment-6ecda151a21f) from Influitive.
+
 Query prefixes may be useful in different scenarios. For example, multi tenant apps running on PostgreSQL would define multiple prefixes, usually one per client, under a single database. The idea is that prefixes will provide data isolation between the different users of the application, guaranteeing either globally or at the data level that queries and commands act on a specific tenants.
 
 Prefixes may also be useful on high-traffic applications where data is partitioned upfront. For example, a gaming platform may break game data into isolated partitions, each named after a different prefix. A partition for a given player is either chosen at random or calculated based on the player information.
--- a/guides/howtos/Multi tenancy with query prefixes.md
+++ b/guides/howtos/Multi tenancy with query prefixes.md
@@ -2,8 +2,9 @@
 
 With Ecto we can run queries in different prefixes using a single pool of database connections. For databases engines such as Postgres, Ecto's prefix [maps to Postgres' DDL schemas](https://www.postgresql.org/docs/current/static/ddl-schemas.html). For MySQL, each prefix is a different database on its own.
 
-> ⚠️ **Caution**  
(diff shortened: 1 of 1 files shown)


greg-rychlewski:
Don't take this as an explicit disapproval, but IMO I'm a bit skeptical of putting things like this in Ecto. 

I wouldn't really suggest someone to learn how to use their database through Ecto. I would suggest they learn how to use their database and then figure out how to transfer that knowledge to Ecto. 

We don't know if Postgres will ever change its implementation and make this comment obsolete (if you know it is basically 0% chance, please take this as a general statement: we don't know when database X will change its internal implementation of feature Y to be more performant). And it would be a lot of pressure for the maintainers to keep up with best practices of each database Ecto supports.

BobbyMcWho:
On the counterpoint, what prompted this PR is someone asking about how they should handle multi tenancy in the Elixir slack, and someone linked both the multi-tenancy docs from Ecto there as reference. 

I think it's really easy to bury yourself in the Postgres schema-based tenancy hole based on libraries like Triplex (or apartment in Ruby), and it's such a super painful place to get yourself out of once you realize the spot you're in. 

I spent a year recently working on a project that ultimately is going to take 18-24 months to remove schema-based tenancy, and once your app is successful and scaling to tens to hundreds of tenants then your development bandwidth and resource budgets are obliterated by a decision that was made at the start of the project

greg-rychlewski:
Maybe then a generic statement along the lines of this?

>The methods referenced in this guide are meant to illustrate the capabilities of Ecto and may not be appropriate for every situation.

We should see what Jose says, but I have the feeling statements like this and links to blogs are out of scope for the Ecto docs. It seems too subjective and subject to change. IIR we only link to official database docs for references to functionality.

josevalim:
There are good concerns all around and I think a compromise would be to indeed link to PostgreSQL official docs on this. Would Postgres docs officially add this disclaimer? If not, why not? Otherwise it becomes our responsibility to judge and intermediate the potential use cases of PostgreSQL and that feels misplaced given Ecto philosophy of mostly exposing the database as is.

josevalim:
Also, I have do have concerns about the linked articles. The one from Influitive says:

#### Migrations

>  Ideally, for things like column additions, you’d get constant O(1) performance. Under a separate schemas approach to tenancy, you now get O(N) performance where N=# tenants.

Sounds fair.

#### Database Resources

> We haven’t exactly dug into this, but I’m quite sure the sheer number of tables we have is an issue.

How can we make this "official Ecto advice" without digging into this and providing numbers?

#### Client Memory Bloat

Not an Ecto issue.

#### Record Identification

Good tip.

---

The CitusData is light on details. They mention connection pools concerns - which I doubt is an issue here. They also mention "Last, as you have more tenants, adding new tables, columns, or indexes can start to go from seconds to minutes to hours across thousands of different tenants." - there is no hard data provided but, as in the other article, it is quite logical. You do have to manage N tenants.

So in my opinion we can add a note similar to:

> When working with multi-tenant databases, beware you will need to apply database migrations (such as adding/removing tables, columns, and indexes) to each tenant. If your application intends to grow to dozens of thousands of tenants, those migrations can eventually become too expensive and take a long time to complete.

Because, from a glance, that's the only issue in common between them and is logical and I don't feel we need to dig deeper into numbers.

guides/howtos/Multi tenancy with query prefixes.md
```diff
@@ -2,6 +2,9 @@
 
 With Ecto we can run queries in different prefixes using a single pool of database connections. For databases engines such as Postgres, Ecto's prefix [maps to Postgres' DDL schemas](https://www.postgresql.org/docs/current/static/ddl-schemas.html). For MySQL, each prefix is a different database on its own.
 
+> ⚠️ **Caution**  
+> When using PostgreSQL, schema-based tenancy may not be the best approach for your application if you plan to scale to tens to hundreds of tenants. For more information, see [this article](https://www.citusdata.com/blog/2016/12/18/schema-sharding-lessons/) from Citus Data, and [this article](https://medium.com/infinite-monkeys/our-multi-tenancy-journey-with-postgres-schemas-and-apartment-6ecda151a21f) from Influitive.
```
josevalim, line 6:
```suggestion
> #### Multi-tenancy and migrations {: .warning}
>
> When working with multi-tenant databases, you need to apply database migrations (such as adding/removing tables, columns, and indexes) to each tenant. If your application grows to dozens of thousands of tenants or more, those migrations can eventually become too expensive and take a long time to complete.
```

josevalim:

josevalim:
Thank you, I pushed an initial version. Feel free to refine it with future PRs. :)

PR #4185, opened 2023-05-12
Improve documentation of `:required` option for `cast_embed` and `cast_assoc`

linusdm:
Nuance what 'required' actually means for one-style and many-style embeds/assocs.

This came from the confusion that one could think that it's enough to have an empty array for has/embeds_many to satisfy the required validation. But when enabling this option it actually has to be an array with at least one element.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1066,7 +1066,7 @@ defmodule Ecto.Changeset do
 
   ## Options
 
-    * `:required` - if the association is a required field. For associations with cardinality
+    * `:required` - if the association is a required field. For associations of cardinality
       one, a non-nil value satisfies this validation. For associations with many entries,
       a non-empty list is satisfactory.
 
@@ -1113,7 +1113,7 @@ defmodule Ecto.Changeset do
 
   ## Options
 
-    * `:required` - if the embed is a required field. For embeds with cardinality
+    * `:required` - if the embed is a required field. For embeds of cardinality
       one, a non-nil value satisfies this validation. For embeds with many entries,
       a non-empty list is satisfactory.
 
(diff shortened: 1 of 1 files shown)


v0idpwn (approved):

lib/ecto/changeset.ex
```diff
@@ -1111,7 +1113,9 @@ defmodule Ecto.Changeset do
 
   ## Options
 
-    * `:required` - if the embed is a required field
+    * `:required` - if the embed is a required field. For embeds with cardinality
+      one, a non-nil value satisfies this validation. For embeds with many entries,
+      a non-empty list is satisfactory.
```
v0idpwn, line 1118: I think `embeds_one` can sound better than "embeds of cardinality one", same for `embeds_many`.

lib/ecto/changeset.ex
```diff
@@ -1066,7 +1066,9 @@ defmodule Ecto.Changeset do
 
   ## Options
 
-    * `:required` - if the association is a required field
+    * `:required` - if the association is a required field. For associations with cardinality
+      one, a non-nil value satisfies this validation. For associations with many entries,
+      a non-empty list is satisfactory.
```
v0idpwn, line 1071: Same thing here

lib/ecto/changeset.ex
josevalim, line 1118:
```suggestion
    * `:required` - if the embed is a required field. For embeds of cardinality
      one, a non-nil value satisfies this validation. For embeds with many entries,
      a non-empty list is satisfactory.
```

lib/ecto/changeset.ex
josevalim, line 1071:
```suggestion
    * `:required` - if the association is a required field. For associations of cardinality
      one, a non-nil value satisfies this validation. For associations with many entries,
      a non-empty list is satisfactory.
```

josevalim:

PR #4187, opened 2023-05-15
fix: support parameterized map types in json path validation

zachdaniel:
Currently if you use a parameterized type that is backed by a `:map` type you will get errors when using bracket access. This adds support for parameterized types in that check.


josevalim:

PR #4191, opened 2023-05-21
Introduce t:Ecto.Changeset.empty_value/0

whatyouhide:
Also cleans up docs for `Ecto.Changeset.empty_values/0`.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -359,7 +359,7 @@ defmodule Ecto.Changeset do
   the module documentation for more information.
   """
   @typedoc since: "3.11.0"
-  @type empty_value :: (term() -> boolean()) | term()
+  @type empty_value :: (term() -> boolean()) | binary() | list() | map() | tuple()
 
   @number_validators %{
     less_than:                {&</2,  "must be less than %{number}"},
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -116,9 +116,10 @@ defmodule Ecto.Changeset do
   empty value inside the array will be removed.
 
   Empty values are stored as a list in the changeset's `:empty_values` field.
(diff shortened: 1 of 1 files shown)


lib/ecto/changeset.ex
```diff
@@ -351,6 +352,15 @@ defmodule Ecto.Changeset do
...
   @type types :: map()
 
+  @typedoc """
+  A possible value that you can pass to the `:empty_values` option.
+
+  See `empty_values/0` and the [*Empty values* section](#module-empty-values) in
+  the module documentation for more information.
+  """
+  @typedoc since: "3.11.0"
+  @type empty_value :: (term() -> boolean()) | term()
```
josevalim, line 362: `term()` already includes anonymous functions. 🤔

josevalim: Maybe we want to break term into strings and all other composite data types?

whatyouhide: @josevalim yes `term()` contains everything ofc, but I think this for documentation is a decent approach. Otherwise, we can also do `(term -> boolean) | String.t | term`, kind of the same. Or you meant something different?

josevalim: I was thinking of:

```suggestion
  @type empty_value :: (term() -> boolean()) | binary() | list() | map() | tuple()
```

Cause i can't think of anything else that would be reasonable as empty value?

whatyouhide: Why not atoms, or numbers (like `0`)?

josevalim: I can't see a reason where it would be sane to consider any of those as empty values.

whatyouhide: Right and actually, folks can use a function if they wanna do something weird, so let's go for that. 👍

josevalim:

PR #4193, opened 2023-05-26
add array_join

ruslandoga:
This PR probes for the core team's opinion on adding support for [`ARRAY JOIN`](https://clickhouse.com/docs/en/sql-reference/statements/select/array-join) as used in ClickHouse. If the core team approves of it, I'll add changes to the docs (mentioning that it's ClickHouse-specific) and tests in subsequent commits.

Here's how it could be used in the Ecto adapter for ClickHouse: https://github.com/plausible/ecto_ch/pull/76/files

```elixir
test "array join" do
  query =
    from at in "arrays_test",
      array_join: a in "arr",
      select: [at.s, a]

  assert all(query) == """
    SELECT a0."s",a1 FROM "arrays_test" AS a0 ARRAY JOIN "arr" AS a1\
    """
end
```

---

Right now `:lateral_join` is used in that Ecto adapter to express array joins:

- [connection.ex](https://github.com/plausible/ecto_ch/blob/0e424d64c3c754b4a35df57d80ddd5743416f564/lib/ecto/adapters/clickhouse/connection.ex#L341-L342)
- https://github.com/plausible/ecto_ch#caveats
- [Plausible examples](https://github.com/search?q=repo%3Aplausible%2Fanalytics%20lateral&type=code)

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -795,7 +795,7 @@ defmodule Ecto.Query do
   defp wrap_in_subquery(queryable), do: %Ecto.SubQuery{query: Ecto.Queryable.to_query(queryable)}
 
   @joins [:join, :inner_join, :cross_join, :cross_lateral_join, :left_join, :right_join, :full_join,
-          :inner_lateral_join, :left_lateral_join]
+          :inner_lateral_join, :left_lateral_join, :array_join, :left_array_join]
 
   @doc """
   Puts the given prefix in a query.
@@ -1053,6 +1053,8 @@ defmodule Ecto.Query do
   defp join_qual(:cross_lateral_join), do: :cross_lateral
   defp join_qual(:left_lateral_join), do: :left_lateral
   defp join_qual(:inner_lateral_join), do: :inner_lateral
+  defp join_qual(:array_join), do: :array
+  defp join_qual(:left_array_join), do: :left_array
 
   defp collect_with_ties([{:with_ties, with_ties} | t], nil),
     do: collect_with_ties(t, with_ties)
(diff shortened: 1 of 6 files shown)


v0idpwn:
For me it seems okay, mostly because I don't see good enough alternative paths for the adapter to support other qualifiers.

josevalim:
The only other way I see to extend this is to stop validating the first argument of `join`, so you can pass anything, but I am not convinced that's better, so this sounds good to me too!

Sounds good to me too, as I don't see another way to extend this.

ruslandoga:
I've added tests and some minimal docs :)

greg-rychlewski:
From what I remember, the array joins are a bit different because they join on a column that exists in the base table instead of a different table. Is that right? 

If so maybe adding some notes to the docs to explain it doesn't work the same as regular joins would help users/maintainers avoid weird issues?  i.e. you can't join on a schema, the source name is actually referring to a column in the base table not a new table, you can't use it with associations, etc... It's also not obvious to me what would happen if it was included in a query with multiple joins and there is ambiguity in which table the column belongs to.

Maybe the link to the Clickhouse docs is enough. Just wanted to put it out there to see what the other maintainers think is helpful.

ruslandoga:
> From what I remember, the array joins are a bit different because they join on a column that exists in the base table instead of a different table. Is that right?

Yes, that's right. I wonder if it's possible to join on array in another table... Will try.

> If so maybe adding some notes to the docs to explain it doesn't work the same as regular joins would help users/maintainers avoid weird issues? i.e. you can't join on a schema, the source name is actually referring to a column in the base table not a new table, you can't use it with associations, etc...

Will add!

>  It's also not obvious to me what would happen if it was included in a query with multiple joins and there is ambiguity in which table the column belongs to.

I think this would be fine since the sources would be aliased for each join / column:

```elixir
  test "array join" do
     query =
       from at in "arrays_test",
         array_join: a in "arr",
         select: [at.s, a]

     assert all(query) == """
            SELECT a0."s",a1 FROM "arrays_test" AS a0 ARRAY JOIN "arr" AS a1\
            """
  end
```

from https://github.com/plausible/ecto_ch/pull/76

greg-rychlewski:
> I think this would be fine since the sources would be aliased for each join / column:
What would happen in a situation like this?

```elixir
Post
|> join(:inner, [_], c in Comment, on: true)
|> join(:array, [_, _], a in "arr", on: true)
```
If "arr" is a column in Post and Comment. What does `a in "arr"` refer to.

ruslandoga:
Ah, I misunderstood. Let me try it with ClickHouse.

greg-rychlewski:
I like the API a lot more, but it would involve some work, i.e. only let it as a source in joins, disallow it for all joins other than array, change the planner, etc...

Not sure if everyone will want to make those changes.  I have to see what Jose/v0idpwn think.

ruslandoga:
```elixir
Post
|> join(:inner, [_], c in Comment, on: true)
|> join(:array, [_, c], a in c.arr, on: true)
```
is invalid according to ClickHouse. It only accepts array join on `FROM` table.

josevalim (approved):

ruslandoga:
@josevalim just in case, I still need to update docs with @greg-rychlewski's suggestions.

ruslandoga:
I've updated the docs but I didn't include all limitations of array joins and instead provided an example and mentioned that only the base table can be used.

greg-rychlewski (approved):

greg-rychlewski:
Thank you!

ruslandoga:
Thank you for accepting it :)

PR #4196, opened 2023-05-30
Add use info for Ecto.{Schema, Type, ParameterizedType}

v0idpwn:
Not 100% sure about both content and location where these are in the module docs.

### DIFF
--- a/lib/ecto/parameterized_type.ex
+++ b/lib/ecto/parameterized_type.ex
@@ -66,6 +66,11 @@ defmodule Ecto.ParameterizedType do
 
   To use this type in places where you need it to be initialized (for example,
   schemaless changesets), you can use `init/2`.
+
+  > #### `use Ecto.ParameterizedType` {: .info}
+  > When you `use Ecto.ParameterizedType`, it will set
+  > `@behaviour Ecto.ParameterizedType` and define default, overridable
+  > implementations for `c:embed_as/2` and `c:equal?/3`.
   """
 
   @typedoc """
--- a/lib/ecto/parameterized_type.ex
+++ b/lib/ecto/parameterized_type.ex
@@ -68,6 +68,7 @@ defmodule Ecto.ParameterizedType do
   schemaless changesets), you can use `init/2`.
 
   > #### `use Ecto.ParameterizedType` {: .info}
+  >
   > When you `use Ecto.ParameterizedType`, it will set
   > `@behaviour Ecto.ParameterizedType` and define default, overridable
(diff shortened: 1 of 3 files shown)


v0idpwn:
I didn't add to `Ecto.Repo` because it has quite extensive info about its `__using__` indirectly in the docs. `Ecto.Schema` is a bit of the same, though not as much, so I added.

lib/ecto/parameterized_type.ex
```diff
@@ -66,6 +66,11 @@ defmodule Ecto.ParameterizedType do
 
   To use this type in places where you need it to be initialized (for example,
   schemaless changesets), you can use `init/2`.
+
+  > #### `use Ecto.ParameterizedType` {: .info}
+  > When you `use Ecto.ParameterizedType`, it will set
+  > `@behaviour Ecto.ParameterizedType` and define default, overridable
+  > implementations for `c:embed_as/2` and `c:equal?/3`.
```
josevalim, line 73:
```suggestion
  > #### `use Ecto.ParameterizedType` {: .info}
  >
  > When you `use Ecto.ParameterizedType`, it will set
  > `@behaviour Ecto.ParameterizedType` and define default, overridable
  > implementations for `c:embed_as/2` and `c:equal?/3`.
```

lib/ecto/schema.ex
```diff
@@ -107,6 +107,14 @@ defmodule Ecto.Schema do
   database with the help of `Ecto.Repo`. On the other hand, embedded
   schemas cannot be queried directly (they are not queryable).
 
+  > #### `use Ecto.Schema` {: .info}
+  > When you `use Ecto.Schema`, it will:
+  > - import `Ecto.Schema` macros `schema/2` and `embedded_schema/1`
+  > - register default values for module attributes that can be overriden, such as
+  > `@primary_key` and `@timestamp_opts`
+  > - register other module attributes that are used by `Ecto.Schema` to provide
+  > reflection through the `__schema__/1` and `__changeset__/1` functions
```
josevalim, line 116:
```suggestion
  > #### `use Ecto.Schema` {: .info}
  >
  > When you `use Ecto.Schema`, it will:
  >
  > - import `Ecto.Schema` macros `schema/2` and `embedded_schema/1`
  > - register default values for module attributes that can be overridden, such as
  > `@primary_key` and `@timestamp_opts`
  > - define reflection functions such as `__schema__/1` and `__changeset__/1`
  >
  > We detail those throughout the module documentation.
```

josevalim (approved):
LGTM! We can do one for Ecto.Repo later :)

Issue #3406, opened 2020-09-11
Casting recursive cte query with prefix not working


D4no0:
Suppose I have the schema:
```
@schema_prefix "company"
schema "organizations" do
  belongs_to :parent, Organization
end
```
I want to get all the children with the query:
```
initial_query = Organization |> where([org], is_nil(org.parent_id))
recursion_query = Organization |> join(:inner, [org], new_org in "cte", on: org.parent_id == new_org.id)
 final_query = initial_query |> union_all(^recursion_query)

{"cte", Organization}
|> recursive_ctes(true)
|> with_cte("cte", as: ^final_query)
|> Repo.all()
```
Doing this query gives the error:
```
** (Postgrex.Error) ERROR 42P01 (undefined_table) relation "company.cte" does not exist
query: WITH RECURSIVE "cte" AS (SELECT o0."id" AS "id", o0."parent_id" AS "parent_id" FROM "company"."organizations" AS o0 WHERE (o0."id" = $1) UNION ALL (SELECT o0."id",o0."parent_id" FROM "company"."organizations" AS o0 INNER JOIN "cte" AS c1 ON o0."parent_id" = c1."id")) SELECT c0."id", c0."parent_id"  FROM "company"."cte" AS c0
```

Judging from the query it seems that it tries to append `company` prefix to `cte`, which makes no sense.

Are there any workarounds for this behavior?

josevalim:
You can set the prefix for each segment. So something like this should work: `from(cte in {"cte", Organization}, prefix: nil) |> ...`

D4no0:
It doesn't seem to work, the same query is generated.

josevalim:
Fixed in master!

D4no0:
Thank you very much!

PR #4203, opened 2023-06-09
docs: Provide CTE prefix replacement

mpotra:
When Ecto schemas have a prefix set, using CTEs references the Schema prefix which breaks the SQL query.
There already exists [an issue closed on this](https://github.com/elixir-ecto/ecto/issues/3406), but [the solution](https://github.com/elixir-ecto/ecto/issues/3406#issuecomment-690953646) doesn't appear to have been documented.

Personally, I'd prefer having a better option, such as maybe `put_query_prefix/2` to handle tuples and set the _From query prefix_ itself, instead of delegating to `Queryable.to_query/1` which sets the schema prefix in the From query and then `put_query_prefix/2` would be setting the prefix on the Query.

Because it took me a couple of hours to figure out the issue and find the solution, I propose that this gets at least mentioned in the documentation as well, to save others using prefixed Ecto schemas some time.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1359,6 +1359,16 @@ defmodule Ecto.Query do
       |> join(:left, [c], p in assoc(c, :products))
       |> group_by([c], c.id)
       |> select([c, p], %{c | products_count: count(p.id)})
+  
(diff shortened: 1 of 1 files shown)


josevalim:

PR #4207, opened 2023-06-13
Add example to Changeset.validate_change/3

jcrevits:
Hello! There's an almost-undocumented other type of return value in this function's typespec - `[{atom, {String.t, Keyword.t}}]` that I nearly made use of. I think it would be nice to add an example? I'm not so sure about the function documentation itself.

It's similar to how `add_error/4` can accept additional keywords to add contextual information. https://github.com/elixir-ecto/ecto/blob/23edc706c49dbfa95ca02eb6f2a559f212d948c2/lib/ecto/changeset.ex#L2202-L2205


josevalim:

PR #4211, opened 2023-06-19
Improve docs and error for Ecto.Repo.delete_all/2

whatyouhide:


lib/ecto/query/planner.ex
```diff
@@ -1995,6 +1995,10 @@ defmodule Ecto.Query.Planner do
                   havings: [], preloads: [], assocs: [], distinct: nil, lock: nil,
                   windows: [], combinations: []} ->
         query
+      _ when operation == :delete_all ->
+        error! query, "`#{operation}` allows only `with_cte`, `where`, `select`, and `join` expressions. " <>
+                      "You can exclude unwanted expressions from a query by using " <>
+                      "Ecto.Query.exclude/2. Error found"
```
whatyouhide, line 2001: Quick and dirty, but at least this mentions `select` too, which was not mentioned before and it's allowed 🙃

PR #4212, opened 2023-06-20
Fix calls to deprecated Logger.warn/2

LouisMT:
I'm not entirely sure if this is the right way. Maybe it's also possible to bump the minimum Elixir version to 1.11, but I guess that would be a breaking change?


josevalim:
Please bump the Elixir requirement and update the CI files too? We won't merge it immediately but we can merge once we start preparing for v3.11. :)

LouisMT:
Alright, done!

josevalim:
master/main is now v3.11 :)

PR #4214, opened 2023-06-21
Pass `nil` through adapter dumpers and loaders

whatyouhide:
This is necessary for our work in [Exandra](https://github.com/vinniefranco/exandra), a Cassandra/Scylla Ecto adapter. Xandra (the underlying driver we use) needs to pass query params to Cassandra as `{type, value}`, so we need every single term to go through our adapter's dumpers and loaders.

This change causes some issues with the MSSQL adapter, but I think they can be fixed, and I'll be working on that next.

cc @josevalim


whatyouhide:
We need to rerun this job with https://github.com/elixir-ecto/ecto_sql/pull/528 merged, yep!

josevalim:

PR #4216, opened 2023-06-21
poc: support virtual embeds

v0idpwn:
The question is if we should. WDYT?
Wrote this while answering a question on slack.

### DIFF
--- a/lib/ecto/embedded.ex
+++ b/lib/ecto/embedded.ex
@@ -28,6 +28,7 @@ defmodule Ecto.Embedded do
     :owner,
     :related,
     :on_cast,
+    virtual: false,
     on_replace: :raise,
     unique: true,
     ordered: true
@@ -144,7 +145,11 @@ defmodule Ecto.Embedded do
   @doc false
   def prepare(changeset, embeds, adapter, repo_action) do
     %{changes: changes, types: types, repo: repo} = changeset
-    prepare(Map.take(changes, embeds), types, adapter, repo, repo_action)
+    #non_virtual_embeds = Enum.reject(embeds, & &1.virtual)
+
+    changes
+    |> Map.take(embeds)
+    |> prepare(types, adapter, repo, repo_action)
   end
 
   defp prepare(embeds, _types, _adapter, _repo, _repo_action) when embeds == %{} do
@@ -153,8 +158,14 @@ defmodule Ecto.Embedded do
 
   defp prepare(embeds, types, adapter, repo, repo_action) do
     Enum.reduce embeds, embeds, fn {name, changeset_or_changesets}, acc ->
-      {:embed, embed} = Map.get(types, name)
-      Map.put(acc, name, prepare_each(embed, changeset_or_changesets, adapter, repo, repo_action))
+      case Map.get(types, name) do
+        {:embed, %{virtual: false} = embed} ->
+          prepared = prepare_each(embed, changeset_or_changesets, adapter, repo, repo_action)
+          Map.put(acc, name, prepared)
+
+        {:embed, %{virtual: true}} ->
+          acc
+      end
     end
   end
 
--- a/test/ecto/query/planner_test.exs
+++ b/test/ecto/query/planner_test.exs
@@ -499,7 +499,7 @@ defmodule Ecto.Query.PlannerTest do
 
   test "plan: generates a cache key" do
     {_query, _cast_params, _dump_params, key} = plan(from(Post, []))
-    assert key == [:all, {:from, {"posts", Post, 111065921, "my_prefix"}, []}]
+    assert key == [:all, {:from, {"posts", Post, 23210922, "my_prefix"}, []}]
 
     query =
       from(
@@ -525,7 +525,7 @@ defmodule Ecto.Query.PlannerTest do
                    {:limit, {true, 1}},
                    {:where, [{:and, {:is_nil, [], [nil]}}, {:or, {:is_nil, [], [nil]}}]},
                    {:join, [{:inner, {"comments", Comment, 38292156, "world"}, true, ["join hint"]}]},
-                   {:from, {"posts", Post, 111065921, "hello"}, ["hint"]},
+                   {:from, {"posts", Post, 23210922, "hello"}, ["hint"]},
                    {:select, 1}]
   end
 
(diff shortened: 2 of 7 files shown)


josevalim:
How would a virtual embed be loaded/dumped? What is the use case? :)

josevalim:
Right. but if we are going to include it in the docs, we need to explain when this would useful. Why do you have a virtual field in this style? What is the use case? When would we recommend it?

SukhikhN:
Virtual embeds could be useful when an Ecto schema is used primarily as application's internal data structure and in the second place for storing it somewhere.

For example, I have a project where I use a schema with several embedded schemas for representing a game state that lives in a GenServer. Ecto changesets are very useful for mutating this state in response to player's actions. 

Only part of the state should be stored (in ETS or database) to be restored when the GenServer (re)starts. So I made some fields virtual in order to skip them when dumping the state to a storage. But I'm forced to declare embeds as virtual maps or map arrays and write special functions for casting them and tracking their changes.

Another example: I cast an external API JSON data into a schema in order to make convenient transformations into internal data format and reject invalid data. Nested JSON objects are casted into embeds. This API uses almost same data structure for requests and responses, so when making a request to the API, I reuse the schema and dump it to JSON. The request structure has less fields than response, so again some fields of the schema are virtual and therefore are not dumped when I make a request. I can't use this approach if one of response embeds should not be send when using the schema for a request.

SukhikhN:
Of course, this problems could be solved in other ways, but making embeds virtual seems more natural to me in this cases.

After all, if we can mark any embedded schema field as virtual so it would not be dumped/loaded, why should embeds be an exception?

josevalim:
Thanks for the clarification. I would probably model your case with different schemas but it doesn't hurt supporting this for consistency, as long as the implementation is straight-forward.

@v0idpwn do you see this growing much in complexity?

v0idpwn:
I don't think it will grow in complexity at all, will add some tests.

pmargreff:
We have a project that would benefit from using `embeds_many ... virtual: true`. Our use case is the following:

We have a complex embedded field that we want to display in the UI more using forms directive. Let's say we have an embed for IP, called `ip` we achieve to render it to the users or APIs by marking the field as `ip_formatted, :string, virtual: true`,  parsing in from and to this virtual fie/d in between fetching from database and rendering it. So, we have both a detailed embed and a formatted version for easy display.

The challenge arises with more complex scenarios, like converting a list of embeds into a simpler list of embeds that are compatible with form validations and other features, such as sorting (we are not using liveview, and unfortunately, it isn't an option in this case). Our current workaround has been to use `load_in_query: false`, which functions similarly to `virtual: true` and is also supported in `embeds_many`.

It works, but it is not ideal (and I imagine not intended to be used in that way either). It always goes with this comment `# load_in_query is to emulate virtual which is not compatible with embeds_many`. That said, we would benefit from this work as well. 

cc: @Lgdev07

josevalim:
Sorry, but that is still unclear then, if you never write it down, what is the benefit of being an embedded? It could as well be `field :children, :any, virtual: true`. If you are not using changesets or queries, which is often the case with virtual fields, why is embedding giving you? It may be even worth asking if it should be coupled in the same schema/struct in the first place. :)

pmargreff:
> If you are not using changesets or queries...

We are using the changesets for validations and rendering errors against the "virtual embed". You have a point about having a different schema that has not persisted. But in that case, we would need to duplicate many other fields where we don't do this type of normalization. Using `any` would also be an option, but it would make it more complex to reason about the structure when passing it through changesets for validation.

josevalim:
@v0idpwn shall we ship this one?

v0idpwn:
@josevalim  I think probably yes! Will give it a rebase and another look :)

lib/ecto/embedded.ex
```diff
@@ -149,7 +150,11 @@ defmodule Ecto.Embedded do
   @doc false
   def prepare(changeset, embeds, adapter, repo_action) do
     %{changes: changes, types: types, repo: repo} = changeset
-    prepare(Map.take(changes, embeds), types, adapter, repo, repo_action)
+    #non_virtual_embeds = Enum.reject(embeds, & &1.virtual)
+
```
v0idpwn, line 154:
```suggestion
```

test/ecto/query/planner_test.exs
```diff
@@ -633,9 +634,18 @@ defmodule Ecto.Query.PlannerTest do
...
            ]
+
+   assert key == [:all,
+       {:lock, "foo"},
+       {:prefix, "foo"},
+       {:limit, {true, 1}},
+       {:where, [{:and, {:is_nil, [], [nil]}}, {:or, {:is_nil, [], [nil]}}]},
+       {:join, [{:inner, {"comments", Comment, 38292156, "world"}, true, ["join hint"]}]},
+       {:from, {"posts", Post, 132715331, "hello"}, ["hint"]},
+       {:select, 1}]
```
greg-rychlewski, line 648:
```suggestion
```
it looks to me like this is an accidental duplicate but let me know if i'm mistaken

greg-rychlewski:
looks good to me. did you want to update the docs with Jose's suggestion here: https://github.com/elixir-ecto/ecto/pull/4216#issuecomment-1600568702

greg-rychlewski:
One thing I just realized is there are edge cases for nested embeds that don't work nicely. For example:

```elixir
defmodule Ecto.Integration.Order do
  use Ecto.Schema

  schema "orders" do
    embeds_one :item, Ecto.Integration.Item
  end
end

defmodule Ecto.Integration.Item do
  use Ecto.Schema

  embedded_schema do
    embeds_one :last_primary_color, Ecto.Integration.ItemColor, virtual: true
  end
end

defmodule Ecto.Integration.ItemColor do
  use Ecto.Schema

  embedded_schema do
    field :name, :string
  end
end
```

If `:last_primary_color` used to be persisted and is changed to virtual, the data will be pulled in from the database because queries are looking at embeds as a whole rather than key by key. So I think it it can lead to these two scenarios:

1. The field being populated when it should be `nil` when using `Repo.all`, etc...
2. The user supplied value being overwritten by what's in the db during `Repo.insert`, etc. This one I haven't checked closely though.

PR #4217, opened 2023-06-24
Ecto.Changeset.validate_number/3 shorthand options

BrandtHill:
Add bash test integer comparison semantics as shorthand options for validate_number/3.
These are the two character shorthands for `:lt`, `:gt`, `:le`, `:ge`, `:eq`, and `:ne`.


josevalim:
Hi @BrandtHill, we don't plan to add the shortcuts so far. Being more explicit doesn't hurt. :)

BrandtHill:
I would respectfully disagree as I nearly fall asleep on my keyboard each time I have to type of `:greater_than_or_equal_to`, but I understand.

PR #4219, opened 2023-06-27
Allow `many_to_many` to order preloads by join table

greg-rychlewski:
This was inspired by this comment: https://github.com/elixir-ecto/ecto/issues/3509#issuecomment-1606909597.

It seems useful to be able to order by the join table, especially if you are adding metadata to the relationship.

The proposal here was the best one I could think of. It feels a little awkward having two options but the other ideas I thought of didn't seem too great:

1. Mixing the source table option into `preload_order` seems a bit messy.
2. I don't think we want to allow bindings or anything like that in the schema definition. It seems complicated and messy.


josevalim:
What about adding `join_preload_order`, to mirror `join_defaults` and `join_where`?

greg-rychlewski:
That sounds like a good option to me. Do you think we should raise if both `join_preload_order` and `preload_order` are specified? I think otherwise the order to apply them would be ambiguous.

Maybe this is overkill, but given ^ do you think we should considering giving users the option to order across the tables? Something like: `order by join_table.col_a, assoc_table.col_b, ...`. The API is maybe not so clean though I think you'd have to do a nested keyword list: `[{:join, asc: :col_a}, {:assoc, asc: :col_b}]`

josevalim:
How hard would it be to check the columns and, if missing, we assume it is from join?

greg-rychlewski:
It would be contained to a relatively small place. Just this function: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/preloader.ex#L318

```elixir
defp preload_order(assoc, query, related_field) do
  custom_order_by = Enum.map(assoc.preload_order, fn
    {direction, field} ->
      {direction, related_key_to_field(query, {0, field})}
    field ->
      {:asc, related_key_to_field(query, {0, field})}
  end)

  [{:asc, related_field} | custom_order_by]
end
```

You have the assoc so you can check if it is many to many. And you have the query so you can get the from source. Then I think you'd just need to use the reflection functions to see if the field exist. Like checking if the type is `nil` or something like that.

The one issue I can see though is you wouldn't be able to distinguish if both tables had the same column name.

josevalim:
Yeah, i am not sure what is the best API. :(

josevalim:
I would avoid adding them at compile-time. Maybe we could allow preload_order as a MFA and in there you should have access to both?

greg-rychlewski:
Sorry if I'm misunderstanding, were you mentioning compile-time due to generating dependencies between modules? 

If so, what I meant was to have "dummy" bindings.  i.e. one fixed binding name for the join and one for the assoc. And then when the option is compiled it will be something like `[{:join, dir1, field}, {:assoc, dir2, field2}, ...]`

I was trying to think of how to use a function too. Though I think we'd have to give them access to modify the entire query rather than generate just the new `order_by` clause. Unless I am not seeing something.

josevalim:
Sorry, what I meant is that extracting a macro expression from an option will be tricky and confusing, especially because none of the other options have macro-like behaviour.

> Though I think we'd have to give them access to modify the entire query rather than generate just the new order_by clause. Unless I am not seeing something.

Gah, good point. I guess the best API so far is the one you proposed as `preload_order: [through: [...], assoc: [...]]`. But I would wait a bit more... maybe inspiration will strike later?

greg-rychlewski:
Yeah let's wait :).

greg-rychlewski:
I was trying to get something like this to work:

```elixir
   preload_order: fn -> dynamic([join, assoc], assoc.field) end
```

But couldn't get it to work. I always get compiler errors either about dynamic not existing or the binding not existing. 

Sorry to bug you but I'm hitting my limits a bit =). Do you know if this is a definite no-no or if there should be something that can make it work?

josevalim:
I am almost sure it has to be written using MFAs (which is why we use MFAs everywhere).

greg-rychlewski:
I think the MFA works pretty nicely. WDYT of this change?

lib/ecto/schema.ex
```diff
@@ -1369,11 +1369,15 @@ defmodule Ecto.Schema do
...
-      See `Ecto.Query.order_by/3` for more examples.
+    * `:preload_order` - Sets the default `order_by` when preloading the association.
+      It may either be a keyword list/list of fields (see `Ecto.Query.order_by/3`) or
+      a MFA tuple, such as `{Mod, fun, []}`. If specifying a keyword list/list of fields,
+      the ordering applies to the association's table and not the join table. Ordering by
+      the join table can be achieved by specifying a MFA tuple that utilizes
+      `Ecto.Query.dynamic/2`. For example, you may specify `{Mod, fun, []}` where `Mod.fun()`
+      returns `[desc: dynamic([assoc, join], join.updated_at)]`. Please note the ordering of
+      the bindings: the association comes before the join table. If you specify a custom preload
+      query, the bindings for that query come before the join table.
```
josevalim, line 1380: This is probably too large to keep as a list item. What if we briefly describe the supported inputs and mention a `## Preload order` section, which can go into all of these cases, including examples?

lib/ecto/repo/preloader.ex
```diff
@@ -316,16 +316,75 @@ defmodule Ecto.Repo.Preloader do
...
+
+            #{Inspect.Ecto.Query.to_string(query)}
+            """
+    end
+
+    {expr, _, _, _, _, _} = Ecto.Query.Builder.Dynamic.fully_expand(%Ecto.Query{}, dynamic)
+
+    case expr do
+      {{:., _, [{:&, _, [_ix]}, _field]}, [], []} = field_ast ->
+        field_ast
```
josevalim, line 379: Do we need to do this validation? Could we simply return the `dynamic` field instead?

greg-rychlewski: I think we have to but this was one of the more difficult parts for me. I could let you know my rationale:

The existing `preload_order` behaviour is to [insert the order before any of the `order_by`s that are already in the preload query](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/preloader.ex#L264):

```elixir
  update_in query.order_bys, fn order_bys ->
    [%Ecto.Query.QueryExpr{expr: preload_order(assoc, query, field), params: [],
                           file: __ENV__.file, line: __ENV__.line}|order_bys]
```

So because of that it basically creates the AST that would normally be created by the builder and injects it. If don't expand the dynamic then it stays as `%DynamicExpr{}`.

greg-rychlewski: Something similar happens for the keyword list version of this as well only it's a lot simpler to achieve it https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/preloader.ex#L329

```elixir
 defp preload_order(assoc, query, related_field) do
  custom_order_by = Enum.map(assoc.preload_order, fn
    {direction, field} ->
      {direction, related_key_to_field(query, {0, field})}
    field ->
      {:asc, related_key_to_field(query, {0, field})}
  end)

  [{:asc, related_field} | custom_order_by]
end

defp related_key_to_field(query, {pos, key}) do
  {{:., [], [{:&, [], [related_key_pos(query, pos)]}, key]}, [], []}
end
```

josevalim: I am just worried this will fail in unpredictable ways. Maybe we can add a fourth argument  to `order_by` which would allow us to prepend? We can also use reverse_order + append + reverse_order. But probably too much work for not much. Thoughts?

greg-rychlewski: Oh I actually like both those ideas. The 4th argument is a bit better IMO because it gives us and other users extra functionality  and is a bit clearer when reading the code. I could implement that if you agree?

josevalim: Please go ahead as you prefer!

greg-rychlewski:
Thanks for the discussion/feedback. I finished the changes we talked about . Some callouts:

1. I added the new `order_by` option to this PR because it was a small change. I could separate it out if you prefer.
2. I didn't add the new `order_by` option to the keyword queries because it's a bit more work and distracts a bit from the other changes. But I could follow up quickly with a PR for that.
3. I changed it so that any `:preload_order` is now using the new `order_by` option even if it's a compile-time keyword list/list.
4. There is a default ordering Ecto applies that goes right at the beginning of the query. This is so the preloaded structs can be placed into their parents in the right order. This still has to use the old way because the ordering is coming from the field definition in the association struct. The field definition has the binding position in it and there is also some custom logic for negative indices. So I thought I better not touch that.

josevalim:
> I didn't add the new order_by option to the keyword queries because it's a bit more work and distracts a bit from the other changes. But I could follow up quickly with a PR for that.

I agree it is not necessary. It is fine to depend on the keyword one for these cases.

lib/ecto/query.ex
```diff
@@ -1752,12 +1752,20 @@ defmodule Ecto.Query do
...
+  can be appended or preprended to the existing ones. The behaviour is controlled
+  through the `:position` option. By default, new expressions are appended.
 
   `order_by` also accepts a list of atoms where each atom refers to a field in
   source or a keyword list where the direction is given as key and the field
   to order as value.
 
+  ## Options
+
+    * `:position` - where to place the order expression relative to the
```
josevalim, line 1764: I think `:position` is a bit too ambiguous, given we already have positional bindings and so on. Maybe "placement"? "mode"?

lib/ecto/query.ex
```diff
@@ -1806,8 +1814,8 @@ defmodule Ecto.Query do
       City |> order_by(^order_by_param) # Keyword list
 
   """
-  defmacro order_by(query, binding \\ [], expr) do
-    Builder.OrderBy.build(query, binding, expr, __CALLER__)
+  defmacro order_by(query, binding \\ [], expr, opts \\ []) do
+    Builder.OrderBy.build(query, binding, expr, opts[:position], __CALLER__)
```
josevalim, line 1818: We cannot access it here because it means we can't do this:

    opts = [position: :prepend]
    order_by(query, [p], p.value, opts)

I think we should either allow dynamic options (and reading it only at runtime) or require an atom to be given (which can be compile-time only).

josevalim (commented):
I have added some minor nits, overall it looks great!

lib/ecto/schema.ex
```diff
@@ -1369,11 +1369,11 @@ defmodule Ecto.Schema do
...
 
-    * `:preload_order` - Sets the default `order_by` of the association.
-      It is used when the association is preloaded.
-      For example, if you set `Post.many_to_many :tags, Tag, join_through: "posts_tags", preload_order: [asc: :foo]`,
-      whenever the `:tags` associations is preloaded, the tags will be order by the `:foo` field.
-      See `Ecto.Query.order_by/3` for more examples.
+    * `:preload_order` - Sets the default `order_by` when preloading the association.
+      It may be a keyword list/list of fields or an MFA tuple, such as `{Mod, fun, []}`.
+      Both cases must resolve to a valid `order_by` expression. See `Ecto.Query.order_by/4`
+      to learn more about about ordering expressions. See the preload order section below to
```
greg-rychlewski, line 1375: @josevalim Is there a way with ex doc to link to sections within the same function doc? Or if not is this something that is theoretically possible and I could investigate adding?

josevalim: It is possible to link to h2 (maybe h3). Run `MIX_ENV=docs mix docs` and see the link. :)

greg-rychlewski: ah beautiful, thank you!

lib/ecto/query.ex
greg-rychlewski: `mode` sounds good to me. done!

lib/ecto/query.ex
greg-rychlewski: Thanks, I updated so it's evaluated at runtime. Does it seem alright?

lib/ecto/query/builder/order_by.ex
```diff
@@ -176,34 +189,48 @@ defmodule Ecto.Query.Builder.OrderBy do
   If possible, it does all calculations at compile time to avoid
   runtime work.
   """
-  @spec build(Macro.t, [Macro.t], Macro.t, Macro.Env.t) :: Macro.t
-  def build(query, _binding, {:^, _, [var]}, env) do
+  @spec build(Macro.t, [Macro.t], Macro.t, Macro.t, Macro.Env.t) :: Macro.t
+  def build(query, _binding, {:^, _, [var]}, opts, env) do
     quote do
-      Ecto.Query.Builder.OrderBy.order_by!(unquote(query), unquote(var), unquote(env.file), unquote(env.line))
+      mode = Ecto.Query.Builder.OrderBy.mode!(unquote(opts)[:mode])
```
josevalim, line 195: Let’s reduce the amount of generated code and pass `opts` to order_by!.

lib/ecto/query/builder/order_by.ex
```diff
@@ -128,6 +131,16 @@ defmodule Ecto.Query.Builder.OrderBy do
 
   defp to_field(field), do: {{:., [], [{:&, [], [0]}, field]}, [], []}
 
+  @doc """
+  Called at runtime to verify the ordering mode
+  """
+  def mode!(nil), do: @default_mode
```
josevalim, line 137: Have this function receive opts as argument to we extract at runtime rather than compile time.

josevalim (approved):
Just one minor nit to do less work inside the quote and ship it!

PR #4220, opened 2023-06-28
Allow updates in CTE

mpotra:
As suggested by https://github.com/elixir-ecto/ecto_sql/pull/532#issuecomment-1611644324 the planner does not allow updates in CTEs, which fail the test [for the PR that introduces](https://github.com/elixir-ecto/ecto_sql/pull/532) data-modifying CTEs for PostgreSQL.

This PR adds `update` to the allowed list for `:all` planner, to enable it in CTEs.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1473,10 +1473,6 @@ defmodule Ecto.Query do
     if !with_query do
       Builder.error!("`as` option must be specified")
     end
-    
-    unless operation in [nil, :all, :update_all, :delete_all, :insert_all] do
-      Builder.error!("`operation` option must be one of :all, :update_all, :delete_all, :insert_all")
-    end
 
     Builder.CTE.build(query, name, with_query, opts[:materialized], operation, __CALLER__)
   end
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1364,6 +1364,9 @@ defmodule Ecto.Query do
     * `:materialized` - a boolean indicating whether the CTE should
     be materialized. If blank, the database's default behaviour
     will be used (only supported by Postgrex, for the built-in adapters)
+    * `:operation` - one of `:all`, `:update_all`, `:delete_all`, `:insert_all` 
+    indicating the operation type of the CTE query. If blank, it defaults to `:all`, 
+    making the CTE query a SELECT query. (only supported by Postgres built-in adapter)
 
   ## Recursive CTEs
 
@@ -1430,6 +1433,25 @@ defmodule Ecto.Query do
       from(cte in {"category_tree", Category}, prefix: nil)
       |> recursive_ctes(true)
       |> with_cte("category_tree", as: ^category_tree_query)
+      
+  For Postgres built-in adapter, it is possible to define data-modifying CTE queries:
+  
+      update_categories_query =
+        Category
+        |> where([c], is_nil(c.parent_id))
+        |> update([c], set: [name: "Root category"])
+        |> select([c], c)
+
+      {"update_categories", Category}
+      |> with_cte("update_categories", as: ^update_categories_query, operation: :update_all)
+      |> select([c], c)
+      
+  Note: In order to retrieve the updates rows from a CTE query, the parent query 
+  must select rows from the CTE table instead of the table referenced by the CTE query.
+  For example, `"update_categories"` will return updates rows for `"category"` table, but
+  selecting from `"category"` table directly will return unaffected rows.
+  For more details see Postgres documentation on data-modifying CTEs and how these work 
+  with snapshots.
 
   Keyword syntax is not supported for this feature.
 
@@ -1446,12 +1468,17 @@ defmodule Ecto.Query do
   '''
   defmacro with_cte(query, name, opts) do
     with_query = opts[:as]
+    operation = opts[:operation]
 
     if !with_query do
       Builder.error!("`as` option must be specified")
     end
+    
+    unless operation in [nil, :all, :update_all, :delete_all, :insert_all] do
+      Builder.error!("`operation` option must be one of :all, :update_all, :delete_all, :insert_all")
+    end
 
-    Builder.CTE.build(query, name, with_query, opts[:materialized], __CALLER__)
+    Builder.CTE.build(query, name, with_query, opts[:materialized], operation, __CALLER__)
   end
 
   @doc """
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1051,7 +1051,7 @@ defmodule Ecto.Query.Planner do
           # We don't want to use normalize_subquery_select because we are
           # going to prepare the whole query ourselves next.
           {_, _, inner_query} = rewrite_subquery_select_expr(inner_query, true)
-          {inner_query, counter} = traverse_exprs(inner_query, :all, counter, fun)
+          {inner_query, counter} = traverse_exprs(inner_query, :all_cte, counter, fun)
 
           # Now compute the fields as keyword lists so we emit AS in Ecto query.
           %{select: %{expr: expr, take: take, aliases: aliases}} = inner_query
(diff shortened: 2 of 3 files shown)


lib/ecto/query/planner.ex
```diff
@@ -1783,7 +1783,7 @@ defmodule Ecto.Query.Planner do
 
   @all_exprs [with_cte: :with_ctes, distinct: :distinct, select: :select, from: :from, join: :joins,
               where: :wheres, group_by: :group_bys, having: :havings, windows: :windows,
-              combination: :combinations, order_by: :order_bys, limit: :limit, offset: :offset]
+              combination: :combinations, order_by: :order_bys, limit: :limit, offset: :offset, update: :updates]
```
v0idpwn, line 1786: Should be `:update`, no? We don't allow multiple update exprs

mpotra: [This test for allowing update](https://github.com/elixir-ecto/ecto_sql/pull/532/commits/f101edf52773ed583650df77bdfe97fd029f0a82#diff-df64a1e98eb69a93a67e388be3fb937b476bc769705a2a682d3e8e1ecdc8a38c) statements inside PostgreSQL CTEs fails unless `update: :updates`; I had tried all other variants for keys and nothing else works. Not sure I understand what you mean by multiple update exprs?

greg-rychlewski:
Maybe we want to have a separate set of fields for CTE? This would allow non-CTE queries to pass through `:update` but in the past it would be silently ignored.

greg-rychlewski (approved):
Thanks looks good to me. Just want to see if v0idpwn/Jose have any objections to the new operation type.

lib/ecto/query/planner.ex
```diff
@@ -1051,7 +1051,7 @@ defmodule Ecto.Query.Planner do
           # We don't want to use normalize_subquery_select because we are
           # going to prepare the whole query ourselves next.
           {_, _, inner_query} = rewrite_subquery_select_expr(inner_query, true)
-          {inner_query, counter} = traverse_exprs(inner_query, :all, counter, fun)
+          {inner_query, counter} = traverse_exprs(inner_query, :all_cte, counter, fun)
```
josevalim, line 1054:
```suggestion
          {inner_query, counter} = traverse_exprs(inner_query, :cte, counter, fun)
```

lib/ecto/query/planner.ex
```diff
@@ -1784,7 +1784,8 @@ defmodule Ecto.Query.Planner do
   @all_exprs [with_cte: :with_ctes, distinct: :distinct, select: :select, from: :from, join: :joins,
               where: :wheres, group_by: :group_bys, having: :havings, windows: :windows,
               combination: :combinations, order_by: :order_bys, limit: :limit, offset: :offset]
-
+  @all_cte_exprs @all_exprs ++ [update: :updates]
```
josevalim, line 1787: The order matters, because it is the orders of bindings, so I believe you will have to list all fields, with `update: :updates` coming immediately after `from: :from`.

josevalim:
New field is fine, we should just call it `:cte` though. I have also added a comment about traversal order.

lib/ecto/query/planner.ex
v0idpwn: oh, its indeed called `:updates` elsewhere, mb

lib/ecto/query/planner.ex
mpotra: Resolved by https://github.com/elixir-ecto/ecto/pull/4220/commits/879d58f52e0f11b40145cda9621add93532ca5c4

lib/ecto/query/planner.ex
mpotra: I've created a new attr list for CTE exprs in https://github.com/elixir-ecto/ecto/pull/4220/commits/879d58f52e0f11b40145cda9621add93532ca5c4 to place :update after :from

mpotra:
My latest commit introduced the  `:operation` option for `with_cte/3`, to support additional operations in Postgres CTEs.
This makes it not necessary to have a separate `:cte` traversal operation and allows for handling CTE operations at adapter level, enabling a non-breaking change for adapters + support for `:update_all` and `:delete_all`

lib/ecto/query/planner.ex
```diff
@@ -1042,16 +1042,23 @@ defmodule Ecto.Query.Planner do
...
+            _ -> :all
+          end
+
           inner_query = put_in(inner_query.aliases[@parent_as], query)
 
           # We don't want to use normalize_subquery_select because we are
           # going to prepare the whole query ourselves next.
           {_, _, inner_query} = rewrite_subquery_select_expr(inner_query, true)
-          {inner_query, counter} = traverse_exprs(inner_query, :all, counter, fun)
+          {inner_query, counter} = traverse_exprs(inner_query, operation, counter, fun)
```
josevalim, line 1061: Can we write this as:

```suggestion
          {inner_query, counter} = traverse_exprs(inner_query, opts.operation, counter, fun)
```

And then we make sure it is normalized to `:all` if `nil` is given when created?

lib/ecto/query.ex
```diff
@@ -1446,12 +1468,17 @@ defmodule Ecto.Query do
...
     with_query = opts[:as]
+    operation = opts[:operation]
 
     if !with_query do
       Builder.error!("`as` option must be specified")
     end
+    
+    unless operation in [nil, :all, :update_all, :delete_all, :insert_all] do
+      Builder.error!("`operation` option must be one of :all, :update_all, :delete_all, :insert_all")
+    end
```
josevalim, line 1479: This validation must happen inside `apply`. Here the validation is too early (if it is inside the macro, it is happening at compile-time). We should validate and convert `nil` to `:all` inside CTE.apply. :)

lib/ecto/query/planner.ex
mpotra: yes, this makes much more sense. Updated in the latest commit

lib/ecto/query.ex
mpotra: I've moved it inside `apply` but now it's referencing the wrong place, hinting at the query instead of the `with_cte/3` call. Are we OK with that? It still feels to me like this should be a compile-time error rather than a runtime one, on invalid option?

```
** (Ecto.Query.CompileError) `operation` option must be one of :all, :update_all, :delete_all, :insert_all
     code: |> select([c], %{id: c.id, desc: c.desc})
     stacktrace:
       test/ecto/adapters/postgres_test.exs:398: (test)
```

lib/ecto/query/planner.ex
```diff
@@ -1051,7 +1051,7 @@ defmodule Ecto.Query.Planner do
           # We don't want to use normalize_subquery_select because we are
           # going to prepare the whole query ourselves next.
           {_, _, inner_query} = rewrite_subquery_select_expr(inner_query, true)
-          {inner_query, counter} = traverse_exprs(inner_query, :all, counter, fun)
+          {inner_query, counter} = traverse_exprs(inner_query, opts[:operation], counter, fun)
```
josevalim, line 1054: since this is a map, this should work, right?

```suggestion
          {inner_query, counter} = traverse_exprs(inner_query, opts.operation, counter, fun)
```

lib/ecto/query/planner.ex
```diff
@@ -1799,6 +1799,10 @@ defmodule Ecto.Query.Planner do
 
   # Traverse all query components with expressions.
   # Therefore from, preload, assocs and lock are not traversed.
+  defp traverse_exprs(query, nil, acc, fun) do
```
josevalim, line 1802: We should remove this clause. It should never be nil.

lib/ecto/query/planner.ex
mpotra: Yes, you're right, I had forgotten it's always a map. Resolved in https://github.com/elixir-ecto/ecto/pull/4220/commits/b1667db38f24b39090b36ae5373589de5b273896

lib/ecto/query/planner.ex
mpotra: Removed in https://github.com/elixir-ecto/ecto/pull/4220/commits/b1667db38f24b39090b36ae5373589de5b273896

lib/ecto/query/planner.ex
```diff
@@ -1042,7 +1042,7 @@ defmodule Ecto.Query.Planner do
 
   defp validate_and_increment(:with_cte, query, with_expr, counter, _operation, adapter) do
     fun = &validate_and_increment(&1, &2, &3, &4, :all, adapter)
-
+
```
josevalim, line 1045:
```suggestion

```

lib/ecto/query/planner.ex
```diff
@@ -1042,7 +1042,6 @@ defmodule Ecto.Query.Planner do
 
   defp validate_and_increment(:with_cte, query, with_expr, counter, _operation, adapter) do
     fun = &validate_and_increment(&1, &2, &3, &4, :all, adapter)
```
josevalim, line 1044:
```suggestion
    fun = &validate_and_increment(&1, &2, &3, &4, :all, adapter)

```

josevalim:

PR #4221, opened 2023-06-29
Data modifying CTE cleanup

greg-rychlewski:
Some small cleanups from https://github.com/elixir-ecto/ecto/pull/4220

1. We aren't able to support `insert_all` right now.
2. The cache key should take the operation into account because it will affect the underlying query


greg-rychlewski:
ah have to fix the tests =/

josevalim (approved):

v0idpwn (approved):

PR #4222, opened 2023-06-29
Validate update_all CTEs

greg-rychlewski:
We could catch invalid update_all CTEs early.


greg-rychlewski:
My bad this is already happening in the planner.

Issue #2570, opened 2018-05-28
Switch to using = ANY(SELECT unnest(?)) instead of = ANY(?) for IN operator


michalmuskala:
The problem of `in (?, ?, ...)` is that it requires different SQL for different number of elements and in effect prevents caching queries. Because of that we switched to `= ANY(?)`, but that prevents using indices for the comparison.

@talentdeficit suggested using `= ANY(SELECT unnest(?))` instead, which supports using indices. The downside is that this relies on support of `unnest` function. To my knowledge the databases that don't support it (redshift) also don't support use of `= ANY(?)`, so it shouldn't be removing any feature.

josevalim:
Can we please do actual tests on this? Let's run some explain commands, benchmark on large data and what not. Otherwise, I don't think we should move forward.

josevalim:
I will close this for now, unless somebody is interested in running some tests and sharing their results. :)

PR #4228, opened 2023-07-13
Add splice to `fragment`

greg-rychlewski:
This is a proposal to add a splice capability to `fragment`. It is taking inspiration from `literal/1` as well as `unquote_splicing`.  

The idea is that the user would supply a list argument to `fragment` and then this would be spliced into the query string. Something like this

```elixir
a = 2
b = 3
query = from p in "posts", where: p.id in fragment("(?, ?, ?)", 1, splice(^[a, b, 4]), 5)

IO.inspect query
#Ecto.Query<from p0 in "posts", where: p0.id in fragment("(?, ?, ?, ?, ?)", 1, 2, 3, 4, 5)>
```

There are several use cases where this could come in handy:

1. Wanting to change the default Ecto behaviour using `fragment` when you require a variable number of arguments. For example changing the `where` operator behaviour: https://github.com/elixir-ecto/ecto/issues/2570#issuecomment-1632895072
2. Wanting to use `VALUE` lists that might have different sizes.
3. Some databases allow variadic functions. This would allow users to supply a variable number of arguments using the same fragment. This is a good fit because fragments would need to be used to call these types of functions anyway.

If the proposal is accepted I could finish up the PR. Just didn't want to get ahead of myself before receiving feedback.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -697,11 +697,11 @@ defmodule Ecto.Query.Builder do
   defp escape_fragment({:splice, _meta, [splice]}, params_acc, vars, env) do
     case splice do
       {:^, _, [value]} = expr ->
-          checked = quote do: Ecto.Query.Builder.splice!(unquote(value))
-          length = quote do: length(unquote(checked))
-          {expr, params_acc} = escape(expr, {:splice, :any}, params_acc, vars, env)
-          escaped =  {:{}, [], [:splice, [], [expr, length]]}
-          {escaped, params_acc}
+        checked = quote do: Ecto.Query.Builder.splice!(unquote(value))
+        length = quote do: length(unquote(checked))
+        {expr, params_acc} = escape(expr, {:splice, :any}, params_acc, vars, env)
+        escaped =  {:{}, [], [:splice, [], [expr, length]]}
+        {escaped, params_acc}
 
       _ ->
         error! "splice/1 in fragment expects an interpolated list, such as splice(^[1, 2, 3]), got `#{Macro.to_string(splice)}`"
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -699,8 +699,8 @@ defmodule Ecto.Query.Builder do
     case spliced do
       {:^, _, [_]} = expr ->
           {expr, params_acc} = escape(expr, {:splice, :any}, params_acc, vars, env)
-          escaped_splice =  {:{}, [], [:splice, [], [expr]]}
-          {escaped_splice, params_acc}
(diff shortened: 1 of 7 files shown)


greg-rychlewski:
It's probably weird they are not collected as parameters. I could see if that could be handled in the planner.

greg-rychlewski:
I think something like this could work actually.

josevalim:
Would it make a difference if we use the notation inside the fragment? Something like "(?, *?, ?)"?

greg-rychlewski:
I believe it will be the same issue. It looks to me like the underlying issue is being able to convert a runtime list from a single param to a dynamic number of params.

WTDY of something like this? I am trying to "flag" the parameters so that they can be treated properly at runtime if needed. Probably this is not a good flag but was just curious what you thought of the general idea?

greg-rychlewski:
Oh I see it now. Thank you for pointing me to that.

greg-rychlewski:
WDYT about this way? If it's good I can add the docs.

Companion ecto sql PR: https://github.com/elixir-ecto/ecto_sql/pull/535

lib/ecto/query/builder.ex
```diff
@@ -74,8 +74,8 @@ defmodule Ecto.Query.Builder do
   with `^index` in the query where index is a number indexing into the
   map.
   """
-  @spec escape(Macro.t, quoted_type | {:in, quoted_type} | {:out, quoted_type}, {list, acc},
-               Keyword.t, Macro.Env.t | {Macro.Env.t, fun}) :: {Macro.t, {list, acc}}
+  @spec escape(Macro.t, quoted_type | {:in, quoted_type} | {:out, quoted_type} | {:splice, quoted_type},
+               {list, acc}, Keyword.t, Macro.Env.t | {Macro.Env.t, fun}) :: {Macro.t, {list, acc}}
```
josevalim, line 78: Question, could we reuse `:in` instead of introducing `:splice`? It seems they behave the same inside the planner when it comes to parameters. In the AST they can still be separate nodes.

josevalim: Oh, maybe a reason why we can't unify is because PG and MySQL will treat them differently, so ignore me.

josevalim:
LGTM! Please don't forget to add splice to Ecto.Query.API as well. :)

josevalim (approved):

lib/ecto/query/builder.ex
greg-rychlewski: > Oh, maybe a reason why we can't unify is because PG and MySQL will treat them differently
Yeah exactly this. At first I wanted to do the same thing but then postgres was removing the `:in`.

Issue #4225, opened 2023-07-12
Prefix missing in association on_delete query


tfwright:
### Elixir version

1.13.4

### Database and Version

postgres 13

### Ecto Versions

3.10

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex

### Current behavior

Similar issue to https://github.com/elixir-ecto/ecto/issues/4183 (affecting `prepare_query`)

Add has_many association with :on_delete option
Delete parent record, loaded from prefix
Observe query.prefix is nil

### Expected behavior

query.prefix should be same as parent

josevalim:
Perfect. so I agree delete_all and nilify_all should use the prefix. PRs welcome too!

PR #4230, opened 2023-07-14
Add parent prefix to association parent on_delete query

tfwright:
Resolved https://github.com/elixir-ecto/ecto/issues/4225

Seeking initial feedback before I try this out in a local project.

From what I can tell, this should be the only place where this needs to be handled for all of the associations that support this option?

I could use some pointers on testing. I didn't see any obvious place to add tests. It doesn't look like the on_delete queries are tested in repo tests, I assume because the query is not returned anywhere in the public API. Maybe integration test?

### DIFF
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -996,7 +996,14 @@ defmodule Ecto.Association.Has do
          parent
        ) do
     if value = Map.get(parent, owner_key) do
-      from x in queryable, where: field(x, ^related_key) == ^value
+      query = from x in queryable, where: field(x, ^related_key) == ^value
+
+      parent
+      |> Ecto.get_meta(:prefix)
+      |> case do
+        nil -> query
+        prefix -> Ecto.Query.put_query_prefix(query, prefix)
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


josevalim:
For the tests, take a look at many_to_many. It seems we have some tests to verify the proper `delete_all` queries are sent and you can start asserting the prefix on those. Also note that integration tests are failing. It may be that you should set the query prefix only the the parent  has a query prefix (otherwise you may discard the default prefix of the schema).

josevalim:
Perhaps that's covered for many to many but not the other association types. So you can mirror those tests into the other associations. :)

lib/ecto/association.ex
```diff
@@ -996,7 +996,14 @@ defmodule Ecto.Association.Has do
...
     if value = Map.get(parent, owner_key) do
-      from x in queryable, where: field(x, ^related_key) == ^value
+      query = from x in queryable, where: field(x, ^related_key) == ^value
+
+      parent
+      |> Ecto.get_meta(:prefix)
+      |> case do
+        nil -> query
+        prefix -> Ecto.Query.put_query_prefix(query, prefix)
+      end
```
greg-rychlewski, line 1006:
```suggestion
      case Ecto.get_meta(parent, :prefix) do
        nil -> query
        prefix -> Ecto.Query.put_query_prefix(query, prefix)
      end
```

josevalim:
Looks good to me @tfwright! Can you please do the same changes for `nillify_all` and include tests? It also has the wrong behaviour. :)

tfwright:
@josevalim I think the change handles both behaviors but I added tests I think should prove it

josevalim (approved):

greg-rychlewski (approved):

greg-rychlewski:
Thanks! For this and the other prefix bugs. You found a lot of them =).

PR #4233, opened 2023-07-19
Allow `nil` map updates in select

greg-rychlewski:
**Reference** 

https://elixirforum.com/t/ecto-merge-select-or-select-merge-into-a-possibly-nil-value-left-join/57133

**Issue**

When performing a query like this, it will fail with `BadMapError` if one of the main rows doesn't have an associated row:

```elixir
from(c in Comment, left_join: p in Post, on: c.post_id == p.id, select: %{p | temp: c.text}
```

**Proposal**

I think we should allow this and simply return `nil`. It seems like the intuitive thing to expect and also there is some precedent. We will return `nil` in this situation

```elixir
from(c in Comment, left_join: p in Post, on: c.post_id == p.id, select: map(p, [:field1, :field2, ...])
```

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1360,6 +1360,17 @@ defmodule Ecto.Integration.RepoTest do
       end
     end
 
+    test "map update on association" do
+      p = TestRepo.insert!(%Post{})
+      TestRepo.insert!(%Comment{post_id: p.id, text: "comment text"})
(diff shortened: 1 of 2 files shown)


josevalim (approved):

greg-rychlewski:
@aiwaiwa You could give this a try :).

aiwaiwa:
Coool. Wow!

Issue #4235, opened 2023-07-19
Embed changesets don't always give errors for duplicates primary keys


mattnenterprise:
### Elixir version

1.15.2

### Database and Version

PostgreSQL 11

### Ecto Versions

3.9.5

### Database Adapter and Versions (postgrex, myxql, etc)

0.16.5

### Current behavior

If an embeds changeset has a duplicate id as apart of the changes it doesn't give a duplicate id error. This works correctly and gives an error if the id is in the data of the changeset.

For example the following test from the ecto codebase passes as it returns an error that the primary key id has already been taken.
```elixir
test "duplicate pk on insert" do
    embeds = [%MyEmbed{x: "xyz", id: @uuid} |> Ecto.Changeset.change,
              %MyEmbed{x: "abc", id: @uuid} |> Ecto.Changeset.change]
    changeset =
      %MySchema{}
      |> Ecto.Changeset.change
      |> Ecto.Changeset.put_embed(:embeds, embeds)
    assert {:error, changeset} = TestRepo.insert(changeset)
    refute changeset.valid?
    errors = Ecto.Changeset.traverse_errors(changeset, fn {msg, _} -> msg end)
    assert errors == %{embeds: [%{}, %{id: ["has already been taken"]}]}
  end
  ```
  
If however we modify the test to where the second embed changeset has the id as part of the changes it give no error and fails.
```elixir
test "duplicate pk on insert" do
    embeds = [%MyEmbed{x: "xyz", id: @uuid} |> Ecto.Changeset.change,
              %MyEmbed{} |> Ecto.Changeset.change(x: "abc", id: @uuid)]
    changeset =
      %MySchema{}
      |> Ecto.Changeset.change
      |> Ecto.Changeset.put_embed(:embeds, embeds)
    assert {:error, changeset} = TestRepo.insert(changeset)
    refute changeset.valid?
    errors = Ecto.Changeset.traverse_errors(changeset, fn {msg, _} -> msg end)
    assert errors == %{embeds: [%{}, %{id: ["has already been taken"]}]}
  end
  ```
  
  The only thing changed in the above test is that the second embed changeset is created with
  ```elixir
  %MyEmbed{} |> Ecto.Changeset.change(x: "abc", id: @uuid)
  ```
  instead of 
  ```elixir
   %MyEmbed{x: "abc", id: @uuid} |> Ecto.Changeset.change
  ```

### Expected behavior

[...]

PR #4237, opened 2023-07-20
Fix duplicate ID check for `many` assocs/embeds

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4235

The underlying issue is that the function that extracts the pk values to check for duplicates is the same as the function that extracts the pk values to match current values to changes. The latter has to extract from `changeset.data` while the former should check for `changeset.changes` first and then `changeset.data`.

### DIFF
--- a/lib/ecto/changeset/relation.ex
+++ b/lib/ecto/changeset/relation.ex
@@ -515,9 +515,9 @@ defmodule Ecto.Changeset.Relation do
   defp change_pk(pks) do
     fn %Changeset{} = cs ->
       Enum.map(pks, fn pk ->
-        case Map.fetch(cs.changes, pk) do
-          {:ok, pk_value} -> pk_value
-          :error -> Map.get(cs.data, pk)
+        case cs.changes do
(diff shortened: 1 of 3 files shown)


lib/ecto/changeset/relation.ex
```diff
@@ -510,6 +512,17 @@ defmodule Ecto.Changeset.Relation do
...
+  defp change_pk(pks) do
+    fn %Changeset{} = cs ->
+      Enum.map(pks, fn pk ->
+        case cs.changes do
+          %{^pk => pk_value} -> pk_value
+          _ -> Map.get(cs.data, pk)
+        end
+      end)
+    end
+  end
```
greg-rychlewski, line 524: It looks to me like anything getting to `changeset.changes` has its key normalized to atom and value casted to the right type. But please correct me if I'm wrong.

josevalim: Correct!

josevalim (approved):

PR #4239, opened 2023-07-21
Allow safe interpolation on Fragment query strings

bamorim:
I was playing around with the idea yesterday and I saw there was already a discussion regarding that in the mailing list, so I'm creating this PR.

## Problem

When writing complex fragments with a lot of arguments, it can become easily a mess due to positional arguments. To know where something would go in the query, you have to count the `?` and then count the `,` (or lines if you have in multiple lines) to be able to know where everything goes.

## Solution

The idea is to allow interpolation using a sigil that just compiles to the same `fragment` call underneath.

One thing to note that I'd like input on:

When I checked the mailing list, I saw @josevalim suggested `~FRAGMENT`, but because it is an uppercase sigil, it does not automatically interpolate the contents (which also means editor support don't treat `#{}` as interpolations and you get worse autocomplete).

We could probably just get the string, parse it and then it would work the same, but we would still have worse support for tooling.

Because of that, I went with `~f` instead, but of course this is something we might want to discuss.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2694,4 +2694,51 @@ defmodule Ecto.Query do
           end)
     }
   end
+
+  @doc """
+  Syntax sugar for writing fragments with interpolations.
+
+      dynamic([user: user], ~f\"""
+        CASE
+          WHEN \#{user.role == :admin} THEN 3
+          WHEN \#{user.role == :manager} THEN 2
+          ELSE 1
+        END
+      \""")
+
+  Is equivalent to
+
+      dynamic([user: user], fragment(
+        \"""
+          CASE
+            WHEN ? THEN 3
+            WHEN ? THEN 2
+            ELSE 1
+          END
+        \""",
+        user.role == :admin,
+        user.role == :manager
+      ))
+
+  But with the benefit of interpolated values being in the right place.
+  """
+  defmacro sigil_f({:<<>>, _meta, pieces}, _) do
+    query =
+      pieces
+      |> Enum.map(fn
+        "" <> binary -> binary
+        _ -> "?"
+      end)
+      |> Enum.join()
+
+    args =
+      Enum.flat_map(pieces, fn
+        {:"::", _, [{_, _, [val]} | _]} -> [val]
+        _ -> []
+      end)
+
+    quote do
+      fragment(unquote_splicing([query | args]))
+    end
+  end
 end
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2694,49 +2694,4 @@ defmodule Ecto.Query do
           end)
     }
   end
-
-  @doc """
-  Syntax sugar for writing fragments with interpolations.
-
-      dynamic([user: user], ~f\"""
-        CASE
-          WHEN \#{user.role == :admin} THEN 3
-          WHEN \#{user.role == :manager} THEN 2
-          ELSE 1
-        END
-      \""")
-
-  Is equivalent to
-
-      dynamic([user: user], fragment(
-        \"""
-          CASE
-            WHEN ? THEN 3
-            WHEN ? THEN 2
-            ELSE 1
-          END
-        \""",
-        user.role == :admin,
-        user.role == :manager
-      ))
-
-  But with the benefit of interpolated values being in the right place.
-  """
-  defmacro sigil_f({:<<>>, meta, pieces}, _) do
-    query =
-      pieces
-      |> Enum.map(fn
-        "" <> binary -> binary
-        _ -> "?"
-      end)
-      |> Enum.join()
-
-    args =
-      Enum.flat_map(pieces, fn
-        {:"::", _, [{_, _, [val]} | _]} -> [val]
-        _ -> []
-      end)
-
-    {:fragment, meta, [query | args]}
-  end
 end
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2722,7 +2722,7 @@ defmodule Ecto.Query do
 
   But with the benefit of interpolated values being in the right place.
   """
-  defmacro sigil_f({:<<>>, _meta, pieces}, _) do
+  defmacro sigil_f({:<<>>, meta, pieces}, _) do
     query =
       pieces
       |> Enum.map(fn
@@ -2737,8 +2737,6 @@ defmodule Ecto.Query do
         _ -> []
       end)
 
-    quote do
-      fragment(unquote_splicing([query | args]))
-    end
+    {:fragment, meta, [query | args]}
   end
 end
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -478,33 +478,33 @@ defmodule Ecto.Query.API do
   """
   def fragment(fragments), do: doc! [fragments]
 
-  @doc """
+  @doc ~S'''
   Syntax sugar for writing fragments with interpolations.
 
-      dynamic([user: user], ~f\"""
+      dynamic([user: user], ~f"""
         CASE
-          WHEN \#{user.role == :admin} THEN 3
-          WHEN \#{user.role == :manager} THEN 2
+          WHEN #{user.role == :admin} THEN 3
+          WHEN #{user.role == :manager} THEN 2
           ELSE 1
         END
-      \""")
+      """)
 
   Is equivalent to
 
       dynamic([user: user], fragment(
-        \"""
+        """
           CASE
             WHEN ? THEN 3
             WHEN ? THEN 2
             ELSE 1
           END
-        \""",
+        """,
         user.role == :admin,
         user.role == :manager
       ))
 
   But with the benefit of interpolated values being in the right place.
-  """
+  '''
   def sigil_f(ast, modifiers), do: doc! [ast, modifiers]
 
   @doc """
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -446,6 +446,16 @@ defmodule Ecto.Query.API do
 
       from p in Post, where: fragment("? in (?,?,?)", p.id, ^1, ^2, ^3)
 
+  ## Interpolation
+
+  As an alternative splicing, fragments also support safe interpolations
+  as a syntax sugar for splicing, so the query above could be rewriten as
+
+      from p in Post, where: fragment("#{p.id} in (#{^1},#{^2},#{^3})")
+
+  Which has the benefit of more easily understanding where each expression
+  is used.
+
   ## Defining custom functions using macros and fragment
 
   You can add a custom Ecto query function using macros.  For example
@@ -477,35 +487,6 @@ defmodule Ecto.Query.API do
 
   """
   def fragment(fragments), do: doc! [fragments]
-
-  @doc ~S'''
-  Syntax sugar for writing fragments with interpolations.
-
-      dynamic([user: user], ~f"""
-        CASE
-          WHEN #{user.role == :admin} THEN 3
-          WHEN #{user.role == :manager} THEN 2
-          ELSE 1
-        END
-      """)
-
-  Is equivalent to
-
-      dynamic([user: user], fragment(
-        """
-          CASE
-            WHEN ? THEN 3
-            WHEN ? THEN 2
-            ELSE 1
-          END
-        """,
-        user.role == :admin,
-        user.role == :manager
-      ))
-
-  But with the benefit of interpolated values being in the right place.
-  '''
-  def sigil_f(ast, modifiers), do: doc! [ast, modifiers]
 
   @doc """
   Allows a literal identifier to be injected into a fragment:
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -479,6 +479,35 @@ defmodule Ecto.Query.API do
   def fragment(fragments), do: doc! [fragments]
 
   @doc """
+  Syntax sugar for writing fragments with interpolations.
+
+      dynamic([user: user], ~f\"""
+        CASE
+          WHEN \#{user.role == :admin} THEN 3
+          WHEN \#{user.role == :manager} THEN 2
+          ELSE 1
+        END
+      \""")
+
+  Is equivalent to
+
+      dynamic([user: user], fragment(
+        \"""
+          CASE
+            WHEN ? THEN 3
+            WHEN ? THEN 2
+            ELSE 1
+          END
+        \""",
+        user.role == :admin,
+        user.role == :manager
+      ))
+
+  But with the benefit of interpolated values being in the right place.
+  """
+  def sigil_f(ast, modifiers), do: doc! [ast, modifiers]
+
+  @doc """
   Allows a literal identifier to be injected into a fragment:
 
       collation = "es_ES"
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -162,7 +162,7 @@ defmodule Ecto.Query.Builder do
   end
 
   # fragments
-  def escape({:sigil_f, _sigil_meta, [{:<<>>, _meta, pieces}, _modifiers]}, given_type, params_acc, vars, env) do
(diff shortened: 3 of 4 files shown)


lib/ecto/query.ex
```diff
@@ -2694,4 +2694,51 @@ defmodule Ecto.Query do
...
+        "" <> binary -> binary
+        _ -> "?"
+      end)
+      |> Enum.join()
+
+    args =
+      Enum.flat_map(pieces, fn
+        {:"::", _, [{_, _, [val]} | _]} -> [val]
+        _ -> []
+      end)
```
bamorim, line 2738: My initial implementation was using `Macro.traverse`, but I think this is much cleaner and easier to understand.

greg-rychlewski:
Thanks for this. I don't have time to review too closely yet but one thing I noticed is we want to avoid creating a macro and basically just pattern match in the builder. Similar to how we pattern match on `{:fragment, _, _}` in `Ecto.Query.Builder`.

bamorim:
@greg-rychlewski is that something that is possible? I think sigils are processed before being passed down to macros like the query ones.

bamorim:
Nevermind, I think I know how to get it working, I was messing things up.

bamorim:
@greg-rychlewski it was an easy change! :)

greg-rychlewski:
Awesome :). 

Probably need to see what Jose is thinking about `~f`. I believe the idea behind `~FRAGMENT` was to be more explicit and also make it clear it's not string interpolation. But not sure how that weighs against the syntax highlighting/tooling. I am biased because my tooling is very basic and most of the time I don't notice the syntax highlights so will let you guys debate it :).

Some other things that probably need thinking/debate:

```elixir
~f"GREATEST(#{t.date}, #{^date})")
```

- I think we want to convert every sigil interpolation into a query parameter instead of having the user use the pin operator in the string. Basically it is all dynamic input so needs to be treated safely.
- I don't think we can infer query bindings in here because they will potentially conflict with maps/structs?

bamorim:
> Some other things that probably need thinking/debate:
> 
> ```elixir
> ~f"GREATEST(#{t.date}, #{^date})")
> ```
> 
> * I think we want to convert every sigil interpolation into a query parameter instead of having the user use the pin operator in the string. Basically it is all dynamic input so needs to be treated safely.
> * I don't think we can infer query bindings in here because they will potentially conflict with maps/structs?

Not sure I understood the points, sorry. In my mind, this is just syntax sugar for `fragment` so everything else is still the same as `fragment` would be. In fact, we just translate as if the user have written `fragment(query, ...frags)`. That is, the user would have to use `^date` in fragment (`fragment("GREATEST(?, ?)", t.date, ^date)`) so I left it the same way.

Would you mind elaborating a little bit more the points?

greg-rychlewski:
Sorry you're right I was getting mixed up.

lib/ecto/query/api.ex
```diff
@@ -478,6 +478,35 @@ defmodule Ecto.Query.API do
   """
   def fragment(fragments), do: doc! [fragments]
 
+  @doc """
```
josevalim, line 481:
```suggestion
  @doc ~S'''
```

This way you don't need to escape `"""` in the examples inside nor interpolations.

wojtekmach:
I think this is pretty great. I think ~f over ~FRAGMENT is a good choice, interpolation is really key here. My only minor concern is, typical to sigils, we'd have ~f in Ecto and there's ~F in Surface. I don't think they'd be commonly used together and in general I don't think it's a deal breaker but still something to consider.

If we had lower-case multi letter sigils would we still go for ~f instead of ~fragment? And if we'd go with ~fragment should we just implement interpolations in fragment, exactly like you did? We can totally do that because up to this point we raised:

```
iex> from u in "users", select: fragment("LOWER(#{u.name})")
** (Ecto.Query.CompileError) to prevent SQL injection attacks, fragment(...) does not allow strings to be interpolated as the first argument via the `^` operator, got: `"LOWER(#{u.name})"`
    (ecto 3.11.0-dev) expanding macro: Ecto.Query.select/3
    iex:4: (file)
    (ecto 3.11.0-dev) expanding macro: Ecto.Query.from/2
    iex:4: (file)
```

and now we'd:

```elixir
iex> from u in "users", select: fragment("LOWER(#{u.name})")
#Ecto.Query<from u0 in "users", select: fragment("LOWER(?)", u0.name)>
```

WDYT?

bamorim:
@wojtekmach OMG, you have some really great points.

At this point in Ecto we are already having full control over the AST and transforming it in multiple ways so to be honest, sigils don't add much here other than giving people a sense that they are still in a familiar "elixir land" space, so all of these options are valid.

Regarding `~f` vs `~fragment`, `~f` has the benefit of being really small and removing some noise, making the query look even more like SQL (`fragment` is not a thing we would see in SQL), so that seems great.

However, Ecto users are already familiar with the idea of a fragment and it is something people are already using, so just adding the functionality of interpolation to that seems like a no-brainer to me, however we would probably have to limit it to `fragment/1` because it would be weird to mix the two styles.

The implementation for that is pretty similar, if you want I can update the PR with that.

bamorim:
@wojtekmach it was quite simple to experiment with the fragment idea, these were the tests I have passing here now:

```elixir
    test "raises at compile time when mixing interpolation and positional arguments" do
      assert_raise Ecto.Query.CompileError,
                   ~r/fragment\(...\) does not allow mixing interpolation and positional arguments/, fn ->
        quote_and_eval from p in "posts", where: fragment("GREATEST(#{t.date}, ?)", t.other)
      end
    end

    test "builds a fragment with one interpolation" do
      assert inspect(select("t", [t], fragment("LOWER(#{t.email})"))) ==
               inspect(select("t", [t], fragment("LOWER(?)", t.email)))
    end

    test "builds a fragment with multiple interpolation" do
      date = ~D[2020-01-01]

      assert inspect(select("t", [t], fragment("GREATEST(#{t.date}, #{^date})"))) ==
               inspect(select("t", [t], fragment("GREATEST(?, ?)", t.date, ^date)))
    end
 ```

File: test/ecto/query_test.exs [hunk 3]
```diff
@@ -1133,4 +1133,22 @@ defmodule Ecto.QueryTest do
...
+               inspect(select("t", [t], fragment("LOWER(?)", t.email)))
+    end
+
+    test "builds a fragment with multiple interpolation" do
+      date = ~D[2020-01-01]
+
+      assert inspect(select("t", [t], ~f"GREATEST(#{t.date}, #{^date})")) ==
+               inspect(select("t", [t], fragment("GREATEST(?, ?)", t.date, ^date)))
+    end
+  end
```
christhekeele, line 1153: We should probably add a test for interpolating the same variable multiple times. It also prompts the discussion: does doing this create multiple positional arguments, repeatedly interpolating the same value? If so, are we ok with that, what are the trade-offs? Are there better ways? etc

lib/ecto/query/builder.ex
```diff
@@ -162,6 +162,24 @@ defmodule Ecto.Query.Builder do
   end
 
   # fragments
+  def escape({:sigil_f, _sigil_meta, [{:<<>>, _meta, pieces}, _modifiers]}, given_type, params_acc, vars, env) do
+    query =
+      pieces
+      |> Enum.map(fn
+        "" <> binary -> binary
```
christhekeele, line 169: Would this read better or more idiomatically as `binary when is_binary(binary) -> binary`? Or is there a semantic difference in this approach that I am missing?

bamorim: To be honest, I haven't thought much about that, just did the "fastest" way to pattern match on a binary. I don't have a strong opinion about that.

test/ecto/query_test.exs
bamorim: I don't think we need a test for interpolating the same variable multiple times as there is no special case for that. We just translate everything to `?` and add the interpolation as argument to `fragment`.

Of course this means we also inherit all the same "semantics" from `fragment`,  which include this repetition and the only current way of fixing that would be using either subqueries or `dynamic`.

IMHO, even though I agree that these are some pain points, I don't know if it is directly related to this PR.

bamorim: The idea from the mailing list tries to address that, the example given by Phillip was this:

```elixir
fragment(
  ":service_year < :year OR (:service_year = :year AND :service_month <= :month)",
  service_year: service.invoice_year,
  service_month: service.invoice_month,
  year: ^max_year,
  month: ^max_month
)
```

The challenge however was that this would require a much more complex parser for the query string.

Maybe we could do something like this

```elixir
fragment(
  "#{arg(:service_year)} < #{^max_year} OR (#{arg(:service_year)} = #{^max_year} AND #{service.invoice_month} <= #{^max_month})",
  service_year: service.invoice_year
)
```

(If we followed the idea of just adding interpolation to `fragment` instead of adding a sigil).

However, this would require adding either a reserved function name like `arg` or  reserve another unary operator, which is probably harder.

This example is not great, of course, but even if `service_year` was another complex fragment, I'm not sure what is the benefit of that vs just doing a subquery to compute and then re-use other than giving more options.

But yeah, the amount of things I wrote show that this is a deeper problem then just allowing interpolation so probably should be discussed separately.

greg-rychlewski:
Take this as another data point and not a disapproval or anything like that. 

IMO if there is a good way to do named parameters that would be a friendlier syntax. When I was look at this syntax my first reaction was about SQL injection before realizing I am seeing a fragment. It was also a bit jarring for me to see bindings and pins inside of a string.

bamorim:
> Take this as another data point and not a disapproval or anything like that. 
> 
> IMO if there is a good way to do named parameters that would be a friendlier syntax. When I was look at this syntax my first reaction was about SQL injection before realizing I am seeing a fragment. It was also a bit jarring for me to see bindings and pins inside of a string.

That is a good argument for doing the interpolation in the fragment instead of ~f since the sigil doesn't make it clear it is a fragment (you have to know that f stands for fragment).

I particularly don't like the idea of named paras because for me they still suffer from the same problem of having to look at two different places to get the full concept, just a little bit better because you now can give better names. The only benefit of named args is that now you can reuse the same value, but I'd still vote for interpolation because for most cases it's much simpler.

josevalim:
I think we go back to the point of it looking too much like string interpolation. With sigils it is at least excusable because sigils often override it. I think our options are either to use a different interpolation syntax (such as `?{...}`) or keywords lists, but I am personally not convinced of any. :(

bamorim:
But is it a problem to be too much like string interpolation if it is still safe?

I mean, I see this more of a feature and a showcase of how great macros can be that you can code in a way similar to the naive way and still be safe rather than a problem.

wojtekmach:
I know I just proposed string interpolation but given the conversation above and in mailing list discussion that I'm just catching up on, yeah, it's tricky and we need to be careful.

I think the danger is we teach people it's OK to do `fragment("...#{foo}")` and they are so used to it that this one time they need a totally custom SQL query they'd do `Repo.query!("...#{foo}")` and they'd be busted. I don't know what the solution is, this is just my read on the situation.

bamorim:
That's a good point. Do you think the sigil mitigates that problem a little based on the argument Valim have above?

josevalim:
I think sigil is not an option, anymore, based on everything said. ~f is too confusing, ~FRAGMENT does not allow interpolation, and if we are going to use ~FRAGMENT with another interpolation, we might as well use the existing `fragment` macro.

bamorim:
I changed the code to interpolate directly on `fragment`, it's now about us to decide whether or not that is bad.

I particularly think that regardless of whether this feature exists or not, we should warn users that `Repo.query!` doesn't do escaping and that they should be careful.

greg-rychlewski:
We could potentially have something like this

```elixir
fragment("?{a} <= ?{b} - ?{a}", a: ^1, b: ^2)
```

But I remember Jose brought up a good point that in the case above 3 query parameters are generated instead of 2. So the side effects could be unexpected. 

Then I think it's a debate of whether having the naming ability outweighs this. This topic seemed to generate a good amount of interest, so maybe it's worth reconsidering. The behaviour could be documented and if the user will end up using the current fragment they will generate that many query parameters anyway.

bamorim:
It think I'll edit the title of this PR to be about interpolation in Fragment and we can open another PR with an implementation for named paras so we can then compare them.

josevalim:
I don't think it is an implementation concern but an API concern. The only API that addresses all complaints is `fragment("?{...}")` and I am just not sure if it is worth it.

tessi:
Hmmm what about

```elixir
fragment("date_add(#{:a}, #{:b})", a: p.created_at, b: ^0)
```

(allowing only symbols to exist in interpolations)

* easy to implement and maintain with a very similar approach @bamorim took in this PR
* safe regarding fears of SQL injections and safety
* by using kw args, duplicated params are not a problem - esp. when they get long
* syntax highlighting works
* it is easy to see what `:a` or `:b` refer to in the interpolations, reducing the learning curve

It doesn't solve locality (the interpolated expression being within the query), but the current "?" approach doesn't solve that either and the discussion (here or in the mailing list) didn't express this as a must-have feature.

josevalim:
A keyword syntax is misleading because repeated arguments will be interpolated multiple times. I don’t think it is an option, unfortunately.

tessi:
if that's a good option, I believe this should do the trick if this is the first `escape({:fragment, ...)` definition:

```elixir
defmodule Ecto.Query.Builder do
  # ...

  # fragments
  def escape({:fragment, _, [{:<<>>, _meta, pieces} | [params]]}, given_type, params_acc, vars, env) when is_list(params) do
    if not Keyword.keyword?(params) do
      error! "fragment(...) with named params expect a keyword list as the last argument, got: #{Macro.to_string(params)}"
    end

    query =
      pieces
      |> Enum.map(fn
        "" <> binary -> binary
        _ -> "?"
      end)
      |> Enum.join()

    frags =
      Enum.flat_map(pieces, fn
        {:"::", _, [{_, _, [val]} | _]} when is_atom(val) -> [Keyword.get(params, val)]
        {:"::", _, [{_, _, [val]} | _]} -> error! "named params in fragment(...) queries must be atoms, got: #{Macro.to_string(val)}"
        _ -> []
      end)

    escape({:fragment, [], [query | frags]}, given_type, params_acc, vars, env)
  end

  # ...
end
```

tessi:
I see. This is because people would expect params to be given to the DB only once where we would duplicate them and give them N times?

One way out would be to forbid use of duplicated names. If `fragment("foo(#{:a}, #{:a})", a: 0)` was used, we could either warn the user (not sure how) or error with a nice message regarding that concern and suggesting the user could use `fragment("foo(#{:a1}, #{:a2})", a1: 0, a2: 0)` instead.

josevalim:
Right, but then I don’t see what the keyword list gives compared to using something like ?{..}  for interpolation. I think those trade-offs are fine for a library but i am torn about adding them to Ecto per se :(

bamorim:
But is really making a safe-interpolation option here a bad thing?

People running `Repo.query!` are already in "unsafe land", so it is worth not adding a convenient feature with fear that this might lead people to use it wrongly? It is barely documented and it is already an advanced use case where people "should know what they are doing ™️".

If we are worried about people misusing `Repo.query`, wouldn't be better to maybe invest sometime in making `Repo.query!` safe then?

For example, maybe we could instead make Repo.query! a macro that would work in string literals and interpolation like `fragment` but would require some sort of opt-in for runtime built strings (like html_safe in Rails).

wojtekmach:
Changing Repo.query to a macro would be a breaking change. An easy one to fix but one nonetheless.

Let's keep discussing.

josevalim:
Doing it in Repo.query has the downside that someone can simply refactor the code. If you do this, it is safe:

    Repo.query("SELECT #{...}", ...)

But if you refer to this, it is unsafe:

    string = "SELECT #{...}"
    Repo.query!(string, ...)

We could force it to always be a string, but then it is an even larger breaking change.

As said above, I really don't think we should "overload" interpolation. Keyword lists will cause confusion too. "The only API that addresses all complaints is fragment("?{...}") and I am just not sure if it is worth it."

josevalim:
Closing this for now, as I am not comfortable with the options here. Sorry and thanks!

Issue #4224, opened 2023-07-11
Ecto.UUID.cast casts a string value that is not a valid UUID


dantuck:
### Elixir version

1.14.0

### Database and Version

PostgreSQL 15.2

### Ecto Versions

3.8.4

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex

### Current behavior

`Ecto.UUID.cast!("warehouse worker")`

The output is `"77617265-686f-7573-6520-776f726b6572"`

### Expected behavior

** (Ecto.CastError) cannot cast "warehouse worker" to Ecto.UUID
    (ecto 3.8.4) lib/ecto/uuid.ex:54: Ecto.UUID.cast!/1
    iex:1: (file)

josevalim:
Yes, we match 16 bytes UUIDs (their raw representation) and any 16 bytes are valid. You can consult previous reports for more info. :)

lytedev:
@josevalim Would you accept a contribution to the documentation for this function that provides this context? Looks like you've posted the same response a couple of other times and perhaps clarifying this case in the functions docstring would be beneficial -- even though its typespec sort of provides some clarity.

Previous reports:

- https://github.com/elixir-ecto/ecto/issues/4174
- https://github.com/elixir-ecto/ecto/issues/3125

Issue #4174, opened 2023-05-03
Ecto.UUID.cast succeeds on non-uuid strings with 16 characters


omginbd:
### Elixir version

Elixir 1.14.0; Erlang/OTP 25

### Database and Version

N/A

### Ecto Versions

3.9.4

### Database Adapter and Versions (postgrex, myxql, etc)

N/A

### Current behavior

Currently, given a non-uuid string of length 16 like "xxxxxxxxxxxxxxxx":

1. Call `Ecto.UUID.cast("xxxxxxxxxxxxxxxx")`
2. The cast succeeds and returns `{:ok, "78787878-7878-7878-7878-787878787878"}`

### Expected behavior

I would expect that given a non-uuid string of length 16 like "xxxxxxxxxxxxxxxx":

1. Calling `Ecto.UUID.cast("xxxxxxxxxxxxxxxx")`
2. Would fail to cast and instead return `:error`

I would expect this to be caught by some of the tests [here](https://github.com/elixir-ecto/ecto/blob/master/test/ecto/uuid_test.exs#L12), but it seems to specifically require a 16 character string.

josevalim:
Yes, we support both binary (low-level) and string (high-level) UUIDs when casting. If you want to restrict, it can be done with a custom type. :)

josevalim:
To further clarify, a UUID is stored as a 16 byte binary, which can be anything, even 16 x in a row.

Issue #3125, opened 2019-09-26
Ecto seems to validate UUIDs that are definitely not valid UUIDs


dhaspden:
### Precheck

* Do not use the issues tracker for help or support requests (try Elixir Forum, Stack Overflow, IRC or mailing lists, etc).
* For proposing a new feature, please start a discussion on [elixir-ecto](https://groups.google.com/forum/#!forum/elixir-ecto).
* For bugs, do a quick search and make sure the bug has not yet been reported.
* Finally, be nice and have fun!

### Environment

* Elixir version (elixir -v): 1.8.1
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): N/A
* Ecto version (mix deps): 3.1.7
* Database adapter and version (mix deps): N/A
* Operating system: OSX 10.14.6

### Current behavior

I was writing some tests for an `embedded_schema` that has a field which is a UUID. I wrote a test just to make sure the UUID validation was returning as expected and couldn't figure out for the life of me why it wasn't working.

```
iex(1)> Ecto.UUID.cast("not a valid uuid")
{:ok, "6e6f7420-6120-7661-6c69-642075756964"}
iex(2)> Ecto.UUID.cast("definitely not a valid uuid")
:error
iex(3)> Ecto.UUID.cast("legit not a uuid")
{:ok, "6e6f7420-6120-7661-6c69-642075756964"}
```

I'm relatively new to Elixir and this library but this doesn't seem right. This is definitely not a valid UUID. Maybe I don't understand the core function of `cast` in this case but it's definitely strange. Also interesting that the first and last example cast to the same UUID.

### Expected behavior

I'd imagine it should play out like this:

```
iex(1)> Ecto.UUID.cast("not a valid uuid")
:error
iex(2)> Ecto.UUID.cast("definitely not a valid uuid")
:error
iex(3)> Ecto.UUID.cast("legit not a uuid")
:error
```

josevalim:
Cast allows either UUIDs in the binary 16 byte format or in the UUID human representation format. In this sense, any 16 byte long string that you pass, is a valid binary UUID which is converted to its human representation.

[...]

PR #4240, opened 2023-07-21
docs: add clarification and examples to `Ecto.UUID.cast/1`

lytedev:
This highlights the fact that it can take a 16-byte binary (naturally, even in the form of a string) which may be unexpected to users. It also adds some recommendations for `Ecto.UUID.load/1` as a possible alternative for some uses.

See the following issues for more context and details:

- https://github.com/elixir-ecto/ecto/issues/3125
- https://github.com/elixir-ecto/ecto/issues/4174
- https://github.com/elixir-ecto/ecto/issues/4224

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -20,8 +20,9 @@ defmodule Ecto.UUID do
 
   @doc """
   Casts either a string in the canonical, human-readable UUID format or a
-  16-byte binary to a UUID in its canonical, human-readable UUID format. If
-  `uuid` is neither of these, `:error` will be returned.
+  16-byte binary to a UUID in its canonical, human-readable UUID format.
+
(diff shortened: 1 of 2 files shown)


lib/ecto/uuid.ex
```diff
@@ -19,9 +19,33 @@ defmodule Ecto.UUID do
   def type, do: :uuid
 
   @doc """
-  Casts to a UUID.
+  Casts either a string in the canonical, human-readable UUID format or a
+  16-byte binary to a UUID in its canonical, human-readable UUID format. If
+  `uuid` is neither of these, `:error` will be returned.
```
josevalim, line 24:
```suggestion
  16-byte binary to a UUID in its canonical, human-readable UUID format.

  If `uuid` is neither of these, `:error` will be returned.
```

josevalim:

PR #4243, opened 2023-07-24
Expand Multi transaction error message

tfwright:
Opening a draft to serve as a starting point for discussion. I have run into this error a few times and I've never been able to decipher exactly what it means. It usually appears to result from invalid error handling inside a Multi. Usually that has been unintentional but in a recent case I ran into when handling a nested multi transaction that failed as a success in the parent multi. I've added that case as an example in the docs to point future wayward souls in the right direction. Please advise on accuracy and clarity, although I do think clearly _something_ should be added to this error because it doesn't seem very helpful/instructive currently.

### DIFF
--- a/lib/ecto/repo/transaction.ex
+++ b/lib/ecto/repo/transaction.ex
@@ -16,9 +16,19 @@ defmodule Ecto.Repo.Transaction do
     return = &adapter.rollback(adapter_meta, &1)
 
     case Ecto.Multi.__apply__(multi, repo, wrap, return) do
-      {:ok, values} -> {:ok, values}
-      {:error, {key, error_value, values}} -> {:error, key, error_value, values}
-      {:error, operation} -> raise "operation #{inspect operation} is manually rolling back, which is not supported by Ecto.Multi"
+      {:ok, values} -> 
+        {:ok, values}
+
+      {:error, {key, error_value, values}} -> 
+        {:error, key, error_value, values}
+
+      {:error, operation} -> 
+        raise """
+          Operation #{inspect operation} is manually rolling back, which is not supported by Ecto.Multi.
+
+          This can occur if you try to commit an outer transaction when the inner transaction has failed 
+          (and rolled back). Instead, allow the outer transaction to fail as well and handle the error case.
+        """
     end
   end
 
--- a/lib/ecto/repo/transaction.ex
+++ b/lib/ecto/repo/transaction.ex
@@ -24,10 +24,10 @@ defmodule Ecto.Repo.Transaction do
 
       {:error, operation} -> 
         raise """
-          Operation #{inspect operation} is manually rolling back, which is not supported by Ecto.Multi.
+        operation #{inspect operation} is manually rolling back, which is not supported by Ecto.Multi.
 
-          This can occur if you try to commit an outer transaction when the inner transaction has failed 
-          (and rolled back). Instead, allow the outer transaction to fail as well and handle the error case.
+        This can occur if you try to commit an outer transaction when the inner transaction has failed 
+        (and rolled back). Instead, allow the outer transaction to fail as well and handle the error case.
         """
     end
   end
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
Hi @tfwright,

The gist of your message looks right, that anything causing the multi to rollback will produce that error. I'm having a hard time picturing your case though. Would you be able to provide a small example of the nested transaction issue you faced?

tfwright:
A pseudo-code version would look like this:

```
def outer_multi do
  Multi.new()
  |> Multi.insert(:something, ...)
  |> Multi.run(:inner_multi, fn _, _ -> 
    inner_multi()
    :ok
  end)
  |> Repo.transaction()
end

def inner_multi do
  Multi.new()
  |> Multi.insert(:something_else, ...)
  |> Repo.transaction()
end
```

greg-rychlewski:
How did you manage to fix it? By flattening the inner transaction into the main transaction?

tfwright:
In this particular case I didn't really need the outer multi because it was fine to just crash the process if what was the first operation failed

lib/ecto/repo/transaction.ex
```diff
@@ -16,9 +16,19 @@ defmodule Ecto.Repo.Transaction do
...
+      {:error, {key, error_value, values}} -> 
+        {:error, key, error_value, values}
+
+      {:error, operation} -> 
+        raise """
+          Operation #{inspect operation} is manually rolling back, which is not supported by Ecto.Multi.
+
+          This can occur if you try to commit an outer transaction when the inner transaction has failed 
+          (and rolled back). Instead, allow the outer transaction to fail as well and handle the error case.
+        """
```
v0idpwn, line 31:
```suggestion
        raise """
        operation #{inspect operation} is manually rolling back, which is not supported by Ecto.Multi.

        This can occur if you try to commit an outer transaction when the inner transaction has failed 
        (and rolled back). Instead, allow the outer transaction to fail as well and handle the error case.
        """
```

Exceptions start with lowercase, docstrings don't need the extra tab.

v0idpwn (commented):
I like the idea, but the suggestion is a bit confusing to me, maybe it could be worded differently to be more clear.

tfwright:
@v0idpwn sure, I am not set on the wording. If anyone has a suggestion for clarifications/additions I'm happy to make those. I think just about anything extra here would be more helpful than the existing (providing it's accurate of course)

greg-rychlewski:
Another thing about the message is that I'm not sure it makes sense to have a transaction nested inside a Multi. It doesn't look to me like you can utilize savepoints so it's essentially the same thing as flattening it out only with more difficult error handling.

lib/ecto/repo/transaction.ex
```diff
@@ -16,9 +16,19 @@ defmodule Ecto.Repo.Transaction do
...
+      {:error, {key, error_value, values}} -> 
+        {:error, key, error_value, values}
+
+      {:error, operation} -> 
+        raise """
+        operation #{inspect operation} is manually rolling back, which is not supported by Ecto.Multi.
+
+        This can occur if you try to commit an outer transaction when the inner transaction has failed 
+        (and rolled back). Instead, allow the outer transaction to fail as well and handle the error case.
+        """
```
greg-rychlewski, line 31:
```suggestion
      {:error, operation} -> 
        raise """
        operation #{inspect operation} is rolling back unexpectedly . 
        
        This can happen if `repo.rollback/1` is manually called, which is not supported by `Ecto.Multi`. It can also occur if a nested transaction has rolled back and its error is not bubbled up to the outer multi. Nested transactions are discouraged when using `Ecto.Multi`. Consider flattening out the transaction instead. 
        """
```

@v0idpwn / @josevalim  Wanted to see if I could close this one. How does this look to you?

v0idpwn: This looks good to me

josevalim: Looks good to me. But please break it into multiple lines and have `\` at the end of each line so we remove the line breaks from when the exception is effectively shown. :) Thanks :heart:

greg-rychlewski:
Thank you @tfwright

PR #4244, opened 2023-07-25
Document subquery can be used in select

Coffei:
Since 3.9.0 `subquery` can be used directly in `select` and `select_merge`. This mentions it in the documentation so it's clear it is a supported construct.


josevalim:

Issue #4247, opened 2023-07-25
Atoms in select cause problems in unions


Coffei:
### Elixir version

1.14.3

### Database and Version

Postgresql 15

### Ecto Versions

3.10.1 and master

### Database Adapter and Versions (postgrex, myxql, etc)

0.16.2

### Current behavior

I have problem with this query
```elixir
Repo.all(
  union_all(
    from(a in "asset",
      where: a.id == "338793453154",
      select: %{id: a.id, x: :xxx}
    ),
    ^from(a in "asset",
      where: a.id == "338079210336",
      select: %{id: a.id, x: :yyy}
    )
  )
)
```
Both queries inside union return a single row but the result looks like this

```elixir
[%{id: "338793453154", x: :xxx}, %{id: "338079210336", x: :xxx}]
```

Even though the second row comes from the second query with `x: :yyy` there is `x: :xxx` in the result. This seems to be caused by the fact that atoms in select are not part of the query and are added to the result later (and Ecto probably can't tell which query the row came from in a union?).

The SQL query looks like 

```sql
SELECT a0."id" FROM "asset" ... UNION ALL (SELECT a0."id" FROM "asset"  ...)
```

If I replace the atoms with strings it works, because strings are added to the query

```sql
SELECT a0."id", 'xxx' FROM "asset" ... UNION ALL (SELECT a0."id", 'yyy' FROM "asset" ...)
```

This can even cause a crash if you combine string, because the each query in the union has different number of columns. So query

```elixir
Repo.all(
  union_all(
    from(a in "asset",
      where: a.id == "338793453154",
      select: %{id: a.id, x: "xxx"}
    ),
    ^from(a in "asset",
      where: a.id == "338079210336",
      select: %{id: a.id, x: :yyy}
    )
  )
)
```

ends with error

```
(Postgrex.Error) ERROR 42601 (syntax_error) each UNION query must have the same number of columns
```

Not sure how much using atoms in a select is supported or not, I didn't see anything in the docs suggesting either. But to me this seems like something that should work.

### Expected behavior

Ideally atoms in select should just work. If they are not supported it could be mentioned in the docs or outright forbidden.

PR #4248, opened 2023-07-26
Add warning for selecting literal atoms from combinations

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4247

I went the documentation route because:

1. I don't believe there is any way to make it work. The literal atoms are part of the post processor and we have no idea how many rows will be returned from each query in the combination.
2. It will be pretty complicated to detect literal atoms in select expressions in combinations and raise if they are different. I think this is an uncommon enough case that it's not worth the extra complexity.

Technically only rows from the parent query are returned from except(_all) and intersect(_all) but I thought I'd add the warning to them just in case people expect the results to honour the literal atoms.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1845,7 +1845,10 @@ defmodule Ecto.Query do
   A union query expression.
 
   Combines result sets of multiple queries. The `select` of each query
-  must be exactly the same, with the same types in the same order.
+  must be exactly the same, with the same types in the same order. When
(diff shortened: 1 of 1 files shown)


Coffei:
Thanks! I though this was not going to be easy to fix.

If I can, I'd suggest adding a word `in` or `across` to the first sentence, seems incomplete without it, like so: _its value must be the same **in** all queries._

josevalim (approved):

PR #4249, opened 2023-07-27
Improve  Ecto.Repo config error messge

fceruti:
The issue is really simple: if you set anything other than `false | Logger.level/0`, you get a really cryptic message.

This error was already discussed in [this link](https://bytemeta.vip/index.php/repo/elixir-ecto/ecto_sql/issues/351), and since both Jose asked for PRs and the issue was still active, here I am.

This is my first dive into Ecto, so I'm not quite sure if I'm doing things correctly, or in the right place, so any guidance to land this PR would be welcomed.

### DIFF
--- a/lib/ecto/repo/supervisor.ex
+++ b/lib/ecto/repo/supervisor.ex
@@ -52,6 +52,7 @@ defmodule Ecto.Repo.Supervisor do
   def compile_config(_repo, opts) do
     otp_app = Keyword.fetch!(opts, :otp_app)
     adapter = opts[:adapter]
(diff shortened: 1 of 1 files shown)


greg-rychlewski:
Hi,

Thank you for opening the PR. It was actually addressed recently in the `ecto_sql` project: https://github.com/elixir-ecto/ecto_sql/pull/519/files.

The reason it went into that project instead of Ecto is because that is the only place the log level takes effect. Essentially just to log out the SQL command.

fceruti:
Thanks for the response, I'm glad it got solved!

PR #4252, opened 2023-07-28
Add Repo.update_all/1

v0idpwn:
Convenience function to be used when query already has updates.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -406,7 +406,7 @@ defmodule Ecto.Repo do
 
       if Ecto.Adapter.Queryable in behaviours do
         if not @read_only do
-          def update_all(queryable, updates, opts \\ []) do
+          def update_all(queryable, updates \\ [], opts \\ []) do
             repo = get_dynamic_repo()
 
             Ecto.Repo.Queryable.update_all(
@@ -1349,13 +1349,13 @@ defmodule Ecto.Repo do
       |> MyRepo.update_all([])
 
       from(p in Post, where: p.id < 10, update: [set: [title: ^new_title]])
-      |> MyRepo.update_all([])
+      |> MyRepo.update_all()
 
       from(p in Post, where: p.id < 10, update: [set: [title: fragment("upper(?)", ^new_title)]])
-      |> MyRepo.update_all([])
+      |> MyRepo.update_all()
 
       from(p in Post, where: p.id < 10, update: [set: [visits: p.visits * 1000]])
-      |> MyRepo.update_all([])
+      |> MyRepo.update_all([], [timeout: 5_000])
 
   """
   @doc group: "Query API"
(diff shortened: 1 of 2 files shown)


v0idpwn:
The only problem I see is that its not great for functions to have more than 1 optional parameter, but considering how common it is to have the update statement in the query and writing a useless `[]`, this seemed reasonable. 

Not sure if it should be merged, though :)

josevalim:
Thoughts @greg-rychlewski?

greg-rychlewski:
I have been caught by this a few times (forgetting the empty list). But in general I'm not a big fan of having multiple optional params. I saw them lead to too much confusion and a lot of the times you end up having to put the middle values anyways.

So I am a soft -1. I don't really recommend it but if enough people think this is a big enough issue I wouldn't complain about it being added.

josevalim:
It is your decision then folks. I am also ok with adding it or not. :)

greg-rychlewski (approved):

greg-rychlewski:
@v0idpwn if you find it useful, go for it. the number of parameters is very small so the multiple defaults is not the end of the world in this case, imo

v0idpwn:
I'm also not 100% sure, I got this idea after some feedback regarding Repo.update_all/3 API.

I will leave this open for a while, maybe we get more feedback from users.

greg-rychlewski:
Ah, got it. For the sake of discussion then, one other thing that is nice about forcing the update to be specified is the intention is more explicit. When reviewing someone else’s code you have a bit more confidence they didn’t accidentally forget about specifying the updates. 

With this change they could throw a bad query in here quickly and not get any feedback. But with the current setup they will at least have to think about why the compiler is complaining about a missing argument.

josevalim:
I guess we should raise if we go in this direction and there no updates in the query either?

greg-rychlewski:
Yeah +1 for that

v0idpwn:
It already raises a nice error about the lack of updates, I think its maybe even friendlier than the compiler complaining about the lack of the second parameter? Of course it happens in runtime, tho.

greg-rychlewski:
yeah agreed with everything you said. my comment was more about catching it before it's deployed and having the intent be more explicit when reading the code.

wojtekmach:
+1 from me but if you are unsure if it’s worth to add it, don’t add it, totally up to you. :)

wojtekmach:
Somewhat similar, we need to do `Postgrex.query("select * from posts", [])`. It would be nice not to have to specify empty list of params. I sneaked in such change so we can simply do `MyXQL.query("select * from posts")` and nobody complained so far. ;)

greg-rychlewski:
I'm just playing devil's advocate, but there is a bit of a difference between ^ and the updates.

The user specified updates will override the ones in the query. So forgetting to specify it means the wrong update could possibly be used. In the query case you just get an error from the database about missing parameters.

I don't know how often people are sending queries into `update_all` with updates that shouldn't happen. But in general it seems like re-using queries with slight modifications is enabled/encouraged. We have those functions that remove X key from the query.

greg-rychlewski:
I think my vote is actually -1 now, given the potential for weird issues I said above. I think the function is too flexible for it to be worth it to remove the explicitness.

v0idpwn:
I will be closing this based on feedback, at least for now.

Issue #4253, opened 2023-07-30
Deprecate literal and kw hints in favor of unsafe_literal


josevalim:
### Elixir version

All

### Database and Version

All

### Ecto Versions

3.8+

### Database Adapter and Versions (postgrex, myxql, etc)

All

### Current behavior

Currently we have both `literal` and kw-values in hints which can be potentially unsafe.

### Expected behavior

We must deprecate `literal` in favor of `unsafe_literal`. Same behaviour, different name. Similarly, for hints, instead of KW we should allow `["SAMPLE", unsafe_literal(...), ...]`.

Related commits:

* https://github.com/elixir-ecto/ecto/commit/6e7bc3f4d757b7c09ced10135e0c5c4ce1f4ea2f
* https://github.com/elixir-ecto/ecto/pull/3592

@greg-rychlewski let me know if you want to take this one, otherwise I shall do it. :)

greg-rychlewski:
I could take it. For `literal` though I think it's safe because it's only being used for identifiers so they are always quoted. The hint/sample stuff is a bit different because it can alter the actual command.

At least this is what it looks like to me. LMK if I'm getting confused.

greg-rychlewski:
Should the hint stuff be `:unsafe_fragment` instead. Similar to this:

```elixir
* `:conflict_target` - A list of column names to verify for conflicts.
      It is expected those columns to have unique indexes on them that may conflict.
      If none is specified, the conflict target is left up to the database.
      It may also be `{:unsafe_fragment, binary_fragment}` to pass any
      expression to the database without any sanitization, this is useful
      for partial index or index with expressions, such as
      `{:unsafe_fragment, "(coalesce(firstname, ""), coalesce(lastname, "")) WHERE middlename IS NULL"}` for
      `ON CONFLICT (coalesce(firstname, ""), coalesce(lastname, "")) WHERE middlename IS NULL` SQL query.
```

josevalim:
A list of strings or an unsafe fragment? Sounds good to me!

PR #4254, opened 2023-07-31
Use unsafe_fragment for dynamic hints.

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4253

I wavered on the `:sample` idea. We would have to duplicate what we are doing here because sampling needs `unsafe_fragment` (the sample method is part of the sql and it can be anything, people can make extensions). Plus there are other things that can come after `from`. So I just tried to improve the docs to make it clear `hints` is the mechanism for anything coming after `from`.

I will update this with the ecto_sql PR

edit: compainion ecto_sql PR https://github.com/elixir-ecto/ecto_sql/pull/546

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -975,7 +975,7 @@ defmodule Ecto.Query do
   >
   > The output of `unsafe_fragment/1` will be injected directly into the
   > SQL statement without being escaped. For this reason, user input should
-  > **never** be used.
+  > **never** be used because it can lead to SQL injections.
 
   ## Keywords examples
 
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -974,8 +974,9 @@ defmodule Ecto.Query do
   > ### Unsafe Fragments {: .warning}
   >
   > The output of `unsafe_fragment/1` will be injected directly into the
-  > SQL statement without being escaped. For this reason, user input should
-  > **never** be used because it can lead to SQL injections.
+  > resulting SQL statement without being escaped. For this reason, input
+  > from uncontrolled sources, such as user input, should **never** be used.
+  > Otherwise, it could lead to harmful SQL injection attacks.
 
   ## Keywords examples
 
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1311,7 +1311,7 @@ defmodule Ecto.Query do
 
   ## Hints
 
-  `join` also supports index hints, as found in databases such as
+  `join` also supports table hints, as found in databases such as
   [MySQL](https://dev.mysql.com/doc/refman/8.0/en/index-hints.html),
   [MSSQL](https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-2017) and
   [Clickhouse](https://clickhouse.tech/docs/en/sql-reference/statements/select/sample/).
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -948,6 +948,35 @@ defmodule Ecto.Query do
   a value that implements the `Ecto.Queryable` protocol
   and the second argument the expression.
 
+  ## Hints
+
+  The `hints` keyword can be used to specify query hints:
+
+      from p in Post,
+        hints: ["USE INDEX FOO"],
+        where: p.title == "title"
+
+  They can also be used as general mechanism for adding statements that
+  come after the `from` clause. For example, they can be used to enabled
+  table sampling:
+
+      from p in Post,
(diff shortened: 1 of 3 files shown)


lib/ecto/query/builder/from.ex
```diff
@@ -145,6 +140,37 @@ defmodule Ecto.Query.Builder.From do
...
+          "`hints` must be a compile time string, an unsafe fragment of the form `unsafe_fragment(^...)`, " <>
+            "or a list containing either, got: `#{Macro.to_string(other)}`"
+        )
+    end
+  end
+
+  def hint!(other) do
+    Builder.error!(
+      "`hints` must be a compile time string, an unsafe fragment of the form `unsafe_fragment(^...)`, " <>
+        "or a list containing either, got: `#{Macro.to_string(other)}`"
```
v0idpwn, line 170:
```suggestion
        "or a compile time list containing either, got: `#{Macro.to_string(other)}`"
```

lib/ecto/query/builder/from.ex
```diff
@@ -145,6 +140,37 @@ defmodule Ecto.Query.Builder.From do
...
+          {:unsafe_fragment, checked}
+        end
+
+      hint when is_binary(hint) ->
+        hint
+
+      other ->
+        Builder.error!(
+          "`hints` must be a compile time string, an unsafe fragment of the form `unsafe_fragment(^...)`, " <>
+            "or a list containing either, got: `#{Macro.to_string(other)}`"
```
v0idpwn, line 162:
```suggestion
            "or a compile time list containing either, got: `#{Macro.to_string(other)}`"
```

v0idpwn (approved):
Both code and concept look great to me.

lib/ecto/query.ex
```diff
@@ -1281,7 +1311,7 @@ defmodule Ecto.Query do
 
   ## Hints
 
-  `from` and `join` also support index hints, as found in databases such as
+  `join` also supports index hints, as found in databases such as
```
josevalim, line 1314:
```suggestion
  `join` also supports table hints, as found in databases such as
```

lib/ecto/query/builder/from.ex
josevalim: I think the initial "compile time" is meant to qualify them all. :)

josevalim (approved):

lib/ecto/query/builder/from.ex
```diff
@@ -145,6 +140,37 @@ defmodule Ecto.Query.Builder.From do
...
+  Validates hints at compile time and runtime
+  """
+  def hint!(hint) when is_binary(hint), do: hint
+
+  def hint!({:unsafe_fragment, _, [fragment]}) do
+    case fragment do
+      {:^, _, [value]} ->
+        quote do
+          checked = Ecto.Query.Builder.From.hint!(unquote(value))
+          {:unsafe_fragment, checked}
```
josevalim, line 153: I think we can actually unpack it here:

```suggestion
          Ecto.Query.Builder.From.hint!(unquote(value))
```

This way we don't need to change the adapters. WDYT?

lib/ecto/query/builder/from.ex
v0idpwn: Then the `an` and `a` need to be changed

lib/ecto/query/builder/from.ex
v0idpwn: Didn't realise adapter would need changes. If yes, it would be relevant to notify people from ecto_ch as they use the hints feature.

greg-rychlewski: Yeah at first I was thinking to keep it for inspect but maybe it is not worth it. Do you think it’s worth it for inspect?

greg-rychlewski: The clickhouse people need to be notified either way though because the kw version is deprecated now

lib/ecto/query/builder/from.ex
greg-rychlewski: Would it read better if it was like this:

> a (list of) compile-time strings or unsafe fragments of the form `unsafe_fragment(^...)`, got: ...

lib/ecto/query/builder/from.ex
josevalim: Yeah, i don't think it is worth it. :)

lib/ecto/query/builder/from.ex
greg-rychlewski: actually nm I get what you're saying now. fixed.

v0idpwn:
@ruslandoga kind of breaking changes for hints

Issue #4262, opened 2023-08-14
Aggregation type load differs from normal queries


Wigny:
### Elixir version

1.15.4

### Database and Version

MySQL 5.7

### Ecto Versions

3.10.3

### Database Adapter and Versions (postgrex, myxql, etc)

myxql 0.6.3

### Current behavior

When I created a custom `Ecto.ParameterizedType` for handling ISO 8601 Durations with Timex, I defined it to be of type `:float`, later when I tried to apply this custom type on an already existent schema field of type `:decimal`, I noticed that loading and dumping into the DB worked fine in normal queries, but it fails when aggregating.

```elixir
Mix.install([
  {:ecto_sql, "~> 3.10"},
  {:myxql, ">= 0.0.0"},
  {:timex, "~> 3.7"}
])

Application.put_env(:my_app, MyApp.Repo,
  username: "root",
  password: "root",
  hostname: "localhost",
  database: "my_app_dev"
)

defmodule MyApp.Repo do
  use Ecto.Repo,
    adapter: Ecto.Adapters.MyXQL,
    otp_app: :my_app
end

defmodule MyApp.Repo.Migrations.CreateTravels do
  use Ecto.Migration

  def change do
    create table("travels") do
      add(:time, :decimal)
    end
  end
end

defmodule MyApp.Type.Duration do
  use Ecto.ParameterizedType

  alias Timex.Duration

  defguardp is_duration(value) when is_struct(value, Duration)

  @impl true
  def type(_params), do: :float

  @impl true
  def init(opts), do: %{unit: opts[:unit]}

  @impl true
  def cast(value, _params) when is_nil(value), do: {:ok, nil}
  def cast(value, _params) when is_duration(value), do: {:ok, value}
  def cast(value, params) when is_float(value), do: {:ok, from_float(value, params.unit)}

  @impl true
  def load(value, _, _params) when is_nil(value), do: {:ok, nil}
  def load(value, _, params) when is_float(value), do: {:ok, from_float(value, params.unit)}

  @impl true
  def dump(value, _, _params) when is_nil(value), do: {:ok, nil}
  def dump(value, _, params) when is_duration(value), do: {:ok, to_float(value, params.unit)}

  defp from_float(value, :microseconds), do: Duration.from_microseconds(value)
  defp from_float(value, :milliseconds), do: Duration.from_milliseconds(value)
  defp from_float(value, :seconds), do: Duration.from_seconds(value)
[...]

PR #4265, opened 2023-08-19
change {:maybe, type} to {:aggregate, type}

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4262

Companion Ecto SQL PR: https://github.com/elixir-ecto/ecto_sql/pull/550

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -413,6 +413,7 @@ defmodule Ecto.Type do
   def type({:parameterized, type, params}), do: type.type(params)
   def type({:array, type}), do: {:array, type(type)}
   def type({:map, type}), do: {:map, type(type)}
+  def type({:maybe, type}), do: type
   def type(type) when type in @base, do: type
   def type(type) when is_atom(type), do: type.type()
   def type(type), do: type


josevalim (approved):

lib/ecto/type.ex
```diff
@@ -602,7 +602,7 @@ defmodule Ecto.Type do
     {:ok, nil}
   end
 
-  def load({:maybe, type}, value, loader) do
+  def load({:aggregate, type}, value, loader) do
```
josevalim, line 605: There is one thing I don’t understand. If aggregate float falls back to float, why is it not loading as expected?

greg-rychlewski: Oh so what I found is the following happens for aggregates ([reference](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/myxql.ex#L152C2-L157C35))

1. `Ecto.Adapters.MyXQL.loaders/2` is called with the following arguments `{:maybe, :float}` and `{:maybe, :float}`. That causes the return value to be `[{:maybe, :float}]`
2. The result from above eventually gets passed to [Ecto](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/type.ex#L605). Then Ecto tries to convert the decimal to float but it's not allowed. So it returns the decimal. 

Now with this change instead of (1) returning `[{:maybe, :float}]` it returns `[&float_decode/1, {:maybe, :float}]`. So first `float_decode/1` converts the decimal into a float. And then the `{:maybe, :float}` part will essentially be the identity function. 

I'm guessing the decimal -> float conversion is not allowed in Ecto itself because for most adapters it will be an unsafe downcast. It seems like MySQL allows it because [the precision is limited in that database](https://dev.mysql.com/doc/refman/8.0/en/precision-math-decimal-characteristics.html) which I'm guessing fits nicely into Elixir floats.

josevalim: I am surprised we are not falling back to calling loaders(:float) in the adapter. Perhaps that’s the change we need to make.

josevalim: I think if we add `def type({:aggregate, sub}), do: sub`, then it will work. as it will force us to pass the real type to adapters. But I am not sure what else we break. if that's the case, then perhaps I would even keep it as `:maybe`. Otherwise this PR is the best way to go indeed.

greg-rychlewski: > I think if we add def type({:aggregate, sub}), do: sub, then it will work. as it will force us to pass the real type to adapters. But I am not sure what else we break.

Going this route makes things a little confusing to me at least. But let me know what you think. When we get to this part:

```elixir
process_loaders(adapter.loaders(type(type), type), {:ok, value}, adapter)
```

We want `adapters.loaders` to be unaware of the `maybe`. But then we want `process_loaders` to be aware of it. So there is a conflict between the return value of `adapter.loaders` and what we want to happen after.

greg-rychlewski: Actually sorry I see what you are saying. Maybe this works I will play around with it.

greg-rychlewski: It seems like it's not breaking anything. Pushing the change to let the integration tests run.

greg-rychlewski: The integration tests pass. When looking through the code I can't see any other place that would break even if tests are missing. `{:maybe, type}` is only set in the planner during `collect_fields` and it's private otherwise. So besides loading/dumping aggregations from the database I don't believe there is any other hidden landmine.

It looks to me like you found the best change.

josevalim: Apologies for not seeing it sooner. :(

josevalim (approved):

lib/ecto/type.ex
josevalim: I only realized it when review the PR :)

greg-rychlewski: No worries! Sorry I didn't see it either.

PR #4266, opened 2023-08-19
Update changelog

greg-rychlewski:
As an aside:

I'm not too sure what the plans are for releasing 3.11. But I was wondering if it could wait a little bit for a couple things:

1. To see if I can come up with a decent proposal for the derived fields
2. I believe I thought of a simple way to support `values`. The other PR has stalled for over a year so I thought maybe I could submit something if my idea works out.


CHANGELOG.md
```diff
@@ -2,9 +2,33 @@
 
 ## v3.11.0-dev
 
+### Enhancements
+
+  * [Ecto.Association] Allow `preload_order` to take MFAs for `many_to_many` associations. This allows ordering by the join table
+  * [Ecto.Query] Add `:operation` option to `with_cte/3`. This allows CTEs to perform updates and deletes
+  * [Ecto.Query] Support `splice(^...)` in `fragment`
+  * [Ecto.Query] Add `:mode` option to `order_by/4`
```
josevalim, line 10: I had to consult the docs to remember what `:mode` does. It makes me think it may not be the best option name. Maybe we should instead have `prepend_order_by`? It would be closer to `or_where` and `or_having` in this case. WDYT?

josevalim:
@greg-rychlewski I don't think there is any pressure for releasing, so take your time. @whatyouhide may want a release, but they are still working on improving the adapter.

Also, I would be very glad to give you owner access on hex.pm if you want to publish the next releases. Just let me know your hex.pm email (you can email it to me). :)

In terms of features, the feature I see most asked is automatic migrations support, as in Django: https://docs.djangoproject.com/en/4.2/topics/migrations/#workflow - so if you are interested in bringing the projects forward, it may be something worth working on. It would be added to EctoSQL and it would require at least two new callbacks in SQL adapters: one for listing all fields for a table and another that checks the field compatibility (the ecto field with the db field).

CHANGELOG.md
greg-rychlewski: agreed! sent a PR

greg-rychlewski:
> Also, I would be very glad to give you owner access on hex.pm if you want to publish the next releases. Just let me know your hex.pm email (you can email it to me). :)

Oh thanks very much. I will send you an email. 

> In terms of features, the feature I see most asked is automatic migrations support, as in Django: https://docs.djangoproject.com/en/4.2/topics/migrations/#workflow - so if you are interested in bringing the projects forward, it may be something worth working on. It would be added to EctoSQL and it would require at least two new callbacks in SQL adapters: one for listing all fields for a table and another that checks the field compatibility (the ecto field with the db field).

This looks interesting I will give it a shot!

josevalim (approved):

PR #4268, opened 2023-08-20
Remove unnecessary constraint in self-referencing many to many guide

greg-rychlewski:
I saw this guide linked on Elixir Forum. When I read through it I noticed that there is an additional unique constraint that is not needed. If the constraint is on the pair then it doesn't matter what order the pair is presented in.

### DIFF
--- a/guides/howtos/Self-referencing many to many.md
+++ b/guides/howtos/Self-referencing many to many.md
@@ -9,7 +9,7 @@ Let's imagine we are building a system that supports a model for relationships b
 ```elixir
 defmodule MyApp.Accounts.Person do
   use Ecto.Schema
-  
+
   alias MyApp.Accounts.Person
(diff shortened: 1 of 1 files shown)


josevalim:
> If the constraint is on the pair then it doesn't matter what order the pair is presented in.

This may not be true on some databases (MySQL). So perhaps we should keep it commented out with a note?

greg-rychlewski:
> This may not be true on some databases (MySQL).

Ah sorry about that. I learned something new today. Just want to make sure I understand: is it because MySQL returns something like "duplicate entries for keys a,b" and so the exact constraint name can't be matched on?

josevalim:
I was thinking more from the performance side of things but I guess we are over-covered here. We already have indexes on individual columns, so I think we are good.

josevalim (approved):

PR #4270, opened 2023-08-25
Support values lists

greg-rychlewski:
Continues the work started here: https://github.com/elixir-ecto/ecto/pull/3929

The approach here is a bit different. Instead of making a macro I'm treating it like `fragment` where we just pattern match on the AST. 

Will update with the ecto sql PR shortly. The integration tests will fail here but pass there.

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -503,6 +503,11 @@ defmodule Ecto.Query.API do
   def splice(list), do: doc! [list]
 
   @doc """
+
+  """
+  def values(values, types), do: doc! [values, types]
+
+  @doc """
   Allows a field to be dynamically accessed.
 
       def at_least_four(doors_or_tires) do
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -518,7 +518,7 @@ defmodule Ecto.Query.API do
   ## Select example
 
       values = [%{id: 1, text: "abc"}, %{id: 2, text: "xyz"}]
-      types = %{id: :integer}
+      types = %{id: :integer, text: :string}
 
       query =
         from v1 in values(values, types),
@@ -529,7 +529,7 @@ defmodule Ecto.Query.API do
 
   ## Delete example
       values = [%{id: 1, text: "abc"}, %{id: 2, text: "xyz"}]
-      types = %{id: :integer}
+      types = %{id: :integer, text: :string}
 
       query =
         from p in Post,
@@ -541,7 +541,7 @@ defmodule Ecto.Query.API do
 
   ## Update example
       values = [%{id: 1, text: "abc"}, %{id: 2, text: "xyz"}]
-      types = %{id: :integer}
+      types = %{id: :integer, text: :string}
 
       query =
         from p in Post,
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -503,7 +503,53 @@ defmodule Ecto.Query.API do
   def splice(list), do: doc! [list]
 
   @doc """
+  Creates a values list/constant table.
 
+  A values list can be used as a source in a query, both in `Ecto.Query.from/2`
+  and `Ecto.Query.join/5`.
+
+  The first argument is a list of maps representing the values of the constant table.
+  Each entry in the list must have exactly the same fields or an error is raised.
+
+  The second argument is a map of types corresponding to the fields in the first argument.
+  Each field must be given a type or an error is raised. Any type that can be specified in
+  a schema may be used.
+
+  ## Select example
+
+      values = [%{id: 1, text: "abc"}, %{id: 2, text: "xyz"}]
+      types = %{id: :integer}
+
+      query =
+        from v1 in values(values, types),
+          join: v2 in values(values, types),
+          on: v1.id == v2.id
+
+      Repo.all(query)
+
+  ## Delete example
+      values = [%{id: 1, text: "abc"}, %{id: 2, text: "xyz"}]
+      types = %{id: :integer}
+
+      query =
+        from p in Post,
+          join: v in values(values, types),
+          on: p.id == v.id,
+          where: p.counter == ^0
+
+      Repo.delete_all(query)
+
+  ## Update example
+      values = [%{id: 1, text: "abc"}, %{id: 2, text: "xyz"}]
+      types = %{id: :integer}
+
+      query =
+        from p in Post,
+          join: v in values(values, types),
+          on: p.id == v.id,
+          update: [set: [text: v.text]]
+
+      Repo.update_all(query, [])
   """
   def values(values, types), do: doc! [values, types]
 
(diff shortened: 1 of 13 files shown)


lib/ecto/query/builder.ex
```diff
@@ -791,6 +791,84 @@ defmodule Ecto.Query.Builder do
...
+  @spec values_list_fields(list(map)) :: map()
+  def values_list_fields(values_list) do
+    fields =
+      Enum.reduce(values_list, %{}, fn values, fields ->
+        Enum.reduce(values, fields, fn {field, _}, fields ->
+          Map.put(fields, field, true)
+        end)
+      end)
+
+    Map.keys(fields)
```
v0idpwn, line 831:
```suggestion
          Map.put(fields, field, true)
        end)
      end)

    Map.keys(fields)
```
This is a MapSet, so you could use it for simplicity

v0idpwn (commented):
Looks great.

lib/ecto/query/api.ex
```diff
@@ -502,6 +502,57 @@ defmodule Ecto.Query.API do
...
+  Each entry in the list must have exactly the same fields or an error is raised.
+
+  The second argument is a map of types corresponding to the fields in the first argument.
+  Each field must be given a type or an error is raised. Any type that can be specified in
+  a schema may be used.
+
+  ## Select example
+
+      values = [%{id: 1, text: "abc"}, %{id: 2, text: "xyz"}]
+      types = %{id: :integer}
```
josevalim, line 521: Don't we need to give a type for each field? Or do we infer in some cases?

lib/ecto/query/builder.ex
```diff
@@ -791,6 +791,84 @@ defmodule Ecto.Query.Builder do
...
+      quote do
+        Ecto.Query.Builder.values_list_types!(unquote(fields), unquote(types))
+      end
+
+    params =
+      quote do
+        Ecto.Query.Builder.values_list_params!(unquote(values_list), unquote(types))
+      end
+
+    num_rows = quote do: length(unquote(values_list))
```
josevalim, line 814: It is generally bad practice to unquote the same variable several times, because it means its expression will be executed several times. Perhaps we should encode only once, as parameter, and have all relevant information in the parameter (even if we have to create a struct to represent it).

lib/ecto/query/planner.ex
```diff
@@ -1156,6 +1159,12 @@ defmodule Ecto.Query.Planner do
     {fragments, acc} = prewalk(fragments, kind, query, expr, acc, adapter)
     {{:fragment, meta, fragments}, acc}
   end
+  defp prewalk_source({:values, meta, [types, num_rows]}, _kind, _query, _expr, acc, _adapter) do
```
josevalim, line 1162: Should we also dump the values in the list? For example, for UUIDs?

lib/ecto/query/builder.ex
josevalim: Although I see now that we need the num_rows must be part of the query cache. Perhaps the best option is to ensure on the from/join layer that the values become an `Ecto.Query.Values` struct and that the AST here only points to a variable. Something like this:

    from v in values([...])

becomes:

    internal_ecto_variable = Ecto.Query.Values.new([...])
    from v in values(internal_ecto_variable)

Then in here, you have something like:

    types = quote do: unquote(var).types
    num_rows = quote do: unquote(var).num_rows
    params = quote do: unquote(var).params ++ unquote(params)
    {{:{}, [], [:values, [], [types, num_rows]]}, params}

Although I think you may be able to write it as:

    params = [var | params]

And then in the planner expand the single "var" parameter into many parameters (I believe we do something similar for "in" already).

greg-rychlewski: > And then in the planner expand the single "var" parameter into many parameters (I believe we do something similar for "in" already).

I could try playing around with this. The reason I didn't do it initially is because cast/dump/load currently expect the `:in` param to be homogenous in terms of types. 

To make it work we would need to extend it to work with a list of types. And there is a bit of a tradeoff with the error reporting because it's for the entire param and not the individual components. So you'd get something like `can't cast [1, "1", ...] to type [:string, :int, ...]`. I can still play around with it but just wanted to see if you thought this was a non-starter before I do.

josevalim: I see. I am fine with expanding it into multiple parameters here as well. :)

lib/ecto/query/api.ex
greg-rychlewski: sorry that was an oversight. no inference in this PR

lib/ecto/query/builder.ex
greg-rychlewski: done! thank you

lib/ecto/query/builder.ex
greg-rychlewski: I gave it a try with what you said about the struct. The one part I'm unsure of is this:

```elixir
values_struct = quote do: Ecto.Query.Values.new(unquote(values_list), unquote(types))
{expr, params} = Builder.escape_values_list(values_struct)
```

I'm not too sure how to get it to work just passing in the ast for the `values_struct` variable instead of the whole quoted expression with the call to `Ecto.Query.Values.new`. I always get a compiler error `undefined function values_struct/0`.

I was also trying to think if I should be putting it somewhere else. But I think once I'm inside the from/join macros I'll have the same issue wherever I go. If you have any advice for this part I'd appreciate it. Thank you.

josevalim: @greg-rychlewski it needs to be done inside from/join at the same level you do the query. You have to define the Values.new variable before you build the query. Going back to the example, we want this:

    internal_ecto_variable = Ecto.Query.Values.new([...])
    from v in values(internal_ecto_variable)

In think in practice, you can do something like this:

1. Add a new element to From.escape to return `{query :: Macro.t(), prelude :: Macro.t(), params :: Keyword.t()}`

2. For everyone, `:prelude` will simply be `:ok`. For the `values` case, prelude will be `quote do: internal_ecto_variable = Ecto.Query.Values.new([...])`. In this case, you don't need to change `Builder.escape` at all, you can fully handle `values` inside `Builder.From.escape_source`. Something like this:

```elixir
  defp escape_source(query, env) do
    case Macro.expand_once(query, env) do
      {:values, _, [values, types]} = fragment->
        prelude = quote do: values = Ecto.Query.Values.new(unquote(values), unquote(types))
        types = quote do: values.types
        num_rows = quote do: values.num_rows
        params = quote do: values.params ++ unquote(params)
        {{:{}, [], [:values, [], [types, num_rows]]}, prelude, params}

      ... -> ...same code as today...
    end
```

3. Then, in `build`, you will do something like this:

```elixir
    {query, prelude, binds} = escape(query, env)

    query =
      case query do
        ... -> ...same code as today...
      end

    if prelude == :ok do
      query
    else
      quote do
        unquote(prelude)
        unquote(query)
      end
    end
```

I didn't check if this works at all, but that's what I had in mind!

greg-rychlewski: Ah I understand now. Thank you! WDYT?

greg-rychlewski: for the join one i'm not sure what's better:

1. extract the quoted join expression into a function so the prelude can be conditionally add it
2. keep it as is

josevalim (approved):
Looks great to me! 💯  If you want to keep the prelude as nil instead of :ok, that also works!

PR #4275, opened 2023-09-06
Allow validate_change to receive previous value

alexcastano:


lib/ecto/changeset.ex
```diff
@@ -2340,10 +2340,13 @@ defmodule Ecto.Changeset do
   `:errors` field of the changeset and the `:valid?` flag will be set to
   `false`.
 
+  Optionally, the validator function can take a third argument with the
+  previous value of the field.
```
greg-rychlewski, line 2344: What do you think about changing the first paragraph to something like this instead? We could probably do a better job describing the old behaviour as well:

```
Validates the given `field` change.

It invokes the `validator` function to perform the validation
only if a change for the given `field` exists and the change
value is not `nil`. If `validator` has an arity of 2, it is passed
`field` as the first argument and the change value as the
second argument. If it has an arity of 3, it is passed a third
argument corresponding to the current value of `field`. The
function must return a list of errors (with an empty list meaning no errors).
```

josevalim:
Hi @alexcastano, thanks for the PR!

What is the use case? I am honestly not sure we should merge it because having validations depending on the previous value is almost always not ideal. Plus, you can close over the changeset and access the changeset inside the function. So it seems we are adding complexity for a rare case that can be tackled in other ways.

lib/ecto/changeset.ex
alexcastano: My intention was to modify as little as possible, but it is much better this way. Thanks

alexcastano:
> Hi @alexcastano, thanks for the PR!
> 
> What is the use case? I am honestly not sure we should merge it because having validations depending on the previous value is almost always not ideal. Plus, you can close over the changeset and access the changeset inside the function. So it seems we are adding complexity for a rare case that can be tackled in other ways.

The use case is when you want to validate the change itself. For example, you don't want a field to be increased more than 10 in the changeset. I am also using it for state changes, you cannot change the status from "published" to "draft", but you can change from "draft"to "publish", or from "idea" to "draft". In both examples, you need both values, the previous one and the new one.

IMO this new clause makes more sense with the name of the function: "validate_change", now it is more like "validate_new_value" only. 

Anyway, I use this a lot, but maybe it is just me :) I made the PR because it would be nice to have it in Ecto directly. Happy to drop it if you think it doesn't make sense.

Thanks for your work!

josevalim:
I see. A state machine is a good example. Last question then: why not use `Map.fetch!(changeset.data, key)` to get the current value?

alexcastano:
> I see. A state machine is a good example. Last question then: why not use `Map.fetch!(changeset.data, key)` to get the current value?

Because of schemaless changesets:

```iex
iex(7)> params = %{first_name: "Jose"}
%{first_name: "Jose"}
iex(8)> types = %{first_name: :string}
%{first_name: :string}
iex(9)> {%{}, types} |> Ecto.Changeset.cast(params, Map.keys(types))
#Ecto.Changeset<
  action: nil,
  changes: %{first_name: "Jose"},
  errors: [],
  data: %{},
  valid?: true
>
iex(10)> |> Map.fetch!(:data) |> Map.fetch!(:first_name)
** (KeyError) key :first_name not found in: %{}
    (stdlib 4.2) :maps.get(:first_name, %{})
    iex:10: (file)
```

alexcastano:
Related, but maybe it should be discussed in another issue, when we don't execute the validator because the change to the `nil` value may be very inconvenient. I mean, if you want to detect the transition to `nil`, you have to create a new function because `validate_change` doesn't work for this case. You can say: "Well, use validate_required to validate it". However, you may allow the `nil` value in some circumstances, but you don't want to change to `nil` when the value is already set, for example.

I believe we cannot do anything about it because it would be a breaking change. This behaviour is documented from the very beginning, so I don't think it is a good idea to change it now. I just to let you know, maybe you have an idea.

josevalim:
I would strongly recommend setting the map for schemaless change sets with all keys. :) in any case, I don’t think we should pursue this for all. Thank you!

alexcastano:
Hi @josevalim , I thought you were happy with the state machine example and PR. The last comment is an aside. If you prefer not to merge, I understand too 👍🏻

PR #4276, opened 2023-09-09
Add warning about unspecified join condition behaviour

greg-rychlewski:
I saw some confusion around this on Elixir Forum so tried to improve the docs

ref: https://elixirforum.com/t/why-does-a-missing-on-option-in-ecto-query-join-5-trigger-this-warning/58197

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1279,7 +1279,7 @@ defmodule Ecto.Query do
   after the join. In the expression syntax, the options are given as
   the fifth argument.
 
-   > ### Unspecified join condition {: .warning}
+  > ### Unspecified join condition {: .warning}
   >
   > Leaving the `on` option unspecified will trigger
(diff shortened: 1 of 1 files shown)


lib/ecto/query.ex
```diff
@@ -1279,6 +1279,15 @@ defmodule Ecto.Query do
...
   the fifth argument.
 
+  > ### Unspecified join condition {: .warning}
+  >
+  > Leaving the `on` option unspecified will trigger
+  > a warning when not performing a cross join. This
+  > is to help users avoid performing expensive cross
+  > joins when they don't mean to. To remove the warning,
+  > change the join type to a cross join or explicitly set
+  > `on: true`.
```
v0idpwn, line 1289: Maybe we should also suggest specifying :on in case it was forgotten?

v0idpwn (approved):

PR #4277, opened 2023-09-15
Add returning to delete

Schultzer:
This PR adds the returning option to the delete function, which aligns with update and insert.

This is useful if you want to listen to telemetry events for deleted rows from another process that would execute: `MyRepo.delete(%MySchema{}, returning: true)`

The Ecto.SQL companion PR can be found here:  [#559](https://github.com/elixir-ecto/ecto_sql/pull/559).

Currently, TDS and Postgres support returning; for MySQL, one could use a temporary table: https://stackoverflow.com/questions/33115526/in-mysql-can-i-return-deleted-rows-after-a-deletion and https://dev.mysql.com/doc/refman/8.0/en/internal-temporary-tables.html.

I copied and pasted the code from insert and update, and I'm unsure if I need to call: `add_read_after_writes` or `fields_to_sources` after `returning`. If anyone has some guidance on this, it would be highly appreciated!


greg-rychlewski:
Thanks for the PR. I think it makes sense but I'm not sure what the stance will be on [breaking the existing behaviour for adapters](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapter/schema.ex#L90).

If it is accepted then some comments:

- We should add integration tests with the tag `:returning`
- We should document the option
- `fields_to_sources` is a must because it maps schema field names to database column names if they are different. also it puts `returning` into the right structure to be used by the adapters
- using `reads_after_writes` makes sense to me since insert/update are doing it. the one argument i see against it is that it will change the return value of existing apps. i'm not sure if that is enough to not use it though.

josevalim:
We could keep it backwards compatible by checking inside the Ecto.Repo.Schema if the adapter implements delete/5 or delete/4 with `function_exported?(module, :delete, 5)`. If there is no delete/5, we don't pass the return types for now. Then we can remove the check on later Ecto versions.

lib/ecto/repo/schema.ex
```diff
@@ -552,14 +558,23 @@ defmodule Ecto.Repo.Schema do
           apply(mod, on_delete, [reflection, changeset.data, name, tuplet])
         end
 
-        schema_meta = metadata(struct, schema.__schema__(:autogenerate_id), opts)
+        autogen_id = schema.__schema__(:autogenerate_id)
+        schema_meta = metadata(struct, autogen_id, opts)
+        {_changes, _cast_extra, _dump_extra, return_types, return_sources} =
+          autogenerate_id(autogen_id, changeset.changes, return_types, return_sources, adapter)
```
josevalim, line 564: I don't believe we need to call this function, as we should not autogenerate keys on deletions. Is there a reason why you call it?

Schultzer:
Thanks for the feedback; I've made it backward-compatible and added additional tests and documentation for the returning option. When writing the new test, I've decided to include the autogenerated fields as the insert function does; it makes sense to me that delete and insert return autogenerated fields and primary_keys. Please let me know if my assumptions are wrong on this matter.

josevalim:
Thanks for the updates. I don't think we should invoke any autogenerated value, so we should revert that. :) Especially primary keys, they are already meant to exist and be generated for delete to work. :)

lib/ecto/repo/schema.ex
Schultzer: I didn't see your comment before I posted it below: https://github.com/elixir-ecto/ecto/pull/4277#issuecomment-1721278859

In short, I followed the other tests, and I think I might have gotten caught up by the ID not being included:

https://github.com/elixir-ecto/ecto/blob/master/test/ecto/repo_test.exs#L506L529

In my mind, it makes sense always to include the primary keys when only the `:parent_id` key has been selected since it might not be enough information to determine which row has been deleted.

FWIW, I'm not opposed to reverting this, but I look forward to hearing your decision.

Schultzer:
> Thanks for the updates. I don't think we should invoke any autogenerated value, so we should revert that. :) Especially primary keys, they are already meant to exist and be generated for delete to work. :)

Thanks. I've reverted and updated the test and documentation to reflect that.

lib/ecto/adapter/schema.ex
```diff
@@ -87,6 +87,6 @@ defmodule Ecto.Adapter.Schema do
   in case there is no record matching the given filters,
   `{:error, :stale}` is returned.
   """
-  @callback delete(adapter_meta, schema_meta, filters, options) ::
+  @callback delete(adapter_meta, schema_meta, filters, returning, options) ::
```
greg-rychlewski, line 90: I think we need to have both arities.

josevalim:

lib/ecto/adapter/schema.ex
josevalim: Older versions will get the warning but still work, that's good enough to me. :)

greg-rychlewski: ah thanks I didn't realize that

PR #4279, opened 2023-09-15
Improve traverse_* typespecs

aaronrenner:
The current typspecs for `traverse_errors/2` and `traverse_validations/2` don't take into account errors returned from nested changesets. 

https://github.com/elixir-ecto/ecto/blob/32875ccf6482126e3ae7928606a1676a1b05d9da/test/ecto/changeset/has_assoc_test.exs#L1233

This PR extracts a custom `traverse_result` type so the typespec can be recursive.


josevalim:

PR #4280, opened 2023-09-19
Respect late binding types

greg-rychlewski:
I was reading Elixir Forum and came across a post that made it seem like as/parent_as types weren't being respected. Reference: https://elixirforum.com/t/ecto-types-and-named-bindiings/58370

When I looked into it I found that `as/parent_as` and `field(as/parent_as, _)` were having their types set to `:any`. When I fixed this I also noticed some tests started failing. So it actually uncovered some faulty tests:

1. Some of the late binding tests were comparing the `code` field to an integer when code is a binary. So casting should fail. When I made my changes the tests started failing. I updated the tests to cast to an integer field to make them pass and to ensure casting was being done properly.
2. Some of the parent_as tests were referencing a field that didn't exist on the parent. Namely, a `deleted` field on the posts schema. It wasn't failing before because the type of the field wasn't being looked up. Now that it's being looked up it failed. I fixed the test in a similar manner described above.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -629,14 +629,7 @@ defmodule Ecto.Query.Builder do
 
   defp escape_field!({kind, _, [value]}, field, _vars)
        when kind in [:as, :parent_as] do
-    value =
-      case value do
-        {:^, _, [value]} ->
-          value
-
-        other ->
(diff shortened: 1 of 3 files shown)


josevalim (approved):
Great find!

Issue #4281, opened 2023-09-19
parent_as/1 does not work in the RHS of union_all/2


marcelloma:
### Elixir version

Elixir 1.14.4 (compiled with Erlang/OTP 24)

### Database and Version

PostgreSQL 11.13

### Ecto Versions

Ecto 3.10.3

### Database Adapter and Versions (postgrex, myxql, etc)

Ecto SQL 3.10.1 / Postgrex 0.17.3

### Current behavior

When using `union_all/2` `parent_as/1` is only available in the LHS query of the union.

```elixir
  # This query is a bit nonsensical it was contrived to show the union_all bug but its a pointless query.
  def ecto_bug() do
    query_a =
      from(tm in TeamMember)
      |> select([tm], tm.member_id)
      |> where([tm], tm.team_id == parent_as(:t).id)

    query_b =
      from(tm in TeamMember)
      |> select([tm], tm.member_id)
      |> where([tm], tm.team_id == parent_as(:t).id)

    query_c = union_all(query_a, ^query_b)

    query = from(t in Team, as: :t, where: t.id in subquery(query_c))
    Repo.all(query)
  end
```

This the error I get from the function above:

```
** (Ecto.SubQueryError) the following exception happened when compiling a subquery.

    ** (Ecto.QueryError) could not find named binding `parent_as(:t)` in query:
    
    from t0 in TeamMember,
      where: t0.team_id == parent_as(:t).id,
      select: t0.member_id
    

The subquery originated from the following query:

from t0 in Team,
  as: :t,
  where: t0.id in subquery(
  #Ecto.Query<from t0 in TeamMember, where: t0.team_id == parent_as(:t).id, union_all: (from t0 in TeamMember,
  where: t0.team_id == parent_as(:t).id,
  select: t0.member_id), select: %{member_id: t0.member_id}>
),
  select: t0

    (elixir 1.14.4) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir 1.14.4) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (ecto 3.10.3) lib/ecto/repo/queryable.ex:211: Ecto.Repo.Queryable.execute/4
    (ecto 3.10.3) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
```

If I comment out `where([tm], tm.team_id == parent_as(:t).id)` from query_b it works fine.

### Expected behavior

[...]

PR #4282, opened 2023-09-20
Pass parent aliases to combination queries

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4281

The issue is that the right hand side of combinations queries are stored in the `:combinations` list of the main query and they weren't being passed the parent aliases.

This only matters during normalization because that is when the parent index is checked. Then after the query is normalized the parent alias is popped out again and is handled later in Ecto SQL. 

Companion Ecto SQL PR: https://github.com/elixir-ecto/ecto_sql/pull/561

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1167,9 +1167,23 @@ defmodule Ecto.Query.Planner do
   end
   defp prewalk_source(%Ecto.SubQuery{query: inner_query} = subquery, kind, query, _expr, counter, adapter) do
     try do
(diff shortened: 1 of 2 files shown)


josevalim (approved):

v0idpwn (approved):

greg-rychlewski:
Sorry for the update. I realized I could do it this way instead. I think this is a bit nicer but what do you guys think?

greg-rychlewski:
@josevalim Sorry just wanted to double check before merging. Are you ok with the small refactor I did?

josevalim (approved):

Issue #4284, opened 2023-09-25
`returning: true` option from `Repo.insert` doesn't respect `load_in_query: true` fields


y86:
### Elixir version

Elixir 1.14.2 (compiled with Erlang/OTP 25)

### Database and Version

PostgreSQL 14

### Ecto Versions

 3.10.3

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex at 0.17.3

### Current behavior

When using the `return: true` option of `Repo.insert`, it returns all fields, even the ones marked with `load_in_query: false` in the Schema.

### Expected behavior

Even though the docs state "When true, returns all fields." for the `returning` option, I expected fields marked with `load_in_query: false` to not be present in this return select query to be consistent with the `Ecto.Query.select` behavior that, when omitted, returns a full schema, but respecting the `load_in_query` option.

PR #4285, opened 2023-09-26
Make `returning` docs more explicit

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4284.

It may or may not make sense to remove anything with `load_in_query: false` but I think since it already returns them the safest thing to do is document it?


josevalim (approved):
I agree this is the best way to go, especially because `load_in_query` explicitly says "selecting".

PR #4286, opened 2023-09-26
docs/schemaless-changesets: common data structure

kevinschweikert:
The documentation for schemaless changesets was confusing, because the struct had different fields than the types. This PR unifies the example to a common data structure, so it is more understandable by the reader.


josevalim:

Issue #4287, opened 2023-09-28
`select_merge` repeating all fields on query


thiagopromano:
### Elixir version

1.15

### Database and Version

Postgres

### Ecto Versions

3.10

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex

### Current behavior

After upgrading to Ecto 3.10.x, we've noticed a significant decrease in performance on our queries, after close analysis, we've noticed that the queries sent to the DB had fields being repeated multiple times on the select statement.

This is what we get on `ecto` and `ecto_sql` 3.10.0:

```elixir
iex(1)> import Ecto.Query
Ecto.Query
iex(2)> q = from(u in User, select: %{id: u.id})
#Ecto.Query<from u0 in User, select: %{id: u0.id}>
iex(3)> q = from(u in q, select_merge: map(u, [:name]))
#Ecto.Query<from u0 in User,
 select: merge(%{id: map(u0, [:name]).id}, map(u0, [:name]))>
iex(4)> q = from(u in q, select_merge: map(u, [:email]))
#Ecto.Query<from u0 in User,
 select: merge(
  merge(%{id: map(u0, [:name, :email]).id}, map(u0, [:name, :email])),
  map(u0, [:name, :email])
)>
iex(5)> q = from(u in q, select_merge: map(u, [:inserted_at]))
#Ecto.Query<from u0 in User,
 select: merge(
  merge(
    merge(
      %{id: map(u0, [:name, :email, :inserted_at]).id},
      map(u0, [:name, :email, :inserted_at])
    ),
    map(u0, [:name, :email, :inserted_at])
  ),
  map(u0, [:name, :email, :inserted_at])
)>
iex(6)> q = from(u in q, select_merge: map(u, [:updated_at])) 
#Ecto.Query<from u0 in User,
 select: merge(
  merge(
    merge(
      merge(
        %{id: map(u0, [:name, :email, :inserted_at, :updated_at]).id},
        map(u0, [:name, :email, :inserted_at, :updated_at])
      ),
      map(u0, [:name, :email, :inserted_at, :updated_at])
    ),
    map(u0, [:name, :email, :inserted_at, :updated_at])
  ),
  map(u0, [:name, :email, :inserted_at, :updated_at])
)>
iex(7)> Repo.print_sql(q)
[...]

PR #4288, opened 2023-09-29
Don't duplicate fields with select_merge and map/2

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4287

This is a bit involved but I'll try my best to explain the issue. To trigger this issue you need the following conditions:

1. To produce a `{:merge, [left, right]}` expression, i.e. when `map/2` is merged with `%{}`
2. To merge something that has a source expression into the result of (1), i.e. map/2 is given the expression `{:&, [], [0]}`

I'll describe below how the issue manifests referencing this query:

```elixir
from(s in "schema", select: %{id: s.id})
|> select_merge([s], map(s, [:x]))
|> select_merge([s], map(s, [:y])
```

1. When you go into `collect_fields` the first time, `from` is `:none`.
2. When collecting the fields from the merge of `%{id: s.id}` and `map(s, [:x])` two things happen:
    - from changes from `:none` to a complex tuple tagged by `:source`
    - `:x` and `:y` get added to the fields list because the `take` from both maps are merged before the planner
3. When collecting the fields from the merge of `map(s, [:y])` with the result of the above it hits this piece of code

```elixir
defp collect_fields({:&, _, [ix]}, fields, from, query, take, _keep_literals?, drop) do
  {expr, taken} = source_take!(:select, query, take, ix, ix, drop)
  {expr, Enum.reverse(taken, fields), from}
end
```
So now it's adding `:x` and `:y` to the fields again and we get the duplicates. The solution is to not touch the fields once you've processed a `{:&, [], [0]}` expression because it is already considering the merged fields.

The code I added actually was there before but removed here https://github.com/elixir-ecto/ecto/commit/75431735644309e20bb030da88743a9e623300e2.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1455,6 +1455,11 @@ defmodule Ecto.Query.Planner do
     {{:source, :from}, fields, {{:source, :from}, expr, taken}}
   end
 
+  defp collect_fields({:&, _, [0]}, fields, from, _query, _take, _keep_literals?, _drop)
+       when from != :never do
+    {{:source, :from}, fields, from}
+  end
+
   defp collect_fields({:&, _, [ix]}, fields, from, query, take, _keep_literals?, drop) do
     {expr, taken} = source_take!(:select, query, take, ix, ix, drop)
     {expr, Enum.reverse(taken, fields), from}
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Thanks for debugging and fixing my mess. :)

PR #4290, opened 2023-10-01
Raise on literal non-base binary/uuids in query

### LINKED ISSUE #4289: Query fails to encode on literal UUID
### Elixir version

All

### Database and Version

PostgreSQL (all)

### Ecto Versions

All

postgrex

### Current behavior

A query on a Ecto.UUID field, where the UUID is literal, emits SQL with invalid bytes:

```
where([s], s.uuid == "017f65d1-80bd-152d-f997-afa6dd33a00f")
```

The reason why this happens is because the Planner calls Ecto.UUID.dump, which returns a binary, which we then promptly inject into the query.

Ecto actually has a mechanism to deal with these, called Ecto.Query.Tagged, and the original parameter is wrapped in such a struct, but the struct is removed during the Planner.prewalk time. The only reason why this doesn't happen to binaries is because we explicit keep the tagged struct:

```
  defp prewalk(%Ecto.Query.Tagged{value: v, type: type} = tagged, kind, query, expr, acc, adapter) do
    if Ecto.Type.base?(type) do
      {tagged, acc}
    else
      {dump_param(kind, query, expr, v, type, adapter), acc}
    end
  end
```

### Expected behavior

We should probably dump the param and check its underlying type, if the value is a string and the underlying type is not a string (or any), then we should probably keep the tagged struct. Note we would still need to support Tagged UUIDs in EctoSQL.

josevalim:
Closes #4289.
Closes https://github.com/elixir-ecto/ecto_sql/issues/520.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1309,7 +1309,28 @@ defmodule Ecto.Query.Planner do
     if Ecto.Type.base?(type) do
       {tagged, acc}
     else
-      {dump_param(kind, query, expr, v, type, adapter), acc}
+      type = field_type!(kind, query, expr, type)
+
+      with {:ok, type} <- normalize_param(kind, type, v),
+           {:ok, value} <- dump_param(adapter, type, v) do
+        # We cannot encode binary/uuid in queries because they would emit
+        # invalid queries with binary parts in them. In theory, we could
+        # wrap them in Ecto.Query.Tagged, but a tagged UUID would most
+        # likely wrap its string representation, not its binary one.
+        # So it is best to be consistent and not support query-dumping of
+        # non-base types.
+        if is_binary(value) and Ecto.Type.type(type) in [:binary_id, :binary, :uuid] do
+          error = "cannot encode value `#{inspect v}` of type `#{inspect(type)}` within a query, please interpolate (using ^) instead"
+          error! query, expr, error
+        else
+          {value, acc}
+        end
+      else
+        {:error, error} ->
+          error = error <> ". Or the value is incompatible or it must be " <>
+                           "interpolated (using ^) so it may be cast accordingly"
+          error! query, expr, error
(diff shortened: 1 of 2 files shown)


v0idpwn (approved):
I like this approach better than all alternatives considered.

lib/ecto/query/planner.ex
```diff
@@ -1309,7 +1309,28 @@ defmodule Ecto.Query.Planner do
     if Ecto.Type.base?(type) do
       {tagged, acc}
     else
-      {dump_param(kind, query, expr, v, type, adapter), acc}
+      type = field_type!(kind, query, expr, type)
```
greg-rychlewski, line 1312: Sorry if this is noise but I remember playing around with this and you might have to call `Ecto.Type.type` on this to get `:uuid`? Otherwise I think it will be `Ecto.UUID`

greg-rychlewski:
I agree this is the cleanest approach.

lib/ecto/query/planner.ex
josevalim: It is all good as is. `field_type` gets the type from a field, which may still be `Ecto.UUID`. But the Ecto.Type calls always reserve the type properly. :)

greg-rychlewski: oh you are already considering that below. ignore me!

greg-rychlewski (approved):

josevalim:

greg-rychlewski:
One thing I forgot to mention is that the old behaviour worked fine for MySQL so this will be a breaking change for them. I'm guessing it's a fair trade to have greater library stability. Just wanted to mention it.

josevalim:
Perfect, we will include it in the changelog. And it was probably fine for mysql query wise, although if the query was logged anywhere, it would be invalid utf8 and likely cause other breakages.

PR #4297, opened 2023-10-14
Add `Ecto.Multi.query/5`

greg-rychlewski:
I was in a situation recently where this would have been handy. Setting special configuration parameters for a transaction.

Not completely sure if it's desirable to add it since `query` behaves differently than the other repo functions. i.e. an adapter might not even expose. But I thought I'd see what you thought because it's still useful to reduce the verbosity of some transactions.

### DIFF
--- a/lib/ecto/multi.ex
+++ b/lib/ecto/multi.ex
@@ -496,6 +496,42 @@ defmodule Ecto.Multi do
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


josevalim:
I believe query is part of EctoSQL repo extension and not an official part of Repo, so that would make this a no-go. :)

greg-rychlewski:
No worries, makes sense.

Issue #4256, opened 2023-08-05
Ecto association doesn't reset after the ID has been updated (from Changeset changes)


francois08:
### Elixir version

1.14.4

### Database and Version

PostgreSQL 15

### Ecto Versions

3.6

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex

### Current behavior

Say in the changeset function I allow to change a foreign key:
```elixir
defmodule MyApp.Transactions.Revenue do
  schema "revenues" do
    # ...
    belongs_to :wallet, Wallet
  end

  def changeset(revenue, attrs) do
    required_fields = ~w(date amount wallet_id)a
    revenue
    |> cast(attrs, required_fields)
    |> validate_required(required_fields)
  end
end
```

```elixir
def update_revenue(%Revenue{} = revenue, attrs) do
  revenue
  |> Revenue.changeset(attrs)
  |> Repo.update()
end
```

If `revenue` had a wallet preloaded (which we can't know as all we know is that we receive a `revenue` struct with preloaded assocs or not) and the ID has been changed, after a successful update the association will not match with the ID.

Can't we reset the associations as we know from the Changeset that the IDs have changed (present in `:changes`)?

If not, I wonder how this problem is solved. It must be pretty common because as an example, I imagine a user may select the wallet through a combo box for the revenue, and so we receive a wallet ID in a form or absinthe request. Assuming the wallet will never be preloaded is dangerous and error-prone (and I opened this issue because what happened to me is that the `wallet` was preloaded - and didn't match with the ID which led to bugs - because the code for my real case is more complex than the example code shown, as it uses transactions for example).

We should know the source of truth in this case from `:changes`.

### Expected behavior

Expected the association to reset to not-loaded

josevalim:
This is “expected”. Ecto never loads and reloads data behind the scenes for associations, Because while it works in some cases, it also leads to surprising and often slow behavior in others. You can chose to preload the association again by calling Repo.preload with the force option.

francois08:
What do you think about something like this? (quickly wrote it down, it would be adapted to make it reusable)

```elixir
  def changeset(revenue, attrs) do
    required_fields = ~w(date amount wallet_id)a
    revenue
    |> cast(attrs, required_fields)
    |> validate_required(required_fields)
    |> unload_assocs()
  end

  defp unload_assocs(changeset) do
    if changed?(changeset, :wallet_id) do
[...]

PR #4299, opened 2023-10-16
Reset belongs_to association if foreign key update results in a mismatch

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4256

Some callouts:

1. This can only take effect if the foreign key is changed and the assoc field is not. If both are changed then mismatches are handled here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L925
2. Given the above, the reset doesn't actually trigger any change in the database. It's only a post-processing cleanup to ensure the returned struct doesn't have any inconsistencies. So the cleanup can be handled in the post-process step.
3. We only have to consider the update operation. During insert everything is surfaced as a change so it is handled by point 1

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -379,7 +379,7 @@ defmodule Ecto.Repo.Schema do
             values = dump_extra ++ values
 
             changeset
-            |> load_changes(:loaded, return_types, values, embeds, autogen, adapter, schema_meta)
+            |> load_changes(:loaded, return_types, values, [], embeds, autogen, adapter, schema_meta)
             |> process_children(user_changeset, children, adapter, assoc_opts)
 
           {:error, _} = error ->
@@ -467,7 +467,7 @@ defmodule Ecto.Repo.Schema do
           case apply(user_changeset, adapter, action, args) do
             {:ok, values} ->
               changeset
-              |> load_changes(:loaded, return_types, values, embeds, autogen, adapter, schema_meta)
+              |> load_changes(:loaded, return_types, values, assocs, embeds, autogen, adapter, schema_meta)
               |> process_children(user_changeset, children, adapter, assoc_opts)
 
             {:error, _} = error ->
@@ -570,7 +570,7 @@ defmodule Ecto.Repo.Schema do
 
         case apply(changeset, adapter, :delete, args) do
           {:ok, values} ->
-            changeset = load_changes(changeset, :deleted, return_types, values, %{}, [], adapter, schema_meta)
+            changeset = load_changes(changeset, :deleted, return_types, values, [], %{}, [], adapter, schema_meta)
             {:ok, changeset.data}
 
           {:error, _} = error ->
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
There were some things I thought about to save iterating through all the association fields again. But I'm not sure they are worth it:

1. Creating a new reflection function to return only the `:belongs_to` assocs. But then we are duplicating some of functions in the schema.
2. Doing the  cleanup in the function that is already iterating over the associations. But I think it would make it pretty busy/messy: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L871

lib/ecto/repo/schema.ex
```diff
@@ -467,7 +467,7 @@ defmodule Ecto.Repo.Schema do
           case apply(user_changeset, adapter, action, args) do
             {:ok, values} ->
               changeset
-              |> load_changes(:loaded, return_types, values, embeds, autogen, adapter, schema_meta)
+              |> load_changes(:loaded, return_types, values, assocs, embeds, autogen, adapter, schema_meta)
```
josevalim, line 470: We have `parents`, which already returns only the parents assocs, reducing the amount we need to traverse. Although I wonder if we could move the cleaning up to `process_parents`?

greg-rychlewski: Right now only the changed parents are kept which are being taken care of here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L925.

But if we start keeping the unchanged parents we can do better. What do you think of the latest commit?

lib/ecto/repo/schema.ex
```diff
@@ -865,29 +865,34 @@ defmodule Ecto.Repo.Schema do
...
               %{^assoc => {:assoc, %{relationship: :child} = refl}} ->
-                {changes, parent, [{refl, value} | child]}
+                {changes, parent, [{refl, value} | child], unchanged_parent}
             end
 
           %{} ->
-            {changes, parent, child}
+            case types do
+              %{^assoc => {:assoc, %{relationship: :parent} = refl}} ->
+                {changes, parent, child, [refl | unchanged_parent]}
```
josevalim, line 888: Is it worth already computing if there is a mismatch here? Because we have changes and data as well, right? So we could directly return the list of fields to reset?

greg-rychlewski: ah good point. WDYT?

lib/ecto/repo/schema.ex
```diff
@@ -897,14 +915,15 @@ defmodule Ecto.Repo.Schema do
...
-  defp process_parents(changeset, user_changeset, assocs, adapter, opts) do
+  defp process_parents(changeset, user_changeset, assocs, reset_assocs, adapter, opts) do
     %{changes: changes, valid?: valid?} = changeset
 
     # Even if the changeset is invalid, we want to run parent callbacks
     # to collect feedback. But if all is ok, still return the user changeset.
     case Ecto.Association.on_repo_change(changeset, assocs, adapter, opts) do
       {:ok, struct} when valid? ->
         changes = change_parents(changes, struct, assocs)
+        struct = Ecto.reset_fields(struct, reset_assocs)
```
josevalim, line 926: Let's add `def reset_fields(struct, []), do: struct` to its definition, to ensure it is fast, and SHIP IT! :D

josevalim (approved):

PR #4301, opened 2023-10-17
Fix unit tests for OTP 26

greg-rychlewski:
just the unit tests for now

### DIFF
--- a/test/ecto/repo_test.exs
+++ b/test/ecto/repo_test.exs
@@ -595,19 +595,31 @@ defmodule Ecto.RepoTest do
         [y: query, z: query]
       ])
 
-      assert_received {:insert_all, %{source: "my_schema"}, rows}
+      assert_received {:insert_all, %{source: "my_schema", header: header}, planned_rows}
 
       assert [
                [x: "x1", yyy: "y1"],
-               [x: {%Ecto.Query{} = query2, [^value]}, z: "z2"],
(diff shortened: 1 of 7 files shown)


test/ecto/repo_test.exs
```diff
@@ -595,19 +595,31 @@ defmodule Ecto.RepoTest do
...
+      {queries_with_index, _} =
+        Enum.reduce(planned_rows, {[], 0}, fn row, {queries, ix} ->
+          Enum.reduce(header, {queries, ix}, fn field, {queries, ix} ->
+            case row[field] do
+              {%Ecto.Query{} = query, _} -> {[{ix, query} | queries], ix + 1}
+              nil -> {queries, ix}
+              _ -> {queries, ix + 1}
+            end
+          end)
+        end)
```
josevalim, line 616:
```suggestion
        for row <- planned_rows, field <- header, reduce: {[], 0} do
          {queries, ix} ->
            case row[field] do
              {%Ecto.Query{} = query, _} -> {[{ix, query} | queries], ix + 1}
              nil -> {queries, ix}
              _ -> {queries, ix + 1}
            end
        end
```

josevalim (approved):

test/ecto/repo_test.exs
greg-rychlewski: nice :)

PR #4305, opened 2023-10-23
Fix rollback spec

NickNeck:


lib/ecto/repo.ex
```diff
@@ -2149,10 +2149,10 @@ defmodule Ecto.Repo do
...
   Rolls back the current transaction.
 
-  The transaction will return the value given as `{:error, value}`.
+  The transaction will return the `value` given as `{:error, value}`.
 
   Note that calling `rollback` causes the code in the transaction to stop executing.
   """
   @doc group: "Transaction API"
-  @callback rollback(value :: any) :: no_return
+  @callback rollback(value) :: {:error, value} when value: any
```
josevalim, line 2157: I believe rollback raises but the transaction returns the error tuple, no?

greg-rychlewski: Yeah it throws and then transaction catches it

```elixir
def rollback(%DBConnection{conn_mode: :transaction} = conn, reason) do
    %DBConnection{conn_ref: conn_ref} = conn
    throw({__MODULE__, conn_ref, reason})
  end

def transaction(%DBConnection{conn_mode: :transaction} = conn, fun, _opts) do
    %DBConnection{conn_ref: conn_ref} = conn

    try do
      result = fun.(conn)
      conclude(conn, result)
    catch
      :throw, {__MODULE__, ^conn_ref, reason} ->
        fail(conn)
        {:error, reason}

      kind, reason ->
        stack = __STACKTRACE__
        fail(conn)
        :erlang.raise(kind, reason, stack)
    else
      result ->
        {:ok, result}
    end
  end
```

PR #4309, opened 2023-10-30
Add FilterModel utility

andrew-combs:
This is something we developed to help with building queries that were very dependent on association filters. This let us quickly compose the right "shape" of the data and then construct a query from it.

After diving into the internals of Ecto to make it more extensible to all our schemas, we realized it was a generally useful module that might have a home in the core Ecto repository.

The interface was a rough-cut based on our primary use cases, but I'd love to hear how the semantics could be evolved if they don't seem very intuitive.

Thanks!

### DIFF
--- a/lib/ecto/query/filter_model.ex
+++ b/lib/ecto/query/filter_model.ex
@@ -1,0 +1,249 @@
+defmodule Ecto.Query.FilterModel do
+  @moduledoc """
+  Allows a user to take any module that implements Ecto.Schema and
+  build a query using a map that represents that schema's data shape.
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @andrew-combs , thanks for the PR. I believe this is best suite as a library or part of a project, because if it becomes part of Ecto, then we have to eventually support all of Ecto's query syntax, which will just lead to complexity in the long term. Libraries and usage in your application will be much more comfortable in accepting a limited scope. :)

andrew-combs:
Makes perfect sense to me! We'll open it up somewhere else as a standalone utility

PR #4312, opened 2023-11-09
Small doc improvement on c:equal?/2

leandrocp:
To make it clear how the terms are compared by default, as defined at https://github.com/elixir-ecto/ecto/blob/8fac7068ed28cbb63c9844bb4cfba09fb48b9a09/lib/ecto/type.ex#L171


josevalim:

PR #4320, opened 2023-11-16
Updates lateral join docs to reflect current behavior

beerlington:
Fragments are no longer required when using lateral joins and can be replaced by `subquery` and `parent_as`.


josevalim:

Issue #4321, opened 2023-11-17
Ecto.Enum doesn't work with module attributes in Ecto.Query


cjbottaro:
### Elixir version

1.15.7

### Database and Version

PostgreSQL 15.3

### Ecto Versions

3.11.0

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.17.3

### Current behavior

```elixir
defmodule Thing do
  use Ecto.Schema

  schema "things" do
    field :type, Ecto.Enum, values: [foo: "Type::Foo", bar: "Type::Bar"]
  end

  @types [:foo, :bar]

  def test do
    require Ecto.Query

    Ecto.Query.from(t in Thing, where: t.type in @types)
    |> Repo.all()
  end
end

```
Calling `Thing.test/0` results in this error:
```
** (Ecto.QueryError) unsupported expression: :foo in query:

from t0 in Thing,
  where: t0.type in [:foo, :bar],
  select: t0

    (ecto_sql 3.11.0) lib/ecto/adapters/postgres/connection.ex:1870: Ecto.Adapters.Postgres.Connection.error!/2
    (elixir 1.15.7) lib/enum.ex:4372: Enum.map_intersperse_list/3
    (ecto_sql 3.11.0) lib/ecto/adapters/postgres/connection.ex:855: Ecto.Adapters.Postgres.Connection.expr/3
    (ecto_sql 3.11.0) lib/ecto/adapters/postgres/connection.ex:828: Ecto.Adapters.Postgres.Connection.paren_expr/3
    (ecto_sql 3.11.0) lib/ecto/adapters/postgres/connection.ex:805: Ecto.Adapters.Postgres.Connection.boolean/4
    (ecto_sql 3.11.0) lib/ecto/adapters/postgres/connection.ex:168: Ecto.Adapters.Postgres.Connection.all/2
    (ecto_sql 3.11.0) lib/ecto/adapters/postgres.ex:132: Ecto.Adapters.Postgres.prepare/2
```

Note that if you change `Thing.test/0` to:
```elixir
  def test do
    require Ecto.Query

    Ecto.Query.from(t in Thing, where: t.type in [:foo, :bar])
    |> Repo.all()
  end
```

Then everything works fine.

### Expected behavior

To be able to use module attributes in the query.

wojtekmach:
In the meantime this should work:

```elixir
where: t.type in ^@types
```

PR #4322, opened 2023-11-18
Allow module attributes to be given to in operator

greg-rychlewski:
My attempt to close https://github.com/elixir-ecto/ecto/issues/4321. 

I'm a bit unsure about this one so opening this for discussion/my own learning as well.

The issue is that when a module attribute is used like `p.field in @list`, it gets [sent here](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder.ex#L357):

```elixir
def escape({:in, _, [left, right]} = expr, type, params_acc, vars, env) do
    assert_type!(expr, type, :boolean)

    ltype = {:out, quoted_type(right, vars)}
    rtype = {:in, quoted_type(left, vars)}

    ....
end
```

And the fact that it is actually a list is not known and the type eventually gets set to :any. Making it so that it can't be cast to the type of `p.field` by the planner. 

Ideally it should [go here](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder.ex#L344C18-L344C18) so it gets tagged with the field type:

```elixir
def escape({:in, _, [left, right]} = expr, type, params_acc, vars, env)
      when is_list(right)
      when is_tuple(right) and elem(right, 0) in ~w(sigil_w sigil_W)a do
    assert_type!(expr, type, :boolean)

    {:array, ltype} = quoted_type(right, vars)
    rtype = {:array, quoted_type(left, vars)}
  
    ...
end
```

My fix attempt is to expand `right` before choosing which path to go down. This should call `Module.get_attribute` from what I know. If I'm not mistaken these escape functions will always be called at compile time so it should be ok.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -341,9 +341,38 @@ defmodule Ecto.Query.Builder do
   end
 
   # in operator
-  def escape({:in, meta, [left, right]}, type, params_acc, vars, env) do
+  def escape({:in, _, [left, right]} = expr, type, params_acc, vars, env)
+       when is_list(right)
+       when is_tuple(right) and elem(right, 0) in ~w(sigil_w sigil_W @)a do
+    assert_type!(expr, type, :boolean)
+
     right = Macro.expand_once(right, get_env(env))
-    escape_in({:in, meta, [left, right]}, type, params_acc, vars, env)
+
+    {:array, ltype} = quoted_type(right, vars)
+    rtype = {:array, quoted_type(left, vars)}
+
+    {left, params_acc} = escape(left, ltype, params_acc, vars, env)
+    {right, params_acc} = escape(right, rtype, params_acc, vars, env)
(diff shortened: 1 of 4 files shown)


lib/ecto/query/builder.ex
```diff
@@ -341,36 +341,9 @@ defmodule Ecto.Query.Builder do
   end
 
   # in operator
-  def escape({:in, _, [left, right]} = expr, type, params_acc, vars, env)
-      when is_list(right)
-      when is_tuple(right) and elem(right, 0) in ~w(sigil_w sigil_W)a do
```
josevalim, line 346: I wonder if the simplest fix is to add ~w(sigil_w sigil_W @) to this line. Then, if necessary, also add a Macro.expand_once in the body?

greg-rychlewski: Good call. That should work because the other function head is meant for fragment, Query.API stuff like type and subquery. So they won't be put into a module attribute.

Changed!

josevalim (approved):

Issue #4326, opened 2023-11-22
Unable to set keys dynamically using select_merge


fuelen:
### Elixir version

1.15.5

### Database and Version

PostgreSQL 14.1

### Ecto Versions

3.10.3

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.17.3

### Current behavior

This code works fine:
```elixir
iex> "users" |> select([u], %{}) |> select_merge([u], %{:test => "1"}) |> limit(1) |> Repo.one 
[debug] QUERY OK source="users" db=0.8ms
SELECT TRUE FROM "users" AS u0 LIMIT 1
↳ anonymous fn/4 in :elixir.eval_external_handler/1, at: src/elixir.erl:376
%{test: "1"}
```
```elixir
"users" |> select([u], %{}) |> select_merge([u], %{^"test" => "1"}) |> limit(1) |> Repo.one
[debug] QUERY OK source="users" db=0.8ms
SELECT 'test' FROM "users" AS u0 LIMIT 1
↳ anonymous fn/4 in :elixir.eval_external_handler/1, at: src/elixir.erl:376
%{"test" => "1"}
```

but this doesn't:
```elixir
iex(3)> "users" |> select([u], %{}) |> select_merge([u], %{^:test => "1"}) |> limit(1) |> Repo.one
[debug] QUERY ERROR source="users" db=1.8ms
SELECT 'test' FROM "users" AS u0 LIMIT 1
↳ anonymous fn/4 in :elixir.eval_external_handler/1, at: src/elixir.erl:376
** (DBConnection.EncodeError) Postgrex expected a binary, got :test. Please make sure the value you are passing matches the definition in your table or in your query or convert the value accordingly.
    (postgrex 0.17.3) lib/postgrex/type_module.ex:947: Postgrex.DefaultTypes.encode_params/3
    (postgrex 0.17.3) lib/postgrex/query.ex:75: DBConnection.Query.Postgrex.Query.encode/3
    (db_connection 2.6.0) lib/db_connection.ex:1403: DBConnection.encode/5
    (db_connection 2.6.0) lib/db_connection.ex:1503: DBConnection.run_prepare_execute/5
    (db_connection 2.6.0) lib/db_connection.ex:1607: DBConnection.run/6
    (db_connection 2.6.0) lib/db_connection.ex:743: DBConnection.parsed_prepare_execute/5
[...]

PR #4327, opened 2023-11-22
Don't send dynamic map keys to the database

greg-rychlewski:
I was reading over the issue from this morning about seeing dynamic map keys (https://github.com/elixir-ecto/ecto/issues/4326). I think I have a decent way to solve it.

The idea is that any interpolated map key doesn't need to be sent to the database because it won't affect the query:
- For subqueries/CTEs, only atom map key are allowed and those are never sent to the database
- For outer queries literals are never sent to the database

This means instead of making map keys query parameters we can simply validate them on the spot and save them into the query struct.

There is a limitation where we can really only interpolate map keys for basic literals like atoms, strings and numbers. Composite literals like tuples and maps would require recursive escaping, which we lose the ability to do because we are at runtime.

I'm also restricting this proposal to just map keys even though I think there might be a case to extend it to anywhere in select. But it is more complicated (i.e. the distinction between outer and inner query becomes important) and more tests break, so I thought I would save it for later.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1959,7 +1959,8 @@ defmodule Ecto.Query do
   >
   > When selecting a literal atom, its value must be the same across
   > all queries. Otherwise, the value from the parent query will be
-  > applied to all other queries.
+  > applied to all other queries. This also holds true for selecting
+  > maps with atom keys.
 
   Union expression returns only unique rows as if each query returned
   distinct results. This may cause a performance penalty. If you need
@@ -2011,7 +2012,8 @@ defmodule Ecto.Query do
   >
   > When selecting a literal atom, its value must be the same across
   > all queries. Otherwise, the value from the parent query will be
-  > applied to all other queries.
+  > applied to all other queries. This also holds true for selecting
+  > maps with atom keys.
 
   Note that the operations `order_by`, `limit` and `offset` of the
   current `query` apply to the result of the union. `order_by` must
@@ -2058,7 +2060,8 @@ defmodule Ecto.Query do
   >
   > When selecting a literal atom, its value must be the same across
   > all queries. Otherwise, the value from the parent query will be
-  > applied to all other queries.
+  > applied to all other queries. This also holds true for selecting
+  > maps with atom keys.
 
   Except expression returns only unique rows as if each query returned
   distinct results. This may cause a performance penalty. If you need
@@ -2110,7 +2113,8 @@ defmodule Ecto.Query do
   >
   > When selecting a literal atom, its value must be the same across
   > all queries. Otherwise, the value from the parent query will be
-  > applied to all other queries.
+  > applied to all other queries. This also holds true for selecting
+  > maps with atom keys.
 
   Note that the operations `order_by`, `limit` and `offset` of the
   current `query` apply to the result of the set difference. `order_by`
@@ -2157,7 +2161,8 @@ defmodule Ecto.Query do
   >
   > When selecting a literal atom, its value must be the same across
   > all queries. Otherwise, the value from the parent query will be
-  > applied to all other queries.
+  > applied to all other queries. This also holds true for selecting
+  > maps with atom keys.
 
   Intersect expression returns only unique rows as if each query returned
   distinct results. This may cause a performance penalty. If you need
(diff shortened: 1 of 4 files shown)


greg-rychlewski:
I think this is the same as what you were thinking with `literal` and can be applied anywhere in select. Though I have to play around with it more to see if that causes any issues.

The reason I omitted `literal` here is because these values don't need to go into the actual query and won't go if they are constants rather than dynamic values.

greg-rychlewski:
This wont work =/. Subqueries keep the literals it's only the outer query that ignores them.

greg-rychlewski:
I requested a review because I wasn't sure if this had slipped by you or if you didn't have time for it right now. If the latter please ignore me.

josevalim:
Thank you for requesting it. It definitely slipped through. I think we just need to check this is also sane for unions/intersections, if you haven't yet!

josevalim (approved):

greg-rychlewski:
@fuelen you could try this out to see how it works for you :)

greg-rychlewski:
@josevalim Actually sorry there is something about combinations I didn't consider. I was hoping to get your opinion.

If there is a combination and the user interpolates a string map key then the string literal will be sent to the adapter. Then it will be up to the adapter to escape string properly to avoid something like this

```elixir
union_query = from(c in Post, select: %{"id" => c.id})
outer_map_key = "id'; drop table posts; select 'id"
from(c in Post, select: %{^outer_map_key => c.id}, union: ^union_query)
```

Adapters are suppose to be escaping string literals properly. Otherwise even literals defined by the dev and not the user won't work properly. But it seems extreme to open them to SQL injection from outside users if they fail to do that.

This means if we want to close this hole, we have to continue to have interpolated map keys get changed to query parameters. And for atoms we need to have an escape hatch. There are two mechanisms we have today that kind of fit but not exactly:

1. `unsafe_fragment`. This is allowed for hints and insert conflicts to inject sql directly without modification. It's not exactly the same as here where we are injecting a literal that will have some handling in the adapter, i.e. quoting/escaping
2. `literal/1`. This is used as an escape hatch inside of fragments that allows the user to inject identifiers directly into the query, i.e. column names, collation names, etc. It's not exactly the same as here because the adapters will see this and quote it as an identifier rather than a literal string.

Based on the above, I can think of a few options:

1. Create something new, like `unsafe_literal/1` and use it here. I don't think we need to change the existing `literal/1` because it is treated specially by adapters so it is safe
2. Say this is ok, the adapter is supposed to be escaping strings properly
3. Say there is no good way to handle this and completely revert

I am personally leaning towards option 1 but this feels hairy so not super confident. WDYT?

greg-rychlewski:
Oh yeah sorry what I meant is that adapters still need to escape string literals properly. And the current ones do.

I was asking more from the angle of new adapters that might have mistakes in their implementation at the beginning.

greg-rychlewski:
I'm good to release. I don't see a release to hold this feature back.

josevalim:
@greg-rychlewski I am reconsidering this one. Here we are using `^foo` but we are not sending it as parameter to the database, right? Do we do this anywhere else? 🤔 It sounds like an exception to how Ecto works, no?

josevalim:
Alternatively, if we don't send it to the database, then we are still not passing it as parameter either. Do we do it anywhere else besides literal? or unsafe_fragment?

greg-rychlewski:
I believe the only two are literal and unsafe_fragment. But they are not really matching this purpose too well.

For most things they won't be sent to the database but if combinations exist then they will. So there is definitely the gap you are saying in that scenario.

I'm pretty pessimistic that we can make combinations work without sending literals to the database. For example:

```sql
select "string1" union select "string2"
```

Something like this would be returned as ["select1", "select1"] from Ecto if the actual values are not sendt to the database. Since we don't know where the main query ends/combination query starts in the results from the driver we can't really extract the literals and place them back in the right place after.

Given the above, I can only think of two options. Either scrap this entirely or introduce something like unsafe_literal in selects.

greg-rychlewski:
It might also be possible in the planner to modify this part so that `keep_literals?` for the key is changed to be always false. I could play around with it to see if that causes issues.

```elixir
defp collect_kv([{key, value} | elems], fields, from, query, take, keep_literals?, acc) do
    {key, fields, from} = collect_fields(key, fields, from, query, take, keep_literals?, %{})
    {value, fields, from} = collect_fields(value, fields, from, query, take, keep_literals?, %{})
    collect_kv(elems, fields, from, query, take, keep_literals?, [{key, value} | acc])
  end
```

greg-rychlewski:
No the above will break existing code for combinations. For example:

```elixir
from c in Comment, select: %{"outer_title" => c.title}, union: ^from(c in Comment, select: %{"inner_title" => c.title})
```

If the map key is sent to the database then it will return correctly

```elixir
[%{"outer_title" => "title"},  %{"inner_title" => "title"}]
```

But if it is not sent to the database it will return this

```elixir
[%{"outer_title" => "title"},  %{"outer_title" => "title"}]
```

greg-rychlewski:
selected_as 1 and 2, as well

PR #4328, opened 2023-11-23
Proposal: Add native support to bitstring

Gigitsu:
Recently, I had the need to store a variable-size bitstring in a PostgreSQL database. I was surprised to find that there isn't native support in Ecto for this data type, and the `:string` type doesn't work for me since I need to store a string where the number of bits is not a multiple of 8.

This PR is currently just a draft, but if this could be a desired feature, I would be happy to implement the entire feature set, including an `Ecto.Query` set of bitwise operators to work with bits and a corresponding `Ecto` schema type.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -2,12 +2,13 @@ VERSION 0.6
 
 all:
     BUILD \
-        --build-arg ELIXIR_BASE=1.15.6-erlang-25.3.2.6-alpine-3.16.7 \
-        --build-arg ELIXIR_BASE=1.15.6-erlang-24.3.4.14-alpine-3.16.7 \
+        --build-arg ELIXIR_BASE=1.15.6-erlang-25.3.2.6-alpine-3.18.4 \
+        --build-arg ELIXIR_BASE=1.15.6-erlang-24.3.4.14-alpine-3.18.4 \
         +integration-test
 
 integration-test-base:
-    ARG ELIXIR_BASE=1.15.6-erlang-25.3.2.6-alpine-3.16.7
+    ARG ELIXIR_BASE=1.15.6-erlang-25.3.2.6-alpine-3.18.4
+    ARG TARGETARCH 
     FROM hexpm/elixir:$ELIXIR_BASE
     RUN apk add --no-progress --update git build-base
     RUN mix local.rebar --force
@@ -17,13 +18,13 @@ integration-test-base:
     RUN apk add --no-progress --update docker docker-compose git postgresql-client mysql-client
 
     RUN apk add --no-cache curl gnupg --virtual .build-dependencies -- && \
-        curl -O https://download.microsoft.com/download/e/4/e/e4e67866-dffd-428c-aac7-8d28ddafb39b/msodbcsql17_17.5.2.1-1_amd64.apk && \
-        curl -O https://download.microsoft.com/download/e/4/e/e4e67866-dffd-428c-aac7-8d28ddafb39b/mssql-tools_17.5.2.1-1_amd64.apk && \
-        echo y | apk add --allow-untrusted msodbcsql17_17.5.2.1-1_amd64.apk mssql-tools_17.5.2.1-1_amd64.apk && \
+        curl -O https://download.microsoft.com/download/3/5/5/355d7943-a338-41a7-858d-53b259ea33f5/msodbcsql18_18.3.2.1-1_${TARGETARCH}.apk && \
+        curl -O https://download.microsoft.com/download/3/5/5/355d7943-a338-41a7-858d-53b259ea33f5/mssql-tools18_18.3.1.1-1_${TARGETARCH}.apk && \
+        echo y | apk add --allow-untrusted msodbcsql18_18.3.2.1-1_${TARGETARCH}.apk mssql-tools18_18.3.1.1-1_${TARGETARCH}.apk && \
         apk del .build-dependencies && rm -f msodbcsql*.sig mssql-tools*.apk
-    ENV PATH="/opt/mssql-tools/bin:${PATH}"
+    ENV PATH="/opt/mssql-tools18/bin:${PATH}"
 
-    GIT CLONE https://github.com/elixir-ecto/ecto_sql.git /src/ecto_sql
+    GIT CLONE --branch feature/bitstring-support https://github.com/Gigitsu/ecto_sql.git /src/ecto_sql
     WORKDIR /src/ecto_sql
     RUN mix deps.get
 
@@ -44,7 +45,7 @@ integration-test:
     ARG MYSQL_IMG="mysql:5.7"
 
     # then run the tests
-    WITH DOCKER --pull "$PG_IMG" --pull "$MCR_IMG" --pull "$MYSQL_IMG"
+    WITH DOCKER --pull "$PG_IMG" --pull "$MCR_IMG" --pull "$MYSQL_IMG" --platform linux/amd64
         RUN set -e; \
             timeout=$(expr $(date +%s) + 60); \
 
@@ -54,7 +55,7 @@ integration-test:
             docker run --name mysql --network=host -d -e MYSQL_ROOT_PASSWORD=root "$MYSQL_IMG"; \
 
             # wait for mssql to start
-            while ! sqlcmd -S tcp:127.0.0.1,1433 -U sa -P 'some!Password' -Q "SELECT 1" >/dev/null 2>&1; do \
+            while ! sqlcmd -C -S tcp:127.0.0.1,1433 -U sa -P 'some!Password' -Q "SELECT 1" >/dev/null 2>&1; do \
                 test "$(date +%s)" -le "$timeout" || (echo "timed out waiting for mysql"; exit 1); \
                 echo "waiting for mssql"; \
                 sleep 1; \
--- a/Earthfile
+++ b/Earthfile
@@ -24,7 +24,7 @@ integration-test-base:
         apk del .build-dependencies && rm -f msodbcsql*.sig mssql-tools*.apk
     ENV PATH="/opt/mssql-tools18/bin:${PATH}"
 
-    GIT CLONE --branch feature/bitstring-support https://github.com/Gigitsu/ecto_sql.git /src/ecto_sql
+    GIT CLONE https://github.com/elixir-ecto/ecto_sql.git /src/ecto_sql
     WORKDIR /src/ecto_sql
     RUN mix deps.get
 
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -400,7 +400,7 @@ defmodule Ecto.Integration.Bitstring do
 
   schema "bitstrings" do
     field :bs,  :bitstring
-    field :bs_with_default, :bitstring, default: <<42::6>>
+    field :bs_with_default, :bitstring
     field :bs_with_size, :bitstring
   end
 end
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -401,6 +401,6 @@ defmodule Ecto.Integration.Bitstring do
   schema "bitstrings" do
     field :bs,  :bitstring
     field :bs_with_default, :bitstring, default: <<42::6>>
-    field :bs_with_size, :bitstring, size: 10
+    field :bs_with_size, :bitstring
   end
 end
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -47,7 +47,6 @@ defmodule Ecto.Integration.Post do
     field :intensities, {:map, :float}
     field :posted, :date
     field :read_only, :string, read_only: true
-    field :token, :bitstring
     has_many :comments, Ecto.Integration.Comment, on_delete: :delete_all, on_replace: :delete
     has_many :force_comments, Ecto.Integration.Comment, on_replace: :delete_if_exists
     has_many :ordered_comments, Ecto.Integration.Comment, preload_order: [:text]
@@ -387,5 +386,21 @@ defmodule Ecto.Integration.ArrayLogging do
   schema "array_loggings" do
     field :uuids, {:array, Ecto.Integration.TestRepo.uuid()}
     timestamps()
+  end
+end
+
+defmodule Ecto.Integration.Bitstring do
+  @moduledoc """
+  This module is used to test:
+
+    * Bitstring type
+
+  """
+  use Ecto.Integration.Schema
+
+  schema "bitstrings" do
+    field :bs,  :bitstring
+    field :bs_with_default, :bitstring, default: <<42::6>>
+    field :bs_with_size, :bitstring, size: 10
   end
 end
--- a/integration_test/support/schemas.exs
+++ b/integration_test/support/schemas.exs
@@ -47,6 +47,7 @@ defmodule Ecto.Integration.Post do
     field :intensities, {:map, :float}
     field :posted, :date
     field :read_only, :string, read_only: true
+    field :token, :bitstring
     has_many :comments, Ecto.Integration.Comment, on_delete: :delete_all, on_replace: :delete
     has_many :force_comments, Ecto.Integration.Comment, on_replace: :delete_if_exists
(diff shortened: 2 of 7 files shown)


josevalim:
We will be happy to add this but I believe your first challenge will be adding support for bit strings in Postgrex itself, unless it already works? :)

Gigitsu:
I'm currently using a custom Ecto type with Postgrex, which does pretty much what I've already implemented in this PR, and everything works happily :D

If you agree, I can go on with this PR and make a sample project to end-to-end test this feature with Postgresx.

josevalim:
For testing, please add some integration tests in the `integration_test` folder. Check the README for more info! You can add some tags so we skip these tests on MySQL and other adapters.

Gigitsu:
Great! Thank you.

lib/ecto/query/builder.ex
```diff
@@ -121,6 +121,11 @@ defmodule Ecto.Query.Builder do
     escape_with_type(op_expr, type, params_acc, vars, env)
   end
 
+  def escape({:type, _, [{bitwise_op, _, [_, _]} = op_expr, type]}, _type, params_acc, vars, env)
+      when bitwise_op in ~w(&&& |||)a do
```
greg-rychlewski, line 125:
```suggestion
      when bitwise_op in ~w(&&& ||| <<< >>>)a do
```
It looks like Postgres also supports the shift operators: https://www.postgresql.org/docs/current/functions-bitstring.html

But note we also need to make changes to Ecto SQL so the adapter can convert these operators into the correct query string (i.e. convert `&&&` to `&`): https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L420

Similarly for the unary operators we'd need to add function head like this: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L875

So that `~~~` works. We can probably do the same for the bitwise functions `band`, `bor`, `bnot`, `bsl`, `bsr`, and `bxor`. Since there is no binary operator for xor this would give users a clean way to use Postgres's xor operator.

Gigitsu:
Hi @greg-rychlewski, thank you very much for your review.

I'm working on a companion ecto_sql PR that will convert everything into a query string too.

lib/ecto/query/builder.ex
Gigitsu: Hi @greg-rychlewski, thank you very much for your review.

I'm working on a companion ecto_sql PR that will convert everything into a query string too.

Gigitsu:
This is the `ecto_sql` companion PR elixir-ecto/ecto_sql#577

greg-rychlewski:
It's looking good, thank you. I think only two things are missing:

**Ecto**
1. Integration tests. We can add them to this repo and tag them `@tag :bitwise` as Jose suggested. Then adapters that don't support it can ignore the tests

**Ecto SQL**
1. I believe we can add support for bitwise functions `band`, `bor`, `bxor`, `bnot`, `bsl`, `bsr`. This would be the only way to expose the Postgres xor operator through Ecto. It would involve creating patterns like this for those function https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L871. Care has to be taken about parentheses. i.e. binary ops need them and unary I believe don't.
2. Similar to point 2 we can create a pattern for the `~~~` operator to support negation

greg-rychlewski:
The integration tests can go here: https://github.com/elixir-ecto/ecto/blob/master/integration_test/cases/type.exs

Gigitsu:
I've just added the pattern for `~~~` :D and I'm working on `ecto_sql` to convert that into a SQL string.

I'll then add support for bitwise functions and integration tests.

Thank you for your support :) really appreciated.

Gigitsu:
Everything reverted also here!

lib/ecto/query/builder.ex
```diff
@@ -121,11 +121,29 @@ defmodule Ecto.Query.Builder do
     escape_with_type(op_expr, type, params_acc, vars, env)
   end
 
+  def escape({:type, _, [{bitwise_op, _, [_, _]} = op_expr, type]}, _type, params_acc, vars, env)
```
josevalim, line 124: I am a bit torn on adding these operations, because in Elixir they work on numbers, and here they work on bitstrings. It feels like a false equivalence to me. So, for now, let's further simplify this PR so it only adds the type but not the new operators? Sorry for the delay.

josevalim: > If you believe it would be better, I can modify this PR to introduce only the Ecto type. Afterwards, we can discuss the possibility of adding bitwise operators.

Yes please!

Gigitsu: Sure! No problem, I'll do it as soon as possible.

greg-rychlewski:
@josevalim there was one other part we were discussing in the ecto_sql PR. it was how to handle literals. one suggestion was to handle it in the builder doing something like this

```elixir
defp do_literal(value, _, current) when current in @always_tagged do
  type = 
    if current == :bitstring do
      quote do: if is_binary(unquote(value)), do: :binary, else: :bitstring
    else
      current
    end

  {:%, [], [Ecto.Query.Tagged, {:%{}, [], [value: value, type: type]}]}
end
```

and the other suggestion was to handle it in the adapter. Do you have any preference?

josevalim:
Let’s handle it in Ecto, I think it would be wrong to say it is a binary when it is actually a bitstring.

Gigitsu:
@josevalim I've removed every implementation of bitwise operators and added support to bitstring in migrations too both here and it the `ecto_sql` companion PR elixir-ecto/ecto_sql#577

I'm also trying to implement integration tests but I could use some help here.

I've managed to add a bitstring test that works for Postgres, but it seems that I'm having some trouble making it work for MySQL.
It appears that the MySQL connector doesn't handle bitstrings in a where clause like Postgres does (I need to investigate this further). If MySQL can't actually handle bitstrings in a where clause, how can I manage these different behaviors between databases since the integration tests are the same?

greg-rychlewski:
@Gigitsu Are you able to rebase on master and resolve the conflicts? We'll be better able to help if we can see the test results here. Thank you.

Gigitsu:
@greg-rychlewski sorry yesterday I run out of time :D rebased

josevalim (approved):

Gigitsu:
Hi @josevalim @greg-rychlewski , I've updated the Earthfile to make it compatible with ARM (Apple Silicon in my case) CPUs. However, I only have experience with Docker and haven't used Earthly before, so please take a look at it.

integration_test/support/schemas.exs
```diff
@@ -388,3 +388,19 @@ defmodule Ecto.Integration.ArrayLogging do
...
+  This module is used to test:
+
+    * Bitstring type
+
+  """
+  use Ecto.Integration.Schema
+
+  schema "bitstrings" do
+    field :bs,  :bitstring
+    field :bs_with_default, :bitstring, default: <<42::6>>
```
greg-rychlewski, line 403:
```suggestion
    field :bs_with_default, :bitstring
```
The reason to remove this is because it will make Ecto populate the value in the schema before reaching the database. So if we want to test the default functionality added to the adapter, we have to leave this blank.

Gigitsu: Right, I missed it. Does it make sense to test the `field` default value moving it to another field?

Something like this:

```elixir
  schema "bitstrings" do
    field :bs,  :bitstring
    field :bs_with_default, :bitstring
    field :bs_with_size, :bitstring, default: <<42::10>>
  end
```

or is it superfluous?

greg-rychlewski: it's alright to leave it out. it's essentially the same as adding the field yourself. so since we are testing that out already we are good

Gigitsu:
> Let’s handle it in Ecto, I think it would be wrong to say it is a binary when it is actually a bitstring.

To correctly tag bitstring literal, I have two possible ways:

1.  I'm trying to implement a function that detects a quoted bitstring like this:
```elixir
  # Tagged
  def quoted_type({:<<>>, _, expr}, _vars), do: is_quoted_binary(expr) && :binary || :bitstring
  def quoted_type({:type, _, [_, type]}, _vars), do: type

  defp is_quoted_binary(args) when is_list(args) do
    Enum.all?(args, fn
      {:"::", _, [_, size]} -> rem(size, 8) == 0
      {:"::", _, [_, {spec, _, size}]} when spec in [:size] -> rem(size, 8) == 0
      _ -> true
    end)
  end
```
but I need to know every possible size specifier that can produce a bitstring that is not a binary

2. The second approach is to use code eval as suggested by @greg-rychlewski [here](https://github.com/elixir-ecto/ecto_sql/pull/577#issuecomment-1829880008)
```elixir
  # Tagged
  def quoted_type({:<<>>, _, _} = expr, _vars), do: is_quoted_binary(expr) && :binary || :bitstring
  def quoted_type({:type, _, [_, type]}, _vars), do: type

  defp is_quoted_binary(expr), do: is_binary(Code.eval_quoted(expr))
```
 

Option 2 is easier and safer than option 1, but I'm unsure if there could be any issues when evaluating the code. Option 1, on the other hand, is harder to implement.

@josevalim  @greg-rychlewski  I would appreciate a second opinion on this :)

greg-rychlewski:
Hi @Gigitsu ,

The latest suggestion I had was to do something like this https://github.com/elixir-ecto/ecto/pull/4328#issuecomment-1950249006.

WDYT?

Gigitsu:
> The latest suggestion I had was to do something like this [#4328 (comment)](https://github.com/elixir-ecto/ecto/pull/4328#issuecomment-1950249006).

Sorry, I forgot about this :(.
So `quoted_type` would always return `:bitstring` and we will do the check a runtime. It could be the easiest and safest solution possible. It will also add a little bit of runtime complexity but I think it's not a big deal. 

I don't have a strong preference for any of the solutions, though. Let's wait for @josevalim.

Thank you @greg-rychlewski

josevalim:
This is a very cheap check, so we can check it more than once. But you are right, at compile-time we don't know the type, we need to move it to runtime. Although I would say at compile-time it should say `:binary` (and we treat bitstring as a specialization).

Gigitsu:
I've updated both PR and I think that they are ready now.

If you agree, I only have to change back the `ecto_sql` deps from my repo to the `ecto` one, ( [here](https://github.com/elixir-ecto/ecto/pull/4328/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47L26) and [here](https://github.com/elixir-ecto/ecto_sql/pull/577/files#diff-dfa6f4ed74c90e5d4fda283d547d366586e690387289bcfd473e3fa5f9ace308L79) ) before merging.

josevalim (approved):
Thank you @Gigitsu! @greg-rychlewski, please ship it if you are also happy!

greg-rychlewski:
@josevalim Sorry just wanted to confirm one thing before shipping. Did you want to send non-binary bitstring literals to the adapters as `:binary` or did you want to convert them to `:bitstring` in Ecto?

greg-rychlewski:
There's a couple reasons I prefer doing it in Ecto:

1. We don't send the wrong type to the adapter
2. With the current change in ecto sql, if a user manually tags the `:binary` type and provides a value that is not binary, it will just be interpreted as `bitstring`. But this can silently allow an error to continue. Probably they should have to be explicit to use the new type?

```elixir
defp expr(%Ecto.Query.Tagged{value: binary, type: :binary}, _sources, _query)
         when is_binary(binary) do
      ["'\\x", Base.encode16(binary, case: :lower) | "'::bytea"]
    end

    defp expr(%Ecto.Query.Tagged{value: bitstring, type: type}, _sources, _query)
         when type in ~w(bitstring binary)a and is_bitstring(bitstring) do
      bitstring_literal(bitstring)
    end
```

josevalim:
If we can do it in Ecto, it would be preferable. Perhaps during normalization?

greg-rychlewski:
We could probably do it here, yeah: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1333

Though is the reason not to do something like this because the exact type might be needed somewhere else when building the query?

```elixir
defp do_literal(value, _, current) when current in @always_tagged do
  type =  quote do: if is_binary(unquote(value)), do: :binary, else: :bitstring

  {:%, [], [Ecto.Query.Tagged, {:%{}, [], [value: value, type: type]}]}
end
```

josevalim:
The do_literal approach may work too.

greg-rychlewski:
@Gigitsu Sorry for all the back and forth. It's your choice:

1. We can merge as is and I will try making it work in ecto in another PR
2. You could try to see if the `do_literal` approach works in this PR

Gigitsu:
@greg-rychlewski no problem, I can do it in this PR. I'm just not sure which is the best approach but you know way better than me the Ecto internals so I trust your judgment.

lib/ecto/query/builder.ex
```diff
@@ -777,7 +777,7 @@ defmodule Ecto.Query.Builder do
     do: do_literal(value, expected, quoted_type(value, vars))
 
   defp do_literal(value, _, current) when current in @always_tagged,
-    do: {:%, [], [Ecto.Query.Tagged, {:%{}, [], [value: value, type: current]}]}
+    do: {:%, [], [Ecto.Query.Tagged, {:%{}, [], [value: value, type: normalize_type(value, current)]}]}
```
Gigitsu, line 780: I preferred to create a new function `normalize_type` here. This way, I can match on the `:binary` current type so if, in future, `@always_tagged` were to receive new values other than `:binary` it wouldn't silently be converted into `:bitstring` but would throw an error.

I tried to put a catch all definition of `normalize_type`, but the compiler warns me that it can never match because `normalize_type(value, :binary)` clauses completely cover the type.

greg-rychlewski: i think it's ok. if we add new types then we would see the error when it fails to normalize.

sorry just one thing. could you please do `quote do: ...` instead of the raw AST? it just makes it easier to read/remember what is happening when we have to come back to it later

Gigitsu: Changed to `quote do: ...` but I had to also change the test to assert a pattern matching instead of an equality. 

One of the last test runs failed and I don't know why. At first glance, it seems unrelated to my changes..

Earthfile
```diff
@@ -17,13 +18,13 @@ integration-test-base:
...
-        echo y | apk add --allow-untrusted msodbcsql17_17.5.2.1-1_amd64.apk mssql-tools_17.5.2.1-1_amd64.apk && \
+        curl -O https://download.microsoft.com/download/3/5/5/355d7943-a338-41a7-858d-53b259ea33f5/msodbcsql18_18.3.2.1-1_${TARGETARCH}.apk && \
+        curl -O https://download.microsoft.com/download/3/5/5/355d7943-a338-41a7-858d-53b259ea33f5/mssql-tools18_18.3.1.1-1_${TARGETARCH}.apk && \
+        echo y | apk add --allow-untrusted msodbcsql18_18.3.2.1-1_${TARGETARCH}.apk mssql-tools18_18.3.1.1-1_${TARGETARCH}.apk && \
         apk del .build-dependencies && rm -f msodbcsql*.sig mssql-tools*.apk
-    ENV PATH="/opt/mssql-tools/bin:${PATH}"
+    ENV PATH="/opt/mssql-tools18/bin:${PATH}"
 
-    GIT CLONE https://github.com/elixir-ecto/ecto_sql.git /src/ecto_sql
+    GIT CLONE --branch feature/bitstring-support https://github.com/Gigitsu/ecto_sql.git /src/ecto_sql
```
greg-rychlewski, line 27:
```suggestion
    GIT CLONE https://github.com/elixir-ecto/ecto_sql.git /src/ecto_sql
```

greg-rychlewski:
thanks @Gigitsu! I'm updated the ecto_sql pointer and will merge this. once it's merged could you please have the ecto sql PR point to the latest commit on Ecto and we will merge it.

Gigitsu:
Sure!! Thank you

Issue #4331, opened 2023-11-30
Using url in config to set database name overrides dynamic_repo config


benschenker:
### Elixir version

1.15.4

### Database and Version

PostgreSQL 16.0

### Ecto Versions

3.11.0

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.17.3

### Current behavior

When following https://underjord.io/ecto-multi-tenancy-dynamic-repos-part-2.html, setting the dynamic repo, swapping the DB name based on a customer id, I was able to get everything working in my dev environment which sets the repo default config params directly via username, password, hostname, database.
Specifically in a phoenix plug I would set RepoManager.set_customer(customer_id)
Then later on in the request handling I would use the Repo normally and it would query the correct database.

In prod, I set the repo default config via the :url option, when doing this it seems to always clobber the dynamic repo such that using Repo will always query the default database specified in the url, rather than the customer specific database.

I even updated the RepoManager code to parse the URL and extract the default values, removing the url from the config before calling start_link, to no avail

The only workaround I found was a line of code in my runtime.exs config like
```
  db_config =
    case System.get_env("DATABASE_URL") do
      url when is_binary(url) ->
        Ecto.Repo.Supervisor.parse_url(url)

      _ ->
        nil
    end

  if(db_config) do
    config :myapp, Myapp.Repo, db_config
  end
```

### Expected behavior

I'd expect that using url, to set the repo default config would end up working the same way as setting the individual values directly.

That being said, my workaround does work, just spent a whole day going a bit crazy :p

greg-rychlewski:
Hi,

If you're able to create a reproducible example we can verify for sure. But based on your link it looks like the issue is here

```elixir
def get_customer_options(customer) do
  customer_database = get_database_name(customer)
  config = Application.get_env(:my_app, MyApp.Repo)

  config
  |> Keyword.put(:name, nil)
  |> Keyword.put(:database, customer_database)
end
```

Your original config will be something like this

```elixir
config :myapp, MyApp.Repo, url: "myurl"
```

What that function does is adds `name: nil` and `database: customer_database` to it. So it ends up like this

```elixir
config :myapp, MyApp.Repo, url: "myurl", name: nil, database: customer_database
```

Now it tries to start the repo. How this works is Ecto pops the url parameter, parses it, and then merges it into the rest:

```elixir
{url, config} =  Application.get_env(:myapp, MyApp.Repo, []) |> Keyword.pop(:url)
[...]

PR #4332, opened 2023-12-01
Add doc about URL config precedence

greg-rychlewski:
Pretty sure this is the issue here: https://github.com/elixir-ecto/ecto/issues/4331.

Adding some docs to clarify the behaviour.


josevalim (approved):
Yup, you are correct!

Issue #4251, opened 2023-07-27
on_confict: :replace_all gives ERROR 42601 (syntax_error) when using fields that source existing fields


benstepp:
### Elixir version

Elixir 1.15.4, OTP 26.0.2

### Database and Version

Postgres 15.3

### Ecto Versions

ecto 3.10.3, ecto_sql 3.10.1

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.17.2

### Current behavior

When using `on_conflict: :replace_all` and using fields that `:source` other existing fields, ecto generates invalid sql. Ecto already provides some mechanisms to get around this issue, such as `{:replace_all_except, fields}`, or `{:replace, fields}` but this can become quite cumbersome with larger schemas.

Here's a test case that reproduces this scenario:

```elixir
# migration
defmodule EctoConflictTarget.Repo.Migrations.CreateUsers do
  use Ecto.Migration

  def change do
    create table(:users) do
      add(:email, :string)
      add(:full_name, :string)
      timestamps()
    end

    create unique_index(:users, [:email])
  end
end

# test case
defmodule EctoConflictTargetTest do
  use EctoConflictTarget.DataCase

  defmodule User do
    @moduledoc false

    use Ecto.Schema

    schema "users" do
      field(:email, :string)
      field(:full_name, :string)
      field(:legacy_full_name_field, :string, source: :full_name)
      timestamps()
    end
  end

  @timestamp NaiveDateTime.truncate(NaiveDateTime.utc_now(), :second)

  test "insert with on_conflict: :replace_all" do
    Repo.insert!(%User{email: "user@ecto.test"})

    assert {:ok, %User{}} =
             Repo.insert(%User{email: "user@ecto.test", full_name: "First Last"},
               conflict_target: [:email],
               on_conflict: :replace_all
             )
  end

  test "insert_all with on_conflict: :replace_all" do
    Repo.insert!(%User{email: "user@ecto.test"})

    assert {1, nil} =
             Repo.insert_all(
               User,
               [
                 %{
                   email: "user@ecto.test",
                   full_name: "First Last",
                   inserted_at: @timestamp,
                   updated_at: @timestamp
                 }
               ],
               conflict_target: [:email],
               on_conflict: :replace_all
             )
  end
end
```

```
  1) test insert with on_conflict: :replace_all (EctoConflictTargetTest)
     test/ecto_conflict_target_test.exs:48
[...]

PR #4335, opened 2023-12-12
Allow schema fields to be set to read only

greg-rychlewski:
### Motivation

The `:mode` option will allow users to choose if a field should be read only or read-write. It is basically the same as adding a `:read_only` option only having a single field will let us add write only fields in the same place later on instead of creating a new field, if we want to add it.

Read only fields are useful for a few reasons:

1. We recently added the `generated` option to migrations. This allows users to put them into their schemas and have them just work™.
2. There are some workflows that benefit from having read only fields. For example, this option issue: https://github.com/elixir-ecto/ecto/issues/4251. The user would like two copies of the same field in their schema.
3. It solves part of the problems stated in this PR: https://github.com/elixir-ecto/ecto/pull/4261. Users can create generated fields in their migrations and utilize them in much the same way that was described in that issue. There are some limitations/differences but there is some overlap there.

### Changes

There were a lot of cases to consider. I will list them below:

**Casting**

- We allow casting read only fields the same way we do for virtual fields. The user might want to work with the changeset but not persist the field.

**Insert**

- When surfacing the changes, we silently filter read only fields, same as virtual fields. This is really the only way it can work because all fields are surfaced on insert so it doesn't make sense to raise.
- We want read only fields to be returned when `returning: true` is specified
- We don't want read only fields to be updated in `on_conflict`. They will be silently ignored for `replace_all` because the user is not specifying them. For `{:replace, fields}` and conflict queries an error will be raised because the user is going out of their way to specify them.

**Insert All**

- We raise immediately if a header contains a read only field. Both for lists of maps and insert queries.
- Same behaviour for `returning` and `on conflict` applies here as it did for `insert`

**Update**

- Same as `insert`, we silently ignore read only fields in the changeset. It might make sense to raise here but we don't do so for virtual fields. If we want to change one I think there's a case to change both.
- Same behaviour as `insert` for `returning`. There is no `on_conflict` to worry about.

**Update All**

- Raise if a read only field is given to the `:update` key in the query. 

**Select**

- The read only fields should be able to be selected
- The read only fields should be part of selecting the entire struct
- The read only fields should be allowed to be used as query parameters

**Delete**

- Read only fields don't come into play here

**Delete All**

- Same behaviour as select queries

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -9,6 +9,7 @@ defmodule Ecto.Query.Planner do
   end
 
   @parent_as __MODULE__
+  @write_only_kinds [:update]
   @aggs ~w(count avg min max sum row_number rank dense_rank percent_rank cume_dist ntile lag lead first_value last_value nth_value)a
 
   @doc """
@@ -1913,8 +1914,13 @@ defmodule Ecto.Query.Planner do
   end
 
   defp type!(kind, query, expr, schema, field, allow_virtuals?) when is_atom(schema) do
+    only_writable? = kind in @write_only_kinds
+
     cond do
-      type = schema.__schema__(:type, field) ->
+      type = only_writable? && schema.__schema__(:writable_type, field) ->
+        type
+
+      type = !only_writable? && schema.__schema__(:type, field) ->
         type
 
       type = allow_virtuals? && schema.__schema__(:virtual_type, field) ->
@@ -1927,6 +1933,9 @@ defmodule Ecto.Query.Planner do
                                 "Did you mean to use `#{owner_key}`?"
           %_{} ->
             error! query, expr, "field `#{field}` in `#{kind}` is an association in schema #{inspect schema}"
+
+          _ when only_writable? ->
+            error! query, expr, "field `#{field}` in `#{kind}` is an unwritable field in schema #{inspect schema}"
 
           _ ->
             error! query, expr, "field `#{field}` in `#{kind}` is a virtual field in schema #{inspect schema}"
@@ -2027,15 +2036,6 @@ defmodule Ecto.Query.Planner do
   defp cte_fields([], [], _aliases), do: []
 
   defp assert_update!(%Ecto.Query{updates: updates} = query, operation) do
-    read_only_fields =
-      case get_source!(:updates, query, 0) do
-        {source, schema, _} when is_binary(source) and schema != nil ->
-          schema.__schema__(:read_only)
-
-        _ ->
-          %{}
-      end
-
     changes =
       Enum.reduce(updates, %{}, fn update, acc ->
         Enum.reduce(update.expr, acc, fn {_op, kw}, acc ->
@@ -2044,13 +2044,7 @@ defmodule Ecto.Query.Planner do
               error! query, "duplicate field `#{k}` for `#{operation}`"
             end
 
-            case read_only_fields do
-              %{^k => _} ->
-                error! query, "cannot update read only field `#{k}`"
-
-              _ ->
-                Map.put(acc, k, v)
-            end
+            Map.put(acc, k, v)
           end)
         end)
       end)
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -9,7 +9,6 @@ defmodule Ecto.Query.Planner do
   end
 
   @parent_as __MODULE__
-  @write_only_kinds [:update]
   @aggs ~w(count avg min max sum row_number rank dense_rank percent_rank cume_dist ntile lag lead first_value last_value nth_value)a
 
   @doc """
@@ -1727,7 +1726,7 @@ defmodule Ecto.Query.Planner do
 
       {:error, {:values, _, [types, _]}} ->
         fields = Keyword.keys(types)
-        dumper = types |> Enum.map(fn {field, type} -> {field, {field, type}} end) |> Enum.into(%{})
+        dumper = types |> Enum.map(fn {field, type} -> {field, {field, type, false}} end) |> Enum.into(%{})
         {types, fields} = select_dump(fields, dumper, ix, drop)
         {{:source, :values, nil, types}, fields}
 
@@ -1750,7 +1749,7 @@ defmodule Ecto.Query.Planner do
     |> Enum.reverse
     |> Enum.reduce({[], []}, fn
       field, {types, exprs} when is_atom(field) and not is_map_key(drop, field) ->
-        {source, type} = Map.get(dumper, field, {field, :any})
+        {source, type, _read_only?} = Map.get(dumper, field, {field, :any, false})
         {[{field, type} | types], [select_field(source, ix) | exprs]}
       _field, acc ->
         acc
@@ -1914,13 +1913,8 @@ defmodule Ecto.Query.Planner do
   end
 
   defp type!(kind, query, expr, schema, field, allow_virtuals?) when is_atom(schema) do
-    only_writable? = kind in @write_only_kinds
-
     cond do
-      type = only_writable? && schema.__schema__(:writable_type, field) ->
-        type
-
-      type = !only_writable? && schema.__schema__(:type, field) ->
+      type = schema.__schema__(:type, field) ->
         type
 
       type = allow_virtuals? && schema.__schema__(:virtual_type, field) ->
@@ -1933,9 +1927,6 @@ defmodule Ecto.Query.Planner do
                                 "Did you mean to use `#{owner_key}`?"
           %_{} ->
             error! query, expr, "field `#{field}` in `#{kind}` is an association in schema #{inspect schema}"
-
-          _ when only_writable? ->
-            error! query, expr, "field `#{field}` in `#{kind}` is an unwritable field in schema #{inspect schema}"
 
           _ ->
             error! query, expr, "field `#{field}` in `#{kind}` is a virtual field in schema #{inspect schema}"
@@ -2036,12 +2027,21 @@ defmodule Ecto.Query.Planner do
   defp cte_fields([], [], _aliases), do: []
 
   defp assert_update!(%Ecto.Query{updates: updates} = query, operation) do
+    schema = schema_for_update(query)
+    dumper = schema && schema.__schema__(:dump)
+
     changes =
       Enum.reduce(updates, %{}, fn update, acc ->
         Enum.reduce(update.expr, acc, fn {_op, kw}, acc ->
           Enum.reduce(kw, acc, fn {k, v}, acc ->
             if Map.has_key?(acc, k) do
               error! query, "duplicate field `#{k}` for `#{operation}`"
+            end
+
+            case dumper do
+              %{^k => {_, _, false}} -> :ok
+              %{}-> error! query, "cannot update unwritable field `#{k}`"
+              nil -> :ok
             end
 
             Map.put(acc, k, v)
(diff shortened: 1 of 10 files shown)


josevalim:
> having a single field will let us add write only fields in the same place later on instead of creating a new field

Do you think it makes sense to have write only fields? I am asking because we already use `read_only` elsewhere, so it would be nice to be consistent.

lib/ecto/schema.ex
```diff
@@ -525,9 +527,13 @@ defmodule Ecto.Schema do
     :type,
     :where,
     :references,
-    :skip_default_validation
+    :skip_default_validation,
+    :mode
   ]
 
+  @default_mode :readwrite
+  @allowed_modes [:readwrite, :readonly]
```
josevalim, line 535: We use `read_only` in the repo, so at least here we should be consistent and have `read_write` and `read_only`?

greg-rychlewski:
> Do you think it makes sense to have write only fields? I am asking because we already use read_only elsewhere, so it would be nice to be consistent.

I actually started with `read_only` because I couldn't think of any use case for write only. And then I started doubting myself that maybe someone will come by with a valid use case later on.

To me it's a strange idea though and I never really encountered that behaviour anywhere. I'm happy to change back to read_only if you also think it's ok.

josevalim:
The implementation looks clean and concise. The addition of `writeable_field` is excellent to encapsulate them.

For now, I would prefer to call them `read_only` for consistency with everything else.

I also wonder if there is a relationship between the following options:

1. virtual fields and read_only fields. It feels they are not related, given virtual fields are not read from repository.

2. read_only fields and read_after_writes. It feels they are orthogonal. A field can be read only and may or may not be read after writes.

 I am not saying there is, just wondering. Any other option we may have missed?

greg-rychlewski:
The other one is `:load_in_query` which makes sure that it is selected when the entire struct is selected. But similar to the other ones it doesn't quite cover the entire case. Because you can still write to these fields.

`:virtual` I don't believe is suitable here because we want `:read_only` fields to be read from the database.

The other field options are completely unrelated:

```elixir
 :default,
    :source,
    :autogenerate,
    :read_after_writes,
    :virtual,
    :primary_key,
    :load_in_query,
    :redact,
    :foreign_key,
    :on_replace,
    :defaults,
    :type,
    :where,
    :references,
    :skip_default_validation,
    :mode
```

greg-rychlewski:
Virtual is pretty close because you could select an existing database field into it if you're explicit. But you would not be able to retrieve it by selecting something like `from p in Post`. And then another issue is you could not use `:read_after_writes` on virtual fields, which is what the user here wanted to do https://github.com/elixir-ecto/ecto/issues/4251

josevalim:
Yeah, `read_only` is actually configuring writes (i.e. writes are never allowed). Both `load_in_query` and `read_after_writes` are configured the reads so they could be unified, but I don't think it would be much better. For example:

  * `read_on: [:reads, :writes]`, the default is `:reads`

There is at least the argument that `load_in_query` should have been called `read_in_query` but even that is a bit misleading (they can be read, they just are not loaded).

Similarly, we could have `write_on: [:writes]` and setting it to `write_on: []` means read only. But yeah, I don't think any of the APIs proposed in this comment are better, so I think we are good to go!

lib/ecto/query/planner.ex
```diff
@@ -2027,14 +2027,29 @@ defmodule Ecto.Query.Planner do
   defp cte_fields([], [], _aliases), do: []
 
   defp assert_update!(%Ecto.Query{updates: updates} = query, operation) do
+    read_only_fields =
+      case get_source!(:updates, query, 0) do
+        {source, schema, _} when is_binary(source) and schema != nil ->
+          schema.__schema__(:read_only)
```
josevalim, line 2033: Do we raise if the field is virtual? If we don't, should we? Maybe this should be called `non_writeable` instead and include both virtual and read only fields?

josevalim (approved):
Just one tiny comment and ship it!

lib/ecto/query/planner.ex
greg-rychlewski: It looks like it's handled when trying to get the type. The validation is shared with selects, etc.

But actually I like consolidating the two here and in other places. I will make the change it if that's alright with you

josevalim: @greg-rychlewski perhaps we could unify it by introducing `writeable_type`? This way we don't need to perform additional traversals.

greg-rychlewski: that's a really good idea. i will see what i can do

greg-rychlewski: What do you think of the latest changes?

There is one area that it is pretty difficult to get rid of the extra traversal. When using insert_all from a query: https://github.com/elixir-ecto/ecto/blob/fe2130587da04fbe1bf957d4e0c78e0511fd8516/lib/ecto/repo/schema.ex#L118

In order to remove the extra traversal you would have to detect inserts in `normalize_select` and then do the writable detection inside of `collect_fields`: https://github.com/elixir-ecto/ecto/blob/fe2130587da04fbe1bf957d4e0c78e0511fd8516/lib/ecto/query/planner.ex#L1392

I think that will get pretty messy.

lib/ecto/query/planner.ex
```diff
@@ -1913,8 +1914,13 @@ defmodule Ecto.Query.Planner do
   end
 
   defp type!(kind, query, expr, schema, field, allow_virtuals?) when is_atom(schema) do
+    only_writable? = kind in @write_only_kinds
```
josevalim, line 1917:
```suggestion
    type_key = if kind in @write_only_kinds, do: :writable_type, else: :type
```

And then pass it as first argument to the type call?

lib/ecto/repo/schema.ex
```diff
@@ -159,6 +171,10 @@ defmodule Ecto.Repo.Schema do
 
   defp init_mapper(schema, dumper, adapter, placeholder_map) do
     fn {field, value}, acc ->
+      unless writable_field?(schema, field) do
+        raise ArgumentError, "cannot give unwritable field `#{field}` to insert_all"
+      end
+
       case dumper do
         %{^field => {source, type}} ->
```
josevalim, line 179: Should we instead not include read only fields in the dumper?

lib/ecto/repo/schema.ex
```diff
@@ -117,9 +117,21 @@ defmodule Ecto.Repo.Schema do
 
     header = case query.select do
       %Ecto.Query.SelectExpr{expr: {:%{}, _ctx, args}} ->
-        Enum.map(args, &elem(&1, 0))
+        Enum.map(args, fn {field, _} ->
+          unless writable_field?(schema, field) do
```
josevalim, line 121: And perhaps we could use the dumper to check this too?

lib/ecto/repo/schema.ex
```diff
@@ -711,7 +727,15 @@ defmodule Ecto.Repo.Schema do
         {{:nothing, [], conflict_target}, []}
 
       {:replace, keys} when is_list(keys) ->
-        fields = Enum.map(keys, &field_source!(schema, &1))
+        fields =
+          Enum.map(keys, fn key ->
+            unless writable_field?(schema, key) do
+              raise ArgumentError, "cannot replace unwritable field `#{key}` in :on_conflict option"
+            end
```
josevalim, line 734: Weird question but, do we need this check? I assume we are not going to send the field anyway, so something else will crash?

lib/ecto/query/planner.ex
```diff
@@ -1913,8 +1914,13 @@ defmodule Ecto.Query.Planner do
...
+
     cond do
-      type = schema.__schema__(:type, field) ->
+      type = only_writable? && schema.__schema__(:writable_type, field) ->
+        type
+
+      type = !only_writable? && schema.__schema__(:type, field) ->
         type
 
       type = allow_virtuals? && schema.__schema__(:virtual_type, field) ->
```
josevalim, line 1926: I was wondering if we could unify this somehow, so we don't have to perform three different queries, but I have no idea. Plus `:virtual_type` is public, so it is not like we can get rid of it.

lib/ecto/repo/schema.ex
greg-rychlewski: Oh we need them in the dumper because of `select_dump` in the planner.

lib/ecto/repo/schema.ex
greg-rychlewski: I was hoping for that too but the only place it crashes is here:

```elixir
field_source!(schema, key)
```

And we'd need to support field sources for read only fields.

lib/ecto/repo/schema.ex
greg-rychlewski: We could make a `write_dumper` and use that here instead?

lib/ecto/repo/schema.ex
greg-rychlewski: The only other thing I can think is if we make `writable_field_source` similar to how we made `writable_type`.

lib/ecto/query/planner.ex
greg-rychlewski: When I was playing around with this they don't really intersect well. Virtual types are only allowed in `type/2` so that's where that restriction comes in. And the writable restriction comes in with being part of an update.

I think it is probably cleaner to put it back into assert_update? We are already traversing the update keys there https://github.com/elixir-ecto/ecto/blob/50d81d8d35317de9f5091c4839cf3775c6786db9/lib/ecto/query/planner.ex#L2029

lib/ecto/repo/schema.ex
josevalim: > Either that or I could add read_only? to the current dumper output

Good call! Let's try that, given the purpose of the dumper was to encapsulate all info in a single place.

lib/ecto/repo/schema.ex
greg-rychlewski: I was going to say maybe I can put read_only? as part of the current field_source output but it's public so I believe that's out.

greg-rychlewski:
I forgot about embeds btw. Do you think we should raise if `read_only` is given as an option to a field in an embedded schema?

I was thinking yes because we cannot really merge the new stuff into the old value. We can only do a complete overwrite.

josevalim:
>  Do you think we should raise if read_only is given as an option to a field in an embedded schema?

Yes, it is a perfectly fine starting point.

greg-rychlewski:
@josevalim thanks for all the feedback. i think this is the cleanest iteration so far. let me know what you think!

lib/ecto/query/planner.ex
```diff
@@ -2027,15 +2027,24 @@ defmodule Ecto.Query.Planner do
...
           Enum.reduce(kw, acc, fn {k, v}, acc ->
             if Map.has_key?(acc, k) do
               error! query, "duplicate field `#{k}` for `#{operation}`"
-            else
-              Map.put(acc, k, v)
             end
+
+            case dumper do
+              %{^k => {_, _, false}} -> :ok
+              %{}-> error! query, "cannot update unwritable field `#{k}`"
```
josevalim, line 2043:
```suggestion
              %{} -> error! query, "cannot update unwritable field `#{k}`"
```

lib/ecto/query/planner.ex
```diff
@@ -2027,15 +2027,24 @@ defmodule Ecto.Query.Planner do
   defp cte_fields([], [], _aliases), do: []
 
   defp assert_update!(%Ecto.Query{updates: updates} = query, operation) do
+    schema = schema_for_update(query)
```
josevalim, line 2030: Call it `dumper_for_update` and skip one level of conditional? :)

```suggestion
    dumper = dumper_for_update(query)
```

josevalim (approved):
Beauitiful, just minor nits!

PR #4339, opened 2023-12-17
Fix misinfo in Repo.start_link/1 docs

v0idpwn:
Ecto never returns `:ok` there.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -639,8 +639,8 @@ defmodule Ecto.Repo do
   @callback config() :: Keyword.t()
 
   @doc """
-  Starts any connection pooling or supervision and return `{:ok, pid}`
-  or just `:ok` if nothing needs to be done.
+  Starts any connection pooling or supervision tree required by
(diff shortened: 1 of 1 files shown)


lib/ecto/repo.ex
```diff
@@ -639,8 +639,8 @@ defmodule Ecto.Repo do
   @callback config() :: Keyword.t()
 
   @doc """
-  Starts any connection pooling or supervision and return `{:ok, pid}`
-  or just `:ok` if nothing needs to be done.
+  Starts any connection pooling or supervision tree required by
+  the Repo.
```
v0idpwn, line 643:
```suggestion
  Starts the Repo supervision tree.
```

josevalim (approved):

lib/ecto/repo.ex
v0idpwn: yay or nay, @josevalim?

josevalim: yay

Issue #4112, opened 2023-02-23
bug in has_one-through scenario: preload not set when combining atom and existing join


dirkholzapfel:
### Elixir version

Elixir 1.14.1 (compiled with Erlang/OTP 25)

### Database and Version

postgres (PostgreSQL) 14.5

### Ecto Versions

3.10.0-dev

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.16.5

### Current behavior

See failing test case in https://github.com/dirkholzapfel/ecto/tree/bug/preload

A preload is not set when mixing an atom and an existing join in the expression:
```
Repo.one(
        from(c in Comment,
          join: pp in assoc(c, :post_permalink),
          join: u in assoc(pp, :user),
          preload: [post_permalink: {pp, [user: u]}]
        )
      )
--> user is preloaded

Repo.one(
        from(c in Comment,
          join: pp in assoc(c, :post_permalink),
          join: u in assoc(pp, :user),
          preload: [post_permalink: {pp, [:post_comments_authors, user: u]}]
        )
      )
--> user is not preloaded (#Ecto.Association.NotLoaded<association :user is not loaded>)      
```

### Expected behavior

The preload should be set

greg-rychlewski:
Right now it's not possible to mix the 2 types of preloads, the "join" preloads and the "separate query" preloads. You can think of it like this:

1. Your main struct has the "join" preloads added.
2. Now Ecto looks for the "separate query" preloads and puts them on the main struct. These ones don't know about the previous preloads so they are erased.

So you'd have to make them both "separate query" preloads in your case. 

To make the compatible is pretty ugly, I think. You'd have to track all the sub-preloads of the first type and then make the second type aware of them. And you have join_through and that kind of stuff. Maybe the best we can do is make a note in the docs. But I'll see what the others say.

dirkholzapfel:
Thanks @greg-rychlewski, I understand the problem better now ❤️.
I am 👍 on raising an error.

greg-rychlewski:
I apologize I misspoke. The issue I mentioned only affects `through` associations and it is indeed a bug. You can test a non-through association like this in the same situation and it works:

```elixir
post = TestRepo.insert!(%Post{})
    user = TestRepo.insert!(%User{})
    comment = TestRepo.insert!(%Comment{post: post, author_id: user.id})
    TestRepo.insert!(%Permalink{post: post, user: user})

    query =
        from(c in Comment,
[...]

PR #4341, opened 2023-12-20
Fix through association nested preloads

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4112

I didn't add tests yet because I wanted to see what you thought about the solution presented here and the alternative solution. This one is really tricky.

## Background

When using `Repo.preload`, it is possible for through associations to lose their nested preloads like this

```elixir
  from c in Comment, 
    join: t in ThroughAssoc,
    on: c.through_id == t.id, 
    join: u in User,
    on: t.user_id == u.id, 
    preload: [through_assoc: {t, [:some_other_assoc, user: u]}]
```

What happens is this:

1. The join preloads put User into ThroughAssoc and ThroughAssoc into Comment
2. The separate query preload puts :some_other_assoc into ThroughAssoc. However, it does this by generating the association chain and then putting it into the through field. This association chain does not have User inside of ThroughAssoc so that nested preload is lost.

## Solution

The solution in this PR is pretty simple. If the through association is already loaded then we preload directly into it. The same way we do for embeds. When determining whether to put the association chain into the through field, we simply don't put the association chain into the through field if it has already been loaded. However, the chain is still calculated and put into the "regular" association fields. 

The drawback of this method is that we are adding additional queries for the loaded through associations. Taking into account nested associations, this could end up with a decent amount of extra queries.

## Alternative Solution

The alternative solution is to not preload the association chain if the through association has been loaded. I tried this out first but it gets quite complex to the point where I'm not sure there is any sane way to do it. There are two main issues:

1. Different through associations can overlap in their association chains. For example:

- posts => author => location
- posts => author => company => department
- posts => author => company => manager => director

Depending on which of these 3  through fields are already loaded, we would have to keep track of which sub-preloads at each nested layer are to be kept.

2. Building off of point (1), the chain of preloads is dependent on which struct you are looking at. Since each struct can have different fields loaded/unloaded, not every one will have the same chain of nested preloads. This goes against the current flow in the code which assumes each struct in a give layer has the same sub-preloads. 

It also means we have to carry over information several layers deep. For example, a field on `post` might control whether `manager` is preloaded. But by the time we get to manager we've lost all the post information because we preload recursively.

### DIFF
--- a/lib/ecto/repo/preloader.ex
+++ b/lib/ecto/repo/preloader.ex
@@ -65,15 +65,16 @@ defmodule Ecto.Repo.Preloader do
     if sample = Enum.find(structs, & &1) do
       module = sample.__struct__
       prefix = preload_prefix(tuplet, sample)
-      {assocs, throughs, embeds} = expand(module, preloads, {%{}, %{}, []})
+      {assocs, throughs, embeds} = expand(module, preloads, {%{}, [], []})
       structs = preload_embeds(structs, embeds, repo_name, tuplet)
+      structs = preload_throughs(structs, throughs, repo_name, tuplet)
 
       {fetched_assocs, to_fetch_queries} =
         prepare_queries(structs, module, assocs, prefix, repo_name, tuplet)
 
       fetched_queries = maybe_pmap(to_fetch_queries, repo_name, tuplet)
       assocs = preload_assocs(fetched_assocs, fetched_queries, repo_name, tuplet)
-      throughs = Map.values(throughs)
+      throughs = Enum.map(throughs, &elem(&1, 0))
 
       for struct <- structs do
         struct = Enum.reduce assocs, struct, &load_assoc/2
@@ -164,7 +165,6 @@ defmodule Ecto.Repo.Preloader do
   defp preload_embeds(structs, [], _repo_name, _tuplet), do: structs
 
   defp preload_embeds(structs, [embed | embeds], repo_name, tuplet) do
-
     {%{field: field, cardinality: card}, sub_preloads} = embed
 
     {embed_structs, counts} =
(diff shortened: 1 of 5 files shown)


josevalim:
Awesome. This looks good to me but it feels it could be slightly backwards compatible, right? Previously we were always preloading through (as it force: true) and now it make it behave like (force: false), right?

greg-rychlewski:
Yeah definitely it changes the behaviour. In addition to the change of force behaviour it also changes some of the nested preload behaviour. For example, if you had this:

```elixir
schema "posts" do
  has_many :comments_authors, through: [:comments, :authors]
  has_many :comments
end

schema "comments" do
  belongs_to :author
end

schema "authors" do
  has_one :permalink
  has_many :posts
end

post = Repo.preload(post, :comments_authors)
post = Repo.preload(post, [comments_authors: [:permalink], comments: [authors: [:posts]]])
```

In the second preload the `:comments_authors` gets expanded to its association chain and `:permalink` and `:posts` end up getting put into the authors struct. Since we were forcing an overwrite in the past, `:posts` would also be nested inside of `:comments_authors` even though it wasn't explicitly set. But with the change it won't get put there.

Now that you mention forcing, if we go ahead with this then I should also take it into account and allow the overwrite if force is true.

josevalim:
Maybe an alternative solution here is to know if the association has been loaded because it happens via Repo.preload or if it happens via query (i.e. preload+join)?

If the preload is filled because of a preload+join, then we can apply the heuristic in this patch, because the user is explicitly saying they want their given struct to win (i.e. it is force: false).

If it is a Repo.preload, then can keep the current semantics of filling in the association chain?

greg-rychlewski:
That's a really good idea. What do you think about the last commit? If you like it, I can write the integration tests.

josevalim:
Wait a second, you mentioned that in this case we will no longer have authors and posts in comments or something.

> post = Repo.preload(post, [comments_authors: [:permalink], comments: [authors: [:posts]]])

But wouldn't we be able to know if comments_authors was set before hand (before calling preload) or set by preload itself? In this case, we could:

1. If it was set before hand, we keep whatever is there, excepting forcing
2. If it is was not set, we preserve today's behaviour

Would this be able to reduce the backwards incompatible changes without adding a new option?

greg-rychlewski:
Sorry I probably am not understanding properly. Would this be the same as my original commit or something different?

josevalim:
Sorry, I am telling you my conclusions but not my thought process. There are two concerns in this PR, as far as I know:

1. force_through changing the behaviour from true to false

2. `post = Repo.preload(post, [comments_authors: [:permalink], comments: [authors: [:posts]]])` not getting expanded into the authors struct

What I am saying is that, we only need to change the behaviour of `preload: [...]` with joins in them, we can keep everything else behaving the exact same. And we can probably do that by passing some additional information here: https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/queryable.ex#L238

greg-rychlewski:
Oh I see what you mean now. There is a case to change the behaviour for `Repo.preload` as well. The reason is to keep parity with non-through preloads. For example

```elixir
post = Repo.preload(post, [non_through_assoc: [:nested_assoc]])
post = Repo.preload(post, :non_through_assoc)
```

In the case above, `post` will still keep `:nested_assoc`. But for the case below it will not because the through field is wiped by the preload chain of "regular" assocs:

```elixir
post = Repo.preload(post, [through_assoc: [:nested_assoc]])
post = Repo.preload(post, through_assoc)
```

I don't personally see a reason to expect non-through and through assocs to behave differently here. But to maintain backwards compatibility I thought it might be worth giving them an option instead of outright changing the behaviour.

greg-rychlewski:
Actually I misspoke a bit for this example:

```elixir
post = Repo.preload(post, [through_assoc: [:nested_assoc]])
post = Repo.preload(post, through_assoc)
```
This will still have `:nested_assoc`. It only won't if you reset the "regular" assoc that is part of the through assoc. So for instance if you have this

```elixir
schema "posts" do
  has_many :through_assoc, through: [:through1, through2]
  has_one :through1
end
```

Then if you do this, the nested assoc would disappear:

```elixir
post = Repo.preload(post, [through_assoc: [:nested_assoc]])
post = Ecto.reset_fields(post, [:through1])
post = Repo.preload(post, through_assoc)
```

I think what you said makes sense then. As long as you are using `Repo.preload` normally and not resetting random fields it will behave consistently. I'll change the PR.

greg-rychlewski:
WDYT?

lib/ecto/repo/preloader.ex
```diff
@@ -590,13 +641,17 @@ defmodule Ecto.Repo.Preloader do
...
-            |> Enum.reduce({fields, query, sub_preloads}, &{nil, nil, [{&1, &2}]})
-            |> elem(2)
+          if Keyword.has_key?(query_assocs, preload) do
+            {assocs, [{info, sub_preloads, true} | throughs], embeds}
+          else
+            through =
+              through
+              |> Enum.reverse()
+              |> Enum.reduce({fields, query, sub_preloads}, &{nil, nil, [{&1, &2}]})
+              |> elem(2)
```
josevalim, line 651:
```suggestion
            {_, _, through} =
              through
              |> Enum.reverse()
              |> Enum.reduce({fields, query, sub_preloads}, &{nil, nil, [{&1, &2}]})
```

lib/ecto/repo/preloader.ex
```diff
@@ -12,14 +12,14 @@ defmodule Ecto.Repo.Preloader do
...
+  @spec query([list], Ecto.Repo.t, list, Access.t, list, fun, {adapter_meta :: map, opts :: Keyword.t}) :: [list]
+  def query([], _repo_name, _preloads, _take, _assocs, _fun, _tuplet), do: []
+  def query(rows, _repo_name, [], _take, _assocs, fun, _tuplet), do: Enum.map(rows, fun)
 
-  def query(rows, repo_name, preloads, take, fun, tuplet) do
+  def query(rows, repo_name, preloads, take, assocs, fun, tuplet) do
     rows
     |> extract()
-    |> normalize_and_preload_each(repo_name, preloads, take, tuplet)
+    |> normalize_and_preload_each(repo_name, preloads, take, assocs, tuplet)
```
josevalim, line 22: Nitpick: let's normalize assocs here, because the other branches have no assocs, so they don't need to call assocs normalization code.

Also, what if we normalize assocs to a map? We are using Keyword.get and has_key?, and those should be more efficient for maps?

lib/ecto/repo/preloader.ex
```diff
@@ -176,23 +178,57 @@ defmodule Ecto.Repo.Preloader do
...
-    do: load_embeds(card, field, structs, embed_structs, counts, [struct | acc])
+  defp preload_throughs(
+         structs,
+         [{_, _, false = _from_query?} | throughs],
+         repo_name,
+         query_assocs,
+         tuplet
+       ) do
+    # We only preload through associations directly if they were previously loaded by a query.
+    # Otherwise, we preload the association chain.
```
josevalim, line 198:
```suggestion
    # We only preload through associations directly if they were previously loaded by a query.
    # Otherwise, we preload the association chain.
```

Can we clarify here? Because we have been struggling with nomenclature, so being more concrete will help. Let's use `Ecto.Query.preload` and `Ecto.Repo.preload`?

josevalim (approved):
Looks great, I have only some minor nits.

To make sure we are on the same page. This should be fully backwards compatible. The only chance in behaviour is when you preload a through with joins and regular assocs, which was broken anyway (i.e. it was the original bug report)?

greg-rychlewski:
> The only chance in behaviour is when you preload a through with joins and regular assocs, which was broken anyway (i.e. it was the original bug report)?

Yeah exactly. The following are backwards compatible:

- Everything with `Repo.preload`
- Everything with `Ecto.Query.preload` not involving throughs
- Everything with `Ecto.Query.preload` involving throughs not used in joins
- Everything with `Ecto.Query.preload` involving throughs used in joins as long as they don't have separate query preloads

The following is not backwards compatible:

- `Ecto.Query.preload` when a through association is used in a join and has a nested separate query preload

In this case, the differences are:

1) the association chain is no longer preloaded and we simply preload directly onto the loaded through association. 
2) as an additional consequence of 1, if the user goes out of their way to specify both the through association and the individual through components in the preload list,there won't be any merging of the sub-preloads here: https://github.com/elixir-ecto/ecto/blob/1e9b87bb08b579ee1e3932c239853f57bc79e5aa/lib/ecto/repo/preloader.ex#L673

Before I merge, could you confirm ^ is ok with you as well? Thanks!

josevalim:
Yup, it is all good to me. Beautifully done!

Issue #4311, opened 2023-11-07
Incompatible behavior of selecting many_to_many relation


v1d3rm3:
### Elixir version

1.14.4

### Database and Version

PostgresSQL 15

### Ecto Versions

3.10.3

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.17.3

### Current behavior

When defining many to many relations in Ecto, the relation is using module name of relations in join relation. For example, if i've these definitions:

```elixir
defmodule EctoAssocRepro.UserSchema do
  alias EctoAssocRepro.OrganizationSchema
  alias EctoAssocRepro.UserOrganizationSchema
  use Ecto.Schema

  schema "users" do
    # many_to_many(:organizations, OrganizationSchema, join_through: UserOrganizationSchema, join_keys: [user_id: :id, organization_id: :id]) # IT WORKS
    many_to_many(:organizations, OrganizationSchema, join_through: UserOrganizationSchema)
    belongs_to :team, EctoAssocRepro.TeamSchema
  end
end

defmodule EctoAssocRepro.OrganizationSchema do
  alias EctoAssocRepro.UserSchema
  alias EctoAssocRepro.UserOrganizationSchema
  use Ecto.Schema

  schema "organizations" do
    many_to_many(:users, UserSchema, join_through: UserOrganizationSchema)
  end
end

defmodule EctoAssocRepro.UserOrganizationSchema do
  alias EctoAssocRepro.OrganizationSchema
  alias EctoAssocRepro.UserSchema
  use Ecto.Schema

  @primary_key false
  schema "users_organizations" do
    belongs_to(:user, UserSchema)
    belongs_to(:organization, OrganizationSchema)
    timestamps()
  end
end
```

and try to query relationship:

```bash
import Ecto.Query
EctoAssocRepro.Repo.all(from u in EctoAssocRepro.UserSchema, preload: [:organizations])
```

It gives me this error:

```bash 
** (Ecto.QueryError) deps/ecto/lib/ecto/query/planner.ex:963: field `user_schema_id` in `select` does not exist in schema EctoAssocRepro.UserOrganizationSchema in query:

from o0 in EctoAssocRepro.OrganizationSchema,
  join: u1 in EctoAssocRepro.UserOrganizationSchema,
  on: o0.id == u1.organization_schema_id,
  where: u1.user_schema_id in ^[3, 2, 1],
  order_by: [asc: type(u1.user_schema_id, :id)],
[...]

PR #4342, opened 2023-12-21
clarify what is called "schema name"

v1d3rm3:
As explicited in #4311, a little reinforcement of the meaning "schema name"

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -35,6 +35,8 @@ defmodule Ecto.Schema do
         end
       end
 
+
+
   By default, a schema will automatically generate a primary key which is named
   `id` and of type `:integer`. The `field` macro defines a field in the schema
   with given name and type. `has_many` associates many posts with the user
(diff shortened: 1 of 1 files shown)


lib/ecto/schema.ex
```diff
@@ -47,9 +47,11 @@ defmodule Ecto.Schema do
...
   `Ecto.Changeset` module.
 
-  Note that the name of the database table does not need to correlate to your
-  module name.  For example, if you are working with a legacy database, you can
-  reference the table name when you define your schema:
+  Note that what we call "schema name" is actually the module name (e.g. `User`).
+  The first argument of `schema/2` is the name of database's table, which does
+  not need to correlate to your module name (schema name).  For example, if you
+  are working with a legacy database, you can reference the table name when
+  you define your schema:
```
greg-rychlewski, line 54:
```suggestion
  The first argument of `schema/2` is the name of database's table, which does
  not need to correlate to your module name (commonly referred to as the schema/schema name).
  For example, if you are working with a legacy database, you can reference the table name
  (`legacy_users`) when you define your schema (`User`):
```

greg-rychlewski:
Thank you!

PR #4348, opened 2024-01-10
Fix compilation error when using prepend_order_by

AdiletAbylov:
If you are trying to use `Ecto.Query.prepend_order_by/3` it fails during compilation with 
```
** (Ecto.Query.CompileError) unsupported :prepend_order_by in keyword query expression
    (ecto 3.11.1) expanding macro: Ecto.Query.from/2
```
This is happening because `prepend_order_by` is not added to the `from` building function guard.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -1103,7 +1103,7 @@ defmodule Ecto.Query do
 
   @from_join_opts [:as, :prefix, :hints]
   @no_binds [:union, :union_all, :except, :except_all, :intersect, :intersect_all]
-  @binds [:lock, :where, :or_where, :select, :distinct, :order_by, :group_by, :windows] ++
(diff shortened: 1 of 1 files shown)


greg-rychlewski:
Thank you for the PR. Could you please also add a regression test?

AdiletAbylov:
Will do it a bit later

josevalim:
I am not sure I want to support prepend_order_by in query. I think it is fine as a regular operation but in queries it feels very out of place. :(

greg-rychlewski:
oh yeah you're right. i forgot we discussed that =(

AdiletAbylov:
So, do you still need test from my side?

greg-rychlewski:
It's ok, Jose is right.

greg-rychlewski:
If there is a good place in the docs to make it clearer and you'd like to submit an improvement, that would be appreciated.

AdiletAbylov:
Ok, I am closing this PR and will open another one for docs addition

PR #4350, opened 2024-01-18
Clarify order on coalesce functionality

gosimitz:


lib/ecto/query/api.ex
```diff
@@ -265,7 +265,8 @@ defmodule Ecto.Query.API do
   def count(value, :distinct), do: doc! [value, :distinct]
 
   @doc """
-  Takes whichever value is not null, or null if they both are.
+  If the first value is not null, it takes that. Otherwise it
+  takes the second value or null if they both are.
```
josevalim:
```suggestion
  Takes the first value which is not null, or null if they both are.
```

josevalim:

PR #4355, opened 2024-01-22
Fix syntax in example

Adzz:
As written the example is not valid elixir because the double quotes need escaping. But in postgres at least double quotes refer to columns whereas here we want an empty string.


josevalim:

PR #4358, opened 2024-01-28
Introduce is_named_binding guard

greg-rychlewski:
What do you think about adding this as a convenience? i.e. instead of

```elixir
def modify_query(query) do
  if has_named_binding?(query, :name) do
    ... do something ...
  else
    ... do something else ...
  end
end
```

you can do

```elixir
def modify_query(query) when is_named_binding(query, :name) do
  ... do something ...
end

def modify_query(query) do
  .. do something else...
end
```

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -529,6 +529,15 @@ defmodule Ecto.Query do
     end
   end
 
+  @doc """
+  Returns `true` if `query` has named binding `name`; otherwise returns `false`.
+
+  For more information on named bindings see "Named bindings" in this module doc.
+  """
+  @doc guard: true
+  defguard is_named_binding(query, name)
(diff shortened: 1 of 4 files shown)


lib/ecto/query.ex
```diff
@@ -529,6 +529,15 @@ defmodule Ecto.Query do
     end
   end
 
+  @doc """
+  Returns `true` if `query` has named binding `name`; otherwise returns `false`.
+
+  For more information on named bindings see "Named bindings" in this module doc.
```
josevalim, line 535: Let's say it is the same as `has_named_binding?` but allowed in guards. I also don't think we need `@doc guard: true` (Elixir uses it for docs but it is not required).

greg-rychlewski: done!

josevalim (approved):

PR #4359, opened 2024-01-28
Fix attach_prefix

greg-rychlewski:
It looks like this broke in one of the recent commits

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,4 +1,4 @@
-VERSION 0.5
+VERSION 0.6
 
 all:
     BUILD \
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1811,8 +1811,8 @@ defmodule Ecto.Query.Planner do
   """
   def attach_prefix(%{prefix: nil} = query, opts) when is_list(opts) do
     case Keyword.fetch(opts, :prefix) do
-      {:ok, prefix} when is_binary(prefix) -> %{query | prefix: prefix}
-      {:ok, prefix} -> raise ArgumentError, ":prefix must be a binary, got: #{inspect(prefix)}"
+      {:ok, prefix} when is_binary(prefix) or is_nil(prefix) -> %{query | prefix: prefix}
+      {:ok, prefix} -> raise ArgumentError, ":prefix must be a binary or `nil`, got: #{inspect(prefix)}"
       :error -> query
     end
   end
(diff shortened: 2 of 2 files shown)


josevalim (approved):

josevalim:
Thank you for cleaning up my mess.

greg-rychlewski:
It seems like some migration tests are failing due to using atom prefixes. And the type specs are saying they allow atom:https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/migration.ex#L485

Maybe it is ok for prefix to be an atom? From the stacktrace it seems like it's raising while checking the completed migrations

```elixir
(ecto 3.11.1) lib/ecto/query/planner.ex:1815: Ecto.Query.Planner.attach_prefix/2
   +integration-test *failed* |        (ecto 3.11.1) lib/ecto/repo/queryable.ex:211: Ecto.Repo.Queryable.execute/4
   +integration-test *failed* |        (ecto 3.11.1) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
   +integration-test *failed* |        (ecto_sql 3.11.1) lib/ecto/migrator.ex:586: anonymous fn/5 in Ecto.Migrator.lock_for_migrations/4
   +integration-test *failed* |        (ecto_sql 3.11.1) lib/ecto/migrator.ex:585: Ecto.Migrator.lock_for_migrations/4
   +integration-test *failed* |        test/ecto/migrator_test.exs:262: anonymous fn/0 in Ecto.MigratorTest."test migrator prefix"/1
   +integration-test *failed* |        (ex_unit 1.15.6) lib/ex_unit/capture_log.ex:113: ExUnit.CaptureLog.with_log/2
   +integration-test *failed* |        (ex_unit 1.15.6) lib/ex_unit/capture_log.ex:75: ExUnit.CaptureLog.capture_log/2
   +integration-test *failed* |        test/ecto/migrator_test.exs:261: (test)
```

So probably atom prefixes work with regular queries?

josevalim:
The query layer requires them to be strings at compile-time and I think normalizing them to strings is a good choice. So I would change migrations typespecs to force them to be strings and add a deprecation warning to attach_prefix if they are atoms. Thoughts?

greg-rychlewski:
> The query layer requires them to be strings at compile-time 

Oh I see what you mean now. What you are saying makes sense to me.

> add a deprecation warning to attach_prefix if they are atoms

Do you mean something like this:

```elixir
def attach_prefix(%{prefix: nil} = query, opts) when is_list(opts) do
      {:ok, prefix} when is_binary(prefix) or is_nil(prefix) ->
        %{query | prefix: prefix}
      
       {:ok, prefix} when is_atom(prefix)->
         Logger.warning("atom prefixes are deprecated")
         %{query | prefix: prefix}
      
      :error ->
        query
    end
  end
```

josevalim:
Yes, but using IO.warn since it can be treated as an error during compilation if using —warnings-as-errors

greg-rychlewski:
done!

josevalim (approved):

PR #4364, opened 2024-02-03
Raise if user tries to query embedded schema

greg-rychlewski:
There are some edge cases around embedded schemas that I think can be cleaned up a little. This one happens when you try to query an embedded schema. Currently, you'll get  an error like this:

> (Postgrex.Error) ERROR 42P01 (undefined_table) relation "nil" does not exist

With this change you'll get this

>  ** (Protocol.UndefinedError) protocol Ecto.Queryable not implemented for Ecto.Integration.Item of type Atom, the given module is an embedded schema. This protocol is implemented for the following type(s): Atom, BitString, Ecto.Query, Ecto.SubQuery, Tuple

### DIFF
--- a/lib/ecto/queryable.ex
+++ b/lib/ecto/queryable.ex
@@ -33,12 +33,15 @@ defimpl Ecto.Queryable, for: Atom do
       UndefinedFunctionError ->
         message =
           if :code.is_loaded(module) do
-            "the given module does not provide a schema"
+            "the given module either does not provide a schema"
           else
             "the given module does not exist"
           end
 
         raise Protocol.UndefinedError, protocol: @protocol, value: module, description: message
+
+      FunctionClauseError ->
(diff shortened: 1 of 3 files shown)


lib/ecto/queryable.ex
```diff
@@ -33,12 +33,15 @@ defimpl Ecto.Queryable, for: Atom do
       UndefinedFunctionError ->
         message =
           if :code.is_loaded(module) do
-            "the given module does not provide a schema"
+            "the given module either does not provide a schema"
```
greg-rychlewski, line 36:
```suggestion
            "the given module does not provide a schema"
```

greg-rychlewski:
I think this makes sense, but I might be missing something. Like if there is a special case for MongoDB or something like that.

v0idpwn (approved):
Beautiful

v0idpwn:
I don't think there's any case you could use this before. But I may also be missing something 🤔

josevalim (approved):

PR #4372, opened 2024-02-13
Fix update_change/3 function documentation

codeanpeace:
Inline function document references put_changed/3 rather than put_change/3


codeanpeace:
Nevermind, already fixed on master

PR #4373, opened 2024-02-13
docs: fix single query preloading in associations cheatsheet

sheerlox:
Based on my testing and the [`preload/3` documentation](https://hexdocs.pm/ecto/Ecto.Query.html#preload/3), the associations cheatsheet contains invalid instructions for using join queries to preload associations in one query.


greg-rychlewski:
Thank you

Issue #4375, opened 2024-02-13
query: aliases are not recognized within `splice/1`


Gladear:
### Elixir version

Elixir 1.15.6 (compiled with Erlang/OTP 26)

### Database and Version

N/A

### Ecto Versions

3.11.1

### Database Adapter and Versions (postgrex, myxql, etc)

N/A

### Current behavior

As far as I see, it's not possible to use "aliases" within `splice/1`
```ex
iex> Mix.install(ecto: "3.11.1")
iex> import Ecto.Query
iex> from seq in "pg_class",
...>     where: seq.relkind == "S",
...>     select: %{
...>       seq_name: fragment("concat_ws(?, ?)", ".", splice(^["public", seq.relname]))
...>     }
error: undefined variable "seq"
  iex:6

** (CompileError) cannot compile code (errors have been logged)
```

outputs
```
error: undefined variable "seq"
  iex:6

** (CompileError) cannot compile code (errors have been logged)
```

Obviously, this is a minimal reproduction. My goal would be to define a macro for `concat_ws`

```ex
  defmacro concat_ws(separator, expressions) do
    quote bind_quoted: [separator: separator, expressions: expressions] do
      fragment("concat_ws(?, ?)", separator, splice(^expressions))
    end
  end
```

Then use it like this

```ex
select: %{
  seq_name: concat_ws(".", ["public", seq.relname])
}
```

### Expected behavior

I expect aliases to be recognized within `splice/1`

josevalim:
I don't think this can work indeed. We should probably document that we can only splice runtime values.

greg-rychlewski:
If there are no conflicts in column names you could potentially do this

```elixir
seq_name: fragment("concat_ws(?, ?)", ".", splice(^["public", "relname"]))
```

LostKobrakai:
Could aliases be made to work by supporting dynamic within the spliced list?

PR #4376, opened 2024-02-13
Document fragment splice behaviour

greg-rychlewski:
I think we can solve https://github.com/elixir-ecto/ecto/issues/4375 this way. WDYT?

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -503,7 +503,7 @@ defmodule Ecto.Query.API do
   You may only splice runtime values. For example, this would not work because
   query bindings are compile-time constructs:
 
-      from p in Post, where: fragment("concat(?)", p.id, splice(^[p.count, " ", "count"]))
+      from p in Post, where: fragment("concat(?)", splice(^[p.count, " ", "count"])
   """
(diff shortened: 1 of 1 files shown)


josevalim:
If it is a literal list, then why not use fragment? The `splice` is about when the list is unknown at compile-time, no?

greg-rychlewski:
Bad brain fart by me =(. I got caught up in the specific example in the issue. I just documented the behaviour instead.

lib/ecto/query/api.ex
```diff
@@ -499,6 +499,11 @@ defmodule Ecto.Query.API do
   The example above will be transformed at runtime into the following:
 
       from p in Post, where: fragment("? in (?,?,?)", p.id, ^1, ^2, ^3)
+
+  You may only splice runtime values. For example, this would not work because
+  query bindings are compile-time constructs:
+
+      from p in Post, where: fragment("concat(?)", p.id, splice(^[p.count, " ", "count"]))
```
greg-rychlewski, line 506:
```suggestion
      from p in Post, where: fragment("concat(?)", splice(^[p.count, " ", "count"]))
```

josevalim (approved):

Issue #3387, opened 2020-08-17
Multi tenancy app not working with cte queries


alexcassol:
### Environment

* Elixir version (elixir -v): **Elixir 1.10.4 (compiled with Erlang/OTP 23)**
* Database and version (PostgreSQL 9.4, MongoDB 3.2, etc.): **PostgreSQL 10.14**
* Ecto version (mix deps): **3.4.6**
* Database adapter and version (mix deps): **postgrex 0.15.5**
* Operating system: **Fedora 32 (Workstation Edition)**

### Current behavior
I have a multi tenancy app that uses recursive queries. 

The problem is that when I give a prefix to the query through `Repo.all(prefix: "tenant_x")`, all the tables and the CTE query uses the prefix.

**Example:**
```elixir
category_tree_initial_query =
  Category
  |> where([c], is_nil(c.parent_id))

category_tree_recursion_query =
  Category
  |> join(:inner, [c], ct in "category_tree", on: c.parent_id == ct.id)

category_tree_query =
  category_tree_initial_query
  |> union_all(^category_tree_recursion_query)

{"category_tree", Category}
|> recursive_ctes(true)
|> with_cte("category_tree", as: ^category_tree_query)
|> select([c], c)
|> Repo.all(prefix: "tenant_x")

```
This SQL is generated:

```sql
WITH RECURSIVE "category_tree" AS (

    SELECT * 
    FROM "tenant_x"."categories" 
    WHERE c.parent_id IS NULL

    UNION ALL (
       SELECT * 
        FROM "tenant_x"."categories" AS c 
        INNER JOIN "tenant_x"."category_tree" AS ct ON c.parent_id == ct.id)
)
SELECT *
FROM "tenant_x"."category_tree" 

```
CTE queries shouldn't have the prefix.

### Expected behavior

I expected that prefix is used only in tables, not CTE query.

```sql
WITH RECURSIVE "category_tree" AS (

    SELECT * 
    FROM "tenant_x"."categories" 
    WHERE c.parent_id IS NULL

    UNION ALL (
       SELECT * 
        FROM "tenant_x"."categories" AS c 
        INNER JOIN "category_tree" AS ct ON c.parent_id == ct.id)
)
SELECT *
FROM "category_tree" 

```

josevalim:
This is by design. You can change this by explicitly setting a prefix in the cte query:

    |> with_cte("category_tree", as: ^%{category_tree_query | prefix: "another one"})

[...]

PR #4377, opened 2024-02-15
Ignore query prefix for CTE sources

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/3387

I tried to figure out the exact scoping rules for the CTE names by playing around with Postgres and MySQL. For example these queries:

```sql
with 
     test as (select 1), 
     test2 as (select 2), 
     test3 as (with test as (select * from test2) select * from test) 
select * from test3;
 ?column? 
----------
        2
```

From what I see, it's basically the same as variable name scoping in regular programming languages. You can see CTE names at your own level or an outer level as long as it has been defined before it gets to you.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -249,27 +249,30 @@ defmodule Ecto.Query.Planner do
     {%{expr | source: subquery}, subquery}
   end
 
-  defp plan_source(query, %{source: {nil, schema}} = expr, _adapter, _cte_names)
+  defp plan_source(query, %{source: {nil, schema}} = expr, _adapter, cte_names)
        when is_atom(schema) and schema != nil do
     source = schema.__schema__(:source)
-    prefix = plan_source_schema_prefix(expr, schema) || query.prefix
+    source_prefix = plan_source_schema_prefix(expr, schema)
+
+    prefix =
+      case cte_names do
+        %{^source => _} -> source_prefix
+        _ -> source_prefix|| query.prefix
+      end
+
     {%{expr | source: {source, schema}}, {source, schema, prefix}}
   end
 
-  defp plan_source(query, %{source: {source, nil}, prefix: prefix} = expr, _adapter, cte_names)
-       when is_binary(source) do
+  defp plan_source(query, %{source: {source, schema}, prefix: prefix} = expr, _adapter, cte_names)
+       when is_binary(source) and is_atom(schema) do
     prefix =
       case cte_names do
         %{^source => _} -> prefix
         _ -> prefix || query.prefix
       end
 
-    {expr, {source, nil, prefix}}
+    {expr, {source, schema, prefix}}
   end
-
-  defp plan_source(query, %{source: {source, schema}, prefix: prefix} = expr, _adapter, _cte_names)
-       when is_binary(source) and is_atom(schema),
-       do: {expr, {source, schema, prefix || query.prefix}}
 
   defp plan_source(_query, %{source: {kind, _, _} = source, prefix: nil} = expr, _adapter, _cte_names)
        when kind in [:fragment, :values],
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -197,15 +197,16 @@ defmodule Ecto.Query.Planner do
   This function is called by the backend before invoking
   any cache mechanism.
   """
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
I have a mistake here where I need to allow `{source, Schema}` to have its prefixed wiped as well when it's used outside the CTE. Trying to fix.

josevalim (approved):
Beautifully done, thank you :heart:

greg-rychlewski:
There are some complications I realized I didn't take care of. Some of them seem like they might not have an answer but still looking into it. I was wondering what you thought about these

**within a CTE**
- we have to be able to wipe prefixes because CTEs can be recursive and can also reference other CTEs
- there are 3 different ways the CTE source name come up:
    - `from c in "cte_name"`, in which case it seems reasonable to assume they want the prefix wiped
    - `from s in Schema` where there source of `Schema` is `cte_name`, in which case it seems reasonable to assume they don't want the prefix wiped
    - `from c in {"cte_name", Schema}`, in which case it's kind of ambiguous. it doesn't really make sense to do this inside of a cte so the intentions are a bit unclear

Outside the schema we treat a `from` queryable like `{"cte_name", Schema}` as wanting to use the CTE. So if we want to be consistent with that we'd wipe the prefix. But then it is indistinguishable from `from s in Schema` inside of the planner which shouldn't be wiped.

**outside a CTE**
- we have to be able to wipe prefixes from `{"cte_name", Schema}` when it is used in `from` because that is how you can make your CTE your `from` queryable. but it is indistinguishable from `from s in Schema` where the source of `Schema` is `cte_name` once you get to the planner. and the latter i think would be surprising to have its prefix wiped
- you can run into a similar issue with joins like this `join(:inner, [c, c1, c2], c3 in ^from(c in Comment), on: true)` but if we could solve the `from` issue we could potentially limit the wipes to `from` and not wipe in joins unless they are strings. though if they specify a join like `join(:inner, [c, c1, c2], c3 in {"cte_name", Schema}, on: true)` it is a bit ambiguous what they want/would slightly lean towards wiping for consistency with `from`

josevalim:
I wouldn't do it based on which syntax you use, it makes it too complex. If CTEs can be recursive, then we should probably apply the rule of ignoring the prefix unless one is given, even inside CTEs, and we take that as the general rule everywhere. WDYT?

greg-rychlewski:
> I wouldn't do it based on which syntax you use, it makes it too complex.

I agree.

Just want to make sure I was clear. For example, let's see say we have this schema:

```elixir
defmodule Schema do
  use Ecto.Schema

  schema "cte_name" do
     ...
  end
end
```

and then we write this query:

``` elixir
from(s in Schema)
|> with_cte("cte_name", as: ^from(...)
|> join(:inner, [_], c in "cte_name", on: true)
```

Then this would treat `from(s in Schema)` as if it were the cte as well. Which seems unexpected t ome.

josevalim:
But that's how it behaves today anyway. It may be counter intuitive but that's how it behaves in the database too, it is as if you rewrite locally what is the meaning of a table.

josevalim:
You can also argue the opposite. A CTE “overrides” all table names. However, if you use a prefix, the CTE lose this property, which is confusing. We are rather making it consistent.

greg-rychlewski:
Ah that's a really good point. I didn't think of that. 

Just want to let it sit for a little bit but probably I push my change later today.  Right now I am only erasing nil schemas and need to do it for all.

greg-rychlewski:
@josevalim I pushed a change so that it will erase the prefix for sources with schema as well. When you have time could you look it over once more?

Thanks and sorry for all the back and forth.

josevalim:
No need to apologize, the double and triple checks are deeply appreciated.

lib/ecto/query/planner.ex
```diff
@@ -216,58 +217,72 @@ defmodule Ecto.Query.Planner do
...
+  defp plan_source(query, %{source: {nil, schema}} = expr, _adapter, cte_names)
        when is_atom(schema) and schema != nil do
     source = schema.__schema__(:source)
-    prefix = plan_source_schema_prefix(expr, schema) || query.prefix
+    source_prefix = plan_source_schema_prefix(expr, schema)
+
+    prefix =
+      case cte_names do
+        %{^source => _} -> source_prefix
+        _ -> source_prefix|| query.prefix
```
josevalim, line 260:
```suggestion
        _ -> source_prefix || query.prefix
```

josevalim (approved):

PR #4382, opened 2024-02-22
Bubble up custom cast errors of the inner type for `{:map, type}` and `{:array, type}`

satom99:
Take the following custom type as an example:
```elixir
  defmodule CustomWithCastError do
    use Ecto.Type

    def type,      do: :any
    def load(_),   do: {:ok, :load}
    def dump(_),   do: {:ok, :dump}
    def cast(_),   do: {:error, foo: :bar}
  end
```
When we attempt to cast this type wrapped in an `array` or a `map`, the current behaviour in Ecto is that we get:
```elixir
    assert cast({:array, CustomWithCastError}, ["foo", "bar", "baz"]) == :error

    assert cast({:map, CustomWithCastError}, %{"x" => "foo", "y" => "bar"}) == :error
```
Where as you can see, the custom errors returned by our custom type are lost. That, and we do not know what item of the array or the map made the cast fail. This PR aims to improve on both of these things.

With the changes in this PR, the behaviour above changes to:
```elixir
    assert cast({:array, CustomWithCastError}, ["foo", "bar", "baz"]) == {:error, [index: 0, foo: :bar]}

    assert cast({:map, CustomWithCastError}, %{"x" => "foo", "y" => "bar"}) == {:error, [key: "x", foo: :bar]}
```
So we do return the custom error, and we add a reference that indicates what item made the cast fail. We include an `index` integer for the array, and the `key` for the map.

### DIFF
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -1369,7 +1369,7 @@ defmodule Ecto.Type do
         :error
 
       {:error, custom_errors} ->
-        {:error, [{:index, index} | custom_errors]}
+        {:error, Keyword.update(custom_errors, :source, [index], &[index | &1])}
     end
   end
 
@@ -1402,7 +1402,7 @@ defmodule Ecto.Type do
         :error
 
       {:error, custom_errors} ->
-        {:error, [{:key, key} | custom_errors]}
+        {:error, Keyword.update(custom_errors, :source, [key], &[key | &1])}
     end
   end
 
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -539,8 +539,10 @@ defmodule Ecto.Type do
     end
   end
 
-  def dump({:array, {_, _, _} = type}, value, dumper), do: array(value, type, dumper, false, [])
-  def dump({:array, type}, value, dumper), do: array(value, type, dumper, true, [])
+  def dump({:array, {_, _, _} = type}, value, dumper),
+    do: array_with_type(value, type, dumper, false, [])
+
+  def dump({:array, type}, value, dumper), do: array_with_type(value, type, dumper, true, [])
   def dump({:map, type}, value, dumper), do: map(value, type, dumper, false, %{})
 
   def dump(:any, value, _dumper), do: {:ok, value}
@@ -633,8 +635,10 @@ defmodule Ecto.Type do
     end
   end
 
-  def load({:array, {_, _, _} = type}, value, loader), do: array(value, type, loader, false, [])
-  def load({:array, type}, value, loader), do: array(value, type, loader, true, [])
+  def load({:array, {_, _, _} = type}, value, loader),
+    do: array_with_type(value, type, loader, false, [])
+
+  def load({:array, type}, value, loader), do: array_with_type(value, type, loader, true, [])
   def load({:map, type}, value, loader), do: map(value, type, loader, false, %{})
 
   def load(:any, value, _loader), do: {:ok, value}
@@ -831,12 +835,12 @@ defmodule Ecto.Type do
 
   defp cast_fun({:array, {:parameterized, _, _} = type}) do
     fun = cast_fun(type)
-    &array(&1, fun, false, [])
+    &array_with_index(&1, fun, false, 0, [])
   end
 
   defp cast_fun({:array, type}) do
     fun = cast_fun(type)
-    &array(&1, fun, true, [])
+    &array_with_index(&1, fun, true, 0, [])
   end
 
   defp cast_fun({:map, {:parameterized, _, _} = type}) do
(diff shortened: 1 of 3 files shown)


lib/ecto/type.ex
```diff
@@ -1352,19 +1352,24 @@ defmodule Ecto.Type do
   defp of_base_type?(:date, value), do: Kernel.match?(%Date{}, value)
   defp of_base_type?(_, _), do: false
 
-  defp array([nil | t], fun, true, acc) do
-    array(t, fun, true, [nil | acc])
+  defp array([nil | t], fun, true, {index, acc}) do
```
josevalim, line 1355: Instead of a tuple, please pass index and acc as separate arguments, it will allocate fewer "objects".

lib/ecto/type.ex
```diff
@@ -1386,9 +1391,14 @@ defmodule Ecto.Type do
...
-      :error -> :error
-      {:error, _custom_errors} -> :error
+      {:ok, value} ->
+        map_each(t, fun, skip_nil?, Map.put(acc, key, value))
+
+      :error ->
+        :error
+
+      {:error, custom_errors} ->
+        {:error, Keyword.put(custom_errors, :key, key)}
```
josevalim, line 1401:
```suggestion
        {:error, [key: key] ++ custom_errors}
```

lib/ecto/type.ex
satom99: Thank you for reviewing this! 😃

That was my first approach but we already have an `array/5` function that takes in a `type` as an extra argument — [see line 1378](https://github.com/elixir-ecto/ecto/blob/v3.11.1/lib/ecto/type.ex#L1378).

I can rename that function to `array_with_type/5` and then do `array_with_index/5` for this one. Wdyt?

lib/ecto/type.ex
```diff
@@ -1352,19 +1352,24 @@ defmodule Ecto.Type do
...
-      :error -> :error
-      {:error, _custom_errors} -> :error
+      {:ok, h} ->
+        array(t, fun, skip_nil?, {index + 1, [h | acc]})
+
+      :error ->
+        :error
+
+      {:error, custom_errors} ->
+        {:error, Keyword.put(custom_errors, :index, index)}
```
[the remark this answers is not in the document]
satom99: I suppose it's the same, but how about
```suggestion
        {:error, [{:index, index} | custom_errors]}
```
? 🤔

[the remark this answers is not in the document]
josevalim: Both are fine!

lib/ecto/type.ex
josevalim: Renaming is good!

josevalim:
I realized that, if you have an array type inside an array type, or a map of arrays, the error information won't be clear either. Thoughts? 🤔

satom99:
> I realized that, if you have an array type inside an array type, or a map of arrays, the error information won't be clear either. Thoughts? 🤔

Funnily enough, I was about to mention that too 😄 For our specific use case, we use an accumulable `:source` field like:
```elixir
assert cast({:array, {:map, CustomWithCastError}}, [%{"x" => "foo"}, %{"y" => "bar"}]) ==
             {:error, [source: [0, "x"], foo: :bar]}
```
And then in the types we do:
```elixir
      {:error, custom_errors} ->
        {:error, Keyword.update(custom_errors, :source, [index], &[index | &1])}
```
And this also plays nicely with the `:source` field that we already use for Changeset errors.

We have a polymorphic type that casts onto Ecto schemas behind the scenes. Here is an example that illustrates it:
```elixir
User -> field :posts, {:array, PolymorphicPostType} -> PremiumPostSchema.new/0 -> invalid :message field
```
and in this case we would get `source: [:posts, 0, :message]` as part of the error.

And the same applies for `:map`s. So if we had `field :posts, {:map, PolymorphicPostType}` where the keys are, say, the ID of the posts, then we would get `source: [:posts, "fa4de6b4-94f8-455b-9382-9573f4c66b6e", :message]`.

Using the `:source` key also makes sure that we keep the full path in case a Changeset is used at some point.

Thoughts? 🤔

josevalim:
Let's give it a try :) Please do the commits on top, in case we want to revert it later, we can always squash automatically before merging.

josevalim (approved):

greg-rychlewski:
@satom99 it looks like a couple changeset tests are failing. would you be able to take a look?

satom99:
> @satom99 it looks like a couple changeset tests are failing. would you be able to take a look?

On it!

satom99:
@greg-rychlewski done, please review :)

greg-rychlewski (approved):

greg-rychlewski:
thanks @satom99 ! nice change =)

PR #4397, opened 2024-03-30
Add info about redacted fields to inspect()

### LINKED ISSUE #4395: IO.inspect doesn't log Changeset.Params
### Elixir version

1.16

### Database and Version

any

### Ecto Versions

3.11.2

any

### Current behavior

```
def changeset(event, attrs) do
    event
    |> cast(attrs, [:type, :received_at, :body, :title, :from, :metadata])
    |> IO.inspect(label: "Changeset after cast")
```

Returns:

```
Changeset after cast: #Ecto.Changeset<
  action: nil,
  changes: %{title: "zzzzzzzzzz"},
  errors: [],
  data: #Events.Eventlog.Event<>,
  valid?: true
>
```

The params are missing. As a beginner first I thought functions like cast_assoc work like magic, because where do they get the data from?? 
Turns out all the data is in params, but its simply not logged. If I do `IO.inspect(changeset.params)` I get them fine.

### Expected behavior

Log the **complete** changeset struct.

I don't know if this is a bug in Ecto or in Elixir. That is for the more wise to decide.

ponychicken:
Fixes #4395 
But less verbose in the output than #4396

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -4174,9 +4174,9 @@ defimpl Inspect, for: Ecto.Changeset do
 
   def inspect(%Ecto.Changeset{data: data} = changeset, opts) do
     list =
-      for attr <- [:action, :changes, :errors, :data, :valid?] do
-        {attr, Map.get(changeset, attr)}
-      end
+      [:action, :changes, :errors, :data, :valid?]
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks, I pushed my initial suggestion instead. :)

Issue #4398, opened 2024-04-05
`Repo.load/2` doesn't cast virtual fields into a Schema struct


nicolasdabreo:
### Elixir version

Elixir 1.16.0 (compiled with Erlang/OTP 26)

### Database and Version

PostgreSQL 14

### Ecto Versions

ecto 3.11.1

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.17.4

### Current behavior

1. I perform a complex manual query to retrieve some data, this data includes additional fields relevant to my Schema in this data context.

```
query = `BIG SCARY SQL`
%{columns: columns, rows: rows} = Repo.query!(query)
```

2. I iterate over the rows to load the data into my Struct, including the additional fields as virtual fields.

```
    for row <- rows do
      casted_data = Repo.load(MyStruct, {columns, row})
      ## IO.inspect(casted_data)
      ## %MyStruct{persisted_field: "foo", virtual_field: nil}
    end
```

### Expected behavior

As the Schema as just one of the representations of my data in this context and not the actual database table itself I should be able to load not just the fields present in the database to my struct using `Repo.load/2`.

josevalim:
Thanks for the report.

Load is exclusively about loading database data and virtual fields are never sent/read from the database, so it is understandable that load won't set virtual fields, and I am not sure we should make an exception here. Can't you set the field after calling load?

nicolasdabreo:
It requires doing a number of additional enumerations over the database, converting to atom keys, zipping the cols/rows up, getting the new data and loading it back into the struct.

This is basically what we're already doing and just not working with the struct at all. I saw this as a refactoring opp to simplify, bring data safety with the struct and reduce enumerations.

josevalim:
We would be glad to consider a PR if it is a small change. But if it requires semantic or large changes, then it will likely be refused. :) Give it a try if you want to!

nicolasdabreo:
Thanks for confirming! I'll take a look this weekend :grin:

PR #4399, opened 2024-04-06
Include virtual fields in Ecto.Schema.__schema__(:load)

nicolasdabreo:
Addresses #4398

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -672,6 +672,7 @@ defmodule Ecto.Schema do
         def __schema__(:loaded), do: unquote(Macro.escape(loaded))
         def __schema__(:redact_fields), do: unquote(redacted_fields)
         def __schema__(:virtual_fields), do: unquote(Enum.map(virtual_fields, &elem(&1, 0)))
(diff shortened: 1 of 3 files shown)


josevalim:
We can't do this change, unfortunately, because it means that we will now load non-mapped fields in regular database cases. For example, imagine i have an embedded schema, with a virtual field "foo", that exists in the database too. Previously it was discarded, now it is maintained.

nicolasdabreo:
Ah I see how `__schema__.(:load)` is being used for Queryable now so this change cross contaminates it. 

I have instead moved it up to `Loader.unsafe_load/3` as this seems exclusively used for non-database uses and just for mapping afaics. I can then pull out the virtual fields and merge them into the types.

Is this getting closer to something usable?

josevalim:
I mean, it is the same issue. This function is used internally by Ecto, so any changes to it will affect existing codebases. However, I think the answer is for you to use the [public reflection API](https://hexdocs.pm/ecto/Ecto.Schema.html#module-reflection) and roll your own implementation of this. This way you don't depend on Ecto. :)

PR #4403, opened 2024-04-10
raise if empty list is given to replace

greg-rychlewski:
do you think this is worth it? it might be better to catch early and also the message might be more understandable


josevalim (approved):

wojtekmach:
@greg-rychlewski since Elixir 1.14, IO.warn accepts a single stack trace entry as a keywords list, see https://hexdocs.pm/elixir/IO.html#warn/2 for more information.

I fixed it in https://github.com/elixir-ecto/ecto_sql/commit/32113be3bab11d9f2e27832a6fb363524edbc69a.

PR #4406, opened 2024-04-23
Propagate parent process logger level to preloader

v0idpwn:
Closes https://github.com/elixir-ecto/ecto_sql/issues/602

### DIFF
--- a/lib/ecto/repo/preloader.ex
+++ b/lib/ecto/repo/preloader.ex
@@ -136,9 +136,11 @@ defmodule Ecto.Repo.Preloader do
       # still necessary.
       opts = Keyword.put_new(opts, :caller, self())
       on_preloader_spawn = Keyword.get(opts, :on_preloader_spawn, fn -> :ok end)
+      parent_logger_level = Logger.get_process_level(self())
 
       preloaders
       |> Task.async_stream(fn preloader ->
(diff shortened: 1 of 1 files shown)


v0idpwn:
Problem: `Logger.get_process_level` and `Logger.put_process_level` function were added on Elixir 1.15.

josevalim:
What if we fix the metadata initially and we address the level once we require Elixir v1.15?

v0idpwn:
I like the idea of propagating logger metadata too, certainly can add that!
With regards to the level, I assume we want to support Elixir <1.15 for at least one year more (1.15.0 is not even one year old yet), and delaying this fix until then doesn't seem to be worth it. The two variants of pmap seem relatively low cost to me. WDYT?

greg-rychlewski:
is everything inside of `Logger.get/put_process_level` also dependent on Elixir 1.15 or is there anything we can inline?

v0idpwn:
It's very Elixir-version dependent, I'd say. 

Before 1.15 there was `enable/1`, `disable/1` and `enabled?/1` for process. We could use that, but it's a more limited feature (and is now deprecated in favor of the new APIs).

v0idpwn:
That's fair enough, @josevalim. I'll close this one.

PR #4417, opened 2024-05-16
Support subqueries in order_by, group_by, distinct, and windows

zachdaniel:
So this seems to work without too much change required, but I do have one main misgiving: This requires a modification to `Ecto.QueryExpr`, which AFAIK is used in other places that won't be anticipate having subqueries in them.

I've got a test that shows this working as-is with `ecto_sql`: https://github.com/elixir-ecto/ecto_sql/pull/607

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -433,7 +433,7 @@ defmodule Ecto.Query do
 
   defmodule ByExpr do
     @moduledoc false
-    defstruct [:expr, :file, :line, params: [], subqueries: [], aliases: %{}]
+    defstruct [:expr, :file, :line, params: [], subqueries: []]
   end
 
   defmodule BooleanExpr do
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -431,6 +431,11 @@ defmodule Ecto.Query do
     defstruct [:expr, :file, :line, params: []]
   end
 
+  defmodule ByExpr do
+    @moduledoc false
+    defstruct [:expr, :file, :line, params: [], subqueries: [], aliases: %{}]
+  end
+
   defmodule BooleanExpr do
     @moduledoc false
     defstruct [:op, :expr, :file, :line, params: [], subqueries: []]
@@ -2866,7 +2871,7 @@ defmodule Ecto.Query do
     schema = assert_schema!(query)
     pks = schema.__schema__(:primary_key)
     expr = for pk <- pks, do: {dir, field(0, pk)}
-    %QueryExpr{expr: expr, file: __ENV__.file, line: __ENV__.line}
+    %ByExpr{expr: expr, file: __ENV__.file, line: __ENV__.line}
   end
 
   defp assert_schema!(%{from: %Ecto.Query.FromExpr{source: {_source, schema}}})
--- a/lib/ecto/query/builder/distinct.ex
+++ b/lib/ecto/query/builder/distinct.ex
@@ -30,11 +30,21 @@ defmodule Ecto.Query.Builder.Distinct do
   Called at runtime to verify distinct.
   """
   def distinct!(query, distinct, file, line) when is_boolean(distinct) do
-    apply(query, %Ecto.Query.QueryExpr{expr: distinct, params: [], line: line, file: file})
+    apply(query, %Ecto.Query.ByExpr{expr: distinct, params: [], line: line, file: file})
   end
   def distinct!(query, distinct, file, line) do
-    {expr, params} = Builder.OrderBy.order_by_or_distinct!(:distinct, query, distinct, [])
-    expr = %Ecto.Query.QueryExpr{expr: expr, params: Enum.reverse(params), line: line, file: file}
+    {expr, params, subqueries, aliases} =
+      Builder.OrderBy.order_by_or_distinct!(:distinct, query, distinct, [])
+
+    expr = %Ecto.Query.ByExpr{
+      expr: expr,
+      params: Enum.reverse(params),
+      line: line,
+      file: file,
+      subqueries: subqueries,
+      aliases: aliases
+    }
+
     apply(query, expr)
   end
 
@@ -54,12 +64,14 @@ defmodule Ecto.Query.Builder.Distinct do
 
   def build(query, binding, expr, env) do
     {query, binding} = Builder.escape_binding(query, binding, env)
-    {expr, {params, _acc}} = escape(expr, {[], %{}}, binding, env)
+    {expr, {params, acc}} = escape(expr, {[], %{subqueries: [], aliases: {:%{}, [], []}}}, binding, env)
     params = Builder.escape_params(params)
 
-    distinct = quote do: %Ecto.Query.QueryExpr{
+    distinct = quote do: %Ecto.Query.ByExpr{
                            expr: unquote(expr),
                            params: unquote(params),
+                           subqueries: unquote(acc[:subqueries]),
+                           aliases: unquote(acc[:aliases]),
                            file: unquote(env.file),
                            line: unquote(env.line)}
     Builder.apply_query(query, __MODULE__, [distinct], env)
--- a/lib/ecto/query/builder/distinct.ex
+++ b/lib/ecto/query/builder/distinct.ex
@@ -69,7 +69,7 @@ defmodule Ecto.Query.Builder.Distinct do
     distinct = quote do: %Ecto.Query.ByExpr{
                            expr: unquote(expr),
                            params: unquote(params),
-                           subqueries: unquote(acc[:subqueries]),
+                           subqueries: unquote(acc.subqueries),
                            file: unquote(env.file),
                            line: unquote(env.line)}
     Builder.apply_query(query, __MODULE__, [distinct], env)
--- a/lib/ecto/query/builder/distinct.ex
+++ b/lib/ecto/query/builder/distinct.ex
@@ -33,7 +33,7 @@ defmodule Ecto.Query.Builder.Distinct do
     apply(query, %Ecto.Query.ByExpr{expr: distinct, params: [], line: line, file: file})
   end
   def distinct!(query, distinct, file, line) do
-    {expr, params, subqueries, aliases} =
+    {expr, params, subqueries} =
       Builder.OrderBy.order_by_or_distinct!(:distinct, query, distinct, [])
 
     expr = %Ecto.Query.ByExpr{
@@ -41,8 +41,7 @@ defmodule Ecto.Query.Builder.Distinct do
       params: Enum.reverse(params),
       line: line,
       file: file,
-      subqueries: subqueries,
-      aliases: aliases
+      subqueries: subqueries
     }
 
     apply(query, expr)
@@ -64,14 +63,13 @@ defmodule Ecto.Query.Builder.Distinct do
 
   def build(query, binding, expr, env) do
     {query, binding} = Builder.escape_binding(query, binding, env)
-    {expr, {params, acc}} = escape(expr, {[], %{subqueries: [], aliases: {:%{}, [], []}}}, binding, env)
+    {expr, {params, acc}} = escape(expr, {[], %{subqueries: []}}, binding, env)
     params = Builder.escape_params(params)
 
     distinct = quote do: %Ecto.Query.ByExpr{
                            expr: unquote(expr),
                            params: unquote(params),
                            subqueries: unquote(acc[:subqueries]),
-                           aliases: unquote(acc[:aliases]),
                            file: unquote(env.file),
                            line: unquote(env.line)}
     Builder.apply_query(query, __MODULE__, [distinct], env)
--- a/lib/ecto/query/builder/group_by.ex
+++ b/lib/ecto/query/builder/group_by.ex
@@ -99,7 +99,7 @@ defmodule Ecto.Query.Builder.GroupBy do
     group_by = quote do: %Ecto.Query.ByExpr{
                            expr: unquote(expr),
                            params: unquote(params),
-                           subqueries: unquote(acc[:subqueries]),
+                           subqueries: unquote(acc.subqueries),
                            file: unquote(env.file),
                            line: unquote(env.line)}
     Builder.apply_query(query, __MODULE__, [group_by], env)
--- a/lib/ecto/query/builder/group_by.ex
+++ b/lib/ecto/query/builder/group_by.ex
@@ -49,21 +49,21 @@ defmodule Ecto.Query.Builder.GroupBy do
   Shared between group_by and partition_by.
   """
   def group_or_partition_by!(kind, query, exprs, params) do
-    {expr, {params, _, subqueries, aliases}} =
-      Enum.map_reduce(List.wrap(exprs), {params, length(params), [], %{}}, fn
+    {expr, {params, _, subqueries}} =
+      Enum.map_reduce(List.wrap(exprs), {params, length(params), []}, fn
         field, params_count when is_atom(field) ->
           {to_field(field), params_count}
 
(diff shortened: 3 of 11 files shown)


josevalim:
> So this seems to work without too much change required, but I do have one main misgiving: This requires a modification to Ecto.QueryExpr, which AFAIK is used in other places that won't be anticipate having subqueries in them.

Let's create a new struct called ByExpr and use the opportunity to support subqueries in both group by and order by?

zachdaniel:
Happy to do that, only concern with doing that will be anywhere that was explicitly matching on `QueryExpr`, but we can suss those out as necessary. Let me give it a try and see if anything breaks :)

lib/ecto/query/builder/windows.ex
```diff
@@ -160,7 +160,14 @@ defmodule Ecto.Query.Builder.Windows do
   end
 
   defp do_runtime_window!([{:order_by, order_by} | kw], query, acc, params) do
-    {order_by, params} = OrderBy.order_by_or_distinct!(:order_by, query, order_by, params)
+    {order_by, params, subqueries, aliases} = OrderBy.order_by_or_distinct!(:order_by, query, order_by, params)
+
+    if subqueries != [] || aliases != %{} do
+      # TODO?
+      raise ArgumentError,
+            "subqueries and aliases are not allowed in runtime windows"
```
zachdaniel, line 168: I think I still need to address this as well, because we ought to support everything we support in order_bys in windows.

zachdaniel:
Alright, made some adjustments. There are two outstanding issues:

1. now this does require changes in `ecto_sql`. I'm working on those changes. They will have to be updated in tandem for this to work.
2. I might need some guidance on what to do with the subqueries/aliases present when building windows. I can't say for sure, but it seems like it might just need more updates in `ecto_sql`.

lib/ecto/query/builder/windows.ex
zachdaniel: I don't mind doing this separately, but AshPostgres does actually rely on the fact that order_by & windows support the same syntax, so I'll need to follow up with supporting this in windows soon after.

josevalim: I see, as you prefer then!

zachdaniel: Always better to limit scope though, so I will do it in a separate PR when we're done here :)

zachdaniel:
Working on wrapping this up today. I'm ending up doing windows support because windows uses the same code that `group_by` uses, so adding subquery support to that naturally flows on to adding it for windows (w/ some extra bits required to make that work).

The big thing I'm realizing based on your comment about testing non-dynamic cases is that I definitely haven't handled the compile time builder side of things. I'm going to try to do that as well, but I'm not sure if I'll be able to pull it off w/o some tips. The primary issue is that the accumulator for compile time expressions appears to support gathering up subqueries but not aliases? I'll follow up and we can see how far I get 👍

lib/ecto/query/builder/windows.ex
zachdaniel: Ended up doing it in this PR after all :)

zachdaniel:
Still have some equivalent tests to write in ecto_sql. I'm curious if the non-dynamic/dynamic split is important to test in `ecto_sql` considering that it's tested here? Regardless, still need to test each query building scenario.

zachdaniel:
Just commenting to point out that this should be ready to review. I can't see (for some reason) how or why the integration tests failed, but I'm assuming that it is because there are required changes in `ecto_sql` for this to work. Can we ship them simultaneously to address that? Do I need to find a way to make them compatible?

zachdaniel:
@greg-rychlewski like I'd update the `mix.exs` to use this branch as a dependency?

greg-rychlewski:
yeah exactly

zachdaniel:
Done :) its building now.

lib/ecto/query/builder/distinct.ex
```diff
@@ -30,11 +30,21 @@ defmodule Ecto.Query.Builder.Distinct do
...
   """
   def distinct!(query, distinct, file, line) when is_boolean(distinct) do
-    apply(query, %Ecto.Query.QueryExpr{expr: distinct, params: [], line: line, file: file})
+    apply(query, %Ecto.Query.ByExpr{expr: distinct, params: [], line: line, file: file})
   end
   def distinct!(query, distinct, file, line) do
-    {expr, params} = Builder.OrderBy.order_by_or_distinct!(:distinct, query, distinct, [])
-    expr = %Ecto.Query.QueryExpr{expr: expr, params: Enum.reverse(params), line: line, file: file}
+    {expr, params, subqueries, aliases} =
+      Builder.OrderBy.order_by_or_distinct!(:distinct, query, distinct, [])
```
josevalim, line 37: Can we actually have subqueries inside Distinct?

lib/ecto/query/builder/distinct.ex
```diff
@@ -54,12 +64,14 @@ defmodule Ecto.Query.Builder.Distinct do
 
   def build(query, binding, expr, env) do
     {query, binding} = Builder.escape_binding(query, binding, env)
-    {expr, {params, _acc}} = escape(expr, {[], %{}}, binding, env)
+    {expr, {params, acc}} = escape(expr, {[], %{subqueries: [], aliases: {:%{}, [], []}}}, binding, env)
```
josevalim, line 67: It seems a bit suspicious you are passing are passing the escaped version of `aliases` here. Do we have tests using aliases inside distinct? Should we even allow subqueries/aliases in distinct?

lib/ecto/query/planner.ex
```diff
@@ -676,6 +680,62 @@ defmodule Ecto.Query.Planner do
...
+  @spec plan_order_bys(Ecto.Query.t(), module, map()) :: Ecto.Query.t
+  defp plan_order_bys(query, adapter, cte_names) do
+    order_bys =
+      Enum.map(query.order_bys, fn
+        %{subqueries: []} = order_by ->
+          order_by
+
+          %{subqueries: subqueries} = order_by ->
+          %{order_by | subqueries: Enum.map(subqueries, &plan_subquery(&1, query, nil, adapter, false, cte_names))}
+      end)
```
josevalim, line 692: We can probably extract this into a shared function. Perhaps we can extract the whole `plan_order_bys` function into a `plan_by(:order_by, query, adapter, cte_names)`.

lib/ecto/query.ex
```diff
@@ -431,6 +431,11 @@ defmodule Ecto.Query do
     defstruct [:expr, :file, :line, params: []]
   end
 
+  defmodule ByExpr do
+    @moduledoc false
+    defstruct [:expr, :file, :line, params: [], subqueries: [], aliases: %{}]
```
josevalim, line 436: Are you sure the `aliases` field is required? I am almost sure it is about handling `selected_as`, which is not a feature we can support in group_by/order_by. It is a select only feature.

lib/ecto/query/builder/distinct.ex
zachdaniel: Yep! I tested it out.

zachdaniel:
```sql
SELECT DISTINCT ON (EXISTS (SELECT * FROM accounts AS a2 WHERE a2.id = a1.id)) * FROM accounts AS a1
```

zachdaniel: (not that you'd want to run that actual query, but confirmed that you can make subqueries in the distinct statement, at least in postgres)

lib/ecto/query.ex
zachdaniel: Ah, interesting, I didn't know what it was for. I can discard the `aliases` information and remove the `aliases` key from `ByExpr` and see if there are any issues though?

zachdaniel: Done.

lib/ecto/query/planner.ex
zachdaniel: The only "problem" with that is the subtly different state for `order_by`, `distinct`, `group_by` and `window`.

Actually I can do it for `order_by` and `group_by` as those actually are the same, but windows is a keyword list and `distinct` is a single `ByExpr`.

zachdaniel: I've done it for the two that seemed reasonable.

lib/ecto/query/builder/distinct.ex
```diff
@@ -54,12 +63,13 @@ defmodule Ecto.Query.Builder.Distinct do
...
     {query, binding} = Builder.escape_binding(query, binding, env)
-    {expr, {params, _acc}} = escape(expr, {[], %{}}, binding, env)
+    {expr, {params, acc}} = escape(expr, {[], %{subqueries: []}}, binding, env)
     params = Builder.escape_params(params)
 
-    distinct = quote do: %Ecto.Query.QueryExpr{
+    distinct = quote do: %Ecto.Query.ByExpr{
                            expr: unquote(expr),
                            params: unquote(params),
+                           subqueries: unquote(acc[:subqueries]),
```
josevalim, line 72:
```suggestion
                           subqueries: unquote(acc.subqueries),
```

lib/ecto/query/builder/group_by.ex
```diff
@@ -93,12 +93,13 @@ defmodule Ecto.Query.Builder.GroupBy do
...
     {query, binding} = Builder.escape_binding(query, binding, env)
-    {expr, {params, _acc}} = escape(:group_by, expr, {[], %{}}, binding, env)
+    {expr, {params, acc}} = escape(:group_by, expr, {[], %{subqueries: []}}, binding, env)
     params = Builder.escape_params(params)
 
-    group_by = quote do: %Ecto.Query.QueryExpr{
+    group_by = quote do: %Ecto.Query.ByExpr{
                            expr: unquote(expr),
                            params: unquote(params),
+                           subqueries: unquote(acc[:subqueries]),
```
josevalim, line 102:
```suggestion
                           subqueries: unquote(acc.subqueries),
```

lib/ecto/query/builder/order_by.ex
```diff
@@ -136,32 +136,50 @@ defmodule Ecto.Query.Builder.OrderBy do
...
   def order_by_or_distinct!(kind, query, exprs, params) do
-    {expr, {params, _}} =
-      Enum.map_reduce(List.wrap(exprs), {params, length(params)}, fn
+    {expr, {params, _, subqueries}} =
+      Enum.map_reduce(List.wrap(exprs), {params, length(params), []}, fn
         {dir, expr}, params_count when dir in @directions ->
-          {expr, params} = dynamic_or_field!(kind, expr, query, params_count)
+          {expr, params} =
+            dynamic_or_field!(kind, expr, query, params_count)
+
```
josevalim, line 144:
```suggestion
          {expr, params} = dynamic_or_field!(kind, expr, query, params_count)
```

lib/ecto/query/builder/order_by.ex
```diff
@@ -136,32 +136,50 @@ defmodule Ecto.Query.Builder.OrderBy do
...
           {{dir, expr}, params}
 
         expr, params_count ->
-          {expr, params} = dynamic_or_field!(kind, expr, query, params_count)
+          {expr, params} =
+            dynamic_or_field!(kind, expr, query, params_count)
+
```
josevalim, line 150:
```suggestion
          {expr, params} = dynamic_or_field!(kind, expr, query, params_count)
```

lib/ecto/query/builder/order_by.ex
```diff
@@ -196,13 +214,14 @@ defmodule Ecto.Query.Builder.OrderBy do
...
-    {expr, {params, _acc}} = escape(:order_by, expr, {[], %{}}, binding, env)
+    {expr, {params, acc}} = escape(:order_by, expr, {[], %{subqueries: []}}, binding, env)
     params = Builder.escape_params(params)
 
     order_by =
-      quote do: %Ecto.Query.QueryExpr{
+      quote do: %Ecto.Query.ByExpr{
               expr: unquote(expr),
               params: unquote(params),
+              subqueries: unquote(acc[:subqueries]),
```
josevalim, line 224:
```suggestion
              subqueries: unquote(acc.subqueries),
```

lib/ecto/query/builder/windows.ex
```diff
@@ -125,55 +125,58 @@ defmodule Ecto.Query.Builder.Windows do
...
 
-  defp build_compile_window({name, compile_acc, _, params}, env) do
+  defp build_compile_window({name, compile_acc, _, params, acc}, env) do
     {name,
      quote do
-       %Ecto.Query.QueryExpr{
+       %Ecto.Query.ByExpr{
          expr: unquote(compile_acc),
          params: unquote(params),
+         subqueries: unquote(acc[:subqueries]),
```
josevalim, line 138:
```suggestion
         subqueries: unquote(acc.subqueries),
```

lib/ecto/query/builder/windows.ex
```diff
@@ -125,55 +125,58 @@ defmodule Ecto.Query.Builder.Windows do
...
          file: unquote(env.file),
          line: unquote(env.line)
        }
      end}
   end
 
-  defp build_runtime_window({name, compile_acc, runtime_acc, params}, _env) do
-    {:{}, [], [name, Enum.reverse(compile_acc), runtime_acc, Enum.reverse(params)]}
+  defp build_runtime_window({name, compile_acc, runtime_acc, params, acc}, _env) do
+    {:{}, [], [name, Enum.reverse(compile_acc), runtime_acc, Enum.reverse(params), {:%{}, [], [subqueries: acc[:subqueries]]}]}
```
josevalim, line 146:
```suggestion
    {:{}, [], [name, Enum.reverse(compile_acc), runtime_acc, Enum.reverse(params), {:%{}, [], Map.to_list(acc)}]}
```

?

lib/ecto/query/builder/windows.ex
```diff
@@ -125,55 +125,58 @@ defmodule Ecto.Query.Builder.Windows do
...
   @doc """
   Invoked for runtime windows.
   """
   def runtime!(query, runtime, file, line) do
     windows =
-      Enum.map(runtime, fn {name, compile_acc, runtime_acc, params} ->
-        {acc, params} = do_runtime_window!(runtime_acc, query, compile_acc, params)
-        expr = %Ecto.Query.QueryExpr{expr: Enum.reverse(acc), params: Enum.reverse(params), file: file, line: line}
+      Enum.map(runtime, fn {name, compile_acc, runtime_acc, params, escape_acc} ->
+        {{acc, subqueries}, params} = do_runtime_window!(runtime_acc, query, {compile_acc, escape_acc[:subqueries]}, params)
```
josevalim, line 155:
```suggestion
        {{acc, subqueries}, params} = do_runtime_window!(runtime_acc, query, {compile_acc, escape_acc.subqueries}, params)
```

test/ecto/query/builder/distinct_test.exs
```diff
@@ -63,6 +63,21 @@ defmodule Ecto.Query.Builder.DistinctTest do
...
+      %{distinct: distinct} = from p in "posts", as: :p, distinct: ^distinct
+      assert distinct.expr ==  [asc: {:exists, [], [subquery: 0]}]
+      assert [_] = distinct.subqueries
+
+      %{distinct: distinct} = from p in "posts", as: :p, distinct: [asc: exists(from other_post in "posts", where: other_post.id == parent_as(:p).id)]
+      assert distinct.expr ==  [asc: {:exists, [], [subquery: 0]}]
+      assert [_] = distinct.subqueries
+    end
+
+
```
josevalim, line 80:
```suggestion
```

josevalim (approved):
Looks great to me. Any other comments @greg-rychlewski? :)

greg-rychlewski:
looks good to me :)

josevalim:

PR #4418, opened 2024-05-17
Add sslmode=verify-full connection string expansion.

chrismccord:
This query param is used by postgres and cockroachdb to verify to force a secure connection,
verify that the server certificate is signed by a known CA, and verify that the server address matches that specified in the certificate.

This requires setting various complex ssl options that we can do for the user automatically.

mysql also uses the sslMode terminology for their ssl options, so I believe this can also align there as well.


josevalim:
Thank you, this helped moved the conversation forward, see PR in https://github.com/elixir-ecto/postgrex/pull/677!

PR #4422, opened 2024-05-26
Consider only sort_param if the assoc param is not provided, not the drop_param.

mathieuprog:
1. The following commit message says it should consider the sort_param, without mentioning the drop_param: https://github.com/elixir-ecto/ecto/commit/afc694ce723f047e9fe7828ad16cea2de82eb217

2. The tests still pass when removing the check `is_list(drop)`

3. Adding the last 2 lines in the tests:
```elixir
changeset = cast(%Author{}, %{sort: ["new"]}, :posts, opts)
assert Enum.map(changeset.changes.posts, & &1.changes[:title]) == [nil] # passes

changeset = cast(%Author{}, %{drop: [0]}, :posts, opts)
assert Enum.map(changeset.changes.posts, & &1.changes[:title]) == [nil] # doesn't pass
```
raises with message
> ** (KeyError) key :posts not found in: %{}

Based on these 3 observations, I suspect the check `is_list(drop)` shouldn't be there.

### DIFF
--- a/test/ecto/changeset/embedded_test.exs
+++ b/test/ecto/changeset/embedded_test.exs
@@ -463,6 +463,9 @@ defmodule Ecto.Changeset.EmbeddedTest do
     changeset = cast(%Author{}, %{sort: ["new"]}, :posts, opts)
     assert Enum.map(changeset.changes.posts, & &1.changes[:title]) == [nil]
 
+    changeset = cast(%Author{}, %{drop: [0]}, :posts, opts)
+    assert Enum.map(changeset.changes.posts, & &1.changes[:title]) == [nil]
+
     changeset = cast(%Author{}, %{posts: posts, sort: [2, 3, 1]}, :posts, opts)
     assert Enum.map(changeset.changes.posts, & &1.changes[:title]) == ~w(two three one)
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1290,7 +1290,7 @@ defmodule Ecto.Changeset do
     drop = opts_key_from_params(:drop_param, opts, params)
 
     changeset =
-      if is_map_key(params, param_key) or is_list(sort) or is_list(drop) do
+      if is_map_key(params, param_key) or is_list(sort) do
         value = Map.get(params, param_key)
         original = Map.get(data, key)
         current = Relation.load!(data, original)


josevalim:
Can you please add the failing test that is failing? I suspect we should correct it to not raise, instead of discarding the option.

mathieuprog:
Could you have a look now @josevalim ?

test/ecto/changeset/embedded_test.exs
```diff
@@ -463,6 +463,9 @@ defmodule Ecto.Changeset.EmbeddedTest do
     changeset = cast(%Author{}, %{sort: ["new"]}, :posts, opts)
     assert Enum.map(changeset.changes.posts, & &1.changes[:title]) == [nil]
 
+    changeset = cast(%Author{}, %{drop: [0]}, :posts, opts)
+    assert Enum.map(changeset.changes.posts, & &1.changes[:title]) == [nil]
```
josevalim, line 467: Isn't this test failing because we are dropping from an empty list and, therefore, there are indeed no changes? What is failing here is the test code, not the actual code?

Can you provide a test code that failed before and passes now? Or is the whole thing an optimization?

mathieuprog: I opened this PR because I don't see why we check for `is_list(drop)`, i.e. which case it covers.

I'm not a user of these options but I have to mimic these options for `polymorphic_embed` lib, that's how I go into these details:)

Can we find a failing test for the omission of this condition, which would prove that this condition is necessary?

Or maybe you added this option for a matter of consistency (just checking both options), despite that there are probably no cases where a missing key+presence of drop option would succeed.

josevalim: That's a better phrasing of the problem, I was missing the context. I can push a test that shows it working.

mathieuprog: Thank you for adding the test. Isn't it weird we have empty changes and an action to replace?
The same test with `sort` option instead of `drop` option leads to an error:
```elixir
changeset = cast(%Author{posts: [%Post{title: "one"}]}, %{drop: [9]}, :posts, opts)
IO.inspect changeset.changes

%{posts: [
   #Ecto.Changeset<action: :replace, changes: %{}, errors: [],
    data: #Ecto.Changeset.EmbeddedTest.Post<>, valid?: true, ...>
]}
```

```elixir
changeset = cast(%Author{posts: [%Post{title: "one"}]}, %{sort: ["new"]}, :posts, opts)
IO.inspect changeset.changes

%{}
```

josevalim: I can’t recall the exact conditions but this was 100% added based on scenarios we ran into building LiveView apps.

PR #4426, opened 2024-06-01
Merge types during `Ecto.Changeset.merge/2`

### LINKED ISSUE #4424: Ecto.Changeset.merge do not merge types and do not raises when types are different
### Elixir version

1.16.1

### Database and Version

PostgreSQL: postgres:13-alpine

### Ecto Versions

3.11.2

postgrex 0.17.4

### Current behavior

During the `Ecto.Changeset.merge/2` operation the type field of the changeset is not merged. Which result in a strange result when applying changes using schemaless changesets

```elixir
data = %{}
changeset1 = Ecto.Changeset.cast({data, %{field1: :string}}, %{field1: "field 1"}, [:field1])
changeset2 = Ecto.Changeset.cast({data, %{field2: :string}}, %{field2: "field 2"}, [:field2])
Ecto.Changeset.merge(changeset1, changeset2) |> Map.get(:types)
# %{field2: :string}

Ecto.Changeset.merge(changeset1, changeset2) |> Ecto.Changeset.apply_changes()
# %{field2: "field 2"}
```

Is it the expected behaviour?
I see in the code the merging of types is performed via `Kernel.||/2` and not a `Map.merge/2` operation

```elixir
  defp cast_merge(cs1, cs2) do
    ...
    new_types = cs1.types || cs2.types
    ...
  end 
```

### Expected behavior

I can see 2 types of behaviour that could be acceptable:

1. Merge the types for easiness to use merge with schema less changeset
2. Raise if the types are not equals (like when the data from the changesets are different)

vinibrsl:
This commit adds support for merging types when calling `Ecto.Changeset.merge/2`.

Closes #4424

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1523,10 +1523,10 @@ defmodule Ecto.Changeset do
 
   defp cast_merge(cs1, cs2) do
     new_params = (cs1.params || cs2.params) && Map.merge(cs1.params || %{}, cs2.params || %{})
+    new_types = (cs1.types || cs2.types) && Map.merge(cs1.types || %{}, cs2.types || %{})
     new_changes = Map.merge(cs1.changes, cs2.changes)
     new_errors = Enum.uniq(cs1.errors ++ cs2.errors)
     new_required = Enum.uniq(cs1.required ++ cs2.required)
-    new_types = cs1.types || cs2.types
     new_valid? = cs1.valid? and cs2.valid?
 
     %{
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1523,7 +1523,7 @@ defmodule Ecto.Changeset do
 
   defp cast_merge(cs1, cs2) do
     new_params = (cs1.params || cs2.params) && Map.merge(cs1.params || %{}, cs2.params || %{})
-    new_types = (cs1.types || cs2.types) && Map.merge(cs1.types || %{}, cs2.types || %{})
+    new_types = Map.merge(cs1.types, cs2.types)
     new_changes = Map.merge(cs1.changes, cs2.changes)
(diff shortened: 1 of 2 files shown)


lib/ecto/changeset.ex
```diff
@@ -1523,10 +1523,10 @@ defmodule Ecto.Changeset do
 
   defp cast_merge(cs1, cs2) do
     new_params = (cs1.params || cs2.params) && Map.merge(cs1.params || %{}, cs2.params || %{})
+    new_types = (cs1.types || cs2.types) && Map.merge(cs1.types || %{}, cs2.types || %{})
```
vinibrsl, line 1526: From the `Ecto.Changeset.t/1` typespec, `types` can be `nil`, so I implemented fallbacks with `||`. However, in the context of `Ecto.Changeset.merge/2`, I'm not sure if `types` can actually be `nil`, since an `(ArgumentError) different :data when merging changesets` error is raised if the `data` fields differ.

Can someone confirm if it's safe to simply `Map.merge(cs1.types, cs2.types)` for a cleaner solution?

josevalim, line 1526:
```suggestion
    new_types = Map.merge(cs1.types, cs2.types)
```

lib/ecto/changeset.ex
```diff
@@ -1523,10 +1523,10 @@ defmodule Ecto.Changeset do
 
   defp cast_merge(cs1, cs2) do
     new_params = (cs1.params || cs2.params) && Map.merge(cs1.params || %{}, cs2.params || %{})
+    new_types = Map.merge(cs1.types, cs2.types)
```
josevalim, line 1526: It seems the assumption is that some people were initializing the changeset directly without the proper API. I don't think we need to support that.

josevalim:

PR #4427, opened 2024-06-02
Add new functions to Repo behaviour

tmbb:
This PR is not suitable for merging right away because of a few reasons (detailed below), but it serves as a template for what I think should be included in Ecto.

## Rationale

Currently there is no way to recursively preload associations in deeply nested changesets. Preloading association in deeply nested changesets is important if one is working with nested forms inside LiveView. LiveView expects forms built from changesets (not ecto structs!) and it expects associations to be preloaded so that the HTML components can iterate over the value's "children".

One could think of dropping changesets from LiveView, but changesets are really convenient to validate data and to make sure that errors are rendered correctly in HTML forms.

Nothing inside this PR refers to LiveView, of course, but it's just the motivating example that led me to write it.

## Changes implemented in this PR

This PR implements two new public callbacks in the `Repo` behaviour, namely:

* `@spec preload_in_result({:ok, Ecto.Schema.t()} | {:error, Ecto.Changeset.t()}, list(), list()) :: {:ok, struct} | {:error, changeset}` which is optimized for piping results returned by functions such as `Repo.insert()` and `Repo.update()`
* `@spec preload_in_changeset(changeset, list(), list())` to preload associations inside a changeset (the function above calls this one when given a `{:error, changeset}` tuple)

The implementation is not very efficient in the sense that it probably generates more queries than is really required, but that can be optimized by walking the nested changesets twice (once to get the IDs of all the parents to get the children in a single query and then to attribute each child to the correct parent based on the ID; changesets without IDs can't have children already in the database, and don't need to be queryeed - I already do this optimization).

## Testing

I'm trying to test this with the `TestRepo` which uses a custom adapter (which I don't understand at all, I've never looked into adapter code), but it seems like the `TestRepo` is not generating IDs correctly when multiple changesets are inserted (?).

## Questions

1. Is this considered to be in scope for Ecto? I believe it should be considered in scope because it's non-trivial functionality which might be useful for a number of projects, but probably too trivial to be in its own package.
2. How can I make the test adapter generate IDs correctly? Should I instead test with a new adapter, even if I have to import something just for testing?

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -587,6 +587,28 @@ defmodule Ecto.Repo do
           )
         end
 
+        def preload_in_result(result, preloads, opts \\ []) do
+          repo = get_dynamic_repo()
+
+          Ecto.Repo.Preloader.preload_in_result(
+            result,
+            repo,
+            preloads,
+            Ecto.Repo.Supervisor.tuplet(repo, prepare_opts(:preload, opts))
+          )
+        end
+
+        def preload_in_changeset(changeset, preloads, opts \\ []) do
+          repo = get_dynamic_repo()
+
+          Ecto.Repo.Preloader.preload_in_changeset(
+            changeset,
+            repo,
+            preloads,
+            Ecto.Repo.Supervisor.tuplet(repo, prepare_opts(:preload, opts))
+          )
+        end
+
         def prepare_query(operation, query, opts), do: {query, opts}
         defoverridable prepare_query: 3
       end
--- a/lib/ecto/repo/preloader.ex
+++ b/lib/ecto/repo/preloader.ex
@@ -696,4 +696,153 @@ defmodule Ecto.Repo.Preloader do
   defp filter_and_reraise(exception, stacktrace) do
     reraise exception, Enum.reject(stacktrace, &match?({__MODULE__, _, _, _}, &1))
   end
+
+  # Preload in changeset or in result
+
+  @doc """
+  Implementation for `Ecto.Repo.preload_in_result/2`.
+  """
+
+  @spec preload({:ok, struct} | {:error, changeset}, atom, atom | list, {adapter_meta :: map, opts :: Keyword.t}) ::
+    {:ok, struct} | {:error, changeset} when struct: Ecto.Schema.t, changeset: Ecto.Changeset.t
+
+  def preload_in_result({:ok, struct}, repo_name, preloads, tuplet) do
+    {:ok, preload(struct, repo_name, preloads, tuplet)}
+  end
+
+  def preload_in_result({:error, changeset}, repo_name, preloads, tuplet) do
+    {:error, preload_in_changeset(changeset, repo_name, preloads, tuplet)}
+  end
+
+  @doc """
+  Implementation for `Ecto.Repo.preload_in_changeset/2`.
+  """
+
+  @spec preload_in_changeset(changeset, atom, atom | list, {adapter_meta :: map, opts :: Keyword.t}) ::
+    changeset when changeset: Ecto.Changeset.t
+
+  def preload_in_changeset(%Ecto.Changeset{} = changeset, repo_name, preloads, tuplet) do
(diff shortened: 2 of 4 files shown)


josevalim:
Hi @tmbb! I need to understand the problem a bit better. Why can't the associations be loaded before you put the data in the changeset? Can you provide a general example with before/after that could benefit for this?

tmbb:
@josevalim the problem of the "non-preloaded" associations happen when I use the `cast_assoc(:assoc, sort_param: :assoc_sort)` functionalot to create a new child association. In that case, neither the child association nor its siblings have their associations preloaded. Maybe this is a problem that should be solved at the level of the `cast_assoc/2` function, but I don't think the `cast_assoc/2` function receives enough data to know which associations should be preloaded, or even what the default value of the preloaded associations should be. And preloading things shouldn't be the job of the `cast_assoc/2` function anyway (it doesn't even have access to the repo).

An example can be found in the test I've added in the newest commit: https://github.com/elixir-ecto/ecto/blob/f46ed7fdfa1e182deef67e2688fee057c41600f8/test/ecto/repo/preload_test.exs#L65-L105

josevalim:
Why can't you preload the associations into the struct before you call `cast`? You already have to load it from the database, why not preload it by then?

greg-rychlewski:
It looks like their issue is when the data is not already in the db but when casting will produce a changeset for the creation of a new entry. 

But I am still not 100% sure I understand why it's an issue. It seems like it's normal for that to exist only in `changes` before the persistence happens. And then once you call `Repo.insert/update`, you can have them in `data`.

tmbb:
> Why can't you preload the associations into the struct before you call `cast`?

As shown in the code above, I am already preloading the data before the cast. It still doesn't give me preloaded associations for the  `books`.

> It looks like their issue is when the data is not already in the db but when casting will produce a changeset for the creation of a new entry.

The problem is not exactly that, the problem is that I want the `:chapters` associations to be preloaded in the data of the book change set (even if the preload is an empty list).

I need the associations to be preloaded in the changeset data (the changeset data, that is, order to be able to iterate over the chapters in nested inputs in a Phoenix form (with `<.nested_inputs_for/>`).

I want the form to work even with changeset that haven't been persisted, in which the user mau be performing edits with incremental validation.

I hope I've made it more clear

josevalim:
I think the simplest solution is to add a flag to cast_assoc that resets the association if unloaded?

greg-rychlewski:
i think something like that makes sense. to me it sounds like when the cast functions are creating changesets for new inserts the associations are defaulting to unloaded and that is causing the issue.

tmbb:
> i think something like that makes sense. to me it sounds like when the cast functions are creating changesets for new inserts the associations are defaulting to unloaded and that is causing the issue.

Exactly, that's the issue. But I'm not sure that `cast_assoc/2` even has enough information to preload the associations or to even known which associations to preload (which in case of creation means that the associations are "empty"). Is the list of preloads stored anywhere in the data or changesets?

greg-rychlewski:
@tmbb I think it can be as simple as [adding this change here](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L1296)

```elixir
force_load = opts[:force_load]
data = if force_load, do: %{data | key => Relation.load!(data, original)}, else: data
changeset = %{force_update(changeset, opts) | data: data, changes: changes, valid?: valid?}
```

By the time you get here you know if it's not loaded then it's a new entry. [Otherwise this would have raised](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L1289).

josevalim:
@greg-rychlewski I didn't check the code (sorry, I am being lazy) but would that only be invoked for new records? For updates it can lead to weird failures.

greg-rychlewski:
I think so but maybe I need to check all the cases more closely. My thinking is that `original` is defined this way:

```elixir
original = Map.get(data, key)
```

josevalim:
@tmbb, please feel free to explore the option to `cast_assoc` route above. We don't plan to move forward with the exact approach in this PR. Thank you!

PR #4428, opened 2024-06-05
Add anchors in documentation

jcrevits:
Hello! I've stumbled upon a few places in the docs that could use anchors instead of having to search through the module page. Feel free to change the formatting! Thank you!


v0idpwn (commented):
I think that when you add anchors you can remove the "on the module documentation" parts

josevalim:
I would keep the "on the module documentation" because it can be helpful in places where the link does not work, such as IEx.

josevalim:

PR #4431, opened 2024-06-13
Allow select: map(x, fields), select_merge: %{} as source query for insert_all. Fixes #4430

### LINKED ISSUE #4430: Support `select: map(x, fields), select_merge: %{...}` in source query for insert_all
### Elixir version

1.16.3

### Database and Version

PostgreSQL 15.0

### Ecto Versions

master

postgrex

### Current behavior

```elixir
Repo.insert_all(Foo, from x in Foo, select: %{foo: x.foo, bar: x.bar}, select_merge: %{baz: some_fun(x.baz)})
```

works as intended.

```elixir
Repo.insert_all(Foo, from x in Foo, select: map(x, [:foo, :bar]), select_merge: %{baz: some_fun(x.baz)})
```

raises an ArgumentError stating that the source query does not select into a map.

### Expected behavior

The above two snippets should both be supported as they should result in the same source query being produced.

I would be happy to take a stab at this if this is indeed the intended behaviour.

nmk:


test/ecto/repo_test.exs
```diff
@@ -680,6 +680,17 @@ defmodule Ecto.RepoTest do
...
 
+    test "takes query selecting on map + select_merge" do
+      query =
+        from s in MySchema,
+          select: map(s, [:x]),
+          select_merge: %{y: s.y}
+
+      TestRepo.insert_all(MySchema, query)
+
+      assert_received {:insert_all, %{source: "my_schema"}, {%Ecto.Query{}, _params}}
```
josevalim, line 691: Is there a way to check the fields here? We need to check that both x and y are being sent.

lib/ecto/repo/schema.ex
```diff
@@ -112,27 +112,23 @@ defmodule Ecto.Repo.Schema do
...
-          case dumper do
-            %{^field => {_, _, false}} -> field
-            %{} -> raise ArgumentError, "cannot select unwritable field `#{field}` for insert_all"
-            nil -> field
-          end
-        end)
+        Enum.map(args, fn {field, _} -> extract_field(field, dumper) end)
+
+      %Ecto.Query.SelectExpr{expr: {:merge, _ctx, merge_args}, take: %{^ix => {_fun, take_fields}}} ->
+        [_, {_, _, merge_fields}] = merge_args
```
greg-rychlewski, line 124: I believe it can be in the reverse order too like this

```
{:merge, [],
   [
     {:%{}, [], [y: {{:., [type: :binary], [{:&, [], [0]}, :yyy]}, [], []}]},
     {:&, [], [0]}
   ]}
```

greg-rychlewski:
We should be extra sure we are catching all the cases here https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/builder/select.ex#L367

greg-rychlewski:
Playing around with it more. This can get pretty dicey with nested merges. For example this query

```elixir
from s in MySchema,
          where: s.x > ^threshold,
          select: %{y: s.y},
          select_merge: map(s, [:x]),
          select_merge: %{x: s.x}
```

turns into this

```elixir
{:merge, [],
   [
     {:merge, [],
      [
        {:%{}, [], [y: {{:., [type: :binary], [{:&, [], [0]}, :yyy]}, [], []}]},
        {:&, [], [0]}
      ]},
     {:%{}, [], [x: {{:., [type: :string], [{:&, [], [0]}, :x]}, [], []}]}
   ]}
```

nmk:
@greg-rychlewski I can see in https://github.com/elixir-ecto/ecto/blob/e0bb7299fe0540998aeb6b10e5b87b63f66f61cb/test/ecto/query/planner_test.exs#L2311 that there is a `fields` field on the `Ecto.Query.SelectExpr` struct. Would it be possible to use this field to extract the headers?

josevalim:
That will lose you the field names. However, in the planner module, we have a function for extracting fields of a subquery, maybe that can be used, but I think in this case it is probably better to improve the error message and have a single select with a map for simplicity, for full control over the database changes.

PR #4432, opened 2024-06-14
Convert structs that implement the Enumerable protocol to lists

andersonmcook:
I found it surprising that I could not use a `MapSet` or another `Enum` when using `in` in a query as I expected it to function similarly to `Kernel.in/2`.
```elixir
ids = MapSet.new([1, 2, 3])
where(Comment, [c], c.id in ^ids) 
```

This PR checks for an implementation of the `Enumerable` protocol and converts it to a list when casting, returning an error if there is none.

If this isn't the direction you want Ecto to go, I'm happy to have this closed. Otherwise, let me know anything else I should be mindful of or any documentation I should update.

Thanks!

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -135,8 +135,9 @@ defmodule Ecto.Query.API do
 
       from p in Post, where: p.id in [1, 2, 3]
 
-  The right side may either be a list, a literal list, a struct that implements
-  the `Enumerable` protocol, or even a column in the database with array type:
+  The right side may either be a literal list, an interpolated list,
+  any struct that implements the `Enumerable` protocol, or even a
+  column in the database with array type:
 
       from p in Post, where: "elixir" in p.tags
 
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -135,8 +135,8 @@ defmodule Ecto.Query.API do
 
       from p in Post, where: p.id in [1, 2, 3]
 
-  The right side may either be a list, a literal list
-  or even a column in the database with array type:
+  The right side may either be a list, a literal list, a struct that implements
+  the `Enumerable` protocol, or even a column in the database with array type:
 
       from p in Post, where: "elixir" in p.tags
 
--- a/lib/ecto/type.ex
+++ b/lib/ecto/type.ex
@@ -1387,6 +1387,13 @@ defmodule Ecto.Type do
     {:ok, Enum.reverse(acc)}
   end
 
+  defp array_with_index(%_{} = struct, fun, skip_nil?, index, acc) do
+    case Enumerable.impl_for(struct) do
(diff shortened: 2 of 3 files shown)


josevalim (approved):

greg-rychlewski:
Thanks. It would be good to update the docs here: https://hexdocs.pm/ecto/Ecto.Query.API.html#in/2

lib/ecto/type.ex
```diff
@@ -1387,6 +1387,13 @@ defmodule Ecto.Type do
...
+  defp array_with_index(%_{} = struct, fun, skip_nil?, index, acc) do
+    case Enumerable.impl_for(struct) do
+      nil -> :error
+      _ -> struct |> Enum.to_list() |> array_with_index(fun, skip_nil?, index, acc)
+    end
+  end
+
   defp array_with_index(_, _, _, _, _) do
     :error
   end
```
greg-rychlewski, line 1399:
```suggestion
  defp array_with_index(other, fun, skip_nil?, index, acc) do
    case Enumerable.impl_for(other) do
      nil -> :error
      _ -> other |> Enum.to_list() |> array_with_index(fun, skip_nil?, index, acc)
    end
  end
```
@josevalim is there any reason to limit it to structs or should we do something like this?

josevalim: The only other built in is maps and I don’t think it would work here?

greg-rychlewski: ah because it turns into a keyword list right? yeah makes sense!

andersonmcook: I updated the documentation specifically calling out structs, as I'm not sure that comparing the key/value pairs may work intuitively?

andersonmcook: We could pull out the values of a keyword/map and call it out in the documentation, as well.

greg-rychlewski: imo it would be a bit confusing to do something like this `from t in Table, where: t.int_row in %{a: 1, b: 2}`. to me the current change looks good. will approve and if jose agrees to leave out maps/kw lists then we are good to go

greg-rychlewski (approved):

lib/ecto/query/api.ex
```diff
@@ -135,8 +135,8 @@ defmodule Ecto.Query.API do
 
       from p in Post, where: p.id in [1, 2, 3]
 
-  The right side may either be a list, a literal list
-  or even a column in the database with array type:
+  The right side may either be a list, a literal list, a struct that implements
+  the `Enumerable` protocol, or even a column in the database with array type:
```
josevalim, line 139:
```suggestion
  The right side may either be a literal list, an interpolated list,
  any struct that implements the `Enumerable` protocol, or even a
  column in the database with array type:
```

josevalim:

PR #4433, opened 2024-06-17
insert_all with query was not using field source

greg-rychlewski:
The issue is that when a query is used as the data for `insert_all`, the map keys are not converted into the field source names.  This is in contrast to when a list of maps/kw lists are given, [which use the field source name](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/schema.ex#L176).

I believe we want to standardize this so it's the same behaviour both ways but it would be good to confirm you agree. Thanks.


v0idpwn (approved):
Looks right to me

josevalim (approved):

PR #4438, opened 2024-06-23
Allow 2-arity preload function

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/pull/4401

I'll work on the subquery preload a bit later because it's more complicated.


.DS_Store
josevalim, line 1: Oops :)

josevalim (approved):

PR #4439, opened 2024-06-24
Allow join preloads with subqueries

greg-rychlewski:
Following our discussion about lateral preloads, I took a stab at allowing subqueries to be both the parent and child of join preloads.

The main deterrent is that subqueries do not have a schema assigned to them, so it's not possible to directly obtain their primary keys or association metadata. These are both important when associating the parents/children once the information is retrieved from the database.

The workaround proposed here is to take the `related` schema from the owner association metadata and assume the subquery complies with it. This is similar but not 100% the same as how it works today. Today you could theoretically preload schema `A` while joining on schema `B` as long as `B` has its own primary key and has the fields needed for the association.

In this PR i took a stab only at joining in plain subqueries and not `assoc(...)`. I have to think a bit more about `assoc(...)` but I think we want to support this anyway....

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1155,15 +1155,22 @@ defmodule Ecto.Query.Planner do
   Prepare association fields found in the query.
   """
   def plan_assocs(query) do
-    plan_assocs(query, 0, query.assocs)
+    plan_assocs(query, 0, query.assocs, nil)
     query
   end
 
-  defp plan_assocs(_query, _ix, []), do: :ok
+  defp plan_assocs(_query, _ix, [], _owner_refl), do: :ok
 
-  defp plan_assocs(query, ix, assocs) do
-    # We validate the schema exists when preparing joins above
-    {_, parent_schema, _} = get_preload_source!(query, ix)
+  defp plan_assocs(query, ix, assocs, owner_refl) do
+    # We validate the schema exists when preparing joins.
+    # If preloading into a subquery, we assume its schema
+    # is the same as the related schema from the association
+    # metadata of the owner.
+    parent_schema =
+      case get_preload_source!(query, ix) do
+        {_, schema, _} -> schema
+        %Ecto.SubQuery{} -> owner_refl.related
+      end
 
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
no I messed up...we can't subtract 1 from its source index to get the owner source. it can be multiple levels up. need to rethink this

greg-rychlewski:
Wasn't so bad to fix that. This is in a state that can be reviewed =)

lib/ecto/query/planner.ex
```diff
@@ -2283,11 +2290,14 @@ defmodule Ecto.Query.Planner do
...
 
+      %Ecto.SubQuery{} = subquery when ix != 0 ->
+        subquery
+
       _ ->
         error!(
           query,
-          "can only preload sources with a schema " <>
-            "(fragments, binary and subqueries are not supported)"
+          "can only preload sources with a schema or subqueries in `join`" <>
```
josevalim, line 2299: If we are assuming it defaults to the association, then we can allow anything, including fragments, no?

josevalim: Definitely, will fully review when back at home

josevalim (approved):

greg-rychlewski:
@josevalim I stumbled across something else that I think might be a better way to do this. When you have some free time, would you be able to let me know what you think?

Instead of assuming the subquery returns the proper schema, I believe we can use the `:select` field on the subquery. This field is populated during subquery source planning, so it will be available to us when planning assocs. [It is populated as one of these](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L431):

```elixir
defp subquery_source(nil, types), do: {:map, types}
defp subquery_source(name, types) when is_atom(name), do: {:struct, name, types}

defp subquery_source({:source, schema, prefix, types}, only) do
    types =
      Enum.map(only, fn {field, {:value, type}} -> {field, Keyword.get(types, field, type)} end)

    {:source, schema, prefix, types}
end
```

So basically we would check that it's ` {:source, schema, prefix, types}` and use `schema`. And this way we can even do it in `from`.

WDYT? Do you think it sounds worth exploring?

josevalim:
Sounds good to me! :)

PR #4442, opened 2024-07-03
Return keyword list of associations for a given module

andersonmcook:
Adds the ability to find the name and type of association between two schema modules via schema reflection. It's a combination of `__schema__(:associations)` and `__schema__(:association, assoc)`.

Doesn't handle traversing `through` associations.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -433,6 +433,9 @@ defmodule Ecto.Schema do
   * `__schema__(:virtual_type, field)` - Returns the type of the given virtual field;
 
   * `__schema__(:associations)` - Returns a list of all association field names;
+  * `__schema__(:associations, module)` - Returns a keyword list of association
+    reflections of the given module;
+
   * `__schema__(:association, assoc)` - Returns the association reflection of the given assoc;
 
   * `__schema__(:embeds)` - Returns a list of all embedded field names;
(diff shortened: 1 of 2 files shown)


josevalim:
What is the use case? Couldn’t it perhaps be something you compute somehow else?

andersonmcook:
> What is the use case? Couldn’t it perhaps be something you compute somehow else?

I have a module that recursively clones records through child associations. So if we have a schema:
```mermaid
erDiagram
  GRANDPARENT ||--o{ PARENT : has
  PARENT ||--o{ CHILD :has
```
We can pass an existing `%Grandparent{}` to the cloning function and it will clone through all its children, copying the fields that should be copied and reassociating the new record to appropriate parent.
```mermaid
graph
  subgraph o[Original]
    gp[Grandparent]
    p1[Parent]
    p2[Parent]
    c1[Child]
    c2[Child]
    c3[Child]
    c4[Child]
    
    gp --> p1 & p2
    p1 --> c1 & c2
    p2 --> c3 & c4
  end

  subgraph c[Clones]
    cgp[Grandparent]
    cp1[Parent]
    cp2[Parent]
    cc1[Child]
    cc2[Child]
    cc3[Child]
    cc4[Child]
     
    cgp --> cp1 & cp2
    cp1 --> cc1 & cc2
    cp2 --> cc3 & cc4
  end
  
  gp -.->|clones|cgp
  p1 -.->|clones|cp1
  p2 -.->|clones|cp2
  c1 -.->|clones|cc1
  c2 -.->|clones|cc2
  c3 -.->|clones|cc3
  c4 -.->|clones|cc4
```

We have some records that are recursive or need to inherit from multiple parents so part of the cloning calls:
```elixir
defp assoc(%{data: %child{}} = changeset, parent) do
  for {_, {:assoc, assoc}} <- parent.__struct__.__changeset__(),
      assoc.related == child,
      assoc.relationship == :child,
      reduce: changeset do
    changeset ->
      Changeset.change(changeset, [{assoc.related_key, Map.fetch!(parent, assoc.owner_key)}])
  end
end
```
This uses the existing `changeset` reflection to get all child associations and associate them to the parent. An alternative would be to call the `associations` reflection and then get the association struct for each key with the `association` reflection.

😆 I recognize this is a niche use-case and I can work around it. An alternative to the proposal here could be to have another reflection that gives the association key and the association struct.

No hard feelings if you don't want this in Ecto 😄

josevalim:
Yes, I believe this is too specific to have in Ecto, given it is metadata that we will now compute and store for every possible schema. Thanks!

PR #4450, opened 2024-07-13
Improve docs of Multi

ponychicken:
The statement "Accepts the same arguments and options" was confusing because the other functions don't have a `name` argument


josevalim:

PR #4458, opened 2024-07-24
Expand order by before recursion

greg-rychlewski:
I was trying to fix the regression introduced by [this PR](https://github.com/elixir-ecto/ecto/pull/4449) where something like this doesn't work anymore for order by

```elixir
defmacro my_custom_order(p) do
  quote do
    [unquote(p).id, some_sql_function(unquote(p))]
  end
end
```

It's caused because we are calling `escape` again when the macro is expanded so each element inside of it gets escaped twice.

I am probably missing something obvious, but the approach I took in this PR is to escape the order by expression before we go down into the recursion, similar to what we do with `from`.

### DIFF
--- a/lib/ecto/query/builder/order_by.ex
+++ b/lib/ecto/query/builder/order_by.ex
@@ -51,6 +51,7 @@ defmodule Ecto.Query.Builder.OrderBy do
           {Macro.t(), {list, term}}
   def escape(kind, expr, params_acc, vars, env) do
     expr
+    |> Macro.expand_once(env)
     |> List.wrap()
     |> Enum.flat_map_reduce(params_acc, &do_escape(&1, &2, kind, vars, env))
   end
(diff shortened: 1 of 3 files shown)


lib/ecto/query/builder/order_by.ex
```diff
@@ -51,39 +51,37 @@ defmodule Ecto.Query.Builder.OrderBy do
...
   end
 
   defp do_escape({dir, expr}, params_acc, kind, vars, env) do
     {ast, params_acc} = Builder.escape(expr, :any, params_acc, vars, env)
-    {[{quoted_dir!(kind, dir), ast}], params_acc}
+    {{quoted_dir!(kind, dir), ast}, params_acc}
   end
 
-  defp do_escape(expr, params_acc, kind, vars, env) do
+  defp do_escape(expr, params_acc, _wtkind, vars, env) do
```
greg-rychlewski, line 82:
```suggestion
  defp do_escape(expr, params_acc, _kind, vars, env) do
```

PR #4463, opened 2024-07-29
Document conditional upserts

nathanl:


josevalim (approved):

lib/ecto/repo.ex
```diff
@@ -1537,6 +1537,28 @@ defmodule Ecto.Repo do
...
+  database features. For example, PostgreSQL supports conditional upserts like
+  `DO UPDATE SET title = EXCLUDED.title WHERE EXCLUDED.version > post.version`.
+  This means that the title will be updated only if the proposed row has a
+  greater version value than the existing row.
+
+  Ecto can support this as follows:
+
+      conflict_query =
+        from(p in Post,
+          update: [set: [title: fragment("EXCLUDED.title")]],
```
greg-rychlewski, line 1553: not sure if it's too pedantic for an example, but probably the version should be updated as well?

greg-rychlewski:
@josevalim This PR is making me think of something. Because I have this kind of query everywhere in my codebase.

The most awkward part is creating the `set` keyword list. It would be pretty nice if we could do what we do for `replace` and just take the names of the fields and convert them to `EXCLUDED` in the adapter. Do you think we could do that here and just raise if it's not part of a conflict query?

josevalim:
@greg-rychlewski I think supporting `replace:` is a bit weird, as it only really works for upserts. :( I am wondering if this is something we need conveniences in Ecto for? You could abstract this behind functions and macros yourself, right?

lib/ecto/repo.ex
nathanl: @greg-rychlewski Good point. I expanded the example to update the version. I also made the title less boring. 😄

greg-rychlewski (approved):

greg-rychlewski:
@nathanl thank you!
@josevalim yeah good point this could be handled easily with macros

PR #4467, opened 2024-08-01
Document Ecto.Schema.has_many/3 :preload_order option with MFA

sygour:
On `has_many` relations, `:preload_order` option is also compatible with MFA.

This is currently stated in `many_to_many/3` documentation, this pull request aims at making those two pieces of documentation consistent.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -856,12 +856,15 @@ defmodule Ecto.Schema do
     * `:where` - A filter for the association. See "Filtering associations" below.
       It does not apply to `:through` associations.
 
-    * `:preload_order` - Sets the default `order_by` of the association.
-      It is used when the association is preloaded.
+    * `:preload_order` - Sets the default `order_by` when preloading the association.
+      It may be a keyword list/list of fields or an MFA tuple, such as `{Mod, fun, []}`.
+      Both cases must resolve to a valid `order_by` expression.
       For example, if you set `Post.has_many :comments, preload_order: [asc: :content]`,
       whenever the `:comments` associations is preloaded,
       the comments will be ordered by the `:content` field.
-      See `Ecto.Query.order_by/3` for more examples.
+      See `Ecto.Query.order_by/3` to learn more about ordering expressions.
+      See the [many_to_many/3 preload order](#many_to_many/3-preload-order) section below to learn how
+      this option can be utilized.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


lib/ecto/schema.ex
```diff
@@ -856,12 +856,15 @@ defmodule Ecto.Schema do
...
-      It is used when the association is preloaded.
+    * `:preload_order` - Sets the default `order_by` when preloading the association.
+      It may be a keyword list/list of fields or an MFA tuple, such as `{Mod, fun, []}`.
+      Both cases must resolve to a valid `order_by` expression.
       For example, if you set `Post.has_many :comments, preload_order: [asc: :content]`,
       whenever the `:comments` associations is preloaded,
       the comments will be ordered by the `:content` field.
-      See `Ecto.Query.order_by/3` for more examples.
+      See `Ecto.Query.order_by/3` to learn more about ordering expressions.
+      See the [many_to_many/3 preload order](#many_to_many/3-preload-order) section below to learn how
```
greg-rychlewski, line 866:
```suggestion
      See [many_to_many/3 preload order](#many_to_many/3-preload-order) to learn how
```

lib/ecto/schema.ex
```diff
@@ -856,12 +856,15 @@ defmodule Ecto.Schema do
...
+    * `:preload_order` - Sets the default `order_by` when preloading the association.
+      It may be a keyword list/list of fields or an MFA tuple, such as `{Mod, fun, []}`.
+      Both cases must resolve to a valid `order_by` expression.
       For example, if you set `Post.has_many :comments, preload_order: [asc: :content]`,
       whenever the `:comments` associations is preloaded,
       the comments will be ordered by the `:content` field.
-      See `Ecto.Query.order_by/3` for more examples.
+      See `Ecto.Query.order_by/3` to learn more about ordering expressions.
+      See [many_to_many/3 preload order](#many_to_many/3-preload-order) to learn how
+      this option can be utilized.
```
greg-rychlewski, line 867:
```suggestion
```

Actually I'm not sure if we should put this here. The focus is really on ordering when there is a join table and I'm a bit worried it will through users off or they will think this can apply to `through` associations`.

sygour: Thanks for reviewing, I was actually wondering if that was a good thing to refer to `many_to_many` documentation

greg-rychlewski:
thank you @sygour !

PR #4468, opened 2024-08-01
remove compile dependency on `:preload_order` MFA in `has_many`

### LINKED ISSUE #4466: `:preload_order` with MFA causes compile-time dependency to referenced module
### Elixir version

Elixir 1.15.6 (compiled with Erlang/OTP 26)

### Database and Version

N/A

### Ecto Versions

3.11.2

N/A

### Current behavior

In a schema definition, using the `:preload_order` on `has_many` with a MFA causes the file to have a compile-dependency to the referenced module.

```ex
schema "foo" do
  # `mix xref trace ...` reports a compile dependency to MyApp.Tag
  has_many :tags, MyApp.Tag, preload_order: {MyApp.Tag, :preload_order, []}
end
```

### Expected behavior

I don't think this is intended, I guess it's probably just a mistake, so I expect no compile-dependency here. I'll try to raise a PR fixing the issue

Gladear:
Expand AST on `has_many` opts. This is already done on `many_to_many` so I figured it shouldn't cause any problem.

Fixes #4466

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1035,6 +1035,7 @@ defmodule Ecto.Schema do
   """
   defmacro has_many(name, schema, opts \\ []) do
     schema = expand_literals(schema, __CALLER__)
+    opts = expand_literals(opts, __CALLER__)
 
     quote do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #4469, opened 2024-08-01
`insert_all` with source (+ update syntax)

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4464

I also think we can make it work with selecting just a source. This could be useful if for instance you are trying to move data to a new table.


greg-rychlewski:
I messed up. `fields` is already translated to `:source`. I'll try to see if there's anything else I can do.

lib/ecto/repo/schema.ex
```diff
@@ -116,6 +116,19 @@ defmodule Ecto.Repo.Schema do
...
+        select_fields = for {{:., _, [{:&, _, [^ix]}, field]}, [], []} <- fields, do: field
+        update_fields = Keyword.keys(args)
+
+        select_fields ++ update_fields
+        |> Enum.map(fn field ->
+          case dumper do
+            %{^field => {source, _, false}} -> source
+            %{} -> raise ArgumentError, "cannot select unwritable field `#{field}` for insert_all"
+            nil -> field
+          end
```
josevalim, line 129: Should we extract this `case` into a `dump_field` or `dump_insert_field` function?

lib/ecto/repo/schema.ex
```diff
@@ -116,6 +116,19 @@ defmodule Ecto.Repo.Schema do
     end
 
     header = case query.select do
+      %Ecto.Query.SelectExpr{expr: {:%{}, [], [{:|, _, [{:&, _, [ix]}, args]}]}, fields: fields} ->
+        select_fields = for {{:., _, [{:&, _, [^ix]}, field]}, [], []} <- fields, do: field
+        update_fields = Keyword.keys(args)
+
+        select_fields ++ update_fields
```
josevalim, line 123: Do we need to solve duplicates somehow?

We need a test for `%{map(p, [:lock_version]) | lock_version: 2}` and make sure the update wins?

josevalim: Btw, whenever possible, I prefer unit tests, because it has a way quicker feedback loop. I am not sure if we are unit testing this functionality though (we might!).

integration_test/cases/repo.exs
```diff
@@ -969,6 +969,80 @@ defmodule Ecto.Integration.RepoTest do
       source = from p in Permalink, select: %{url: p.title, post_id: ^post_id}
       assert {1, _} = TestRepo.insert_all(Permalink, source)
     end
+
+    @tag :upsert_all
+    @tag :with_conflict_target
+    @tag :returning
+    test "insert_all with query and source" do
```
josevalim, line 976: Is this feature related at all to conflict targets? They are orthogonal no? Could it be written in a simpler style, as the test above, or is it harder to test without a upsert?

lib/ecto/repo/schema.ex
```diff
@@ -116,6 +116,19 @@ defmodule Ecto.Repo.Schema do
     end
 
     header = case query.select do
+      %Ecto.Query.SelectExpr{expr: {:%{}, [], [{:|, _, [{:&, _, [ix]}, args]}]}, fields: fields} ->
```
greg-rychlewski, line 119: @josevalim There is an issue with taking `select.fields` because the fields have already been dumped. So the second dump will fail with the field has a `:source` option. And removing the second dump means we don't catch `read_only` fields.

The simplest way I can see to address this is to insert something into the field metadata when it is dumped into the planner [here](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L2242).

Something like:

`{{:., [read_only: true/false], [{:&, [], [ix]}, field]}, [], []}` 

AFAIK this will break some tests but besides that it should be fair game to use the meta like this? We do it for something else [here](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1629). 

But please let me know what you think

josevalim: Yeah, I think it's fine to use meta for that!

greg-rychlewski:
@josevalim thanks for the great feedback. i believe i addressed all of it :)

lib/ecto/repo/schema.ex
```diff
@@ -271,6 +273,22 @@ defmodule Ecto.Repo.Schema do
     {rows, Enum.reverse(cast_params), counter}
   end
 
+  defp insert_all_select_dump!({{:., dot_meta, [{:&, _, [_]}, field]}, [], []}) do
+    if dot_meta[:read_only] == true do
```
josevalim, line 277: Just to confirm: we should check this only on upserts, right? A regular insert should allow writes even to read only columns, right?

greg-rychlewski: Right now we will ignore for even regular inserts with `Repo.insert` without conflict.  I think I understand your point though, you are asking how the data even gets in the table in the first place right?

I'm not 100% what is best. It feels a little odd to me to allow any kind of write. And there are other ways to get the data into the table like through a different schema or creating Ecto on top of an existing database or derived fields. But I think I get your point too.

greg-rychlewski: Maybe we need to change read_only to writable: true/false/insert_only? I think we need to support no inserts for calculated fields

josevalim: Ok, if we don't support them for inserts today, we need to be consistent.

josevalim:
@greg-rychlewski I have been thinking about this PR and I am not really sure if we should support the update the syntax. The biggest issue is that, if you write this:

    select: %{p | foo: 321}

It tells me you are selecting **all* of p and updating the value of `:foo`. The fact we are selecting only a subset is a bit awkward to me, even though it is the most logical. What would we do if the user really wants to select all of them in the future? Sorry for not noticing this earlier, but we can probably keep the other warts of this PR (except the update)?

lib/ecto/repo/schema.ex
greg-rychlewski: @josevalim We haven't released `read_only` yet so I could still change it to something else in another PR. Do you think it's better to change it to an option that supports 3 choices (write allowed, write never allowed, write allowed on inserts)

josevalim: I have a feeling we discussed this but I can’t recall the details. Maybe writeable: :always, :insert, :update or :never?

read_only is nice that it mirrors the repo option but probably not granular enough here.

Thoughts?

greg-rychlewski: We had discussed something similar: read_only, write_only, read_and_write. Though I think we rightly dismissed that idea.

I like this `writeable: :always, :insert, :update or :never`. It makes sense someone might want to insert a value and then never have it be touched again. Though if you agree, I would want to leave out the :update option at the beginning for simplicity and see if someone asks for it. It's a big tougher for me to envision a realistic scenario for it.

josevalim: Agreed with leaving :update out.

greg-rychlewski: So for next steps, how does this sound:

1. I update this PR based on your comment about update syntax but keep the `read_only` option for now
2. After 1 is merged I create a different PR to switch read_only for `writeable` (it will be a decent size PR so prefer splitting)

BTW before we go ahead, I understood your comment about selecting a subset of fields being weird with the update syntax. The only thing that is giving me pause is that we do the same for select queries like `Repo.all`. So would it make sense to support it to be consistent, or did you mean it's particularly weird for `insert`.

josevalim: Perfect!

> The only thing that is giving me pause is that we do the same for select queries like Repo.all. So would it make sense to support it to be consistent, or did you mean it's particularly weird for insert.

I meant it is particularly weird for insert. When I do `select: %{p | foo: 321}`, that returns all fields in `p` AND the updated `foo`. But in the insert it is returning only `foo`. Does it make sense? When we use subqueries we are also returning all fields. We could make it consistent and return all fields, but I don't think that's what people want?

greg-rychlewski: Ah I think I see what you are saying now. Looking at the code, the situation is more nuanced than I thought. 

It seems like we return both fields if the update is on a literal map, but if the update is on a source then we only select the update:

```elixir
TestRepo.insert!(%Barebone{})
    q1 = from b in Barebone, select: %{b | num: ^"1"}
    q2 = from b in Barebone, select: %{%{num: b.num} | num: ^"1"}
    sq1 = from s in subquery(q1), select: s
    sq2 = from s in subquery(q2), select: s

    TestRepo.one(q1, log: :error)
     # 07:20:55.377 [error] QUERY OK source="barebones" db=0.7ms queue=0.3ms
     # SELECT $1 FROM "barebones" AS b0 ["1"]

    TestRepo.one(q2, log: :error)
     # 07:20:55.379 [error] QUERY OK source="barebones" db=0.2ms queue=0.2ms
     # SELECT b0."num", $1 FROM "barebones" AS b0 ["1"]

    TestRepo.one(sq1, log: :error)
     # 07:20:55.380 [error] QUERY OK db=0.1ms queue=0.3ms
     # SELECT s0."num" FROM (SELECT $1 AS "num" FROM "barebones" AS sb0) AS s0 ["1"]
    
    TestRepo.one(sq2, log: :error)
     # ** (Ecto.SubQueryError) the following exception happened when compiling a subquery.
```

And the one that selects both does not work in a subquery. I think that makes sense though because from the parent query there would be no way to choose one or the other since they are both called the same thing. 

I wasn't around so can't be 100% sure, but based on the code it seems intentional that when updating a source the field is dropped:

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L474
https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L1962

greg-rychlewski: I might be missing context but the way I make sense of the above:

1. It makes logical sense to drop the field so that it can be used in subqueries. But definitely there is a bit of a disconnect with the semantics, like you said.
2. I believe the reason it is not also done for literal maps is because the keys of the literal map are not necessarily what you are selecting in the query. So you can't really know what to drop.

josevalim: @greg-rychlewski the regular non-subquery case is pretty much an optimization: we don't bring data if we know it will be replaced. We could apply it to more cases too.

However, this looks like a bug to me:

     TestRepo.one(sq1, log: :error)
     # 07:20:55.380 [error] QUERY OK db=0.1ms queue=0.3ms
     # SELECT s0."num" FROM (SELECT $1 AS "num" FROM "barebones" AS sb0) AS s0 ["1"]

The query `q1 = from b in Barebone, select: %{b | num: ^"1"}` should mean "bring all of `b` and update `num`". If we are bringing only part of `b`, we need to fix that. If this is too hard to fix, we should deprecate this construct in subqueries.

We could make updating map literals in subquery work too, as long as it has the same semantics.

greg-rychlewski: For the sub query case, if we differentiate the two then how is the update to be interpreted? Just as an additional field that can be selected? Or if there is an update when does it happen exactly

josevalim: I would implement it as select all fields of `b` (which I think that's how `select: b` already works) and then replacing the field `num` by the expression given on update. Does that make sense?

greg-rychlewski: FWIW the update syntax to me always looks like we are throwing away the original value, subquery or not. Because the update is part of the select statement of the subquery it feels like it should be evaluated before it’s retuned to the parent. For example if you call a function on some columns you would only expect to get access to the result in the subquery but not the underlying columns.

josevalim: I think we want them to semantically behave as if they were outside of the subquery, right? Which should mirror their Elixir semantics as closely as possible too. So `%{b | num: ...}` should be returning all of `b` with `num`?

greg-rychlewski: > I would implement it as select all fields of b (which I think that's how select: b already works) and then replacing the field num by the expression given on update. Does that make sense?

What you said makes sense but to me it's also consistent with what happens today, namely this:

```elixir
 SELECT s0."num" FROM (SELECT $1 AS "num" FROM "barebones" AS sb0) AS s0 ["1"]
```

So I'm probably not understanding something. What would that query look like if you had to change it?

josevalim: To me it should return this:

    SELECT s0."fielda", so."fieldB", s0."num" FROM (SELECT "fielda", "fieldb", $1 AS "num" FROM "barebones" AS sb0) AS s0 ["1"]

The result above is only correct if "num" is the only field in Barebones! Maybe that's the case, which is why this discussion could be confusing? :D

greg-rychlewski: I have a feeling we might be saying the same thing but I'm not explaining myself properly. So in this PR and also all the subquery stuff, the only thing that happens is the fields being updated are replaced by the update expression.

josevalim: Sounds like a miscommunication indeed. :) And this test shows what I have been saying:

```
    test "supports update in maps" do
      query = from p in Post, select: %{p | text: p.title}
      query = plan(from(subquery(query), [])) |> elem(0)
      assert "%{id: &0.id(), title: &0.title(), text: &0.title()}" =
             Macro.to_string(query.from.source.query.select.expr)
    end
```

So I think we are all good. :)

greg-rychlewski: Ah my bad. I picked bare-bones too try to simplify the discussion but it backfired.

Since insert_all is functioning like a subquery does that also mean you’re ok to use the update syntax there as well?

josevalim: If the insert is inserting all fields in the source, then yes!

greg-rychlewski: oh yes 100% it will always insert all the fields. it's not possible with this change to insert less. it might be this part looks weird:

```elixir
%Ecto.Query.SelectExpr{expr: {:%{}, [], [{:|, _, [{:&, _, [ix]}, args]}]}, fields: fields} ->
    select_fields = for {{:., _, [{:&, _, [^ix]}, _]}, [], []} = expr <- fields, do: insert_all_select_dump!(expr)
    update_fields = args |> Keyword.keys() |> Enum.map(&insert_all_select_dump!(&1, dumper))
    select_fields ++ update_fields
```

the reason for the filtering comprehension is because the planner will replace the field with the update expression in `select.fields`. so we can't extract the field name from it and instead we need to take it from the keys in `args`.

if this sounds ok then the current PR is ready for review

josevalim (approved):

lib/ecto/repo/schema.ex
josevalim: Apologies then, I misunderstood the code, and this was all my confusion. I guess there at least no stones left unturned. :)

greg-rychlewski: No worries at all. It also took me a lot of digging to make sure this is making sense. I could have done a better job pointing out the weird parts.

greg-rychlewski:
@josevalim I'll get the `read_only` -> `writeable` change done this week so we are not stuck not being able to release for long. In case someone asks you to

Issue #4470, opened 2024-08-03
Should we allow disabling of Ecto cache on specific queries?


greg-rychlewski:
### Elixir version

all

### Database and Version

all

### Ecto Versions

all

### Database Adapter and Versions (postgrex, myxql, etc)

all

### Current behavior

I believe all queries are using the Ecto cache with no API no turn it off.

### Expected behavior

This forum post makes me think there is a valid use case to make it configurable per query: https://elixirforum.com/t/ecto-query-cache-seems-to-grow-without-limits/65326

The use case is they are issuing "one and done" kind of queries to do some clean up. These queries are utilizing `values` which is cached based on a count of the number of entries. So each new count creates a new cache entry would could potentially be very big in size (they are saying 200-600 params per query). Since these are big queries that aren't meant to be repeatedly used it might make sense to let them disable the cache?

josevalim:
We should disable the cache for `values` queries unless the values are somehow passed as parameters.

josevalim:
We do something similar for `in` in SQL queries.

PR #4471, opened 2024-08-03
Disable cache for values lists

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4470

The cache key used to be like this `{:values, [], [[num: :integer], 2]}`. Basically we need the types (due to needing to cast the values) and the number of rows in order for the cache to work properly.

This has the potential to produce a lot of cache entries if the types or number of rows varies. In the attached issue we had a report of someone issuing many queries with 200-600 values in them. So each stored query string has the potential to be quite large.

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -521,6 +521,8 @@ defmodule Ecto.Query.API do
   Each field must be given a type or an error is raised. Any type that can be specified in
   a schema may be used.
 
+  Queries using a values list are not cacheable by Ecto.
+
   ## Select example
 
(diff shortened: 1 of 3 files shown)


josevalim (approved):

josevalim:
We should potentially updated documentation to say that it disables the cache.

greg-rychlewski:
@dlindenkreuz I think this was based on your report. Just giving you an FYI

greg-rychlewski:
Also FWIW, the `prepare: unnamed/named` option being discussed in the forum post are related to Postgres. Whether they save prepared statements or not. So wouldn't affect the issue you had. Also `insert` does not have the issue because it does not use the query planner in the same way.

PR #4473, opened 2024-08-05
Fix edge case for insert_all with selecting a source with updates

greg-rychlewski:
I missed an edge case yesterday.

If the user updates with a field from the same source we have to worry about duplicate headers. What happens is the updated field gets moved to the bottom of `select.fields` but keeps its expression. So the for comprehension we are using to filter out updates won't know it's an update.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -125,7 +125,10 @@ defmodule Ecto.Repo.Schema do
                 do: insert_all_select_dump!(expr)
 
           updated_fields = args |> Keyword.keys() |> Enum.map(&insert_all_select_dump!(&1, dumper))
-          unchanged_fields ++ updated_fields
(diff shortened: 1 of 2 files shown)


lib/ecto/repo/schema.ex
```diff
@@ -120,11 +120,19 @@ defmodule Ecto.Repo.Schema do
     header =
       case query.select do
         %Ecto.Query.SelectExpr{expr: {:%{}, [], [{:|, _, [{:&, _, [ix]}, args]}]}, fields: fields} ->
+          {updated_fields, updated_set} =
+            args
+            |> Keyword.keys()
+            |> Enum.map_reduce(MapSet.new(), fn field, set ->
```
josevalim, line 126:
```suggestion
            Enum.map_reduce(args, MapSet.new(), fn {field, _}, set ->
```

josevalim (approved):

PR #4476, opened 2024-08-05
Improve `insert_all` select expression message

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4474

You're right there are too many issues trying to make this work At minimum we'd have to traverse through the select expressions again to find any parameters, even if there are none.

It's not as obvious how to catch the subquery case so will look into that and follow up.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -370,7 +370,11 @@ defmodule Ecto.Query.Planner do
     {schema_or_source, expr, %{select: select} = query} =
       rewrite_subquery_select_expr(query, source?)
 
-    {expr, _} = prewalk(expr, :select, query, select, 0, adapter)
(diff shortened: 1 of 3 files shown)


josevalim (approved):
For the sub query, perhaps we can detect if there is old_params too?

greg-rychlewski:
What do you think about my last commit for subqueries? 

I believe this should work because if there is a mismatch in these two numbers then it should be an error in Ecto. `select_merge` should be the only thing that can erase a select expression from existence, if I'm not missing anything.

I don't think we can port this over to `insert_all` though because we will have the issue that we will have to try extracting the field names from an unbounded amount of nested `:merge` nodes.

josevalim (approved):
It all looks great to me! :heart:

PR #4477, opened 2024-08-08
Always keep map literal in `select_merge` if keys are distinct

greg-rychlewski:
Previously we discussed if `insert_all` can support `select_merge` with multiple map literals when one of the inner maps has interpolations. For example:

```elixir
from p in "posts", select: %{title: ^"title"}, select_merge: %{visits: 1}
```

Because the inner map has an interpolation, we don't know whether it will be wiped out and our query params will be incorrect. So we turn the select expression into `merge(%{...}, %{....})`.  And this merge expression doesn't work with `insert_all` due to difficulty extracting fields from `merge`. 

There is one condition we can be certain the interpolation won't be wiped out: when all the merges have distinct keys.  In this case we can keep the map literal and it will work with `insert_all`.  

This would give a lot more flexibility to user. If they have dynamic keys/values they want to turn into multiple merge statements, they can always remove duplication in Elixir before going to Ecto land.

For example this:

```elixir
conditions = [k1: v1, k2: v2, k3: v3, ...]
base = from p in "posts", select: %{k0: v0}

Enum.reduce(conditions, base, fn {k, v} -> from b in base, select_merge: %{^k => ^v}
```

They can ensure no duplications in `conditions` beforehand.

### DIFF
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -387,19 +387,24 @@ defmodule Ecto.Query.Builder.Select do
         {{:source, meta, ix}, {:source, _, ix}} ->
           {:&, meta, [ix]}
 
-        {{:struct, meta, name, old_fields}, {:map, _, new_fields}} when old_params == [] ->
-          cond do
-            new_fields == [] ->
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Really nice improvement! :heart:

PR #4482, opened 2024-08-11
Add `:allow_stale` option for Repo.{insert, update, delete}

### LINKED ISSUE #4481: Improve support for soft deletes
### Elixir version

All

### Database and Version

PostgreSQL 14+

### Ecto Versions

All

Postgrex

### Current behavior

I am trying to enable soft-delete behaviour in an application, following the rules in this article: https://evilmartians.com/chronicles/soft-deletion-with-postgresql-but-with-logic-on-the-database

Assuming a "item" table with the following field:

```elixir
  add :_deleted_at, :utc_datetime
```

I can trivially add soft deletes like this:

```elixir
    execute """
            CREATE OR REPLACE RULE "phx_sync_soft_deletion" AS ON DELETE TO "items"
            DO INSTEAD UPDATE items SET _deleted_at = NOW() WHERE id = OLD.id RETURNING OLD.*;
            """,
            """
            DROP RULE IF EXISTS phx_sync_soft_deletion ON items;
            """
```

As well as a view that hides all deleted rows:

```
    execute "CREATE OR REPLACE VIEW visible_items AS SELECT * FROM items WHERE _deleted_at IS NULL",
            "DROP VIEW IF EXISTS visible_items"
```

And then I change my Ecto schema to the following:

```elixir
  schema "visible_items" do
```

This means I only see the non soft-deleted rows. But if I want to see them all, I can do `Repo.all({"items", Item})`.

However, this has one issue: because the delete never happen, the underlying `Repo.delete(item)` operation returns zero rows. Ecto then assumes this is happening because the record does not exist, and return an `Ecto.StaleEntryError`. We can workaround this by calling `{:error, _changeset} = Repo.delete(item, stale_error_field: :id)` but it still returns an error tuple.

### Expected behavior

Perhaps we should allow a `allow_stale: true` option? Or `allow_no_return: true`?

I thought about enabling this by default, because a `delete` should be idempotent, but perhaps that's not the best idea. Some applications may want to detect when two deletes race each other?

v0idpwn:
Closes #4481.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1651,6 +1651,8 @@ defmodule Ecto.Repo do
     * `:stale_error_message` - The message to add to the configured
       `:stale_error_field` when stale errors happen, defaults to "is stale".
 
+    * `:allow_stale` - Doesn't error if delete is stale. Defaults to `false`.
+
   See the ["Shared options"](#module-shared-options) section at the module
   documentation for more options.
 
@@ -1847,6 +1849,8 @@ defmodule Ecto.Repo do
     * `:stale_error_message` - The message to add to the configured
       `:stale_error_field` when stale errors happen, defaults to "is stale".
 
+    * `:allow_stale` - Doesn't error if delete is stale. Defaults to `false`.
+
   See the ["Shared options"](#module-shared-options) section at the module
   documentation for more options.
 
@@ -1890,6 +1894,8 @@ defmodule Ecto.Repo do
       `Ecto.StaleEntryError`. Only applies to updates.
     * `:stale_error_message` - The message to add to the configured
       `:stale_error_field` when stale errors happen, defaults to "is stale".
+      Only applies to updates.
+    * `:allow_stale` - Doesn't error if delete is stale. Defaults to `false`.
       Only applies to updates.
 
   See the ["Shared options"](#module-shared-options) section at the module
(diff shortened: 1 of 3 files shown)


lib/ecto/repo.ex
```diff
@@ -1651,6 +1651,8 @@ defmodule Ecto.Repo do
     * `:stale_error_message` - The message to add to the configured
       `:stale_error_field` when stale errors happen, defaults to "is stale".
 
+    * `:allow_stale` - Doesn't error if delete is stale. Defaults to `false`.
```
josevalim, line 1654:
```suggestion
    * `:allow_stale` - Doesn't error if insert is stale. Defaults to `false`.
      This may happen if there are rules or triggers in the database that
      reject the insert operation.
```

lib/ecto/repo.ex
```diff
@@ -1847,6 +1849,8 @@ defmodule Ecto.Repo do
...
```
josevalim, line 1852:
```suggestion
    * `:allow_stale` - Doesn't error if update is stale. Defaults to `false`.
      This may happen if the struct has been deleted from the database before
      the update or if there is a rule or a trigger on the database that rejects
      the update operation.
```

lib/ecto/repo.ex
```diff
@@ -1951,6 +1957,8 @@ defmodule Ecto.Repo do
...
```
josevalim, line 1960:
```suggestion
    * `:allow_stale` - Doesn't error if delete is stale. Defaults to `false`.
      This may happen if the struct has been deleted from the database before
      this deletion or if there is a rule or a trigger on the database that rejects
      the delete operation.
```

josevalim (approved):
Just nits to the docs and ship it :)

PR #4486, opened 2024-08-12
fix extra backtick

mayel:


josevalim:
Already fixed, will push new docs soon. Thx!

PR #4487, opened 2024-08-12
Update Enum.values/2 spec

### LINKED ISSUE #4484: Undocumented or Incorrect dialyzer spec changes
### Elixir version

1.17

### Database and Version

---

### Ecto Versions

3.12.0

postgrex

### Current behavior

mix dialyzer warning:
```
Ecto.Enum.values(MyApp.MySchema, :status)

will never return since the success typing is:
(map(), atom()) :: [atom()]

and the contract is
(map() | Ecto.Schema.t(), atom()) :: [atom()]
```

### Expected behavior

Since the code above still works, it seems like the typespec is incorrect. If it is incorrect, I missed anything regarding this in the Changelog. This is from live view generated code.

wkirschbaum:
fixes https://github.com/elixir-ecto/ecto/issues/4484

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -235,7 +235,7 @@ defmodule Ecto.Enum do
       #=> [:foo, :bar, :baz]
 
   """
-  @spec values(map | Ecto.Schema.t(), atom) :: [atom()]
+  @spec values(module | map, atom) :: [atom()]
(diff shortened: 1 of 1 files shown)


nathany-copia:
Thanks for this fix. 

Now Dialyzer is passing for us too.

PR #4489, opened 2024-08-12
Allow literal struct in `insert_all` + add recent changes to docs

greg-rychlewski:
I was adding the recent changes to the `insert_all` docs and thought we should probably allow literal structs too.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -1496,9 +1496,13 @@ defmodule Ecto.Repo do
 
   ## Examples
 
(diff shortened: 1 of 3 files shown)


josevalim (approved):

josevalim:
@greg-rychlewski after some thought, I am reverting this change. I think a struct is a bit semantically weird because it has all fields. So does it mean you include all fields, considering the struct defaults? 🤔 Since I am shipping v3.12.1, I don't want to include it in.

PR #4490, opened 2024-08-12
remove subquery parameter length check

zachdaniel:
This comes from the check here:

```elixir
    {expr, counter} = prewalk(expr, :select, query, select, 0, adapter)

    if counter != length(select.params) do
      error!(query, "subqueries cannot select_merge onto an existing field that has an interpolation")
    end
```

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/query/planner.ex#L375

I'm working on a fix, but I'm not actually sure what the heuristic here should be. It seems like each subquery "uses" a parameter slot, or something along those lines, which must be accounted for here. However, adding `+ length(select.subqueries)` can't be right, because it doesn't pass both tests.

This currently affects a portion of Ash users.

### DIFF
--- a/test/ecto/query/subquery_test.exs
+++ b/test/ecto/query/subquery_test.exs
@@ -563,14 +563,5 @@ defmodule Ecto.Query.SubqueryTest do
 
       assert Macro.to_string(planned_q.from.source.query.select.expr) == "%{id: ^0, text: ^1}"
     end
-
-    test "overlapping interpolations are allowed" do
-      msg = ~r"subqueries cannot select_merge onto an existing field that has an interpolation"
-
-      assert_raise Ecto.SubQueryError, msg, fn ->
-        q = from c in Comment, select: %{id: ^1}, select_merge: %{id: ^2}
-        from(s in subquery(q)) |> plan()
-      end
-    end
   end
 end
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -370,11 +370,7 @@ defmodule Ecto.Query.Planner do
     {schema_or_source, expr, %{select: select} = query} =
       rewrite_subquery_select_expr(query, source?)
 
-    {expr, counter} = prewalk(expr, :select, query, select, 0, adapter)
-
-    if counter != length(select.params) do
-      error!(query, "subqueries cannot select_merge onto an existing field that has an interpolation")
-    end
+    {expr, _} = prewalk(expr, :select, query, select, 0, adapter)
 
     {{:map, types}, fields, _from} =
       collect_fields(expr, [], :none, query, select.take, true, %{})


greg-rychlewski:
The check there is meant to ensure we have the right number of non-subquery parameters. 

Is the issue you are seeing because `select.params` contains `{:subquery, ix}`? Maybe we just need to filter those outs.

zachdaniel:
It could be, let me try doing that 👍

zachdaniel:
```elixir
    if counter != Enum.count(select.params, &!match?({:subquery, _}, &1)) do
      error!(query, "subqueries cannot select_merge onto an existing field that has an interpolation")
    end
```

That doesn't appear to work. It seems that `counter` is counting params inside of subqueries. I think we need to modify `prewalk` to not include subqueries (probably with a new argument to not break other usages of it).

zachdaniel:
For instance, `select.params` in the failing test is `[{:subquery, 0}]` but the `counter` is returning `2`, I believe because the subquery in the select itself has `2` parameters.

greg-rychlewski:
The only thing I can think is if instead of dong `length(params)` we recursively count the subquery params each time we see `{:subquery, ix}`. 

But then it's a matter of do we want to do this or take a step back and see if there is a better way. If this is breaking existing code we can remove the check and try again later. The check is not crucial it's just mean to provide a better error message instead of the driver complaining about wrong number of query params.

greg-rychlewski:
My feeling is we should remove the check and try again later. Recursively counting the subquery params is quadratic and the prewalker is pretty sensitive. I would be hesitant to change it for this.

@josevalim what do you think?

zachdaniel:
However we resolve it, if we could get a patch release at some point that would be wonderful ❤️ I'm happy to take whatever approach we think should be taken. 

I imagine there is quite a bit of complexity in recursively counting all params, because we have to traverse each type of expression in each key of a query. Although I imagine there may be a utility that does that or something already?

Regardless, happy to help address in whatever way is preferred 👍

greg-rychlewski:
Another thing that might work is resetting the count when we encounter the subquery in the prewalker, the same we do here:

```elixir
{expr, counter} = prewalk(expr, :select, query, select, 0, adapter)
```

But this is invasive enough that I wouldn't be confident pushing it as a quick patch release. Would still prefer getting rid of the check first to fix your issue quickly.

But let's see what Jose says too.

greg-rychlewski:
After looking close, we might be able to do it comparing like this:

```elixir
num_parent_params = length(params)
num_subquery_params =   Enum.reduce(select.subqueries, 0, &(&2 + length(&1.params))))

if counter != num_parent_params + num_subquery_params do
  error!(query, "subqueries cannot select_merge onto an existing field that has an interpolation")
end
```

but I have not been able to think of all possible edge cases atm

josevalim:
@greg-rychlewski we added the check under the assumption the query was going to fail later anyway, no? So it seems some queries were working?

josevalim:
Maybe we are failing on the "zero" case? Where nothing is expected but we are getting something on the other side?

zachdaniel:
I believe the issue in this case is that `select.params` needs to equal the number of parameter references that are *not* in subqueries, but `prewalk` counts the number of params everywhere.

josevalim:
Ok, we should just remove the check then. The cases we want to detect fail anyway, just at the DB layer.

zachdaniel:
I will amend this PR to remove the check then 👍

zachdaniel:
Sorry, my formatter formatted the files 😢 will fix.

test/ecto/query/subquery_test.exs
```diff
@@ -563,14 +563,5 @@ defmodule Ecto.Query.SubqueryTest do
 
       assert Macro.to_string(planned_q.from.source.query.select.expr) == "%{id: ^0, text: ^1}"
```
zachdaniel, line 564: only remaining question is that this test above ^ was added as part of the PR that initially produced this check, but this test does pass. I left this test, but can remove it if desired.

josevalim (approved):

zachdaniel:
I believe the test failure is unrelated to my changes.

greg-rychlewski:
thanks very much for finding this @zachdaniel and apologies for the bug

zachdaniel:
Not a problem at all :) Thanks for the quick response and for all the hard work everyone! 3.12 is a great release :)

PR #4495, opened 2024-08-16
Update Ecto.Changeset.get_assoc/3 typespecs for belongs_to associations

cserino:
For belongs_to associations, it returns nil or a changeset/struct which the previous typespec did not cover.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1689,7 +1689,8 @@ defmodule Ecto.Changeset do
       [%Post{id: 1, title: "world"}]
 
   """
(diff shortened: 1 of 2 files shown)


josevalim (approved):

greg-rychlewski:
thanks very much

warmwaffles:
This just came up with dialyzer complaining that only `[t()]` was possible. Is there a timeline on when this gets sweeped up into a patch?

josevalim:
There are some prefix related fixes on the pipeline, so as soon as that is addressed, we can ship a new one.

PR #4496, opened 2024-08-18
Load context into struct metadata

JesseStimpson:
The goal of the change is to have the :context track along with the schema metadata so that it can be used by an adapter for its own purposes.

The :prefix option already has good semantics for this, except that it is restricted to String.t() and subjected to is_binary guards. We can certainly duplicate the way prefix is treated with the context field, but this would result in a large changeset (particularly because we would need `%Query{context: ...}` or similar.

As is, this PR is a bit of a compromise in that it loads the context onto the schema on the basic operations, but does not attempt to duplicate all of the Query's prefix semantics. (subqueries, etc)

### DIFF
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -54,9 +54,9 @@ defmodule Ecto.Repo.Queryable do
           raise Ecto.QueryError, query: query, message: "preloads are not supported on streams"
         end
 
-        preprocessor = preprocessor(from, preprocess, adapter)
+        preprocessor = preprocessor(from, preprocess, adapter, opts)
         stream = adapter.stream(adapter_meta, query_meta, prepared, dump_params, opts)
-        postprocessor = postprocessor(from, postprocess, take, adapter)
+        postprocessor = postprocessor(from, postprocess, take, adapter, opts)
 
         stream
         |> Stream.flat_map(fn {_, rows} -> rows end)
@@ -228,9 +228,9 @@ defmodule Ecto.Repo.Queryable do
           from: from
         } = select
 
-        preprocessor = preprocessor(from, preprocess, adapter)
+        preprocessor = preprocessor(from, preprocess, adapter, opts)
         {count, rows} = adapter.execute(adapter_meta, query_meta, prepared, dump_params, opts)
-        postprocessor = postprocessor(from, postprocess, take, adapter)
+        postprocessor = postprocessor(from, postprocess, take, adapter, opts)
 
         {count,
          rows
@@ -239,69 +239,71 @@ defmodule Ecto.Repo.Queryable do
     end
   end
 
-  defp preprocessor({_, {:source, {source, schema}, prefix, types}}, preprocess, adapter) do
-    struct = Ecto.Schema.Loader.load_struct(schema, prefix, source)
+  defp preprocessor({_, {:source, {source, schema}, prefix, types}}, preprocess, adapter, opts) do
+    schema_meta = %{source: source, prefix: prefix, context: Keyword.get(opts, :context)}
+    struct = Ecto.Schema.Loader.load_struct(schema, schema_meta)
 
     fn row ->
       {entry, rest} = struct_load!(types, row, [], false, struct, adapter)
-      preprocess(rest, preprocess, entry, adapter)
+      preprocess(rest, preprocess, entry, adapter, opts)
     end
   end
 
-  defp preprocessor({_, from}, preprocess, adapter) do
+  defp preprocessor({_, from}, preprocess, adapter, opts) do
     fn row ->
-      {entry, rest} = process(row, from, nil, adapter)
-      preprocess(rest, preprocess, entry, adapter)
+      {entry, rest} = process(row, from, nil, adapter, opts)
+      preprocess(rest, preprocess, entry, adapter, opts)
     end
   end
 
-  defp preprocessor(:none, preprocess, adapter) do
+  defp preprocessor(:none, preprocess, adapter, opts) do
     fn row ->
-      preprocess(row, preprocess, nil, adapter)
+      preprocess(row, preprocess, nil, adapter, opts)
     end
   end
 
-  defp preprocess(row, [], _from, _adapter) do
+  defp preprocess(row, [], _from, _adapter, _opts) do
     row
   end
 
-  defp preprocess(row, [source | sources], from, adapter) do
-    {entry, rest} = process(row, source, from, adapter)
-    [entry | preprocess(rest, sources, from, adapter)]
+  defp preprocess(row, [source | sources], from, adapter, opts) do
+    {entry, rest} = process(row, source, from, adapter, opts)
+    [entry | preprocess(rest, sources, from, adapter, opts)]
   end
 
-  defp postprocessor({:any, _}, postprocess, _take, adapter) do
+  defp postprocessor({:any, _}, postprocess, _take, adapter, opts) do
     fn [from | row] ->
-      row |> process(postprocess, from, adapter) |> elem(0)
+      row |> process(postprocess, from, adapter, opts) |> elem(0)
     end
   end
 
-  defp postprocessor({:map, _}, postprocess, take, adapter) do
+  defp postprocessor({:map, _}, postprocess, take, adapter, opts) do
     fn [from | row] ->
-      row |> process(postprocess, to_map(from, take), adapter) |> elem(0)
+      row |> process(postprocess, to_map(from, take), adapter, opts) |> elem(0)
     end
   end
 
-  defp postprocessor(:none, postprocess, _take, adapter) do
-    fn row -> row |> process(postprocess, nil, adapter) |> elem(0) end
+  defp postprocessor(:none, postprocess, _take, adapter, opts) do
+    fn row -> row |> process(postprocess, nil, adapter, opts) |> elem(0) end
   end
 
-  defp process(row, {:source, :from}, from, _adapter) do
+  defp process(row, {:source, :from}, from, _adapter, _opts) do
     {from, row}
   end
 
-  defp process(row, {:source, {source, schema}, prefix, types}, _from, adapter) do
-    struct = Ecto.Schema.Loader.load_struct(schema, prefix, source)
+  defp process(row, {:source, {source, schema}, prefix, types}, _from, adapter, opts) do
+    schema_meta = %{source: source, prefix: prefix, context: Keyword.get(opts, :context)}
+    struct = Ecto.Schema.Loader.load_struct(schema, schema_meta)
     struct_load!(types, row, [], true, struct, adapter)
   end
 
-  defp process(row, {:source, :values, _prefix, types}, _from, adapter) do
+  defp process(row, {:source, :values, _prefix, types}, _from, adapter, _opts) do
     values_list_load!(types, row, [], true, adapter)
   end
 
-  defp process(row, {:merge, left, right}, from, adapter) do
-    {left, row} = process(row, left, from, adapter)
-    {right, row} = process(row, right, from, adapter)
+  defp process(row, {:merge, left, right}, from, adapter, opts) do
+    {left, row} = process(row, left, from, adapter, opts)
+    {right, row} = process(row, right, from, adapter, opts)
 
     data =
       case {left, right} do
@@ -338,18 +340,18 @@ defmodule Ecto.Repo.Queryable do
     {data, row}
   end
 
-  defp process(row, {:struct, struct, data, args}, from, adapter) do
-    case process(row, data, from, adapter) do
+  defp process(row, {:struct, struct, data, args}, from, adapter, opts) do
+    case process(row, data, from, adapter, opts) do
       {%{__struct__: ^struct} = data, row} ->
-        process_update(data, args, row, from, adapter)
+        process_update(data, args, row, from, adapter, opts)
 
       {data, _row} ->
         raise BadStructError, struct: struct, term: data
     end
   end
 
-  defp process(row, {:struct, struct, args}, from, adapter) do
-    {fields, row} = process_kv(args, row, from, adapter)
+  defp process(row, {:struct, struct, args}, from, adapter, opts) do
+    {fields, row} = process_kv(args, row, from, adapter, opts)
 
     case Map.merge(struct.__struct__(), Map.new(fields)) do
       %{__meta__: %Ecto.Schema.Metadata{state: state} = metadata} = struct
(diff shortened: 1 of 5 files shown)


greg-rychlewski:
So I understand the use case better, what situations is this handling that [Ecto.put_meta/2](https://hexdocs.pm/ecto/Ecto.html#put_meta/2) is not?

JesseStimpson:
> So I understand the use case better, what situations is this handling that [Ecto.put_meta/2](https://hexdocs.pm/ecto/Ecto.html#put_meta/2) is not?

Pulling in some details from the [mailing list](https://groups.google.com/g/elixir-ecto/c/fM0r13D4gHQ):

In my adapter for FoundationDB I've been misusing the :prefix option to pass an opaque data structure from the caller to my adapter. (Specifically this is a tuple of the type `{:erlfdb_tenant, reference()}`, which is the handle to an open FDB Tenant).

It seems that the :context may be a more appropriate way of doing this, which is necessary as of 3.12, since the :prefix is now subjected to is_binary checks.

However, there is a nice feature of the :prefix that isn't present with :context.

```
      tenant = open_tenant()
      user = TestRepo.get(User, id, context: tenant)
      assert ^tenant = Ecto.get_meta(user, :context) # fail
```

The assertion here fails, with the :context entry being nil. If you do something similar with :prefix, it works as expected.

---

This change would allow the caller to carry the context on the schema struct without an explicit call to Ecto.put_meta/2.

greg-rychlewski:
One important difference between prefix and this is with preloads. Each preloaded schema can define `@schema_prefix` to ignore the one given to the repo operation. Or `prefix` can be given individual `join`s. But with this it would just apply in a blanket manner to everything. 

For the example you gave, it seems this is meant to be a convenience using an option on the repo rather than having to call a separate function afterwards. Is that the main advantage or is there something else?

```elixir
tenant = open_tenant()
user = TestRepo.get(User, id, context: tenant)
assert ^tenant = Ecto.get_meta(user, :context) # fail
```

JesseStimpson:
My adapter isn't attempting to use context as an option, but rather as the context in which a particular schema struct is relevant.

FoundationDB has many differences to the more commonly supported DBs, so I am happy to answer any specifics, but I'll try to avoid too much detail. :)

In the FDB adapter, the `tenant` object is a resource that's acquired by the caller during runtime and used across several transactions. At a high level it can be considered an entirely separate database to achieve multi-tenancy.

When writing a schema to a particular tenant, we wish to consider that schema struct as intrinsically tied to that tenant, so that future operations (e.g. Changeset) remain tied to that tenant. This protects the caller from accidentally writing a struct from one tenant into another tenant. (To do so the caller should explicitly remove the context from the struct metadata)

It's true that we could force the caller to call put_meta on each struct, but it would be prone to bugs and vulnerabilities.

I believe this use case to be similar to one of the prefix use cases: a way to implement multi-tenancy. However, my `tenant` is not a string. Instead, it can be considered a token (opaque data structure) to the tenant. This allows the developer to centralize some of their authorization checks to acquire that tenant resource, and have confidence that the tenancy won't be broken.

greg-rychlewski:
In your changes it is not obvious to me how this is handled:

>One important difference between prefix and this is with preloads. Each preloaded schema can define @schema_prefix to ignore the one given to the repo operation. Or prefix can be given individual joins. But with this it would just apply in a blanket manner to everything.

Could you perhaps write a test where you are retrieving data from a query like this `from schema_a, join: schema_b` where `schema_a` and `schema_b` are given different contexts?

>It's true that we could force the caller to call put_meta on each struct, but it would be prone to bugs and vulnerabilities.

I might be misunderstanding your use case, could you correct me if I have the wrong idea:

1. You are proposing the user passes a correct context into the repo call where they first fetch the data. Ecto will insert the context into the schema metadata automatically.
2. Currently the user needs to pass the correct context to `Ecto.put_meta/2` after they first fetch the data. 

It is not too clear to me how one is more error prone than the other. If there is more to this, like your adapter is using the `context` during the first repo call then that use case makes sense to me. But if it's just about how the metadata is first being put into the struct, that is less clear to me.

greg-rychlewski:
> I didn't attempt to cover this use case yet

No worries, the smaller initial PR is appreciated. 

I am basically just trying to figure out the right path forward. Because `context` is a schema property, if we were to put it as an option in the `Repo` functions then it would be a bit of a problem for it to always apply to every schema in the query. For this use case it might not be a problem, but it wouldn't really be a general solution or in line with the way other schema properties behave, like `prefix`.

>Indeed, the adapter uses the context on the first repo call. In my use case, the context is not a property of the Schema definition, but rather a property of the struct during runtime.

This part I'm still having trouble understanding, apologies. If it is not a property of the schema then what would you adapter to with `context` in this case:

```elixir
tenant = open_tenant()
user = TestRepo.get(User, id, context: tenant)
```

Since the struct does not exist until after the database operation has finished, what is your adapter doing with `context` during this call?

JesseStimpson:
> the way other schema properties behave, like `prefix`.

It's worth noting that I've successfully been using `prefix` for this exact purpose up until Ecto 3.12. In 3.11, the type was still String.t(), so my adapter could be considered out of spec, but there were no is_binary guards, at least not in the code paths my adapter exercised. There are guards in 3.12 that have broken my adapter. I was out of spec, so no complaints from me here.

In trying to resolve, I want context to behave exactly the way prefix does, but without the string requirement. I am stopping short of requesting that prefix remove the string requirement in the Ecto API, because this appears to be a design choice.

> If it is not a property of the schema

Hm, terminology may be failing us here. If I may adopt an OOP analogy: in my adapter, the `context: tenant` is a property of the Schema object, not of the Schema class.

> Since the struct does not exist until after the database operation has finished, what is your adapter doing with `context` during this call?

The adapter requires the`context: tenant` to issue a transaction on the database. If the record is found, the resulting struct should have `context: tenant` in the metadata for future operations.

```elixir
# Adapter pseudo-code. Background: FDB is a key-value store which maintains a separate
# keyspace for each tenant. Tenant A can have /users/id123 and Tenant B can have /users/id123,
# and the values are independent.
FoundationDB.transaction(tenant,
  fn tx ->
    case :erlfdb.get(tx, "/users/#{id}") do
      :not_found ->
        nil
      data ->
        {:ok, data}
  end)
```

If I were able to translate the above into SQL, it would look something like this
```elixir
RDBMS.exec("""
SELECT * FROM #{prefix}.users
WHERE id = $1
""", [id])
```

greg-rychlewski:
Thanks for answering all my questions. Apologies for being so unfamiliar with your db.

So to me it sounds like you are using context to direct your queries to the right place. Which is basically what prefix does. 

I don't think we want to turn context into an alternate prefix. As you alluded to the changes would be quite large if we want behaviour to be general/consistent. 

So then the question is do we open up `prefix` to non-strings or I think Jose was alluding to something like an adapter callback that can take the metadata you put into the repo options and then do whatever it wants to with it on the result of the query. At least I think that's what he meant.

Since Jose responded to you in the mailing list let's get his opinion on this too. At the moment I can't remember why we made prefixes strings only. I will have to go back and look.

JesseStimpson:
Thanks for sticking in there with me @greg-rychlewski ! 🤝 

By all means I'm open to other clever solutions here.

I admit I'm confused by what Ecto considers to be the high level difference between prefix and context.

From [Ecto.Schema.Metadata](https://hexdocs.pm/ecto/Ecto.Schema.Metadata.html):

On `:context`, the docs have this:

> The `:context` field represents additional state some databases require for proper updates of data.

On `:prefix`:

> Tracks the source prefix in the data storage.

Naively, I read this as `:prefix` being a subset of `:context`. But there may be some nuance I'm missing.

greg-rychlewski:
It was before my time, but just looking at the code `context` is quite limited compared to `prefix`. There is a lot of logic for ensuring `prefix` can be handled individually across different schema in a query. And setting precedence rules across different specifications.

For `context` it seems like it was meant for the user to add something manually onto a previously fetched struct.

Schultzer:
@JesseStimpson have you explored using the process dictionary to store this information?

You would be able to sidestep Ecto completely. But I also see the value in leveraging context, although I can’t remember seeing any adapter currently using it.

josevalim:
The context was added precisely to track schema information across insert to update and delete, such as a tenant, and I think it is reasonable to extend it to queries.

However, I think the similarities with prefixes is worth discussing. Prefixes have very fine grained control. You can set them per query, per join, per preload. @JesseStimpson, can you have a query where the context is different between a join or a preload? If so, using prefixes may be the best option indeed, as you will have more control, as the `:context` would rather apply to all equally.

So our options are:

1. Keep :prefix as a binary and FoundationDB uses it. @JesseStimpson, although the tenant is a reference today, could it be a name as well? Maybe the node name the tenant is running on?
2. We remove the restriction that prefix is a binary
3. We add context as this PR

greg-rychlewski:
My main comment about keeping the changes in the PR is that it feels a bit inconsistent to say `:context` applies to everything and `:prefix` has fine grained controls. Since they are both just schema metadata I don't personally see why one should behave different than the other. 

Let's say Foundation DB doesn't need that control but then later on Foundation DB 2 did. Then I feel they would be right to question the difference in behaviour. And then we would basically need to re-implement `prefixes` again but with a different name.

josevalim:
That's a good point. Shall we just revert the binary checks from prefix then?

greg-rychlewski:
That's my feeling but I couldn't remember why we forced them to be binary. Was going to take a look at our old convos to make sure we won't get bit some other way.

JesseStimpson:
> have you explored using the process dictionary to store this information?

This wouldn't be ideal, I think. The caller would have to be aware of this behavior when they send a struct to a different process.

> can you have a query where the context is different between a join or a preload?

I'd like to support this someday in the FDB adapter. It is not currently implemented, but it's a perfectly reasonable query.

> although the tenant is a reference today, could it be a name as well?

The tenant is identified by a string name, but it's important to me to keep the semantics of "acquiring the tenant resource" so that an application developer can centralize authorization checks.

I have been thinking on how to have my cake (string) and eat it too (resource-based). It's possible but a bit awkward as I would have to maintain a registry for "open tenants" and then figure out how to garbage collect them. Currently I am taking advantage of ref counting on enif_resource to which the reference points.

That being said, if a change to the Ecto API is not possible, I believe I would rather choose to use context as is, and work around the current shortcomings.

> Shall we just revert the binary checks from prefix then?

I would love this ❤️ 😄 ! If I may, I'd also request changing the type from `String.t()` to a type that supports some kind of container `String.t() | map()`, `String.t() | tuple()`, `any()`, etc.

josevalim:
> I have been thinking on how to have my cake (string) and eat it too (resource-based). It's possible but a bit awkward as I would have to maintain a registry for "open tenants" and then figure out how to garbage collect them. 

How many tenants can you have? Are they are always opened and closed dynamically? Keeping open connections in Erlang/Elixir is very cheap, so if the number is limited (hundreds and not dozens of thousands), wouldn't it be more efficient to keep them open (unless FDB itself is not happy with this arrangement).

JesseStimpson:
I love the opportunity to evangelize FDB :)

It may be important for me to note that FDB is a kv store with ACID transactions, and the EctoFDB adapter takes an opinionated stance on how data is laid out in the keyspace ([Layer Concept](https://apple.github.io/foundationdb/layer-concept.html)). I'll just focus on FDB itself when discussing limitations, so I can remove any opinions I've baked into EctoFDB.

In FDB, when a tenant is created, it's internally assigned an 8-byte integer. When a tenant is opened, FDB guarantees that any transaction created from it can only access keys that are prefixed with that 8-byte integer. This prefix is an implementation detail internal to FDB and **the prefix itself is not exposed to the database client**.

> How many tenants can you have?

At the FDB storage level, since a tenant is simply an 8-byte prefix to a key, there is effectively no limit on the number of tenants. The total number of keys will be limited by the hardware on which the FDB cluster runs. (Billions of keys)

How many tenants can you have open from one client? I'm not aware of any limit other than system memory. Each open tenant is **not a connection** to the database. Instead, consider it a rule imposed on any transactions created from that tenant. I don't know of any reason why a sufficient system would not be able to open 100k+ concurrent tenants.

At the FDB C API, I believe the "open tenant" is just a C struct.

> Are they are always opened and closed dynamically?

Yes, the tenant must be opened before transacations can be created from it. However, the use of tenants itself is optional in FDB. One can write to the keyspace without the concept of a tenant. [ In EctoFDB I've chosen to require a tenant, at least for now ]

All that being said, my adapter already prefers to keep many tenants open long term. For example, when a LiveView mounts, we can check that the client is authorized to access a certain tenant, and open it. That tenant can be used for the lifetime of the LiveView. When the LiveView is terminated, the open tenant will be garbage collected.

Maybe some example code will help demonstrate. I believe this is the same way a Postgres app would use prefix, except my prefix is fundamentally not a string.

```elixir
defmodule MyAppWeb.InitAssigns do
  def on_mount(:user, _params, session, socket) do
    user_identity = get_user_identity(session)

    if user_identity do

      # ** centralized authorization check for acquiring the tenant resource **
      tenant = MyApp.Repo.open_tenant(user_identity.id)
      
      {:cont,
       socket
       |> assign(:tenant, tenant)}
    else
      {:halt,
       socket
       |> put_flash(:error, "You must sign in")}
    end
  end
end
```

```elixir
# router
  live_session :authenticated, on_mount: {MyAppWeb.InitAssigns, :user} do
    scope "/", MyAppWeb do
      pipe_through :browser
      live "/dashboard", DashboardLive, :home
      # ...
    end
  end
```

```elixir
defmodule MyAppWeb.DashboardLive do
  def mount(_params, _session, socket) do

    # ** retrieves all notifications for a given tenant **
    notifications =
        MyApp.Repo.all(Notification, context: tenant) # or prefix: tenant

    {:ok, socket |> assign(:notifications, notifications}
  end

  def handle_event("mark_all_read", _params, socket) do
    for n <- socket.assigns.notifications do
      changeset = Notification.changeset(n, %{state: :read})

      # ** tenant is in metadata, does not need to be specified **
      MyApp.Repo.update(changeset)

    end
    {:noreply, socket |> assign(:notifications, [])}
  end
end
```

You can probably ignore my earlier comment about a registry for tenants by string. I did a poor job explaining it, and it's not something I would choose to implement due to the complexity. I suppose I was simply trying to get the message across that my desired `:prefix`/`:context` is not a string.

One thing I don't want to do is change my adapter to this

```elixir
MyApp.Repo.get(Notification, prefix: user_identity.id) # id is a string
MyApp.Repo.insert(notif, prefix: user_identity.id) # id is a string
...
```

This style would impose that the adapter internally opens the tenant for each transaction. It's important to my vision of the EctoFDB adapter that the application developer is responsible for opening tenants. (Authorization checks and on-line Migrations)

josevalim:
Alright, thank you! @greg-rychlewski said he would look into blockers to reverting prefix to non-binary, so if there is nothing, we are happy to revert it :)

greg-rychlewski:
There's nothing that stands out to me from the conversation around the change. It seemed mostly like standardizing around the way `Ecto.Query.put_query_prefix` works.

But maybe to be safe when making the change, it would be good to have tests for everything with non-binary prefix:

- @schema_prefix
- Ect.Query.put_query_prefix
- :prefix option on repo
- :prefix option on from/join

We also need to change migrations in Ecto SQL to accept non-strings.

josevalim:
I think it is fine for Ecto SQL to expect and impose strings. The rationale is that it should be an adapter concern, not an Ecto one.

@JesseStimpson would you please send a PR? :)

JesseStimpson:
@josevalim @greg-rychlewski Thanks for being open to this change. I'll work on a new PR for it. 

What would you like to do with this PR about the context? I will no longer being using :context in my adapter with the change to :prefix, so I have no need for it, and leave it to your judgment.

josevalim:
I believe we can close this then and reopen if there is a need. :)

PR #4497, opened 2024-08-21
Support non-string prefix

JesseStimpson:
Discussed in elixir-ecto/ecto#4496

These changes allow an adapter to support a `:prefix` and `@schema_prefix` of any type, rather than being restricted to String.t().

1. At compile time, `from` and `join` query portions still require either string literal or interpolated value. The interpolated value can be anything. Additional literals can be supported here if desired. (from.ex:87 and join.ex:146)
2. Happy to add more tests, let me know if you see gaps
3. I also verified ecto_sql's tests still pass

Thanks!

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -882,7 +882,7 @@ defmodule Ecto.Query do
     subquery = wrap_in_subquery(query)
 
     case Keyword.fetch(opts, :prefix) do
-      {:ok, prefix} when is_binary(prefix) or is_nil(prefix) ->
+      {:ok, prefix} ->
         put_in(subquery.query.prefix, prefix)
 
       :error ->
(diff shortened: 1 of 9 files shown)


josevalim (approved):

josevalim:
We should probably send a PR to `ecto_sql` that validates the string to be a prefix (or nil) on that side with a clear error message. :)

greg-rychlewski:
> At compile time, from and join query portions still require either string literal 

If we are saying adapters control the prefix then what would be the issue with `from p in Post, prefix: 123`

josevalim:
We can probably allow it to be "any literal" but I think it is ok to be a bit more conservative when relaxing those rules. WDYT?

JesseStimpson:
The only downside I see to expanding the allowed literals is that the error
message has to change in lockstep. Since there are so many ecto_sql users,
I was hesitant to change the error message and possibly increase confusion
for ecto_sql users.

greg-rychlewski (approved):

greg-rychlewski:
Thanks for the change. I understand wanting to be conservative about the literals for now.

PR #4503, opened 2024-08-27
Add `ecto_query: :preload` option to preload query

andrewtimberlake:
This pull request, based on [this discussion](https://groups.google.com/g/elixir-ecto/c/Y_S6u0WWecc), adds an option to a preload query so that they can be filtered out when using [Multi tenancy with foreign keys](https://hexdocs.pm/ecto/multi-tenancy-with-foreign-keys.html#adding-org_id-to-read-operations)

This will also change migrations to pass `opts[:ecto_query] = :schema_migration` instead of `opts[:schema_migration] = true`

### DIFF
--- a/guides/howtos/Multi tenancy with foreign keys.md
+++ b/guides/howtos/Multi tenancy with foreign keys.md
@@ -27,7 +27,7 @@ defmodule MyApp.Repo do
   @impl true
   def prepare_query(_operation, query, opts) do
     cond do
-      opts[:skip_org_id] || opts[:schema_migration] ->
+      opts[:skip_org_id] || opts[:ecto_query] ->
         {query, opts}
 
       org_id = opts[:org_id] ->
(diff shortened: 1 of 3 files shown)


guides/howtos/Multi tenancy with foreign keys.md
```diff
@@ -27,7 +27,7 @@ defmodule MyApp.Repo do
   @impl true
   def prepare_query(_operation, query, opts) do
     cond do
-      opts[:skip_org_id] || opts[:schema_migration] ->
+      opts[:skip_org_id] || opts[:ecto_query] ->
```
josevalim, line 30:
```suggestion
      opts[:skip_org_id] || opts[:ecto_query] in [:schema_migrations, :preload] ->
```

josevalim: So if we add more types in the future, it doesn't mean they get automatically skipped.

josevalim:

guides/howtos/Multi tenancy with foreign keys.md
andrewtimberlake: It should be in [:schema_migration] (singular)

PR #4507, opened 2024-09-09
Update examples for `reverse_order/1` to correctly refer to `order_by/1` instead of `order/1`

rmoorman:


josevalim:
Btw, the phoenix_html bug you reported is still in my inbox, I am just short on time :)

rmoorman:
Good to know! Seeing a lot of great stuff happening all around regarding elixir land (elixir conf, livebook, lv native to name a few)  so I can imagine🫀

PR #4516, opened 2024-10-01
`apply_changes/1`: convert nested schemaless validations

### LINKED ISSUE #4514: `apply_changes/1` does not convert nested schemaless validation changeset
### Elixir version

1.16.3

### Database and Version

13.16

### Ecto Versions

3.12.3

0.19.1

### Current behavior

Sometimes it's useful to validate maps without a schema, to avoid having to define another embedded struct or to do dynamic validations:
```elixir
defmodule MySchema do
  use Ecto.Schema
  import Ecto.Changeset

  @primary_key false
  embedded_schema do
    field :extended, :map
  end

  def changeset(struct \\ %__MODULE__{}, params) do
    struct
    |> cast(params, [:extended])
    |> update_change(:extended, fn extended ->
      {%{}, %{foo: :integer, bar: :integer}}
      |> cast(extended || %{}, [:foo, :bar])
      |> validate_required([:bar])
    end)
  end
end

test "apply_changes/1 for nested schemaless changeset" do
  params = %{"extended" => %{"foo" => 1, "bar" => 2}}
  changeset = MySchema.changeset(%MySchema{}, params)

  %MySchema{extended: %{foo: 1, bar: 2}} = apply_changes(changeset)
end
```

### Expected behavior

`apply_changes/1` should convert the nested changeset to a map with atom keys.

azizk:
Fixes #4514

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2208,6 +2208,9 @@ defmodule Ecto.Changeset do
         {:ok, {tag, relation}} when tag in @relations ->
           apply_relation_changes(acc, key, relation, value)
 
+        {:ok, :map} when is_struct(value, Changeset) ->
+          Map.put(acc, key, apply_changes(value))
+
         {:ok, _} ->
           Map.put(acc, key, value)
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2190,6 +2190,9 @@ defmodule Ecto.Changeset do
   regardless if the changeset is valid or not. See `apply_action/2`
   for a similar function that ensures the changeset is valid.
 
+  If a field of type `:map` contains a schemaless changeset,
+  its changes too will be applied and returned as a map with atom keys.
+
   ## Examples
 
       iex> changeset = change(%Post{author: "bar"}, %{title: "foo"})
--- a/test/ecto/changeset_test.exs
+++ b/test/ecto/changeset_test.exs
@@ -1073,6 +1073,23 @@ defmodule Ecto.ChangesetTest do
     assert changed_post.category == nil
   end
 
+  test "apply_changes/1 with nested schemaless validation" do
+    params = %{"seo_metadata" => %{"keywords" => ["foo", "bar"], "slug" => "my-post-1"}}
+
+    changeset =
+      %Post{}
+      |> changeset(params)
+      |> update_change(:seo_metadata, fn seo_metadata ->
+        {%{}, %{keywords: {:array, :string}, slug: :string}}
+        |> cast(seo_metadata || %{}, [:keywords, :slug])
+        |> validate_required([:keywords, :slug])
+      end)
+
+    changed_post = apply_changes(changeset)
+
+    assert changed_post.seo_metadata == %{keywords: ["foo", "bar"], slug: "my-post-1"}
+  end
+
   describe "apply_action/2" do
     test "valid changeset" do
       post = %Post{}


josevalim (approved):

greg-rychlewski:
Sorry if I am misunderstanding. But was the intent of the original issue to be able to validate maps before persisting? If so will this not ignore the validation

josevalim:
Good point, we will need to carry those accordingly and also document the feature, since this is specific to apply changes, we don’t handle changesets in maps otherwise.

greg-rychlewski:
@azizk Do you want to take that on? If not we can probably merge this and I could tackle the other stuff.

azizk:
Yeah, I suspected something else was missing to make this a complete feature. I think I'd rather hand it over to you. Too many functions to look at. :sweat_smile:

Thanks!

greg-rychlewski:
No worries. Thank you for this change. @josevalim are you ok to merge just this one for now?

greg-rychlewski:
Or maybe I am getting ahead of myself. I have doubts now which one of these 2 we are going with 

1. we only ever make the change for apply_changes and document it 
2. We also try to do it for persistence. I.e resolve the value or report an error during repo operations 

the second point sounds like a good feature to me but I haven’t evaluated whether it’s possible

josevalim:
The best would be likely to support this everywhere, because this solves the problem of us supporting nested schemaless changesets, but I am unsure how practical it is, especially if it will require us special casing map types everywhere.

warmwaffles:
> special casing map types everywhere.

Sounds like introducing a schema 😉 

I haven't come across the need to do nested schemaless changesets. I would think if you reach for nested schemaless, you should probably define a schema because it's about to get complex.

greg-rychlewski:
One thing I can see is if you have one or a few functions that do some basic validations on maps that can be reused. Like a function that takes in a list of keys and checks if they are present. As opposed to creating separate schemas for every type of map you might want to do this on.

greg-rychlewski:
sorry @azizk. so to close this out would you be able to add into the docs for `apply_changes/1` that we allow this kind of schemaless changes? and then we'll take it away to see if it makes sense to extend it further

greg-rychlewski (approved):
thank you!

azizk:
@greg-rychlewski
> One thing I can see is if you have one or a few functions that do some basic validations on maps that can be reused. Like a function that takes in a list of keys and checks if they are present. As opposed to creating separate schemas for every type of map you might want to do this on.

That's a good example. I'd like to mention another one that came to me yesterday. It's about saving storage space.

I have been working on a work project dealing with OpenRTB. I have an embedded schema for the `Bid` object which has a lot of fields. When I store a `Creative` schema I only want to store a limited number of fields from the `Bid` embedded schema. So I define `field :bid, :map` instead of `embeds_one :bid, Bid`, because otherwise Ecto stores all properties from `Bid` even if the values are `nil`. I still want to validate the `bid` map field in the `Creative` changeset function to be sure and to avoid future surprises.

What I do atm is to use `validate_change/3` and add an error if the `Bid` changeset is invalid. 

Initially we wanted to allow schemaless changesets, but maybe it's worth thinking about allowing any kind of changeset. Otherwise it becomes necessary – in this case at least – to convert a schema changeset to a schemaless changeset (I guess by doing `Map.update!(changeset, :data, &Map.from_struct/1`?).

PS: I didn't want to open a new issue, so commented here, if that's okay.

greg-rychlewski:
@azizk Jose had a good idea where custom types can be utilized. I'm not sure if you saw this before but you could basically make your own type and control these things through the `cast` callback. And you could even make the type parameterized to make it generalizable.

Let us know what you think:

https://hexdocs.pm/ecto/Ecto.Type.html
https://hexdocs.pm/ecto/Ecto.ParameterizedType.html

greg-rychlewski:
As one example, you could get very close to mimicking `cast_embed` by making a parameterized type like `MyValidatedMap` that takes a `:with` parameter. That parameter is a function that does all the changeset functions and then as an extra step at the end checks if the changeset is valid and then applies the changes or returns an error.

azizk:
Hi, Greg (@greg-rychlewski)!

Been a long time but I still wanted to follow up here.

While Ecto’s casting mechanism is powerful for form-based, database-bound data, I've come to believe it's not the best fit for (strict) JSON validation. Ecto is fundamentally centered around the concept of "changes" and "changesets", which makes a lot of sense in the context of forms and persistence — but feels mismatched for pure data validation where no underlying DB schema is involved. Of course, embedded schemas don't require a DB, however there's still the idea of changesets and data maybe be type-converted — e.g. the string `"123"` is implicitly cast as the integer `123`.

For use cases that require strict JSON type checking, I'm thinking that a library like [Xema](https://hexdocs.pm/xema) is better suited. Xema is inspired by [JSON Schema](http://json-schema.org/) and provides data casting and validation without the semantics of changesets.

Thanks for your earlier suggestions about custom and parameterized types — definitely useful to keep in mind for cases where they make sense.

Just wanted to share this perspective, and I do think think it's worth clarifying in the `Ecto.Changeset` documentation to help avoid surprises and potential frustration.

PR #4517, opened 2024-10-02
Include a preload joins vs. no joins discussion in the preload docs

PragTob:
I'm not sure if Ecto wants to provide this level of guidance in the docs, so I understand if this PR won't be merged.

I've just seen many well-meaning people do this wrong (imo) by defaulting to joining all associations just to preload them, resulting in a lot more code and potentially worse performance.

I also vaguely remember a tweet from José calling this out, but couldn't find it again.

Here I tried to summarize what I think I know about the topic, hoping that it's accurate and helpful.

As usual, thank you all very much for your work! :green_heart:

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2667,6 +2667,34 @@ defmodule Ecto.Query do
                  where: l.inserted_at > c.updated_at,
                  preload: [:author, comments: {c, likes: l}]
 
+  ## Choosing Between Preloading with Joins vs. Separate Queries
+
+  Deciding between preloading associations via joins, a single large
+  query, (`preload: [comments: c]`) or separate smaller queries
+  (`preload: [:comments]`) depends on the specific use case.
+  Here are some factors to guide your decision:
+
+  * **Joins reduce database round trips:** By fetching data in a single
+  query, joins can minimize database round trips, potentially reducing
+  overall latency.
+  * **Potential for data duplication:** Joins may lead to duplicated
+  data in the result set, which requires more processing by Ecto
+  and consumes more bandwidth when transmitting the results.
+  * **Increased database load:** Joins can be taxing on the database,
+  increasing the demand on working memory. In some cases, the database
+  may need to write intermediate results to disk, which can slow down
+  query performance.
+  * **Parallelism with separate queries:** When using separate queries
+  outside of a transaction, Ecto can parallelize the preload queries,
+  which can speed up the overall operation.
+
+  In general, a good default is to only use joins in preloads if you're
+  already joining the associations in the main query. For example,
+  in the last query in the section above, comments and likes are already
+  joined, so they are included in the preload.
+  However, the author is not joined in the main query, so it is preloaded
+  via a separate query.
+
   ## Preload queries
 
   Preload also allows queries to be given, allowing you to filter or
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2680,10 +2680,6 @@ defmodule Ecto.Query do
   * **Potential for data duplication:** Joins may lead to duplicated
   data in the result set, which requires more processing by Ecto
   and consumes more bandwidth when transmitting the results.
-  * **Increased database load:** Joins can be taxing on the database,
-  increasing the demand on working memory. In some cases, the database
-  may need to write intermediate results to disk, which can slow down
-  query performance.
   * **Parallelism with separate queries:** When using separate queries
   outside of a transaction, Ecto can parallelize the preload queries,
   which can speed up the overall operation.
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -2667,7 +2667,7 @@ defmodule Ecto.Query do
                  where: l.inserted_at > c.updated_at,
                  preload: [:author, comments: {c, likes: l}]
 
-  ## Choosing Between Preloading with Joins vs. Separate Queries
+  ## Choosing between preloading with joins vs. separate queries
 
   Deciding between preloading associations via joins, a single large
   query, (`preload: [comments: c]`) or separate smaller queries


lib/ecto/query.ex
```diff
@@ -2667,6 +2667,34 @@ defmodule Ecto.Query do
                  where: l.inserted_at > c.updated_at,
                  preload: [:author, comments: {c, likes: l}]
 
+  ## Choosing Between Preloading with Joins vs. Separate Queries
```
PragTob, line 2670: I wasn't sure whether to make it a separate section with a heading or not, but I ended up doing it because it got too big to just be part of the normal text.

I also put it "first" as it seemed most relevant to the preceeding setion flow-wise. Happy to re-evaluate and change!

lib/ecto/query.ex
```diff
@@ -2667,6 +2667,34 @@ defmodule Ecto.Query do
...
+  * **Joins reduce database round trips:** By fetching data in a single
+  query, joins can minimize database round trips, potentially reducing
+  overall latency.
+  * **Potential for data duplication:** Joins may lead to duplicated
+  data in the result set, which requires more processing by Ecto
+  and consumes more bandwidth when transmitting the results.
+  * **Increased database load:** Joins can be taxing on the database,
+  increasing the demand on working memory. In some cases, the database
+  may need to write intermediate results to disk, which can slow down
+  query performance.
```
greg-rychlewski, line 2686: This all looks good to me but this part I have some hesitation for a couple reasons:

1. I think this is going a little bit out of Ecto's domain. 
2. In my professional life I've often seen people too scared to use joins when they are appropriate due to not understanding the internals of the database very well. So I'm hesitant to give people information like this who likely will not take the step further to understand when this becomes a real issue vs when the issue is all in their head. Also the stated issues can happen even with non-join queries.

PragTob: Hey, thanks for the comments! I definitely don't want people to be afraid of joins in general. What I want to discourage is "joins just for the sake of preloading" and even that not completely.

To me this is one of the bigger reasons why I think that overusing joins for preloads is a problem hence I wanted to include it.

Do you have a solution in mind? Get rid off the point? Soften it up? Link to some other source, as you're right and it's complex and we can't aim to capture all the nuance involved here.

greg-rychlewski: I would personally leave out anything about database internals all together. I think it's the kind of topic that you need to do a lot of digging and self discovery and not something that you can get from a single source. 

But I am not so opposed that I would insist on leaving it out if both you and Jose want it in. I just personally find it more misleading than helpful.

warmwaffles: In SQLite it is not recommended to preload with joins and it is actively encouraged to do N+1 queries as it is more memory efficient. The underlying adapter implementation _could_ clarify these.

PragTob: @greg-rychlewski to me it's "I'm just a contributor, you know the project, audience, needs & style better than me" so I'm always more inclined to do what the maintainers want :) And as @warmwaffles points out (thanks, didn't know that!) it can vary wildly between database implementations. That said, taking PostgreSQL as the standard for ecto in the elixir community doesn't sound too bad.

Should I remove the point or should we wait for some others to chime in? I don't think there is a fully DB agnostic way to make that point, although I think it's fair to assume that for most DBs a join is an amount of more work vs. a single query.

greg-rychlewski: I would wait to see what Jose says. He is definitely more in tune with what would be best for a wider audience.

warmwaffles: Yea @PragTob it's counter intuitive for SQLite, but that's mainly because it's an embedded database so you don't have to pay for that TCP round trip retrieving data. You are limited on disk speed access. SQLite will also consume way more memory when you try to preload with joins. Just fun things to think about.

josevalim: Agreed that it is hard for us to estimate or argue the database load side. :)

PragTob: Alright, thank you everyone - I removed the section about database load!

PragTob:
Updated, removed section about database load. Thanks for the feedback everyone!

josevalim (approved):

lib/ecto/query.ex
```diff
@@ -2667,6 +2667,30 @@ defmodule Ecto.Query do
...
+  * **Parallelism with separate queries:** When using separate queries
+  outside of a transaction, Ecto can parallelize the preload queries,
+  which can speed up the overall operation.
+
+  In general, a good default is to only use joins in preloads if you're
+  already joining the associations in the main query. For example,
+  in the last query in the section above, comments and likes are already
+  joined, so they are included in the preload.
+  However, the author is not joined in the main query, so it is preloaded
+  via a separate query.
```
greg-rychlewski, line 2692: Sorry one more comment =(. Please correct me if I'm misunderstanding, but this kind of reads backwards to me. 

i.e. we are already not joining so just use a separate query preload. But it doesn't necessarily answer whether it would have been better to write the main query as a join with authors to begin with.

If it is a 1:1 relationship between posts and authors I feel like it would be more performant to do the join. In this case there is no duplication of the main rows. So you avoid the separate query and in a lot of cases will be doing the combining of the data on the more performant database server.

greg-rychlewski: This is again I think quite a nuanced topic. In general I think the advice to use joins for 1:1 and separate query for 1:many is a good starting point. But different cases may yield different results.

For instance if you are joining an enormous amount of of tables it may actually be faster in some cases to take advantage of the parallel preloader. But then again some databases, like postgres, are able to spawn workers to do the query in parallel. But then you have to be careful about memory issues if you use too much parallelism on the database side.

I would personally give the advice to start with join = 1:1, separate query = 1:many and then benchmark. I'm not sure we can do any better than that. 

But again it would be nice to hear other opinions like yours and @josevalim 's

PragTob: @greg-rychlewski you are right and the 1:1 distinction vs. 1:many distinction is a good one.

My main guidance (so far) is: Don't add joins just to preload. If you do that, you should be _sure_ it makes things faster/matters (as it's also more code). 

So, that's why I included it like this in the example. Everything else may also depend too much on the underlying database engine.

Overall, this all may be more a topic for a more detailed book (and it has a 1-2 page discussion in programming ecto) but I still think it's worth giving people some guidance. My main worry is that I've seen queries with 6+ joins added on just for the purpose of preloading and I truly believe that ain't the way :)

greg-rychlewski: I might be misunderstanding. When you say adding joins just for preloading what do you mean? If you do a join and want that table's information you won't get it returned without the preload.

greg-rychlewski: Are you saying you don't advise doing joins for information retrieval purposes? That it needs to have another purpose like filtering?

PragTob: Yes, that's my overall advice right now for ecto. If I just want to retrieve records and populate their associations a normal `preload` is enough/better imo (due to parallelism and the other problems we discussed).

That said, I'm not an omega DB expert. The docs also adhere by this: every time a join is used for a preload it also fulfills some other function in the query. 

To make it an example:

```
# good
Repo.all from p in Post,
           join: c in assoc(p, :comments),
           where: c.published_at > p.updated_at,
           preload: [comments: c]

Repo.all from p in Post,
           preload: :comments

# bad
Repo.all from p in Post,
           join: c in assoc(p, :comments),
           preload: [comments: c]
```

It's of course more nuanced than that (and in general I'd rather use `Repo.preload` then but different discussion) but that's the overall guideline I'd give people. 

But I also need to rewrite it for clarity if that wasn't clear :)

josevalim: That's my general guideline too: prefer preloads, use joins only if you need them. And Ecto's syntax literally push towards that: loading associations with joins is doing the preload syntax plus a specified join.

PragTob: @greg-rychlewski just general guidelines as the "default place to start" imo. Everything else is where I'd want to see benchmarks or some more arguments. The "Bad" here wasn't meant as "always/usually bad" but as "bad in terms of what I'd recommend as a default" - they can still be very good/useful depending on the circumstances.

@josevalim thanks for chiming in!

greg-rychlewski: Understood. I would personally start with 1:1 vs 1:many but just wanted to see if someone like Jose or another maintainer who has been here longer thinks this is ok starting advice. So even though it's not clicking with me personally I at least feel comfortable approving.

lib/ecto/query.ex
```diff
@@ -2667,6 +2667,30 @@ defmodule Ecto.Query do
                  where: l.inserted_at > c.updated_at,
                  preload: [:author, comments: {c, likes: l}]
 
+  ## Choosing Between Preloading with Joins vs. Separate Queries
```
josevalim, line 2670:
```suggestion
  ## Choosing between preloading with joins vs. separate queries
```

josevalim:

PR #4521, opened 2024-10-06
Use persistent_term for faster adapter config lookups

josevalim:


lib/ecto/repo/registry.ex
```diff
@@ -18,9 +18,8 @@ defmodule Ecto.Repo.Registry do
   end
 
   def lookup(repo) when is_atom(repo) do
-    GenServer.whereis(repo)
-    |> Kernel.||(raise "could not lookup Ecto repo #{inspect repo} because it was not started or it does not exist")
-    |> lookup()
+    :persistent_term.get(repo) ||
```
wojtekmach, line 21: Double checking: I think get/1 either returns :undefined or raises so wondering if we should use `get(repo, nil)` instead. If we are setting it prior, nevermind!

josevalim: Yes, that's what we wanted, thank you!!!

josevalim:
```suggestion
    :persistent_term.get(repo, nil) ||
```

josevalim:

Issue #4522, opened 2024-10-08
SQL Syntax error if empty list is passed to Ecto.Query.API.values/2


fuelen:
### Elixir version

Elixir 1.17.3 (compiled with Erlang/OTP 27)

### Database and Version

Postgres 14

### Ecto Versions

ecto 3.12.4, ecto_sql 3.12.1

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.19.1

### Current behavior

```elixir
(from r in "users", join: v in values([], %{id: :binary_id}), as: :values, on: r.id == v.id) |> Repo.exists?()
```
generates this SQL:
```sql
SELECT TRUE FROM "users" AS u0 INNER JOIN (VALUES (),()) AS v1 () ON u0."id" = v1."id" LIMIT 1
```
and fails with
```
** (Postgrex.Error) ERROR 42601 (syntax_error) syntax error at or near ")"

    query: SELECT TRUE FROM "users" AS u0 INNER JOIN (VALUES (),()) AS v1 () ON u0."id" = v1."id" LIMIT 1
    (ecto_sql 3.12.1) lib/ecto/adapters/sql.ex:1096: Ecto.Adapters.SQL.raise_sql_call_error/1
    (ecto_sql 3.12.1) lib/ecto/adapters/sql.ex:994: Ecto.Adapters.SQL.execute/6
    (ecto 3.12.4) lib/ecto/repo/queryable.ex:232: Ecto.Repo.Queryable.execute/4
    (ecto 3.12.4) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
    (ecto 3.12.4) lib/ecto/repo/queryable.ex:136: Ecto.Repo.Queryable.exists?/3
```

### Expected behavior

Generate a valid SQL or fail early (not on postgrex level) with friendly error message if the list is empty

PR #4523, opened 2024-10-08
Raise when empty list is given to `values/2`

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4522

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -485,6 +485,10 @@ defmodule Ecto.Query do
     @moduledoc false
     defstruct [:types, :num_rows, :params]
 
+    def new([], _types) do
+      raise ArgumentError, "must provide a non-empty list to values/2"
+    end
+
     def new(values_list, types) do
       fields = fields(values_list)
       types = types!(fields, types)
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -515,7 +515,8 @@ defmodule Ecto.Query.API do
   and `Ecto.Query.join/5`.
 
   The first argument is a list of maps representing the values of the constant table.
-  Each entry in the list must have exactly the same fields or an error is raised.
+  An error is raised if the list is empty or if every map does not have exactly the
+  same fields.
 
   The second argument is a map of types corresponding to the fields in the first argument.
   Each field must be given a type or an error is raised. Any type that can be specified in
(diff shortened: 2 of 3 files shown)


josevalim (approved):

fuelen:
I think it would be great to mention this in docs as well. It is not obvious that empty list must be handled explicitly. I caught this error in production.

greg-rychlewski:
@fuelen Thank you that's a good callout. I will update the docs.

Out of curiosity do you know of any other standard way that this is handled?

fuelen:
> Out of curiosity do you know of any other standard way that this is handled?

If we talk about elegant ways, then I don't know any

josevalim:
I have asked chatgpt and it suggested generating `(SELECT NULL WHERE false)`, which I think is better than failing. WDYT?

fuelen:
It will work, but it is no longer a VALUES statement :slightly_smiling_face: Don't know whether this is a problem.
Columns and types should be defined as well, so in a more complex scenario it will look like this
```sql
SELECT * FROM (SELECT NULL::uuid, NULL::text, NULL::int WHERE false) AS v1 ("id", "name", "age")
```

josevalim:
I don't think it is a problem in this case, but maybe @greg-rychlewski disagrees. :)

greg-rychlewski:
I think it could be a bit jarring to see such a different query in the logs if you're investigating unexpected input. But I also agree it is not good that it raises.

Since I can't think of anything better then I would go with it because I agree raising is the worst option.

warmwaffles:
@josevalim is it really wise to send that along the connection and wait for the db to respond though?

greg-rychlewski:
It could be part of a left join like this

```sql
select * from test left join (select null where false) as n  on true;
```

greg-rychlewski:
I am still a bit unsure but after thinking about it more, it might not be so bad to let the query go through and then raise at the driver level. My reasoning:

1. It keeps the query the same as the user wrote it, so that it is not surprising to see something different in your logs.
2. It is closer to the behaviour you will get in other systems like `psql`. Where you will get an error if you try to use an empty values list.
3. It leaves the option open for databases that want to support empty values lists. I don't know of any but I know in general some databases introduce nonstandard behaviour for convenience.

Probably point 2 is the most contentious because sometimes Ecto helps and sometimes it doesn't. In this case I could see a case for saying "ok you asked for an empty list, you got it". Similar to if you ask not a non-existent table.

If we go this route then I still to clean up the SQL because we have some issues like missing types.

josevalim:
Any route is fine by me. :+1:

Issue #4524, opened 2024-10-10
`field/3`'s `:writable` seems to ignore the `:insert` option


lucavenir:
### Elixir version

1.16.2

### Database and Version

PostgreSQL 16.1

### Ecto Versions

3.12.4

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.19.1

### Current behavior

I wanted to apply this `:writable` option mentioned in [field/3' documentation](https://hexdocs.pm/ecto/Ecto.Schema.html#field/3). Unluckily `writable: insert` seems to be ignored.

Given this schema:
```ex
defmodule MyApp.Accounts.Account do
  use Ecto.Schema
  import Ecto.Changeset

  schema "accounts" do
    field :firebase_uid, :string, writable: :insert
    field :deactivated_at, :utc_datetime

    timestamps(type: :utc_datetime)
  end

  @doc false
  def changeset(account, attrs) do
    account
    |> cast(attrs, [:firebase_uid, :deactivated_at])
    |> validate_required([:firebase_uid])
  end
end
```
This tests fails:
```ex
test "update_account/2 doesn't allow firebase_uid to change" do
  account = account_fixture()
  invalid_attrs = %{ firebase_uid: "another different uid" }

  # this wont work
  # assert {:error, %Ecto.Changeset{}} = Accounts.update_account(account, invalid_attrs)
  # instead, these pass (???)
  {:ok, account} = Accounts.update_account(account, invalid_attrs)
  assert account.firebase_uid == "another different uid"  # !!
end
```

### Expected behavior

I'd expect the previous test to have the commented parts to pass, and the uncommented parts to fail, i.e. I'd expect this test to pass:

```ex
test "update_account/2 doesn't allow firebase_uid to change" do
  account = account_fixture()
  invalid_attrs = %{ firebase_uid: "another different uid" }

  assert {:error, %Ecto.Changeset{}} = Accounts.update_account(account, invalid_attrs)
  # or
  # {:ok, account} = Accounts.update_account(account, invalid_attrs)
  # assert account.firebase_uid != "another different uid" 
end
```

(meaning: I'm not sure if the `:insert` option ignores the write or returns an error, but surely it shouldn't allow an edit!)

greg-rychlewski:
Hi @lucavenir ,

[...]

PR #4525, opened 2024-10-11
fix writable changes

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4524

The complexity comes from the fact that we want changes in virtuals to show up in the returned struct but not changes in read-only fields. So we need to filter twice: once to remove read only fields but not virtuals. And then right before the db operation to filter out virtuals again.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -382,7 +382,6 @@ defmodule Ecto.Repo.Schema do
     changeset = Relation.surface_changes(changeset, struct, keep_fields ++ assocs)
     changeset = update_in(changeset.changes, &Map.drop(&1, drop_fields))
 
-
     wrap_in_transaction(adapter, adapter_meta, opts, changeset, assocs, embeds, prepare, fn ->
       assoc_opts = assoc_opts(assocs, opts)
       user_changeset = run_prepare(changeset, prepare)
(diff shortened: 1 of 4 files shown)


lib/ecto/repo/schema.ex
```diff
@@ -379,7 +381,9 @@ defmodule Ecto.Repo.Schema do
     # On insert, we always merge the whole struct into the
     # changeset as changes, except the primary key if it is nil.
     changeset = put_repo_and_action(changeset, :insert, repo, tuplet)
-    changeset = Relation.surface_changes(changeset, struct, insertable_fields ++ assocs)
+    changeset = Relation.surface_changes(changeset, struct, surfaceables)
+    changeset = update_in(changeset.changes, &Map.take(&1, virtuals ++ surfaceables))
```
josevalim, line 385: I wonder if we should change `insertable_fields` to return a tuple with the insertable fields and the fields that need to be dropped, so we don't need to compute it here? And then this becomes a Map.drop? We only need to drop fields with writable that do not include `:insert`.

greg-rychlewski: good call :) updated

lib/ecto/repo/schema.ex
```diff
@@ -379,7 +379,9 @@ defmodule Ecto.Repo.Schema do
     # On insert, we always merge the whole struct into the
     # changeset as changes, except the primary key if it is nil.
     changeset = put_repo_and_action(changeset, :insert, repo, tuplet)
-    changeset = Relation.surface_changes(changeset, struct, insertable_fields ++ assocs)
+    changeset = Relation.surface_changes(changeset, struct, keep_fields ++ assocs)
+    changeset = update_in(changeset.changes, &Map.drop(&1, drop_fields))
+
```
josevalim, line 384:
```suggestion
```

josevalim (approved):

PR #4528, opened 2024-10-18
Pass allow_stale opt to assocs

tomconroy:
When implementing the soft delete detailed [here](https://dashbit.co/blog/soft-deletes-with-ecto) I ran into errors when soft deleting assocs, it seems we just need to pass the allow_stale option on to the assocs to allow this


josevalim (commented):
Can we please update the docs as well? Thank you!

josevalim (commented):
Just mention it cascades to associations. :)

tomconroy:
Done! I also updated the language slightly and removed the "Only applies to updates." note on insert_or_update as it will now affect inserts with associations. I think this change doesn't affect `delete`, I'm not sure though.

josevalim:

PR #4534, opened 2024-10-29
Improve exception message for Jason.Encoder errors

ivanov:
Thanks for all that you do, including continued maintenance of `ecto`, and now taking the time to consider this small pull request.

I decided it was too much to replicate the full contextual block that was already there just to add one alternative `@derive...` line, and thought maybe a commented out approach would be both simple and useful.

### DIFF
--- a/lib/ecto/json.ex
+++ b/lib/ecto/json.ex
@@ -10,15 +10,16 @@ if Code.ensure_loaded?(Jason.Encoder) do
           Repo.preload(#{inspect(owner)}, #{inspect(field)})
 
       Or choose to not encode the association when converting the struct \
-      to JSON by explicitly listing the JSON fields in your schema, or by \
-      excluding it (commented out):
+      to JSON by explicitly listing the JSON fields in your schema:
 
           defmodule #{inspect(owner)} do
             # ...
 
             @derive {Jason.Encoder, only: [:name, :title, ...]}
-            # @derive {Jason.Encoder, except: [#{inspect(field)}]}
             schema ... do
+
+      You can also use the :except option instead of :only if you would \
+      prefer to skip some fields.
       """
     end
   end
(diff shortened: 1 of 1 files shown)


lib/ecto/json.ex
```diff
@@ -10,12 +10,14 @@ if Code.ensure_loaded?(Jason.Encoder) do
...
-      to JSON by explicitly listing the JSON fields in your schema:
+      to JSON by explicitly listing the JSON fields in your schema, or by \
+      excluding it (commented out):
 
           defmodule #{inspect(owner)} do
             # ...
 
             @derive {Jason.Encoder, only: [:name, :title, ...]}
+            # @derive {Jason.Encoder, except: [#{inspect(field)}]}
             schema ... do
```
warmwaffles, line 21: One alternative here is to show two separate examples without having a directive to uncomment a line.

```elixir
    @doc """


      Or choose to not encode the association when converting the struct \
      to JSON by explicitly listing the JSON fields in your schema:

          defmodule #{inspect(owner)} do
            # ...
            @derive {Jason.Encoder, only: [:name, :title, ...]}
            schema ... do

      Or choose to not encode the association by excluding it in your schema:

          defmodule #{inspect(owner)} do
            # ...
            @derive {Jason.Encoder, except: [#{inspect(field)}]}
            schema ... do

 
    """
```

ivanov: I'm happy to do convert to that style, if that's preferred (also happy if maintainers want to adjust this in that direction). The only thing to add is that this isn't part of narrative documentation, this is traceback you will get if your code causes this error, so that's why I was trying to lean towards being more terse.

josevalim:
I believe except is anti-pattern because it is easy to leak information you don't want but it is ok to mention it. :)

josevalim:

Issue #4537, opened 2024-11-02
Inspect.Error when bind name is pinned


fuelen:
### Elixir version

Elixir 1.17.3 (compiled with Erlang/OTP 27)

### Database and Version

Irrelevant

### Ecto Versions

3.12.4

### Database Adapter and Versions (postgrex, myxql, etc)

irrelevant

### Current behavior

```elixir
iex(1)> dynamic([{^:a, r}], r)
#Inspect.Error<
  got Ecto.QueryError with message:

      """
      unknown bind name `:a` in query:

      from q0 in query
      """

  while inspecting:

      %{
        line: 1,
        file: "iex",
        fun: #Function<42.39164016/1 in :erl_eval.expr/6>,
        binding: [{{:^, [line: 1], [:a]}, {:r, [line: 1], nil}}],
        __struct__: Ecto.Query.DynamicExpr
      }

  Stacktrace:

    (ecto 3.12.4) lib/ecto/query/builder.ex:1026: Ecto.Query.Builder.count_alias!/2
    (elixir 1.17.3) src/elixir.erl:386: :elixir.eval_external_handler/3
    (stdlib 6.1) erl_eval.erl:904: :erl_eval.do_apply/7
    (stdlib 6.1) erl_eval.erl:648: :erl_eval.expr/6
    (stdlib 6.1) erl_eval.erl:271: :erl_eval.exprs/6
    (stdlib 6.1) erl_eval.erl:648: :erl_eval.expr/6
    (stdlib 6.1) erl_eval.erl:271: :erl_eval.exprs/6
    (ecto 3.12.4) lib/ecto/query/builder/dynamic.ex:83: Ecto.Query.Builder.Dynamic.expand/3
    (ecto 3.12.4) lib/ecto/query/builder/dynamic.ex:53: Ecto.Query.Builder.Dynamic.fully_expand/2
    (ecto 3.12.4) lib/ecto/query/inspect.ex:22: Inspect.Ecto.Query.DynamicExpr.inspect/2
    (elixir 1.17.3) lib/inspect/algebra.ex:347: Inspect.Algebra.to_doc/2
    (elixir 1.17.3) lib/kernel.ex:2381: Kernel.inspect/2
    (iex 1.17.3) lib/iex/evaluator.ex:376: IEx.Evaluator.io_inspect/1
    (iex 1.17.3) lib/iex/evaluator.ex:335: IEx.Evaluator.eval_and_inspect/3
[...]

PR #4541, opened 2024-11-02
fix inspecting dynamc with interpolated named bindings

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4537


greg-rychlewski:
I did it two ways....the first commit keeps the pin in the inspect and the second one doesn't. I think the second is nicer but can change it back if you prefer the first.

josevalim (approved):
We shouldn't keep the pin i think because we don't keep it for regular queries.

PR #4545, opened 2024-11-09
erase from persistent_term without blocking registry

SteffenDE:
Note: this is a draft. I talked to @josevalim and we're not 100% sure how to handle this yet. Another idea was to just not erase entries from persistent_term.

This was an interesting problem. It started with me updating the dependencies in phoenix_live_dashboard and noticing test failures related to our test of the Ecto Stats page. The stats page calls `Ecto.Repo.all_running` to get a list of repos.

The problem was that the test case still got an old repo as running, even though the previous test already stopped it (using start_supervised). It turns out that the :persistent_term.erase was slowing the registry GenServer down too much, therefore it wasn't processing the DOWN messages fast enough and not deleting the old repo entries from ets before the new test was already running.

### DIFF
--- a/lib/ecto/repo/registry.ex
+++ b/lib/ecto/repo/registry.ex
@@ -45,7 +45,10 @@ defmodule Ecto.Repo.Registry do
   @impl true
   def handle_info({:DOWN, ref, _type, pid, _reason}, table) do
     [{^pid, ^ref, name, _}] = :ets.lookup(table, pid)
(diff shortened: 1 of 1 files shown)


SteffenDE:
Relates to: https://github.com/phoenixframework/phoenix_live_dashboard/pull/456

josevalim:
I ended up reverting it. Thank you for spelunking!

Issue #4547, opened 2024-11-12
wrong(?) `unknown_binding` warnings when upgrading to Ecto 3.12


Miradorn:
### Elixir version

Elixir 1.17.2 (compiled with Erlang/OTP 26)

### Database and Version

postgres 14.9

### Ecto Versions

3.12.4

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.19.2

### Current behavior

I'm trying to update from Ecto 3.11.2 to 3.12.4 and suddenly some of our queries can no longer be compiled with the warning `invalid query has specified more bindings than bindings available in `where` (look for `unknown_binding!` in the printed query below) in query`, but the query itself doesn't have this warning and it does look correct to me (and it did work before the update :D). Here is the query logged by the error (with the names of stuff anonymized):

```
from e0 in MySchema1,
           as: :my_schema_1,
           join: p1 in assoc(e0, :my_schema_2),
           as: : my_schema_2,
           where: my_schema_1.field_1 == ^53091,
           where: exists(
           subquery(
             #Ecto.Query<from e0 in MySchema3, where: e0.field_1 == parent_as(:my_schema_1).id, where: parent_as(:my_schema_2).field_2 not in ^["X"] and e0.field_2 >= ^~D[2024-11-11], or_where: e0.field_3 == parent_as(:my_schema_1).id and
           parent_as(: my_schema_2).field_3 == ^"Y" and e0.end_date >= ^~D[2024-10-11], select: 1>
           )
         ),
           where: p1.field_4 != :xyz,
           select: e0
```

I'm not sure wether this is a bug or a change in 3.12 is just surfacing an error by us when creating the query.

If you need any more information, please let me know.

❤️ Thanks!

### Expected behavior

The query should probably still compile and work as before

josevalim:
Hi @Miradorn, I think there is no `unknown_binding!` in the query you pasted? Can you please try to provide a snippet that reproduces the issue, rather than the printed query? You can probably do it by using "strings" in from, such as `from s in "schema"`, just remember to do `select: s.something`.

Miradorn:
>  I think there is no unknown_binding! in the query you pasted?

Exactly, that's my main problem in understanding what goes wrong. all bindings seem to be good 🙈 
I'll try to create a reproduction.

Miradorn:
ok, here we go:

```
defmodule EctoTest.Stuff do
  import Ecto.Query

  def da_query do
    base_query =
[...]

PR #4548, opened 2024-11-12
Plan sources before creating plan_subquery closure

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4547

Because `where` and friends might need to cast parameters to field types, we need to be able to cast the sources first. This way `query.sources` is populated.


josevalim (approved):
Thank you for cleaning up after me :heart:

Issue #4536, opened 2024-10-31
Discussion: Selecting parts of embeds with `Ecto.Query.API.struct/2`


greg-rychlewski:
### Elixir version

all

### Database and Version

all

### Ecto Versions

3.12

### Database Adapter and Versions (postgrex, myxql, etc)

all

### Current behavior

Currently you can use `struct/2` to retrieve only certain fields from the underlying table and still have the result return as the struct you want and still have preloads.

My coworker had a situation where they also wanted to restrict the attributes coming from an embedded field within the table. The best I was able to suggest was to use `select_merge`, like this:

```elixir
from t in Table, 
  select: struct([t], [:field1, :field2, ...]), 
  select_merge: %{embed_field: json_extract_path(t.embed_field, ["some", "path"]))
```

There may be a more standard way to do this that I am not thinking of. But assuming the way above is the best way to do it currently, would it make sense to allow something like this

```elixir
from t in Table, 
  select: struct([t], [:field1, :field2, ..., embed_field: [:sub_field1, :sub_field2, sub_field3: [...]])
```

### Expected behavior

TBD

josevalim:
Yes, I think allowing that extension on map/struct can be nice.

greg-rychlewski:
One thing I realized is it would probably be helpful to have a way to extract subsets of json for this behaviour. I think it can get a bit tricky to select each path individually and then try to rebuild the potentially nested object at the end.

The way this is done in Postgres/MySQL is to use functions like this:


```sql
json(b)_build_object(key1, value1, key2, value2, ...) (postgres)
json_object(key1, value1, key2, value2, ...) mysql)
```

What would you think if we introduced `Ecto.Query.API.json_object/3`?  It would take the following arguments:

1. the json(b) field
2. a path list with potentially nested lists, like `["field1", "field2", "field3" => [...]]`
3. the type (`:jsonb` or `:json`) because postgres can't use the same syntax for both

greg-rychlewski:
Or I guess the API I suggested makes more sense with the name `json_subset`. And if we want to add `json_object` we keep the alternating key/value arguments.

greg-rychlewski:
Closing based on discussion in PR.

PR #4558, opened 2024-12-22
Allow map/2 and struct/2 to take subsets of embed fields

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4536

The general idea is to select the fields individually and then combine them in Elixir after they are retrieved from the database. There are more corner cases than I was anticipating. Here are some points for consideration:

1. I am using `json_extract_path` to select the individual fields from the embed assuming this abstraction is meant to work for most databases. If it's not mean to be, then is it better to create a general expression for embed fields and have the adapter convert it?
2. Right now this only works for `embeds_one` and not `embeds_many` because of the way extractions are being done. There is a similar limitation on `json_extract_path`. There is a chance it can be extended to `embeds_many` but I am not sure there is a simple operator to extract the same key from every object in a json array. So if we want to keep this we might have to be ok with it only working for `embeds_one`
3. Right now taking associations inside of embeds is not supported. I think this can be done but if it's not a big deal would prefer to do it in a follow-up PR. I think there are enough complications with the current PR to consider.
4. There is a current inconsistency, maybe intentional, between taking associations and taking embeds that is carrying over to this PR. If doing `map(p, [assoc_field: [:assoc_sub_field])` the `:assoc_field` will also be a map. But if it's an embed field it will be a struct. This happens with or without the current PR. I kept the current behaviour but wanted to highlight it in case we want to make these consistent in a different PR. It will probably be a larger change to see why it behaves this way today.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1425,6 +1425,7 @@ defmodule Ecto.Integration.RepoTest do
       assert p3 == %{id: pid3}
     end
 
+    @tag :json_extract_path
     test "take with embed" do
       today = Date.utc_today()
 
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -1443,7 +1443,7 @@ defmodule Ecto.Integration.RepoTest do
       [o1, o2, o3] =
         Order
         |> select([o], map(o, [:metadata, item: [:price, primary_color: [:name]]]))
-        |> order_by([o], o.item["price"])
+        |> order_by([o], o.id)
         |> TestRepo.all()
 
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you @greg-rychlewski, this looks neatly done but, after looking at the implementation, I wonder if it is worth it?

I am also worried about having partial representations of the data, because we only load part of the map or embedded schema. In a way this is expected, but this will become increasingly problematic with the type system, as we need to effectively assume all fields to be nil.

In this case, it is probably best to write:

    select: {map(o, [:metadata]), o.item.price.primary_color.name}

greg-rychlewski:
> after looking at the implementation, I wonder if it is worth it?

Yeah I agree. I had a hard time letting it go but I was also thinking the same thing after finishing. Thanks for reviewing.

PR #4559, opened 2024-12-22
Document usage of ParametrizedType inside of a composite type

tomekowal:
Hi!

I was recently struggling with finding how to use a parametrized type inside of an array.
https://elixirforum.com/t/syntax-for-an-array-of-parameterized-types-in-ecto/68114
@wojtekmach suggested adding that to the documentation.

Let me know if that is OK or if I should improve anything!


josevalim:

PR #4561, opened 2024-12-25
Add expression to nil comparison error message

greg-rychlewski:
Was reading Elixir Forum and found this: https://elixirforum.com/t/ecto-argumenterror-comparison-with-nil-is-forbidden-when-querying-for-null-field-in-phoenix-elixir/68343/3

This should help make the issue easier to identify.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -357,11 +357,10 @@ defmodule Ecto.Query.Builder do
   def escape({comp_op, _, [left, right]} = expr, type, params_acc, vars, env)
       when comp_op in ~w(== != < > <= >=)a do
     assert_type!(expr, type, :boolean)
-    compare_str = Macro.to_string(expr)
 
     if is_nil(left) or is_nil(right) do
       error!(
-        "comparison with nil in `#{compare_str}` is forbidden as it is unsafe. " <>
+        "comparison with nil in `#{Macro.to_string(expr)}` is forbidden as it is unsafe. " <>
           "If you want to check if a value is nil, use is_nil/1 instead"
       )
     end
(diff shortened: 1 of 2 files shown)


lib/ecto/query/builder.ex
```diff
@@ -357,10 +357,11 @@ defmodule Ecto.Query.Builder do
   def escape({comp_op, _, [left, right]} = expr, type, params_acc, vars, env)
       when comp_op in ~w(== != < > <= >=)a do
     assert_type!(expr, type, :boolean)
+    compare_str = Macro.to_string(expr)
```
josevalim, line 360: My only concern is that we are now storing strings in the generated artifact. And those can be large. What if we store only the opposite side, to reduce the amount of generated code?

lib/ecto/query/builder.ex
```diff
@@ -357,10 +357,11 @@ defmodule Ecto.Query.Builder do
...
 
     if is_nil(left) or is_nil(right) do
       error!(
-        "comparison with nil is forbidden as it is unsafe. " <>
+        "comparison with nil in `#{compare_str}` is forbidden as it is unsafe. " <>
```
josevalim, line 364: Build it lazily:

```suggestion
        "comparison with nil in `#{Macro.to_string(expr)}` is forbidden as it is unsafe. " <>
```

lib/ecto/query/builder.ex
```diff
@@ -372,7 +373,8 @@ defmodule Ecto.Query.Builder do
     {right, params_acc} = escape(right, rtype, params_acc, vars, env)
 
     {params, acc} = params_acc
-    {{:{}, [], [comp_op, [], [left, right]]}, {params |> wrap_nil(left) |> wrap_nil(right), acc}}
+    params = params |> wrap_nil(left, compare_str) |> wrap_nil(right, compare_str)
```
josevalim, line 376:
```suggestion
    params = params |> wrap_nil(left, Macro.to_string(right)) |> wrap_nil(right, Macro.to_string(left))
```

lib/ecto/query/builder.ex
```diff
@@ -1184,13 +1186,14 @@ defmodule Ecto.Query.Builder do
   @doc """
   Called by escaper at runtime to verify that a value is not nil.
   """
-  def not_nil!(nil) do
+  def not_nil!(nil, compare_str) do
     raise ArgumentError,
-          "comparison with nil is forbidden as it is unsafe. " <>
+          "comparison with nil in `#{compare_str}` is forbidden as it is unsafe. " <>
```
josevalim, line 1191:
```suggestion
          "comparing `#{compare_str}` with `nil` is forbidden as it is unsafe. " <>
```

josevalim:
I dropped some suggestions, WDYT?

Btw, happy holidays @greg-rychlewski!

greg-rychlewski:
oh yeah that's a really good point. updated with your suggestions.

happy holidays to you as well!

josevalim (approved):

PR #4562, opened 2025-01-08
Allow numbers in literal/1

greg-rychlewski:
There was a discussion where someone was using a database that does not allow query parameters in limit expressions but it's still desirable to have dynamic limit values. The simplest thing I can think of is to expand `literal/1` to also allow numbers.

Some other considerations that I did not pursue:

- Should we allow literal outside of fragments? That would be a bigger change and would not really make sense with the original use case of `literal/1` (quoting identifiers). Also we probably don't want to encourage using this too much as it may blow up the Ecto cache.
- Should we allow more than identifiers and numbers? Then we have an issue where a binary may be interpreted as a literal string or an identifier so we need to add `literal/2` or something like that. I'm not sure it is worth it since the use cases are pretty limited.

Companion Ecto SQL PR: https://github.com/elixir-ecto/ecto_sql/pull/651

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -480,17 +480,20 @@ defmodule Ecto.Query.API do
   def fragment(fragments), do: doc!([fragments])
 
   @doc """
-  Allows a literal identifier to be injected into a fragment:
+  Allows a literal identifier or number to be injected into a fragment:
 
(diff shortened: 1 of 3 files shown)


josevalim (approved):
I am fine with expanding it slowly as needed, as this looks good to me :)

PR #4563, opened 2025-01-09
Revert "Allow numbers in literal/1 (#4562)" and add identifier/1 and constant/1 instead

greg-rychlewski:
Based on our discussions, it seems there is agreement that the old `literal/1` should be deprecated and we should make two new "functions": one for identifiers and one for constant values.

There was agreement by some people for `identifier/1` and `constant/1`. If you are not in favour of those names let me know! It won't affect the PR much.

Also, I'm not sure what is the normal deprecation procedure. I think we need to keep accepting `literal/1` to not break apps? If so should we remove it from the documentation or put some notes in the documentation about it being deprecated and replaced by these 2?

Companion Ecto SQL PR: https://github.com/elixir-ecto/ecto_sql/pull/652

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -493,6 +493,38 @@ defmodule Ecto.Query.API do
   def literal(binary), do: doc!([binary])
 
   @doc """
+  Allows a dynamic identifier to be injected into a fragment:
+
+      collation = "es_ES"
+      select("posts", [p], fragment("? COLLATE ?", p.title, identifier(^"es_ES")))
+
+  The example above will inject the value of `collation` directly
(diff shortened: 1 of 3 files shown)


josevalim:
For deprecation we can remove the literal and make it sure it emits a warning, a compile-time warning being best. Internally, we can rewrite `literal` to `identifier` in the AST. It is probably best if we do these two changes in the Builder, so the warning is compile-time. But yeah, other than that, we need to keep on supporting for quite some time.

josevalim (approved):

greg-rychlewski:
@josevalim sorry before I merge, does the last commit look alright to you?

josevalim:
ship it!!!!

PR #4569, opened 2025-01-20
Allow exclude/2 to remove windows

greg-rychlewski:
While thinking about adding windows preloads or windowed limits to Ecto.Query I noticed we can't do this right now.

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -997,7 +997,7 @@ defmodule Ecto.Query do
   end
 
   defp do_exclude(%Ecto.Query{} = query, {:windows, window_names}) when is_list(window_names) do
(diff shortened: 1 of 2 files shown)


lib/ecto/query.ex
```diff
@@ -988,6 +996,11 @@ defmodule Ecto.Query do
     %{query | joins: remaining, aliases: aliases}
   end
 
+  defp do_exclude(%Ecto.Query{} = query, {:windows, window_names}) when is_list(window_names) do
+    remaining = Enum.filter(query.windows, fn {name, _} -> name not in window_names end)
```
greg-rychlewski, line 1000: i have a hard time imagining the number of windows being used is worth creating a set out of `window_names` for this....but please let me know if this is too lazy of an implementation

josevalim (approved):

PR #4577, opened 2025-02-13
Add Repo.transaction_with/2

wojtekmach:
References:
- https://tomkonidas.com/repo-transact/
- https://elixirforum.com/t/seeking-thoughts-on-advantages-of-the-repo-transact-pattern-vs-disadvantages-i-ve-read-about-ecto-multi/61733

I believe the interface and implementation are fairly uncontroversial and straightforward. I think the bigger undertaking is potentially revamping the docs, de-emphasizing Multi in favor of this, given that in a lot of (if not the vast majority of) cases, it achieves the same outcomes with less and simpler code. There are no plans to deprecate Multi.

The name of this function is not ideal, some other options considered were `Repo.transact` and `Repo.with_transaction`. I think `Repo.transaction_with` wins because basically:

```
iex(1)> Repo.transaction<tab>
transaction/1         transaction/2         transaction_with/1    transaction_with/2
```

That is, it will show up alongside `transaction` in docs, autocomplete, etc.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -290,6 +290,16 @@ defmodule Ecto.Repo do
           )
         end
 
+        def transaction_with(fun, opts \\ []) do
+          repo = get_dynamic_repo()
+
+          Ecto.Repo.Transaction.transaction_with(
+            repo,
+            fun,
+            Ecto.Repo.Supervisor.tuplet(repo, prepare_opts(:transaction, opts))
+          )
+        end
+
         def in_transaction? do
           Ecto.Repo.Transaction.in_transaction?(get_dynamic_repo())
         end
@@ -2076,6 +2086,8 @@ defmodule Ecto.Repo do
   A successful transaction returns the value returned by the function
   wrapped in a tuple as `{:ok, value}`.
 
+  See also `c:transaction_with/2`.
+
   ### Nested transactions
 
   If `c:transaction/2` is called inside another transaction, the function
@@ -2131,6 +2143,8 @@ defmodule Ecto.Repo do
   `{:error, failed_operation, failed_value, changes_so_far}` will be returned.
 
   Explore the `Ecto.Multi` documentation to learn more and find detailed examples.
+
+  See also `c:transaction_with/2`.
 
   ## Aborted transactions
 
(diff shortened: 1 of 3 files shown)


lib/ecto/repo.ex
```diff
@@ -2200,6 +2214,81 @@ defmodule Ecto.Repo do
...
+      {:error, :oops}
+
+      iex> Repo.transaction_with(fn ->
+      ...>   Repo.insert!(%User{}) # will be rolled back
+      ...>   raise "oops"
+      ...> end)
+      ** (RuntimeError) oops
+  """
+  @doc group: "Transaction API"
+  @callback transaction_with(fun, opts :: Keyword.t()) ::
```
warmwaffles, line 2288: I am wondering if changing from `fun` to (-> {:ok, term()} | {:error, term()})` is a better hint.

wojtekmach: Good call, done in 9a9f133e.

novaugust:
I'm grateful to see this being brought into Ecto ❤️ it'll help with onboarding new engineers to have `transact` functionality in ecto's docs rather than our own impls

I'd really like to encourage using `transact` over `transact_with`:

- existing codebases, inspired by sasa's post, are already using it (so no need to find-and-replace when updating to latest ecto)
- `transact` is a verb, same as the existing `get`/`update`/`delete` repo functions (fwiw, we also have the bangified `Repo.transact!` to fit the theme here)
- `_with` doesn't really do anything in this name. it feels superfluous in other contexts as well: `Enum.map_with` for example
- for tab auto completion, i'm going to hit tab well before writing out the whole name,(eg `Repo.trans<tab>`), which would still give `transact` alongside all the other existing transaction functions

Regardless of how it ends up named, thanks for bringing it in!

sasa1977:
Thanks for bring this to Ecto :heart: 

Any thoughts about also supporting returning plain `:ok` and `:error` (which would also be the result of `transaction_with`)? In practice I found this convenient.

lib/ecto/repo.ex
```diff
@@ -2200,6 +2214,80 @@ defmodule Ecto.Repo do
...
+  If this function returns `{:ok, result}`, it means the transaction
+  was successfully committed. On the other hand, if it returns `{:error, reason}`,
+  it means the transaction was rolled back.
+
+  If an Elixir exception occurs the transaction will be rolled back
+  and the exception will bubble up from the transaction function.
+  If no exception occurs, the transaction is committed when the
+  function returns. A transaction can be explicitly rolled back
+  by calling `c:rollback/1`, this will immediately leave the function
+  and return the value given to `rollback` as `{:error, value}`.
```
novaugust, line 2232: i might just be failing at reading comprehension this morning, but this doesn't seem to explicitly detail that the transaction is rolled back if `fun` returns `{:error, reason}` (or really, anything other than `{:ok, result}`)

```suggestion
  If an Elixir exception occurs the transaction will be rolled back
  and the exception will bubble up from the transaction function.
  If no exception occurs, the transaction is committed if the function 
  returns `{:ok, result}`. Returning `{:error, result}` will rollback the transaction
  and this function will return `{:error, result}` as well. 
  A transaction can be explicitly rolled back
  by calling `c:rollback/1`, this will immediately leave the function
  and return the value given to `rollback` as `{:error, value}`.
```

v0idpwn (commented):
Love this! 

Should probably be made an optional callback (like transaction is).

zlatkovlasic:
Very happy to see this, been using Sasa's version for some time. Strongly against proposed function name, tho. Since one of the benefits is simpler and more readable code (with less LOC) - `Repo.transaction_with(with ...` feels wrong

Damirados:
> I'd really like to encourage using `transact` over `transact_with`:
>  ...
> Regardless of how it ends up named, thanks for bringing it in!

100% to this comment. I will definitely rename it to `transact` on a project level if it comes in with a different name.

tomkonidas (approved):

sasa1977:
FWIW as the original author of the name `transact`, I'm fine with `transaction_with`. Given that we're keeping `transaction`, I actually find that `transaction_with` better communicates the purpose of the new function.

I think that ideally, the current `transaction` would become `transaction!`, and the new function would become `transaction`, but I guess this isn't possible given the amount of code out there. Alternatively, perhaps `transact` and `transact!` with the deprecated `transaction`, but not sure if it's worth it.

sodapopcan:
> but I guess this isn't possible given the amount of code out there.

Unless Ecto v4?  Could ship with a re-writer task, though perhaps I'm oversimplifying this.

josevalim:
I have discussed this with @wojtekmach:

1. Let's call this `transact`, it will accept either a function or a multi, the function can return only `{:ok, ...}` or `{:error, ...}` for now
2. Let's see if we can implement `transaction` on top of `transact`
3. `transaction` is soft-deprecated (to be fully deprecated later on)

zlatkovlasic:
Any news on this :)

wojtekmach:
I hope tknfonosh it in the next week or two, sorry for the delay.

wojtekmach:
Closing in favor of #4618, thanks everyone for feedback and apologies for delay!

Issue #4575, opened 2025-02-07
is_nil not allowed in order_by since v3.12.0


justincy:
### Elixir version

1.15.7

### Database and Version

PostgreSQL 15.6

### Ecto Versions

This error began in 3.12.0. It was allowed in 3.11.2 and earlier versions.

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex 0.20

### Current behavior

Using `is_nil` in `order_by` generates an `Ecto.Query.CompileError`.

```
== Compilation error in file test/ecto/query_test.exs ==
** (Ecto.Query.CompileError) comparison with nil in `b == nil` is forbidden as it is unsafe. If you want to check if a value is nil, use is_nil/1 instead
    (ecto 3.13.0-dev) expanding macro: Ecto.Query.order_by/3
    test/ecto/query_test.exs:82: Ecto.QueryTest."test query building allows is_nil in order_by"/1
    (ecto 3.13.0-dev) expanding macro: Ecto.Query.from/2
    test/ecto/query_test.exs:82: Ecto.QueryTest."test query building allows is_nil in order_by"/1
```

That error was generated by adding this test to the ecto repo:

```
test "allows is_nil in order_by" do
  _ = from(p in "posts", left_join: b in "blogs", order_by: is_nil(b))
end
```

### Expected behavior

No error should be generated since it's using `is_nil` as directed.

justincy:
I wonder if the changes for [macro expansion](https://github.com/elixir-ecto/ecto/pull/4449) in `order_by` cause it to forget that `is_nil` is being used.

greg-rychlewski:
Thanks @justincy I believe your analysis is correct.

@josevalim It looks like this is the culprit: https://github.com/greg-rychlewski/ecto/blob/d4e7dbf87f946ea8911b68d9a68dff65099b3635/lib/ecto/query/builder/order_by.ex#L54

Is there any way to put the `env` that we don't want `is_nil` to be expanded?

josevalim:
I think what we can do is to pass `{env, function}` to builder, so the builder does the expansion and calls back to us, like we do in select? https://github.com/greg-rychlewski/ecto/blob/d4e7dbf87f946ea8911b68d9a68dff65099b3635/lib/ecto/query/builder/select.ex#L130

PR #4578, opened 2025-02-14
Fix macros inside of order_by

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4575

### DIFF
--- a/lib/ecto/query/builder/order_by.ex
+++ b/lib/ecto/query/builder/order_by.ex
@@ -51,41 +51,57 @@ defmodule Ecto.Query.Builder.OrderBy do
           {Macro.t(), {list, term}}
   def escape(kind, expr, params_acc, vars, env) do
     expr
-    |> Macro.expand_once(get_env(env))
     |> List.wrap()
-    |> Enum.map_reduce(params_acc, &do_escape(&1, &2, kind, vars, env))
+    |> Enum.flat_map_reduce(params_acc, &do_escape(&1, &2, kind, vars, env))
   end
 
   defp do_escape({dir, {:^, _, [expr]}}, params_acc, kind, _vars, _env) do
-    {{quoted_dir!(kind, dir),
-      quote(do: Ecto.Query.Builder.OrderBy.field!(unquote(kind), unquote(expr)))}, params_acc}
+    {[{quoted_dir!(kind, dir),
+      quote(do: Ecto.Query.Builder.OrderBy.field!(unquote(kind), unquote(expr)))}], params_acc}
   end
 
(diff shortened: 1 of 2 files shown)


josevalim (approved):
Beautiful!

Gigitsu:
Hello,

Unfortunately, it seems that this PR may have introduced a regression.

Given the following setup:

```elixir
defmodule MyQueryUtils do
  defmacro foo(table) do
    quote(do: unquote(table).priority > 0)
  end
end

defmodule Post do
  use Ecto.Schema

  @primary_key false
  schema "posts" do
    field :title, :string
    field :priority, :integer
  end
end
```

This query no longer compiles:

```elixir
import MyQueryUtils
import Ecto.Query

select(Post, [o], %{
  struct(o, [:title]) |
    row_number: over(row_number(), order_by: [desc: foo(o)])
})
```

With the following error:

```
** (KeyError) key :module not found in: {#Macro.Env<
   aliases: [],
   ...
 >, #Function<4.25497058/5 in Ecto.Query.Builder.Select.escape_expansion>}

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (elixir 1.18.3) lib/macro/env.ex:539: Macro.Env.expand_import/5
    (elixir 1.18.3) lib/macro.ex:1877: Macro.do_expand_once/2
    (elixir 1.18.3) lib/macro.ex:1824: Macro.expand_once/2
    (ecto 3.13.1) lib/ecto/query/builder.ex:1101: Ecto.Query.Builder.try_expansion/5
    (ecto 3.13.1) lib/ecto/query/builder/order_by.ex:79: Ecto.Query.Builder.OrderBy.do_escape/5
    (elixir 1.18.3) lib/enum.ex:1316: anonymous fn/3 in Enum.flat_map_reduce/3
    (elixir 1.18.3) lib/enum.ex:4968: Enumerable.List.reduce/3
    (elixir 1.18.3) lib/enum.ex:1315: Enum.flat_map_reduce/3
```

A possible fix could be to handle this specific case [here](https://github.com/Gigitsu/ecto/blob/master/lib/ecto/query/builder/order_by.ex#L76) by adding:

```elixir
defp do_escape(expr, params_acc, kind, vars, {env, _}) do
  do_escape(expr, params_acc, kind, vars, env)
end
```

Let me know what you think — I’d be happy to help test and open a pull request for this.

Thanks again!

greg-rychlewski:
@Gigitsu Yeah basically that. In other places like the main builder function we define a `get_env` helper like this that might be a bit easier to read

```elixir
defp get_env({env, _}), do: env
  defp get_env(env), do: env
```

PR #4582, opened 2025-02-25
Documentation tweaks for changesets and embedded schemas

jdmarshall:
A few tweaks to the docs, covering some grey areas where I could have used a little extra help.

Critiques welcome. New Guy Edits the Docs is valuable both to the next person to read them, and as a teachable moment if they get something hilariously wrong.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1,6 +1,6 @@
 defmodule Ecto.Changeset do
   @moduledoc ~S"""
-  Changesets allow filtering, type conversion, validation and
+  Changesets allow filtering, casting, validation and
   definition of constraints when manipulating structs, usually in
   preparation for inserting and updating entries into a database.
 
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -5,13 +5,13 @@ defmodule Ecto.Changeset do
   preparation for inserting and updating entries into a database.
 
   There is an example of
-  [working with changesets](Ecto.html#module-changesets) in the
+  [working with changesets](`Ecto#module-changesets`) in the
   introductory documentation in the `Ecto` module. In a nutshell, there
   are two main functions for creating a changeset. The `cast/4` function
-  is used to receive external parameters from a form, API or command line
-  and convert them to the types defined in your `Ecto.Schema`. `change/2`
-  is used to modify data directly from your application, assuming the
-  data given is valid and matches the existing types.
+  is used to receive external parameters from a form, API or command
+  line, and convert them to the types defined in your `Ecto.Schema`.
+  `change/2` is used to modify data directly from your application,
+  assuming the data given is valid and matches the existing types.
 
   The remaining functions in this module, such as validations,
   constraints, association handling, are about manipulating
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -2,14 +2,16 @@ defmodule Ecto.Changeset do
   @moduledoc ~S"""
   Changesets allow filtering, type conversion, validation and
   definition of constraints when manipulating structs, usually in
-  preparation for insertion into the database.
+  preparation for inserting and updating entries into a database.
 
-  There is an example of working with changesets in the introductory
-  documentation in the `Ecto` module. The functions `cast/4` and
-  `change/2` are the usual entry points for creating changesets.
-  The first one is used to cast and validate external parameters,
-  such as parameters sent through a form, API, command line, etc.
-  The second one is used to change data directly from your application.
+  There is an example of
+  [working with changesets](Ecto.html#module-changesets) in the
+  introductory documentation in the `Ecto` module. In a nutshell, there
+  are two main functions for creating a changeset. The `cast/4` function
+  is used to receive external parameters from a form, API or command line
+  and convert them to the types defined in your `Ecto.Schema`. `change/2`
+  is used to modify data directly from your application, assuming the
+  data given is valid and matches the existing types.
 
   The remaining functions in this module, such as validations,
   constraints, association handling, are about manipulating
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -147,8 +147,8 @@ defmodule Ecto.Changeset do
 
   ## Associations, embeds, and on replace
 
-  Using changesets you can work with associations as well as with embedded
-  structs. There are two primary APIs:
+  Using changesets you can work with associations as well as with
+  [embedded](embedded-schemas.html) structs. There are two primary APIs:
 
     * `cast_assoc/3` and `cast_embed/3` - those functions are used when
       working with external data. In particular, they allow you to change
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1,7 +1,8 @@
 defmodule Ecto.Changeset do
   @moduledoc ~S"""
-  Changesets allow filtering, casting, validation and
-  definition of constraints when manipulating structs.
+  Changesets allow filtering, type conversion, validation and
+  definition of constraints when manipulating structs, usually in
+  preparation for insertion into the database.
 
   There is an example of working with changesets in the introductory
   documentation in the `Ecto` module. The functions `cast/4` and
@@ -18,13 +19,13 @@ defmodule Ecto.Changeset do
 
   Changesets allow working with two kinds of data:
 
-    * internal to the application - for example programmatically generated,
-      or coming from other subsystems. This use case is primarily covered
-      by the `change/2` and `put_change/3` functions.
-
-    * external to the application - for example data provided by the user in
+    * external to the application - for example user input from
       a form that needs to be type-converted and properly validated. This
       use case is primarily covered by the `cast/4` function.
+
+   * internal to the application - for example programmatically generated,
+      or coming from other subsystems. This use case is primarily covered
+      by the `change/2` and `put_change/3` functions.
 
   When working with external data, the data is typically provided
   as maps with string keys (also known as parameters). On the other hand,


lib/ecto/changeset.ex
```diff
@@ -18,14 +19,14 @@ defmodule Ecto.Changeset do
 
   Changesets allow working with two kinds of data:
 
-    * internal to the application - for example programmatically generated,
-      or coming from other subsystems. This use case is primarily covered
-      by the `change/2` and `put_change/3` functions.
-
-    * external to the application - for example data provided by the user in
+    * external to the application - for example user input from
```
jdmarshall, line 22: The external case seems to be the one that a new user will encounter chronologically first. And also everywhere except the bullets, external is mentioned prior to internal, so I reordered them to be consistent.

lib/ecto/changeset.ex
```diff
@@ -1,7 +1,8 @@
 defmodule Ecto.Changeset do
   @moduledoc ~S"""
-  Changesets allow filtering, casting, validation and
```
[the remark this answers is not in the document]
josevalim: "cast" is the term consistently used by all four major SQL databases: [PostgreSQL](https://www.postgresql.org/docs/current/sql-createcast.html), [MySQL](https://dev.mysql.com/doc/refman/8.4/en/cast-functions.html), [SQLServer](https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver16), and [SQLite3](https://www.sqlite.org/lang_expr.html#cast_expressions) (although MySQL and SQLServer do support `CONVERT` too).

In particular, type conversion mostly breaks into coercion (implicit) and casting (explicit), and we are performing casting here. So saying "type conversion" is like saying "we have a number", while we could be more precise and instead say the exact type of number we have, "integer" or "float". Also, regarding jargons, the documentation will be _full_ of jargons. "structs", "database", even "integers" and "floats" are jargons but it wouldn't be productive to assume familiarity with them. That said, I don't mind us expanding exactly what `cast` and `change` means but I do believe the terminology here aligns with industry. :)

lib/ecto/changeset.ex
```diff
@@ -147,8 +148,8 @@ defmodule Ecto.Changeset do
 
   ## Associations, embeds, and on replace
 
-  Using changesets you can work with associations as well as with embedded
-  structs. There are two primary APIs:
+  Using changesets you can work with associations as well as with
+  [embedded](embedded-schemas.html) structs. There are two primary APIs:
```
josevalim, line 152: 💯

lib/ecto/changeset.ex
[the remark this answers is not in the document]
jdmarshall: We aren’t in the database though. We’re in docs that someone who has just run phx.gen for the first time and is now trying to learn a bunch of jargon terms.

Type conversion I copied and pasted from three paragraphs down.

I’ve only run into this definition of casting in Ruby, and while you have been using Ruby for what, 20 years? Elixir users may be coming from NodeJS or Python or PHP.

lib/ecto/changeset.ex
```diff
@@ -1,7 +1,8 @@
 defmodule Ecto.Changeset do
   @moduledoc ~S"""
-  Changesets allow filtering, casting, validation and
-  definition of constraints when manipulating structs.
+  Changesets allow filtering, type conversion, validation and
+  definition of constraints when manipulating structs, usually in
+  preparation for insertion into the database.
 
   There is an example of working with changesets in the introductory
   documentation in the `Ecto` module. The functions `cast/4` and
```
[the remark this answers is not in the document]
jdmarshall: Also I thought you could block select lines but maybe I’m confusing it with bitbucket?

lib/ecto/changeset.ex
jdmarshall: This link was suggested to me by someone and it helps a lot. The problem I suspect is it’s in another tab in the menu. I spent ten minutes this morning trying to decide if there was another way to organize those to make the guides more obvious to someone coming in via a Google search, and… I’ve got nothing that isn’t worse than the status quo. But I might double check the media queries.  I think I was running into horizontal scrollbars on desktop and I need to doublecheck.

lib/ecto/changeset.ex
```diff
@@ -18,14 +21,14 @@ defmodule Ecto.Changeset do
...
       a form that needs to be type-converted and properly validated. This
       use case is primarily covered by the `cast/4` function.
 
+   * internal to the application - for example programmatically generated,
+      or coming from other subsystems. This use case is primarily covered
+      by the `change/2` and `put_change/3` functions.
+
   When working with external data, the data is typically provided
   as maps with string keys (also known as parameters). On the other hand,
   when working with internal data, you typically have maps of atom keys
```
jdmarshall, line 34: test

lib/ecto/changeset.ex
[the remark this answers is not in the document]
josevalim: My point is precisely that we are using this nomenclature because of the database, not because of Ruby, my experience with Ruby (which I haven't touched in more than a decade) is not relevant to this discussion at all. It is a wrong assumption and I still do not understand why it is being brought up. I will drop a comment below.

lib/ecto/changeset.ex
```diff
@@ -1,14 +1,17 @@
...
-  Changesets allow filtering, casting, validation and
-  definition of constraints when manipulating structs.
-
-  There is an example of working with changesets in the introductory
-  documentation in the `Ecto` module. The functions `cast/4` and
-  `change/2` are the usual entry points for creating changesets.
-  The first one is used to cast and validate external parameters,
-  such as parameters sent through a form, API, command line, etc.
-  The second one is used to change data directly from your application.
+  Changesets allow filtering, type conversion, validation and
```
josevalim, line 3:
```suggestion
  Changesets allow filtering, type casting, validation and
```

Let's please either call it "casting" or "type casting", as that's the term we consistently use through the docs. We already detail it below now, which should be enough.

lib/ecto/changeset.ex
```diff
@@ -1,14 +1,17 @@
...
+  definition of constraints when manipulating structs, usually in
+  preparation for inserting and updating entries into a database.
+
+  There is an example of
+  [working with changesets](Ecto.html#module-changesets) in the
```
josevalim, line 8:
```suggestion
  [working with changesets](`Ecto#module-changesets`) in the
```

lib/ecto/changeset.ex
```diff
@@ -1,14 +1,17 @@
...
+  introductory documentation in the `Ecto` module. In a nutshell, there
+  are two main functions for creating a changeset. The `cast/4` function
+  is used to receive external parameters from a form, API or command line
```
josevalim, line 11:
```suggestion
  is used to receive external parameters from a form, API or command line,
```

lib/ecto/changeset.ex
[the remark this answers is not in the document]
warmwaffles: > We aren’t in the database though.

You have to realize that Ecto was built originally, exclusively to interact with a Database. It's only fairly recently in the last 5 years. There are tons of this lineage still hanging around in the `Repo` that are only pertinent to an SQL database and don't quite work in document stores.

> Java and I think C# are just swizzling pointers most of the time they are “casting”.

Those are people coming from a statically typed language. They should know the difference between the two and isn't a reason to change the "jargon" alone.

lib/ecto/changeset.ex
jdmarshall: Still learning hexdoc syntax

lib/ecto/changeset.ex
jdmarshall: If you don't want this I obviously can't make you, but I will just say again, in different words: this is trying to be technically precise in the part of the documentation that will be most seen by people who need a 5th grade English description of the purpose of this file, and be glossed over by the people who have graduated to using it purely as a reference.

Also it is referred to as type conversion in the External section and in several other places in the docs, and a bunch of places in the issues database. I didn't make it up, or introduce it, I borrowed from what is already there.

josevalim: Correct. It is ok if we expand and use alternative names later on. But this is literally the one sentence summary. We should be precise and use appropriate naming. The breakdown comes later.

josevalim: Another way to put it down: the first sentence is what will briefly recap what you read in the past, when you are browsing through the docs or when you need to remember exactly what this module does. It is unreasonable for the audience you describe to expect them to read the first sentence and be like "oh, I know exactly what it does". They have to read the rest of the document.

lib/ecto/changeset.ex
josevalim: Adding more hyper links is good. If you are new to a library, there is so much you can learn, and you don't have the complete understand of where to go next and how things relate, the links help with that. :+1:

lib/ecto/changeset.ex
jdmarshall: > Another way to put it down: the first sentence is what will briefly recap what you read in the past, when you are browsing through the docs or when you need to remember what this module does, and casting is a very 

Introductions are meant to be vague and refined as you go. 'Recapping' a term that hasn't been defined yet is not recapping.

"What is a changeset?" now requires me to learn what casting is. Which is probably okay if "What is a changeset?" isn't already the second or third question in a chain of terminology yak shaving.

But that's exactly where I was when I was reading this section, and exactly the sorts of conversations I've had with other people making similar edits to internal company docs. Many, many times. 

I'll change it back, but I'm doing it under duress.  This revert undermines most of the changes made further down.

lib/ecto/changeset.ex
jdmarshall: I still haven't figured out the bug that brought me here in the first place.

josevalim:
Thank you for the pull request @jdmarshall.

However, I would like to ask you to be respectful of all maintainers. We don't believe it is more important to be accurate than accessible, we believe we should do both, a consistent language is also important. However, we have _no proof_ that "type conversion" is any more accessible than "type casting". I am sure it is more accessible to _some_, including you, but it is hard to generalize to all of the backgrounds that may learn. Don't assume we make decisions based on Ruby, even when we provide evidence that the name was aligned with SQL databases. Ecto. We understand Ecto is divisive, some people love its abstractions, some people really struggle getting into it, and we want to improve it. We want to receive feedback and improve it but we should keep a respectful tone.

josevalim:
I have pushed a separate commit that breaks down each of the items in the first sentence, because we should not expect any word, either "type casting" or "type conversion", to do the work that needs to be actually done by a whole paragraph. We should not assume the user will know what filtering and constraints mean either: https://github.com/elixir-ecto/ecto/commit/6c612ca358b567242a13fee1fcc3fceb2edce6a6

PR #4583, opened 2025-02-25
Change introduction doc grouping

TylerWitt:
The introduction doc grouping reads as a step-by-step tutorial for creating the `friends` example.

The Embedded Schema docs being included can be confusing for someone using the bottom nav, since they would get moved from the friends example to an unrelated user profile example for embedded schemas. It seems to be more fitting as a how-to.

I also moved out the testing docs for similar reasons. In Phoenix, testing has its own section, so I did the same for Ecto (since the testing documentation is disconnected from `friends`.


josevalim:

PR #4586, opened 2025-03-01
Load virtual into schema struct

jswanner:
References:
- https://github.com/elixir-ecto/ecto/issues/4398

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -2344,6 +2344,7 @@ defmodule Ecto.Schema do
           {name, type}
         end
       end
+      |> Keyword.merge(virtual_fields)
 
     dump =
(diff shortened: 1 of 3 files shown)


josevalim:
It looks good to me, although I am a bit worried about loading virtual fields. Do you have thoughts @greg-rychlewski? Worst case scenario we require it to be opt-in via a flag.

greg-rychlewski:
I think it would have to be opt-in because the function is shared by embedded loading. So if it wasn't opt-in virtual embed fields would start appearing in results.

Using the schema as the type definition feels a bit off to me if wanting to load virtual fields. They are ignored in all the other Schema API functions, as far as I can see.

Maybe it would make sense to expand the reflection api so the user can create their own type map?

Issue #4591, opened 2025-03-12
Wrong binding when a query uses join with a subquery


freevova:
### Elixir version

elixir 1.16.2-otp-26

### Database and Version

postgres:15.7

### Ecto Versions

latest

### Database Adapter and Versions (postgrex, myxql, etc)

postgrex

### Current behavior

This is an example of the code:
```elixir
defmodule MyTest do
  def run_wrong_query do
    query = User

    from tasks in base_query(),
      join: users in subquery(
        from users in query, 
          join: profiles in assoc(users, :profile),
          select: users.id
      ),
      on: tasks.user_id == users.id,
      select: {tasks.id, users.id}
  end

  def run_right_query do
    from tasks in base_query(),
      join: users in subquery(
        from users in User, 
          join: profiles in assoc(users, :profile),
          select: users.id
      ),
      on: tasks.user_id == users.id,
      select: {tasks.id, users.id}
  end

  defp base_query do
    from tasks in Task,
      join: items in assoc(tasks, :item)
  end
end

```

When I run `MyTest.run_wrong_query()` I get:
```elixir
#Ecto.Query<from t0 in Task,
 join: i1 in assoc(t0, :item),
 join: u2 in subquery(from u0 in User,
  join: p1 in assoc(u0, :profile),
  select: u0.id),
 on: u2.id == t0.task_id, select: {t0.id, i1.id}>
```

And when I run `MyTest.run_right_query()` I get:
```elixir
#Ecto.Query<from t0 in Task,
 join: i1 in assoc(t0, :item),
 join: u2 in subquery(from u0 in User,
  join: p1 in assoc(u0, :profile),
  select: u0.id),
 on: u2.id == t0.task_id, select: {t0.id, u2.id}>
```

### Expected behavior

In the `MyTest.run_wrong_query()` I actually expect select to be `{t0.id, u2.id}>` but I `get {t0.id, i1.id}>`

PR #4592, opened 2025-03-14
Fix join_count conflicts

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4591

To me it looks like this is happening:

1. The `from` source can only be determined at runtime, so we don't know the starting join count. We quote the expression used to evaluate the join count.
2. The count from (1) is stored in a variable `join_count` in the join builder module.
3. Now let's say we have a subquery that also has a `from` source that can only be determined at runtime. This is setting the variable `join_count` to what makes sense for it. And this is conflicting with (2)

So the solution I made is to not set a `join_count` variable and instead evaluate the `Builder.count_binds` function each time.

### DIFF
--- a/test/ecto/query/builder/join_test.exs
+++ b/test/ecto/query/builder/join_test.exs
@@ -202,4 +202,17 @@ defmodule Ecto.Query.Builder.JoinTest do
       end, [], __ENV__)
     end
   end
+
+  test "join count is incremented correctly for runtime sources" do
+    users_table = "users"
+    source_query = fn -> from p in "posts", join: c in "comments", on: true end
+
+    q =
+      from s in source_query.(),
+        join: j in subquery(from u in users_table, join: p in "profiles", on: true, select: u.id),
+        on: true,
+        select: {s.id, j.id}
+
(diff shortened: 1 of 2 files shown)


test/ecto/query/builder/join_test.exs
```diff
@@ -202,4 +202,17 @@ defmodule Ecto.Query.Builder.JoinTest do
...
+    users_table = "users"
+    source_query = fn -> from p in "posts", join: c in "comments", on: true end
+
+    q =
+      from s in source_query.(),
+        join: j in subquery(from u in users_table, join: p in "profiles", on: true, select: u.id),
+        on: true,
+        select: {s.id, j.id}
+
+    assert {:{}, _, [_, {{:., [], [{:&, [], [2]}, :id]}, [], []}]} = q.select.expr
```
greg-rychlewski, line 216: Without the change this will fail because it will be `{{:., [], [{:&, [], [2]}, :id]}, [], []}`

The `join_count` is reset to 1 by the join in the subquery

josevalim (approved):
This is one where I trust CI more than myself. So if CI is green, ship it!

PR #4594, opened 2025-03-18
compile time comments

dkuku:
compile time commnts


gmile:
@dkuku out of curiosity, why was this closed?

PR #4596, opened 2025-03-29
Ecto.Enum.cast_value/3

cjbottaro:
@josevalim Is this what you meant by adding a function to Ecto to make this easier
https://github.com/phoenixframework/phoenix_html/issues/461#issuecomment-2764189902

Let me know if this is what you were thinking or any changes need to be made!

### DIFF
--- a/lib/ecto/enum.ex
+++ b/lib/ecto/enum.ex
@@ -280,9 +280,9 @@ defmodule Ecto.Enum do
   end
 
   @doc """
-  Helper to cast "dump values" to their corresponding value.
+  Helper to cast values through `cast/2`.
 
-  Note that if `value` is already a valid casted value, it returns `{:ok, value}`.
+  Provides an ergonomic API to `cast/2`.
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


cjbottaro:
Originally I was thinking to define `cast_value/3` in terms of `cast/2`, but it felt "heavy" having to call `init/1` and figuring out the options with which to call it. But maybe the "heaviness" isn't as bad as I imagine?

cjbottaro:
Ahh, I figured out that you can get the params via the __changeset__() function. So now `cast_value/3` is defined in terms of `cast/2`.

lib/ecto/enum.ex
```diff
@@ -279,6 +279,55 @@ defmodule Ecto.Enum do
     |> Keyword.values()
   end
 
+  @doc """
+  Helper to cast values through `cast/2`.
+
+  Provides an ergonomic API to `cast/2`.
```
josevalim, line 285: I think `cast/2` is hidden from docs:

```suggestion
  Casts a value from the given `schema` and `field`.
```

lib/ecto/enum.ex
```diff
@@ -317,25 +366,31 @@ defmodule Ecto.Enum do
...
-      %{^field => _} -> raise ArgumentError, "#{field} is not an Ecto.Enum field"
-      %{} -> raise ArgumentError, "#{field} does not exist"
-    end
+  def mappings(schema_or_types, field) do
+    get_params(schema_or_types, field)
+    |> Map.fetch!(:mappings)
   end
 
-  def mappings(schema, field) when is_atom(schema) do
+  @spec get_params(module | map, atom) :: map | no_return
```
josevalim, line 374: We don't spec private functions :)

```suggestion
```

josevalim (approved):
I am happy! Let's see if other maintainers agree too!

cjbottaro:
On really nitpicky thing... should the function be called `cast_data` instead of `cast_value`? `data` seems to be more inline with the terminology used throughout the file.

josevalim:
I am happy with `cast_value` but let's see if someone has a different opinion.

greg-rychlewski (approved):
I like `cast_value`. It seems more inline with the other function names.

josevalim:

PR #4597, opened 2025-03-30
Document `:log_stacktrace_mfa`

v0idpwn:
Ecto-side documentation for https://github.com/elixir-ecto/ecto_sql/pull/658

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -69,9 +69,10 @@ defmodule Ecto.Repo do
 
     * `:log_stacktrace_mfa` - A `{module, function, arguments}` tuple that customizes
       which part of the stacktrace is included in query logs. The specified function
-      must accept at least two arguments (a stacktrace and the repo module) and return
-      a filtered stacktrace. Additional arguments can be passed in the third element
-      of the tuple. For `Ecto.Adapters.SQL`, defaults to
+      must accept at least two arguments (stacktrace and metadata) and return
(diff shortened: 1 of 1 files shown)


lib/ecto/repo.ex
```diff
@@ -64,9 +64,18 @@ defmodule Ecto.Repo do
...
 
-    * `:stacktrace`- when true, publishes the stacktrace in telemetry events
+    * `:stacktrace`- when `true`, publishes the stacktrace in telemetry events
       and allows more advanced logging.
 
+    * `:log_stacktrace_mfa` - A `{module, function, arguments}` tuple that customizes
+      which part of the stacktrace is included in query logs. The specified function
+      must accept at least two arguments (a stacktrace and the repo module) and return
+      a filtered stacktrace. Additional arguments can be passed in the third element
+      of the tuple. For `Ecto.Adapters.SQL`, defaults to
```
josevalim, line 74:
```suggestion
    * `:log_stacktrace_mfa` - A `{module, function, arguments}` tuple that customizes
      which part of the stacktrace is included in query logs. The specified function
      must accept at least two arguments (stacktrace and metadata) and return
      a filtered stacktrace. The metadata is a map with keys such as `:repo` and other
      adapter specific information. Additional arguments can be passed in the third
      element of the tuple. For `Ecto.Adapters.SQL`, defaults to
```

josevalim: I propose one small change in case we need to pass more information in the future and you can merge it!

josevalim (approved):

PR #4599, opened 2025-04-02
Add @schema_redact: :all_except_primary_keys module attribute to Ecto Schema

Nezteb:
Initial proposal: https://groups.google.com/g/elixir-ecto/c/kQJN-7HPn6E

@josevalim initially approved of the second of my two proposed options:
```
@schema_redact :all_except_primary_keys
schema "users" do
  # ...
end
```

While implementing, I realized that it's probably not best to redact _all_ fields. I opted to keep the primary keys of a schema even if `@schema_redact` is set to something like `:all`, hence the rename to `:all_except_primary_keys`.

TODO:
- [X] Document new property once implementation/naming is settled

Future ideas:
- Actually add `@schema_redact: :all`, but would that be useful?
- Add separate attributes only deriving inspect for a list of specific fields?

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -130,6 +130,9 @@ defmodule Ecto.Schema do
   inspect on the schema unless the schema module is tagged with
   the option `@derive_inspect_for_redacted_fields false`.
 
+  A schema module tagged with `@schema_redact :all_except_primary_keys` will
+  redact all fields except primary keys.
+
   ## Schema attributes
 
   Supported attributes for configuring the defined schema. They must
@@ -154,6 +157,10 @@ defmodule Ecto.Schema do
     * `@schema_context` - configures the schema context. Defaults to `nil`,
       which generates structs and queries without context. Context are not used
       by the built-in SQL adapters.
+
+    * `@schema_redact` - If set to `:all_except_primary_keys`, Ecto will
+      treat all non-primary key fields as if they were individually marked
+      as redacted.
 
     * `@foreign_key_type` - configures the default foreign key type
       used by `belongs_to` associations. It must be set in the same
@@ -1992,7 +1999,14 @@ defmodule Ecto.Schema do
     writable = opts[:writable] || :always
     put_struct_field(mod, name, Keyword.get(opts, :default))
 
-    if Keyword.get(opts, :redact, false) do
+    redact_field? = Keyword.get_lazy(opts, :redact, fn ->
+      case Module.get_attribute(mod, :schema_redact) do
+        :all_except_primary_keys -> not pk?
+        _ -> false
+      end
+    end)
+
+    if redact_field? do
       Module.put_attribute(mod, :ecto_redact_fields, name)
     end
 
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -160,7 +160,8 @@ defmodule Ecto.Schema do
 
     * `@schema_redact` - If set to `:all_except_primary_keys`, Ecto will
       treat all non-primary key fields as if they were individually marked
-      as redacted.
+      as redacted. Defaults to `false`, as no fields are redacted by default.
+      The value set here can be changed per field through the `:redact` option.
 
     * `@foreign_key_type` - configures the default foreign key type
       used by `belongs_to` associations. It must be set in the same
@@ -2000,9 +2001,9 @@ defmodule Ecto.Schema do
     put_struct_field(mod, name, Keyword.get(opts, :default))
 
     redact_field? = Keyword.get_lazy(opts, :redact, fn ->
-      case Module.get_attribute(mod, :schema_redact) do
+      case Module.get_attribute(mod, :schema_redact, false) do
         :all_except_primary_keys -> not pk?
-        nil -> false
+        false -> false
       end
     end)
 
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -2002,7 +2002,7 @@ defmodule Ecto.Schema do
     redact_field? = Keyword.get_lazy(opts, :redact, fn ->
       case Module.get_attribute(mod, :schema_redact) do
         :all_except_primary_keys -> not pk?
-        _ -> false
+        nil -> false
       end
     end)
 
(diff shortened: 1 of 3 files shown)


solnic:
Fantastic! PKs by default sounds very reasonable. Thank you for working on this 💜

josevalim:
Hi @Nezteb, in this case, I would call it `:all_except_primary_keys`.

Btw, implementation wise, I'd implement it here:

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/schema.ex#L1995-L1997

So it does something like this:

    Keyword.get(opts, :redact, fn -> compute_default_redact_from_module_attribute(...) end)

This way, you don't need to change changesets nor nothing else, we reuse the existing redact infrastructure. WDYT?

lib/ecto/changeset.ex
```diff
@@ -4370,6 +4359,17 @@ defimpl Inspect, for: Ecto.Changeset do
...
+  defp get_redacted_fields(%type{}) do
+    type
+    |> function_exported?(:__schema__, 1)
+    |> get_redacted_fields(type)
+  end
+
+  defp get_redacted_fields(_), do: []
+
+  defp get_redacted_fields(true, type), do: type.__schema__(:redact_fields)
+  defp get_redacted_fields(_, _type), do: []
```
Nezteb, line 4371: Now that implementation is moved to the schema module, I figured I'd leave this here as cleanup. 😄

lib/ecto/schema.ex
```diff
@@ -1991,8 +1991,10 @@ defmodule Ecto.Schema do
     pk? = opts[:primary_key] || false
     writable = opts[:writable] || :always
     put_struct_field(mod, name, Keyword.get(opts, :default))
+    redact_primary_key? = Module.get_attribute(mod, :redact) == :all_except_primary_keys and not pk?
+    redact_field? = Keyword.get(opts, :redact, false)
 
-    if Keyword.get(opts, :redact, false) do
+    if redact_primary_key? or redact_field? do
       Module.put_attribute(mod, :ecto_redact_fields, name)
     end
```
Nezteb, line 1999: @josevalim What do you think of this? 😄

lib/ecto/changeset.ex
josevalim: Let's keep it as it was, as it is an unrelated change as of now (and I am not particularly sure it is cleaner, as it has more indirection). :)

lib/ecto/schema.ex
```diff
@@ -2324,9 +2326,9 @@ defmodule Ecto.Schema do
 
     struct_fields = Module.get_attribute(module, :ecto_struct_fields) |> Enum.reverse()
     derive = Module.get_attribute(module, :derive)
+    already_derives_inspect? = List.keymember?(derive, Inspect, 0)
 
-    if redacted_fields != [] and not List.keymember?(derive, Inspect, 0) and
-         derive_inspect?(module) do
+    if redacted_fields != [] and not already_derives_inspect? and derive_inspect_redacted?(module) do
```
josevalim, line 2331: These changes are also no longer required I believe, let's undo them! :)

lib/ecto/schema.ex
josevalim: We only need to read the module attribute if there is no option, so the approach I originally suggested should be better suited:

```elixir
    redact_field? = Keyword.get_lazy(opts, :redact, fn ->
      Module.get_attribute(mod, :schema_redact) == :all_except_primary_keys and not pk?
    end)
```

Also, let's call the option `@schema_redact`, closer to the other module attributes, and let's add some documentation to the module docs (there is already a section on module attributes).

lib/ecto/changeset.ex
Nezteb: Done!

lib/ecto/schema.ex
Nezteb: I'm always blown away at how succinct your implementations are; great idea!

Updated. 😄

lib/ecto/schema.ex
Nezteb: Undone!

lib/ecto/schema.ex
```diff
@@ -1992,7 +1999,14 @@ defmodule Ecto.Schema do
     writable = opts[:writable] || :always
     put_struct_field(mod, name, Keyword.get(opts, :default))
 
-    if Keyword.get(opts, :redact, false) do
+    redact_field? = Keyword.get_lazy(opts, :redact, fn ->
+      case Module.get_attribute(mod, :schema_redact) do
+        :all_except_primary_keys -> not pk?
+        _ -> false
```
josevalim, line 2005: So we raise with unknown values!

```suggestion
        nil -> false
```

Nezteb: That causes the test for `@schema_redact false` to fail. 😅

Nezteb: Or are you saying I should raise an exception here? And write a corresponding test? Can do!

lib/ecto/schema.ex
```diff
@@ -1992,7 +1999,14 @@ defmodule Ecto.Schema do
     writable = opts[:writable] || :always
     put_struct_field(mod, name, Keyword.get(opts, :default))
 
-    if Keyword.get(opts, :redact, false) do
+    redact_field? = Keyword.get_lazy(opts, :redact, fn ->
+      case Module.get_attribute(mod, :schema_redact) do
+        :all_except_primary_keys -> not pk?
+        nil -> false
+      end
```
josevalim, line 2006:
```suggestion
      case Module.get_attribute(mod, :schema_redact, false) do
        :all_except_primary_keys -> not pk?
        false -> false
      end
```

lib/ecto/schema.ex
```diff
@@ -155,6 +158,10 @@ defmodule Ecto.Schema do
       which generates structs and queries without context. Context are not used
       by the built-in SQL adapters.
 
+    * `@schema_redact` - If set to `:all_except_primary_keys`, Ecto will
+      treat all non-primary key fields as if they were individually marked
+      as redacted.
```
josevalim, line 163:
```suggestion
      as redacted. Defaults to `false`, as no fields are redacted by default.
      The value set here can be changed per field through the `:redact` option.
```

lib/ecto/schema.ex
josevalim: Indeed, I changed it to default to `false`. :) I think we are good now!

josevalim:

PR #4604, opened 2025-04-12
Allow fields in json_extract_path

greg-rychlewski:
companion to https://github.com/elixir-ecto/ecto_sql/pull/660 in Ecto SQL


josevalim (approved):
Very nice! :heart: I think we didn't consider this syntax before.

It may not be supported in all adapters but I think that's fine. Unsupported ones can raise.

PR #4605, opened 2025-04-14
Add Repo.prepare_transaction/2

milmazz:
### What I'm doing

Adding a new callback `prepare_transaction/2` to `Ecto.Repo`.

### Why?

To offer the possibility of further modifying the given `Ecto.Multi` or options in transaction operations before it is transformed and sent to the database. @josevalim [suggested](https://github.com/elixir-ecto/postgrex/pull/709#issuecomment-2773787388) to consider adding this function to inject the `commit_comment` option if the users wanted to extract that comment based on the `stacktrace`.

### Approach

1. Create the new `prepare_transaction/2` callback.
2. Provide a default implementation of the callback.
3. Invoke that callback from the `transaction` operation in `Ecto.Repo`.
4. Add unit tests.
5. Update the `Ecto.TestAdapter` to include the `opts` given to the transaction operation.
6. Update broken unit tests (after the previous change).

### Alternate approaches considered

As mentioned in this [discussion](https://github.com/elixir-ecto/postgrex/pull/709#issuecomment-2773559339), if the user wants to inject a `commit_comment` based on the `stacktrace`, the only option available today is the following _hack_:

```elixir
defoverridable transaction: 1, transaction: 2

def transaction(fun_or_multi, opts \\ []) do
  # The opts[:stacktrace] option is not available at this point, so, we probably have to invoke
  # Ecto.Repo.Supervisor.tuplet/2 twice. Sad times :-(
  # Is that, or do we avoid the `super` at the end and do everything ourselves here?
  opts = Keyword.put_new_lazy(opts, :commit_comment, fn -> extract_comment(opts) end)

  super(fun_or_multi, opts)
end
```

And we certainly don't want to promote these things ;)

### I’d like feedback on

The current implementation of `prepare_transaction/2` takes a `fun_or_multi` as the first argument, but it doesn't make too much sense to accept the `fun` part (a function with arity zero or one) because I don't see how that would be useful or how it can be extended.  The original implementation of `prepare_transaction/2` accepts a `fun` to keep things as simple as possible. As always, I'm open to suggestions.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -292,12 +292,12 @@ defmodule Ecto.Repo do
         def transaction(fun_or_multi, opts \\ []) do
           repo = get_dynamic_repo()
 
-          Ecto.Repo.Transaction.transaction(
-            __MODULE__,
-            repo,
-            fun_or_multi,
+          {adapter_meta, opts} =
             Ecto.Repo.Supervisor.tuplet(repo, prepare_opts(:transaction, opts))
-          )
+
+          {fun_or_multi, opts} = prepare_transaction(fun_or_multi, opts)
+
+          Ecto.Repo.Transaction.transaction(__MODULE__, repo, fun_or_multi, {adapter_meta, opts})
         end
 
         def in_transaction? do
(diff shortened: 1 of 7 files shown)


josevalim (approved):

josevalim:

PR #4607, opened 2025-04-16
Clarify the interaction between `reload` and associations

PragTob:
I was pretty sure what the behavior was, went ahead and tested it but thought it may be helpful to add it to the docs as it could also be reasonable to assume that all preloaded associations are reloaded as well.

Thanks for all your work! :green_heart:


PragTob:
IMG_20171221_144956_Bokeh

greg-rychlewski (approved):

greg-rychlewski:
That is a good callout! Thanks

PR #4613, opened 2025-06-01
Add Ecto migrations formatter config example to "Getting Started" guide

arcanemachine:
After merging #4612, I took a look around to find instructions for adding the migration-specific formatter rules to a vanilla Ecto project.

Surprisingly, a few Google searches didn't turn up any instructions (that I could find)... so I figured it might just be easier to add the instructions to the part of the Getting Started guide that deals with migrations.

It does break the flow of the tutorial briefly, but I wasn't sure if there was a better place for it. Feel free to make any needed modifications.

### DIFF
--- a/guides/introduction/Getting Started.md
+++ b/guides/introduction/Getting Started.md
@@ -155,7 +155,31 @@ defmodule Friends.Repo.Migrations.CreatePeople do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


arcanemachine:
Works for me!

PR #4618, opened 2025-06-11
Add Repo.transact/2 and soft-deprecate transaction/2

wojtekmach:


lib/ecto/repo/transaction.ex
warmwaffles, line 1: I only see the multi test being updated. Can we add some smoke tests for the function with arity 0 and 1 cases?

warmwaffles: Oh I got so damn confused, it was first on the list of changes. I usually look for tests at the bottom because `test/` comes after `lib` 🤦🏻

dustinfarris:
Is the motivation for this documented anywhere?  Trying to get up to speed.

warmwaffles:
@wojtekmach probably wise to link to the original PR in the body or just copy paste the old PR description here for the next passer-by

PR #4624, opened 2025-06-21
Add support for :writable option on belongs_to

sheharyarn:
PR for my [proposal](https://groups.google.com/g/elixir-ecto/c/JDL69kx5nTA) to add `:writable` option to `belongs_to`.

This PR simply whitelists the `:writable` option in `@valid_belongs_to_options`, which has already been implemented in #4335 & #4472.

`cast_assoc` & `put_assoc` currently do not respect this option, so if it makes sense, I can send a follow-up PR for that.

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -1106,6 +1106,12 @@ defmodule Ecto.Schema do
     * `:where` - A filter for the association. See "Filtering associations"
       in `has_many/3`.
 
+    * `:writable` - Defines when the underlying field is allowed to modified, Must be
+      one of `:always`, `:insert`, or `:never`. If set to `:always`, the field can be
+      modified by any repo operation. If set to `:insert`, the field can be inserted
+      but cannot be further modified, even in an upsert. If set to `:never`, the field
+      becomes read only. Defaults to `:always`.
+
   ## Examples
 
       defmodule Comment do
@@ -2138,6 +2144,7 @@ defmodule Ecto.Schema do
     :defaults,
     :primary_key,
     :source,
+    :writable,
     :where
   ]
 
--- a/test/ecto/repo/belongs_to_test.exs
+++ b/test/ecto/repo/belongs_to_test.exs
@@ -35,6 +35,15 @@ defmodule Ecto.Repo.BelongsToTest do
     end
   end
 
+  defmodule RestrictedSchema do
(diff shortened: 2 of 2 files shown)


josevalim:
Looks good to me, but I wonder if @greg-rychlewski has any additional thoughts?

> cast_assoc & put_assoc currently do not respect this option, so if it makes sense, I can send a follow-up PR for that.

How do you think it would be handled? I can still see someone using cast_assoc for something else, even though they never write the field, so perhaps we should not change anything.

josevalim:

greg-rychlewski:
Reminding myself I need to look to see if this is affected at all https://github.com/elixir-ecto/ecto/pull/4299

greg-rychlewski:
@josevalim There is a bit of a weird interaction. This test passes:

```elixir
test "insert unwritable assoc" do
    assoc = %MyAssoc{x: "xyz"}

    changeset =
      %RestrictedSchema{}
      |> Ecto.Changeset.change
      |> Ecto.Changeset.put_assoc(:unwritable, assoc)
    schema = TestRepo.insert!(changeset)
    assoc = schema.unwritable
    assert assoc.id
    assert assoc.id == schema.unwritable_id
  end
```

So if you insert the foreign key directly the value is ignored. If you insert the assoc then the foreign key is saved. 

I'm thinking it's weird if the assoc record is inserted but the foreign key silently isn't updated. And it's also weird if it just saves the value and ignores the option.

Maybe the best option is to also not insert the assoc? We might have to add a new `writable` key to the assoc struct or something like that.

greg-rychlewski:
I’ll give it a shot seeing what it would take to make it not written

sheharyarn:
I'm also happy to contribute if you can point me in the right direction.

greg-rychlewski:
It's also possible you can combine all those checks I mentioned above in the `pop_assocs` helper function in `lib/ecto/repo/schema.ex`. But I'm not too sure yet if that's possible. But that would be what I would try first. It could reduce the number of places we have to change code pretty dramatically

josevalim:
I reverted this for now, as we had to do a new release. Please include the old commit once you push the changes above. :) Thank you!

greg-rychlewski:
@josevalim The revert doesn't seem to be in the master branch

josevalim:
done!

PR #4627, opened 2025-06-23
Docs: Don't reference deprecated `literal`

ftes:
This is probably not the best fix for the docs.

Additional ideas (not done):
- use different wording instead of `literal`
- reference the `identifier` and `constant` function docs
- explain the difference between `identifier` and `constant`

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -399,7 +399,7 @@ defmodule Ecto.Query.API do
 
       fragment("lower(?)", p.title) == type(^title, :string)
 
-  ## Literals
+  ## Identifiers (literals)
 
   Sometimes you need to interpolate a literal value into a fragment,
   instead of a parameter. For example, you may need to pass a table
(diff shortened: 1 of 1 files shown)


greg-rychlewski:
Maybe something like this

```
## Identifiers and Constants

  Sometimes you need to interpolate an identifier or a constant value into a fragment,
  instead of a query parameter. The latter can happen if your database does not allow
  parameterizing certain clauses. For example:

      collation = "es_ES"
      fragment("? COLLATE ?", ^name, ^collation)

      limit = "10"
      "posts" |> select([p], p.title) |> limit(fragment("?", ^limit))

  The first example above won't work because `collation` needs to be quoted as an identifier. 
  The second example won't work on databases that do not allow passing query parameters
  as part of `limit`.

  You can address this by telling Ecto to treat these values differently than a query parameter:

      fragment("? COLLATE ?", ^name, identifier(^collation))
      "posts" |> select([p], p.title) |> limit(fragment("?", ^constant(limit))

  Ecto will make these values directly part of the query, handling quoting and escaping where necessary.

  > #### Query caching {: .warning}
  >
  > Because identifiers and constants are made part of the query, each different
  > value will generate a separate query, with its own cache.
```

ftes:
Thanks, I like hour suggestions and copy & pasted them into this PR in case you want to merge.

greg-rychlewski (approved):

PR #4637, opened 2025-07-01
Propagate opts configured in `prepare_query/3` to preloads

derekkraan:
This ensures that when `:on_preloader_spawn` is set in the `prepare_query/3` callback, it is propagated to the preloader.

(For avoidance of XY issues, I am trying to find a way to propagate opentelemetry tracing context into the preloads, so that preloaded queries can appear in my opentelemetry traces like normal, and not appear as orphaned spans. The `on_preloader_spawn` option appears to be designed for exactly this. But, there is currently no convenient way to set this globally. Currently it only works when you add it to the opts in Repo.all/3 etc)

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -644,11 +644,17 @@ defmodule Ecto.Repo do
         def preload(struct_or_structs_or_nil, preloads, opts \\ []) do
           repo = get_dynamic_repo()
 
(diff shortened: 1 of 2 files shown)


josevalim:
Fantastic, could we perhaps have a test? We have some unit tests in Ecto.Repo where we send messages or use the process dictionary for the options, perhaps this could be added there? Thank you.

PR #4639, opened 2025-07-11
Replace deprecated BadStructError by ArgumentError

sabiwara:
Reusing the same message as [`BadStructError`](https://github.com/elixir-lang/elixir/blob/v1.19.0-rc.0/lib/elixir/lib/exception.ex#L1460-L1472).

This is triggering a (slightly confusing) warning on the latest RC:
Screenshot 2025-07-12 at 7 32 02


integration_test/cases/repo.exs
```diff
@@ -1378,7 +1378,7 @@ defmodule Ecto.Integration.RepoTest do
         TestRepo.all(from p in Post, select: %{p.title | title: "new title"})
       end
 
-      assert_raise BadStructError, fn ->
+      assert_raise ArgumentError, fn ->
```
v0idpwn, line 1381:
```suggestion
      assert_raise ArgumentError, r/expected a struct named/, fn ->
```

v0idpwn (approved):

PR #4643, opened 2025-07-28
Provide function group descriptions in `Ecto.Repo`

rhcarvalho:
- Add descriptions to "Query API", "Schema API", etc.
- Move `Ecto.Repo.all_running/0` to "Process API" group.

---

I was reading the docs for `update_all` and was surprised to find it under "Query API", while other update-related callback functions are part of the "Schema API".

~I understand "Query API" is for all APIs related to reading from the database, and "Schema API" for writing. In that interpretation, `load`, `reload`, `reload!`, and `preload` would also fit better within Query API.~
Update: the groups are based on the data they operate on, see https://github.com/elixir-ecto/ecto/pull/4643#issuecomment-3126689881.

The `all_running/0` function was the only one outside any group, but it fits with the "Process API" group.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -214,6 +214,24 @@ defmodule Ecto.Repo do
 
   """
 
+  @moduledoc groups: [
+               %{title: "Query API", description: "Functions that operate on an `Ecto.Query`."},
+               %{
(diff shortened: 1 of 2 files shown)


josevalim:

PR #4644, opened 2025-08-06
docs: link `parent_as/1` to definition

kevinschweikert:
This just makes the `parent_as` clickable to jump to the definition


josevalim:

PR #4645, opened 2025-08-06
Tweak formatting and language in Multi

DavidOliver:
Capitalisation of 'Multi' in normal text was inconsistent. I opted to capitalise as I believe that, unlike 'query', 'schema', 'type', etc., 'multi' is not a "common noun". (If I've got the terminology correct here - I am not a linguist.)

In case it's important, a couple of error messages were also tweaked.

If any of this needs reverting or further tweaking, do let me know. Thanks.


josevalim:

PR #4653, opened 2025-09-06
ensure select statements in joined queries have access to parent bindings

zachdaniel:
Ultimately, I don't fully understand how parent binding sources are managed well enough to really say if this is the most correct fix. Since `parent_as` works in other, non-select contexts using this formulation, it feels like there may be a more targeted/specific fix available?

Regardless, the newly added test will fail without the fix, and the fix does not cause any additional test failures. Open to adjusting it if as needed.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -669,6 +669,8 @@ defmodule Ecto.Query.Planner do
         distinct: nil,
         lock: nil
       } ->
+        join_query = put_in(join_query.aliases[@parent_as], query)
+
         join_query = rewrite_prefix(join_query, query.prefix)
         from = rewrite_prefix(join_query.from, prefix)
         {from, source} = plan_source(join_query, from, adapter, cte_names)
--- a/test/ecto/query/planner_test.exs
+++ b/test/ecto/query/planner_test.exs
@@ -1459,6 +1459,20 @@ defmodule Ecto.Query.PlannerTest do
     assert Macro.to_string(hd(hd(query.joins).source.query.wheres).expr) =~ "in %Ecto.SubQuery{"
   end
 
+  test "normalize: parent_as in subquery wrapped in a query should not raise an error" do
+    child = from(p in Post,
+      where: p.status == ^:published,
+      select: %{field: parent_as(:outer).posted > ^~N[2025-01-01 00:00:00]}
+    )
+
+    query = from thing in Post, 
+      as: :outer, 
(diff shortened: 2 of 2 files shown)


josevalim:
Thanks for the PR! I would like to hear @greg-rychlewski's opinion but my gut feeling is that the test should not pass indeed. In this case:

```
    query = from thing in Post, 
      as: :outer, 
      inner_lateral_join: sub in ^(from row in subquery(child)),
      on: true
```

The parent query of `subquery(child)` is `from row in subquery(child)`. I don't think the subquery should bypass the boundary of its actual query into the parent one, unless you have another subquery. Otherwise, we are mixing composition rules with parent/child rules.

Forget subqueries for a second. Imagine you have `child`. When you do this:

```
another_query = from ...
query = from row in another_query
```

In this case, `query` and `another_query` are doing regular composition. They share the same `as`. It is not a parent/child relationship. `query` is not the parent of `another_query`.


Therefore, when you do this:

```
another_query = from ...
query = from Post, join: row in another_query
```

Once again, `another_query` is not subquery, `query` is not its parent. This PR changes them to be a child/parent relationship, which is just wrong, given by the fact there are no subqueries involved.

josevalim:
The walking upwards does not matter, because, as per my first comment, there is no parent/child relationship between the `from` and the query given in `join`. The parent/child relationship only happens when you wrap them in subqueries.

You have three queries:

    QUERY 1 = from thing in Post, ...
    QUERY 2 = from(row in subquery(child), join: "outers", as: :outer)
    QUERY 3 = child

QUERY 1 and 2 is regular query composition. There is no parent/child relationship in there. The same way `another_query` `query = from row in another_query`  is not a child of `query`.

greg-rychlewski:
Yeah after looking at this I have the same opinion as @josevalim. I think the crux is these two things are not the same

```
inner_lateral_join: s in subquery(q)
```

```
inner_lateral_join: s in ^(from s in subquery(q))
```

The subquery in the latter example doesn't know anything outside of `from s in subquery(q))`. 

I can understand the confusion though because if I was just writing the SQL by hand and joining on a query I would be forced to wrap the whole thing in a subquery or it would be invalid SQL. But with Ecto you have to think in terms of composition.

greg-rychlewski:
I think the issue is that in raw SQL the joined query must be in a subquery whether not it is lateral. So when Ecto is allowing a join on a query under any circumstance, it is performing composition rather than an implicit subquery.

PR #4657, opened 2025-09-13
Update Repo Documentation: Paranthesis on function calls

TylerPachal:
If any of these examples were in application code, calling `mix format` would add parenthesis.  So I think it makes sense to include the parenthesis in the examples to mimic how the usage would look in a code base.


josevalim:

PR #4659, opened 2025-09-18
Allow 2-ariety functions as preload function in Query

orlera:
As described in the error message, "Preloads can be a query, a function expecting one or two arguments".

Right now we are now allowing 2-ariety functions receiving `ids` and `assoc` as arguments.


orlera:
@josevalim do we know when v3.13.3 will be released?

Issue #4651, opened 2025-08-28
If a change fails a constraint with composite key only the first key name is reported in the errors


ponychicken:
### Elixir version

Elixir 1.18.4 (compiled with Erlang/OTP 27)

### Database and Version

--

### Ecto Versions

ecto 3.13.2

### Database Adapter and Versions (postgrex, myxql, etc)

--

### Current behavior

If you have a table with a composite pkey:

```

  def change do
    create table(:widgets, primary_key: false) do
      add :key1, :integer, primary_key: true
      add :key2, :integer, primary_key: true
      add :name, :string
    end
  end
```

and then set it in the changeset def
```

  @primary_key false
  schema "widgets" do
    field :key1, :integer, primary_key: true
    field :key2, :integer, primary_key: true
    field :name, :string
  end

  def changeset(widget, attrs) do
    widget
    |> cast(attrs, [:key1, :key2, :name])
    |> unique_constraint([:key1, :key2], name: :widgets_pkey)
  end
```

Then when a changeset fails the constraint you will receive a message like:
`#Ecto.Changeset<action: :insert, changes: %{name: "XXCCXX", key1: 1, key2: 2}, errors: [key1: {"has already been taken", [constraint: :unique, constraint_name: "widgets_pkey"]}], data: ....<>, valid?: false, ...>`

### Expected behavior

I dont know if you want to threat this as a bug, i found it confusing and it made me check the db if the constraint was set-up correctly. I thing either putting both keys as into the errors would be more clear.

greg-rychlewski:
Would it make sense to put another metadata field like `constraint_keys`

tfantina:
If this is is a feature worth adding I'd be interested in taking a crack at it.

PR #4661, opened 2025-09-21
Adds option to list composite keys for unique constraint

tfantina:
From #4651, this is a possible option for showing the composite keys in the error message. By adding all composite fields to the `fields` list in the error meta it should provide enough additional information about the composite keys making up the constraint.

 I'm happy to change the approach based on the recommendations of reviewers. Also I'm happy to forget this feature altogether but if there is any interest from the maintainers in this I'd like to try and make it happen.

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -450,6 +450,25 @@ defmodule Ecto.Integration.RepoTest do
 
   @tag :join
   @tag :unique_constraint
+  test "unique constraint with composite primary keys" do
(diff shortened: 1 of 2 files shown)


josevalim:
> When you say "perhaps if you stored them indeed in a separate called fields" do you mean something like a new meta field? 

Yes, exactly. Then we don't need an option. Because I also can't see how the current implementation would work, as it doesn't seem we support the field to be a list on the result of `add_constraint`.

But I am not sure if we should go ahead with this issue unless there is a use case for it. :) I am not sure if the proposed solutions in the original issue will solve the original problem.

PR #4663, opened 2025-09-27
Allow atoms (interpolated or not) as source in field/2

greg-rychlewski:
I wasn't 100% confident this is what you were thinking but please have a look!

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -603,6 +603,10 @@ defmodule Ecto.Query.API do
   @doc """
   Allows a field to be dynamically accessed.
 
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you @greg-rychlewski! This is exactly what I had in mind but I actually did not remember we already supported `as(:posts)`. Given that, perhaps we only need to support `as(^atom)` (perhaps we already do) and update `Ecto.Query.API.field/2` docs?

greg-rychlewski:
@josevalim We support both. Updated docs :)

josevalim (approved):
Thank you and sorry for sending you down the wrong rabbit hole!

PR #4678, opened 2025-11-08
support selecting a subset of a subquery as a struct

zachdaniel:
This increases the general flexibility of subquery selecting. For example, there might be a case where you want to do something like this:

```elixir
sub = from post in Post, select: post

from post in subquery(sub),
  where: post.special,
  select: struct(post, [:id, :name])
```

The `special` field being filtered outside of the subquery requires selecting it from the subquery, but it doesn't need to be present in the response.

That's a contrived example because you would just filter in the subquery, but real life has gotten me to this point in a more complicated scenario 😄

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -2208,7 +2208,7 @@ defmodule Ecto.Query.Planner do
         error!(query, "it is not possible to return a struct subset of a fragment")
 
       {{:ok, {:struct, fields}}, %Ecto.SubQuery{select: select}} ->
-        handle_subquery_struct(select, fields, ix, query)
+        subquery_select_fields(select, fields, ix, query)
 
       {{:ok, {_, []}}, {_, _, _}} ->
         error!(
@@ -2280,7 +2280,7 @@ defmodule Ecto.Query.Planner do
     end)
   end
 
-  defp handle_subquery_struct(select, requested_fields, ix, query) do
+  defp subquery_select_fields(select, requested_fields, ix, query) do
     available_fields = subquery_source_fields(select)
     requested_fields = List.wrap(requested_fields)
 


zachdaniel:
I don't think that integration test is failing due to my changes but there is something weird w/ the results, I can't see them or rerun the task.

lib/ecto/query/planner.ex
```diff
@@ -2280,6 +2280,39 @@ defmodule Ecto.Query.Planner do
     end)
   end
 
+  defp handle_subquery_struct(select, requested_fields, ix, query) do
+    available_fields = subquery_source_fields(select)
+    requested_fields = List.wrap(requested_fields)
+
+    Enum.each(requested_fields, fn field ->
+      unless field in available_fields do
```
```suggestion
      if field not in available_fields do
```

zachdaniel: lol, old habits dying hard 😆

lib/ecto/query/planner.ex
```diff
@@ -2280,6 +2280,39 @@ defmodule Ecto.Query.Planner do
...
+
+    types =
+      Enum.map(requested_fields, fn field ->
+        {:ok, type} = subquery_type_for(select, field) 
+        {field, type}
+      end)
+
+    field_exprs =
+      Enum.map(types, fn {field, type} ->
+        {{:., [type: type], [{:&, [], [ix]}, field]}, [], []}
```
josevalim: We should probably use `select_field` to build this information. In particular, we don't attach the type information anywhere in the other `take!` calls, so perhaps we shouldn't do so here.

lib/ecto/query/planner.ex
```diff
@@ -2280,6 +2280,39 @@ defmodule Ecto.Query.Planner do
...
+    schema =
+      case select do
+        {:struct, schema, _} -> schema
+        {:source, {_, schema}, _, _} when not is_nil(schema) -> schema
+        _ ->
+          error!(query, "it is not possible to return a struct subset of a subquery that does not return a schema or a struct")
+      end
+
+    types =
+      Enum.map(requested_fields, fn field ->
```
josevalim: Let's avoid double traversal of the fields. We can probably merge `subquery_type_for` and match on errors to detect the field is invalid.

lib/ecto/query/planner.ex
```diff
@@ -2207,8 +2207,8 @@ defmodule Ecto.Query.Planner do
       {{:ok, {:struct, _}}, {:fragment, _, _}} ->
         error!(query, "it is not possible to return a struct subset of a fragment")
 
-      {{:ok, {:struct, _}}, %Ecto.SubQuery{}} ->
-        error!(query, "it is not possible to return a struct subset of a subquery")
+      {{:ok, {:struct, fields}}, %Ecto.SubQuery{select: select}} ->
+        handle_subquery_struct(select, fields, ix, query)
```
josevalim: Given all functions seems to have a `select_` prefix, we should probably follow that here:

```suggestion
        select_subquery_struct(select, fields, ix, query)
```

Or `subquery_select_fields`.

lib/ecto/query/planner.ex
```diff
@@ -2280,6 +2280,35 @@ defmodule Ecto.Query.Planner do
     end)
   end
 
+  defp handle_subquery_struct(select, requested_fields, ix, query) do
+    available_fields = subquery_source_fields(select)
```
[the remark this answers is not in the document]
zachdaniel: 🤔 I don't think I'm currently adding the query to any error message, but maybe I'm missing it. Could you point out the line?

josevalim (approved):

lib/ecto/query/planner.ex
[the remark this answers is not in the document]
greg-rychlewski: oh sorry I meant that `available_fields` is only used here:

```elixir
error!(query, "field `#{field}` in struct/2 is not available in the subquery. " <>
                         "Subquery only returns fields: #{inspect(available_fields)}")
```

and i think it might be unnecessary because the `error!` function already prints the entire query. it raises `Ecto.QueryError` which takes care of it. so you are already told which fields exist in that way. 

it's not a big deal either way it was just a tiny cleanup (if my assumptions are right)

greg-rychlewski:
thanks!

PR #4682, opened 2025-11-19
has_one does not support preload_order option

bkilshaw:
Currently `has_one` and `has_many` both validate against the same `@valid_has_options` allowing you to pass a `preload_order` option into `has_one` despite it not being supported.

There is a PR (https://github.com/elixir-ecto/ecto/pull/4505) that's been open for over a year with some discussion around `has_one` supporting `preload_order` but it would probably be good to clean this up in the meantime.


greg-rychlewski:
Thanks :)

PR #4684, opened 2025-12-02
Fix `search_path` parameter description and example

Gigitsu:
This PR fixes an error I made in one of my previous PRs. The `:search_path` parameter is no longer a list of strings, but a comma separated string of prefixes.


greg-rychlewski (approved):
Thanks :)

PR #4693, opened 2026-01-04
Support compile-time fragment splicing

greg-rychlewski:
There are a couple of people in this old issue that would benefit from this change: https://github.com/elixir-ecto/ecto/issues/4375

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -529,7 +529,7 @@ defmodule Ecto.Query.API do
       from p in Post, where: fragment("? in (?,?,?)", p.id, ^1, ^2, ^3)
 
   You may also splice compile-time lists. This allows you to combine query parameters
-  with litreals and constructs like query bindings
+  with literals and constructs like query bindings
 
       sep = " "
       from p in Post, select: fragment("concat(?)", splice([p.count, ^sep, "count"]))


josevalim:
Hi @greg-rychlewski, there is something that feels a bit awkward about this. The issue is that dynamic is just one of the things we could potentially splice? What if we want to splice something that has a constant or an identifier? We are also mixing parameters with expressions (which I guess is something dynamic does anyway...) but this time it is nested within a list.

Anyway, I wonder if there is another way we could tackle this. For example, one idea would be to introduce `fragment_apply(:fun, [arg1, arg2, arg3])`, and we safely convert it into `fragment("fun(?, ?, ...)", ...)` without any of the issues here. Perhaps that would be even enough to deprecate splice?

greg-rychlewski:
Or if we created a new API to build up lists with DSL in them. I can't think of a good API at the moment but the "spirit" would be something like

```
list = dynamic_list(bindings, exprs) 
list = list |> append(bindings, expr1) |> prepend(bindings, expr2)
```

I'm not sure if that counts as less awkward than putting a bunch of dynamics into a list but just throwing anything I think of out there. Maybe this is really bad because people will eventually ask to duplicate all the Elixir list API.

josevalim:
Probably the biggest reason this doesn't work is because databases support splice but Elixir doesn't, so there is a mismatch.

For the original issue, perhaps what needs to be done is to move the splice to the caller:

    defmacro concat_ws(left, right) do
      quote do
        fragment("concat_ws(?, ?)", unquote(left), unquote(right))
      end
    end

Then we could make this work:

    concat_ws("123", splice([q.foo, q.bar]))

As well as this:

    concat_ws("123", splice([q.foo, ^runtime_value]))

Then in the query language:

  * `splice([...])` gets immediately spliced into the fragment (could even be done at compile time)
  * `splice(^list)` remain as is (or we splice it at runtime, so we don't have to change the adapter)

But the above should give a compile-time feature, no dynamic, and without adapter changes. We just need to make sure that a splice outside of a `fragment` raises (we could also make it so a `splice` on any query function just splices it too, so it becomes a generic feature, but I don't think we have such functions in our API today).

greg-rychlewski:
Oh that's a really good idea. Ok so to summarize:

1. I'll change this PR to do this `splice([...]) gets immediately spliced into the fragment (could even be done at compile time)`
2. If it's alright with you I'll send a separate PR to remove the need for the adapter to do anything with `splice(^list)`
3. After those two can think about `we could also make it so a splice on any query function just splices it too, so it becomes a generic feature` just to see if it makes sense

And then the caveat is we won't be able to put query bindings in `splice(^list)`  but we're making that trade off for a cleaner feature.

Does that seem ok to you as well?

josevalim:
> I'll change this PR to do this splice([...]) gets immediately spliced into the fragment (could even be done at compile time)

I think splice is either `splice(^list)` (runtime) or `splice([...])` (compile). And the compile-time can always be done at compile time. In the initial PR, I would only add the compile-time behaviour, don't change how the runtime one works currently (that can be another PR).

> And then the caveat is we won't be able to put query bindings in splice(^list) but we're making that trade off for a cleaner feature.

We could also allow a splice inside a splice. This means you could do this: `splice([q.foo, q.bar, splice(^runtime)])`.

---

:shipit:

greg-rychlewski:
@josevalim Ready for review!

josevalim (approved):
Beautiful!

PR #4694, opened 2026-01-04
Fix: Enable autogenerate options for binary_id primary keys

ChristopheNonato:
Hello!!
I'm Chris.

I'm new to Elixir and while creating a personal project in Phoenix I realized that there was only autogenerated uuidv4 available as of ecto 3.13.5. But while taking a look at the repo I saw a few PRs introducing UUIDv7 recently merged into master, that's great!

I tried it in my project but the following provided syntax was not working:
`@primary_key {:id, :binary_id, autogenerate: [version: 7, monotonic: true]}`

I realized that changing :binary_id to Ecto.UUID fixed it:
`@primary_key {:id, Ecto.UUID, autogenerate: [version: 7, monotonic: true]}`

But it seems to me that making it work with the :binary_id type is expected as this is the syntax provided in [uuid.ex](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/uuid.ex#L22), and as it works for uuidv4 with this syntax.

I started to investigate this issue from the following compilation error I was facing:
image

To fix this issue when passing autogenerate options to :binary_id primary keys this Pr introduces the following changes:
  - Modified lib/ecto/schema.ex to detect when autogenerate options are provided for ID types and route them to `store_autogenerate_id!` instead of `store_mfa_autogenerate!`, storing options as a 4-tuple {name, source, type, **opts**}
  - Updated lib/ecto/type.ex to add `adapter_autogenerate/3` that handles :binary_id with options by calling `Ecto.UUID.bingenerate(opts)` directly
  - Modified lib/ecto/repo/schema.ex and lib/ecto/embedded.ex to handle both 3-tuple (backward compatible) and 4-tuple formats, passing options through to the UUID generator

Not sure if it's the best approach but I wanted to propose a solution, not only create an issue describing the problem :)

### DIFF
--- a/test/ecto/schema_test.exs
+++ b/test/ecto/schema_test.exs
@@ -1298,7 +1298,7 @@ defmodule Ecto.SchemaTest do
     end
   end
 
-  test "binary_id with autogenerate options" do
+  test "binary_id with autogenerate options stores metadata correctly" do
     assert SchemaWithUuidV7PrimaryKey.__schema__(:autogenerate_id) ==
              {:id, :id, :binary_id, [version: 7, monotonic: true]}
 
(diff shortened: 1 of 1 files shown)


ChristopheNonato:
> Hi @ChristopheNonato, thank you for the PR!
> 
> `binary_id` means "use whatever the database chooses", which may not support UUID at all. So I think using Ecto.UUID is indeed the best way to go. Alternatively, we may make the default binary_id configurable, but I don't think we should allow options to be given or assume we can generate v7.

Got it @josevalim !
Thanks for the explanation!
I`ll close this PR then, I created another one just to change the syntax where it was not accurate:
https://github.com/elixir-ecto/ecto/pull/4695

PR #4696, opened 2026-01-09
Fragment constant/1 doesn't need adapter

greg-rychlewski:
Right now we are passing `{:constant, [], [value]}` to the adapter and the adapter is taking care of the unpacking. But we can pass the constant value directly from Ecto and the existing adapter functionality can take care of it without the need for handling a `:constant` tuple.


josevalim (approved):

josevalim:
Sounds good!

PR #4709, opened 2026-03-12
Correctly extract passwords with colons from URL

pnezis:
```elixir
iex(1)> Ecto.Repo.Supervisor.parse_url("ecto://user:top:secret@host:12345/mydb")
[
  hostname: "host",
  scheme: "ecto",
  username: "user",
  password: "top",
  database: "mydb",
  port: 12345
]
```

It correctly works if the password is URL encoded, but since passwords with colon are very common it is better to support them directly as well. 

Surprisingly I noticed that **`mix format` does not work in the repo** and there is no CI check for formatted code. The reason is that `.formatter.exs` is missing `:inputs`. Is this intentional?

### DIFF
--- a/lib/ecto/repo/supervisor.ex
+++ b/lib/ecto/repo/supervisor.ex
@@ -111,7 +111,7 @@ defmodule Ecto.Repo.Supervisor do
       raise Ecto.InvalidURLError, url: url, message: "path should be a database name"
     end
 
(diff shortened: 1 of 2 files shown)


josevalim:
> Surprisingly I noticed that mix format does not work in the repo and there is no CI check for formatted code. The reason is that .formatter.exs is missing :inputs. Is this intentional?

Yes, we are still formatting the codebase bit by bit.

PR #4714, opened 2026-04-05
chore: accept a `counter` option in `prepare_query/3`

zachdaniel:


lib/ecto/adapter/queryable.ex
```diff
@@ -104,13 +104,14 @@ defmodule Ecto.Adapter.Queryable do
   def prepare_query(operation, repo_name_or_pid, queryable, opts \\ []) do
     %{adapter: adapter, cache: cache} = Ecto.Repo.Registry.lookup(repo_name_or_pid)
 
-    query_cache? = Keyword.get(opts, :query_cache, true)
+    counter = Keyword.get(opts, :counter, 0)
+    query_cache? = if counter != 0, do: false, else: Keyword.get(opts, :query_cache, true)
```
zachdaniel: I had to disable the cache, but I'm wondering if we should maybe instead include the counter in the cache key?

lib/ecto/adapter/queryable.ex
```diff
@@ -104,13 +104,20 @@ defmodule Ecto.Adapter.Queryable do
...
 
-    query_cache? = Keyword.get(opts, :query_cache, true)
+    counter = Keyword.get(opts, :counter, 0)
+
+    query_cache? =
+      if counter != 0 do
+        false
+      else
+        Keyword.get(opts, :query_cache, true)
+      end
```
josevalim: Yes, we make it part of the cache and GG!

greg-rychlewski:
Out of curiosity what problem does this solve?

josevalim:
@zachdaniel is implementing `MERGE`, which is basically multiple insert, update, etc. queries spliced into one.

lib/ecto/adapter/queryable.ex
zachdaniel: Done 😄 I kept the original cache key unchanged if the index is 0 just for the sake of simplicity/not updating a bunch of tests etc.

lib/ecto/query/planner.ex
```diff
@@ -136,7 +136,14 @@ defmodule Ecto.Query.Planner do
...
     {cast_params, dump_params} = Enum.unzip(params)
-    key = if query_cache?, do: key, else: :nocache
+
+    key =
+      cond do
+        not query_cache? -> :nocache
+        counter != 0 -> {key, counter}
+        true -> key
+      end
+
```
```suggestion
    key = if query_cache?, do: [key | counter], else: :nocache
```

lib/ecto/query/planner.ex
```diff
@@ -136,7 +136,7 @@ defmodule Ecto.Query.Planner do
   def query(query, operation, cache, adapter, counter, query_cache?) do
     {query, params, key} = plan(query, operation, adapter)
     {cast_params, dump_params} = Enum.unzip(params)
-    key = if query_cache?, do: key, else: :nocache
+    key = if query_cache?, do: [key | counter], else: :nocache
```
```suggestion
    key = if query_cache?, do: {key, counter}, else: :nocache
```

lib/ecto/query/planner.ex
```diff
@@ -136,7 +136,7 @@ defmodule Ecto.Query.Planner do
   def query(query, operation, cache, adapter, counter, query_cache?) do
     {query, params, key} = plan(query, operation, adapter)
     {cast_params, dump_params} = Enum.unzip(params)
-    key = if query_cache?, do: key, else: :nocache
+    key = if query_cache?, do: {key, counter}, else: :nocache
```
josevalim: @zachdaniel I thought it would be better to always add the counter, to reduce the number of branches, but it seems that either using a list is wrong or we are messing with the key in an incompatible way. Can you please take a deeper look?

zachdaniel: Sorry I'm not quite following.

> it seems that either using a list is wrong or we are messing with the key in an incompatible way. 

WDYM by using a list?

Is this about the test failures? The seem like timeouts in the GH actions to me but maybe I'm missing something.

josevalim: I am not so sure. I have cloned ecto, pulled your branch, rebased master, and `ECTO_PATH=../ecto ECTO_ADAPTER=pg mix test` in `ecto_sql` fails for me.

And I decided to always return a tuple with the counter so we don't have hidden behaviour depending if counter is zero or not.

zachdaniel: I will do the same and investigate 👌

zachdaniel: @josevalim I could be missing something, but when I do the same I get all tests passing 🤔

josevalim: What did you run and which project? CI is also failing but it was not failing before my changes (and passed with yours).

zachdaniel: Sorry. I understand now. I didn't realize you had pushed to the branch 🤦‍♂️. The CI failures here seem like timeouts and my branch passed ecto_sql tests locally but I didn't pull. Will investigate 😄

zachdaniel: Okay so the problem is that the system uses the sentinal key `:nocache`, but that was getting wrapped in a tuple.

josevalim:

PR #4717, opened 2026-04-08
feat (Repo): Raise an error on query-like keyword opts to Repo functions

s3cur3:
Consider the following, written by a new-to-Elixir developer:

```elixir
Repo.delete_all(User, where: [account_id: user.account_id])
```

The developer mistakenly believed that the `opts` to the Repo function worked like the Ecto.Query DSL, when in fact these keyword options were entirely ignored (making this an unscoped deletion of all users).

With this change, Ecto will raise a runtime error when given `opts` keys that seem to indicate the user was trying to modify the queryable passed in to:

- Repo.all/2
- Repo.all_by/3
- Repo.delete_all/2
- Repo.exists?/2
- Repo.get/3
- Repo.get_by/3
- Repo.one/2
- Repo.stream/2

I've taken the simple approach to the validation (adding a `validate_query_opts!/2` call at the top of each function). I had considered bottlenecking this to do the validation only in `Ecto.Repo.Queryable.execute/4`, but doing so would have required passing `execute/4` the name of the function that called it (since right now it only receives the `operation`, one of `:all`, `:update_all`, or `:delete_all`, even for functions like `get` or `get_by`). I'm happy to refactor it to go through `execute` if that's a desirable tradeoff, though.

### DIFF
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -342,7 +342,6 @@ defmodule Ecto.Repo do
         :group_by,
         :distinct,
         :select,
-        :from,
         :update
       ]
 
--- a/lib/ecto/repo.ex
+++ b/lib/ecto/repo.ex
@@ -326,6 +326,38 @@ defmodule Ecto.Repo do
         |> Keyword.merge(opts)
       end
 
+      # Keyword list keys a user may try to pass to something like Repo.all/2, under
+      # the mistaken belief that the keyword arguments operate like the Ecto.Query DSL.
+      # We want to proactively raise an error on something like:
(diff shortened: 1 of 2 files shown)


lib/ecto/repo.ex
```diff
@@ -326,6 +326,38 @@ defmodule Ecto.Repo do
...
+      # did not intend.
+      @unsupported_query_opts [
+        :where,
+        :preload,
+        :order_by,
+        :limit,
+        :group_by,
+        :distinct,
+        :select,
+        :from,
```
```suggestion
```

josevalim:

PR #4721, opened 2026-04-26
Simplify UUIDv7 options

ryanwinchester:
Simplify the UUIDv7 options and make them more opinionated and easier to use.

* Too much flexibility means too easy to use in a weird way.
* There isn't really a good reason to use monotonic in combination with milliseconds.
* Nanoseconds wasn't _really_ nanoseconds, but sub-milliseconds based on nanoseconds, stepped across 12 bits

Since there hasn't been a release yet since the UUIDv7 options were added, it's a good time to do this.

### Before

Options:
  * `precision: :millisecond | :nanosecond`
  * `monotonic: true | false`
  
### After

Options:
  * `precision: :millisecond | :monotonic`

### DIFF
--- a/lib/ecto/application.ex
+++ b/lib/ecto/application.ex
@@ -3,7 +3,6 @@ defmodule Ecto.Application do
   use Application
 
   def start(_type, _args) do
-    :ok = :persistent_term.put({Ecto.UUID, :millisecond}, :atomics.new(1, signed: false))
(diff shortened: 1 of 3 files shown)


josevalim:
Love it! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

ryanwinchester:
Is there a timeline/approximate expected next release?

josevalim:
None but we can try to ship a RC within the next month.

PR #4722, opened 2026-04-27
Add `Ecto.Changeset.reorder_assoc`

### LINKED ISSUE #4713: unique_constraint doesn't work when using cast_assoc from parent
### Elixir version

Erlang/OTP 27 [erts-15.1.3] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit]  
Elixir 1.18.2 (compiled with Erlang/OTP 27)

### Database and Version

docker postgres:17

### Ecto Versions

ecto 3.13.5

postgrex 0.22.0

### Current behavior

Parent schema (workflow)
```
defmodule Devhub.Workflows.Schemas.Workflow do
  use Ecto.Schema

  import Ecto.Changeset

  alias Devhub.Workflows.Schemas.Step

  schema "workflows" do
    has_many :steps, Step, preload_order: [:order], on_replace: :delete, on_delete: :delete_all

    timestamps()
  end

  def changeset(schema \\ %__MODULE__{}, attrs) do
    schema
    |> cast_assoc(:steps,
      with: &Step.changeset/3,
      sort_param: :step_sort,
      drop_param: :step_drop
    )
  end
end
```

Child schema (step)
```
defmodule Devhub.Workflows.Schemas.Step do
  use Ecto.Schema

  import Ecto.Changeset

  schema "workflow_steps" do
    field :order, :integer
    field :name, :string

    timestamps()
  end

  def changeset(schema, attrs, order) do
    schema
    |> cast(attrs, [:workflow_id, :name])
    |> put_change(:order, order)
    |> unique_constraint([:workflow_id, :name],
      error_key: :name,
      name: "workflow_steps_workflow_id_name_unique",
      message: "step names must be unique"
    )
  end
end
```

Then if you update the workflow if crashes on Repo.update
```
workflow
|> Workflow.changeset(%{
  "steps" => %{
    "0" => %{"name" => "test"},
    "1" => %{"name" => "test"}
  }
})
|> Repo.update()
```

Error log
```
[error] Postgrex.Protocol (#PID<0.11502.0> ("db_conn_5")) disconnected: ** (Postgrex.Error) ERROR 23505 (unique_violation) duplicate key value violates unique constraint "workflow_steps_workflow_id_name_unique"

    table: workflow_steps
    constraint: workflow_steps_workflow_id_name_unique

Key (workflow_id, name)=(wf_01KJVBR7EYV4H22BN96F1QEYYY, test) already exists.
[error] GenServer #PID<0.16119.0> terminating
** (Postgrex.Error) ERROR 23505 (unique_violation) duplicate key value violates unique constraint "workflow_steps_workflow_id_name_unique"

    table: workflow_steps
    constraint: workflow_steps_workflow_id_name_unique

Key (workflow_id, name)=(wf_01KJVBR7EYV4H22BN96F1QEYYY, test) already exists.
    (db_connection 2.9.0) lib/db_connection.ex:1782: DBConnection.run_transaction/4
    (devhub 0.1.0) lib/devhub/workflows/actions/update_workflow.ex:15: Devhub.Workflows.Actions.UpdateWorkflow.update_workflow/2
    (devhub 0.1.0) lib/devhub_web/live/workflows/edit_workflow.ex:288: DevhubWeb.Live.Workflows.EditWorkflow.handle_event/3
    (phoenix_live_view 1.1.17) lib/phoenix_live_view/channel.ex:530: anonymous fn/3 in Phoenix.LiveView.Channel.view_handle_event/3
    (telemetry 1.4.1) /Users/michael/Code/Devhub/devhub-private/deps/telemetry/src/telemetry.erl:359: :telemetry.span/3
    (phoenix_live_view 1.1.17) lib/phoenix_live_view/channel.ex:260: Phoenix.LiveView.Channel.handle_info/2
    (stdlib 6.1.2) gen_server.erl:2345: :gen_server.try_handle_info/3
    (stdlib 6.1.2) gen_server.erl:2433: :gen_server.handle_msg/6
    (stdlib 6.1.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
```

### Expected behavior

Repo.update to return `{:error, %Ecto.Changeset{}}` with the appropriate error

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4713

I had a few different thoughts how to do this...some of the alternatives I thought about:

- Accepting a list of `:insert`, `:update`, `:delete` and letting the user choose which order they want. I thought it might get messy though validating that the list makes sense (i.e. if there are duplicates of certain atoms or things are missing) and also if potentially people will want to handle `:replace` specially
- Accepting a sorting function the user defines. I thought it might be better to start simpler and add this later if we get requests for more flexibility. There are some gotchas with translating `:replace` into repo actions that might be tricky to handle
- The name of the option value `:unique_safe` I thought about making it something like `:remove_first` or `:delete_first` or `:delete_upsert_insert`. The first two I discarded because I thought they were a bit confusing/not completely accurate. The last one I thought is a bit long and opens it up to having all the permutations as option names. I decided to try to find some kind of short user-friendly name that encompasses the specific purpose.

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1247,6 +1247,25 @@ defmodule Ecto.Changeset do
   The next two arguments are the changesets to be compared for sorting. You must return
   a `true` if the first changeset precedes or is in the same place as the second changeset
   and `false` otherwise.
+
+  ## Example
+
+      iex> sort_fn = refl, cs1, _cs2 ->
+      ...>   # ensure inserts and updates come first
+      ...>   case cs1.action do
+      ...>     :insert -> true
+      ...>     :update -> true
+      ...>     :replace when refl.on_replace == :nilify -> true
+      ...>     _ -> false
+      ...>   end
+      ...> end
+      iex> # assume `:comments` association has `on_replace: delete`
+      iex> cs = %Post{comments: [%Comment{id: 1, body: "hello"}, %Comment{id: 2, body: "bye"}]}
+      ...> |> change()
+      ...> |> put_assoc(:comments, [%Comment{id: 2, body: "hello"}, %Comment{id: 3, body: ""}])
+      ...> |> reorder_assoc(:comments, sort_fn)
+      iex> cs.changes.comments
+      [%Ecto.Changeset{data: %Comment{id: 2}}, %Ecto.Changeset{data: %Comment{id: 3}}, %Ecto.Changeset{data: %Comment{id: 1}}]
   """
   @spec cast_assoc(t, atom(), (term(), t, t -> boolean())) :: t
   def reorder_assoc(%Changeset{} = changeset, name, sort_fn)
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1219,19 +1219,32 @@ defmodule Ecto.Changeset do
 
   This function should be used when wanting to re-order the list of changes
   for an association with cardinality `:many` before writing to the database.
-  The 2-arity version of this function sorts the changes in a way that is safe
-  for use with unique constraints.
+  The 2-arity version sorts the changes in a way that is safe for use with
+  unique constraints.
 
   For example, if you have a unique constraint on the field `:name` and your list
-  of changes might introduce conflicts, you can use this function to sort changes
-  by deletes, then updates and then inserts. The `:on_replace` behavour will be
+  of changes might introduce conflicts, you can use this to sort changes by deletes
+  first, then updates and then inserts. The `:on_replace` behavour will be
   handled automatically.
 
   Using this function is preferable to relying on deferred constraints because the
   resulting error cannot be mapped back into the correct changeset and your transaction
   will simply raise.
 
-  See `reorder_assoc/3` if you would like to use your own custom sorting function.
+  Care must be taken when using this in conjunction with the `:sort_param` option
+  in `cast_assoc/3`. They both change the internal ordering of the association so you
+  must isolate the effects of this function to only the database operation.
+
+  See `reorder_assoc/3` if you would like to use a custom sorting function.
+
+  ## Example
+      iex> # assume `:comments` association has `on_replace: delete`
+      iex> cs = %Post{comments: [%Comment{id: 1, body: "hello"}, %Comment{id: 2, body: "bye"}]}
+      ...> |> change()
+      ...> |> put_assoc(:comments, [%Comment{id: 3, body: ""}, %Comment{id: 2, body: "hello"}])
+      ...> |> reorder_assoc(:comments, sort_fn)
+      iex> cs.changes.comments
+      [%Ecto.Changeset{data: %Comment{id: 1}}, %Ecto.Changeset{data: %Comment{id: 2}}, %Ecto.Changeset{data: %Comment{id: 3}}]
   """
   @spec cast_assoc(t, atom()) :: t
   def reorder_assoc(%Changeset{} = changeset, name) when is_atom(name) do
@@ -1241,11 +1254,11 @@ defmodule Ecto.Changeset do
   @doc """
   Reorders the changes for a given association using a custom sorting function.
 
-  This function behaves similarly to `reorder_assoc/2` except it allows the user
-  to define their own sorting function. The function must be of arity 3 where the
-  first argument is the reflection struct of the association, such as `Ecto.Association.Has`.
-  The next two arguments are the changesets to be compared for sorting. You must return
-  a `true` if the first changeset precedes or is in the same place as the second changeset
+  Behaviour is similar to `reorder_assoc/2` except it allows the user to define
+  their own sorting function. It must be of arity 3 where the first argument is
+  the reflection struct of the association, such as `Ecto.Association.Has`. The
+  next two arguments are the changesets to be compared. You must return `true`
+  if the first changeset precedes or is in the same place as the second changeset
   and `false` otherwise.
 
   ## Example
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -1215,6 +1215,85 @@ defmodule Ecto.Changeset do
   end
 
   @doc """
+  Reorders the changes for a given association.
+
+  This function should be used when wanting to re-order the list of changes
+  for an association with cardinality `:many` before writing to the database.
+  The 2-arity version of this function sorts the changes in a way that is safe
+  for use with unique constraints.
+
+  For example, if you have a unique constraint on the field `:name` and your list
+  of changes might introduce conflicts, you can use this function to sort changes
+  by deletes, then updates and then inserts. The `:on_replace` behavour will be
+  handled automatically.
+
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
As a side note I don't think `on_replace: :nilify` is working properly for `has_many`. When I tried to do a test for this change with that option it basically didn't register the update. And when I investigated the changes were empty. So I think we are not adding the foreign key change into the changes map.

If it's alright I can fix that in a follow up.

greg-rychlewski:
Actually nm I'm just dumb and didn't set up the test data correctly. Added the test

josevalim:
@greg-rychlewski could this be a changeset function instead? Something when casting associations or a pass we do on the changeset data after cast/put_assoc?

greg-rychlewski:
@josevalim What about something like this

```elixir
cast_assoc(changeset, :assocs, change_order: :unique_safe)
```

We would still do the sorting where it is in this PR but now the user has the flexibility to change the order instead of it being hard coded onto the schema. 

The reason I'm not sure about `or a pass we do on the changeset data after cast/put_assoc?` is because of the interactions with `:sort_param`. Since they are orthogonal concerns the two orders should be allowed to be different. But if they are different then they both can't be satisfied after calling `cast_assoc`

josevalim:
I was thinking something like: `cast_assoc(changeset, :assocs) |> reorder_assoc(:assocs)`. For `put`, we could make it emit the correct order if we want (I believe the order is not guaranteed). For cast, we actually don't to change the order, because the order is often the one coming from the UI/interface.

greg-rychlewski:
Ah ok. Just to make sure I understand correctly you are saying reorder_assoc always affects the database operation order right? So cast 
_assoc ignores it when you’re passing around the struct but when we are doing the database operations would respect it?

josevalim:
So there are two possible ways for the changesets for an assoc to be created:

1. cast_assoc, which often preserves UI order
2. put_assoc, which we can probably change the order to whatever we like

So my idea is that you can call reorder_assoc before you write to the database. We can provide a default function (no argument, but you can also give a custom one). Does that make sense?

greg-rychlewski:
Oh yes I got your idea now. Thanks!

greg-rychlewski:
@josevalim I'm having a bit of doubt and was curious what you thought. I started to implement it and then I thought that `changes` is already a public field so people could basically just use `Enum.sort` instead of `reorder_assoc`. We have the default function which we could set as the one in this PR but that kind of feels weird to make the default of a generic reorder function so specific to this one issue.

Do you think it would make sense to expose specific sort functions for users to apply themselves? I can't think of a good name right now but the idea would be something like `Ecto.Changeset.sort_delete_update_insert` which would take a list of changesets and order them the way this PR is. It seems to me the value is taking care of the complexities with `:replace` that users might forget about/not know how to do.

josevalim:
We could have a reorder_assoc that has a default function (which implements the semantics you mentioned above) but it could allow a custom function to be given for convenience. :)

greg-rychlewski:
@josevalim Let me know what you think of this when you have time :)

lib/ecto/changeset.ex
```diff
@@ -1214,6 +1214,117 @@ defmodule Ecto.Changeset do
...
+  Reorders the changes for a given association.
+
+  This function should be used when wanting to re-order the list of changes
+  for an association with cardinality `:many` before writing to the database.
+  The 2-arity version sorts the changes in a way that is safe for use with
+  unique constraints.
+
+  For example, if you have a unique constraint on the field `:name` and your list
+  of changes might introduce conflicts, you can use this to sort changes by deletes
+  first, then updates and then inserts. The `:on_replace` behavour will be
```
```suggestion
  first, then updates and then inserts. The `:on_replace` behaviour will be
```

lib/ecto/changeset.ex
```diff
@@ -1214,6 +1214,117 @@ defmodule Ecto.Changeset do
...
+  Behaviour is similar to `reorder_assoc/2` except it allows the user to define
+  their own sorting function. It must be of arity 3 where the first argument is
+  the reflection struct of the association, such as `Ecto.Association.Has`. The
+  next two arguments are the changesets to be compared. You must return `true`
+  if the first changeset precedes or is in the same place as the second changeset
+  and `false` otherwise.
+
+  ## Example
+
+      iex> sort_fn = refl, cs1, _cs2 ->
```
josevalim: The reflection is not public API, as far as I know, so we cannot "show" it for the custom function, unfortunately :(

lib/ecto/changeset.ex
```diff
@@ -1214,6 +1214,117 @@ defmodule Ecto.Changeset do
...
+  first, then updates and then inserts. The `:on_replace` behavour will be
+  handled automatically.
+
+  Using this function is preferable to relying on deferred constraints because the
+  resulting error cannot be mapped back into the correct changeset and your transaction
+  will simply raise.
+
+  Care must be taken when using this in conjunction with the `:sort_param` option
+  in `cast_assoc/3`. They both change the internal ordering of the association so you
+  must isolate the effects of this function to only the database operation.
```
josevalim, line 1236: Is this function meant to be used mostly with `cast_assoc`? Do we need to use it with put_assoc or put_assoc already returns the correct order?

lib/ecto/changeset.ex
greg-rychlewski: removed!

lib/ecto/changeset.ex
greg-rychlewski: Right now I left `put_assoc` as it was which basically keeps it in the order the user gave and adds the replaced changes to the front. 

We could call `reorder_assoc` at the end of it because I don't think there are any public guarantees on the order right now. I was just a bit hesitant to rock the boat.

lib/ecto/changeset.ex
```diff
@@ -1214,6 +1214,109 @@ defmodule Ecto.Changeset do
...
+  ## Example
+      iex> # assume `:comments` association has `on_replace: delete`
+      iex> cs = %Post{comments: [%Comment{id: 1, body: "hello"}, %Comment{id: 2, body: "bye"}]}
+      ...> |> change()
+      ...> |> put_assoc(:comments, [%Comment{id: 3, body: ""}, %Comment{id: 2, body: "hello"}])
+      ...> |> reorder_assoc(:comments, sort_fn)
+      iex> cs.changes.comments
+      [%Ecto.Changeset{data: %Comment{id: 1}}, %Ecto.Changeset{data: %Comment{id: 2}}, %Ecto.Changeset{data: %Comment{id: 3}}]
+  """
+  @spec cast_assoc(t, atom()) :: t
```
josevalim: Wrong specs :)

lib/ecto/changeset.ex
josevalim: replace in the front look sane enough, because, if the replace is a delete, then it is in the front and if it is a nilify it matches updates, so we don't need to change `put_assoc`.

josevalim (approved):
Beautiful job! Fix the specs and ship it :)

PR #4726, opened 2026-05-13
Add on_join_through_conflict option for ManyToMany Associations

### LINKED ISSUE #4725: Upsert, `many_to_many` and `cast_assoc`: insert in join table generated even when the upsert turns out to be an update
### Elixir version

1.19.5

### Database and Version

Postgres 16.1

### Ecto Versions

3.13.5

postgrex

### Current behavior

Let's assume I have 3 tables `country`, `city` and `country_cities` (the join table).

I insert and update the countries and there associated cities using `cast_assoc` on a `many_to_many` association from the `Country` schema using an **upsert**.

For example this is the params of the changeset:

```elixir
%{
  "cities" => %{
    "0" => %{
      "_persistent_id" => "0",
      "id" => "4215",
      "name" => "Lyon modified"
    },
    "1" => %{
      "_persistent_id" => "1",
      "id" => "2456",
      "name" => "Grenoble"
    }
  },
  "cities_drop" => [""],
  "cities_sort" => ["0", "1"],
  "name" => "France",
}
```

We then apply these params to the existing `Country` and get a changeset passed to `Repo.insert/...`.

It doesn't work and the transaction is rolled back because a insert **in the join table** is generated, even though the `Country` exists and the association (in this case `cities`) is loaded:

```
[debug] QUERY OK source="country" db=0.3ms
INSERT INTO "country" ("id","name","updated_at","inserted_at") VALUES ($1,$2,$3,$4) ON CONFLICT ("id") DO UPDATE SET "updated_at" = EXCLUDED."updated_at","inserted_at" = EXCLUDED."inserted_at","traffic" = EXCLUDED."traffic","name" = EXCLUDED."name" RETURNING "updated_at","inserted_at","name","id" [1267, "France",  ~U[2026-05-12 17:54:37.577948Z], ~U[2026-04-28 18:07:49.000000Z]]

[debug] QUERY OK source="city" db=0.3ms
UPDATE "city" SET "name" = $1, "updated_at" = $2 WHERE "id" = $3 ["Lyon modified", ~U[2026-05-12 17:54:37.579648Z], 4215]

[debug] QUERY ERROR source="country_cities" db=0.3ms
INSERT INTO "country_cities" ("country_id","city_id","inserted_at","updated_at") VALUES ($1,$2,$3,$4) [1267, 4215, ~N[2026-05-12 17:54:37], ~N[2026-05-12 17:54:37]]
```

As you can see the last statement is an insert in the join table.

It is generated because:
1. in https://github.com/elixir-ecto/ecto/blob/270acebe3d76779dcb0e97c40b0271b2103cbfc9/lib/ecto/repo/schema.ex#L453 the action is set to `:insert`
2. then in https://github.com/elixir-ecto/ecto/blob/270acebe3d76779dcb0e97c40b0271b2103cbfc9/lib/ecto/association.ex#L1595 Ecto determines it's a new, fresh insert and therefore a association record has to be added in the join table

### Expected behavior

First it remains to be determined if it is a bug.

In my opinion it is, because the changeset contains all the required information (preloaded assoc, ids in the changeset generated by cast_assoc) to determine whether we should insert in the join table or not. Moreover, when using upserts, we except it to be an insert or an update.

I think this bug is a reminiscence of Ecto before upserts were implemented. In this case it made sense to assume that a `Repo.insert/...` could only be an insert, but this no longer the case since. Maybe just removing [this line](https://github.com/elixir-ecto/ecto/blob/270acebe3d76779dcb0e97c40b0271b2103cbfc9/lib/ecto/association.ex#L1595) would make sense, because as far as I know there is no generic way to know if the result of an upsert is a new record or an existing one.

PS: the example is trivial and doesn't represent what I'm really doing, and I don't really have the option to edit cities separately. I'm building a collaborative subsystem based on Ecto changesets. In a nutshell: changes are proposed by the user, then params are stored in the DB and later applied to the current stored object.

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4725

A couple of callouts:

- only allowing `:nothing` or `:raise` because the stuff the Ecto does is only inserts the two foreign keys and doesn't use a schema. So I don't think replace makes sense here
- didn't add for insert_all because it doesn't look to me like it automatically handles associations
- this covers all the associations because i can't really think of a valid reason to do it for some but not others. though if someone ever makes a convincing case we could potentially accept a keyword list

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -1163,7 +1163,7 @@ defmodule Ecto.Repo.Schema do
   defp assoc_opts([], _opts), do: []
 
   defp assoc_opts(_assocs, opts) do
-    Keyword.take(opts, [:timeout, :log, :telemetry_event, :prefix, :allow_stale])
+    Keyword.take(opts, [:timeout, :log, :telemetry_event, :prefix, :allow_stale, :on_join_table_conflict])
   end
 
   defp process_parents(changeset, user_changeset, assocs, reset_assocs, adapter, opts) do
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -1163,7 +1163,7 @@ defmodule Ecto.Repo.Schema do
   defp assoc_opts([], _opts), do: []
 
   defp assoc_opts(_assocs, opts) do
-    Keyword.take(opts, [:timeout, :log, :telemetry_event, :prefix, :allow_stale, :on_join_table_conflict])
+    Keyword.take(opts, [:timeout, :log, :telemetry_event, :prefix, :allow_stale])
   end
 
   defp process_parents(changeset, user_changeset, assocs, reset_assocs, adapter, opts) do
--- a/lib/ecto/association.ex
+++ b/lib/ecto/association.ex
@@ -1300,6 +1300,7 @@ defmodule Ecto.Association.ManyToMany do
   @behaviour Ecto.Association
   @on_delete_opts [:nothing, :delete_all]
   @on_replace_opts [:raise, :mark_as_invalid, :delete]
+  @on_join_through_conflict_opts [:raise, :nothing]
 
   defstruct [
     :field,
@@ -1312,6 +1313,7 @@ defmodule Ecto.Association.ManyToMany do
     :join_keys,
     :join_through,
     :on_cast,
+    :on_join_through_conflict,
     where: [],
     join_where: [],
     defaults: [],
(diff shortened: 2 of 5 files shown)


lib/ecto/repo/schema.ex
```diff
@@ -1163,7 +1163,7 @@ defmodule Ecto.Repo.Schema do
   defp assoc_opts([], _opts), do: []
 
   defp assoc_opts(_assocs, opts) do
-    Keyword.take(opts, [:timeout, :log, :telemetry_event, :prefix, :allow_stale])
+    Keyword.take(opts, [:timeout, :log, :telemetry_event, :prefix, :allow_stale, :on_join_table_conflict])
```
josevalim: Instead of passing it through here, can we store this value in the ManyToMany reflection struct instead? This way we validate it when we create the association!

josevalim: So basically we would do this:

    many_to_many :comments, Comment, on_join_table_conflict: :nothing

josevalim: Or whatever is the best name for the option!

greg-rychlewski: done!

josevalim (approved):

PR #4727, opened 2026-05-15
Redact repo config from SASL progress reports (#4718)

### LINKED ISSUE #4718: Ecto.Repo.Supervisor.start_child leaks credentials
### Elixir version

1.19.5

### Database and Version

PostgreSQL 16.4

### Ecto Versions

3.13.5

postgrex 0.22.0

### Current behavior

Ecto.Repo.Supervisor.start_child leaks credentials.

### Expected behavior

I believe it would be important to always wrap configuration in a structure like what was done for https://github.com/elixir-ecto/db_connection/issues/340.

dl-alexandre:
Closes #4718.

`Ecto.Repo.Supervisor.wrap_child_spec/2` stores the adapter's `start` MFA in the child spec's `start:` field. That MFA's args carry the full repo config (`:username`, `:password`, `:url`, …) by value. When `:logger` is set with `handle_otp_reports: true` and `handle_sasl_reports: true` (Schultzer's setup in the issue), the supervisor's `progress` report inspects the child spec verbatim, and the password lands in stdout / log sinks.

The fix mirrors elixir-ecto/db_connection#340: wrap the inner MFA in a small `Ecto.Repo.Supervisor.SensitiveData` struct that derives `Inspect` with no visible fields, then unwrap it on the receiving side of the supervisor boundary in `start_child/4`. The struct only exists while the args sit in the child-spec registry — exactly the window in which SASL formats them. After `apply/3` runs, adapter behavior is unchanged.

The existing 3-tuple `start_child/4` clause is preserved for back-compat with any external caller. The only internal caller (`wrap_child_spec/2`) goes through the wrapped clause.

## Reproducer

```elixir
# Run from a project with ecto + ecto_sql + postgrex.
:logger.add_handlers(:sasl)
Application.put_env(:logger, :handle_otp_reports, true)
Application.put_env(:logger, :handle_sasl_reports, true)

defmodule LeakRepo do
  use Ecto.Repo, otp_app: :leak, adapter: Ecto.Adapters.Postgres
end

Application.put_env(:leak, LeakRepo,
  username: "user", password: "PLEASE-DO-NOT-LEAK-ME",
  hostname: "127.0.0.1", database: "nope",
  pool_size: 1, backoff_type: :stop
)

ExUnit.CaptureLog.capture_log([level: :info], fn ->
  {:ok, _} = LeakRepo.start_link()
  Process.sleep(500)
end)
|> tap(&IO.puts(if &1 =~ "PLEASE-DO-NOT-LEAK-ME", do: "LEAK", else: "REDACTED"))
```

Before this PR: `LEAK` (password visible inside `=PROGRESS REPORT====`).
With this PR: `REDACTED` (the MFA is rendered as `#Ecto.Repo.Supervisor.SensitiveData<...>`).

## Tests

Two unit tests added under `test/ecto/repo/supervisor_test.exs`:

1. `inspect/1` on the wrapper renders no credential fields.
2. `start_child/4` correctly unwraps the wrapper and delegates to the existing code path.

`mix test test/ecto/repo/supervisor_test.exs` — 21 tests, 0 failures.

## Open questions

- The migrator and the `[:ecto, :repo, :init]` telemetry event still pass raw opts. They're not SASL/log leaks per se (telemetry handlers are in-process), but worth a follow-up if you'd like opts treated uniformly.

### DIFF
--- a/test/ecto/repo/supervisor_test.exs
+++ b/test/ecto/repo/supervisor_test.exs
@@ -223,31 +223,4 @@ defmodule Ecto.Repo.SupervisorTest do
       end
     end
   end
-
-  describe "child spec credential redaction (#4718)" do
-    test "the wrapper hides credentials from inspect/1" do
-      mfa =
-        {SomePool, :start_link,
-         [[username: "u", password: "s3cret-do-not-leak", url: "ecto://u:s3cret@h/d"]]}
-
-      rendered = inspect(%Ecto.Repo.Supervisor.SensitiveData{data: mfa})
-
-      refute rendered =~ "s3cret"
-      refute rendered =~ "password"
-      refute rendered =~ "url"
-      assert rendered =~ "SensitiveData"
-    end
-
-    test "start_child/4 unwraps the SensitiveData wrapper" do
-      # The wrapper clause must delegate to the same code path as the raw
-      # MFA clause. Verify with a no-op MFA that returns `:ignore`.
-      defmodule WrapperProbe do
-        def go, do: :ignore
-      end
-
-      wrapped = %Ecto.Repo.Supervisor.SensitiveData{data: {WrapperProbe, :go, []}}
-
-      assert Ecto.Repo.Supervisor.start_child(wrapped, :nope, :adapter, %{}) == :ignore
-    end
-  end
 end
--- a/test/ecto/repo/supervisor_test.exs
+++ b/test/ecto/repo/supervisor_test.exs
@@ -223,4 +223,31 @@ defmodule Ecto.Repo.SupervisorTest do
       end
     end
   end
+
+  describe "child spec credential redaction (#4718)" do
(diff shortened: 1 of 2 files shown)


test/ecto/repo/supervisor_test.exs
```diff
@@ -223,4 +223,31 @@ defmodule Ecto.Repo.SupervisorTest do
...
+      # MFA clause. Verify with a no-op MFA that returns `:ignore`.
+      defmodule WrapperProbe do
+        def go, do: :ignore
+      end
+
+      wrapped = %Ecto.Repo.Supervisor.SensitiveData{data: {WrapperProbe, :go, []}}
+
+      assert Ecto.Repo.Supervisor.start_child(wrapped, :nope, :adapter, %{}) == :ignore
+    end
+  end
```
josevalim:
```suggestion
```

josevalim:

PR #4729, opened 2026-05-22
Ecto.Schema: Allow embedded_schema to have non-virtual :any fields

wojtekmach:


lib/ecto/schema.ex
```diff
@@ -1997,7 +1997,10 @@ defmodule Ecto.Schema do
     # better to raise unknown type first than unsupported option.
     type = check_field_type!(mod, name, type, opts)
 
-    if type == :any && !opts[:virtual] do
+    fields = Module.get_attribute(mod, :ecto_struct_fields)
+    schema_with_source? = match?(%Ecto.Schema.Metadata{}, fields[:__meta__])
+
+    if schema_with_source? && type == :any && !opts[:virtual] do
```
josevalim: What if we store `@ecto_source` and then we check if it is nil? Loading all fields from attributes can get expensive.

We should also move the expensive check last:

```suggestion
    if type == :any && !opts[:virtual] &&
         match?(%Ecto.Schema.Metadata{}, Module.get_attribute(mod, :ecto_struct_fields)[:__meta__]) do
```

josevalim (approved):

PR #4730, opened 2026-05-27
Expand macros in window frames

dannote:
Allow `:frame` to expand macros before validating that the result is a fragment.

Today this works:

```elixir
windows: [
  recent: [
    order_by: result.date,
    frame: fragment(
      "RANGE BETWEEN INTERVAL 10 DAYS PRECEDING AND INTERVAL 10 DAYS FOLLOWING EXCLUDE CURRENT ROW"
    )
  ]
]
```

but an equivalent helper is rejected before expansion:

```elixir
defmacro range_between(start_boundary, end_boundary, opts \\ []) do
  exclude = Keyword.get(opts, :exclude, :no_others)

  quote do
    fragment(
      unquote("RANGE BETWEEN #{start_boundary} AND #{end_boundary} EXCLUDE #{exclude}")
    )
  end
end

windows: [
  recent: [
    order_by: result.date,
    frame:
      range_between(
        "INTERVAL 10 DAYS PRECEDING",
        "INTERVAL 10 DAYS FOLLOWING",
        exclude: "CURRENT ROW"
      )
  ]
]
```

I ran into this while working on [QuackDB](https://github.com/elixir-vibe/quackdb), a new DuckDB adapter for Elixir. DuckDB supports richer frame clauses such as `ROWS`, `RANGE`, `GROUPS`, and `EXCLUDE`, and QuackDB could expose reusable helpers for those clauses instead of requiring users to write frame fragments inline.

The validation stays the same: after expansion, the frame still has to be a fragment. Non-fragment expansions continue to raise the existing error.

### DIFF
--- a/lib/ecto/query/builder/windows.ex
+++ b/lib/ecto/query/builder/windows.ex
@@ -72,8 +72,21 @@ defmodule Ecto.Query.Builder.Windows do
   defp escape_frame({:fragment, _, _} = fragment, params_acc, vars, env) do
     Builder.escape(fragment, :any, params_acc, vars, env)
   end
-  defp escape_frame(other, _, _, _) do
-    Builder.error!("expected a dynamic or fragment in `:frame`, got: `#{inspect other}`")
+
+  defp escape_frame(other, params_acc, vars, env) do
+    macro_env =
+      case env do
+        {env, _} -> env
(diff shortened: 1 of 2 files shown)


josevalim:

Issue #4719, opened 2026-04-19
Add mix ecto.query


josevalim:
This is inspired by `rails query`:

* Add a `read_only: true` option to Repo.transaction (databases that do not support it should raise)
* Add `mix ecto.query "from(p in Post, where: ...)"` which will:
  * Load the current dot iex if any
  * import Ecto.Query
  * Start a read-only transaction
  * Eval the code
  * And pass it to Repo.all using the default repo
* Print the results using inspect(limit: :infinity) but no more than 100 entries (controlled by `--limit` flag)

Question: do we print the high-level schema data (if yes, then we should remove all association fields) or the {MyXQL,Postgrex,...}.Result  struct? Rails prints the result struct. If we go with the result, then this belongs in `ecto_sql` (which could be beneficial as we could support a `--sql` flag too).

cdvx:
I'd like to take this on.

I'm thinking of starting in `ecto` with `mix ecto.query`, returning the high-level schema data. For printing, I'd clean up schema structs by dropping `__meta__` and association fields before inspecting with `limit: :infinity`.
For `read_only: true`, I'd pass the option through from `ecto`, then handle the adapter-specific behavior in `ecto_sql`. Adapters that can't support read-only transactions would raise there.
`--sql` would be handled as a follow-up in `ecto_sql`, since that depends on SQL capable adapters.

Does that split sound right?

v0idpwn:
I think there is value in doing it via schema due to things like `:redact` and custom types, even if I'd prefer to write SQL a lot of the times.

v0idpwn:
If printing the result, I think there could be adapter specific logic so we get something nicer than %Postgrex.Result{} (e.g.: a list of maps is usually way more readable)

cdvx:
That makes sense. I'll start with the schema/ecto path then, given it preserves schema level behavior like redacted fields and the like.
For raw result output / `--sql`, I'll handle that in a separate `ecto_sql` follow-up where adapter results can be formatted into something more readable like maps.

PR #4731, opened 2026-05-29
Add mix ecto.query task

cdvx:
Refs #4719
- adds mix ecto.query to evaluate an Ecto query expression, run it through the selected/default repo inside a read-only transaction, and print schema level data.

### DIFF
--- a/lib/ecto/adapter/transaction.ex
+++ b/lib/ecto/adapter/transaction.ex
@@ -11,6 +11,9 @@ defmodule Ecto.Adapter.Transaction do
   Returns `{:ok, value}` if the transaction was successful where `value`
   is the value returned by the function or `{:error, value}` if the transaction
   was rolled back where `value` is the value given to `rollback/1`.
+
+  If the `:read_only` option is true, the adapter must run the transaction
+  in read-only mode or raise if read-only transactions are not supported.
   """
   @callback transaction(adapter_meta, options :: Keyword.t(), function :: fun) ::
               {:ok, any} | {:error, any}
--- a/lib/mix/tasks/ecto.query.ex
+++ b/lib/mix/tasks/ecto.query.ex
@@ -1,0 +1,164 @@
+defmodule Mix.Tasks.Ecto.Query do
+  use Mix.Task
+  import Mix.Ecto
+
+  @shortdoc "Runs a query against the repository"
+
+  @switches [
+    limit: :integer,
+    repo: [:string, :keep],
+    no_compile: :boolean,
+    no_deps_check: :boolean
+  ]
+
+  @aliases [
+    r: :repo
+  ]
+
+  @moduledoc """
+  Runs the given query against the repository.
+
+  The query is evaluated as Elixir code after loading the current
+  `.iex.exs` file, if one exists, and importing `Ecto.Query`.
+
+  ## Examples
+
+      $ mix ecto.query "from p in Post, where: p.published"
+      $ mix ecto.query -r Custom.Repo "from p in Post, limit: 10"
+
+  ## Command line options
+
+    * `-r`, `--repo` - the repo to query
+    * `--limit` - limits the number of printed entries. Defaults to 100.
+
+  """
+
+  @default_limit 100
+
+  @impl true
+  def run(args) do
+    repos = parse_repo(args)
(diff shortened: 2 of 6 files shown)


lib/mix/tasks/ecto.query.ex
```diff
@@ -0,0 +1,164 @@
...
+    do: Map.new(entry, fn {key, value} -> {key, clean_entry(value)} end)
+
+  defp clean_entry(entry), do: entry
+end
+
+defmodule Mix.Tasks.Ecto.Query.Schema do
+  @moduledoc false
+
+  defstruct [:schema, :fields]
+end
```
v0idpwn, line 151: Why does this struct need to exist? I think it would probably be nicer to rely on the existing inspect implementation for the structs...

lib/mix/tasks/ecto.query.ex
```diff
@@ -0,0 +1,164 @@
...
+  end
+
+  defp clean_entries(entries) do
+    Enum.map(entries, &clean_entry/1)
+  end
+
+  defp clean_entry(%{__struct__: schema} = struct) do
+    if function_exported?(schema, :__schema__, 1) do
+      drop_fields = [
+        :__meta__ | schema.__schema__(:associations) ++ schema.__schema__(:redact_fields)
```
v0idpwn, line 123: I wonder if associations should only be deleted if they aren't preloaded

lib/ecto/adapter/transaction.ex
```diff
@@ -11,6 +11,9 @@ defmodule Ecto.Adapter.Transaction do
   Returns `{:ok, value}` if the transaction was successful where `value`
   is the value returned by the function or `{:error, value}` if the transaction
   was rolled back where `value` is the value given to `rollback/1`.
+
+  If the `:read_only` option is true, the adapter must run the transaction
+  in read-only mode or raise if read-only transactions are not supported.
```
v0idpwn, line 16: I like this option conceptually, but the need to add it makes me wonder if the whole thing should be in `ecto_sql` 🤔

lib/mix/tasks/ecto.query.ex
cdvx: This is only there to keep schema shaped output after dropping `__meta__`, redacted fields, and association fields. Without it, the cleaned value becomes a map and loses the schema name.
I found the `SchemaName{...}` shape easier to read when testing, but I can simplify to maps or regular inspect output if preferred.

lib/mix/tasks/ecto.query.ex
cdvx: That's a good point. I interpreted the issue as dropping all association fields, but keeping preloaded associations is more useful. I’ll change it to keep preloaded associations.

lib/ecto/adapter/transaction.ex
cdvx: Yeah, I had the same concern. I started here because the command evaluates an Ecto query and calls `Repo.all`, so the schema-level output felt like the more natural first pass.
The `read_only: true` part is the bit that crosses into adapter behavior. My thinking was to document and pass the option through in Ecto, then handle the actual SQL adapter support in an `ecto_sql` follow-up, together with the `--sql` option.

lib/mix/tasks/ecto.query.ex
josevalim: Inspect has a default_inspect_fun. I wonder if we can override it during the execution of this task, so we don't have to traverse everything and create a fake `Mix.Tasks.Ecto.Query.Schema`. Basically we intercept all structs with a `__meta__` field pointing to Ecto, otherwise we render them as is.

lib/ecto/adapter/transaction.ex
josevalim: Agreed on moving it to `ecto_sql`!

cdvx: Thanks, in that case I'll close this PR and handle this in `ecto_sql`

lib/mix/tasks/ecto.query.ex
cdvx: I'll look into overriding default_inspect_fun for this

cdvx:
to be handled in `ecto_sql`

PR #4732, opened 2026-06-03
Add Ecto.Query.label/2 query expression for tagging queries with an S…

alesasnouski:
Hi,
I added a label query expression (and label: keyword) that prepends a /* ... */ comment to the generated statement, right before the SELECT/UPDATE/DELETE keyword, like:
```
from(p in Post, label: "get_post_titles_q", select: p.title)
```
Tag (or label) makes it easy to identify a query's call site in database logs, slow-query logs, etc. The comment is rendered leading (not trailing) so it survives truncation of long statements in those logs.

I went with label rather than comment because :comment is already a Postgrex connection option (it appends a trailing comment to the statement), so reusing the name would be confusing.

Tested against Postgres, MySQL and SQL Server.


lib/ecto/repo.ex
```diff
@@ -1732,6 +1746,13 @@ defmodule Ecto.Repo do
     * `:placeholders` - A map with placeholders. This feature is not supported
       by all databases. See the ["Placeholders" section](#c:insert_all/3-placeholders) for more information.
 
+    * `:label` - A string to tag the generated `INSERT` with a leading SQL
+      comment, such as `/* import_users */ INSERT ...`, to identify the statement
+      in database logs and monitoring tools. The string is embedded verbatim into
+      the SQL and therefore cannot contain `/*`, `*/`, or null bytes. Prefer a
+      stable identifier per call site, as the label becomes part of the query
+      cache key.
```
josevalim: I don't think we should support it as a repo option and query option, let's go with one or the other.

lib/ecto/query/builder/label.ex
```diff
@@ -0,0 +1,82 @@
+import Kernel, except: [apply: 2]
+
+defmodule Ecto.Query.Builder.Label do
+  @moduledoc false
+
+  alias Ecto.Query.Builder
+
+  @forbidden ["*/", "/*", <<0>>]
```
josevalim: We should let the adapters do the escaping in case some adapters use a different style?

lib/ecto/query.ex
```diff
@@ -2537,6 +2540,40 @@ defmodule Ecto.Query do
...
+  Adds the given text to the generated statement as a leading SQL comment,
+  immediately before the statement keyword:
+
+      /* get_username_q */ SELECT ...
+
+  This is useful to tag and identify queries in database logs and monitoring
+  tools. The label is rendered *leading* (rather than trailing) so it survives
+  truncation of long statements in logs.
+
+  Because the label becomes part of the generated SQL, it is also part of the
```
greg-rychlewski: maybe this should be in an admonition block. there have been a few times the past couple of years where we had to add workarounds for people whose cache's blew up without them expecting it.

maybe we should also highlight that if they want to use high  cardinality values they can disable the cache for that query. i never used sql comments for logging though so i'm not too sure if this is some kind of extremely bad practice that we shouldn't be encouraging at all.

greg-rychlewski: @josevalim If comments are always prepended do we need to make them part of the cache? Can we just leave them out and then prepend them to the cached sql in the adapter.

lib/ecto/repo.ex
alesasnouski: It reads more like a query option than a repo option:
```
from(s in Site, label: "foobar") |> Repo.all()   vs   from(s in Site) |> Repo.all(label: "foobar")
```
But Repo.insert/update/delete get a struct/changeset and insert_all gets a list of maps, which is why we ended up with the mixed version.

We could switch to the repo option everywhere (same as :prefix). What are your thoughts?

lib/ecto/query.ex
josevalim: They need to be part of the cache, yes!

lib/ecto/repo.ex
josevalim: We already have :comment on Postgrex, so if that's your goal, we can add `:comment` to MyXQL and TDS, and there is no need to change Ecto? :)

alesasnouski: The truth is that the comment should be at the beginning of the query. By default in Postgres (for instance) track_activity_query_size == 1024 bytes, long queries are truncated (in pg_stat_statements, pg_stat_activity and, as a result, in GCP Query Insights console). Changing track_activity_query_size requires a DB server restart.
GCP allows you to change query length up to 4500 bytes, for instance (+ server restart needed).
image
Also adding a comment at the end (in the current implementation) disables the query cache (prep statements).

josevalim: It is at the end because that’s what sqlcommenter uses. Does it also work at the beginning? Another option is to have a flag to control it.

In any case, the disabling of query caching is not related to the position. If we implement it in Ecto and make it part of query cache, then it can work regardless. One option is to support precomment and postcomment options, and deprecate the Postgres comment one.

josevalim: Ping @dkuku.

dkuku: I added it at the end because that's what sql commenter uses, but I also think that most libraries that parse it does not care where it is. Beginning of the query is definetly better for visibility.

dkuku: I did similar change some time ago https://github.com/elixir-ecto/ecto/compare/master...dkuku:ecto:comments_in_query

alesasnouski: With sqlcommenter, comments are dynamic, which is why the prepared-statement cache must be disabled.
So we have three variables: the comment, its position (head/tail), and the cache (on/off) ...

josevalim: > With sqlcommenter, comments are dynamic, which is why the prepared-statement cache must be disabled.

They are dynamic in some sense but, once the code is compiled, you have a limited set of callers to Ecto.Repo. So it could go through the cache. Unless, of course, you start adding dynamic content to it (such as `user_id`). Then the amount of combinations are just too high and I think it makes sense to disable cache.

One option is for us to add two macros: pre_comment and post_comment, that comment before and after (although we store it in a single field called `comments: [{:pre, string}, {:post, string}]`. And yeah, I'd force them to be static (so no repo option).

alesasnouski: How about adding a :cache option ?

dkuku: I did work on this some time ago, and one idea I had was to allow caching the query when the comment is:
- an atom e.g. `:get_all_users`
- keyword list where values are only atoms such as `[app: :my_app, context: :rpc]`
- compile-time string, e.g. `label(query, "my_app")` which the macro system is aware of when used as a Query macro.

This way, by design, you have a limited set of tags for a single query. If you are generating atoms dynamically, you probably have bigger problems in your application, and standard code inspection tools would catch it.

dkuku:
I'm not sure how existing tools like datadog will parse this:
```
/* tag */ select 1 /* sql=commenter */
```
I think it may be undefined behaviour, what about:
```
-- tag
select 1 /* sql=commenter */
```

alesasnouski:
Do you propose to make it exclusive: prefix or suffix, not both?
I think it's ok to add such label like "sql=..."

dkuku:
> Do you propose to make it exclusive: prefix or suffix, not both? 

I'm just not sure if the obsevability tools will continue to parse a query when there is more than 1 comment wrapped in /**/. Otherwise when someone adds a label and a comment at the end it may stop working.

greg-rychlewski:
I’m ok to go with the consensus but if we’re are doing pre_comment and post_comment I’m still a bit unsure why we would cache them.

 If they get passed through as options to the drivers then I think we could just pass [pre_comment, sql, post_comment] to the socket since it accepts iodata? Then we don’t have to worry about blowing up the cache. It’s possible I’m misremembering how things work though

greg-rychlewski:
Oh I’m dumb for some reason I thought you could inject comments as part of the execute phase but it’s only prepare phase. Gg :(

lib/ecto/repo.ex
alesasnouski: We can omit the repo option, but then we can't tag schema statements, like
```Repo.insert(changeset, comments: [pre: "create_user"])```

Also, the `post_comment` macro is static, so it can't replace Postgrex's native `:comment` (which is dynamic, for sqlcommenter). A dynamic trailing comment can only come from the repo option comments: [post: ...].
So deprecating the Postgrex option actually requires the repo option.

One caveat: unlike Postgrex (which auto-disables the prepared-statement cache when a `comment` is present), our option doesn't - so on query operations the user must pass `query_cache: false` for dynamic values.

PR #4733, opened 2026-06-09
Add Repo.select_all 

dkuku:
```
{count, result} = Repo.select_all(query)
```

I proposed this on the Google Group but didn't receive any replies. This has been sitting on my disk, so I'll try to at least get it upstream.

I think this is a missing piece — we have `insert_all`, `delete_all`, and `update_all`, all of which return the count, but there's no equivalent for select. Currently, `Repo.all` has the count available internally but calls `elem(1)` on the tuple, so when we actually need the count we have two options:

1. Calling `Enum.count/1` on the results.
2. Running a separate `Repo.aggregate`.

I think this fits nicely inside the current api - you have either the long version `select_all` that gives you the full result with count, or the short version `all` that gives you only the result. 
The count is useful with pagination
Examples:
 https://github.com/duffelhq/paginator/blob/main/lib/paginator.ex#L329 uses Enum.count
 https://github.com/mojotech/scrivener_ecto/blob/main/lib/scrivener/paginater/ecto/query.ex#L71 uses sql count
 https://github.com/ash-project/ash/blob/main/lib/ash/actions/read/relationships.ex#L1617 uses Enum.split(limit) and checks if there is anything left

### DIFF
--- a/integration_test/cases/repo.exs
+++ b/integration_test/cases/repo.exs
@@ -98,6 +98,60 @@ defmodule Ecto.Integration.RepoTest do
     assert TestRepo.all_by(Post, title: "b") |> Enum.sort() == [post3]
   end
 
+  test "select_all" do
+    post1 = TestRepo.insert!(%Post{title: "a"})
+    post2 = TestRepo.insert!(%Post{title: "b"})
+    post3 = TestRepo.insert!(%Post{title: "c"})
+
+    # Test with full schema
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you but unfortunately I don't believe it is worth enlarging the Ecto API for this functionality. Doing a `length` again is not ideal but unlikely to be problematic compared to all of the cost of going over the wire and instantiating all of the data returned by a query.

PR #4734, opened 2026-06-14
Add ability to configure the behavior of writing to a non-writable field

green-david:
This PR adds the ability to change the default behavior of silently ignoring changes to non-writable fields as discussed [on the mailing list](https://groups.google.com/g/elixir-ecto/c/kloQ5uJ-Xss).

Attempting to write to a non-writable field can lead to code which _appears_ to update the field but actually doesn't, leading to confusing behavior or hidden bugs. By allowing the user to specify that a warning should be logged or an exception raised, these errant writes can be detected and remedied.


lib/ecto/repo/schema.ex
```diff
@@ -624,6 +625,31 @@ defmodule Ecto.Repo.Schema do
...
+    message = "attempted to write to non-writable field #{inspect(name)} during #{action}"
+
+    case on_writable_violation do
+      :raise ->
+        raise ArgumentError, message
+      :warn ->
+        Logger.warning(message)
+      _ ->
+        :ok
+    end
```
green-david: The message here likely needs to be more robust and contain additional information, open to feedback. This is just a simple message for now to illustrate the functionality.

lib/ecto/schema.ex
```diff
@@ -700,6 +701,13 @@ defmodule Ecto.Schema do
       be further modified, even in an upsert. If set to `:never`, the field becomes
       read only. Defaults to `:always`.
 
+    * `:on_writable_violation` - Defines what action to take when performing an insert or update
+      attempts to modify a field that should not be modified according to it's `:writable` value.
+      Must be one of `:nothing`, `:warn`, or `:raise`. If set to `:nothing`, the modification is
+      silently ignored. If set to `:warn`, the modification is ignored and a warning is logged. If set
+      to `:raise`, an exception is raised and the operation is aborted. If `:writable` is set to `:always`,
+      `:on_writable_violation` must be set to `:nothing`. Defaults to `:nothing`.
```
green-david, line 710: I added this as "the path of least resistance" to illustrate the functionality, but I am not confident this is the desired way forward as to the public facing API of how to configure this behavior.

Some other options:

`writable: [when: :never, on_violation: :raise]`
`writable: {:never, :raise}`
`writable: [:never, :raise]`

Open to suggestions :)

josevalim:
Looking at the code, I like the on_writable_violation approach. However, I wonder if the implementation should be `module.__schema__(:on_writable_violation)`. It should be much simpler and we only need to invoke it if there is a violation indeed.

josevalim:
WDYT?

green-david:
> Looking at the code, I like the on_writable_violation approach. However, I wonder if the implementation should be `module.__schema__(:on_writable_violation)`. It should be much simpler and we only need to invoke it if there is a violation indeed.

Sounds good, I will make this change 👍

green-david:
@josevalim I pushed a commit simplifying things a bit by using the `__schema__(:on_writable_violation)` approach you suggested. Let me know if that is what you had in mind👍 

Would also like your thoughts on the warning and exception. It feels like the messages could be more useful by including possibly the schema name? It also seems like the warning could benefit from a stacktrace, but im not sure of the best way to go about that or if that is overkill / a potential performance killer.

I also considered a structured `Ecto.WritableViolationError` instead of a generic `ArgumentError`, but wasn't sure if that was overkill as well. Seems like it could be nice.

Thank you for the feedback as always!

lib/ecto/repo/schema.ex
```diff
@@ -624,6 +625,33 @@ defmodule Ecto.Repo.Schema do
...
 
+  defp drop_non_writable_changes!(changes, non_writable_fields, schema, action) do
+    Enum.reduce(non_writable_fields, changes, fn field, changes ->
+      case Map.pop(changes, field) do
+        {nil, changes} ->
+          changes
+        {_change, changes} ->
+          handle_writable_violation(field, schema, action)
+          changes
+      end
```
josevalim: This can accidentally ignore a field being written to nil. The best is probably:

```
case changes do
  %{^field => _} ->
    handle_writable_violation(field, schema, action)
    Map.delete(changes, field)
  %{} ->
    changes
```

lib/ecto/schema.ex
```diff
@@ -2069,6 +2078,10 @@ defmodule Ecto.Schema do
         raise ArgumentError, "autogenerated fields must always be writable"
       end
 
+      if writable == :always and on_writable_violation != :nothing do
+        raise ArgumentError, "on_writable_violation must be :nothing for always writable fields"
+      end
+
```
josevalim: We probably don't need this check. And maybe users want to set `on_writable_violation: :warn` as a default somewhere and this would get in the way. :)

lib/ecto/schema.ex
```diff
@@ -2077,7 +2090,7 @@ defmodule Ecto.Schema do
         Module.put_attribute(mod, :ecto_query_fields, {name, type})
       end
 
-      Module.put_attribute(mod, :ecto_fields, {name, {type, writable}})
+      Module.put_attribute(mod, :ecto_fields, {name, {type, {writable, on_writable_violation}}})
```
josevalim: We should not need to change this now.

lib/ecto/repo/schema.ex
green-david: 🤦 Thank you for catching this! Will fix 👍

lib/ecto/schema.ex
green-david: I might be missing something - doesn't this need to be included here so that I can reference use the value to populate the data for the `__schema__(:on_writable_violation)` call below? 

Or are you suggesting to separately populate something like `:ecto_on_writable_violations` and use that value in `__schema__` instead perhaps?

I put it here so it was next to `writable` which was pre-existing 🤔

josevalim: Yeah, to use a separate attribute so we don't "pollute" all other fields to store this information.

green-david:
@josevalim Hopefully this is a bit closer! Thank you for the feedback 🙇

josevalim (approved):
@greg-rychlewski, please merge if you see no issues :)

greg-rychlewski:
@green-david sorry do you mind rebasing from main? one of our docker images in the ci had an issue so the unit tests didn't even start. if you rebase it will let us make sure all the tests pass on that job

green-david:
@greg-rychlewski No problem!

I noticed while I was rebasing that there was an issue where we were still silently ignoring surfaced changes even if configured otherwise which I pushed a fix and tests for in a new commit if you wouldn't mind taking a look.

I think there might still be a conflict between the `default` option and the `on_writable_violation` option as with a default value set we always have to surface that value and insert it at insertion time. Perhaps we should only let you configure `on_writable_violation: :nothing` if a `default` is set on the schema field to prevent you from shooting yourself in the foot? WDYT?

I also wonder if it would be desirable to have a Schema attribute like `@on_writable_violation` to be able to configure a global default similar to `@schema_redact` 🤔

greg-rychlewski:
> I also wonder if it would be desirable to have a Schema attribute like @on_writable_violation to be able to configure a global default similar to @schema_redact

That's a good idea. I think it would be good for a follow up PR though because you basically got this one working and don't want to rock the boat.

> I think there might still be a conflict between the default option and the on_writable_violation option as with a default value set we always have to surface that value and insert it at insertion time

I see what you are saying. I'm a bit conflicted because having `violation: :nothing` and something like `writable: :update` means you can use the same schema for inserts and updates and don't have to make two identical ones with the options changed. Maybe `violation: :nothing` with `writable: :never` is a good compromise because the default will never be used at all?

greg-rychlewski:
Or even just `default` with `writable: :never` should be forbidden because even with other violation behaviour you'll just get spammed with raises or warnings

green-david:
> Or even just `default` with `writable: :never` should be forbidden because even with other violation behaviour you'll just get spammed with raises or warnings

Yeah, really the only use case I can think of for `default` with `writable_never` would perhaps be that you have a generated field in the database and for an in-memory built struct you want to default it to a reasonable value so it isn't `nil`, but then you don't want to actually write anything to the DB and let the generated column take over?

greg-rychlewski:
Maybe the best first step is to just leave it for now and see how users react to it and what reports we get. Because if people will start doing complicated things like you shared above then it's better to make less assumptions unless we see it's causing a lot of people problems.

green-david:
> Maybe the best first step is to just leave it for now and see how users react to it and what reports we get. Because if people will start doing complicated things like you shared above then it's better to make less assumptions unless we see it's causing a lot of people problems.

Sounds good, thank you for thinking it through with me!!

greg-rychlewski (approved):

greg-rychlewski:
nice work and great idea :). if you wanted to send another PR for the schema attribute that would be appreciated

PR #4735, opened 2026-06-16
Add cursor fragment splice example

greg-rychlewski:
I stumbled across an interesting use case for fragment splicing today. I thought it might be good for the docs since pagination a relatively non-trivial and popular

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,8 +70,8 @@ jobs:
       fail-fast: false
       matrix:
         elixirbase:
-          - "1.17.3-erlang-27.2-alpine-3.17.9"
-          - "1.17.3-erlang-25.0.4-alpine-3.17.9"
+          - "1.17.3-erlang-27.2-alpine-3.20.9"
+          - "1.17.3-erlang-25.0.4-alpine-3.20.9"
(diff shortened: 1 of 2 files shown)


greg-rychlewski:
some more docker image blues i guess =( i'll look into it

josevalim (approved):

greg-rychlewski:
@josevalim Letting you know what error was happening just in case this is related to something hexpm is doing. It's hard for me to tell if this is erlang or OS or hexpm related:

```
** (Mix) httpc request failed with: {:failed_connect, [{:to_address, {~c"builds.hex.pm", 443}}, {:inet, [:inet], {:tls_alert, {:unsupported_certificate, ~c"TLS client: In state wait_cert_cr at ssl_handshake.erl:2199 generated CLIENT ALERT: Fatal - Unsupported Certificate\n {key_usage_mismatch,{{'Extension',{2,5,29,15},true,[keyCertSign,cRLSign]},\n {'Extension',{2,5,29,37},false,[{1,3,6,1,5,5,7,3,1}]}}}"}}}]}
```

This was happening on certain combinations of erlang and alpine. For example erlang 27.2 and alpine 20.9 for integration tests. And erlang 27.1 using `erlef/setup-beam@v1` action in the unit tests.

PR #4736, opened 2026-06-16
Do not reflect changes ignored due to :writable in returned struct

green-david:
Oops @greg-rychlewski, I realized just slightly too late that I accidentally broke #4524 / #4525 in #4734. This PR fixes the regression and adds accompanying unit tests.

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -456,8 +456,8 @@ defmodule Ecto.Repo.Schema do
     # On insert, we need to nilify non-writable fields in
     # the underlying data so that the returned struct reflects
     # that the write was not actually performed.
-    changeset = update_in(changeset.data, &nilify_unsurfaced_non_writable_data!(&1, drop_fields, schema))
-    changeset = update_in(changeset.changes, &drop_non_writable_changes!(&1, drop_fields, schema, :insert))
+    changeset = nilify_unsurfaced_non_writable_data!(changeset, drop_fields, schema)
+    changeset = drop_non_writable_changes!(changeset, drop_fields, schema, :insert)
 
     wrap_in_transaction(adapter, adapter_meta, opts, changeset, assocs, embeds, prepare, fn ->
       assoc_opts = assoc_opts(assocs, opts)
@@ -526,9 +526,9 @@ defmodule Ecto.Repo.Schema do
     {:error, put_repo_and_action(changeset, :insert, repo, tuplet)}
   end
 
-  defp nilify_unsurfaced_non_writable_data!(data, non_writable_fields, schema) do
+  defp nilify_unsurfaced_non_writable_data!(changeset, non_writable_fields, schema) do
     updates = Enum.reduce(non_writable_fields, [], fn field, updates ->
-      case data do
+      case changeset.data do
         %{^field => value} when value != nil ->
           handle_writable_violation(field, schema, :insert)
 
@@ -538,7 +538,7 @@ defmodule Ecto.Repo.Schema do
       end
     end)
 
-    struct!(data, updates)
+    %{changeset | data: struct!(changeset.data, updates)}
   end
 
   @doc """
@@ -580,7 +580,7 @@ defmodule Ecto.Repo.Schema do
     # fields into the changeset. All changes must be in the
     # changeset before hand.
     changeset = put_repo_and_action(changeset, :update, repo, tuplet)
-    changeset = update_in(changeset.changes, &drop_non_writable_changes!(&1, drop_fields, schema, :update))
+    changeset = drop_non_writable_changes!(changeset, drop_fields, schema, :update)
 
     if changeset.changes != %{} or force? do
       wrap_in_transaction(adapter, adapter_meta, opts, changeset, assocs, embeds, prepare, fn ->
@@ -644,8 +644,8 @@ defmodule Ecto.Repo.Schema do
     {:error, put_repo_and_action(changeset, :update, repo, tuplet)}
   end
 
-  defp drop_non_writable_changes!(changes, non_writable_fields, schema, action) do
-    Enum.reduce(non_writable_fields, changes, fn field, changes ->
+  defp drop_non_writable_changes!(changeset, non_writable_fields, schema, action) do
+   changes = Enum.reduce(non_writable_fields, changeset.changes, fn field, changes ->
       case changes do
         %{^field => _change} ->
           handle_writable_violation(field, schema, action)
@@ -655,6 +655,8 @@ defmodule Ecto.Repo.Schema do
           changes
       end
     end)
+
+    %{changeset | changes: changes}
   end
 
   defp handle_writable_violation(field, schema, action) do
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -626,27 +626,29 @@ defmodule Ecto.Repo.Schema do
   end
 
   defp process_non_writable_fields!(changeset, non_writable_fields, schema, action) do
-    changes =
-      Enum.reduce(non_writable_fields, changeset.changes, fn field, changes ->
+    {changes, data_updates} =
+      Enum.reduce(non_writable_fields, {changeset.changes, []}, fn field, {changes, data_updates} ->
         case changes do
           %{^field => _change} ->
             handle_writable_violation(field, schema, action)
-            Map.delete(changes, field)
+
+            # On insert, we need to nilify non-writable fields in
+            # the underlying data so that the returned struct reflects
+            # that the write was not actually performed.
+            data_updates = case action do
+              :insert -> [{field, nil} | data_updates]
+              _ -> data_updates
+            end
+
+            {Map.delete(changes, field), data_updates}
           %{} ->
-            changes
+            {changes, data_updates}
         end
       end)
 
-    # On insert, if the underlying struct has a value for a non-writable
-    # field, we need to nilify it so that the struct returned from the
-    # insert operation reflects that the write was not actually performed.
-    # Discarding the change above only prevents the actual write from happening.
-    data = case action do
-      :insert ->
-        updates = Enum.map(non_writable_fields, fn field -> {field, nil} end)
-        struct!(changeset.data, updates)
-      _ ->
-        changeset.data
+    data = case data_updates do
+      [] -> changeset.data
+      data_updates -> struct!(changeset.data, data_updates)
     end
 
     %{changeset | data: data, changes: changes}
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -452,8 +452,12 @@ defmodule Ecto.Repo.Schema do
     # On insert, we always merge the whole struct into the
     # changeset as changes, except the primary key if it is nil.
     changeset = put_repo_and_action(changeset, :insert, repo, tuplet)
-    changeset = Relation.surface_changes(changeset, struct, keep_fields ++ drop_fields ++ assocs)
-    changeset = process_non_writable_fields!(changeset, drop_fields, schema, :insert)
+    changeset = Relation.surface_changes(changeset, struct, keep_fields ++ assocs)
+    # On insert, we need to nilify non-writable fields in
+    # the underlying data so that the returned struct reflects
+    # that the write was not actually performed.
+    changeset = update_in(changeset.data, &nilify_unsurfaced_non_writable_data!(&1, drop_fields, schema))
+    changeset = update_in(changeset.changes, &drop_non_writable_changes!(&1, drop_fields, schema, :insert))
 
     wrap_in_transaction(adapter, adapter_meta, opts, changeset, assocs, embeds, prepare, fn ->
       assoc_opts = assoc_opts(assocs, opts)
@@ -522,6 +526,21 @@ defmodule Ecto.Repo.Schema do
     {:error, put_repo_and_action(changeset, :insert, repo, tuplet)}
   end
 
+  defp nilify_unsurfaced_non_writable_data!(data, non_writable_fields, schema) do
+    updates = Enum.reduce(non_writable_fields, [], fn field, updates ->
+      case data do
+        %{^field => value} when value != nil ->
+          handle_writable_violation(field, schema, :insert)
+
+          [{field, nil} | updates]
+        %{} ->
+          updates
+      end
+    end)
+
+    struct!(data, updates)
+  end
+
   @doc """
   Implementation for `Ecto.Repo.update/2`.
   """
@@ -561,7 +580,7 @@ defmodule Ecto.Repo.Schema do
     # fields into the changeset. All changes must be in the
     # changeset before hand.
     changeset = put_repo_and_action(changeset, :update, repo, tuplet)
-    changeset = process_non_writable_fields!(changeset, drop_fields, schema, :update)
+    changeset = update_in(changeset.changes, &drop_non_writable_changes!(&1, drop_fields, schema, :update))
 
     if changeset.changes != %{} or force? do
       wrap_in_transaction(adapter, adapter_meta, opts, changeset, assocs, embeds, prepare, fn ->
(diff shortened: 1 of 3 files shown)


greg-rychlewski:
Wait do I understand correctly that before we were not removing the fields that should be dropped? I'm really confused why our old tests passed then

greg-rychlewski:
Oh I guess the old tests were only checking that we sent the correct fields to the write but not that we were dropping them correctly from the returned struct?

green-david:
> Oh I guess the old tests were only checking that we sent the correct fields to the write but not that we were dropping them correctly from the returned struct?

Exactly! I didnt apply Jose's suggestion correctly in my last PR. We filtered them later before the write.

test/ecto/repo_test.exs
greg-rychlewski:
```suggestion
test "update with on_writable_violation: :nothing saves changes for writable: :always and ignores changes for writable: :insert/:never" do
%{never: nil} =
%MySchemaWritable{id: 1}
|> Ecto.Changeset.change(%{always: 10, never: 11, insert: 12})
|> TestRepo.update!()

assert_received {:update, %{changes: [always: 10]}}
end
```
I think this is enough and then you can also make this minimal change to the existing insert test.

green-david:
@greg-rychlewski I realized as well that even with this fix (and even prior to any of my changes I think) that non-writable _changes_ are not reflected in the returned struct as mentioned in #4524, but that if a non-writable field has a value in the underlying data that is surfaced as a change during insert, the returned struct _does_ reflect an updated value which I dont think should be happening?

For example, this seems to fail:

```elixir
%{never: nil} = 
%SchemaWritable{never: 10}
|> Ecto.Changeset.change(%{})
|> TestRepo.insert!()
```

And this seems to succeed:

```elixir
%{never: nil} = 
%SchemaWritable{}
|> Ecto.Changeset.change(%{never: 10})
|> TestRepo.insert!()
```

I can try to come up with a fix later tonight. Perhaps we just force the underlying field in the data to nil, not sure. Would you prefer a separate PR for that specifically or put it in here?

greg-rychlewski:
I think this PR is ok but I also have to think about it. This is kind of a weird case and I'm a bit wary of over-complicating the already complicated code.

greg-rychlewski:
I guess `nil`ing them out at the end in  the same place we clean up the changes is probably best. But this is weird enough I'd like @josevalim to also take a look after you put the change in

lib/ecto/repo/schema.ex
greg-rychlewski:
```suggestion
    changeset = update_in(changeset.changes, &drop_non_writable_changes!(&1, drop_fields, schema, :insert))
    changeset = update_in(changeset.data, &best_way_to_nil_out_keys_in_elixir(&1, drop_fields))
```
@green-david something like this...i'm not sure off the top of my head if there is a built-in function to do this rather than rolling our own reducer but this is the idea

green-david:
@greg-rychlewski @josevalim This appears to work, I think its ready for another look. 🙇‍♂️

lib/ecto/repo/schema.ex
```diff
@@ -625,16 +625,31 @@ defmodule Ecto.Repo.Schema do
...
+    # field, we need to nilify it so that the struct returned from the
+    # insert operation reflects that the write was not actually performed.
+    # Discarding the change above only prevents the actual write from happening.
+    data = case action do
+      :insert ->
+        updates = Enum.map(non_writable_fields, fn field -> {field, nil} end)
+        struct!(changeset.data, updates)
+      _ ->
+        changeset.data
+    end
```
greg-rychlewski:
```suggestion
   {changes, data_updates} =
      Enum.reduce(non_writable_fields, {changeset.changes, []}, fn field, {changes, data_updates} ->
        case changes do
          %{^field => _change} ->
            handle_writable_violation(field, schema, action)
            data_updates =  if action == :insert, do: [{field, nil} | data_updates], else: data_updates
            {Map.delete(changes, field), data-updates}
          %{} ->
            {changes, data_updates}
        end
      end)

    data = if data_updates != [], do: struct!(changeset.data, data_updates), else: changeset.data
```

maybe something like this...then you only iterate once and also you only update fields that were not `nil` to begin with

test/ecto/repo_test.exs
```diff
@@ -2464,18 +2466,19 @@ defmodule Ecto.RepoTest do
...
+        |> TestRepo.update!()
       end
 
-      %MySchemaWritableRaise{id: 3}
-      |> Ecto.Changeset.change(%{always: 12})
-      |> TestRepo.update()
+      %{always: 12, never: nil, insert: nil} =
+        %MySchemaWritableRaise{id: 3}
+        |> Ecto.Changeset.change(%{always: 12})
+        |> TestRepo.update!()
```
greg-rychlewski: I don't believe we need to change this since it is not trying to update things that shouldn't be updated and we have other tests for the happy path

test/ecto/repo_test.exs
```diff
@@ -2576,12 +2584,13 @@ defmodule Ecto.RepoTest do
...
         %MySchemaWritableRaise{id: 1}
         |> Ecto.Changeset.change(%{never: 10})
-        |> TestRepo.insert()
+        |> TestRepo.insert!()
       end
 
-      %MySchemaWritableRaise{id: 2}
-      |> Ecto.Changeset.change(%{insert: 11, always: 12})
-      |> TestRepo.insert()
+      %{always: 12, never: nil, insert: 11} =
```
greg-rychlewski: I believe we don't need this change either since we are not attempting to change something that shouldn't be changed and we have other tests already asserting the fields that should be changed are present

lib/ecto/repo/schema.ex
green-david: done :)

test/ecto/repo_test.exs
green-david: removed 👍

test/ecto/repo_test.exs
green-david: removed 👍

josevalim:
I am a bit unsure what to do in this case. It feels surfacing the field as changed, only to discard it, is a bit confusing. Perhaps what we whould do is to not surface it and then to a separate pass on insert on the struct for drop fields. So in a nutshell, we remove the `drop_fields` from `surface_changes` and then we do:

```
if insert
  for field <- drop_fields do
    case struct do
      %{^field => value} when value != nil -> 
        # warn
```

green-david:
> I am a bit unsure what to do in this case. It feels surfacing the field as changed, only to discard it, is a bit confusing. Perhaps what we whould do is to not surface it and then to a separate pass on insert on the struct for drop fields. So in a nutshell, we remove the `drop_fields` from `surface_changes` and then we do:
> 
> ```
> if insert
>   for field <- drop_fields do
>     case struct do
>       %{^field => value} when value != nil -> 
>         # warn
> ```

Yeah, I see where you are coming from. 

To my grug brain the current approach is slightly less confusing as it seems like first pass -> figure out what the user is trying to do (surface the changes), second pass -> detect which of those things are bad (look for violations). 

I also kindof like that in the current approach the logic is as similar as possible between update and insert, but that is mostly aesthetic.

Im not stuck on the current approach though if we think special handling it will be more obvious / straightforward to maintain in the long run.

josevalim:
Let’s go with the one I proposed then. We already have special handling for insert in this patch too and I generally dislike when we have code that undoes what we did earlier (in this case, we are unsurfacing the changes). Thanks!

green-david:
> Let’s go with the one I proposed then. We already have special handling for insert in this patch too and I generally dislike when we have code that undoes what we did earlier (in this case, we are unsurfacing the changes). Thanks!

Done - I moved the special handling for insertion out to its own function that is only called in `do_insert` and not `do_update` and no longer surface the non-writable changes. Let me know if this is what you had in mind, hopefully it is close 😄

greg-rychlewski (approved):

lib/ecto/repo/schema.ex
```diff
@@ -522,6 +526,21 @@ defmodule Ecto.Repo.Schema do
     {:error, put_repo_and_action(changeset, :insert, repo, tuplet)}
   end
 
+  defp nilify_unsurfaced_non_writable_data!(data, non_writable_fields, schema) do
+    updates = Enum.reduce(non_writable_fields, [], fn field, updates ->
+      case data do
+        %{^field => value} when value != nil ->
+          handle_writable_violation(field, schema, :insert)
+
+          [{field, nil} | updates]
```
josevalim: Is this how the previous code would behave? If we didn't surface it, I thought we would just leave the field value as is, no?

This is important because if a field has a default value different from nil, we will start setting it back to nil now.

lib/ecto/repo/schema.ex
```diff
@@ -452,7 +452,11 @@ defmodule Ecto.Repo.Schema do
     # On insert, we always merge the whole struct into the
     # changeset as changes, except the primary key if it is nil.
     changeset = put_repo_and_action(changeset, :insert, repo, tuplet)
-    changeset = Relation.surface_changes(changeset, struct, keep_fields ++ drop_fields ++ assocs)
+    changeset = Relation.surface_changes(changeset, struct, keep_fields ++ assocs)
+    # On insert, we need to nilify non-writable fields in
+    # the underlying data so that the returned struct reflects
+    # that the write was not actually performed.
+    changeset = update_in(changeset.data, &nilify_unsurfaced_non_writable_data!(&1, drop_fields, schema))
```
josevalim: Write this as:

```suggestion
    changeset = nilify_unsurfaced_non_writable_data!(changeset, drop_fields, schema)
```

And make it a no-op if drop_fields is an empty list, so we don't do work in the most common case (drop_fields is empty).

lib/ecto/repo/schema.ex
green-david: The behavior is intentionally different as it appeared to be a bug:

#4524 proposed that the most obvious return value for a field whose value was not updated due to `:writable` would be the value actually committed to the database, and #4525 made the changes to do so, but only for actual changes, not surfaced changes.

This led to an odd discrepancy in the previous behavior:

```elixir
# Returned struct does not reflect database state when inserted via a surfaced change
%{never: 10} = 
%SchemaWritable{never: 10}
|> Ecto.Changeset.change(%{})
|> TestRepo.insert!()
```

```elixir
# Returned struct does reflect database state when inserted via an actual change
%{never: nil} = 
%SchemaWritable{}
|> Ecto.Changeset.change(%{never: 10})
|> TestRepo.insert!()
```

This change aligns the behavior amongst the two so at least we are consistent, as the former certainly appears at first glance to be a bug.

If the field has a default value different from nil, and we go to insert the struct, and the field is writable `:never`, we wouldn't have inserted the default value, so should we not return nil to reflect that? (That's the idea anyways)

lib/ecto/repo/schema.ex
green-david: Shall I make the same changes to this?

```elixir
changeset = update_in(changeset.changes, &drop_non_writable_changes!(&1, drop_fields, schema, :insert))
```

to

```elixir
changeset = drop_non_writable_changes!(changeset, drop_fields, schema, :insert)
```

for consistency?

lib/ecto/repo/schema.ex
josevalim: > If the field has a default value different from nil, and we go to insert the struct, and the field is writable :never, we wouldn't have inserted the default value, so should we not return nil to reflect that? (That's the idea anyways)

The problem with this assumption is that it assumes nil is a reasonable value for the field, which may not be the case. If you declare a field with a default of `0`, now we are inserting nil where no nil was ever expected. :(

At the same time, if said field is marked as non-writable, then it means it will always warn, regardless of what the value is, forcing users to use `nil` when an actual value such as `0` may be better.

So I think the best solution overall may be to not warn nor change the value at all in those cases. :(

green-david: > The problem with this assumption is that it assumes nil is a reasonable value for the field, which may not be the case. If you declare a field with a default of `0`, now we are inserting nil where no nil was ever expected. :(

Was this not already the case due to the fact that we filtered out all of the changes before the write silently? I can double check. I think it would get set to the default value in the database, which wouldn't get returned previously unless `:returning` is specified so im not sure the actual write behavior actually changed? (I could be misunderstanding you).

> At the same time, if said field is marked as non-writable, then it means it will always warn, regardless of what the value is, forcing users to use `nil` when an actual value such as `0` may be better.

Yeah, there is a definite not-intuitive conflict between `default` and `writable`. Greg and I discussed it a bit in the original PR [here](https://github.com/elixir-ecto/ecto/pull/4734#issuecomment-4719763955).

It wouldn't have to _always_ warn, no? If they configured `on_writable_violation: :nothing` (which is the default still), it would continue to behave the same (I think?)

> So I think the best solution overall may be to not warn nor change the value at all in those cases. :(

This would certainly be unfortunate. If we do end up having to go this route where surfaced changes can't be detected as writable violations, I guess the best we can do is document it.

josevalim: > Was this not already the case due to the fact that we filtered out all of the changes before the right silently? I can double check.

We don't use database values unless you opt-in via returning. In that case, it makes sense whatever is returned from the database is always ignored.

> It wouldn't have to always warn, no? If they configured on_writable_violation: :nothing (which is the default still), it would continue to behave the same (I think?)

You are right, given you can always opt-out at by saying nothing, then that concern is fine (which is the same conclusion you arrived, sorry for being a slow poke). I'd still say though writable changes the behaviour of `default`, which we should probably avoid. The good news is that it would be just removing code!

green-david: @josevalim Would you mind humoring me with a small summary of what you are suggesting as the path forward? I am happy to make the changes; I just want to be 100% sure I understand what you are proposing before doing so. 🙏

lib/ecto/repo/schema.ex
```diff
@@ -522,6 +526,21 @@ defmodule Ecto.Repo.Schema do
...
+        %{^field => value} when value != nil ->
+          handle_writable_violation(field, schema, :insert)
+
+          [{field, nil} | updates]
+        %{} ->
+          updates
+      end
+    end)
+
+    struct!(data, updates)
```
josevalim: I am propsing something like this:

```suggestion
    if non_writeble_fields != [] and on_writable_violation_for_field(field) != :nothing do
      Enum.each(non_writable_fields, fn field ->
        case data do
          %{^field => value} when value != nil ->
            handle_writable_violation(field, schema, :insert)
          _ ->
            updates
        end
      end)
    end
```

But I am not sure, so I appreciate all of the push back!

josevalim: I pushed an enhancement to the branch which meaqns `on_writable_violation_for_field(field)` is the same as `module.__schema__(:on_writable_violation, field)` :)

green-david: @josevalim Thank you! Just double-checking that by "the branch" you did intend to push to `master`, which is where that got committed, and not this branch 😉

josevalim: Correct!!!

PS: I just want to say that interacting with you on all of these threads has been fantastic, thank you! :heart:

green-david:
@josevalim I went ahead and addressed your other comments (passing the whole changeset to the subfunctions and short-circuiting for an empty list of `drop_fields`). I also went ahead and restored the previous behavior where surfaced changes which are ignored due to `:writable` _are_ reflected in the returned struct whereas "normal" changes which are ignored due to `:writable` are _not_ reflected in the returned struct. We can at least warn or raise if so configured for attempting to write said non-writable surfaced changes now though, which is an improvement.

I'll have to do some more thinking and see if I can come up with something better to address the inconsistency between the returned struct behavior of the two methodologies, but that can be in a future PR perhaps? If we can get this merged, it will at least fix what I broke in #4734 and restore the previous behavior of #4525.

josevalim (approved):

greg-rychlewski (approved):

greg-rychlewski:
thanks :) nice job

PR #4738, opened 2026-06-19
Improve warning/error message for writable violations

green-david:
This PR improves the error message introduced in #4734 to hopefully improve the user experience of actually receiving this warning / error. The updated message is patterned off of the existing message you get with `on_replace: :raise` when trying to replace an association.

The goal of the new message is twofold:

1) By including the schema in the message, it makes clear which actual schema module the field is defined in. If the field is a common field name like `:value`, `:status`, `:archived_at`, etc., it would be difficult to find out exactly which schema with said field is being violated otherwise.
2) By including references to the `:writable` and `:on_writable_violation` options in the message itself, it gives the user a jumping off point to reference the docs directly to figure out how to address the warning / error rather than having to try to figure out where it is coming from.

Open to any other suggestions 😄

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -665,7 +665,15 @@ defmodule Ecto.Repo.Schema do
   defp handle_writable_violation(field, schema, action) do
     on_writable_violation = schema.__schema__(:on_writable_violation, field)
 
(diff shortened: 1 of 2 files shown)


josevalim (approved):

greg-rychlewski:
thanks again :)

PR #4739, opened 2026-07-03
Add UUID v7 helper functions

drtheuns:
Adds `timestamp/1`, `datetime/1`, and `version/1` helpers for dealing with UUIDs.

My main use case is dealing with importing data with existing timestamps but no
UUIDv7 yet. There's currently no way to control the timestamp that's used for
the manual creation of the UUID.

Some parallels with Postgres:

- `uuid_extract_timestamp` -> `Ecto.UUID.timestamp`
- `uuid_extract_version` -> `Ecto.UUID.version`

### DIFF
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -235,6 +235,9 @@ defmodule Ecto.UUID do
    * `:precision` - The timestamp precision for version 7 UUIDs. Supported
      values are `:millisecond` and `:monotonic`. Defaults to `:millisecond`.
 
+   * `:timestamp` - A specific unix timestamp to generate the UUID with.
+     Only works for `precision: :millisecond`.
+
   > #### Monotonic precision {: .info}
   >
   > When using `:monotonic`, sub-millisecond precision is encoded in the
@@ -293,11 +296,12 @@ defmodule Ecto.UUID do
 
   defp bingenerate_v7(opts) do
     {precision, rest} = Keyword.pop(opts, :precision, :millisecond)
+    {timestamp, rest} = Keyword.pop(rest, :timestamp)
     if rest != [], do: raise(ArgumentError, "unsupported options for v7: #{inspect(rest)}")
 
     case precision do
       :millisecond ->
-        timestamp = System.system_time(:millisecond)
+        timestamp = timestamp || System.system_time(:millisecond)
         <<rand_a::12, _::6, rand_b::62>> = :crypto.strong_rand_bytes(10)
         <<timestamp::48, @version_7::4, rand_a::12, @variant::2, rand_b::62>>
 
@@ -374,4 +378,33 @@ defmodule Ecto.UUID do
   defp e(13), do: ?d
   defp e(14), do: ?e
   defp e(15), do: ?f
+
+  @doc """
+  Returns the timestamp from the UUID. Only works for UUID v7.
+
+  Raises `Ecto.ArgumentError` when a UUID is given that's not v7.
+  """
+  def timestamp(<<milliseconds::48, @version_7::4, _::76>>), do: milliseconds
+
+  def timestamp(<<_::48, version::4, _::76>>),
+    do: raise(ArgumentError, "timestamp only supports v7 UUIDs, got v#{version}")
+
+  def timestamp(<<_::288>> = uuid), do: uuid |> dump!() |> timestamp()
+
+  @doc """
+  Tries to fetch the `DateTime` from the UUID. Only works for UUID v7.
+
+  Raises `Ecto.ArgumentError` when a UUID is given that's not v7.
+  """
+  def datetime(uuid) do
+    uuid
+    |> timestamp()
+    |> DateTime.from_unix!(:millisecond)
+  end
+
+  @doc """
+  Returns the version number for the UUID.
+  """
+  def version(<<_::48, version::4, _::76>>), do: version
+  def version(<<_::288>> = uuid), do: uuid |> dump!() |> version()
 end
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -395,12 +395,16 @@ defmodule Ecto.UUID do
   Raises `ArgumentError` for non-v7 UUIDs.
   """
   @spec to_unix(binary()) :: integer()
-  def to_unix(<<unix::48, @version_7::4, _::76>>), do: unix
+  def to_unix(uuid, precision \\ :second)
 
-  def to_unix(<<_::48, version::4, _::76>>),
+  def to_unix(<<unix::48, @version_7::4, _::76>>, precision),
+    do: System.convert_time_unit(unix, :millisecond, precision)
+
+  def to_unix(<<_::48, version::4, _::76>>, _precision),
     do: raise(ArgumentError, "to_unix doesn't support UUID v#{version}")
 
-  def to_unix(uuid), do: uuid |> dump!() |> to_unix()
+  def to_unix(uuid, precision),
+    do: uuid |> dump!() |> to_unix(precision)
 
   @doc """
   Returns the version number for the UUID.
--- a/lib/ecto/uuid.ex
+++ b/lib/ecto/uuid.ex
@@ -235,9 +235,6 @@ defmodule Ecto.UUID do
    * `:precision` - The timestamp precision for version 7 UUIDs. Supported
      values are `:millisecond` and `:monotonic`. Defaults to `:millisecond`.
 
-   * `:timestamp` - A specific unix timestamp to generate the UUID with.
-     Only works for `precision: :millisecond`.
-
   > #### Monotonic precision {: .info}
   >
   > When using `:monotonic`, sub-millisecond precision is encoded in the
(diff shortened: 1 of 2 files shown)


lib/ecto/uuid.ex
```diff
@@ -374,4 +378,33 @@ defmodule Ecto.UUID do
...
+    do: raise(ArgumentError, "timestamp only supports v7 UUIDs, got v#{version}")
+
+  def timestamp(<<_::288>> = uuid), do: uuid |> dump!() |> timestamp()
+
+  @doc """
+  Tries to fetch the `DateTime` from the UUID. Only works for UUID v7.
+
+  Raises `Ecto.ArgumentError` when a UUID is given that's not v7.
+  """
+  def datetime(uuid) do
```
greg-rychlewski: I'm not sure this belongs in Ecto since it's simply composing two different functions.

If Jose would like it in here though, it will currently raise the following error for invalid input which is probably confusing `timestamp only supports v7 UUIDs`

lib/ecto/uuid.ex
```diff
@@ -374,4 +378,33 @@ defmodule Ecto.UUID do
   defp e(13), do: ?d
   defp e(14), do: ?e
   defp e(15), do: ?f
+
+  @doc """
+  Returns the timestamp from the UUID. Only works for UUID v7.
+
+  Raises `Ecto.ArgumentError` when a UUID is given that's not v7.
+  """
+  def timestamp(<<milliseconds::48, @version_7::4, _::76>>), do: milliseconds
```
greg-rychlewski: I'm thinking this is probably a footgun because we are assuming a certain interpretation of the bytes when we accept multiple interpretations. What I mean is if someone is using `monotonic` precision they might be surprised to just get milliseconds back.

lib/ecto/uuid.ex
```diff
@@ -235,6 +235,9 @@ defmodule Ecto.UUID do
    * `:precision` - The timestamp precision for version 7 UUIDs. Supported
      values are `:millisecond` and `:monotonic`. Defaults to `:millisecond`.
 
+   * `:timestamp` - A specific unix timestamp to generate the UUID with.
```
greg-rychlewski: We should probably:

1. Be more explicit what format we are expecting
2. Probably think a bit more what input format would be most useful for users. Not sure if integer is the best or something else.

lib/ecto/uuid.ex
drtheuns: I wasn't entirely sure about this one either. My main motivation for including it was to avoid having to do `uuid |> Ecto.UUID.timestamp() |> DateTime.from_unix!()` everywhere, or adding yet more utils to every single project (that used uuidv7).

lib/ecto/uuid.ex
drtheuns: The function could also be renamed to `milliseconds` or similar. I don't know what the inverse of the monotonic time UUID would be, and I don't think there's even a way to reliably detect which version was used (milliseconds or monotonic) unless perhaps we smuggle 1 bit in one of the random fields. That would only work for UUIDs generated by Ecto though, and it would reduce the randomness

lib/ecto/uuid.ex
drtheuns: I can make it accept either `DateTime` or an integer.

lib/ecto/uuid.ex
drtheuns: Taking a quick peek at Postgres' source code, they don't seem to do anything with the added clock precision for monotonicity when extracting the timestamp:

`uuid_extract_timestamp`: https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/uuid.c#L759

`generate_uuidv7` (called by `uuidv7`): https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/uuid.c#L636

greg-rychlewski:
I don't use UUIDv7 but one thing I am also wondering is why Postgres only allows shifts from the current time rather than setting precise timestamps. I'm guessing it screws up the guarantees somehow and might be insecure. So I think we need to understand this before opening up the option in Ecto.

josevalim:
Let's please revert the `:timestamp` option because it opens up a pandora box. What means to pass it with monotonic precision? We can discuss it separately.

`datetime` is convenient in my opinion because it can also encode the correct precision in the field. One option is to call `to_unix(precision \\ :second)` to mirror datetime. But then should we call the other one `to_datetime`?

We also need to decide if we should raise or return `ok/error` tuples. Note conversion functions in Elixir raises, which would align with the rename proposed above. We should also add `@spec`s.

drtheuns:
I found the following email on the removal of `uuidv7(timestamp)` from Postgres during development: https://www.postgresql.org/message-id/CAGECzQSdRW1PxqRM9F%3DDZ9daCc8D32i7HO7nHE1_ep3mOcv_1g%40mail.gmail.com

The use-case they outline is pretty much exactly my usecase:
> I do agree with others however, that being able to pass in an
arbitrary timestamp for UUID generation would be very useful. For
example to be able to partition by the timestamp in the UUID and then
being able to later load data for an older timestamp and have it be
added to to the older partition

I understand why it shouldn't be a general part of this UUIDv7 implementation though, so I'll add it as a (temp) function for my migration.

---

I've removed the `timestamp` function & option. The `datetime` function was renamed to `to_datetime` & `to_datetime!` to match the functions in `DateTime` (`to_iso8610`, `to_unix`, etc).

josevalim:
Sorry, my feedback was unclear, so let me expand. We should have both to_unix and to_datetime. They raise as before. There are no bang variants for conversion functions in Elixir.

drtheuns:
Hopefully I got it right this time :)

lib/ecto/uuid.ex
```diff
@@ -374,4 +374,38 @@ defmodule Ecto.UUID do
...
+
+  Raises `ArgumentError` for non-v7 UUIDs.
+  """
+  @spec to_unix(binary()) :: integer()
+  def to_unix(<<unix::48, @version_7::4, _::76>>), do: unix
+
+  def to_unix(<<_::48, version::4, _::76>>),
+    do: raise(ArgumentError, "to_unix doesn't support UUID v#{version}")
+
+  def to_unix(uuid), do: uuid |> dump!() |> to_unix()
```
josevalim:
```suggestion
  def to_unix(uuid, precision \\ :second)

  def to_unix(<<unix::48, @version_7::4, _::76>>), do: String.convert_time_unit(unix, :millisecond, precision)

  def to_unix(<<_::48, version::4, _::76>>, _precision),
    do: raise(ArgumentError, "to_unix doesn't support UUID v#{version}")

  def to_unix(uuid, precision), do: uuid |> dump!() |> to_unix(precision)
```

And then please add a test for the precision!

josevalim:
cc @ryanwinchester

ryanwinchester:
> I found the following email on the removal of `uuidv7(timestamp)` from Postgres during development: https://www.postgresql.org/message-id/CAGECzQSdRW1PxqRM9F%3DDZ9daCc8D32i7HO7nHE1_ep3mOcv_1g%40mail.gmail.com
> 
> The use-case they outline is pretty much exactly my usecase:
> 
> > I do agree with others however, that being able to pass in an
> > arbitrary timestamp for UUID generation would be very useful. For
> > example to be able to partition by the timestamp in the UUID and then
> > being able to later load data for an older timestamp and have it be
> > added to to the older partition
> 
> I understand why it shouldn't be a general part of this UUIDv7 implementation though, so I'll add it as a (temp) function for my migration.

Having the "from timestamp" helper function is nice if you need to backfill records and want them with a specific timestamp (I do do this sometimes, and just make my own helper). However, I agree with the reasons they stated that it doesn't fit UUIDv7 spec and that making it UUIDv8 makes sense.

Probably deserves a discussion outside of this PR.

In my opinion, just wait for the postgres folks to make a decision and then copy that, since they have more minds dedicated to these discussions.

lib/ecto/uuid.ex
```diff
@@ -374,4 +374,42 @@ defmodule Ecto.UUID do
...
+    do: raise(ArgumentError, "to_datetime doesn't support UUID v#{version}")
+
+  def to_datetime(uuid), do: uuid |> dump!() |> to_datetime()
+
+  @doc """
+  Extracts the unix timestamp from the UUID. Only works for v7 UUIDs.
+
+  Raises `ArgumentError` for non-v7 UUIDs.
+  """
+  @spec to_unix(binary()) :: integer()
```
ryanwinchester: missing `precision` in spec

ryanwinchester (approved):
Works like postgres' `uuid_extract_timestamp` 👍

warmwaffles:
> Having the helper function is nice if you need to backfill records and want them with a specific timestamp (I do do this sometimes, and just make my own helper).

We had convert integer PKs to UUIDv7s and I wanted to set the timestamp based on the `inserted_at` column that was with the record to remain faithful to the v7 spec. I resorted to my own helper.

Something like this with `:uniq` (for the next person who is searching for help on how to do this with a helper 😉)

```elixir
  def generate_uuid(unixtime) do
    <<rand_a::12, _::6, rand_b::62>> = :crypto.strong_rand_bytes(10)

    raw = <<unixtime::biguint(48), 7::4, rand_a::12, 2::2, rand_b::62>>
    Uniq.UUID.format(raw, :default)
  end
```

Edit: with `Ecto.UUID`

```elixir
  def generate_uuid(unixtime) do
    <<rand_a::12, _::6, rand_b::62>> = :crypto.strong_rand_bytes(10)

    raw = <<unixtime::biguint(48), 7::4, rand_a::12, 2::2, rand_b::62>>
    Ecto.UUID.cast!(raw)
  end
```

drtheuns:
> > Having the helper function is nice if you need to backfill records and want them with a specific timestamp (I do do this sometimes, and just make my own helper).
> 
> We had convert integer PKs to UUIDv7s and I wanted to set the timestamp based on the `inserted_at` column that was with the record to remain faithful to the v7 spec. I resorted to my own helper.
> 
> Something like this with `:uniq` (for the next person who is searching for help on how to do this with a helper 😉)
> 
> ```elixir
>   def generate_uuid(unixtime) do
>     <<rand_a::12, _::6, rand_b::62>> = :crypto.strong_rand_bytes(10)
> 
>     raw = <<unixtime::biguint(48), 7::4, rand_a::12, 2::2, rand_b::62>>
>     Uniq.UUID.format(raw, :default)
>   end
> ```

You can also replace the `Uniq.UUID.format` with just `Ecto.UUID.cast!(raw)`.

josevalim:

PR #4741, opened 2026-07-03
Fix wrong placeholder numbering when a `from` source is a `{fragment, schema}` tuple

### LINKED ISSUE #4740: Wrong placeholder numbering when a `from` source is a `{fragment, schema}` tuple
### Elixir version

Elixir 1.19.4 (compiled with Erlang/OTP 28)

### Database and Version

any

### Ecto Versions

3.14.0

any

### Current behavior

When a query source is a fragment with an attached schema (i.e. `from w in {fragment(...), Schema}`) any parameters interpolated into that fragment are not counted when Ecto assigns final placeholder positions. As a result, parameters in subsequent clauses (where, join, etc.) are numbered as if the fragment's parameters didn't exist, producing colliding placeholders ($1 reused) and a broken SQL query.

Steps to reproduce
A plain fragment source numbers placeholders correctly:

```elixir
query =
  from w in fragment("my_function(?)", ^value),
    where: w.some_field in ^allowed_list,
    select: w.id
```

```sql
SELECT f0."id" FROM my_function($1) AS f0 WHERE (f0."some_field" = ANY($2)) [~U[2026-07-03 15:00:00Z], ["foo", "bar"]]
```

Attaching a schema to the same fragment source breaks the numbering:

```elixir
query =
  from w in {fragment("my_function(?)", ^value), MyApp.MyEntity},
    where: w.some_field in ^allowed_list,
    select: w
```

```sql
SELECT f0."id", ... , ... FROM my_function($1) AS f0 WHERE (f0."some_field" = ANY($1)) [~U[2026-07-03 15:00:00Z], ["foo", "bar"]]
```

The where clause emits `= ANY($1)`, binding the datetime to `ANY(...)` instead of the allowed list, so the database raises a type error (`ANY expects an array`).

### Expected behavior

The fragment's interpolation should occupy `$1` and the where list should be numbered `$2`, exactly as in the schema-less case.

Gigitsu:
Fixes #4740

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1522,6 +1522,11 @@ defmodule Ecto.Query.Planner do
     {{:fragment, meta, fragments}, acc}
   end
 
+  defp prewalk_source({{:fragment, meta, fragments}, schema}, kind, query, expr, acc, adapter) do
+    {fragments, acc} = prewalk(fragments, kind, query, expr, acc, adapter)
+    {{{:fragment, meta, fragments}, schema}, acc}
+  end
+
   defp prewalk_source({:values, meta, [types, num_rows]}, _kind, _query, _expr, acc, _adapter) do
     length = num_rows * length(types)
     # Adapters will use the schema types to cast the values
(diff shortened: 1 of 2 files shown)


josevalim (approved):

greg-rychlewski (approved):

greg-rychlewski:
Thanks very much!

PR #4743, opened 2026-07-04
Match on correct tuple shape in insert_all placeholder type mismatch check

### LINKED ISSUE #4742: Repo.insert_all silently corrupts placeholders
### Elixir version

any

### Database and Version

any

### Ecto Versions

3.14.0

any

### Current behavior

`Repo.insert_all/3` corrupts placeholders when the same placeholder key is reused across columns with different types.

Repro:

```elixir
defmodule PlaceholderSchema do
  use Ecto.Schema

  schema "placeholder_schema" do
    field :bid, :binary_id
    field :str, :string
  end
end

ph = {:placeholder, :k}
uuid = Ecto.UUID.generate()

TestRepo.insert_all(
  PlaceholderSchema,
  [%{bid: ph, str: ph}],
  placeholders: %{k: uuid}
)

receive do
  {:insert_all, meta, rows} ->
    IO.inspect({meta.placeholders, rows})
end
```

Result:
```
{
  [<<...16-byte uuid...>>],
  [[bid: {:placeholder, 2}, str: {:placeholder, 1}]]
}
```

The adapter receives only one placeholder param, but the row references placeholder indexes 1 and 2, `$2` is bound to wrong value. The bug was introduced in 479484c2 which changed
`Ecto.Repo.Schema.extract_placeholder/6` to match 4-tuples but left error case matching a legacy 3-tuple:
https://github.com/elixir-ecto/ecto/blob/32d07c880caeb3894c20e2ef58c3e6c3177a3b99/lib/ecto/repo/schema.ex#L283-L307
the error handling case is a dead code and the placeholder entry is overwritten instead of raising.

The existing test passes by chance - it references a field that does not exist on the test schema, which also happens to raise ArgumentError.

### Expected behavior

Reusing the same placeholder key for columns with different types should raise

lukaszsamson:
Fixes #4742

### DIFF
--- a/lib/ecto/repo/schema.ex
+++ b/lib/ecto/repo/schema.ex
@@ -285,7 +285,7 @@ defmodule Ecto.Repo.Schema do
       %{^key => {idx, ^type, _, _}} ->
         {{:placeholder, idx}, placeholder_dump, counter}
 
-      %{^key => {_, type, _}} ->
+      %{^key => {_, type, _, _}} ->
         raise ArgumentError,
               "a placeholder key can only be used with columns of the same type. " <>
                 "The key #{inspect(key)} has already been dumped as a #{inspect(type)}"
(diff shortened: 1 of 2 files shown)


josevalim:

PR #4746, opened 2026-07-07
Warn on parentheses in query field access

### LINKED ISSUE #4588: Ecto allows parentheses in field names without raising an error
### Elixir version

latest

### Database and Version

any?

### Ecto Versions

main

ecto without adapters

### Current behavior

Adding parentheses to field names (e.g., `q.title()`) is accepted and processed as if the parentheses weren't present. This behavior could be misleading.

Example:
```elixir
# This works:
order_by("q", [q], [asc: q.title()])

# Same as:
order_by("q", [q], [asc: q.title])
```

### Expected behavior

I would expect to get an error

arpitjain099:
Closes #4588. Writing a field access with empty parentheses like `q.title()` is currently accepted and behaves exactly like `q.title`, which is almost always a typo. Per josevalim's note on the issue ("we should definitely warn and then convert to an error in the future"), this warns instead of raising for now.

The two forms only differ by `no_parens: true` on the outer call metadata (confirmed in the thread), so the escape clause now binds that metadata and emits a deprecation warning via `IO.warn/2` when it is absent, then proceeds unchanged. Added a builder test asserting the warning fires for `x.y()` and stays quiet for `x.y`.

One heads-up: I don't have an Elixir toolchain on this machine, so I verified by code-read rather than running `mix test` locally, and I am counting on CI here. Also, a handful of existing builder tests use the parenthesized form (`x.y()`), so they will now print this warning during the run without failing; happy to follow up and quiet those if you would like.

### DIFF
--- a/lib/ecto/query/builder.ex
+++ b/lib/ecto/query/builder.ex
@@ -84,8 +84,19 @@ defmodule Ecto.Query.Builder do
   def escape(expr, type, params_acc, vars, env)
 
   # var.x - where var is bound
-  def escape({{:., _, [callee, field]}, _, []}, _type, params_acc, vars, _env)
+  def escape({{:., _, [callee, field]}, call_meta, []}, _type, params_acc, vars, env)
       when is_atom(field) do
+    if call_meta[:no_parens] != true do
+      stacktrace = Macro.Env.stacktrace(get_env(env))
+
+      IO.warn(
+        "using parentheses after a field access, such as `#{Macro.to_string(callee)}.#{field}()`, " <>
+          "is deprecated and will raise in future Ecto versions. Remove the parentheses to keep " <>
+          "the current behaviour: `#{Macro.to_string(callee)}.#{field}`",
+        stacktrace
+      )
+    end
+
     {escape_field!(callee, field, vars), params_acc}
   end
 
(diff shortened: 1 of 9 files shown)


greg-rychlewski:
Thanks for the PR @arpitjain099. I think we have a bunch of tests that unfortunately trigger this warning. If you run locally you will see a lot of these

```elixir
Warning:   test/ecto/query/builder/distinct_test.exs:19: Ecto.Query.Builder.DistinctTest."test escape handles expressions and params"/1

warning: using parentheses after a field access, such as `x.x()`, is deprecated and will raise in future Ecto versions. Remove the parentheses to keep the current behaviour: `x.x`
```

and the test is like this

```elixir
 assert {Macro.escape(quote(do: [asc: &0.x(), asc: &1.y()])), {[], %{}}} ==
               escape(quote(do: [x.x(), y.y()]), {[], %{}}, [x: 0, y: 1], __ENV__)
```

Would you have time to change the tests as well? If not it's ok I can do a cleanup PR right after.

arpitjain099:
Thanks @greg-rychlewski, happy to. I went through the builder tests and dropped the empty parens from the field accesses that were emitting the warning (distinct, filter, group_by, order_by, select, update, windows, and the main builder_test). Those were all on the `escape/1` input side, and since the escaped output is the same either way the assertions didn't need to change. I left the parens on the one case in builder_test that intentionally checks the warning is raised.

Ran the full suite locally and it's green again with no more `using parentheses` warnings. Let me know if you'd rather I split the test cleanup into its own commit or anything else.

greg-rychlewski (approved):

josevalim:

PR #4748, opened 2026-07-08
Fix data-modifying CTE planner cache traversal

### LINKED ISSUE #4747: Data-modifying CTEs drop/misorder params and collide query-cache keys
### Elixir version

any

### Database and Version

Ecto.TestAdapter and postgesql

### Ecto Versions

3.14.0

postgrex

### Current behavior

`with_cte(..., operation: :update_all | :delete_all)` is documented for data-modifying CTEs, but the query-cache/params planner pass traverses inner CTE queries as `:all`:
```elixir
{_, params, inner_cache} = traverse_cache(query, :all, {[], params}, adapter)
```

This causes three bugs

1. update params are dropped

Repro: 
```elixir
inner =
  from(c in "cats",
    where: c.id == ^10,
    update: [set: [name: ^"Root"]],
    select: c.id
  )

outer =
  "outer"
  |> with_cte("cte", as: ^inner, operation: :update_all)
  |> select([o], o.id)

{_query, params, _cache_key} = Ecto.Query.Planner.plan(outer, :all, Ecto.TestAdapter)
IO.inspect(params)
```
Actual:
```
[{10, 10}]
```

Notce "Root" update value is missing. The normalized query still numbers two placeholders, so the prepared statement has two placeholders but only one param is sent.

2. cache-key collision

Repro:
```elixir
i1 = from(c in "cats", update: [set: [visits: 1]], select: c.id)
i2 = from(c in "cats", update: [set: [visits: 2]], select: c.id)

o1 = "outer" |> with_cte("cte", as: ^i1, operation: :update_all) |> select([o], o.id)
o2 = "outer" |> with_cte("cte", as: ^i2, operation: :update_all) |> select([o], o.id)

{_, _, k1} = Ecto.Query.Planner.plan(o1, :all, Ecto.TestAdapter)
{_, _, k2} = Ecto.Query.Planner.plan(o2, :all, Ecto.TestAdapter)

IO.inspect(k1 == k2)
```
Actual:
```
true
```

The prepared statement for `visits: 1` can be reused for `visits: 2`

3. `delete_all` CTE params are ordered using `:all` traversal order, so select params come before where params, contrary to delete normalization order.

### Expected behavior

The cache/params planner pass should traverse each CTE query using the CTE’s declared operation

lukaszsamson:
Fixes #4747

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -998,7 +998,7 @@ defmodule Ecto.Query.Planner do
     # We could group them to avoid multiple keys, but since they are uncommon, we keep it simple.
     Enum.reduce(queries, cache_and_params, fn
       {name, opts, %Ecto.Query{} = query}, {cache, params} ->
-        {_, params, inner_cache} = traverse_cache(query, :all, {[], params}, adapter)
+        {_, params, inner_cache} = traverse_cache(query, opts.operation, {[], params}, adapter)
 
         {merge_cache(
            {key, name, opts[:materialized], opts[:operation], inner_cache},
--- a/test/ecto/query/planner_test.exs
+++ b/test/ecto/query/planner_test.exs
@@ -1111,6 +1111,93 @@ defmodule Ecto.Query.PlannerTest do
              ] = cte_cache
     end
 
+    test "on update_all with data-modifying CTE" do
+      update_comments =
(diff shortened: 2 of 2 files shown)


greg-rychlewski (approved):

greg-rychlewski:
thank you

PR #4750, opened 2026-07-08
Fix splice params in dynamic expressions

lukaszsamson:
This PR does 2 changes

1. dynamic.ex changes `expand/4` to recurese accumulating splices into a map. Previously the postwalk appended the shared {list, {:splice, :any}} param once per placeholder occurrence resulting in k placeholders x k copies of params and mis-bound placeholders. Now the first occurrence appends the param, so the following ones can reuse it
2. builder.ex changes `length(unquote(params))` into `unquote(length(params))` as params is a plain compile-time list

Fixes #4748


josevalim:

PR #4756, opened 2026-07-18
Fix stale belongs_to key in apply_changes

### LINKED ISSUE #4755: `apply_changes/1` leaves a stale `belongs_to` FK when the association is removed
### Elixir version

any

### Database and Version

any

### Ecto Versions

3.14.1

any

### Current behavior

When a `belongs_to` association is removed with `put_assoc/3`, `apply_changes/1` sets the association field to `nil` but leaves the existing foreign key unchanged. This produces an internally inconsistent struct. Persisting the same changeset through `Repo.update/2` correctly clears the foreign key.

Repro:
```elixir
defmodule Category do
  use Ecto.Schema

  schema "categories" do
  end
end

defmodule Post do
  use Ecto.Schema

  schema "posts" do
    belongs_to :category, Category
  end
end

post = %Post{category_id: 5}

changed_post =
  post
  |> Ecto.Changeset.change()
  |> Ecto.Changeset.put_assoc(:category, nil)
  |> Ecto.Changeset.apply_changes()

IO.inspect(changed_post.category) # nil
IO.inspect(changed_post.category_id) # 5, should be nil
```

`apply_action/2` is also affected

Note that e028a909 added a test asserting the correct behavior but its fixture FK is already `nil` and it passes despite the bug.

### Expected behavior

`apply_changes/1` clears FK

lukaszsamson:
Fixes #4755

### DIFF
--- a/lib/ecto/changeset.ex
+++ b/lib/ecto/changeset.ex
@@ -4323,14 +4323,13 @@ defmodule Ecto.Changeset do
 
     acc = Map.put(acc, key, relation_changed)
 
-    with %Ecto.Association.BelongsTo{related_key: related_key, owner_key: owner_key} <- relation do
-      case relation_changed do
-        %{^related_key => id} -> Map.put(acc, owner_key, id)
-        nil -> Map.put(acc, owner_key, nil)
-        _ -> acc
-      end
-    else
-      _ -> acc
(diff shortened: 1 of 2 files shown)


josevalim:

PR #4758, opened 2026-07-18
Fix map update source field caching

### LINKED ISSUE #4757: Map update on the `from` binding poisons other full-source references in `select`
### Elixir version

any

### Database and Version

any

### Ecto Versions

3.14.1

any

### Current behavior

Selecting a map update of the from binding before another full reference to that binding causes the planner to omit updated fields from the shared source selection. The result depends on tuple order.

Repro:
```elixir
defmodule Post do
  use Ecto.Schema

  schema "posts" do
    field :title, :string
    field :body, :string
  end
end

query1 =
  from(p in Post,
    select: {%{p | title: nil}, p}
  ) |> normalize()

query2 =
  from(p in Post,
    select: {p, %{p | title: nil}}
  ) |> normalize()

query1.select.fields # [&0.id(), &0.body()], title dropped
query2.select.fields # [&0.id(), &0.title(), &0.body()]
```

### Expected behavior

Both query forms should select all required fields

lukaszsamson:
Fixes #4757

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -1785,7 +1785,7 @@ defmodule Ecto.Query.Planner do
 
     {fields, preprocess, from} =
       case from do
-        {from_expr, from_source, from_fields} ->
+        {from_expr, from_source, from_fields, _drop} ->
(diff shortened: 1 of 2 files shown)


josevalim:

PR #4760, opened 2026-07-21
Allow selecting fragment source with columns

greg-rychlewski:
Follow up to the last PR.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -2251,12 +2251,13 @@ defmodule Ecto.Query.Planner do
         {{:map, Enum.map(fields, &{&1, {:value, :any}})},
          Enum.map(fields, &select_field(&1, ix, :always))}
 
-      {:error, {:fragment, [column_names: columns], _}} ->
-        {types, fields} = select_dump(columns, %{}, ix, drop)
-        {{:source, :fragment, nil, types}, fields}
-
-      {:error, {:fragment, _, _}} ->
-        {{:value, :map}, [{:&, [], [ix]}]}
+      {:error, {:fragment, meta, _}} ->
+        if columns = meta[:column_names] do
+          {{:map, Enum.map(columns, &{&1, {:value, :any}})},
+           Enum.map(columns, &select_field(&1, ix, :always))}
+        else
+          {{:value, :map}, [{:&, [], [ix]}]}
+        end
 
       {:error, {:values, _, [types, _]}} ->
         fields = Keyword.keys(types)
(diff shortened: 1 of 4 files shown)


lib/ecto/query/planner.ex
```diff
@@ -2248,6 +2251,10 @@ defmodule Ecto.Query.Planner do
         {{:map, Enum.map(fields, &{&1, {:value, :any}})},
          Enum.map(fields, &select_field(&1, ix, :always))}
 
+      {:error, {:fragment, [column_names: columns], _}} ->
+        {types, fields} = select_dump(columns, %{}, ix, drop)
+        {{:source, :fragment, nil, types}, fields}
+
       {:error, {:fragment, _, _}} ->
         {{:value, :map}, [{:&, [], [ix]}]}
```
greg-rychlewski: I don't think this is used, at least by our SQL adapters. It looks like maybe this is for Mongo or something?

josevalim: I think that was the original idea, yes.

lib/ecto/query/planner.ex
```diff
@@ -2248,6 +2251,10 @@ defmodule Ecto.Query.Planner do
         {{:map, Enum.map(fields, &{&1, {:value, :any}})},
          Enum.map(fields, &select_field(&1, ix, :always))}
 
+      {:error, {:fragment, [column_names: columns], _}} ->
```
josevalim: Since the meta is a keyword list, it would be better to read its contents rather than pattern match on it:

```
{:fragment, meta, _} ->
  if meta[:column_names] do
    ...
  else
    raise ...
  end
```

josevalim (approved):
One comment and ship it!!!!

lib/ecto/query/planner.ex
```diff
@@ -2248,8 +2251,13 @@ defmodule Ecto.Query.Planner do
         {{:map, Enum.map(fields, &{&1, {:value, :any}})},
          Enum.map(fields, &select_field(&1, ix, :always))}
 
-      {:error, {:fragment, _, _}} ->
-        {{:value, :map}, [{:&, [], [ix]}]}
+      {:error, {:fragment, meta, _}} ->
+        if columns = meta[:column_names] do
+          {{:map, Enum.map(columns, &{&1, {:value, :any}})},
+           Enum.map(columns, &select_field(&1, ix, :always))}
```
greg-rychlewski, line 2257: @josevalim Sorry while making your change I actually noticed we already had a way to handle map creation without adding a new `{:source, :fragment, ...}` construct. Just letting you take a look again before mering

josevalim (approved):

PR #4762, opened 2026-07-23
Fix preloader validation for multiple combinations

lukaszsamson:
The previous code matched on exactly one combination leading to a cryptic DB error or misassociated preloads.
This PR relaxes the pattern so an expected `ArgumentError` can be risen


josevalim:

PR #4764, opened 2026-07-24
Fix subqueries in interpolated join wheres

### LINKED ISSUE #4763: Interpolating a query containing a where subquery into a join crashes the planner
### Elixir version

any

### Database and Version

any

### Ecto Versions

3.14

any

### Current behavior

Repro:
```elixir
import Ecto.Query

sq =
  from s in "subposts",
    select: s.id

join_query =
  from p in "posts",
    where: p.id in subquery(sq)

query =
  from p in "parent_posts",
    join: p2 in ^join_query,
    on: true

Ecto.Query.Planner.plan(query, :all, Ecto.TestAdapter)
```
Result:
```
** (KeyError) key :subqueries not found in:

%Ecto.Query.QueryExpr{
  expr: {:in, ..., [{:subquery, 0}]},
  params: [subquery: 0],
  ...
}
```

### Expected behavior

`where` expression should be folded into the join’s on expression and planned normally

lukaszsamson:
Fixes #4763

### DIFF
--- a/lib/ecto/query.ex
+++ b/lib/ecto/query.ex
@@ -428,7 +428,7 @@ defmodule Ecto.Query do
 
   defmodule QueryExpr do
     @moduledoc false
-    defstruct [:expr, :file, :line, params: []]
+    defstruct [:expr, :file, :line, params: [], subqueries: []]
   end
 
   defmodule ByExpr do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   * [Ecto.Query] Fix data-modifying CTE planner cache traversal
   * [Ecto.Query] Fix splice parameters in dynamic expressions
   * [Ecto.Query] Fix subquery order in query builders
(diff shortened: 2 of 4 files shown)


lib/ecto/query.ex
```diff
@@ -428,7 +428,7 @@ defmodule Ecto.Query do
 
   defmodule QueryExpr do
     @moduledoc false
-    defstruct [:expr, :file, :line, params: []]
+    defstruct [:expr, :file, :line, params: [], subqueries: []]
```
josevalim, line 431: Can we instead use a BooleanExpr or add subqueries to the JoinExpr?

lukaszsamson: I can try that

lukaszsamson: @josevalim I had fable create an alternative basing on your suggestion https://github.com/elixir-ecto/ecto/pull/4765

Note the `JoinExpr.subqueries` alternative would be more problematic as it requires threading context through traversals.

PR #4765, opened 2026-07-24
Fix subqueries in interpolated join wheres

lukaszsamson:
## Summary

Fix planning an interpolated join query whose `where` expression contains a subquery.

This is an alternative implementation to #4764 that incorporates the review suggestion to keep subqueries off `QueryExpr`:

* Fold interpolated join filters into a `BooleanExpr`, which already owns subqueries.
* Plan, merge, reindex, and cache those subqueries with the join expression.
* Convert the temporary `BooleanExpr` back to `QueryExpr` after normalization, preserving the contract expected by SQL adapters.
* Allow query inspection while the temporary boolean join expression is present.

## Root cause

Interpolated query filters are folded into the join `on`. Previously that produced a `QueryExpr` without the `subqueries` metadata required to resolve `{:subquery, index}` placeholders, causing a `KeyError` during parameter planning.

The join cache key also needs the nested subquery cache keys; otherwise structurally different subqueries can collide and reuse the wrong prepared SQL.

## Validation

* `mix compile --warnings-as-errors`
* `mix test` — 1,581 tests passed

Regression coverage verifies planning, subquery parameter/index merging, cache-key differentiation, normalization-time inlining, and restoration of `JoinExpr.on` to `QueryExpr`.

### DIFF
--- a/lib/ecto/query/planner.ex
+++ b/lib/ecto/query/planner.ex
@@ -50,12 +50,12 @@ defmodule Ecto.Query.Planner do
   defp merge_expr_and_params(
          op,
          %BooleanExpr{expr: left_expr, params: left_params, subqueries: left_subqueries} = struct,
-         %{expr: right_expr, params: right_params} = right
+         %BooleanExpr{
+           expr: right_expr,
+           params: right_params,
+           subqueries: right_subqueries
+         }
        ) do
-    # The right side may be a QueryExpr (an explicit join `on`),
-    # which holds no subqueries
-    right_subqueries = Map.get(right, :subqueries, [])
-
     merge_expr_and_params(
       op,
       struct,
@@ -71,10 +71,8 @@ defmodule Ecto.Query.Planner do
   defp merge_expr_and_params(
          op,
          %QueryExpr{expr: left_expr, params: left_params} = struct,
-         %{expr: right_expr, params: right_params} = right
+         %BooleanExpr{expr: right_expr, params: right_params, subqueries: []}
        ) do
-    # A QueryExpr cannot hold subqueries, so none may come from the right side
-    [] = Map.get(right, :subqueries, [])
     merge_expr_and_params(op, struct, left_expr, left_params, [], right_expr, right_params, [])
   end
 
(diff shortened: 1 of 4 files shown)


lib/ecto/query/planner.ex
```diff
@@ -47,14 +47,73 @@ defmodule Ecto.Query.Planner do
     end
   end
 
+  defp merge_expr_and_params(
+         op,
+         %BooleanExpr{expr: left_expr, params: left_params, subqueries: left_subqueries} = struct,
+         %{expr: right_expr, params: right_params} = right
+       ) do
+    # The right side may be a QueryExpr (an explicit join `on`),
+    # which holds no subqueries
```
josevalim: Can we make the right-side be a BooleanExpr as well always?

lukaszsamson: done

lib/ecto/query/planner.ex
```diff
@@ -1422,7 +1503,7 @@ defmodule Ecto.Query.Planner do
     Enum.map_reduce(exprs, counter, fn join, acc ->
       {source, acc} = prewalk_source(join.source, :join, query, join, acc, adapter)
       {on, acc} = prewalk(:join, query, join.on, acc, adapter)
-      {%{join | on: on, source: source, params: nil}, acc}
+      {%{join | on: on_to_query_expr(on), source: source, params: nil}, acc}
```
josevalim: Why do we need to rollback? Let's keep it as boolean everywhere... if necessary, we change ecto_sql.

lukaszsamson: Currently ecto_sql pattern matches on QueryExpr. Relaxing that is doable. I created a [draft PR](https://github.com/elixir-ecto/ecto_sql/pull/747). The tests pass but I'm not convinced this is the right way. Providers copy the code patterns from ecto and will remain with silent failures (for expressions will filter out  %BooleanExpr{}) if ecto_sql is loosly pinned.

Besides, after this change `join.on` would be mixed-type: BooleanExpr/QueryExpr. If you want booleans everywhere we would need a bigger refactor making the join builder emit BooleanExpr for all `on`s

josevalim: > Besides, after this change join.on would be mixed-type: BooleanExpr/QueryExpr. If you want booleans everywhere we would need a bigger refactor making the join builder emit BooleanExpr for all ons

Correct. That's what we should do. We already did this change a while ago to support subqueries in `where`. This is just a natural continuation of it.

lukaszsamson: Is that what you had in mind?

josevalim (approved):
@greg-rychlewski you also happy with this one?

greg-rychlewski (approved):

greg-rychlewski:
LGTM!

josevalim:

PR #4766, opened 2026-07-25
Propagate the caller's Logger level to preload tasks

makoto-developer:
Preloads run in parallel tasks (`Task.async_stream` in `Ecto.Repo.Preloader`), and those tasks don't inherit the caller's per-process Logger level. So `Logger.put_process_level/2` (or `put_module_level`) silences the main query but not the queries issued while preloading.

This carries the caller's process level into each preload task. Doing it by default was suggested by the maintainers in elixir-ecto/ecto_sql#602.

`Logger.get_process_level/1` and `put_process_level/2` are Elixir 1.15+, so the propagation is guarded to keep 1.14 support (it's a no-op there).

Added a test that sets a process level and checks the preload tasks pick it up.


josevalim:

makoto-developer:
Thank you!

PR #4773, opened 2026-07-29
Fix autogenerate options for composite types

### LINKED ISSUE #4772: `autogenerate: [opts]` crashes on parameterized and composite field types
### Elixir version

any

### Database and Version

any

### Ecto Versions

3.14.1

any

### Current behavior

Using the list form of the `:autogenerate` field option with a parameterized or composite type compiles successfully but crashes on the first Repo.insert/2.
This appears to be a regression introduced by e6da70d7

Repro:
```elixir
defmodule MyParameterizedType do
  use Ecto.ParameterizedType

  def init(opts), do: opts
  def type(_params), do: :string

  def cast(value, _params), do: {:ok, value}
  def load(value, _loader, _params), do: {:ok, value}
  def dump(value, _dumper, _params), do: {:ok, value}

  def autogenerate(opts), do: Keyword.fetch!(opts, :value)
end

defmodule Example do
  use Ecto.Schema

  schema "examples" do
    field :value, MyParameterizedType,
      autogenerate: [value: "generated"]
  end
end

Repo.insert!(%Example{})
```

Result:
```
** (ArgumentError) you attempted to apply a function on
{:parameterized, {MyParameterizedType, ...}}.
Modules (the first argument of apply) must always be an atom
```

The same problem affects composite types
```elixir
field :values, {:array, MyType},
  autogenerate: [value: ["generated"]]
```

### Expected behavior

No crash

lukaszsamson:
Fixes #4772

### DIFF
--- a/lib/ecto/schema.ex
+++ b/lib/ecto/schema.ex
@@ -2066,7 +2066,8 @@ defmodule Ecto.Schema do
           store_mfa_autogenerate!(mod, name, type, gen)
 
         autogenerate_opts when is_list(autogenerate_opts) ->
-          store_mfa_autogenerate!(mod, name, type, {type, :autogenerate, [autogenerate_opts]})
+          mfa = {autogenerate_module(type), :autogenerate, [autogenerate_opts]}
+          store_mfa_autogenerate!(mod, name, type, mfa)
 
         true ->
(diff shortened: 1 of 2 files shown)


josevalim:

PR #4777, opened 2026-08-04
fix selecting list of fields and map/2 from subquery

### LINKED ISSUE #4776: `select: map(source, fields)` and `select: [:fields]` lose field types when the source is a subquery
### Elixir version

1.20.2 (Erlang/OTP 29)

### Database and Version

PostgreSQL 18.4

### Ecto Versions

ecto 3.14.1, also reproduced on `main` at e7144cd5

postgrex 0.22.3, ecto_sql 3.14.0

### Current behavior

Over a subquery, `map/2` and the `select: [:field, ...]` list form return values that skipped the
loader: a `binary_id` comes back as a raw 16-byte binary and an `Ecto.Enum` as a string. The SQL is
identical to the working forms, so it fails silently.

The list form is wrong twice, because it also loses its container. [The docs for `select/3`](https://hexdocs.pm/ecto/Ecto.Query.html#select/3) state that `select: [:name]` is equivalent to `select: struct(city, [:name])`, and off a schema source it does return a struct. Off a subquery it returns a plain map.

```elixir
Mix.install([{:ecto_sql, "~> 3.14"}, {:postgrex, ">= 0.0.0"}])

defmodule Repo do
  use Ecto.Repo, otp_app: :repro, adapter: Ecto.Adapters.Postgres
end

defmodule Post do
  use Ecto.Schema

  @primary_key {:id, :binary_id, autogenerate: true}
  schema "posts" do
    field :status, Ecto.Enum, values: [:draft, :published]
  end
end

defmodule Repro do
  import Ecto.Query

  def run do
    fields = [:id, :status]
    posts = from post in Post

    show = fn label, query -> IO.puts("#{label} #{inspect(Repo.one!(query))}") end

    show.("subquery, struct(row, ^fields)", from(row in subquery(posts), select: struct(row, ^fields)))
    show.("subquery, map(row, ^fields)   ", from(row in subquery(posts), select: map(row, ^fields)))
    show.("subquery, ^fields             ", from(row in subquery(posts), select: ^fields))
    show.("schema source, map(p, ^fields)", from(post in Post, select: map(post, ^fields)))
    show.("schema source, ^fields        ", from(post in Post, select: ^fields))
  end
end

Application.put_env(:repro, Repo,
  database: "ecto_map_subquery_repro",
  username: "postgres",
  hostname: "localhost",
  pool_size: 2,
  log: false
)

_ = Repo.__adapter__().storage_down(Repo.config())
:ok = Repo.__adapter__().storage_up(Repo.config())
{:ok, _} = Repo.start_link()

Repo.query!("CREATE TABLE posts (id uuid PRIMARY KEY, status text NOT NULL)")
Repo.query!("INSERT INTO posts (id, status) VALUES (gen_random_uuid(), 'published')")

Repro.run()
```

```
subquery, struct(row, ^fields) %Post{__meta__: #Ecto.Schema.Metadata<:loaded>, id: "690ab8f7-1c70-40c6-87bf-6cb0801b78ee", status: :published}
subquery, map(row, ^fields)    %{id: <<105, 10, 184, 247, 28, 112, 64, 198, 135, 191, 108, 176, 128, 27, 120, 238>>, status: "published"}
subquery, ^fields              %{id: <<105, 10, 184, 247, 28, 112, 64, 198, 135, 191, 108, 176, 128, 27, 120, 238>>, status: "published"}
schema source, map(p, ^fields) %{id: "690ab8f7-1c70-40c6-87bf-6cb0801b78ee", status: :published}
schema source, ^fields         %Post{__meta__: #Ecto.Schema.Metadata<:loaded, "posts">, id: "690ab8f7-1c70-40c6-87bf-6cb0801b78ee", status: :published}
```

### Expected behavior

Both forms load the selected fields with the types the subquery's select carries, the way `struct/2` and plain `row.field` already do, and `select: [...]` returns a struct.

greg-rychlewski:
Closes https://github.com/elixir-ecto/ecto/issues/4776

### DIFF
--- a/test/ecto/query/subquery_test.exs
+++ b/test/ecto/query/subquery_test.exs
@@ -440,7 +440,7 @@ defmodule Ecto.Query.SubqueryTest do
       query = normalize(from(c in Comment, join: p in subquery(subquery), on: true, select: p.title))
       assert query.select.fields == [{{:., [type: :string], [{:&, [], [1]}, :title]}, [], []}]
 
-      subquery = from p in Post, select: %{id: p.id, title: p.title}
+      subquery = from p in Post, select: struct(p, [:id, :title])
       query = normalize(from(p in subquery(subquery), select: [:title]))
       assert [{{:., _, [{:&, [], [0]}, :title]}, [], []}] = query.select.fields
 
--- a/test/ecto/query/subquery_test.exs
+++ b/test/ecto/query/subquery_test.exs
@@ -440,7 +440,7 @@ defmodule Ecto.Query.SubqueryTest do
       query = normalize(from(c in Comment, join: p in subquery(subquery), on: true, select: p.title))
       assert query.select.fields == [{{:., [type: :string], [{:&, [], [1]}, :title]}, [], []}]
 
-      subquery = from p in Post, select: struct(p, [:id, :title])
+      subquery = from p in Post, select: %{id: p.id, title: p.title}
       query = normalize(from(p in subquery(subquery), select: [:title]))
       assert [{{:., _, [{:&, [], [0]}, :title]}, [], []}] = query.select.fields
 
--- a/lib/ecto/query/builder/select.ex
+++ b/lib/ecto/query/builder/select.ex
@@ -36,7 +36,7 @@ defmodule Ecto.Query.Builder.Select do
       take?(other) ->
         {
           {:{}, [], [:&, [], [0]]},
-          {[], %{take: %{0 => {:any, Macro.expand(other, env)}}, subqueries: [], aliases: %{}}}
+          {[], %{take: %{0 => {:struct, Macro.expand(other, env)}}, subqueries: [], aliases: %{}}}
         }
 
       maybe_take?(other) ->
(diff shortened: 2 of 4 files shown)


greg-rychlewski:
ah there is some inspect stuff and old tests i need to fix up

josevalim (approved):

test/ecto/query/subquery_test.exs
```diff
@@ -440,7 +440,7 @@ defmodule Ecto.Query.SubqueryTest do
       query = normalize(from(c in Comment, join: p in subquery(subquery), on: true, select: p.title))
       assert query.select.fields == [{{:., [type: :string], [{:&, [], [1]}, :title]}, [], []}]
 
-      subquery = from p in Post, select: %{id: p.id, title: p.title}
+      subquery = from p in Post, select: struct(p, [:id, :title])
```
greg-rychlewski: @josevalim The fix was surfacing an error in this test. Basically it fails now because `select: [atom()]` is considered the same as `struct/2` more strictly.

Just want to make sure this is still the right interpretation of the list of atoms. It's in the docs but not sure if it's out of date

> It is also possible to select a struct and limit the returned fields at the same time:
> 
> from(City, select: [:name])
> 
> The syntax above is equivalent to:
> 
> from(city in City, select: struct(city, [:name]))

josevalim: Yeah, `select: [:foo, :bar]` keeps whatever the from clause is. Structs if there are source+schema, maps if we only have the source.

greg-rychlewski: ah ok I believe what I just pushed now is the correct solution then. sorry for the back and forth. would you mind taking a look once more?

josevalim (approved):

PR #4781, opened 2026-08-12
add map/1 to query api

greg-rychlewski:
Trying again to introduce `map/1`. I think I simplified to a good place in my head but let me try to explain my reasoning. For general strategy I think we have two options:

1. Alter the AST somehow to mark it as a map and then use that marked up AST to remove the schema in the planner's `source_take!`
2. Introduce something like `take: :all` and keep the AST the same

The reason I chose (1) is because in general I think we want to treat this like a source and only care about the map part during post processing. So as much as we can keep the same the better. Also dealing with `take: :all` during `select_merge` adds some complexity

Now given I chose (1) I saw two ways to change the AST:

1. Change the meta
2. Keep it as `{:map, [], {:&, [], [ix]}}`. 

This PR does (1) but I actually started with (2) and then I noticed it was a bit cumbersome dealing with the different pattern matching in `collect_fields` and some of the subquery stuff. It's not impossible but it is more than what this current change is doing.

Let me know what you think!

### DIFF
--- a/lib/ecto/query/api.ex
+++ b/lib/ecto/query/api.ex
@@ -829,6 +829,25 @@ defmodule Ecto.Query.API do
   def map(source, fields), do: doc!([source, fields])
 
   @doc """
+  Used in `select` to return a source as a map.
+
+  Ordinarly, selecting a source will return the result as a struct:
+
+      # Returns [%Post{}, ...]
+      Repo.all(from p in Post, select: p)
+
+  There may be situations where you would like to return the result as a map instead.
+  For example, if you would like to immediately encode your results to JSON without
+  needing to handle the Ecto specific metadata fields.
+
+  To this end, you can use `map/1` in the following way:
+
+      # Returns [%{...}, ...] with all of the query fields in the schema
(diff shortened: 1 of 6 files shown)


greg-rychlewski:
Now I am unsure actually. This has a pretty bad limitation...you cannot do the `map/1` equivalent of this association handling:

```elixir
from(city in City, preload: :country,
     select: %{city | country: map(country: [:id, :population])})
```

Or similarly couldn't do it with virtual fields. 

It might be surprising as well that it's only the query fields in the resulting map. I am pretty sour on this at the moment but will try to think on it some more.

greg-rychlewski:
I have one idea...right now this change is converting the post processing schema to `nil`:

```elixir
schema = if map_source?, do: nil, else: schema
        {{:source, {source, schema}, prefix || query.prefix, types}, 
```
This could be changed so that it keeps the schema but then records it's suppose to be a map in the end. And then in the actual post processing step we can keep whichever keys we want. Maybe we want everything except for the metadata fields.

Maybe something like `{{:map_source, {source, schema}, prefix || query.prefix, types}`

PR #4785, opened 2026-08-21
Sort escaped map/struct fields for deterministic compile output

asweet-confluent:
Several builder escape functions re-embed a compile-time-computed map/struct as a literal AST via bare `Map.to_list/1`. `Map.to_list/1`'s field order isn't guaranteed stable across separate compiler invocations, so two builds of identical source can produce different compiled bytecode whenever any of these get folded into a literal at compile time.


josevalim:

PR #4790, opened 2026-09-06
Fix `exists?` semantics for query combinations

### LINKED ISSUE #4789: `Repo.exists?` changes the membership semantics of `EXCEPT` and `INTERSECT`
### Elixir version

any

### Database and Version

any

### Ecto Versions

3.14

any

### Current behavior

`exists?` replaces the select with `SELECT 1`. For set difference and intersection, the selected values determine membership. Replacing distinct values with the same constant leads to invalid returned value

Repro:
Insert two Rows with ids 1 and 2, then
```elixir
a = from r in Row, where: r.id == 1, select: r.id
b = from r in Row, where: r.id == 2, select: r.id

Repo.all(except(a, ^b))              # [1] - correct
Repo.exists?(except(a, ^b))      # false - wrong, should be true
Repo.all(intersect(a, ^b))          # [] - correct
Repo.exists?(intersect(a, ^b))  # true - wrong, should be false
```

Both `except_all` and `intersect_all` are also affected

Example invalid rendered SQL
```sql
SELECT 1 FROM "rows" AS f0 WHERE (f0."id" = 1)
EXCEPT (SELECT 1 FROM "rows" AS f0 WHERE (f0."id" = 2)) LIMIT 1
```

### Expected behavior

Correct semantics and query results

lukaszsamson:
This PR addresses invalid query involving set membership rewrite to `SELECT 1` by preserving the query form and wrapping it in subquery

Fixes #4789

### DIFF
--- a/lib/ecto/repo/queryable.ex
+++ b/lib/ecto/repo/queryable.ex
@@ -140,12 +140,12 @@ defmodule Ecto.Repo.Queryable do
 
   def exists?(name, queryable, opts) do
     queryable =
-      Query.exclude(queryable, :select)
-      |> Query.exclude(:preload)
+      Query.exclude(queryable, :preload)
+      |> rewrite_combinations()
+      |> Query.exclude(:select)
       |> Query.exclude(:distinct)
       |> Query.select(1)
       |> Query.limit(1)
-      |> rewrite_combinations()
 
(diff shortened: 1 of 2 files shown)


josevalim:

PR #4791, opened 2026-09-11
Update filter/2 docs to note limited database support

ColaCheng:
The previous note said FILTER was Postgres-only; SQLite has supported it since 3.30.


greg-rychlewski:
Thanks for the PR. It might make more sense to make the statement more general so that we don’t risk repeatedly going out of date. Something like not all databases support this operation please check your database documentation.

ColaCheng:
> Thanks for the PR. It might make more sense to make the statement more general so that we don’t risk repeatedly going out of date. Something like not all databases support this operation please check your database documentation.

Thank you for the suggestion! Updated it. Please take a look!

warmwaffles:
Can confirm, in `ecto_sqlite3` `FILTER` is supported.

greg-rychlewski (approved):

greg-rychlewski:
thanks very much!

